PR #4673, opened 2021-02-01
[circleci] Add tag of latest to orc8r dockers

AyliD:
Signed-off-by: Ayelet Regev Dabah <ayelet.regev@gmail.com>

[circleci] Add tag of latest to orc8r dockers

## Summary

for xwf team to be able to retrieve latest docker hash without slack dependency

## Test Plan

circleci

## Additional Information

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -544,6 +544,10 @@ jobs:
       - tag-push-docker:
           project: orc8r
           images: "nginx|controller"
+      - tag-push-docker:
+          project: orc8r
+          images: "nginx|controller"
+          tag-latest: true
       - persist-githash-version:
           file_prefix: orc8r
       - notify-magma:


.circleci/config.yml
```diff
@@ -544,6 +544,10 @@ jobs:
       - tag-push-docker:
           project: orc8r
           images: "nginx|controller"
+      - tag-push-docker:
+          project: orc8r
+          images: "nginx|controller"
+          tag: "latest"
```
tmdzk: Let's use `tag-latest: true`

Cause if we do `tag: "latest"` it will only tag the binaries with latest and not the commit hash, if we pass the above parameter it will double tag it with `latest` and `commit hash`

tmdzk (changes requested):

tmdzk (approved):

PR #4677, opened 2021-02-01
[AGW][MME] Remove stale eNB state with stateless MME

### LINKED ISSUE #4189: [Stateless AGW] [Sctpd] Test GRPC failures in Stateless mode
Need to validate that an Sctpd client write failing towards the MME doesn't cause the sctp socket to close as well. 
For example, in stateless AGW mode, the GRPC from Sctpd to MME will fail while the MME is restarting/upgrading. In this case, the GRPC failure towards MME should not close the Sctp connection with eNB.

ssanadhya:
## Summary

In stateless mode, MME could be down while Sctpd gets an SCTP shutdown from eNB. In such case, the eNB state is never cleaned at MME. This change cleans up the stale eNB state, when a new S1Setup from the same eNB ID comes.  

## Test Plan

- Regression testing: `make integ_test`
- Functional testing: Added new test case and verified that num_enb_connected is zero at the end of the test.


rdefosse:
MAGMA-OAI-TESTS

ulaskozat (approved):
LGTM

ardzoht (approved):

Issue #4536, opened 2021-01-23
[Orc8r] Fix metric scaling issues on AGW


karthiksubraveti:

karthiksubraveti:
closed-by ##4686

PR #4686, opened 2021-02-02
[magmad][metrics] Update metrics reporting to sync once per gw service

mpgermano:

## Summary

This PR corrects an issue we have seen in live deployments. Currently, 
metrics upload can start failing due to gRPC message size being larger
than the max that the channel has been configured with. While we
have a manual configuration for the max message size, this will not
scale well for large deployments.

This PR updates the metrics upload by sending a metrics upload once
per gateway service. Hence, each gateway has it's own metrics queue
and sync cycle. This should allow metrics to scale much better in the future
without hitting the max gRPC message size. 

## Test Plan

Ensure metrics upload works locally with gw and orc8r.
Simulate connection failure between gw and orc8r and
ensure that retry queue logic works appropriately.

### DIFF
--- a/orc8r/gateway/python/magma/magmad/metrics_collector.py
+++ b/orc8r/gateway/python/magma/magmad/metrics_collector.py
@@ -13,9 +13,11 @@ limitations under the License.
 import asyncio
 import calendar
 import logging
+import math
 import prometheus_client
 from prometheus_client.parser import text_string_to_metric_families
 import requests
+import sys
 import time
 from typing import Callable, List, Optional, Dict, NamedTuple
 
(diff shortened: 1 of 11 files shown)


orc8r/gateway/python/magma/magmad/metrics_collector.py
```diff
@@ -93,8 +98,10 @@ def sync(self):
                 # If we throw an exception here, we'll have no idea whether
                 # something was postprocessed or not, so I guess try and make it
                 # idempotent?  #m sevchicken
-                self.post_processing_fn(self._samples)
-            samples = self._retry_queue + self._samples
+                self.post_processing_fn(self._samples_for_service[service_name])
+            samples = self._retry_queue_for_service[service_name] + \
```
karthiksubraveti: Given that every service might have non uniform number of metrics, It would be good to do some batching of metrics per service name as well.

orc8r/gateway/python/magma/magmad/metrics_collector.py
```diff
@@ -167,7 +174,22 @@ def _add_uptime_metric(self, service_name, family):
...
+
+    def _make_sync_done_func(self, service_name, chunk):
+        return lambda future: self._loop.call_soon_threadsafe(
+                               self.sync_done, service_name, chunk,
+                               future)
+
+    def _chunk_samples(self, samples):
+        sample_size_bytes = sys.getsizeof(samples)
+        buckets = math.ceil(
+          sample_size_bytes / self.grpc_max_msg_size_bytes)
```
karthiksubraveti, line 188: minor nit: should we leave any space for any additional overhead ? can we typically send grpc_max_msg_size_bytes size of msg info without any issues ?

orc8r/gateway/python/magma/magmad/metrics_collector.py
```diff
@@ -93,37 +97,40 @@ def sync(self):
...
-                                         self.sync_done, samples, future))
-            self._retry_queue.clear()
-            self._samples.clear()
-        self._loop.call_later(self.sync_interval, self.sync)
-
-    def sync_done(self, samples, collect_future):
+                self.post_processing_fn(self._samples_for_service[service_name])
+
+            samples = self._samples_for_service[service_name]
+            sample_chunks = self._chunk_samples(samples)
```
karthiksubraveti, line 103: minor nit: might be more efficient if chunk_samples returned an iterator.

karthiksubraveti (approved):
Changes LGTM. Left a few comments.

karthiksubraveti:
closes #4536

PR #4695, opened 2021-02-02
[lte][agw] Mitigation for timed out S6a GRPC calls

ulaskozat:
## Summary

- To mitigate GRPC timeouts during authentication step:
-- increased the RPC timeout from 3 seconds to 10 seconds.
-- added a 2nd retry for updating db with random backoff. 
- To allow number of workers configurable via service yml files, I removed the hardwired value for number of workers passed during service initialization. 

## Test Plan

S1AP tests.
Spirent scaling tests.

## Additional Information


uri200:
Hi @ulaskozat  with a 10 second timeout and one retry we may have a potential 20s delay. By that time, will not the UE quit and do an attach again? Is it possible we end up with 2 autentication request for the same UE?

ulaskozat:
@uri200 please comment on https://github.com/magma/magma/pull/4698 instead. Retry is not happening at GRPC side, it is on the call to db itself. 10 second is a hard limit.

PR #4696, opened 2021-02-02
AGW: pipelined: Increase OVS inactivity_probe

pshelar:
Default inactivity_probe is 5 sec, it is increased to 300. Default does not
work in some cases were pipelined is busy with init activity. Increasing this
value fixes the issue.

## Additional Information

### DIFF
--- a/lte/gateway/deploy/roles/magma/files/systemd/magma_pipelined.service
+++ b/lte/gateway/deploy/roles/magma/files/systemd/magma_pipelined.service
@@ -25,6 +25,7 @@ EnvironmentFile=/etc/environment
 ExecStartPre=/usr/bin/ovs-vsctl set bridge gtp_br0 protocols=OpenFlow10,OpenFlow13,OpenFlow14 other-config:disable-in-band=true
 ExecStartPre=/usr/bin/ovs-vsctl set-controller gtp_br0 tcp:127.0.0.1:6633 tcp:127.0.0.1:6654
 ExecStartPre=/usr/bin/ovs-vsctl set-fail-mode gtp_br0 secure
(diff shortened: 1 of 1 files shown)


lte/gateway/deploy/roles/magma/files/systemd/magma_pipelined.service
```diff
@@ -25,6 +25,7 @@ EnvironmentFile=/etc/environment
 ExecStartPre=/usr/bin/ovs-vsctl set bridge gtp_br0 protocols=OpenFlow10,OpenFlow13,OpenFlow14 other-config:disable-in-band=true
 ExecStartPre=/usr/bin/ovs-vsctl set-controller gtp_br0 tcp:127.0.0.1:6633 tcp:127.0.0.1:6654
 ExecStartPre=/usr/bin/ovs-vsctl set-fail-mode gtp_br0 secure
+ExecStartPre=/bin/bash -c 'for id in `sudo /usr/bin/ovsdb-client dump Controller _uuid|tail -n +4` ; do  sudo /usr/bin/ovs-vsctl set Controller $id inactivity_probe=300  ; done'
```
koolzz, line 28: Will controller be initialized when we run this? 
Does it get added when we run this line?
```
ExecStartPre=/usr/bin/ovs-vsctl set-controller gtp_br0 tcp:127.0.0.1:6633 tcp:127.0.0.1:6654
```

pshelar: This is added to OVSDB, so controller does not need to be initialized for this setting. Even in case of race this will take effect in second iteration of reconnect, so it would be ok.

koolzz (approved):

ardzoht (approved):

PR #4698, opened 2021-02-03
[lte][agw]Mitigate GRPC timeouts in S6a requests

ulaskozat:
## Summary

To mitigate GRPC timeouts during authentication step:
-- increased the RPC timeout from 3 seconds to 10 seconds. This is the hard time out.
-- implemented file level sharding of subscribers into 100 buckets based on their last 2 digits in IMSI
-- open/close connections per api call.

To allow number of workers configurable via service yml files, I removed the hardwired value for number of workers passed during service initialization.

## Test Plan

- integ tests
- spirent scaling test

## Additional Information

### DIFF
--- a/lte/gateway/c/oai/lib/s6a_proxy/S6aClient.h
+++ b/lte/gateway/c/oai/lib/s6a_proxy/S6aClient.h
@@ -83,7 +83,7 @@ class S6aClient : public GRPCReceiver {
   S6aClient();
   static S6aClient& get_instance();
   std::unique_ptr<feg::S6aProxy::Stub> stub_;
-  static const uint32_t RESPONSE_TIMEOUT = 3;  // seconds
+  static const uint32_t RESPONSE_TIMEOUT = 10;  // seconds
 };
 
 // There are 3 services which can handle authentication:
--- a/lte/gateway/configs/subscriberdb.yml
+++ b/lte/gateway/configs/subscriberdb.yml
@@ -20,7 +20,7 @@ host_address: 0.0.0.0 # Bind to all interfaces
 enable_streaming: True
 
 # Keep subscriberdb in a file.
-# Comment out db_path to default to :memory: to store subscribers in memory.
+# If this is omitted default path of /var/opt/magma/ will be assumed.
 db_path: /var/opt/magma/
 
 # Bucketize subscribers based on last sid_last_n digits
--- a/lte/gateway/configs/subscriberdb.yml
+++ b/lte/gateway/configs/subscriberdb.yml
@@ -43,3 +43,6 @@ relay_retry_count: 0
 # shall set to false if intend to use Free-Diameter for s6a interface
 # TODO replace boolean with list of protocols: "s6a_over_grpc", "s6a_over_fd"
 s6a_over_grpc: True
+
+# Number of grpc workers
+grpc_workers: 1
--- a/lte/gateway/configs/subscriberdb.yml
+++ b/lte/gateway/configs/subscriberdb.yml
@@ -20,8 +20,11 @@ host_address: 0.0.0.0 # Bind to all interfaces
 enable_streaming: True
 
 # Keep subscriberdb in a file.
-# Change filename to :memory: to store subscribers in memory.
-db_path: file:/var/opt/magma/subscriber.db?cache=shared
+# Comment out db_path to default to :memory: to store subscribers in memory.
+db_path: /var/opt/magma/
+
+# Bucketize subscribers based on last sid_last_n digits
+sid_last_n: 2
 
 # S6A Peer Configurations
 mme_host_name: hss.magma.com
--- a/lte/gateway/configs/subscriberdb.yml
+++ b/lte/gateway/configs/subscriberdb.yml
@@ -45,4 +45,4 @@ relay_retry_count: 0
 s6a_over_grpc: True
 
 # Number of grpc workers
-grpc_workers: 1
+grpc_workers: 10
--- a/lte/gateway/configs/subscriberdb.yml
+++ b/lte/gateway/configs/subscriberdb.yml
@@ -47,5 +47,5 @@ relay_retry_count: 0
 # TODO replace boolean with list of protocols: "s6a_over_grpc", "s6a_over_fd"
 s6a_over_grpc: True
 
-# Number of grpc workers
+# Number of thread workers for the gRPC server
 grpc_workers: 10
--- a/lte/gateway/python/magma/subscriberdb/main.py
+++ b/lte/gateway/python/magma/subscriberdb/main.py
@@ -29,8 +29,7 @@ from lte.protos.mconfig import mconfigs_pb2
 def main():
     """ main() for subscriberdb """
     service = MagmaService('subscriberdb',
-                           mconfigs_pb2.SubscriberDB(),
-                           workers=1)
+                           mconfigs_pb2.SubscriberDB())
 
     # Initialize a store to keep all subscriber data.
     store = SqliteStore(service.config['db_path'], loop=service.loop)
--- a/lte/gateway/python/magma/subscriberdb/main.py
+++ b/lte/gateway/python/magma/subscriberdb/main.py
@@ -28,8 +28,7 @@ from lte.protos.mconfig import mconfigs_pb2
 
 def main():
     """ main() for subscriberdb """
-    service = MagmaService('subscriberdb',
-                           mconfigs_pb2.SubscriberDB())
+    service = MagmaService('subscriberdb', mconfigs_pb2.SubscriberDB())
 
     # Initialize a store to keep all subscriber data.
     store = SqliteStore(service.config['db_path'], loop=service.loop)
--- a/lte/gateway/python/magma/subscriberdb/main.py
+++ b/lte/gateway/python/magma/subscriberdb/main.py
@@ -31,7 +31,8 @@ def main():
     service = MagmaService('subscriberdb', mconfigs_pb2.SubscriberDB())
 
     # Initialize a store to keep all subscriber data.
-    store = SqliteStore(service.config['db_path'], loop=service.loop)
+    store = SqliteStore(service.config['db_path'], loop=service.loop,
+                        sid_digits=service.config['sid_last_n'])
 
     # Initialize the processor
     processor = Processor(store,
--- a/lte/gateway/python/magma/subscriberdb/store/sqlite.py
+++ b/lte/gateway/python/magma/subscriberdb/store/sqlite.py
@@ -11,8 +11,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import logging
+import random
 import sqlite3
 import threading
+import time
 from contextlib import contextmanager
 
 from lte.protos.subscriberdb_pb2 import SubscriberData
@@ -25,6 +28,9 @@ from .base import (
 )
 from .onready import OnDataReady
 
+EXECUTE_TIMEOUT = 5 # wait period in sec if db is locked
+RAND_UPPERBOUND = 250 # maximum delta wait in 2nd attempt (unnormalized)
+RAND_LOWERBOUND = 10 # minimum delta wait after failed first attempt (unnormalized)
 
 class SqliteStore(BaseStore):
     """
@@ -46,7 +52,7 @@ class SqliteStore(BaseStore):
         Returns a thread local connection to the sqlite db.
         """
         if not getattr(self._tlocal, 'conn', None):
-            self._tlocal.conn = sqlite3.connect(self._db_location, uri=True)
+            self._tlocal.conn = sqlite3.connect(self._db_location, timeout=EXECUTE_TIMEOUT, uri=True)
         return self._tlocal.conn
 
     def _create_store(self):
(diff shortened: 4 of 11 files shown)


lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -90,10 +96,19 @@ def edit_subscriber(self, subscriber_id):
             subscriber_data.ParseFromString(row[0])
             yield subscriber_data
             data_str = subscriber_data.SerializeToString()
-            self.conn.execute(
-                "UPDATE subscriberdb SET data = ? " "WHERE subscriber_id = ?",
-                (data_str, subscriber_id),
-            )
+            try:
+                self.conn.execute(
```
karthiksubraveti: Can you use this as a context manager https://docs.python.org/2/library/sqlite3.html#using-the-connection-as-a-context-manager

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -90,10 +96,19 @@ def edit_subscriber(self, subscriber_id):
...
+                    "UPDATE subscriberdb SET data = ? " "WHERE subscriber_id = ?",
+                    (data_str, subscriber_id),
+                )
+            except sqlite3.OperationalError:
+                logging.error("sqlite db error for subscriber: %s", subscriber_id)
+                time.sleep(random.randint(RAND_LOWERBOUND, RAND_UPPERBOUND)/100)
+                # Try one more time after random wait
+                self.conn.execute(
```
karthiksubraveti: What happens if you hit exception here ? Can we perhaps wrap this in a async block ? Does this failure trigger a failure at the caller ?

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -90,10 +96,19 @@ def edit_subscriber(self, subscriber_id):
             subscriber_data.ParseFromString(row[0])
             yield subscriber_data
             data_str = subscriber_data.SerializeToString()
-            self.conn.execute(
-                "UPDATE subscriberdb SET data = ? " "WHERE subscriber_id = ?",
-                (data_str, subscriber_id),
-            )
+            try:
```
shaddi: If we're going to do retries, let's put this in a loop w/ some max retry limit -- until we actually fix the DB lock problem then it's not clear to me that two retries is going to cut it.

shaddi (commented):
Like the direction you're going with this but suggest we do this in a loop rather than a one-shot retry until we fix the underlying issue.

lte/gateway/python/magma/subscriberdb/main.py
```diff
@@ -29,8 +29,7 @@
 def main():
     """ main() for subscriberdb """
     service = MagmaService('subscriberdb',
-                           mconfigs_pb2.SubscriberDB(),
-                           workers=1)
+                           mconfigs_pb2.SubscriberDB())
```
talkhasib: nit: single line

lte/gateway/python/magma/subscriberdb/store/sqlite.py
ulaskozat: The second attempt is really about dithering (i.e., to break synchronization in a burst of requests). Since UE will timeout on its end if we do more backoffs, it has diminishing benefit (or no real benefit).

lte/gateway/python/magma/subscriberdb/store/sqlite.py
ulaskozat: it already is.

lte/gateway/python/magma/subscriberdb/store/sqlite.py
ulaskozat: It is caught and prints a trace in syslog.

rdefosse:
MAGMA-OAI-TESTS

rdefosse:
@ulaskozat 

Running also on our pipeline

rdefosse:
MAGMA-OAI-TESTS

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -25,6 +26,8 @@
 )
 from .onready import OnDataReady
 
+SID_DIGITS = 2 # number of least significant digits from IMSI to use as table buckets
+N_SHARDS = 10 ** SID_DIGITS # number of shards to distribute UEs
```
talkhasib: I suggest moving these to yml config file. Or at least the SID_DIGITS.

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -35,93 +38,131 @@ class SqliteStore(BaseStore):
...
-    def conn(self):
-        """
-        Returns a thread local connection to the sqlite db.
-        """
-        if not getattr(self._tlocal, 'conn', None):
-            self._tlocal.conn = sqlite3.connect(self._db_location, uri=True)
-        return self._tlocal.conn
+
+    def _create_db_locations(self, db_location):
+        # db_location in subscriberdb.yml: file:/var/opt/magma/subscriber.db?cache=shared
```
talkhasib: The yml config is confusing
Suggest changing this config to reflect that it is no longer a single file.
Maybe, change it to specify the folder

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -35,93 +38,131 @@ class SqliteStore(BaseStore):
...
+        # no sharding if it is not a file
+        if i == -1:
+            return [db_location]
+
+        # file name is passed, use it as a base
+        for n_shard in range(N_SHARDS):
+            if n_shard < 10: # prefix i with '0'
+                db_location_list.append(db_location[:i] + '0' + str(n_shard) + db_location[i:])
+            else:
+                db_location_list.append(db_location[:i] + str(n_shard) + db_location[i:])
```
talkhasib: Maybe you can use "{:02d}".format(n_shard)

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -35,93 +38,131 @@ class SqliteStore(BaseStore):
     """
 
     def __init__(self, db_location, loop=None):
-        self._db_location = db_location
-        self._tlocal = threading.local()
+        self._db_location = self._create_db_locations(db_location)
```
talkhasib: s/self._db_location/self._db_locations/g

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -35,93 +38,131 @@ class SqliteStore(BaseStore):
...
-            self.conn.execute("CREATE TABLE IF NOT EXISTS subscriberdb"
-                              "(subscriber_id text PRIMARY KEY, data text)")
+        for db_location in self._db_location:
+            conn = sqlite3.connect(db_location, uri=True)
+            try:
+                with conn:
+                    conn.execute("CREATE TABLE IF NOT EXISTS subscriberdb"
+                                      "(subscriber_id text PRIMARY KEY, data text)")
+            finally:
+                conn.close()
```
talkhasib: I see that we don't maintain the connection to db and open a new connection every time we need to access. Does this affect performance?

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -35,93 +38,131 @@ class SqliteStore(BaseStore):
...
         data_str = subscriber_data.SerializeToString()
-        with self.conn:
-            res = self.conn.execute("SELECT data FROM subscriberdb WHERE "
-                                    "subscriber_id = ?", (sid, ))
-            if res.fetchone():
-                raise DuplicateSubscriberError(sid)
-
-            self.conn.execute("INSERT INTO subscriberdb(subscriber_id, data) "
-                              "VALUES (?, ?)", (sid, data_str))
+        db_location = self._db_location[int(sid[-SID_DIGITS:])]
```
talkhasib: Check sid size before access.
Also, how does int() behave if the string has non digits chars
same for other methods below

I suggest moving this part to a function that returns the location or connection as it is commonly used

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -164,26 +217,35 @@ def resync(self, subscribers):
...
+                    # Clear all subscribers
+                    conn.execute("DELETE FROM subscriberdb")
+
+                    # Add the subscribers with the current state
+                    for sid in bucket_subs[i]:
+                        if sid in current_state:
+                            sub.state.CopyFrom(current_state[sid])
+                            data_str = sub.SerializeToString()
+                            conn.execute("INSERT INTO subscriberdb(subscriber_id, data) "
+                                         "VALUES (?, ?)", (sid, data_str))
```
talkhasib: is indentation correct here??
You are changing behavior

lte/gateway/python/magma/subscriberdb/store/sqlite.py
ulaskozat: I have not seen a meaningful performance hit on that. I erred on the site of simplicity and not keep track of left over connections.

lte/gateway/python/magma/subscriberdb/store/sqlite.py
ulaskozat: make sense. I will specify the path in the subscriberdb.yml. if the file path is not included, it will default to in-memory db.

lte/gateway/python/magma/subscriberdb/store/sqlite.py
ulaskozat: I am moving SID_DIGITS into yml config. N_SHARDS is dependent on it, I will just create object fields to store them (and set to a default value in the init call if absent).

lte/gateway/python/magma/subscriberdb/store/sqlite.py
ulaskozat: actually I will eliminate this, it is a residual from earlier commits where I wasn't keeping a list...

lte/gateway/python/magma/subscriberdb/store/sqlite.py
ulaskozat: good catch! indentation is wrong and there are more problems with it...

ulaskozat: on line 223 I should have stored the subs and on 241 items should be subs not sid, etc...

lte/gateway/python/magma/subscriberdb/store/sqlite.py
ulaskozat: will do...

rdefosse:
MAGMA-OAI-TESTS

lte/gateway/c/oai/lib/s6a_proxy/S6aClient.h
```diff
@@ -83,7 +83,7 @@ class S6aClient : public GRPCReceiver {
   S6aClient();
   static S6aClient& get_instance();
   std::unique_ptr<feg::S6aProxy::Stub> stub_;
-  static const uint32_t RESPONSE_TIMEOUT = 3;  // seconds
+  static const uint32_t RESPONSE_TIMEOUT = 10;  // seconds
```
ssanadhya, line 86: Considering the UE will re-try Attach in around 15 seconds, if it does not receive an Attach Accept, shouldn't we fail faster?

ulaskozat: @ssanadhya the rationale for this is that if there is a db lock issue and if it is a transient thing to see large latency in these calls, the rest of the procedures would still have meaningful time budget to complete the remaining steps (e.g., the remaining ~5sec). If this is not a transient thing and timeouts still happen, then prolonging the timeout can help push back the attach request of UEs relieving the flash crowds a bit. Does this make sense?

ssanadhya: Makes sense! So in case of a severe failure, the UEs will be stalled for 10s before re-trying an attach, slowing down the request flow into subscriberdb, correct?

ssanadhya (approved):
Changes in `S6aClient.h` look good to me!

lte/gateway/c/oai/lib/s6a_proxy/S6aClient.h
ulaskozat: yes...

mpgermano (approved):

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -206,3 +265,16 @@ def _update_apn(self, apn_config, apn_data):
...
         apn_config.ambr.max_bandwidth_dl = apn_data.ambr.max_bandwidth_dl
+
+
+    def _sid2bucket(self, subscriber_id):
+        """
+        Maps Subscriber ID to bucket
+        """
+        try:
+            bucket = int(subscriber_id[-self._sid_digits:])
+        except:
```
ardzoht:
```suggestion
        except (TypeError, ValueError):
```

lte/gateway/configs/subscriberdb.yml
```diff
@@ -43,3 +46,6 @@ relay_retry_count: 0
 # shall set to false if intend to use Free-Diameter for s6a interface
 # TODO replace boolean with list of protocols: "s6a_over_grpc", "s6a_over_fd"
 s6a_over_grpc: True
+
+# Number of grpc workers
```
ardzoht:
```suggestion
# Number of thread workers for the gRPC server
```

ardzoht (approved):
minor nits, overall LGTM

rdefosse:
MAGMA-OAI-TESTS

PR #4700, opened 2021-02-03
[s1ap] Add logging messages to show content of directoryd

uri200:

## Summary

**NOTE this is a temporal change to get some logs on CI runtime**

This PR adds some code to try to find out the error on CI you can see in the link below.

We some messages and a retry mechanism in case of failure 

The idea is not to use the retry mechanism in the future, but trying to figure out why sessionid is not found.

https://app.circleci.com/pipelines/github/magma/magma/13765/workflows/2e8cedc3-0366-42e3-bc4f-9d5bb1b41240/jobs/123496

## Test Plan

```
/magma/lte/gateway/python/integ_tests$ make integ_test TESTS=s1aptests/test_attach_detach_rar_tcp_data.py
...
Ran 1 test in 29.702s

OK
sleep 1

```

## Additional Information

### DIFF
--- a/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
+++ b/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
@@ -60,6 +60,8 @@ from lte.protos.ha_service_pb2 import (
     StartAgwOffloadRequest,
     EnbOffloadType,
 )
+from orc8r.protos.common_pb2 import Void
+
 DEFAULT_GRPC_TIMEOUT = 10
 
 
@@ -159,7 +161,7 @@ class S1ApUtil(object):
         with self._cond:
             rc = self._test_api(cmd_type.value, c_req)
             if rc:
-                logging.error("Error executing command %s" % repr(cmd_type))
+                print("Error executing command %s" % repr(cmd_type))
                 return rc
         return 0
 
@@ -1260,18 +1262,35 @@ class SessionManagerUtil(object):
         qos = QoSInformation(qci=qos["qci"])
 
         # Get sessionid
-        req = GetDirectoryFieldRequest(id=imsi, field_key="session_id")
-        try:
-            res = self._directorydstub.GetDirectoryField(
-                req, DEFAULT_GRPC_TIMEOUT
-            )
-        except grpc.RpcError as err:
-            logging.error(
-                "GetDirectoryFieldRequest error for id: %s! [%s] %s",
-                imsi,
-                err.code(),
-                err.details(),
-            )
+        #TODO: remove retries
+        i = 0
+        MAX = 3
+        res = None
+        while i < MAX:
+            req = GetDirectoryFieldRequest(id=imsi, field_key="session_id")
+            try:
+                res = self._directorydstub.GetDirectoryField(
+                    req, DEFAULT_GRPC_TIMEOUT
+                )
+            except grpc.RpcError as err:
+                print(
+                    "error: GetDirectoryFieldRequest error for id: %s! [%s] %s",
+                    imsi,
+                    err.code(),
+                    err.details(),
+                )
+            if req != None:
+                i = MAX
+            else:
+                i+=1
+                print("warning: directoryd failed to return sessionId for %s. Retrying", imsi)
+                time.sleep(3)
+
+        if res == None:
+            print("error: Couldnt find sessionid. Directoryd content:")
+            allRecordsResponse = self._directorydstub.GetAllDirectoryRecords(Void(), DEFAULT_GRPC_TIMEOUT)
+            for record in allRecordsResponse.recordsResponse:
+                print("%s", str(record))
 
         self._session_stub.PolicyReAuth(
             PolicyReAuthRequest(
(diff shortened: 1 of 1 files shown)


lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -1260,18 +1262,34 @@ def create_ReAuthRequest(self, imsi, policy_id, flow_list, qos):
...
+                )
+            except grpc.RpcError as err:
+                print(
+                    "error: GetDirectoryFieldRequest error for id: %s! [%s] %s",
+                    imsi,
+                    err.code(),
+                    err.details(),
+                )
+            if req != None:
+                i = MAX
```
ulaskozat, line 1283: We don't need to check this. If you really want to it is better done just after line 1269 and use break. Also should not this be `req == None`?

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -1260,18 +1262,34 @@ def create_ReAuthRequest(self, imsi, policy_id, flow_list, qos):
...
+            else:
```
ulaskozat, line 1284: eliminate else

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -1260,18 +1262,34 @@ def create_ReAuthRequest(self, imsi, policy_id, flow_list, qos):
...
+                i+=1
+                print("warning: directoryd failed to return sessionId for %s. Retrying", imsi)
+                time.sleep(3)
+
+        if req == None:
+            print("error: Couldnt find sessionid. Directoryd content:")
+            ret = self._directorydstub.GetAllDirectoryRecords(Void(), DEFAULT_GRPC_TIMEOUT)
+            for record in ret.records:
+                print("%s", str(record))
```
ulaskozat: why do we want to clutter the test output like this?

ulaskozat:
If this is intended for a temporary use, let's not land it...

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
uri200: this is just temporal
This is only failing on CI and right now we were not getting any message. 
Once i get the information it will be gone

themarwhal (approved):

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
uri200: This is a temporal function. Unfortunately the issue only happens on CI.

Once it is fixed i will remove it

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
uri200: Hey @ulaskozat  this code is temporal. Right now we are not getting any message when it fails. This is only failing on CI. So i want to get all the content possible

The logic with this retries is: if there is a request we exit the loop making the i=MAX. But if `req == None` we increase `i` and retry again (up to 3 times)

This should be gone once I see its results on CI

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -1260,18 +1262,35 @@ def create_ReAuthRequest(self, imsi, policy_id, flow_list, qos):
...
+                    imsi,
+                    err.code(),
+                    err.details(),
+                )
+            if req != None:
+                i = MAX
+            else:
+                i+=1
+                print("warning: directoryd failed to return sessionId for %s. Retrying", imsi)
+                time.sleep(3)
```
ssanadhya, line 1287: @uri200 , I believe @ulaskozat is suggesting the following change, which simplifies the indentation. It does not change the logic, but makes it more readable

```suggestion
            if req != None:
               break
           
            i+=1
            print("Warning: directoryd failed to return sessionId for %s. Retrying", imsi)
            time.sleep(3)
```

ssanadhya (approved):
Minor nit, rest looks good!

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -1260,18 +1262,35 @@ def create_ReAuthRequest(self, imsi, policy_id, flow_list, qos):
...
+                    req, DEFAULT_GRPC_TIMEOUT
+                )
+            except grpc.RpcError as err:
+                print(
+                    "error: GetDirectoryFieldRequest error for id: %s! [%s] %s",
+                    imsi,
+                    err.code(),
+                    err.details(),
+                )
+            if req != None:
```
ulaskozat, line 1282: why is this still req, but not res??

uri200: for the record after our conversation, this has been fixed at 
https://github.com/magma/magma/pull/4744

Issue #4630, opened 2021-01-28
[nms] Alert receivers reported not working


andreilee:
### Your Environment

- **Version: master
- **Affected Component:** NMS
- **Affected Subcomponent:** Alerts Receivers
- **Deployment Environment:** e.g. AWS

### Describe the Issue

Set up alerts receivers are reported not working. Requires additional investigation.

**Expected behavior**
N/A

**Screenshots**
N/A

**Additional context**
N/A

PR #4742, opened 2021-02-03
[helm][tf] Backport alert fixes1.4

Scott8440:
## Summary
Adds fixes to alertmanager configurer to 1.4.

## Test Plan

Deploy following 1.4 instructions

### DIFF
--- a/docs/readmes/orc8r/upgrade_1_4.md
+++ b/docs/readmes/orc8r/upgrade_1_4.md
@@ -57,14 +57,14 @@ module orc8r {
 module orc8r-app {
   source = "github.com/magma/magma//orc8r/cloud/deploy/terraform/orc8r-helm-aws?ref=v1.4"
   # ...
-  orc8r_chart_version   = "1.5.10"
+  orc8r_chart_version   = "1.5.11"
   orc8r_tag             = "MAGMA_TAG"       # from build step, e.g. v1.4.0
   orc8r_deployment_type = DEPLOYMENT_TYPE   # valid options are: ["fwa", "federated_fwa", "all"]
 }
 ```
 
 Set `cluster_version` to the Kubernetes version found during the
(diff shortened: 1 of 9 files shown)


karthiksubraveti (approved):
Changes LGTM

orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
```diff
@@ -378,3 +378,20 @@ variable "analytics_category_name" {
   type = string
   default = "magma"
 }
+
```
mpgermano, line 399: In this file there is an `orc8r_chart_version` variable set to 1.5.10. Can you update this to 1.5.11. Also update the `upgrade_1_4.md` at `magma/docs/orc8r/` to reference 1.5.11 instead of 1.5.10.

mpgermano (changes requested):
I don't think we should be merging directly to the v1.4 branch. Does this change already exist on master? If so, just find the old PR and apply the `apply-v1.4` label. We have a script to backport all labeled PRs. If not, then just set the target to master and add the label

docs/readmes/orc8r/upgrade_1_4.md
```diff
@@ -57,7 +57,7 @@ module orc8r {
 module orc8r-app {
   source = "github.com/magma/magma//orc8r/cloud/deploy/terraform/orc8r-helm-aws?ref=v1.4"
   # ...
-  orc8r_chart_version   = "1.5.10"
+  orc8r_chart_version   = "1.5.11"
```
mpgermano, line 60: There's a reference to 1.5.10 below a few lines. Can you update this as well.

mpgermano (approved):

karthiksubraveti:
Fixes https://github.com/magma/magma/issues/4630

PR #4744, opened 2021-02-03
[s1ap] Remove loop on ReAuthRequest

uri200:

## Summary

This PR removes a retries loop on `ReAuthRequest` but adds a new print message showing the content of directoryd in case directory doesn't find the session id

## Test Plan

make integ_test

## Additional Information


ssanadhya (approved):
Thanks for the cleanup!

Issue #4654, opened 2021-01-29
Address lgtm.com potential slice out-of-bounds accesses


electronjoe:
@bbarritt kicked off a test run of lgtm.com for the Magma code base, see [results](https://lgtm.com/projects/g/magma/magma?mode=list&severity=error).

The tooling detected several "Off-by-one index comparison against length may lead to out-of-bounds". E.g. such that a malformed SMS-CP packet could cause an out-of-bounds access.  This Issue will look at the following analysis outputs claiming off-by-one indexing:

* [lte/cloud/go/sms_ll/sms_cp.go](https://lgtm.com/projects/g/magma/magma/snapshot/74a1db2c7ab9b948e1cccad6776140dad3744f18/files/lte/cloud/go/sms_ll/sms_cp.go#xf829089322a7014f:1)
* [feg/radius/src/modules/eap/packet/packet.go](https://lgtm.com/projects/g/magma/magma/snapshot/74a1db2c7ab9b948e1cccad6776140dad3744f18/files/feg/radius/src/modules/eap/packet/packet.go#x2e7be3788c40c64c:1)
* [feg/gateway/services/eap/attribute.go](https://lgtm.com/projects/g/magma/magma/snapshot/74a1db2c7ab9b948e1cccad6776140dad3744f18/files/feg/gateway/services/eap/attribute.go#x94ce2637181819ff:1)

electronjoe:
addressed by PR #4626

PR #4745, opened 2021-02-03
[feg][radius] Fix potential out-of-bounds access in EAP packet

electronjoe:
## Summary

This PR addresses one instance covered in Issue #4654. Previously, malformed EAP packets could cause an out-of-bounds access. This PR addresses this issue by returning an error under such circumstances.

### New Tests

Augmented existing unit tests to include off-nominal binary packet representations with incorrect size for their EAP packet type.

### Existing Presubmit Process

Confirmed success of:

> cd feg/radius/src/
> ./run.sh make
> ./run.sh test

### DIFF
--- a/feg/radius/src/modules/eap/packet/packet.go
+++ b/feg/radius/src/modules/eap/packet/packet.go
@@ -23,9 +23,10 @@ import (
 
 // EAP constants
 const (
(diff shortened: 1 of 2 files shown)


electronjoe:
Addressed by @hcgatewood

PR #4750, opened 2021-02-03
[orc8r] Poll Spec Servicers

AndyLKhuu:

## Summary
This PR is one of the remaining changes needed to convert our merging of swagger specs into runtime. Specifically, this PR implements a polling function which reaches out to services with registered spec servicers and then calling upon them to retrieve the comprehensive list of swagger specs to process.

image
[Project Wise Diagram: Changes from current PR outlined in blue] 

Notable Changes in PR 
- [x] Implementation of polling function to retrieve and generate combined swagger spec
- [x] Test Suite dedicated to sanity checking polling function
- [x] Addition of doc.go for package
- [x] Slight refactoring of existing tests in package   

## Test Plan

## Additional Information


AndyLKhuu:
A naming concern I was having was in regards to the naming of swagger related packages. Right now we have two packages named swagger. 1.) a library dealing with purely the combination aspect of specs in orc8r/cloud/go/swagger and 2.) the working directory of this PR. Currently, an easy fix I have is just aliasing the swagger library used within the working directory as “swagger_lib” but a more concrete fix might be too just rename the package to something else? @hcgatewood thoughts?

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -26,14 +26,11 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-var (
```
hcgatewood, line 29: :+1:

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -26,14 +26,11 @@ import (
...
-	invalidPath      = "invalidPath"
-	testFile         = "test.swagger.v1.yml"
-	testFileContents = "test yaml spec"
-	tmpDir           = "/etc/magma/configs/orc8r/swagger_specs/"
-)
-
 func TestSpecServicer_NewSpecServicerFromFile(t *testing.T) {
+	testFile := "test.swagger.v1.yml"
+	testFileContents := "test yaml spec"
+	tmpDir := "/etc/magma/configs/orc8r/swagger_specs/"
```
hcgatewood: Tiny nit: leave off trailing slash

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -52,16 +49,18 @@ func TestSpecServicer_NewSpecServicerFromFile(t *testing.T) {
 	res, err := servicer.GetSpec(context.Background(), req)
 	assert.NoError(t, err)
 
-	assert.Equal(t, res.SwaggerSpec, testFileContents)
+	assert.Equal(t, testFileContents, res.SwaggerSpec)
```
hcgatewood, line 52: :+1:

orc8r/cloud/go/obsidian/swagger/registry.go
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+package swagger
+
+import (
+	"magma/orc8r/cloud/go/orc8r"
+	"magma/orc8r/lib/go/registry"
+)
+
+// GetSpecServicers returns all the registered Swagger spec servicers
```
hcgatewood: Nit: remove the "the"

orc8r/cloud/go/obsidian/swagger/registry.go
```diff
@@ -0,0 +1,33 @@
...
+func GetSpecServicers() ([]remoteSpec, error) {
+	services, err := registry.FindServices(orc8r.SpecServicerLabel)
+	if err != nil {
+		return []remoteSpec{}, err
```
hcgatewood: Return nil for empty slice / slice zero value

orc8r/cloud/go/obsidian/swagger/doc.go
```diff
@@ -0,0 +1,19 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package swagger
+
+// Package swagger provides a simple implementation of the per service spec servicer
```
hcgatewood: Wrap doc lines to 79 chars

hcgatewood: For doc.go files -- use multiline comment style, and place the comment directly above the `package` line. The latter makes sure the godoc tool will recognize the comment as a package doc

hcgatewood: Nit: "per-service"

orc8r/cloud/go/obsidian/swagger/doc.go
```diff
@@ -0,0 +1,19 @@
...
+// as well as the interface needed to poll every servicer for their respective
+// Swagger specs. This package is meant to serve as the entry point of injecting
```
hcgatewood: "This package serves as the entry point for requresting and combinging Swagger specs at runtime."

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/golang/glog"
+)
+
+var (
+	commonSpecPath = "/etc/magma/configs/orc8r/swagger_specs/common/swagger-common.yml"
+)
+
+// GetCombinedSwaggerSpecs polls every servicer registered with
+// a Swagger spec and merges them together to return a combined spec.
+func GetCombinedSwaggerSpecs() (string, error) {
```
hcgatewood: Package name is `swagger`, so you don't need to include `Swagger` in this function name

hcgatewood: "GetCombinedSpec" (singular)

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,65 @@
...
+	servicers, err := GetSpecServicers()
+	if err != nil {
+		return "", err
+	}
+
+	// Retrieve common spec
```
hcgatewood: These types of "topic sentence" comments are a bad code smell. Each code paragraph should be skimmable for what it does. If it's not skimmable, that's a good sign you should break it out into a separate private function. Then the name of the private function is (a) documentation of what that section does and (b) easier to skim within the main function

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,65 @@
...
+	data, err := ioutil.ReadFile(commonSpecPath)
+	if err != nil {
+		glog.Fatalf("Error retrieving common Swagger spec %+v", err)
```
hcgatewood: `Fatal` should be used with extreme care, and only as a last resort. That is, the service should be effectively unable to function when you're fatal-ing

In this case, this function will be called from the obsidian service. We don't want to kill *all* HTTP access to the orc8r just because a Swagger spec file is missing (e.g. moved)

So just return the error

orc8r/cloud/go/obsidian/swagger/registry.go
```diff
@@ -0,0 +1,33 @@
...
+
+package swagger
+
+import (
+	"magma/orc8r/cloud/go/orc8r"
+	"magma/orc8r/lib/go/registry"
+)
+
+// GetSpecServicers returns all the registered Swagger spec servicers
+func GetSpecServicers() ([]remoteSpec, error) {
```
hcgatewood: Okay so I just re-read through your `remoteSpec` definition. Two small changes:
- make remoteSpec type public (`RemoteSpec`). Since we're going to be using that type outside the package, we want the type to be public
- add a `GetService() string` method to return the name of the service pointed to

The other existing service mesh examples of this registry pattern return an interface type. But that's because (a) we want to be able to easily mock those objects cause they're more complex and (b) we were refactoring from local objects to remote objects. In this case, (a) we don't need to mock the behavior cause it's simple enough to spin up the backing service and (b) the plugin architecture is dead

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	yamlCommon := string(data)
+
+	// Retrieve specs
+	var yamlSpecs []string
+	for _, s := range servicers {
+		yamlSpec, err := s.GetSpec()
+		if err != nil {
+			glog.Errorf("Invalid response from spec servicer \n %+v", err)
```
hcgatewood:
```
err = errors.Wrapf(err, "get Swagger spec from %s service", s.GetService())
glog.Error(err)
```

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,65 @@
...
+		} else {
+			yamlSpecs = append(yamlSpecs, yamlSpec)
+		}
+	}
+
+	// Combine specs.
+	combined, warnings, err := swagger_lib.Combine(yamlCommon, yamlSpecs)
+	if err != nil {
+		glog.Fatal(err)
```
hcgatewood: Same as above: just return the err

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	if warnings != nil {
+		glog.Infof("Warnings: %+v \n", warnings)
```
hcgatewood: "Warnings" isn't enough context if you see this in a log file -- what's actually happened?

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	// Merge specs
```
hcgatewood: This comment is unnecessary

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"magma/orc8r/cloud/go/obsidian/swagger"
+	"magma/orc8r/cloud/go/obsidian/swagger/protos"
+	"magma/orc8r/cloud/go/orc8r"
+	"magma/orc8r/cloud/go/test_utils"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_GetCombinedSwaggerSpecs(t *testing.T) {
```
hcgatewood: This test doesn't need to create any files. In fact, don't even use any golden files. Just create super-short swagger specs inline in this file. We're not testing whether combine "correctly" combines in this test, we're testing the e2e datapath is correctly hooked up

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,97 @@
...
+	assert.NoError(t, err)
+
+	data, err := ioutil.ReadFile(goldenFilePath)
+	assert.NoError(t, err)
+	expected := string(data)
+
+	assert.Equal(t, expected, combined)
+}
+
+func StartTestServiceInternal(
```
hcgatewood: This function is over-engineered. Just call it `func setup(t *testing.T)`
- Doesn't need to be public
- No need to pass in the service names (now that we're not storing config files at all for this test)

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,97 @@
...
+	t *testing.T,
+	serviceOne string,
+	serviceTwo string,
+	serviceThree string,
+) {
+	labels := map[string]string{}
+
+	labels[orc8r.SpecServicerLabel] = "true"
```
hcgatewood: Coalesce this into the previous line -- you can just create the labels map and set this key-value pair in a single declaration

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	srv1, lis1 := test_utils.NewTestOrchestratorService(t, orc8r.ModuleName, "test_service1", labels, nil)
```
hcgatewood: Up to you: I prefer to zero-index my test variable names, so e.g. `services[0] = srv0`, but some people prefer to one-index

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,97 @@
...
+	srv2, lis2 := test_utils.NewTestOrchestratorService(t, orc8r.ModuleName, "test_service2", labels, nil)
+	srv3, lis3 := test_utils.NewTestOrchestratorService(t, orc8r.ModuleName, "test_service3", labels, nil)
+
+	protos.RegisterSwaggerSpecServer(srv1.GrpcServer, swagger.NewSpecServicerFromFile(serviceOne))
+	protos.RegisterSwaggerSpecServer(srv2.GrpcServer, swagger.NewSpecServicerFromFile(serviceTwo))
+	protos.RegisterSwaggerSpecServer(srv3.GrpcServer, swagger.NewSpecServicerFromFile(serviceThree))
+
+	go srv1.RunTest(lis1)
+	go srv2.RunTest(lis2)
+	go srv3.RunTest(lis3)
```
hcgatewood: This declare+register+run block can occur within a loop over the service names, rather than as an unwound loop

hcgatewood (changes requested):
Couple changes. Mainly: you're fatal-ing too aggressively, and for the combine_test we don't need to do anything with actual swagger spec files

hcgatewood:
> A naming concern I was having was in regards to the naming of swagger related packages. Right now we have two packages named swagger. 1.) a library dealing with purely the combination aspect of specs in orc8r/cloud/go/swagger and 2.) the working directory of this PR. Currently, an easy fix I have is just aliasing the swagger library used within the working directory as “swagger_lib” but a more concrete fix might be too just rename the package to something else? @hcgatewood thoughts?

An immediate solution would be to move everything to orc8r/cloud/go/obsidian/swagger. We don't/likely won't use Swagger outside the context of obsidian, so I think it's reasonable to move everything to that pkg

You might have to rename some of the files to make that happen, but I think it's the cleanest solution

karthiksubraveti (commented):
Went through the changes. Changes LGTM.

orc8r/cloud/go/obsidian/swagger/combine.go
AndyLKhuu: Got it, I just copied the name from the project spec but I agree this is more concise given the context.

orc8r/cloud/go/obsidian/swagger/combine.go
AndyLKhuu: Got it, thanks for the explanation. I was thinking the comments looked kind of off given how I haven't seen similarly pattern comments in the codebase but I just left it in as a gamble.

orc8r/cloud/go/obsidian/swagger/combine_test.go
AndyLKhuu: Got it, I feel that it's been more natural for me to stick with one indexing these types of vars so unless you have a strong opinion otherwise, I'll probably keep it this way.

orc8r/cloud/go/obsidian/swagger/combine_test.go
AndyLKhuu: Having inline swagger specs work as the test case was actually the first thing I considered and I actually talked this over with Michael a bit as well. The main reason why I decided to use the golden files pattern was because I found it to be actually quite "ugly" having swagger specs declared in-line, mainly just due to the fact that the minimum boilerplate tags needed (magma-gen-meta, info, base-path, etc) already takes up about 6-7 lines. Coupled with the fact that we want to at least have one unique tag per service as a general sanity test + common spec, I thought it would optimal to just stick with the golden files pattern that we have used in the swagger lib file. Note that the test files I have right now are already extremely minimal(just one unique tag each). I think the big benefit with the golden file patterns is actually in mind the simplicity? I do see how having specs inline may actually be beneficial in the sense that we want to avoid IO operations as much as possible so that we minimize any potential conflicts when running concurrent tests. Besides what I listed above, are there any main benefits that I'm missing with transitioning to in-line specs? @hcgatewood

AndyLKhuu: Update: I just realized that the IO benefit of the spec inline isn't actually there since we would have to create a temp directory to store our common-spec regardless.

hcgatewood: As a starting point: I/O makes tests messy. Except where we're specifically testing the I/O's correct functioning, we want to avoid including I/O in our tests where achieving this goal doesn't come with unreasonable trade-offs

In this case, I think there's a clean way to have this test do zero I/O

1. Handling the inline specs

A. The magma-gen-meta is for model generation rather than API specification, so it can be discarded from these toy specs. (In fact, a service should be able to send a generic Swagger spec and everything would work fine.) B. If you still dislike the length of the multiline strings (I'm assuming these strings will be less than 10 lines each), create the per-service specs as spec.Swagger objects first, then serialize to string before passing to the servicer constructor

2. Handling the common file I/O

Let's rework the poll+combine command slightly to handle this elegantly. A. The `GetCombinedSpec` method can become `GetCombinedSpec(yamlCommon string) (string, error)` and B. add a method `GetCommonSpec() (string, error)`. We'll still need an I/O test for `GetCommonSpec`, but this will allow our test for `GetCombinedSpec` to be I/O-free. (It will also allow an optimization where the obsidian service can read the common spec once, during initialization, instead of re-reading it on every call.)

Notice how this rearranges things such that we're testing the I/O function in isolation, while the larger business logic function an be tested as a pure function. This is a healthy testing pattern

orc8r/cloud/go/obsidian/swagger/combine.go
AndyLKhuu: Got it, I was basing this off how it was being done in the executable currently but I realize the environment in which error logs may show up means that a runtime warning should be more descriptive for the person parsing through logs.

AndyLKhuu:
> > A naming concern I was having was in regards to the naming of swagger related packages. Right now we have two packages named swagger. 1.) a library dealing with purely the combination aspect of specs in orc8r/cloud/go/swagger and 2.) the working directory of this PR. Currently, an easy fix I have is just aliasing the swagger library used within the working directory as “swagger_lib” but a more concrete fix might be too just rename the package to something else? @hcgatewood thoughts?
> 
> An immediate solution would be to move everything to orc8r/cloud/go/obsidian/swagger. We don't/likely won't use Swagger outside the context of obsidian, so I think it's reasonable to move everything to that pkg
> 
> You might have to rename some of the files to make that happen, but I think it's the cleanest solution

AndyLKhuu:
> > A naming concern I was having was in regards to the naming of swagger related packages. Right now we have two packages named swagger. 1.) a library dealing with purely the combination aspect of specs in orc8r/cloud/go/swagger and 2.) the working directory of this PR. Currently, an easy fix I have is just aliasing the swagger library used within the working directory as “swagger_lib” but a more concrete fix might be too just rename the package to something else? @hcgatewood thoughts?
> 
> An immediate solution would be to move everything to orc8r/cloud/go/obsidian/swagger. We don't/likely won't use Swagger outside the context of obsidian, so I think it's reasonable to move everything to that pkg
> 
> You might have to rename some of the files to make that happen, but I think it's the cleanest solution

I definitely agree that moving the Lib Code into the Obsidian directory might be the cleanest solution. For the sake of ensuring that this PR stays true to it's name and task(not mixing in the migration/refactor), I'm thinking I will just open a quick refactoring PR to do that after this current PR is merged?

hcgatewood:
> > > A naming concern I was having was in regards to the naming of swagger related packages. Right now we have two packages named swagger. 1.) a library dealing with purely the combination aspect of specs in orc8r/cloud/go/swagger and 2.) the working directory of this PR. Currently, an easy fix I have is just aliasing the swagger library used within the working directory as “swagger_lib” but a more concrete fix might be too just rename the package to something else? @hcgatewood thoughts?
> 
> > 
> 
> > An immediate solution would be to move everything to orc8r/cloud/go/obsidian/swagger. We don't/likely won't use Swagger outside the context of obsidian, so I think it's reasonable to move everything to that pkg
> 
> > 
> 
> > You might have to rename some of the files to make that happen, but I think it's the cleanest solution
> 
> 
> 
> I definitely agree that moving the Lib Code into the Obsidian directory might be the cleanest solution. For the sake of ensuring that this PR stays true to it's name and task(not mixing in the migration/refactor), I'm thinking I will just open a quick refactoring PR to do that after this current PR is merged?

Either way is fine

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,66 @@
...
+		yamlSpec, err := s.GetSpec()
+		if err != nil {
+			err = errors.Wrapf(err, "get Swagger spec from %s service", s.GetService())
+			glog.Error(err)
+		} else {
+			yamlSpecs = append(yamlSpecs, yamlSpec)
+		}
+	}
+
+	combined, warnings, err := swagger_lib.Combine(yamlCommon, yamlSpecs)
```
hcgatewood, line 50: Add a test case to handle where len(yamlSpecs) == 0

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return "", err
+	}
+	if warnings != nil {
+		glog.Infof("Swagger spec traits were overwritten or unable to be read: %+v \n", warnings)
```
hcgatewood: Trailing newlines (and trailing spaces) are superfluous for logging strings

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	return combined, nil
+}
+
+// GetCommonSpec returns the YAML string of the Swagger Common Spec
+func GetCommonSpec() (string, error) {
+	data, err := ioutil.ReadFile(commonSpecPath)
+	if err != nil {
+		return "", err
```
hcgatewood, line 66: We always want to wrap errors that come from outside our codebase. Aside from giving context specific to our codebase, it also adds an error stack

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"magma/orc8r/cloud/go/test_utils"
+
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/yaml.v2"
+)
+
+func Test_GetCommonSpec(t *testing.T) {
+	specPath := "/etc/magma/configs/orc8r/swagger_specs"
+	commonSpecDir := "/etc/magma/configs/orc8r/swagger_specs/common"
+	commonSpecFilePath := filepath.Join(commonSpecDir, "swagger-common.yml")
```
hcgatewood: Just manually construct this as a literal -- there's no need to use filepath.Join since we're already constructing the commonSpecDir as a literal

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	os.RemoveAll(specPath)
+	defer os.RemoveAll(specPath)
+
+	err := os.MkdirAll(commonSpecDir, os.ModePerm)
+	assert.NoError(t, err)
+
+	commonTag := swagger_lib.TagDefinition{Name: "Tag Common"}
+	commonSpec := swagger_lib.Spec{Tags: []swagger_lib.TagDefinition{commonTag}}
```
hcgatewood: Just construct commonSpec as a single large (multi-line) literal -- no need to separate out commonTag into a separate variable

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,103 @@
...
+	yamlCommon := marshalToYAML(t, commonSpec)
+
+	tags := []swagger_lib.TagDefinition{
+		{Name: "Tag 1"},
+		{Name: "Tag 2"},
+		{Name: "Tag 3"},
+	}
+	services := []string{"test_service1", "test_service2", "test_service3"}
+	assert.Equal(t, len(tags), len(services))
```
hcgatewood: This check is unnecessary. If you really want to perform this type of check, you can do it (and a len check) inside the `setup` function

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	setup(t, tags, services)
+
+	combined, err := swagger.GetCombinedSpec(yamlCommon)
+	assert.NoError(t, err)
+
+	assert.Equal(t, expectedYaml, combined)
+}
+
+func setup(t *testing.T, tags []swagger_lib.TagDefinition, services []string) {
```
hcgatewood: Nit: I like argument order of `t, services, tags`, since you *need* services during setup, but it's conceivable that the tags part could change at some point

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,103 @@
...
+	labels := map[string]string{
+		orc8r.SpecServicerLabel: "true",
+	}
+
+	for i, s := range services {
+		srv, lis := test_utils.NewTestOrchestratorService(t, orc8r.ModuleName, s, labels, nil)
+		tag := tags[i]
+		spec := swagger_lib.Spec{Tags: []swagger_lib.TagDefinition{tag}}
+
+		yamlSpec := marshalToYAML(t, spec)
```
hcgatewood: Coalesce the `tag :=` and `spec :=` lines into this line

orc8r/cloud/go/obsidian/swagger/registry.go
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+package swagger
+
+import (
+	"magma/orc8r/cloud/go/orc8r"
+	"magma/orc8r/lib/go/registry"
+)
+
+// GetSpecServicers returns all registered Swagger spec servicers
```
hcgatewood: End doc comments with a period -- should be full sentences

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -52,16 +49,18 @@ func TestSpecServicer_NewSpecServicerFromFile(t *testing.T) {
...
 }
 
 func TestSpecServicer_GetSpec(t *testing.T) {
```
hcgatewood, line 55: Add a short check for the GetService method

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -26,14 +26,11 @@ import (
...
-	tmpDir           = "/etc/magma/configs/orc8r/swagger_specs/"
-)
-
 func TestSpecServicer_NewSpecServicerFromFile(t *testing.T) {
+	testFile := "test.swagger.v1.yml"
+	testFileContents := "test yaml spec"
+	tmpDir := "/etc/magma/configs/orc8r/swagger_specs"
+
 	os.RemoveAll(tmpDir)
 	defer os.RemoveAll(tmpDir)
```
hcgatewood, line 36: Nit for NewSpecServicerFromFile line: prefer naming things e.g. `"test_spec_servicer"` rather than just `"test"`, as debugging can be a bit confusing when `"test"` is found throughout the debugged object -- i.e. adding a bit more info makes debugging easier

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -26,14 +26,11 @@ import (
 	"github.com/stretchr/testify/assert"
```
hcgatewood, line 26: For the `swagger_protos` import line: here and elsewhere: don't alias imports unless the alias is necessary

*One asterisk to this rule: we have certain packages that we *always* import, like `merrors`. This is a quirk of our codebase, and I have a todo to rename that package to `merrors` to avoid having to do the permanent alias

orc8r/cloud/go/obsidian/swagger/combine.go
hcgatewood: "Some Swagger spec traits..." to emphasize that it's only a partial issue

hcgatewood (changes requested):
Looking good, couple small things then I'll be content

orc8r/cloud/go/obsidian/swagger/remote_spec.go
```diff
@@ -24,17 +24,17 @@
...
-	// service name of the remoteSpec
+type RemoteSpec struct {
+	// service name of the RemoteSpec
 	// should always be lowercase to match service registry convention
 	service string
 }
 
-func NewRemoteSpec(serviceName string) remoteSpec {
-	return remoteSpec{service: strings.ToLower(serviceName)}
+func NewRemoteSpec(serviceName string) RemoteSpec {
```
hcgatewood, line 35: P.S. add doc comments for the public objects/methods in this file

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	err = ioutil.WriteFile(commonSpecFilePath, []byte(yamlCommon), 0644)
+	assert.NoError(t, err)
+
+	actual, err := swagger.GetCommonSpec()
+	assert.NoError(t, err)
+	assert.Equal(t, yamlCommon, actual)
+}
+
+func Test_GetCombinedSwaggerSpecs(t *testing.T) {
```
hcgatewood, line 56: P.P.S. add a test case to handle when there are warnings

orc8r/cloud/go/obsidian/swagger/combine_test.go
AndyLKhuu: I see, my reasoning for adding this check was so that it would be easier for any developer updating the test case in the future to think "oh there must be some reason why" and then resultingly check that we are doing a indexed based for loop which would break if the lengths were unequal. I can see how it's simple enough to not be needed though?

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
AndyLKhuu: Got it, thanks for the catch. Also the highlighting ^ is off again.

orc8r/cloud/go/obsidian/swagger/remote_spec.go
AndyLKhuu: Sounds good, I was thrown off about whether to include it since the interface of mconfig builders didn't comment its public function.

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
AndyLKhuu: GetService is actually a public method of the RemoteSpec object("client" side) and the SpecServicer Object doesn't actually have a GetService method nor do I think it needs one right? We only ever interact directly with the RemoteSpec object in combine.go. 

Update: Discussed Offline to add a registry_test.go

orc8r/cloud/go/obsidian/swagger/combine_test.go
AndyLKhuu: Oh! How come? If we were to do that, we would be kind of testing the "correctness" of the Combine() function and that's not what this test is about right?

AndyLKhuu: I understand the sentiment to add a test case where len(yamlSpecs) == 0 aka no registered Servicers exists but adding a test case to handle for warnings when merging seems to be edging towards testing the "correctness" of combine() rather than the e2e of GetCombinedSwaggerSpec

orc8r/cloud/go/obsidian/swagger/remote_spec.go
```diff
@@ -24,17 +24,20 @@ import (
 	"github.com/golang/glog"
 )
 
-type remoteSpec struct {
-	// service name of the remoteSpec
+// RemoteSpec identifies a remote spec
+type RemoteSpec struct {
```
AndyLKhuu, line 28: @hcgatewood You mentioned in a previous comment that the struct should be made public but since we only ever initialize the object with NewRemoteSpec() we don't necessarily need to have the object be public right? We can still access the objects methods without it the object itself being exported.

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,129 @@
...
+	assert.Equal(t, expectedYaml, combined)
+
+	// Clean up registry
+	registry.RemoveService(services[0])
+	registry.RemoveService(services[1])
+	registry.RemoveService(services[2])
+	registry.RemoveService(serviceDuplicate)
+}
+
+func registerTestSpecServicer(t *testing.T, service string, tag swagger_lib.TagDefinition) {
```
AndyLKhuu: @hcgatewood Any naming suggestions?

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	var yamlSpecs []string
+	for _, s := range servicers {
+		yamlSpec, err := s.GetSpec()
+		if err != nil {
+			err = errors.Wrapf(err, "get Swagger spec from %s service", s.GetService())
+			glog.Error(err)
```
hcgatewood, line 44: Remember to add explanatory comment when we're swallowing errors

Basically, the only time you *don't* need to explain swallowing an error is in a main.go's `main` function

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,129 @@
...
+	for i, s := range services {
+		registerTestSpecServicer(t, s, tags[i])
+	}
+
+	combined, err = swagger.GetCombinedSpec(yamlCommon)
+	assert.NoError(t, err)
+
+	assert.Equal(t, expectedYaml, combined)
+
+	// Success even with merge warnings(duplicate tag)
```
hcgatewood: Missing space before open paren

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,129 @@
...
+	combined, err = swagger.GetCombinedSpec(yamlCommon)
+	assert.NoError(t, err)
+
+	assert.Equal(t, expectedYaml, combined)
+
+	// Clean up registry
+	registry.RemoveService(services[0])
+	registry.RemoveService(services[1])
+	registry.RemoveService(services[2])
+	registry.RemoveService(serviceDuplicate)
```
hcgatewood: Move these lines to directly before you register the first test servicer, and you can replace them with just `defer registry.RemoveServicesWithLabel(orc8r.SpecServicerLabel)`

orc8r/cloud/go/obsidian/swagger/combine_test.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func registerTestSpecServicer(t *testing.T, service string, tag swagger_lib.TagDefinition) {
+	labels := map[string]string{
+		orc8r.SpecServicerLabel: "true",
```
hcgatewood: My preference: for uniformity let's name this orc8r const `SwaggerSpecLabel` -- as in "Swagger specifier" or "has a Swagger specification"

orc8r/cloud/go/obsidian/swagger/combine_test.go
hcgatewood: `registerServicer` is fine -- `Test` is implicit since it's in a test file, and `Spec` is unneeded unless you're also registering other servicers

orc8r/cloud/go/obsidian/swagger/doc.go
```diff
@@ -0,0 +1,20 @@
...
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+/*
+	Package swagger provides a simple implementation of the per-service spec
```
hcgatewood, line 15: I think Codacy wants you to not block-indent the package comment. Try that out and let me know

orc8r/cloud/go/obsidian/swagger/registry_test.go
```diff
@@ -0,0 +1,54 @@
...
+		go srv.RunTest(lis)
+	}
+
+	servicers, err = swagger.GetSpecServicers()
+	assert.NoError(t, err)
+
+	assert.Equal(t, services[0], servicers[0].GetService())
+	assert.Equal(t, services[1], servicers[1].GetService())
+
+	// Clean up registry
```
hcgatewood: Same as previous registry cleanup comment

orc8r/cloud/go/obsidian/swagger/registry_test.go
```diff
@@ -0,0 +1,54 @@
...
+	for _, s := range services {
+		srv, lis := test_utils.NewTestOrchestratorService(t, orc8r.ModuleName, s, labels, nil)
+		go srv.RunTest(lis)
+	}
+
+	servicers, err = swagger.GetSpecServicers()
+	assert.NoError(t, err)
+
+	assert.Equal(t, services[0], servicers[0].GetService())
+	assert.Equal(t, services[1], servicers[1].GetService())
```
hcgatewood: This ordering is not guaranteed. This test needs to be resilient against `servicers` containing the correct values but in different order. Consider creating a map

orc8r/cloud/go/obsidian/swagger/spec_servicer.go
```diff
@@ -46,6 +46,7 @@ func NewSpecServicer(spec string) protos.SwaggerSpecServer {
 	return &specServicer{spec}
 }
 
+// GetSpec returns the associated spec as a YAML string
```
hcgatewood: This comment is superfluous since the method is an impl of the protos.SwaggerSpecServer method. Only include this per-impl comment if it's documenting something idiosyncratic to the particular implementation

orc8r/cloud/go/obsidian/swagger/registry_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRegistry_GetSpecServicers(t *testing.T) {
+	// Success with no registered servicers
+	servicers, err := swagger.GetSpecServicers()
+	assert.NoError(t, err)
+
+	assert.Equal(t, 0, len(servicers))
```
hcgatewood: assert.Empty

orc8r/cloud/go/obsidian/swagger/registry_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"magma/orc8r/cloud/go/obsidian/swagger"
+	"magma/orc8r/cloud/go/orc8r"
+	"magma/orc8r/cloud/go/test_utils"
+	"magma/orc8r/lib/go/registry"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRegistry_GetSpecServicers(t *testing.T) {
+	// Success with no registered servicers
```
hcgatewood, line 28: Test code is the correct place for these topic-sentence comments, they strike a good balance of readability vs. non-rigid test code. Thanks for adding

hcgatewood (changes requested):
Content is great, couple small style items then I'll be ready to merge

I figured you might run into an issue with clearing the local service registry. What did your debugging process look like?

orc8r/cloud/go/obsidian/swagger/spec_servicer.go
AndyLKhuu: Hmm I was thinking the same thing but you had a previous comment saying to add a comment for all public methods and it also seems like the Codacy Check requires it?

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -0,0 +1,69 @@
...
+	servicers, err := GetSpecServicers()
+	if err != nil {
+		return "", err
+	}
+
+	var yamlSpecs []string
+	for _, s := range servicers {
+		yamlSpec, err := s.GetSpec()
+		if err != nil {
+			// Swallowing GetSpec error because the polling should continue
```
hcgatewood: Nit: just "Swallow" is sufficient, the "-ing" indicates it's "currently" happening, which is not strictly true

orc8r/cloud/go/obsidian/swagger/doc.go
```diff
@@ -0,0 +1,20 @@
...
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+/*
+Package swagger provides a simple implementation of the per-service spec
```
hcgatewood: Since we realized Codacy isn't blocking: I actually like including the block indent for multi-line package comments. But I'll leave the decision up to you

orc8r/cloud/go/obsidian/swagger/registry_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	for s := range services {
+		srv, lis := test_utils.NewTestOrchestratorService(t, orc8r.ModuleName, s, labels, nil)
+		go srv.RunTest(lis)
+	}
+
+	servicers, err = swagger.GetSpecServicers()
+	assert.NoError(t, err)
+
+	_, exists := services[servicers[0].GetService()]
```
hcgatewood: These checks are insufficient -- what if both returned servicers were `test_spec_service1`?

I remembered there's a good way to handle this actually: go back to the way you had it before, but use assert.ElementsMatch

hcgatewood (approved):
Mainly just fix that one test then I'll merge

orc8r/cloud/go/obsidian/swagger/registry_test.go
AndyLKhuu: Good point, also thanks for the tip on the cleaner method.

themarwhal (approved):

PR #4776, opened 2021-02-04
AGW: pipelineD: qos: Fix scale test issues

pshelar:
Rather than using async loop initialization is moved to
sync processing. This would avoid any race with policy
installation right after pipelineD boot.
other changes:
1. code readability improvements
2. remove tc filter delete call before create
3. improve error handling
4. add locking to avoid running concurrent TC commands.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/qos/common.py
+++ b/lte/gateway/python/magma/pipelined/qos/common.py
@@ -12,7 +12,6 @@ limitations under the License.
 """
 from typing import List, Dict  # noqa
 
-import asyncio
 import logging
 from enum import Enum
 from lte.protos.policydb_pb2 import FlowMatch
@@ -23,7 +22,7 @@ from magma.pipelined.qos.utils import QosStore
 from magma.configuration.service_configs import load_service_config
 
 LOG = logging.getLogger("pipelined.qos.common")
-#LOG.setLevel(logging.DEBUG)
+# LOG.setLevel(logging.DEBUG)
 
 
 def normalize_imsi(imsi: str) -> str:
@@ -72,11 +71,14 @@ class SubscriberSession(object):
 
 
 class SubscriberState(object):
+    """
+     Map subscriber -> sessions
+    """
     def __init__(self, imsi: str, qos_store : Dict):
         self.imsi = imsi
-        self.rules = {}
-        self.sessions = {}
-        self._qos_store = qos_store
+        self.rules = {}   # rule -> qos handle
+        self.sessions = {}  # IP -> [sessions(ip, qos_argumets, rule_no), ...]
+        self._redis_store = qos_store
 
     def check_empty(self,) -> bool:
         return not self.rules and not self.sessions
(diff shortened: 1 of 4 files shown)


koolzz (approved):

lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -208,67 +210,63 @@ def _setupInternal(self):
...
+            try:
+                # populate state from db
+                in_store_qid = set()
+                purge_store_set = set()
+                for k, v in self._redis_store.items():
+                    if v not in qos_state:
+                        purge_store_set.add(k)
+                        continue
+                    in_store_qid.add(v)
+                    _, imsi, ip_addr, rule_num, d = get_key(k)
```
ardzoht: not related to these changes, can `d` param be more descriptive on what it represents? it's not very clear from the context of the function as of now

lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -208,67 +210,63 @@ def _setupInternal(self):
...
-                    self._clean_restart = True
-                    self.setup()
-
-            asyncio.ensure_future(self.impl.read_all_state(), loop=self._loop).add_done_callback(callback)
+            qos_state = self.impl.read_all_state()
+            try:
+                # populate state from db
+                in_store_qid = set()
+                purge_store_set = set()
+                for k, v in self._redis_store.items():
```
ardzoht: `k` can be `sub_key`, for clarity

lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -208,67 +210,63 @@ def _setupInternal(self):
...
+
+                # purge unreferenced qos configs from system
+                for qos_handle in qos_state:
+                    if qos_handle not in in_store_qid:
+                        LOG.debug("removing qos_handle %d", qos_handle)
+                        self.impl.remove_qos(qos_handle,
+                                             qos_state[qos_handle]['direction'], recovery_mode=True)
+
+                LOG.info("init complete with state recovered successfully")
+            except Exception as e:  # pylint: disable=broad-except
```
ardzoht, line 259: what are the possible errors thrown here? we should aim to be as restrictive as possible while catching exceptions

lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -175,7 +177,7 @@ def debug(cls, _, __, ___):
 
     def redisAvailable(self):
```
ardzoht:
```suggestion
    def is_redis_available(self):
```

ardzoht (changes requested):

lte/gateway/python/magma/pipelined/qos/common.py
pshelar: I would like to keep it as as, since this is initialization and we don't was to get into crash loop due to unknown exception.

ardzoht (approved):

PR #4784, opened 2021-02-05
AGW: pipelined: add missing match fields in diff generator.

pshelar:
Current code is missing some match fields while comparing
new vs existing flows. This can cause diff generator to
drop new required flows from startup flow list.

## Additional Information


koolzz (approved):
Oh nice, thanks for finding this

PR #4790, opened 2021-02-05
Temporarily roll back pipelined polling time for CI

koolzz:

## Additional Information

### DIFF
--- a/lte/gateway/configs/pipelined.yml
+++ b/lte/gateway/configs/pipelined.yml
@@ -67,7 +67,7 @@ meter:
   idle_timeout: 0
 
 enforcement:
-  poll_interval: 10
+  poll_interval: 2
   default_drop_flow_name: 'internal_default_drop_flow_rule'
 
 # Enable polling mobilityd to identify which subscriber sessions need to be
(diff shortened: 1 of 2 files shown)


uri200 (approved):

ardzoht (approved):

uri200:
Thanks @koolzz  

This interval is key for stats. The less frequent pipelined reports stats, the less responsive systems will be (not because pipelined being sessiond clock, but because the information will not be know promptly enough)

Note we have built our system around `2s` interval, our code, tests and knowledge is based on that value. So it would be better than any increase in that specific value were backed up with thorough testing (lte_integ_test, cwf_integ_test and teravm test) and a meaningful reason.

Note accuracy of stats means `$`. Increasing the responsive we create a leak there.

koolzz:
> Thanks @koolzz
> 
> This interval is key for stats. The less frequent pipelined reports stats, the less responsive systems will be (not because pipelined being sessiond clock, but because the information will not be know promptly enough)
> 
> Note we have built our system around `2s` interval, our code, tests and knowledge is based on that value. So it would be better than any increase in that specific value were backed up with thorough testing (lte_integ_test, cwf_integ_test and teravm test) and a meaningful reason.
> 
> Note accuracy of stats means `$`. Increasing the responsive we create a leak there.

Every 2second polling is too heavy for the current system with a large number of UEs. We need to redesign it with per thread polling as we discussed before. I'll check if there is some bandaid fix I can apply

PR #4797, opened 2021-02-06
[NMS] Fix multiple NMS bugs

### LINKED ISSUE #4785: [NMS][1.4]nat_enabled not disabled
[non-nat-enabled-true.har.zip](https://github.com/magma/magma/files/5935262/non-nat-enabled-true.har.zip)

### Your Environment

- **Version:**1.4
- **Affected Component:** NMS
- **Affected Subcomponent:** NMS
- **Deployment Environment:**  AWS (orc8r), bare-metal (AGW)

### Describe the Issue

Adding a new gateway with a non-nat SGi, disabling NAT at the last screen, keeps the NAT feature enabled. This causes mme to crash on the AGW as in the network settings we set ip_allocation to dhcp_broadcast.

**To Reproduce**
1. Add a new gateway in NMS; set NAT to off.
2. Go back to gateway configuration and note NAT configuration for gateway. 
3. Check gateway.mconfig on the AGW and note that `nat_enabled` is set to `true`

**Expected behavior**
See above.

karthiksubraveti:

## Summary

1. EPC configs was being overriden by RAN config component, this patch fixes the issue. This patch also contains gateway config tests which edit configurations across all the config tabs and saves the gateway config. We verify if the edits across all individual config tab is reflected in the final gateway config.
2. This patch also fixes a minor issue with pre defined alerts where gateway checkin alert was missing a "}".
3. Fixes a minor issue where the bulk upload csv file didn't support policies field
4. Also fixes a minor issue with subscribers add dialog, here we update the subscriber state in every loop, setState however is async and when we 
```
 ...
    setSubscriberMap({...subscriberMap, [key]: value}); <==== it doesn't get set immediately. 
````
So we keep overwriting the subscriber map present in the context, when we exit the dialog, since we only have the last one in subscriber map, we don't refresh the subscriber overview immediately. Fixed this by 
- cleaning up subscriber overview, removed grids and used TopBars and CardTitleRow instead
- Triggered setLastRefreshTime from the delete and add dialog handler rather than in useEffect which relied on the number of subscribers in subscriber context to change for triggering refresh.

## Test Plan

- Add an additional unit test to verify gateway config add
- Verified predefined alerts by manually syncing the alerts
-
Screen Shot 2021-02-08 at 5 08 40 AM

## Additional Information

### DIFF
--- a/nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
+++ b/nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
@@ -125,8 +125,42 @@ type SubscriberInfo = {
   state: 'INACTIVE' | 'ACTIVE',
   dataPlan: string,
   apns: Array<string>,
-  policies: Array<string>,
+  policies?: Array<string>,
 };
+
+const SUB_NAME_OFFSET = 0;
+const SUB_IMSI_OFFSET = 1;
+const SUB_AUTH_KEY_OFFSET = 2;
+const SUB_AUTH_OPC_OFFSET = 3;
+const SUB_STATE_OFFSET = 4;
+const SUB_DATAPLAN_OFFSET = 5;
+const SUB_APN_OFFSET = 6;
+const SUB_POLICY_OFFSET = 7;
+const SUB_MAX_FIELDS = 8;
+
+function parseSubscriber(line: string) {
+  const items = line.split(',').map(item => item.trim());
+  if (items.length > SUB_MAX_FIELDS) {
+    throw new Error(
+      `Too many fields to parse, expected ${SUB_MAX_FIELDS} fields, received ${items.length} fields`,
+    );
+  }
+
+  return {
+    name: items[SUB_NAME_OFFSET],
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js [hunk 1]
```diff
@@ -157,19 +167,21 @@ function parseSubscriberFile(fileObj: File) {
...
           subscribers.push({
-            name: items[0],
-            imsi: items[1],
-            authKey: items[2],
-            authOpc: items[3],
-            state: items[4] === 'ACTIVE' ? 'ACTIVE' : 'INACTIVE',
-            dataPlan: items[5],
-            apns: items[6]
-              .split('|')
+            name: items[SUB_NAME_OFFSET],
```
Scott8440: I think it would be cleaner to have a function to parse an individual subscriber and then just call it on all the lines here. You could then pass in the format to that function and have this all be more resilient to change.

Scott8440 (approved):
One nit, everything else lgtm

PR #4598, opened 2021-01-26
[Redis] Add a timeout + renewal logic to Redis Lock

themarwhal:
Signed-off-by: Marie Bremner <marwhal@fb.com>

## Additional Information

- [ ] This change is backwards-breaking

PR #4819, opened 2021-02-09
[Redis] Add a timeout + renewal logic to Redis Lock, and some test refactor

### LINKED ISSUE #4896: [DirectoryD] Add a timeout for locks to prevent stale locks
Version: 1.3.3

In production, we saw that DirectoryD would sometimes get stuck on Redis connections. This was due to some locks being leftover in Redis probably due to an ungraceful service restart.

themarwhal:
## Summary
Two things.
1. Use `python-redis-lock` to support locks with a timeout. (https://pypi.org/project/python-redis-lock/#description) This module adds a small LUA script to enhance the Redis lock to support a timeout behavior. (releases lock in X amount of time with renewal, etc.) We've been seeing a few times in production that an ungraceful service shutdown had left locks in place, leading the entire directoryd service to halt. 
   ```"requires_python": ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*", "size": 12165 ```
2. Refactor all tests that rely on `MockRedis` to use `fakeredis.FakeStringRedis` instead. This is so that we dont have to re-write the Redis functionality. (Change 1 introduced more functionality requirements on `MockRedis`). I also removed a bunch of `@mock.patch("redis.Redis", MockRedis)` that wasn't doing anything. (This dependency is only for the dev setup, so it will not affect production package)
 

## Test Plan
make test
built a package with the new dependency, it builds fine.

## Additional Information


themarwhal:
Previously: https://github.com/magma/magma/pull/4598

ulaskozat (approved):
LGTM

ardzoht (approved):
this is awesome, lgtm

mpgermano (approved):
This is great

PR #4851, opened 2021-02-09
AGW: pipelined: qos: initialize single instance of QoS manager

pshelar:
Gy and Enforcement APP both need QoS manager, create API
to maintain a single instance of the QoS manager object.
This patch also fixes QoS parsing errors in initialization.

## Test Plan
`make test`
Ran integ tests manually on dev.

## Additional Information


lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -207,6 +219,8 @@ def setup(self):
             self._loop.call_later(self._redis_conn_retry_secs, self.setup)
 
     def _setupInternal(self):
+        if self._initialized:
+            return
```
koolzz, line 223: Maybe debug log that we're already initialized

koolzz (approved):

lte/gateway/python/magma/pipelined/qos/common.py
pshelar: I do not sure if this will help.

PR #4855, opened 2021-02-10
[NMS] Run prod version of NMS instead of dev 

karthiksubraveti:

## Additional Information


Scott8440:
I think you need to bump the chart versions and such

Scott8440 (approved):

PR #4857, opened 2021-02-10
Fix envoy_controller init logic

koolzz:

## Additional Information


feg/gateway/services/envoy_controller/control_plane/control_plane.go
```diff
@@ -197,8 +197,6 @@ func GetControllerClient() *ControllerClient {
 	go RunManagementServer(ctx, srv, port)
 	go RunManagementGateway(ctx, srv, gatewayPort)
 
-	<-signal
```
uri200, line 200: what is this channel for? meaning, the cb object is placing something there. What is that

Also note that if the cb object does not place anything there that will be blocked. Maybe it is better to use a timeout to make sure you never block
```
select {
case <-signal:
case <-timeout:
    // log error
}
```

uri200 (approved):

themarwhal (approved):

PR #4769, opened 2021-02-04
Added code to handle duplicate attach requests

rsarwad:
[lte][agw]: Added code to handle duplicate attach requests
## Summary
MME had the support of handling duplicate attach requests as per nas spec; but while handling re-transmitted attach request when attach accept is not sent. MME would initiate implicit detach procedure for previous attach request and simultaneously new attach request would be processed. Implicit detach was leading to release of session at SPGW. On reception of Delete Session Response from SPGW; MME would send UE context release request and delete the UE contexts; So while handling new attach request, at some point MME was not able to fetch UE context, so it was leading un-expected behavior, one such scenario is, on reception of ULA from HSS, mme was not able to fetch UE context; So new attach request would fail.
To address the synchronization issue; code is modified such that new attach request will be processed after receiving the Delete Session Response.

## Test Plan
Executed s1ap sanity test suite
To re-create the duplicate attach scenario, added code to drop first create session response from spgw;
Below is the diff from sgw_handlers.c file to drop first create session response
 322a323 
>   static int drop_first_csr                                     = 0;
414c415,422
<
---
>   if (drop_first_csr == 0) {
>     OAILOG_DEBUG_UE(
>         LOG_SPGW_APP, imsi64,
>         "Rashmi drop first CSRsp for teid " TEID_FMT " \n",
>         create_session_response_p->teid);
>     ++drop_first_csr;
>     OAILOG_FUNC_RETURN(LOG_SPGW_APP, RETURNok);
>   }

Modified code in S1ap tester to send re-transmitted attach with KSI value changed. Below are few lines changed in ue_app.c file
1173    if (ueAppCb->set_ksi == 1) {
1174    /*NAS key set identifier IE*/
1175    attachReq->nasKsi.pres = TRUE;
1176    attachReq->nasKsi.id = CM_EMM_NONAS_KEY_AVAILABLE;
1177    attachReq->nasKsi.tsc = CM_EMM_NASKEYSID_NATIVE_SEC;
1178    }
1179    if (ueAppCb->set_ksi == 0) {
1180      ueAppCb->set_ksi = 1;
1181    }

 Below variable needs to be added UeAppCB in ueAppdbm.x file
 
305 typedef struct _ueAppCb
306 {
307    TskInit          init;
308    Pst              nbPst;
309    Pst              fwPst;
310    U32              trfGenIfAddr; /* IP address of eth port connected to
311                                      traffic generator*/
312    U8               NASProcGuardtimer; /* timer value */
[...]

PR #4860, opened 2021-02-10
Added test case to send second attach request when Attach Accept or Reject is not received

rsarwad:
[lte][agw]: Added test case to send second attach request when Attach Accept or Reject is not received
## Summary
Added this test case to validate the changes done in PR, https://github.com/magma/magma/pull/4769/files


ssanadhya (changes requested):
@rsarwad , this test case would only work with the temporary changes in the Test Plan for https://github.com/magma/magma/pull/4769. Let's attach this test case to the same PR instead of committing it in the repo itself.

ssanadhya:
> @rsarwad , this test case would only work with the temporary changes in the Test Plan for #4769. Let's attach this test case to the same PR instead of committing it in the repo itself.

@rsarwad , could you please upload this test to #4769 and close this PR?

ulaskozat:
can we close this PR?

PR #4881, opened 2021-02-10
[NMS] Fix edit tabs

### LINKED ISSUE #4882: [NMS] Edit tabs not opening on the selected tab
### Describe the Issue

Edit tabs not opening on the selected tab

**To Reproduce**
1. Go to Gateway config view
2. Click on edit RAN, a modal opens on the RAN tab
3. Close the modal
4. Click again on edit RAN, a modal opens on Config tab (first tab)
5. We can reproduce this for every config view (Enodebs, Subscribers)

**Expected behavior**
Click on edit specific config should open the associated tab

HannaFar:

## Summary

Fix edit tabs. We were always setting edit tab to 0 (first tab) when we closed edit modal.

## Additional Information

### DIFF
--- a/nms/app/packages/magmalte/app/views/equipment/EnodebDetailConfigEdit.js
+++ b/nms/app/packages/magmalte/app/views/equipment/EnodebDetailConfigEdit.js
@@ -161,7 +161,6 @@ function EnodeEditDialog(props: DialogProps) {
   const onClose = () => {
     // clear existing state
     setEnb({});
-    setTabPos(0);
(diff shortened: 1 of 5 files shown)


karthiksubraveti (commented):
One bug which might still be lurking is that, when we try to open the create dialog, it might open at a different tabPos, since that state isn't reset, I wonder if we should useEffect and set the tabPos when editProps changes.

karthiksubraveti (approved):

PR #4895, opened 2021-02-10
[orc8r] Expose the combine Swagger spec process to the obsidian service

AndyLKhuu:

## Summary

This PR is one of the conclusive PRs in the series of changes needed to migrate our build time dependency on merging swagger specs into runtime. Specifically this PR exposes the combined Swagger Spec to our obsidian services by adding middleware to create the combined Swagger Spec on page refresh. 

image
(The features in this PR are outlined in blue)

Notable Features Added

* Feature toggle to enable combine of Swagger specs at runtime
* Middleware that combines Swagger Specs

========================

An **alternative design** discussed involved adding a caching mechanism in which we periodically call our GenerateSwaggerSpec handler at a fixed interval and then comparing it with our cached spec. If the new spec and cached spec differed, only then would we write the new spec to file. This design was avoided mainly due to these reasons 

* There is negligible performance gains from this design
* Regardless of the caching, the user will still have to refresh the page to see changes
* More complex than needed since we are under the assumption that the page refresh rate is low. 

## Test Plan
- [x] Manual inspection of logs and UI

## Additional Information


hcgatewood (changes requested):
Good start, but not quite the way we want to handle this

For starters, you're attaching `GenerateSwaggerSpec` as middleware (rather than a route handler), meaning it's being called on *every* REST API request, regardless of whether the request is for Swagger-related resources. Swagger-related requests will be a tiny minority of total requests, so this implementation effectively adds a full over-the-network round-trip plus contention on writing to a single file for *every* request

What we want to do instead is more similar to what I saw when you were walking me through your code: attach a dynamic handler via `e.GET` to forward to the relevant Swagger-generating method.

The only consideration here is whether the dynamic route at `.../apidocs/swagger.yml` (or whatever it is) will play nicely with the static route at `.../apidocs/`. You may have to play around with the order in which you attach those routes. If you can't get them to play nicely, you'll have to move the swagger.yml location (for both static and dynamic cases) outside the catch-all static resource directory

.
.
.

The other consideration: we should be able to accomplish the dynamic route *without* needing to read/write to any files (except perhaps the common spec file, but that should be handled by the GetCommonSpec method, or whatever it was named, that you added previously)

AndyLKhuu:
> Good start, but not quite the way we want to handle this
> 
> For starters, you're attaching `GenerateSwaggerSpec` as middleware (rather than a route handler), meaning it's being called on _every_ REST API request, regardless of whether the request is for Swagger-related resources. Swagger-related requests will be a tiny minority of total requests, so this implementation effectively adds a full over-the-network round-trip plus contention on writing to a single file for _every_ request
> 
> What we want to do instead is more similar to what I saw when you were walking me through your code: attach a dynamic handler via `e.GET` to forward to the relevant Swagger-generating method.
> 
> The only consideration here is whether the dynamic route at `.../apidocs/swagger.yml` (or whatever it is) will play nicely with the static route at `.../apidocs/`. You may have to play around with the order in which you attach those routes. If you can't get them to play nicely, you'll have to move the swagger.yml location (for both static and dynamic cases) outside the catch-all static resource directory
> 
> .
> .
> .
> 
> The other consideration: we should be able to accomplish the dynamic route _without_ needing to read/write to any files (except perhaps the common spec file, but that should be handled by the GetCommonSpec method, or whatever it was named, that you added previously)

Yeah the routing with GET didn't play nicely as you described so that's why I settled for using it as middleware. I see the drawbacks of it now though. Changing the swagger.yml location may solve the previous issue. Will keep you updated.

orc8r/cloud/go/obsidian/tests/unit_test_utils.go
```diff
@@ -110,6 +111,18 @@ func GetHandlerByPathAndMethod(t *testing.T, handlers []obsidian.Handler, path s
...
+}
+
+type yamlToJsonConverter struct {
+	yaml string
+}
+
+func (j *yamlToJsonConverter) MarshalBinary() (data []byte, err error) {
+	return yaml.YAMLToJSON([]byte(j.yaml))
+}
+
```
AndyLKhuu, line 125: This is a bit of an ugly workaround but SwaggerUIBundle doesn't play well with the output of the existing json.Marshal() and yaml.YAMLToJson seems to require the least amount of extra work. Resultingly, I had to settle for adding this construct to satisfy the encoding.BinaryMarshaler field of TestCase

AndyLKhuu:
Update: No need to review this PR fo rnow. With some offline discussions, my current approach is definitely way over engineered.

orc8r/cloud/go/obsidian/tests/unit_test_utils.go
```diff
@@ -86,14 +87,18 @@ func RunUnitTest(t *testing.T, e *echo.Echo, test Test) {
...
-		if assert.NoError(t, handlerErr) && test.ExpectedResult != nil {
+	} else if test.ExpectedResult != nil {
+		if assert.NoError(t, handlerErr) {
 			expectedBytes, err := test.ExpectedResult.MarshalBinary()
 			if assert.NoError(t, err) {
 				// Convert to string for more readable assert failure messages
 				assert.Equal(t, string(expectedBytes)+"\n", recorder.Body.String())
 			}
 		}
+	} else if test.ExpectedString != "" {
```
AndyLKhuu, line 98: Do we have to be concerned about the case in which a future test case may actually want to test that the expected string is the empty string?

orc8r/cloud/configs/obsidian.yml
```diff
@@ -9,3 +9,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# When true, poll and merge Swagger specs at runtime
+# When false, do not poll or merge Swagger specs at runtime
+enable_runtime_specs : true
```
hcgatewood: "enable_dynamic_swagger_specs"

Also, for the false option: document the fact that we're instead falling back to the static asset stored locally on the obsidian service

orc8r/cloud/configs/obsidian.yml
```diff
@@ -9,3 +9,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# When true, poll and merge Swagger specs at runtime
```
hcgatewood: Prefer Go-style doc comments pretty much everywhere -- i.e., prefix the doc comment with the name of the object it's documenting

orc8r/cloud/go/obsidian/obsidian/main.go
```diff
@@ -59,6 +59,12 @@ func main() {
 		log.Fatalf("Error creating service: %s", err)
 	}
 
+	enableRunTimeSpec := srv.Config.MustGetBool(obsidian.EnableRunTimeSpecs)
```
hcgatewood: I don't think there's a great need to make this config value override-able via CLI flag, so we can remove these additions

orc8r/cloud/go/obsidian/server/server.go
```diff
@@ -43,6 +44,10 @@ func Start() {
 	e.Use(CollectStats)
 	e.Use(middleware.Recover())
 
+	if obsidian.CombineSpecAtRuntime {
+		e.GET(obsidian.StaticURLPrefix+"/v1/swagger.yml", handlers.GenerateCombinedSpecHandler)
```
hcgatewood: Add a note that this is implicitly overriding the analogous static endpoint

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -0,0 +1,39 @@
...
+	"magma/orc8r/cloud/go/obsidian"
+	"magma/orc8r/cloud/go/obsidian/swagger"
+
+	"github.com/labstack/echo"
+)
+
+// GenerateCombinedSpecHandler is a routing handler which creates and serves the
+// combined Swagger Spec.
+func GenerateCombinedSpecHandler(c echo.Context) error {
+	yamlCommon, err := swagger.GetCommonSpec()
```
hcgatewood: Hoist the reading of GetCommonSpec into the server.go so that we only ever have to do it once

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"net/http"
+
+	"magma/orc8r/cloud/go/obsidian"
+	"magma/orc8r/cloud/go/obsidian/swagger"
+
+	"github.com/labstack/echo"
+)
+
+// GenerateCombinedSpecHandler is a routing handler which creates and serves the
```
hcgatewood: Wrap doc line length to 79 chars

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -0,0 +1,117 @@
...
+	swagger_lib "magma/orc8r/cloud/go/swagger"
+	"magma/orc8r/cloud/go/test_utils"
+	"magma/orc8r/lib/go/registry"
+
+	"github.com/labstack/echo"
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/yaml.v2"
+)
+
+func Test_GenerateCombinedSpecHandler(t *testing.T) {
```
hcgatewood, line 33: I envision this being a quick sanity-check test, since the handler is a thin wrapper around the existing functionality. I don't think you'll need to do any file I/O

orc8r/cloud/go/obsidian/swagger/spec_servicer.go
```diff
@@ -30,6 +31,7 @@ type specServicer struct {
 // NewSpecServicerFromFile intializes a spec servicer
 // given a service name.
 func NewSpecServicerFromFile(service string) protos.SwaggerSpecServer {
+	service = strings.ToLower(service)
```
hcgatewood, line 34: :+1:

For my understanding: did you run into an issue or did you just read how we do this in other areas of the codebase?

orc8r/cloud/go/obsidian/config.go
```diff
@@ -24,15 +24,20 @@ const (
 	DefaultStaticFolder  = "/var/opt/magma/static"
 	StaticURLPrefix      = "/apidocs"
 	ServiceName          = "OBSIDIAN"
+	// EnableRunTimeSpecs is a parameter name in the obsidian service config.
+	// When true, poll and merge Swagger specs at runtime
+	// When false, do not poll or merge Swagger specs at runtime
+	EnableRunTimeSpecs = "enable_runtime_specs"
```
hcgatewood: Rename this one too to match the comment in obsidian.yml

orc8r/cloud/go/obsidian/tests/unit_test_utils.go
```diff
@@ -122,6 +134,18 @@ func (j *jsonMarshaler) MarshalBinary() (data []byte, err error) {
 	return json.Marshal(j.v)
 }
 
+func YAMLMarshaler(v string) encoding.BinaryMarshaler {
```
hcgatewood: What you have right now is actually StringMarshaler

The first-best way to handle this is probably to have specs implement the encoding.BinaryMarshaler interface directly. Then you won't need this new test-oriented YAMLMarshaler

That is, add a Spec.MarshalBinary method. It can either replace or forward to the existing marshalToYAML private method. Then, you can just pass a spec directly to ExpectedResult, rather than having to first marshal it to string

hcgatewood (changes requested):
Looking good. I think we can simplify the handler test (just needs to be a sanity check), and can simplify how we pass specs to ExpectedResult. After that I think we'll be there

orc8r/cloud/go/obsidian/swagger/spec_servicer.go
AndyLKhuu: When I was checking the runtime logs I noticed that services that were in all CAPS(e.g. CTRACED) were unable to register so I dig some digging and found this to be the best solution.

orc8r/cloud/go/obsidian/tests/unit_test_utils.go
AndyLKhuu: Ahhh got it, that's definitely a cleaner solution.

orc8r/cloud/go/obsidian/obsidian/main.go
AndyLKhuu: If this is the case, I'm thinking that we would get rid of the config in obsidian.yml and just have a boolean flag in config.go? @hcgatewood

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
AndyLKhuu: Hmm I was thinking the same thing but I couldn't see a easy way to store the common spec? Would we just have it as a global var in server.go?

orc8r/cloud/go/obsidian/obsidian/main.go
```diff
@@ -53,6 +53,10 @@ func main() {
 		&obsidian.StaticFolder, "static_folder", obsidian.DefaultStaticFolder,
 		"Folder containing the static files served",
 	)
+	flag.BoolVar(
```
hcgatewood: Sorry maybe I was unclear before: we want the config values to come from the obisidian.yml config file, and here (in main.go) we set the relevant config.go global based on the value from the config file. No CLI arguments involved

As an aside, as we previously discussed, the way we handle configs in orc8r is a bit haphazard, here in particular. But let's go with the way I describe. This will allow operators to flip things by editing their config file, rather than having to potentially upload new Helm charts (idk if it would be easy or hard, but I know config changes are intentionally easy)

orc8r/cloud/go/obsidian/server/server.go
```diff
@@ -43,6 +45,18 @@ func Start() {
 	e.Use(CollectStats)
 	e.Use(middleware.Recover())
 
+	// If the feature is toggled on, serve the dynamically generated Swagger
+	// Spec instead of the static asset.
+	if obsidian.EnableDynamicSwaggerSpecs {
+		yamlCommon, err := swagger.GetCommonSpec()
+		if err != nil {
+			log.Printf("An error occurred while retrieving the Swagger common spec %+v", err)
```
hcgatewood: - Use glog instead of log
- `Error retrieving Swagger common spec: %+v`

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/labstack/echo"
+)
+
+// GetGenerateCombinedSpecHandler returns a routing handler which creates
+// and serves the combined Swagger Spec.
+func GetGenerateCombinedSpecHandler(yamlCommon string) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		combined, err := swagger.GetCombinedSpec(yamlCommon)
+		if err != nil {
+			return obsidian.HttpError(err, http.StatusNotFound)
```
hcgatewood: This should be a 500 (`http.StatusInternalServerError`)

orc8r/cloud/go/obsidian/config.go
```diff
@@ -35,4 +35,9 @@ var (
 	ClientCAPoolPath   string
 	AllowAnyClientCert bool
 	StaticFolder       string
+	// EnableDynamicSwaggerSpecs is a config in the obsidian
+	// service config.
+	// When true, poll and merge Swagger specs at runtime
+	// When false, fall back to serving the static Swagger Spec asset
```
hcgatewood: Spec -> spec

orc8r/cloud/go/swagger/spec.go
```diff
@@ -36,3 +36,12 @@ type TagDefinition struct {
 	Description string
 	Name        string
 }
+
+// MarshalBinary marshals the Spec structure into a byte form.
```
hcgatewood: - `MarshalBinary marshals the spec to bytes.`
- Also you can optionally leave the doc out, since it's an iface impl. Up to you

hcgatewood (approved):
Lgtm. Fix nits and pulling the feature toggle from config file rather than CLI flag, then msg me to merge

orc8r/cloud/configs/obsidian.yml
```diff
@@ -9,3 +9,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# enable_dynamic_swagger_specs is a config in the obsidian
+# service config.
+# When true, poll and merge Swagger specs at runtime
+# When false, fall back to serving the static Swagger spec asset
+enable_dynamic_swagger_specs : true
```
hcgatewood: Nit: remove trailing space before colon

orc8r/cloud/go/obsidian/obsidian/main.go
```diff
@@ -59,6 +59,8 @@ func main() {
 		log.Fatalf("Error creating service: %s", err)
 	}
 
+	obsidian.CombineSpecAtRuntime = srv.Config.MustGetBool(obsidian.EnableDynamicSwaggerSpecs)
```
hcgatewood: No reason to rename this config -- just have them both be `EnableDynamicSwaggerSpecs`

orc8r/cloud/go/obsidian/config.go
```diff
@@ -24,15 +24,21 @@ const (
 	DefaultStaticFolder  = "/var/opt/magma/static"
 	StaticURLPrefix      = "/apidocs"
 	ServiceName          = "OBSIDIAN"
+	// EnableDynamicSwaggerSpecs is a config in the obsidian
+	// service config.
+	// When true, poll and merge Swagger specs at runtime
```
hcgatewood: Nit: merge -> combine

I've been trying to differentiate between "combine", the high-level handling of logically combining multiple specs together, vs. "merge", the low-level implementation detail of how any two specs are merged

orc8r/cloud/go/obsidian/server/server.go
```diff
@@ -43,6 +46,18 @@ func Start() {
 	e.Use(CollectStats)
 	e.Use(middleware.Recover())
 
+	// If the feature is toggled on, serve the dynamically generated Swagger
+	// Spec instead of the static asset.
```
hcgatewood: Can remove this comment -- it's clear from the documentation of the config value

Instead, replace it with a description of how we're doing something a little funky -- implicitly overriding the static endpoint with a dynamic endpoint, and how that functionality isn't explicitly documented

I know we're planning on changing this in the next PR but it's good practice to keep up high quality, because you never know how long something might end up in the codebase if e.g. you suddenly need to pivot what you're working on

hcgatewood (approved):
Small nits, msg me when handled and I'll turn on auto-merge

themarwhal (approved):

PR #4905, opened 2021-02-11
[orc8r][docs] More 1.4 upgrade feedback

hcgatewood:
## Summary

Self-explanatory

## Test Plan

None

## Additional Information

### DIFF
--- a/docs/readmes/orc8r/upgrade_1_4.md
+++ b/docs/readmes/orc8r/upgrade_1_4.md
@@ -50,7 +50,6 @@ container versions
 module orc8r {
   source = "github.com/magma/magma//orc8r/cloud/deploy/terraform/orc8r-aws?ref=v1.4"
   # ...
-
   cluster_version = "1.17"   # set to Kubernetes version found above. 1.17 is used as an example here
 }
 
@@ -58,8 +57,8 @@ module orc8r-app {
   source = "github.com/magma/magma//orc8r/cloud/deploy/terraform/orc8r-helm-aws?ref=v1.4"
   # ...
   orc8r_chart_version   = "1.5.12"
-  orc8r_tag             = "MAGMA_TAG"       # from build step, e.g. v1.4.0
-  orc8r_deployment_type = DEPLOYMENT_TYPE   # valid options are: ["fwa", "federated_fwa", "all"]
+  orc8r_tag             = "MAGMA_TAG"  # from build step, e.g. v1.4.0
+  orc8r_deployment_type = "all"        # valid options: ["fwa", "federated_fwa", "all"]
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/readmes/orc8r/upgrade_1_4.md
```diff
@@ -78,11 +78,23 @@ terraform refresh
 
 ## Terraform apply
 
+Terraform the upgrade. We'll also need to manually rescale the Prometheus
+deployment, as a workaround for quirks in how its deployment handles upgrading.
```
karthiksubraveti: minor nit: Can you reference the prometheus crashlooping issue here or mention it, so that users won't be confused on why they might have to do this additionally.

docs/readmes/orc8r/upgrade_1_4.md
```diff
@@ -58,8 +58,8 @@ module orc8r-app {
   source = "github.com/magma/magma//orc8r/cloud/deploy/terraform/orc8r-helm-aws?ref=v1.4"
   # ...
   orc8r_chart_version   = "1.5.11"
-  orc8r_tag             = "MAGMA_TAG"       # from build step, e.g. v1.4.0
-  orc8r_deployment_type = DEPLOYMENT_TYPE   # valid options are: ["fwa", "federated_fwa", "all"]
+  orc8r_tag             = "MAGMA_TAG"  # from build step, e.g. v1.4.0
+  orc8r_deployment_type = "all"        # valid options: ["fwa", "federated_fwa", "all"]
```
sudhikan, line 61: can we make this `fwa` instead? As that is what we anticipate folks would do?

sudhikan: The other change I was requesting was: 

If this orc8r is a 2nd instance on this AWS **account**, then the service linked role `AWSServiceRoleForAmazonElasticsearchService` may already exist and you may see an error when running the final `terraform apply`. In such events, the variable `deploy_elasticsearch_service_linked_role` should be set to false under module `orc8r`.

docs/readmes/orc8r/upgrade_1_4.md
```diff
@@ -78,11 +78,23 @@ terraform refresh
...
 
 ```bash
 terraform apply     # Check this output VERY carefully!
 ```
 
-After applying Terraform changes, all your application components should be
-upgraded.
+Once the above command indicates it's applying the Helm upgrade, switch to
+a separate terminal tab and manually scale the Prometheus deployment. You'll
+have a 10-minute window to run the command. You can re-run the Terraform
```
Scott8440: What's the point of saying there's a 10 minute window? Seems kinda stressful lol. I'd also mention briefly the reasoning for running this command, and also say that data loss will be minimal if not none. e.g. "Rescaling the prometheus deployment allows for the upgraded pod to obtain the lock on the data directory. Only a few datapoints will be lost during this process, if any at all, due to how quickly this process happens."

docs/readmes/orc8r/upgrade_1_4.md
hcgatewood: Ah I see. I think since, for now at least, we explicitly mention we only support a single deployment per AWS account (hacky but wygd), perhaps this information belongs in the FAQ page. Thoughts?

In any case I'll leave it as a separate PR. Are you down to take that on since you have more context @sudhikan ?

docs/readmes/orc8r/upgrade_1_4.md
```diff
@@ -50,16 +50,15 @@ container versions
...
 }
 
 module orc8r-app {
   source = "github.com/magma/magma//orc8r/cloud/deploy/terraform/orc8r-helm-aws?ref=v1.4"
   # ...
   orc8r_chart_version   = "1.5.12"
-  orc8r_tag             = "MAGMA_TAG"       # from build step, e.g. v1.4.0
-  orc8r_deployment_type = DEPLOYMENT_TYPE   # valid options are: ["fwa", "federated_fwa", "all"]
+  orc8r_tag             = "MAGMA_TAG"  # from build step, e.g. v1.4.0
+  orc8r_deployment_type = "all"        # valid options: ["fwa", "federated_fwa", "all"]
```
mpgermano, line 61: I think using "fwa" here might be better. I don't foresee an operator needing to deploy all.

mpgermano (approved):

hcgatewood:
Auto-merge got to this before I could make the change. See https://github.com/magma/magma/pull/5015

PR #4923, opened 2021-02-12
[lte][agw] Adding null check for ht keys on UEs context on stale eNB state

ardzoht:

## Summary

- Adding null check for keys on UEs coll hashtable for s1ap eNB association ref, as it can return NULL if the UE coll is not initialized / is empty

## Test Plan

- make integ_test

## Additional Information

### DIFF
--- a/lte/gateway/c/oai/tasks/s1ap/s1ap_mme.c
+++ b/lte/gateway/c/oai/tasks/s1ap/s1ap_mme.c
@@ -563,7 +563,6 @@ void s1ap_remove_ue(s1ap_state_t* state, ue_description_t* ue_ref) {
       OAILOG_INFO(LOG_S1AP, "Deleting eNB \n");
       set_gauge("s1_connection", 0, 1, "enb_name", enb_ref->enb_name);
       s1ap_remove_enb(state, enb_ref);
-      update_mme_app_stats_connected_enb_sub();
     }
   }
 }
@@ -577,4 +576,5 @@ void s1ap_remove_enb(s1ap_state_t* state, enb_description_t* enb_ref) {
   hashtable_uint64_ts_destroy(&enb_ref->ue_id_coll);
   hashtable_ts_free(&state->enbs, enb_ref->sctp_assoc_id);
   state->num_enbs--;
+  update_mme_app_stats_connected_enb_sub();
 }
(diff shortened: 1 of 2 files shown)


lte/gateway/c/oai/tasks/s1ap/s1ap_mme_handlers.c
```diff
@@ -339,6 +339,11 @@ void clean_stale_enb_state(
   // Remove the S1 context for UEs associated with old eNB association
   hashtable_key_array_t* keys =
       hashtable_uint64_ts_get_keys(&stale_enb_association->ue_id_coll);
+  if (keys == NULL) {
```
ssanadhya: if (keys == NULL) then we should skip the for loop to clear the UEs, but still call s1ap_remove_enb()

ssanadhya (changes requested):

ssanadhya (approved):

lte/gateway/c/oai/tasks/s1ap/s1ap_mme_handlers.c
```diff
@@ -339,12 +339,15 @@ void clean_stale_enb_state(
...
+      ue_ref = s1ap_state_get_ue_mmeid((mme_ue_s1ap_id_t) keys->keys[i]);
+      s1ap_remove_ue(state, ue_ref);
+    }
+    FREE_HASHTABLE_KEY_ARRAY(keys);
   }
-  FREE_HASHTABLE_KEY_ARRAY(keys);
+
   // Remove the old eNB association
   s1ap_remove_enb(state, stale_enb_association);
   update_mme_app_stats_connected_enb_sub();
```
ssanadhya: Nit comment: Can we move the call to `update_mme_app_stats_connected_enb_sub()` inside `s1ap_remove_enb()` ? 
In this case, the metric will be decremented twice, once when eNB is removed as part of last call to `s1ap_remove_ue()` and once in line 353 here

lte/gateway/c/oai/tasks/s1ap/s1ap_mme.c
```diff
@@ -577,4 +576,5 @@ void s1ap_remove_enb(s1ap_state_t* state, enb_description_t* enb_ref) {
   hashtable_uint64_ts_destroy(&enb_ref->ue_id_coll);
   hashtable_ts_free(&state->enbs, enb_ref->sctp_assoc_id);
   state->num_enbs--;
+  update_mme_app_stats_connected_enb_sub();
```
ulaskozat, line 579: if this is moved to here, please check if there are other places where s1ap_remove_enb is called as well.

ulaskozat (changes requested):
please double check the usage of s1ap_remove_enb paired with update_mme_app_stats_connected_enb_sub()

rdefosse:
MAGMA-OAI-TESTS

rdefosse:
MAGMA-OAI-TESTS

ulaskozat (approved):
LGTM

PR #4961, opened 2021-02-16
[feg] s6a handle more than one Regional-Subscription-Zone-Code AVP

### LINKED ISSUE #4943: FEG support for TAC restriction
Add support for Regional-Subscription-Zone-Code AVP on Update Location Response coming from HSS

etsi.org/deliver/etsi_ts/129200_129299/129272/09.07.00_60/ts_129272v090700p.pdf

7.3.52 Regional-Subscription-Zone-Code
The Regional-Subscription-Zone-Code AVP is of type OctetString. Up to 10 zone codes shall all be defined as the
tracking or routing areas into which the subscriber is allowed to roam. See 3GPP TS 23.003 [3].
NOTE: Any internal list of zone code roaming restrictions that may be generated by the MME/SGSN from the
information in this AVP is an implementation issue only.

uri200:

## Summary

This pr extends https://github.com/magma/magma/pull/4858 to handle more than one Regional-Subscription-Zone-Code AVP.

It also modifies protobuf to support repeated values

## Test Plan

make precommit

## Additional Information


themarwhal (approved):

feg/protos/s6a_proxy.proto
```diff
@@ -168,7 +168,7 @@ message UpdateLocationAnswer {
 
     NetworkAccessMode network_access_mode = 8;
 
-   bytes regional_subscription_zone_code = 9;
+    repeated string regional_subscription_zone_code = 9;
```
themarwhal, line 171: is it just me or is the spacing weird here?

uri200: I think it was off before

PR #4984, opened 2021-02-16
[ARTIFACTORY]Removing useless id check

tmdzk:

[ARTIFACTORY]Removing useless id check

## Summary

We were hardcoding the gpg key idea which is totally useless.


koolzz (approved):
cool

PR #4989, opened 2021-02-17
[lte][agw] Crash del sess resp

ulaskozat:
## Summary

During delete session response processing, a null pointer access occurs. The PR adds the additional check for non-null pointer.

## Test Plan

S1AP integ tests.

## Additional Information


ulaskozat:
aborting this pr.

PR #4992, opened 2021-02-17
Rewrite pipelined stateless stats recovery

koolzz:

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/enforcement_stats.py
+++ b/lte/gateway/python/magma/pipelined/app/enforcement_stats.py
@@ -83,10 +83,7 @@ class EnforcementStatsController(PolicyMixin, RestartMixin, MagmaController):
         self._clean_restart = kwargs['config']['clean_restart']
         self._redis_enabled = kwargs['config'].get('redis_enabled', False)
         # Store last usage excluding deleted flows for calculating deltas
-        if self._redis_enabled:
-            self.last_usage_for_delta = UsageDeltaDict()
-        else:
-            self.last_usage_for_delta = {}
+        self.last_usage_for_delta = defaultdict(RuleRecord)
         self.failed_usage = {}  # Store failed usage to retry rpc to sessiond
         self._unmatched_bytes = 0  # Store bytes matched by default rule if any
         self._default_drop_flow_name = \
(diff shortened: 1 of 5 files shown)


lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -631,6 +615,13 @@ def _delta_usage_maps(self, current_usage, last_usage):
                 new_usage[key] = current
         return new_usage
 
+    def recover_state(self, flows):
+        for flow in flows[self.tbl_num]:
+            self.last_usage_for_delta = self._update_usage_from_flow_stat(
+                self.last_usage_for_delta, flow)
+        self.logger.error("Recovered stats:")
```
koolzz: will change to info

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -83,10 +83,7 @@ def __init__(self, *args, **kwargs):
         self._clean_restart = kwargs['config']['clean_restart']
         self._redis_enabled = kwargs['config'].get('redis_enabled', False)
         # Store last usage excluding deleted flows for calculating deltas
-        if self._redis_enabled:
-            self.last_usage_for_delta = UsageDeltaDict()
-        else:
-            self.last_usage_for_delta = {}
+        self.last_usage_for_delta = defaultdict(RuleRecord)
         self.failed_usage = {}  # Store failed usage to retry rpc to sessiond
         self._unmatched_bytes = 0  # Store bytes matched by default rule if any
```
pshelar, line 87: We could clear enforcement table flow stats using OFPFF_RESET_COUNTS on stat up, by re-adding existing flows to OVS.

https://github.com/faucetsdn/ryu/blob/a394673993468ad3ee6cd784eb0d52aab967a8b3/ryu/ofproto/ofproto_v1_4_parser.py#L4353

koolzz: That is quite nice, I do have 1 concern though. We're aiming for flat rate reporting in near term future, if I wipe stats on restart then it won't work well with flat rate

pshelar: right. flat rate reporting is better solution. this is more of stopgap.

pshelar (approved):
Lets get this change in, we can handle lost stats in separate PR.

PR #5016, opened 2021-02-18
{WIP}[lte][agw] Add 50msec RPC delay for create session and teid update handling

ulaskozat:
## Summary

Experimental PR to test minimum of 50 msec pacing between RPC calls.

## Test Plan

spirent tests for backported version to v1.4

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalSessionManagerHandler.cpp
+++ b/lte/gateway/c/session_manager/LocalSessionManagerHandler.cpp
@@ -37,6 +37,8 @@ LocalSessionManagerHandlerImpl::LocalSessionManagerHandlerImpl(
       current_epoch_(0),
       reported_epoch_(0),
       retry_timeout_ms_(std::chrono::milliseconds{5000}),
+      min_rpc_delta_(std::chrono::milliseconds{50}),
+      last_rpctime_pipelined_(std::chrono::milliseconds{0}),
       is_setting_up_pipelined_(false) {}
 
 void LocalSessionManagerHandlerImpl::ReportRuleStats(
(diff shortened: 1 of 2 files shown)


uri200:
Hey @ulaskozat, is this just a test, or are we adding some kind of buffering for pipelined?

Note since we splitted `create_session` into two, now `create_session` doesn't have any interaction with pipelined. All calls to pipelined are done during `upadate_tunnel_id`.

But If we want to reduce around 50% the amount of calls to pipelined during the creation of the session, we should take a look on the issue noted below where pipelined can't take static and dynamic rules on the same `ActivateFlowsRequest` . Right now we are doing two separate calls per session to activate the flows.

```
  // TODO: Activate static rules and dynamic rules separately until bug
  //  is fixed in pipelined which crashes if activated at the same time
```
http://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/PipelinedClient.cpp#L324-L324


Also, if we moved forward with buffering approach, can we consider to do this buffering as soon as we receive the requests in MME? Otherwise we are rushing the rest of the system unnecessarily. If we are going to wait in sessiond maybe it is better to wait even before so all the systems can benefit from that

Finally, why isn't this buffering done at pipelined level? Pieplined doesn't return anything meaningful to sessiond so it can buffer the request as much as needed.

ulaskozat:
@uri200 I was playing with some pacing ideas to be able to resolve grpc timeouts for the calls from sessiond to pipelined in the scaling tests. But, these did not generate the results I expected.

ulaskozat:
@uri200 since I was testing this for v1.4 branch, I included pacing for both create session and update tunnel id stages...

PR #5017, opened 2021-02-18
[SessionD] Reduce GRPC calls from SessionD to PipelineD

themarwhal:
## Summary
We might still have the pipelined issue where dynamic+static rules cannot be installed together, but we can reduce the number of RPCs by checking.

## Test Plan
make precommit_sm

## Additional Information


uri200:
Hey Marie, I thought on this too but if we have static and dynamic rules we still have the same problem

Maybe we don't see the issue on the test anymore, but wi may see it on the field

themarwhal:
@uri200 Yeah I'm not fixing the problem, just making it less bad. We'll have to fix this on pipelined eventually, but in the meantime I wanted to reduce the number of empty calls.

uri200 (approved):

themarwhal:
removing tag, going to merge manually

PR #5018, opened 2021-02-18
[SessionD] Terminate session after pipelined GRPC failure

themarwhal:
## Summary
When we fail to send an activate flows request to PipelineD, don't retry and terminate the session.

## Additional Information


uri200 (approved):

uri200:
I think it is better not to retry and let the UE decide if it needs to send a new attach. 
Also, retries on same vm services are normally caused due to congestion. And that is exactly the moment when you don't want to retry. 

So this looks like an improvement in case of congestion

PR #5030, opened 2021-02-18
[lte][agw] Fixing null ptr access for PDN context sgw teid on handling create session response

ardzoht:

## Summary

- There's a potential null ptr access during create session response on SGW teid property, this PR adds a check for PDN context existence.

## Test Plan

- make integ_test

## Additional Information

### DIFF
--- a/lte/gateway/c/oai/tasks/mme_app/mme_app_bearer.c
+++ b/lte/gateway/c/oai/tasks/mme_app/mme_app_bearer.c
@@ -1123,8 +1123,10 @@ int mme_app_handle_create_sess_resp(
             "Bad pdn id (%d) for bearer\n", pdn_cx_id);
         continue;
       }
-      ue_context_p->pdn_contexts[pdn_cx_id]->s_gw_teid_s11_s4 =
-          create_sess_resp_pP->s11_sgw_fteid.teid;
+      if (ue_context_p->pdn_contexts[pdn_cx_id]) {
+        ue_context_p->pdn_contexts[pdn_cx_id]->s_gw_teid_s11_s4 =
+            create_sess_resp_pP->s11_sgw_fteid.teid;
+      }
       transaction_identifier = current_bearer_p->transaction_identifier;
     }
 


karthiksubraveti:
@ardzoht just curious why would this be NULL ?

lte/gateway/c/oai/tasks/mme_app/mme_app_bearer.c
```diff
@@ -1123,8 +1123,10 @@ int mme_app_handle_create_sess_resp(
             "Bad pdn id (%d) for bearer\n", pdn_cx_id);
         continue;
       }
-      ue_context_p->pdn_contexts[pdn_cx_id]->s_gw_teid_s11_s4 =
-          create_sess_resp_pP->s11_sgw_fteid.teid;
+      if (ue_context_p->pdn_contexts[pdn_cx_id]) {
```
karthiksubraveti, line 1126: @ardzoht just curious why would this be NULL ?

electronjoe: As to why this would be NULL - could one of the many generators of messages on the ZMQ bus be allocating a MessageDef e.g. [Here](https://github.com/magma/magma/blob/327600a/lte/gateway/c/oai/lib/itti/intertask_interface.c#L267) but then not filling the message data (leaving it zero-initialized as seen) - which then would be interpreted as a NULL ptr (well it would be a null ptr)?

electronjoe: I realize my comment is off - the ZMQ message is not at play here.

ulaskozat: If the pdn context is not there, either there should be a delete session request processed or default bearer should be deactivated, while there is still a create session request pending. @ardzoht what is the context here?

ulaskozat (approved):
LGTM

rdefosse:
MAGMA-OAI-TESTS

lte/gateway/c/oai/tasks/mme_app/mme_app_bearer.c
ardzoht: There's a delete session request procedure for theUE on the logs I've looked at related to the coredumps, so I think would be the former case you mention @ulaskozat

PR #5034, opened 2021-02-18
[helm][tf] provide default grafana volumes

Scott8440:

## Summary
Partner engineer found a helm templating problem when deploying on bare metal due to grafana volume values being empty. The default values for these volumes were passed in as empty objects `{}` because of the way helm merges key/values from values files. If a real default was provided then the "override" would not actually override and instead provide to volume specs for the same volume which would cause templating to fail. To solve this, I use a field `defaultVolumeSpec` in all the grafana volume values and check if a real value is provided. If so, the real value is used otherwise it falls back to the default. 

## Test Plan
Deployed with Terraform:
```
Volumes:
  config:
    Type:      ConfigMap (a volume populated by a ConfigMap)
    Name:      grafana-config-file
    Optional:  false
  datasources:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  grafanadatasources
    ReadOnly:   false
  dashboardproviders:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  grafanaproviders
    ReadOnly:   false
  dashboards:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  grafanadashboards
    ReadOnly:   false
  grafana-data:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  grafanadata
    ReadOnly:   false
```

Deployed on minikube with/without values and saw expected results

### DIFF
--- a/docs/readmes/orc8r/upgrade_1_4.md
+++ b/docs/readmes/orc8r/upgrade_1_4.md
@@ -56,14 +56,14 @@ module orc8r {
 module orc8r-app {
   source = "github.com/magma/magma//orc8r/cloud/deploy/terraform/orc8r-helm-aws?ref=v1.4"
   # ...
-  orc8r_chart_version   = "1.5.15"
+  orc8r_chart_version   = "1.5.16"
   orc8r_tag             = "MAGMA_TAG"  # from build step, e.g. v1.4.0
   orc8r_deployment_type = "fwa"        # valid options: ["fwa", "federated_fwa", "all"]
 }
 ```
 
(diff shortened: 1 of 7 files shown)


karthiksubraveti (approved):

mpgermano (approved):
There are some merge conflicts. Looks good otherwise

PR #5035, opened 2021-02-18
AGW: pipelined: qos: fix recovery of APN AMBR config

pshelar:
Rather than using TC as only source of APN AMBR, use Redis
to store the qid->APN_qid and APN_leaf->APN_qid info.
This avoid complicated recovery in stateless mode.
This PR also adds apn_ambr_enabled config to enable APN AMBR feature.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/qos/common.py
+++ b/lte/gateway/python/magma/pipelined/qos/common.py
@@ -18,7 +18,8 @@ from enum import Enum
 from lte.protos.policydb_pb2 import FlowMatch
 from magma.pipelined.qos.qos_meter_impl import MeterManager
 from magma.pipelined.qos.qos_tc_impl import TCManager, TrafficClass
-from magma.pipelined.qos.types import QosInfo, get_json, get_key, get_subscriber_key
+from magma.pipelined.qos.types import QosInfo, get_key_json, get_key, get_subscriber_key,\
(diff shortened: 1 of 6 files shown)


lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -104,26 +105,33 @@ def find_session_with_rule(self, rule_num: int) -> SubscriberSession:
...
-        self._redis_store[get_json(k)] = qos_handle
+    def _update_rules_map(self, ip_addr: str, rule_num: int, d: FlowMatch.Direction,
+                    qos_data) -> None:
 
         if rule_num not in self.rules:
             self.rules[rule_num] = []
 
         session = self.get_or_create_session(ip_addr)
         session.rules.add(rule_num)
-        self.rules[rule_num].append((d, qos_handle))
```
koolzz, line 117: should the name be a bit more descriptive, rules is confusing because we're accessing it by rule_num to retrieve what exactly?

koolzz (approved):
this looks quite painful thanks for fixing

lte/gateway/python/magma/pipelined/qos/common.py
pshelar: I agree, But I would do it as followup patch to avoid complicating review.

ardzoht (approved):

PR #5052, opened 2021-02-19
[FeG][Orc8r] Add directoryd configs

themarwhal:
## Summary
the python dockerfile in `feg/gateway/docker/python` uses feg/gateway/configs to build a python image
the python dockerfile in `orc8r/gateway/docker` uses orc8r/gateway/configs to build a python images

They're both currently lacking the directoryd config, which leads the service to crashloop.
This should fix the cwf integration test failures.

## Additional Information

### DIFF
--- a/cwf/gateway/configs/directoryd.yml
+++ b/cwf/gateway/configs/directoryd.yml
@@ -1,0 +1,15 @@
+---
+#
+# Copyright 2020 The Magma Authors.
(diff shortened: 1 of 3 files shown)


uri200 (approved):

uri200:
I missed that on the addition of grpc!
Thanks!

PR #5055, opened 2021-02-19
[SessionD][v1.4] Deactivate flows for dead sessions on usage reporting

themarwhal:
## Summary
On some failure cases, we can have flows leftover in PIpelineD that are not cleaned up.
Whenever we receive flows for a session that does not exist anymore, we should re-send deactivate flows to pipelined to help cleanup.

## Test Plan
unit tests

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -234,14 +234,16 @@ void LocalEnforcer::aggregate_records(
   // Insert the IMSI+SessionID for sessions we received a rule record into a set
   // for easy access
   std::unordered_set<ImsiAndSessionID> sessions_with_reporting_flows;
-  for (const RuleRecord& record : records.records()) {
+  // In some failure cases, PipelineD may still hold onto flows for sessions
+  // that do not exist in SessionD. In this case, send DeactivateFlowsRequest
+  std::vector<RuleRecord> dead_sessions_to_cleanup;
+  for (const RuleRecord record : records.records()) {
     const std::string &imsi = record.sid(), &ip = record.ue_ipv4();
     // TODO IPv6 add ipv6 to search criteria
     SessionSearchCriteria criteria(imsi, IMSI_AND_UE_IPV4_OR_IPV6, ip);
     auto session_it = session_store_.find_session(session_map, criteria);
     if (!session_it) {
-      MLOG(MERROR) << "Could not find session for " << imsi << " and " << ip
-                   << " during record aggregation";
+      dead_sessions_to_cleanup.push_back(record);
       continue;
     }
     auto& session          = **session_it;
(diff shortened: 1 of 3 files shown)


koolzz (approved):
neat

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -234,14 +234,16 @@ void LocalEnforcer::aggregate_records(
...
+  std::vector<RuleRecord> dead_sessions_to_cleanup;
+  for (const RuleRecord record : records.records()) {
     const std::string &imsi = record.sid(), &ip = record.ue_ipv4();
     // TODO IPv6 add ipv6 to search criteria
     SessionSearchCriteria criteria(imsi, IMSI_AND_UE_IPV4_OR_IPV6, ip);
     auto session_it = session_store_.find_session(session_map, criteria);
     if (!session_it) {
-      MLOG(MERROR) << "Could not find session for " << imsi << " and " << ip
-                   << " during record aggregation";
+      dead_sessions_to_cleanup.push_back(record);
```
uri200, line 246: did you see that case?
I am asking because i wasn't able to ever see it in teravm
But it make sense add this here

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -260,6 +262,26 @@ void LocalEnforcer::aggregate_records(
...
+void LocalEnforcer::cleanup_dead_sessions(
+    std::vector<RuleRecord>& dead_sessions_to_cleanup) {
+  if (dead_sessions_to_cleanup.size() == 0) {
+    return;
+  }
+  for (const RuleRecord& record : dead_sessions_to_cleanup) {
+    MLOG(MINFO) << "Removing all flows for " << record.sid() << ", "
+                << record.ue_ipv4()
+                << " since the UE no longer exists in SessionD";
+    Teids empty_teids;  // Teids are not used in 1.4
```
uri200, line 277: We will have to remember to add the Teids here once they are included on the report

Maybe @koolzz can add the teids to the message so we can start using teids here

uri200 (approved):

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: Yea for v1.4 we have to either have pipelined send teid or do something else :)

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: @themarwhal Yeah seeing a lot in v1.4. I expect to start seeing it in master also because we start terminating sessions that cannot install flows properly.

themarwhal: (meaning, that if the grpc calls to activate flows are failing, the deactivate flows calls when terminating might fail too.)

uri200: ohh right. In teravm we haven't been able to run load tests like you are doing. That is why.

PR #5060, opened 2021-02-20
[orc8r] Add Routes for Per Service UI

AndyLKhuu:

## Summary
This PR exposes new routes in the Swagger UI documentation of our API so that it is possible to retrieve  a targeted spec of service in both raw YAML and UI form. These changes were made in preparation for the UI overhaul into a more ergonomic tabbed view of the API and importantly includes a significant migration of Swagger related assets. 

Listed below is an overview of the routing changes

**NOTE:** With the changes in routing made by this PR, the **existing /apidocs/v1 route will be deprecated**, for now this PR has added changes which will redirect the user to the **appropriate new route (/swagger/v1/ui/)** so please be conscious of migrating to the new route. 

**Added Routes** 
/swagger/v1/spec/ → serves combined spec as YAML
/swagger/v1/spec/{service} → serve service spec as YAML

_Example_: 
/swagger/v1/spec/ctraced 
Screen Shot 2021-02-22 at 5 33 10 PM

/swagger/v1/ui/ → serve combined spec with UI
/swagger/v1/ui/{service} → serve servie spec with UI 

_Example_: 
/swagger/v1/ui/ctraced 
Screen Shot 2021-02-22 at 5 33 01 PM

**Resource Migrations**
/apidocs/v1/swagger.yml → /swagger./v1/spec/swagger.yml
/apidocs/v1/index.html → /swagger/v1/ui/index.html

## Test Plan
- [x] Manual inspection of routes 

## Additional Information


orc8r/cloud/docker/controller/apidocs/v1/indexTemplated.html
```diff
@@ -0,0 +1,102 @@
+<!-- HTML for static distribution bundle build -->
+<!-- This file is pulled from swagger-ui repository -->
+<!-- If you bump swagger-ui version, make sure to update this file as well! -->
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Magma API Docs</title>
```
AndyLKhuu, line 8: Don't worry about this extra html, I added it temporarily as a means of making sure that I dont make silly mistakes.

orc8r/cloud/go/obsidian/tests/unit_test_utils.go
```diff
@@ -134,6 +134,18 @@ func (j *jsonMarshaler) MarshalBinary() (data []byte, err error) {
...
+	return &stringMarshaler{s}
+}
+
+type stringMarshaler struct {
+	s string
+}
+
+func (m *stringMarshaler) MarshalBinary() (data []byte, err error) {
+	return []byte(m.s), nil
+}
```
AndyLKhuu, line 147: @hcgatewood Very similar situation to the last PR but it doesn't seem like we have the luxury to add a MarshalBinary() method to the template in this case. For now, it seems like we would inevitably have to add a StringMarshaler but I would love to know if there is a more efficient solution.

orc8r/cloud/docker/controller/Dockerfile
```diff
@@ -105,6 +105,8 @@ RUN mkdir -p /var/opt/magma/envdir
 # Build artifacts
 ARG CNTLR_FILES=src/magma/orc8r/cloud/docker/controller
 COPY --from=builder /${CNTLR_FILES}/apidocs /var/opt/magma/static/apidocs
+COPY --from=builder /${CNTLR_FILES}/apidocs/v1/index.html /var/opt/magma/static/swagger/v1/ui/index.html
```
hcgatewood: After thinking about it, let's go ahead and do this
- Move source code `.../apidocs/...` to `.../swagger/...`
- Have prod image follow `/var/opt/magma/static/swagger/v1/{ui,spec,static}/...`, where we are including the Swagger UI static assets (e.g. dist/, etc) under the `v1/static/` directory
- Have HTTP endpoints match the above-mentioned prod image paths. So basically same as you have it now, but also put the dist/ etc under `.../v1/static/...`

As a side note: the end-result container image should only contain one copy of index.html

orc8r/cloud/docker/controller/apidocs/v1/index.html
```diff
@@ -37,7 +37,7 @@
      style="position:absolute;width:0;height:0">
   <defs>
     <symbol viewBox="0 0 20 20" id="unlocked">
-          <path d="M15.8 8H14V5.6C14 2.703 12.665 1 10 1 7.334 1 6 2.703 6 5.6V6h2v-.801C8 3.754 8.797 3 10 3c1.203 0 2 .754 2 2.199V8H4c-.553 0-1 .646-1 1.199V17c0 .549.428 1.139.951 1.307l1.197.387C5.672 18.861 6.55 19 7.1 19h5.8c.549 0 1.428-.139 1.951-.307l1.196-.387c.524-.167.953-.757.953-1.306V9.199C17 8.646 16.352 8 15.8 8z"></path>
+      <path d="M15.8 8H14V5.6C14 2.703 12.665 1 10 1 7.334 1 6 2.703 6 5.6V6h2v-.801C8 3.754 8.797 3 10 3c1.203 0 2 .754 2 2.199V8H4c-.553 0-1 .646-1 1.199V17c0 .549.428 1.139.951 1.307l1.197.387C5.672 18.861 6.55 19 7.1 19h5.8c.549 0 1.428-.139 1.951-.307l1.196-.387c.524-.167.953-.757.953-1.306V9.199C17 8.646 16.352 8 15.8 8z"></path>
```
hcgatewood: Revert these indent/style-based changes. We don't really "own" this file since it's copied from an online source, so no need to mess with blame history

orc8r/cloud/docker/controller/apidocs/v1/index.html
```diff
@@ -70,30 +70,33 @@
...
-window.onload = function() {
-
-  // Build a system
-  const ui = SwaggerUIBundle({
-    url: window.location.origin + '/apidocs/v1/swagger.yml',
+  window.onload = function() {
+
+    // Build a system
+    const ui = SwaggerUIBundle({
+      url: {{if enableDynamicSwaggerSpecs}}
```
hcgatewood: Using a control block is overkill here. Just template the value to the url key, and pass in the value when templating

orc8r/cloud/docker/controller/apidocs/v1/index.html
```diff
@@ -70,30 +70,33 @@
...
+    window.location.origin + '/swagger/v1/spec/' + '{{.}}'
```
hcgatewood: Use a struct template value rather than a string, for forward compatibility

orc8r/cloud/go/obsidian/server/server.go
```diff
@@ -46,27 +47,33 @@
...
 
-	// With this, we implicitly override the static endpoint serving a static
-	// Swagger spec with a dynamic endpoint which dynamically serves the spec.
-	// Note that this behavior is not documented in the Echo framework.
+	yamlCommon, err := swagger.GetCommonSpec()
+	if err != nil {
+		glog.Errorf("Error retrieving Swagger common spec %+v", err)
+	}
+
 	if obsidian.EnableDynamicSwaggerSpecs {
```
hcgatewood: Move this whole thing into RegisterSpecHandlers, where you pass the boolean value of the feature toggle into the function

orc8r/cloud/go/obsidian/server/server.go
```diff
@@ -46,27 +47,33 @@
...
+	tmpl, err := template.New("index.html").Funcs(template.FuncMap{
+		"enableDynamicSwaggerSpecs": func() bool {
+			return obsidian.EnableDynamicSwaggerSpecs
+		},
+	}).ParseFiles(obsidian.StaticFolder + "/swagger/v1/ui/index.html")
+	if err != nil {
+		glog.Errorf("Error retrieving Swagger UI template %+v", err)
+	}
+
+	handlers.RegisterSpecHandlers(e, yamlCommon, tmpl)
```
hcgatewood: "RegisterSwaggerHandlers"

orc8r/cloud/go/obsidian/server/server.go
```diff
@@ -46,27 +47,33 @@
...
-			e.GET(obsidian.StaticURLPrefix+"/v1/swagger.yml", handler)
-		}
+		handler := handlers.GetGenerateCombinedSpecHandler(yamlCommon)
+		e.GET(obsidian.StaticURLPrefix+"/spec/", handler)
 	}
 
-	// Serve static pages for the API docs
-	e.Static(obsidian.StaticURLPrefix, obsidian.StaticFolder+"/apidocs")
-	e.Static(obsidian.StaticURLPrefix+"/swagger-ui/dist", obsidian.StaticFolder+"/swagger-ui/dist")
+	tmpl, err := template.New("index.html").Funcs(template.FuncMap{
```
hcgatewood: This funcs functionality is unnecessary. From my index.html comments: we don't need to handle the feature flag in the index.html, just inject the proper URL in whichever handler

orc8r/cloud/go/obsidian/config.go
```diff
@@ -22,7 +22,7 @@ const (
 	DefaultServerCertKey         = "server_cert.key.pem"
 	DefaultClientCAs             = "ca_cert.pem"
 	DefaultStaticFolder          = "/var/opt/magma/static"
-	StaticURLPrefix              = "/apidocs"
+	StaticURLPrefix              = "/swagger/v1"
```
hcgatewood: You'll need to handle `StaticURLPrefixLegacy = "/apidocs" in https://github.com/magma/magma/blob/074a1a8b9cf7718d3a954d8d087b9b911ac6f3a9/orc8r/cloud/go/obsidian/access/rest_middleware.go#L53

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -14,14 +14,35 @@
...
 import (
+	"bytes"
+	"html/template"
 	"net/http"
+	"strings"
 
 	"magma/orc8r/cloud/go/obsidian"
 	"magma/orc8r/cloud/go/obsidian/swagger"
+	"magma/orc8r/cloud/go/orc8r"
+	swagger_lib "magma/orc8r/cloud/go/swagger"
```
hcgatewood: Don't forget to add a PR merging the two swagger pkgs

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -14,14 +14,35 @@
...
+	"magma/orc8r/lib/go/registry"
 
+	"github.com/golang/glog"
 	"github.com/labstack/echo"
+	"github.com/pkg/errors"
 )
 
+// RegisterSpecHandlers registers routes for the Swagger UI.
```
hcgatewood: "... for Swagger specs and the Swagger UI."

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -14,14 +14,35 @@
...
+func RegisterSpecHandlers(e *echo.Echo, yamlCommon string, tmpl *template.Template) {
+	e.GET(obsidian.StaticURLPrefix+"/spec/:service", GetGenerateSpecHandler(yamlCommon))
+
+	e.GET(obsidian.StaticURLPrefix+"/ui/:service", GetGenerateSpecUIHandler(tmpl))
+	e.GET(obsidian.StaticURLPrefix+"/ui/", GetGenerateSpecUIHandler(tmpl))
+
+	// Redirect requests for apidocs/v1/ to swagger/v1/ui/
+	e.GET("/apidocs/v1/", func(c echo.Context) error {
```
hcgatewood: Use the `StaticURLPrefixLegacy` mentioned in config.go comments

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -14,14 +14,35 @@
...
+		return c.Redirect(http.StatusMovedPermanently, obsidian.StaticURLPrefix+"/ui/")
+	})
+}
+
 // GetGenerateCombinedSpecHandler returns a routing handler which creates
 // and serves the combined Swagger Spec.
 func GetGenerateCombinedSpecHandler(yamlCommon string) echo.HandlerFunc {
```
hcgatewood: "GetSpecHandler"

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +54,76 @@
 		return c.String(http.StatusOK, combined)
 	}
 }
+
+// GetGenerateSpecHandler returns a routing handler which creates and
+// serves the raw YAML spec of a service.
+func GetGenerateSpecHandler(yamlCommon string) echo.HandlerFunc {
```
hcgatewood: "GetPerServiceSpecHandler"

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +54,76 @@
...
+			glog.Infof("Some Swagger spec traits were overwritten or unable to be read: %+v", warnings)
+		}
+
+		return c.String(http.StatusOK, combined)
+	}
+}
+
+// GetGenerateSpecUIHandler returns a routing handler which serves
+// the UI of a service.
+func GetGenerateSpecUIHandler(tmpl *template.Template) echo.HandlerFunc {
```
hcgatewood: "GetUIHandler"

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +54,76 @@
...
+			return obsidian.HttpError(err, http.StatusInternalServerError)
+		}
+
+		return c.HTML(http.StatusOK, tpl.String())
+	}
+}
+
+// isValidService returns if the service is a registered
+// service with a Swagger spec.
+func isValidService(service string) (string, error) {
```
hcgatewood: `func getServiceName(service string) (string, bool, error)`.

Then call it as

```
service, ok, err := getServiceName(requestedService)
if err != nil {
  // internal server error
}
if !ok {
  // 404
}
// ...
```

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +54,76 @@
...
+	services, err := registry.FindServices(orc8r.SwaggerSpecLabel)
+	if err != nil {
+		return "", err
+	}
+
+	// If no service is provided, default to empty string to serve combined spec
```
hcgatewood, line 158: Move this check out of this function. Because if there's no service, we want to template in the URL for the monolithic Swagger spec

hcgatewood, line 158: Side note: double check for the spec and ui monolithic endpoints, as well as the legacy apidocs endpoint -- they should all work *with and without* the trailing slash. Maybe a good way to check this is to encode that expectation into a test

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +54,76 @@
...
+	if service == "" {
+		return "", nil
+	}
+
+	for _, s := range services {
+		if service == s {
+			return strings.ToLower(s), nil
```
hcgatewood: Add comment here about how we're explicitly *not* returning the user-generated string, as a defense-in-depth measure against XSS

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +54,76 @@
...
+		}
+	}
+
+	return "", errors.New("Service provided is not registered.")
```
hcgatewood: General: errors should start with lowercase letter (unless the first word desired caps, e.g. "HTTP") and ends with no punctuation

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +54,76 @@
...
+func GetGenerateSpecUIHandler(tmpl *template.Template) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		service := c.Param("service")
+
+		service, err := isValidService(service)
+		if err != nil {
+			return obsidian.HttpError(err, http.StatusInternalServerError)
+		}
+
+		var tpl bytes.Buffer
```
hcgatewood: "ui" or "buf"

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +54,76 @@
...
+func GetGenerateSpecHandler(yamlCommon string) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		service := c.Param("service")
+
+		service, err := isValidService(service)
+		if err != nil {
+			return obsidian.HttpError(err, http.StatusInternalServerError)
+		}
+
+		remoteSpec := swagger.NewRemoteSpec(service)
```
hcgatewood: Move the logic of this function to swagger pkg, e.g. how you call swagger.GetCombinedSpec above

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -80,6 +83,111 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
+
+	// Fail with invalid service name.
+	tc := tests.Test{
+		Method:         "GET",
+		URL:            testURLRoot,
+		Payload:        nil,
+		ParamNames:     []string{"service"},
+		ParamValues:    []string{"invalid_test_spec_service"},
+		Handler:        handlers.GetGenerateSpecHandler(yamlCommon),
+		ExpectedStatus: http.StatusInternalServerError,
```
hcgatewood: Should be 404

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -80,6 +83,111 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
+		ExpectedError:  "Service provided is not registered.",
```
hcgatewood: "Not found" or equivalent

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -80,6 +83,111 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
+	}
+	tests.RunUnitTest(t, e, tc)
+
+	// Success with valid service name.
+	tag := swagger_lib.TagDefinition{Name: "Tag 1"}
+	testSpec := swagger_lib.Spec{
```
hcgatewood: "expected" or "expectedSpec"

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -80,6 +83,111 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
 
+func Test_GenerateSpecHandler(t *testing.T) {
+	e := echo.New()
+	testURLRoot := "/magma/v1"
+
+	commonTag := swagger_lib.TagDefinition{Name: "Tag Common"}
+	commonSpec := swagger_lib.Spec{
+		Tags: []swagger_lib.TagDefinition{commonTag},
+	}
+	yamlCommon := marshalToYAML(t, commonSpec)
```
hcgatewood, line 91: Side note: update the marshalToYAML impl to use the spec's MarshalBinary method, rather than calling yaml.marshal directly

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -80,6 +83,111 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
+
+	// Fail with invalid service name
+	tc := tests.Test{
+		Method:                 "GET",
+		URL:                    testURLRoot,
+		Payload:                nil,
+		ParamNames:             []string{"service"},
+		ParamValues:            []string{"fake_test_service"},
+		Handler:                handlers.GetGenerateSpecUIHandler(tmpl),
+		ExpectedStatus:         500,
```
hcgatewood: 404

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -80,6 +83,111 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
+		ExpectedErrorSubstring: "Service provided is not registered.",
```
hcgatewood: "Not found" or equivalent

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -80,6 +83,111 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
+		ExpectedResult: testSpec,
+	}
+	tests.RunUnitTest(t, e, tc)
+}
+
+func Test_GenerateSpecUIHandler(t *testing.T) {
+	e := echo.New()
+	testURLRoot := "/magma/v1"
+
+	tmpl, err := template.New("test_template.html").Parse(`{{.}}`)
```
hcgatewood: Give the template some placeholder content. Otherwise we can't tell whether a returned empty string is incidental or intentional

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -80,6 +83,111 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
+		Payload:                nil,
+		ParamNames:             []string{"service"},
+		ParamValues:            []string{"fake_test_service"},
+		Handler:                handlers.GetGenerateSpecUIHandler(tmpl),
+		ExpectedStatus:         500,
+		ExpectedErrorSubstring: "Service provided is not registered.",
+	}
+	tests.RunUnitTest(t, e, tc)
+
+	// Success with empty service name
```
hcgatewood: Add the note that this is a success because we should be pointing to the monolithic swagger spec

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -80,6 +83,111 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
+
+	tc = tests.Test{
+		Method:         "GET",
+		URL:            testURLRoot,
+		Payload:        nil,
+		ParamNames:     []string{"service"},
+		ParamValues:    []string{testService2},
+		Handler:        handlers.GetGenerateSpecUIHandler(tmpl),
+		ExpectedStatus: 200,
+		ExpectedResult: tests.StringMarshaler(tpl.String()),
```
hcgatewood: Manually construct this expected result as a literal, rather than doing any templating

orc8r/cloud/go/obsidian/tests/unit_test_utils.go
hcgatewood: Yeah this is fine. Since you can't create methods for builtin types, we'd have to wrap the string in some object one way or another, and this is a simple way to handle it

hcgatewood (changes requested):
Thanks for adding. Comments mostly on conciseness, and some corner cases

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
AndyLKhuu: Yup that'll be the immediate PR I work on after this one is merged.

orc8r/cloud/docker/controller/Dockerfile
```diff
@@ -103,11 +103,13 @@ COPY src/magma/orc8r/cloud/docker/wait-for-it.sh /usr/local/bin
 RUN mkdir -p /var/opt/magma/envdir
 
 # Build artifacts
```
hcgatewood, line 105: General notes
- The goswagger-config.yml is part of swaggergen, not combine_swagger. So it needs to stay in its original spot. You could consider renaming the original location to get rid of the "apidocs" name from vcs filesystem, but in any case that file doesn't belong in this v1 pkg
- Where are we copying over the swagger ui static assets?

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -50,7 +50,7 @@ func Middleware(next echo.HandlerFunc) echo.HandlerFunc {
 		}
 
 		perms := getRequestedPermissions(req, decorate)
-		staticReadOnly := perms == accessprotos.AccessControl_READ && strings.HasPrefix(c.Path(), obsidian.StaticURLPrefix)
+		staticReadOnly := perms == accessprotos.AccessControl_READ && strings.HasPrefix(c.Path(), obsidian.StaticURLPrefixLegacy)
```
hcgatewood: This is probably hard to sanity check locally because we auto-permit requests if they come from localhost. Here's what you want

```
isStatic := strings.HasPrefix(c.Path(), obsidian.StaticURLPrefix) || strings.HasPrefix(c.Path(), obsidian.StaticURLPrefixLegacy)
isStaticReadOnly := isStatic && perms == ...
```

orc8r/cloud/go/obsidian/server/server.go
```diff
@@ -46,27 +45,17 @@
...
-		yamlCommon, err := swagger.GetCommonSpec()
-		if err != nil {
-			glog.Errorf("Error retrieving Swagger common spec: %+v", err)
-		} else {
-			handler := handlers.GetGenerateCombinedSpecHandler(yamlCommon)
-			e.GET(obsidian.StaticURLPrefix+"/v1/swagger.yml", handler)
-		}
+	err := handlers.RegisterSwaggerHandlers(e)
+	if err != nil {
+		glog.Errorf("error occurred while registering Swagger handlers %+v", err)
```
hcgatewood: - Add note on swallowing error
- "Error registering ..."

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -14,17 +14,50 @@
...
+// RegisterSwaggerHandlers registers routes for Swagger specs and
+// the Swagger UI.
+func RegisterSwaggerHandlers(e *echo.Echo) error {
+	trailSlashMiddleware := middleware.AddTrailingSlashWithConfig(middleware.TrailingSlashConfig{
+		RedirectCode: http.StatusMovedPermanently,
+	})
+
+	err := registerSwaggerSpecHandlers(e, trailSlashMiddleware)
+	if err != nil {
+		return obsidian.HttpError(err, http.StatusInternalServerError)
```
hcgatewood: No need to return http errors from this function, since the errors aren't being propagated back to an http client

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +66,115 @@
...
+	e.GET(obsidian.StaticURLPrefix+"/ui/", GetUIHandler(tmpl, obsidian.EnableDynamicSwaggerSpecs))
+	e.GET(obsidian.StaticURLPrefix+"/ui/:service", GetUIHandler(tmpl, obsidian.EnableDynamicSwaggerSpecs))
+
+	return nil
+}
+
+// getServiceName returns if the service is a registered
+// service with a Swagger spec.
+func getServiceName(service string) (string, bool, error) {
+	// If no service is provided, default to empty string to serve combined spec.
```
hcgatewood: Tiny nit/up to you: for inline comments (rather than doc comments), I prefer to not add a period unless it's serving as a full-stop before a new sentence

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -14,17 +14,50 @@
...
 
 	"github.com/labstack/echo"
+	"github.com/labstack/echo/middleware"
+	"github.com/pkg/errors"
 )
 
-// GetGenerateCombinedSpecHandler returns a routing handler which creates
+// RegisterSwaggerHandlers registers routes for Swagger specs and
+// the Swagger UI.
+func RegisterSwaggerHandlers(e *echo.Echo) error {
```
hcgatewood, line 41: This is the only identifier (function/object/type) that needs to be public in this file, the rest can be private. Once you've done that, reorder your functions such that they approximately follow a top-down call hierarchy. E.g. `registerSwaggerSpecHandlers` should be near the top, under `RegisterSwaggerHandlers`, because it's called directly by `RegisterSwaggerHandlers`

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +66,115 @@
...
+		if err != nil {
+			return obsidian.HttpError(err, http.StatusInternalServerError)
+		}
+
+		return c.HTML(http.StatusOK, buf.String())
+	}
+}
+
+// registerSwaggerSpecHandlers registers routes for Swagger Specs.
+func registerSwaggerSpecHandlers(e *echo.Echo, trailSlashMiddleware echo.MiddlewareFunc) error {
```
hcgatewood: "registerSpecHandlers"

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +66,115 @@
...
+
+	e.GET(obsidian.StaticURLPrefix+"/spec", nil, trailSlashMiddleware)
+	e.GET(obsidian.StaticURLPrefix+"/spec/", GetCombinedSpecHandler(yamlCommon))
+	e.GET(obsidian.StaticURLPrefix+"/spec/:service", GetSpecHandler(yamlCommon))
+
+	return nil
+}
+
+// registerSwaggerUIHandlers registers routes for the Swagger UI.
+func registerSwaggerUIHandlers(e *echo.Echo, trailSlashMiddleware echo.MiddlewareFunc) error {
```
hcgatewood: "registerUIHandlers"

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +66,115 @@
...
+			return obsidian.HttpError(err, http.StatusInternalServerError)
+		}
+
+		return c.String(http.StatusOK, combined)
+	}
+}
+
+// SpecInfo defines the URL to be injected into
+// the Swagger UI template.
+type SpecInfo struct {
```
hcgatewood: "UIInfo" or "UITmplVars" etc

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +66,115 @@
...
+		combined, err := swagger.GetCombinedSpecFromService(yamlCommon, service)
+		if err != nil {
+			return obsidian.HttpError(err, http.StatusInternalServerError)
+		}
+
+		return c.String(http.StatusOK, combined)
+	}
+}
+
+// SpecInfo defines the URL to be injected into
```
hcgatewood: The fact that this template variables struct contains specifically the URL info should be handled in the doc comment for the URL field

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +66,115 @@
...
+// the Swagger UI template.
+type SpecInfo struct {
+	URL string
```
hcgatewood: "url" (unless making private doesn't play nicely with the templating pkg)

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +66,115 @@
 		return c.String(http.StatusOK, combined)
 	}
 }
+
+// GetSpecHandler returns a routing handler which creates and
+// serves the raw YAML spec of a singular service.
```
hcgatewood: Nit: singular -> particular

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +66,115 @@
...
+func GetSpecHandler(yamlCommon string) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		requestedService := c.Param("service")
```
hcgatewood: Inline this directly as an argument to `getServiceName` below

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -58,6 +58,26 @@
 	return combined, nil
 }
 
+// GetCombinedSpecFromService polls a service for its spec and combines
+// it with the common spec to return a combined spec.
+func GetCombinedSpecFromService(yamlCommon string, service string) (string, error) {
```
hcgatewood: "GetServiceSpec" or "GetPerServiceSpec"

orc8r/cloud/go/obsidian/swagger/combine.go
```diff
@@ -58,6 +58,26 @@
...
+	remoteSpec := NewRemoteSpec(service)
```
hcgatewood, line 64: Inline this into the next line

orc8r/cloud/go/obsidian/swagger/v1/index.html
```diff
@@ -70,14 +70,14 @@
...
+<script src="../static/swagger-ui/dist/swagger-ui-standalone-preset.js"> </script>
 <script>
 window.onload = function() {
 
-  // Build a system
-  const ui = SwaggerUIBundle({
-    url: window.location.origin + '/apidocs/v1/swagger.yml',
+    // Build a system
+    const ui = SwaggerUIBundle({
+    url: window.location.origin + '/swagger/v1/spec/' + '{{.URL}}',
```
hcgatewood: For simplicity, let's just acquire full control over the location. Like this

```
url: window.location.origin + '{{.url}}',
```

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +66,115 @@
...
+		err = tmpl.Execute(&buf, SpecInfo{URL: service})
+		if err != nil {
+			return obsidian.HttpError(err, http.StatusInternalServerError)
+		}
+
+		return c.HTML(http.StatusOK, buf.String())
+	}
+}
+
+// registerSwaggerSpecHandlers registers routes for Swagger Specs.
```
hcgatewood: Specs -> specs

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -73,13 +73,130 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
+		ExpectedResult: expected,
+	}
+	tests.RunUnitTest(t, e, tc)
+}
+
+func Test_GenerateSpecUIHandler(t *testing.T) {
+	e := echo.New()
+	testURLRoot := "/magma/v1"
+
+	tmplStr := "Test Result is "
```
hcgatewood: Nit: just manually re-create the full literal in each ExpectedResult. Is easier to read/skim the test cases without having to refer back up to this variable

Corollary then is also use a full literal in the Parse method just below

hcgatewood (changes requested):
Looks good. Style comments, main correctness item is the access middleware point

orc8r/cloud/docker/controller/Dockerfile
AndyLKhuu: 1.) Ah I see, weird that the goswagger-config.yml was originally placed within apidocs/ but okay. I definitely agree that it doesn't belong under the v1 scope, I'll move it back to its original location
2.) We copy over the swagger ui static assets within the Dockerfile::92. There wasn't any changes on it since we discussed to have them remain where they were right?

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
AndyLKhuu: Gotchu, thanks for the catch

orc8r/cloud/go/obsidian/swagger/combine.go
AndyLKhuu: This would actually require us to change the methods of RemoteSpec to use value receivers instead of pointer receivers? For the sake of uniformity, as it seems like most user defined types in the code base use pointer receivers, I'm thinking we should just ignore this one line change? Unless you have a strong preference to maybe switch over to value receivers for the methods, which I also understand the benefits of.

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
AndyLKhuu: Yeah both the field and the struct have to be public in order for it to properly template. The reason why I went with "URL" instead of "Url" is because I've never seen the latter being used.

hcgatewood: Gotcha good to know. So yeah then URL is correct

orc8r/cloud/go/obsidian/swagger/combine.go
hcgatewood: From offline: is Go quirk, leave as-is

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -14,17 +14,84 @@
...
+	"magma/orc8r/cloud/go/orc8r"
+	"magma/orc8r/lib/go/registry"
 
 	"github.com/labstack/echo"
+	"github.com/labstack/echo/middleware"
+	"github.com/pkg/errors"
 )
 
-// GetGenerateCombinedSpecHandler returns a routing handler which creates
+// Defines the struct to be templated into Swagger UI.
```
hcgatewood: "UIInfo contains the templating variables injected into the Swagger UI template."

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -14,17 +14,84 @@
...
+type UIInfo struct {
+	// Defines the URL to be injected into the Swagger UI template
```
hcgatewood: "URL of the underlying Swagger spec."

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -14,17 +14,84 @@
...
+	URL string
+}
+
+// RegisterSwaggerHandlers registers routes for Swagger specs and
+// the Swagger UI.
+func RegisterSwaggerHandlers(e *echo.Echo) error {
+	trailSlashMiddleware := middleware.AddTrailingSlashWithConfig(middleware.TrailingSlashConfig{
+		RedirectCode: http.StatusMovedPermanently,
```
hcgatewood, line 43: Is this the default? I.e. do you have to use `WithConfig` ?

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +100,79 @@ func GetGenerateCombinedSpecHandler(yamlCommon string) echo.HandlerFunc {
...
+		combined, err := swagger.GetServiceSpec(yamlCommon, service)
+		if err != nil {
+			return obsidian.HttpError(err, http.StatusInternalServerError)
+		}
+
+		return c.String(http.StatusOK, combined)
+	}
+}
+
+// getUIHandler returns a routing handler which serves
```
hcgatewood: Nit here and elsewhere: you're wrapping the doc comments too aggressively. Let them flow out toward the 79-character limit, then wrap

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +100,79 @@ func GetGenerateCombinedSpecHandler(yamlCommon string) echo.HandlerFunc {
...
+
+		service, ok, err := getServiceName(requestedService)
+		if err != nil {
+			return obsidian.HttpError(err, http.StatusInternalServerError)
+		}
+		if !ok {
+			return obsidian.HttpError(errors.New("service not found"), http.StatusNotFound)
+		}
+
+		// If runtime Swagger spec is off, serve static spec
```
hcgatewood: This code paragraph is wrong
- UI shouldn't care about dynamic swagger spec toggle
- If service is empty, just point to the `.../spec/` endpoint, which should serve the monolithic spec

The impl of the `.../spec/` endpoint is where we handle the feature toggle

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -11,13 +11,13 @@
  * limitations under the License.
  */
 
-package handlers_test
+package handlers
```
hcgatewood: We want to use separate test pkg where at all possible. If you need to export `getCombinedSpecHandler` and similar, that's fine

I may have given conflicting PR advice about exporting -- but in this case, exporting those identifiers is better than having an in-pkg test file

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -73,13 +72,129 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
+		ExpectedResult: expected,
+	}
+	tests.RunUnitTest(t, e, tc)
+}
+
+func Test_GenerateSpecUIHandler(t *testing.T) {
+	e := echo.New()
+	testURLRoot := "/magma/v1"
+
+	tmpl, err := template.New("test_template.html").Parse("Test Result is {{.URL}}")
```
hcgatewood: `"Test result is"` doesn't provide semantic meaning to the test. Instead let's use something like `"swagger_spec_url: {{.URL}}"`

orc8r/cloud/go/obsidian/config.go
```diff
@@ -22,7 +22,8 @@ const (
...
```
hcgatewood: This is actually just `/swagger`. Will be important for permissions for reading Swagger static assets

hcgatewood (changes requested):
Looks good, couple small things

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
AndyLKhuu: Yup, we have to use add this version of TrailingSlash(withConfig) because otherwise it won't redirect.

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
AndyLKhuu: Got it, will keep this in mind.

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
AndyLKhuu: Got it, that's what I initially thought but I changed it because of the previous round of revision.

orc8r/cloud/go/obsidian/server/server.go
```diff
@@ -46,27 +45,19 @@ func Start() {
...
+		// Swallow RegisterHandlerError because the obsidian service should
+		// continue to run even if Swagger handlers aren't registered.
+		glog.Errorf("Error registering Swagger handlers %+v", err)
 	}
 
-	// Serve static pages for the API docs
-	e.Static(obsidian.StaticURLPrefix, obsidian.StaticFolder+"/apidocs")
-	e.Static(obsidian.StaticURLPrefix+"/swagger-ui/dist", obsidian.StaticFolder+"/swagger-ui/dist")
+	// Serve static assets for the Swagger UI
+	e.Static(obsidian.StaticURLPrefix+"/v1/static/swagger-ui/dist", obsidian.StaticFolder+"/swagger-ui/dist")
```
hcgatewood: Nit: I think we aligned on having the static assets be *not* scoped by version, so e.g. just `/swagger/static/...`

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +73,105 @@
...
+// registerSpecHandlers registers routes for Swagger specs.
+func registerSpecHandlers(e *echo.Echo, trailSlashMiddleware echo.MiddlewareFunc) error {
+	yamlCommon, err := swagger.GetCommonSpec()
+	if err != nil {
+		return err
+	}
+
+	if obsidian.EnableDynamicSwaggerSpecs {
+		e.GET(obsidian.StaticURLPrefix+"/v1/spec/", GetCombinedSpecHandler(yamlCommon))
+	} else { // If dynamic Swagger spec is off, serve static spec
```
hcgatewood: Nit: we don't use this style of trailing comment in orc8r. Either move it above the if line, or move it inside the else block

orc8r/cloud/go/obsidian/swagger/handlers/handlers.go
```diff
@@ -33,3 +73,105 @@
...
+		e.GET(obsidian.StaticURLPrefix+"/v1/spec/", func(c echo.Context) error {
+			return c.File(obsidian.StaticURLPrefix + "/v1/spec/swagger.yml")
+		})
+	}
+
+	e.GET(obsidian.StaticURLPrefix+"/v1/spec", nil, trailSlashMiddleware)
```
hcgatewood, line 134: Swap this line with the blank line above. Cause we wanna group the handlers for the monolithic endpoint, then group the (single) handler for the per-service endpoint

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -73,13 +73,115 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
...
+		ParamNames:     []string{"service"},
+		ParamValues:    []string{""},
+		Handler:        handlers.GetUIHandler(tmpl),
+		ExpectedStatus: 200,
+		ExpectedResult: tests.StringMarshaler("swagger_spec_url: /swagger/v1/spec/"),
+	}
+	tests.RunUnitTest(t, e, tc)
+
+	// Success with valid service name
+	testService2 := "test_spec_service2"
```
hcgatewood: Nit: rather than creating this variable, just inline the service name literal wherever the variable would have been used

hcgatewood (approved):
Looks good. Fix nits then msg to merge.

P.S. next week add a subsequent PR updating the current docs with the new default monolithic endpoints

PR #5070, opened 2021-02-22
[MME] Fix MME compile errors from compiling with GCC/G++8

themarwhal:
## Summary
I was experimenting with gcc/g++ 8 (we currently compile with gcc6), and noticed a bunch of compilation errors. Fixing that here as it's probably relevant for gcc6 too.

MME still compiles with these changes with gcc6.

Errors:
```
/home/vagrant/magma/lte/gateway/c/oai/include/state_converter.h:80:12: error: ‘std::function’ has not been declared
/home/vagrant/magma/lte/gateway/c/oai/lib/event_client/EventClientAPI.cpp:56:3: warning: control reaches end of non-void function [-Wreturn-type]
/home/vagrant/magma/lte/gateway/c/oai/lib/s6a_proxy/s6a_client_api.cpp:63:1: warning: control reaches end of non-void function [-Wreturn-type]
/home/vagrant/magma/lte/gateway/c/oai/tasks/mme_app/mme_events.cpp:50:1: warning: no return statement in function returning non-void [-Wreturn-type]
/home/vagrant/magma/lte/gateway/c/oai/test/mobility_client/test_mobility_client.cpp:71:9: warning: control reaches end of non-void function [-Wreturn-type]
/home/vagrant/magma/lte/gateway/c/oai/test/mobility_client/test_mobility_client.cpp:100:9: warning: control reaches end of non-void function [-Wreturn-type]
```

## Test Plan
unit tests

## Additional Information

### DIFF
--- a/lte/gateway/c/oai/include/mme_events.h
+++ b/lte/gateway/c/oai/include/mme_events.h
@@ -26,7 +26,7 @@ extern "C" {
 /**
  * Helper function to initiate AsyncEventdClient in its own thread
  */
-int event_client_init(void);
+void event_client_init(void);
 
 /**
  * Logs Attach successful event
--- a/lte/gateway/c/oai/include/state_converter.h
+++ b/lte/gateway/c/oai/include/state_converter.h
@@ -36,6 +36,7 @@ extern "C" {
 #include <functional>
 
 #include <google/protobuf/map.h>
+#include <functional>
 
 #include "lte/protos/oai/common_types.pb.h"
 
--- a/lte/gateway/c/oai/lib/event_client/EventClientAPI.cpp
+++ b/lte/gateway/c/oai/lib/event_client/EventClientAPI.cpp
@@ -45,14 +45,15 @@ int log_event(const Event& event) {
     if (status.ok()) {
       std::cout << "[DEBUG] Success logging event: " << event.event_type()
                 << std::endl;
-    } else if (
-        status.error_code() == DEADLINE_EXCEEDED ||
-        status.error_code() == UNAVAILABLE) {
-      return 0;  // Suppress error logs if eventd is unavailable
-    } else {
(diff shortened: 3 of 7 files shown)


lte/gateway/c/oai/include/mme_events.h
```diff
@@ -26,7 +26,7 @@ extern "C" {
 /**
  * Helper function to initiate AsyncEventdClient in its own thread
  */
-int event_client_init(void);
+void event_client_init(void);
```
ulaskozat, line 29: any reason to change the return type to void?

themarwhal: there is no return value
original:
```
int event_client_init(void) {
  init_eventd_client(); <- this also returns void
  // error here since no return value
}
```

change:
```
void event_client_init(void) {
  init_eventd_client(); <- this also returns void
}
```

ulaskozat: It may be better to update these other functions so that they actually return a success or failure code. Do we want the services get started in case of initialization failure for this client? I see that CHECK_INIT_RETURN pattern is not used for this, but probably we should. @ssanadhya what do you think?

ssanadhya: @ulaskozat  @themarwhal , Since Eventd is not a mandatory service for MME to run, the MME startup should not be blocked if it is unable to talk to Eventd. Let's not use CHECK_INIT_RETURN here as it will assert if the initialization failed. But, it will be good to log whether MME is able to communicate with Eventd or not. @ardzoht can comment if we already have some logs/alerts to track whether events are being received from MME or not.

ssanadhya (approved):

lte/gateway/c/oai/include/mme_events.h
ssanadhya: If that check is needed, we can do it in a separate Pull Request

PR #5085, opened 2021-02-22
[MME] Reduce mobilityd client in MME's GRPC timeout from 30->10 sec

themarwhal:

## Summary
Currently the GRPC timeout for the IP allocation call is longer than the UE re-attach timeout (~15 seconds).
On 1.4, we are observing an issue where due to a separate issue in mobilityd, the IP allocation occasionally takes longer than the UE re-attach. This introduces some unexpected call flows in MME that crashes the service. 

This change is **not** a solution to either of the issues above, but it helps to avoid the bad scenario. 
TODO: make this value configurable by mme.yml

## Test Plan
unit tests

## Additional Information


ssanadhya (approved):

ulaskozat (approved):
LGTM

PR #5094, opened 2021-02-23
[pipelined]Fix Null Rule Removal Error

koolzz:

## Additional Information


lte/gateway/python/magma/pipelined/app/enforcement.py
```diff
@@ -257,6 +257,9 @@ def _deactivate_flow_for_rule(self, imsi, ip_addr, rule_id):
         except KeyError:
             self.logger.error('Could not find rule id %s', rule_id)
             return
+        if num is None:
+            self.logger.error('Rule num is None for rule %s', rule_id)
+            return
```
pshelar, line 262: Dont we need fix on sessionD side to send rule number?

pshelar (approved):

PR #5106, opened 2021-02-24
[helm] update default service_regsitry.mode to k8s

Scott8440:

## Summary

When deploying on minikube (or bare metal) orc8r-nginx would crashloop due to a misconfiguration caused by the value `controller.service_registry.mode=yaml`, when for 1.4 it should be `k8s`.

## Test Plan

Deploy following minikube instructions (but using 1.4 and most recent helm charts), orc8r-nginx does not crash anymore, and deployment is the same as if deployed on terraform which has been validated to work.

### DIFF
--- a/cwf/cloud/helm/cwf-orc8r/Chart.yaml
+++ b/cwf/cloud/helm/cwf-orc8r/Chart.yaml
@@ -13,7 +13,7 @@ apiVersion: v2
 appVersion: "1.0"
 description: A Helm chart for magma orchestrator's cwf module
 name: cwf-orc8r
-version: 0.2.0
+version: 0.2.1
 engine: gotpl
 sources:
   - https://github.com/magma/magma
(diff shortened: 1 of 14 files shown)


Scott8440:
FYI I've only really tested this with the orc8r chart since that's generally what I deploy with minikube. LMK if this is not right for all the other charts.

hcgatewood (commented):
Lgtm but I'll defer to @mpgermano

mpgermano:
If we're going to apply this to v1.4 we need to update the documentation and all of the defaults in the variables.tf

mpgermano (changes requested):
Doc and tf updates needed for v1.4

Scott8440:
Updated documentation and tf values

ShubhamTatvamasi:
I faced a lot of issues due to this. :expressionless:

mpgermano (approved):

PR #5125, opened 2021-02-24
[AGW][S1ap tests] Handle SMC retransmissions for MME restarts

ssanadhya:

## Summary

The `test_no_smc_with_mme_restart_reattach.py` restarts the MME while it is running timer T3460 for the SMC it sent. In such cases, depending on when the MME restarted, it might retransmit the SMC, leading to flakiness in the test. This change adds a loop to absorb all the retransmissions before asserting on UE Context Release

## Test Plan

Ran test `test_no_smc_with_mme_restart_reattach.py` multiple times without any failure


ulaskozat (approved):
LGTM

PR #5127, opened 2021-02-24
[orc8r][configurator] Update configurator to allow for paginated entity loads

mpgermano:

## Summary

This PR adds support for pagination at the configurator level.  Entity loads can now be performed 
by specifying a page size and token for which that page should start from. Responses will include 
the page of entities and the next token to be used to retrieve subsequent pages.

A stacked PR implements the first pagination of an API endpoint via the list subscribers call.

Note: Only type-filtered loads permit pagination as of now. Otherwise we would need both a type and key
token for the page. This introduces more complexity that doesn't seem necessary given that "ListAll" entity loads
always contain a type filter.

## Test Plan

Added unit test and integ test for SQL.
Tested via subscriber API changes (in subsequent PR) to ensure
that e2e pagination works as expected.


orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -73,25 +79,29 @@ func (store *sqlConfiguratorStorage) getLoadEntitiesSelectBuilder(networkID stri
...
+			andClause = append(andClause, sq.Eq{fmt.Sprintf("ent.%s", entKeyCol): filter.KeyFilter.Value})
+		}
+		if filter.TypeFilter != nil {
+			andClause = append(andClause, sq.Eq{fmt.Sprintf("ent.%s", entTypeCol): filter.TypeFilter.Value})
+		}
+		if criteria.PageSize > 0 {
+			andClause = append(andClause, sq.Gt{fmt.Sprintf("ent.%s", entKeyCol): criteria.PageToken})
+		}
+		selectBuilder = selectBuilder.Where(andClause)
+		if criteria.PageSize > 0 {
```
hcgatewood: Always set query limit. Use criteria.GetPageSize() described in other comment. Let's make the default max like 10 or 15k, with a note in the service config that thing will start to break much above 15k

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -51,12 +53,16 @@ func (store *sqlConfiguratorStorage) loadFromEntitiesTable(networkID string, fil
...
 
-func (store *sqlConfiguratorStorage) getLoadEntitiesSelectBuilder(networkID string, filter EntityLoadFilter, criteria EntityLoadCriteria) sq.SelectBuilder {
+func (store *sqlConfiguratorStorage) getLoadEntitiesSelectBuilder(networkID string, filter EntityLoadFilter, criteria EntityLoadCriteria) (sq.SelectBuilder, error) {
 	// SELECT ent.pk, ent.key, ent.type, ent.physical_id, ent.version, graph.graph_id, ent.name, ent.description, ent.config,
 	// [[ acl.id, acl.scope, acl.permission, acl.type, acl.id_filter, acl.version ]]
 	// FROM cfg_entities AS ent
 	// [[ LEFT JOIN cfg_acls AS acl ON acl.entity_pk = ent.pk ]]
 	// [[ WHERE (ent.network_id = $1 AND ent.key = $2 AND ent.type = $3) OR (ent.network_id ...) ... ]]
+	// [[ ORDER BY ent.key ASC LIMIT page_size ]]
+	if err := validatePaginatedLoadParameters(filter, criteria); err != nil {
```
hcgatewood: Hoist this validation up to the caller -- there's only one call site, and not returning an error feels cleaner if we can help it

Generally speaking, I like to validate parameters as far toward the edge of the interface as possible. That way all internal code can follow the "if it exists then it's a valid value" rule, which makes for simpler implementations and easier to reason about

orc8r/cloud/go/services/configurator/storage/storage.proto
```diff
@@ -208,11 +208,20 @@ message EntityLoadCriteria {
     bool load_assocs_from_this = 4;
 
     bool load_permissions = 5;
+
+    // The following parameters allow for pagination of entity loads.
+    // Size of page to return
+    uint64 page_size = 6;
+    // Token for which all returned keys will be greater than
+    string page_token = 7;
```
hcgatewood, line 217: Both these page tokens should be the base64-encoded serialized value of an `EntityPageToken` proto message, defined in this file. That gives us stronger flexibility in forward compatibility

Since this page token will get passed all the way out through the REST API, doing things this way also adds the affordance of "this is an opaque page token*

orc8r/cloud/go/services/configurator/types.go
```diff
@@ -439,6 +439,14 @@ type EntityLoadCriteria struct {
 
 	LoadAssocsToThis   bool
 	LoadAssocsFromThis bool
+
+	// The following parameters allow for pagination of entity loads. These
+	// load criteria parameters should be used in combination with one another.
+
+	// Maximum size of entities returned
```
hcgatewood: Here and elsewhere -- follow go doc comment conventions, e.g. "PageSize is the maximum number of entities returned per load request."

orc8r/cloud/go/services/configurator/types.go
```diff
@@ -439,6 +439,14 @@ type EntityLoadCriteria struct {
...
+	PageSize uint64
+	// Token for which all entity keys returned are greater than
```
hcgatewood: This doc comment is peering into implementation details. You can add this as a "for reference this is now it's used", but the leading doc comment should be something like "PageToken is the opaque token provided to load the next page of entities."

orc8r/cloud/go/services/configurator/types.go
```diff
@@ -467,6 +477,9 @@ type EntityLoadResult struct {
 	Entities []NetworkEntity
 	// Entities which were not found
 	EntitiesNotFound []storage2.TypeAndKey
+	// Token which should be used to return the next page for paginated
```
hcgatewood: Same type of comments from above

orc8r/cloud/go/services/configurator/storage/storage.proto
```diff
@@ -208,11 +208,20 @@ message EntityLoadCriteria {
     bool load_assocs_from_this = 4;
 
     bool load_permissions = 5;
+
+    // The following parameters allow for pagination of entity loads.
+    // Size of page to return
+    uint64 page_size = 6;
```
hcgatewood: I think uint32 is sufficient. Consider a max uint32 page size retrieving 1-byte objects -- that's a 4gb page. And more realistic is like, 30-byte objects => 130gb. I don't anticipate us running into this requirement

orc8r/cloud/go/services/configurator/storage/storage.proto
```diff
@@ -208,11 +208,20 @@ message EntityLoadCriteria {
     bool load_assocs_from_this = 4;
 
     bool load_permissions = 5;
+
+    // The following parameters allow for pagination of entity loads.
```
hcgatewood, line 212: Add space below this doc line

orc8r/cloud/go/services/configurator/storage/storage.proto
```diff
@@ -208,11 +208,20 @@ message EntityLoadCriteria {
...
+    // Size of page to return
```
hcgatewood: Throughout this proto file: follow go-style doc conventions in proto files. So e.g. here "page_size is ..."

orc8r/cloud/go/services/configurator/storage/sql_test.go
```diff
@@ -539,6 +540,80 @@ func TestSqlConfiguratorStorage_LoadEntities(t *testing.T) {
 			EntitiesNotFound: []*storage.EntityID{{Type: "hello", Key: "world"}},
 		},
 	}
+	// Load  of 2 entities
+	loadPageRegex := regexp.MustCompile("SELECT ent\\.network_id, ent\\.pk, ent\\.\"key\", ent\\.type, ent\\.physical_id, ent\\.version, ent\\.graph_id, ent\\.name, ent\\.description, ent\\.config " +
```
hcgatewood: Use Go's backtick syntax to avoid having to escape in regex strings

orc8r/cloud/go/services/configurator/storage/sql_test.go
```diff
@@ -539,6 +540,80 @@ func TestSqlConfiguratorStorage_LoadEntities(t *testing.T) {
...
+		"FROM cfg_entities AS ent " +
+		"WHERE \\(ent\\.network_id \\= \\$1 AND ent\\.type \\= \\$2 AND ent\\.\"key\" \\> \\$3\\) " +
+		"ORDER BY ent\\.\"key\" ASC " +
+		"LIMIT 2")
```
hcgatewood: For multi-line arg to function, the opening and closing parens should be on their own line. E.g.

```
MustCompile(
  "SELECT ..." +
  "FROM ..." +
  ...
  "LIMIT ..."
)

File: orc8r/cloud/go/services/configurator/storage/sql.go [hunk 12]
```diff
@@ -472,6 +472,12 @@ func (store *sqlConfiguratorStorage) LoadEntities(networkID string, filter Entit
 	}
 	sort.Slice(ret.Entities, func(i, j int) bool { return entComparator(ret.Entities[i], ret.Entities[j]) })
 
+	// Ensure there are more pages to return
```
hcgatewood: "Set next page token when there may be more pages to return"

orc8r/cloud/go/services/configurator/storage/sql.go
```diff
@@ -472,6 +472,12 @@ func (store *sqlConfiguratorStorage) LoadEntities(networkID string, filter Entit
...
+	if len(ret.Entities) != 0 && uint64(len(ret.Entities)) == loadCriteria.PageSize {
```
hcgatewood:
```
if len(ret.Entities) == loadCriteria.GetPageSize() {
  ...
}


// and elsewhere

// and note that lc.PageSize of 0 is magic value
GetPageSize() int {
  max := configurator default
  if lc.PageSize == 0 {
    return max
  }
  return min(lc.PageSize, max)
}
```

orc8r/cloud/go/services/configurator/storage/sql.go
```diff
@@ -472,6 +472,12 @@ func (store *sqlConfiguratorStorage) LoadEntities(networkID string, filter Entit
 	}
 	sort.Slice(ret.Entities, func(i, j int) bool { return entComparator(ret.Entities[i], ret.Entities[j]) })
```
hcgatewood, line 475: Side note: update the doc.go of the configurator and/or configurator/storage pkgs to briefly mention the pagination functionality

hcgatewood (changes requested):
Looks good. Some small style stuff, main big things are
- page token should be a proto, rather than a bare string
- configurator service's configurable max page size -- and using a LoadCriteria helper to read out the "actual" page size for the request

orc8r/cloud/go/services/configurator/storage/sql.go
mpgermano: Given that the loadCriteria here is a proto, I had to make this an unexported function instead

orc8r/cloud/go/services/configurator/storage/storage.proto
```diff
@@ -208,11 +208,26 @@ message EntityLoadCriteria {
...
     repeated NetworkEntity entities = 1;
     repeated EntityID entities_not_found = 2;
+    // next_page_token is a serialized entity page token for paginated loads.
+    string next_page_token = 3;
+}
+
+// EntityPageToken is an opaque token provided to load the next page of
+// entities.
+message EntityPageToken {
+  string token = 1;
```
hcgatewood: Once we're inside the page token, the fields can have semantic meanings. So this should be `last_included_imsi`, with accompanying description of how it's the imsi of the last subscriber included in the returned response

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -15,23 +15,37 @@ package storage
...
 	sq "github.com/Masterminds/squirrel"
+	"github.com/golang/protobuf/proto"
 	"github.com/pkg/errors"
 	"github.com/thoas/go-funk"
 )
 
+const (
+	// MaxEntityLoadSize is the maximum number of entities that can be loaded
+	// in a request.
+	MaxEntityLoadSize = 15000
```
hcgatewood: Let's make this a service config, since e.g. it's conceivable that 15k might cause problems in the future and we'd want to have a one-stop way to clamp down the max

orc8r/cloud/go/services/configurator/doc.go
```diff
@@ -36,6 +36,12 @@
 	Code calling configurator should take care to define entity types and
 	relations which would invariably result in an acyclic graph.
 
+	Configurator supports token based pagination for entity loads. Clients can
```
hcgatewood, line 39: Let's also add a note on how to determine whether you have the last page. Because a client might be under the impression that if requestedPageSize > size(returnedPage), that they have the final page. But that's not actually true since we have a max page size in configurator.

But probably a better place for this is the LoadEntities doc. Basically just a note that, to exhaustively read all pages, you have to continue until you receive an empty page token in the response -- that's the only way to be sure

Also we should probably add, if don't have already, an explicit test for that invariant

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -390,3 +414,47 @@ func calculateEntitiesNotFound(entsByPk map[string]*NetworkEntity, requestedIDs
 	}
 	return ret
 }
+
+// getEntityLoadPageSize returns the maximum number of entities to return based
+// on the EntityLoadCriteria specified. A page size of 0 will default to the
+// maximum load size.
+func getEntityLoadPageSize(loadCriteria EntityLoadCriteria) int {
```
hcgatewood: Let's make this a method on the `EntityLoadCriteria` type

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -390,3 +414,47 @@ func calculateEntitiesNotFound(entsByPk map[string]*NetworkEntity, requestedIDs
...
+	if loadCriteria.PageSize == 0 {
+		return MaxEntityLoadSize
+	}
+	return util.MinInt(int(loadCriteria.PageSize), MaxEntityLoadSize)
+}
+
+func getNextPageToken(entities []*NetworkEntity) (string, error) {
```
hcgatewood, line 419: Let's make this a method on the `NetworkEntities` type, e.g. https://github.com/magma/magma/blob/5738d8c022dbaeeb677e6944eeb7ac5107ba6a65/orc8r/cloud/go/services/configurator/types.go#L255-L263

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -390,3 +414,47 @@ func calculateEntitiesNotFound(entsByPk map[string]*NetworkEntity, requestedIDs
...
+	lastEntity := entities[len(entities)-1]
+	nextPageToken := &EntityPageToken{Token: lastEntity.Key}
+	return serializePageToken(nextPageToken)
+}
+
+func serializePageToken(token *EntityPageToken) (string, error) {
```
hcgatewood, line 425: Same, let's make method on `*EntityPageToken` type

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -73,25 +93,29 @@ func (store *sqlConfiguratorStorage) getLoadEntitiesSelectBuilder(networkID stri
...
-			}
-			selectBuilder = selectBuilder.Where(andClause)
+		andClause := sq.And{sq.Eq{fmt.Sprintf("ent.%s", entNidCol): networkID}}
+		if filter.KeyFilter != nil {
+			andClause = append(andClause, sq.Eq{fmt.Sprintf("ent.%s", entKeyCol): filter.KeyFilter.Value})
+		}
+		if filter.TypeFilter != nil {
+			andClause = append(andClause, sq.Eq{fmt.Sprintf("ent.%s", entTypeCol): filter.TypeFilter.Value})
+		}
+		if criteria.PageSize != 0 {
```
hcgatewood: Should be `if pageToken != nil {`

And add regression test for that logic

orc8r/cloud/go/services/configurator/storage/sql.go
```diff
@@ -472,6 +472,13 @@ func (store *sqlConfiguratorStorage) LoadEntities(networkID string, filter Entit
...
 
+	// Set next page token when there may be more pages to return
+	if len(ret.Entities) == getEntityLoadPageSize(loadCriteria) {
+		ret.NextPageToken, err = getNextPageToken(ret.Entities)
+		if err != nil {
+			return ret, err
+		}
+	}
```
hcgatewood, line 483: Tiny nit: newline before final return for long fns

hcgatewood (changes requested):
Looks great. Couple small fixes then I'll be ready to merge

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
mpgermano: We can't. The type being used is a proto, so we can't make this a method.

orc8r/cloud/go/services/configurator/storage/storage.proto
mpgermano: An entity in configurator isn't always going to be a subscriber entity. I think this should stay generic.

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
mpgermano: Similarly this is also a proto from the storage proto (not the struct defined in `types.go`) so we can't make this a method (as much as I'd like to).

mpgermano:
Addressed comments.

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -73,25 +87,29 @@ func (store *sqlConfiguratorStorage) getLoadEntitiesSelectBuilder(networkID stri
...
+		}
+		if criteria.PageToken != "" {
+			andClause = append(andClause, sq.Gt{fmt.Sprintf("ent.%s", entKeyCol): pageToken.LastIncludedEntity})
 		}
-	}
 
-	return selectBuilder
+		selectBuilder = selectBuilder.Where(andClause)
+		selectBuilder = selectBuilder.OrderBy(fmt.Sprintf("ent.%s ASC", entKeyCol))
+		selectBuilder = selectBuilder.Limit(uint64(pageSize))
```
hcgatewood: Nit: these 3 lines can be coalesced to a `.Where(...).OrderBy(...).Limit(...)` single-assignment

orc8r/cloud/go/services/configurator/client_api.go
```diff
@@ -657,10 +657,10 @@ func DoesInternalEntityExist(entityType, entityKey string) (bool, error) {
 }
 
 // LoadAllEntitiesOfType fetches all entities of specified type in a network
```
hcgatewood: Let's also include that note about how to exhaustively read all pages here at the client api, since it's more visible for callers

orc8r/cloud/go/services/configurator/doc.go
```diff
@@ -36,6 +36,12 @@
...
+	specify a page size and token for a given entity type. The load response
+	will contain the entity page and a page token to be used on subsequent
+	requests. If a page size is not specified, configurator defaults to the
+	maximum supported page size (15k).
```
hcgatewood: Add short note that this max is configurable by service config

hcgatewood (approved):
Lgtm, please see 3 short comments before merging

themarwhal (approved):

PR #5134, opened 2021-02-24
[AGW] Fix coredump generation in local setup

themarwhal:

## Summary
1. fix coredump generation for local dev VM setup. We were not setting the ulimit and core_pattern properly, so doing that in ansible
2. For local devVM the dpkg command fails since there is not a magma package installed. I've changed it to a less nice option of just outputting a filtered `apt-cache policy magmag`. :'( 

## Test Plan
```
vagrant destroy magma
vagrant up magma
...
force MME to crash
...
vagrant@magma-dev:~/magma/lte/gateway$ ls -la /var/core
total 8844
drwxr-xr-x  2 root root    4096 Feb 24 22:21 .
drwxr-xr-x 13 root root    4096 Feb 24 22:21 ..
-rw-rw-rw-  1 root root 1295716 Feb 24 22:02 core-1614204134-sessiond-27822.tgz
-rw-rw-rw-  1 root root 2661162 Feb 24 22:04 core-1614204289-mme-29304.tgz
-rw-rw-rw-  1 root root 2499931 Feb 24 22:06 core-1614204409-mme-32224.tgz
-rw-rw-rw-  1 root root 2581291 Feb 24 22:21 core-1614205273-mme-10665.tgz
```

## Additional Information

### DIFF
--- a/lte/gateway/deploy/roles/magma/tasks/main.yml
+++ b/lte/gateway/deploy/roles/magma/tasks/main.yml
@@ -71,6 +71,18 @@
   copy: src=99-magma.conf dest=/etc/sysctl.d/99-magma.conf
   when: full_provision
 
+- name: Add the sysctl config from the step before
+  become: yes
+  command: 'sysctl -p /etc/sysctl.d/99-magma.conf'
+  when: full_provision
+
+- name: Set core ulimit to unlimited
+  pam_limits:
+    domain: '*'
+    limit_item: core
+    limit_type: '-'
+    value: unlimited
+
 - name: Create the /var/core directory
   file: path=/var/core state=directory
   when: full_provision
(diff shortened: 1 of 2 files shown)


lte/gateway/deploy/roles/magma/tasks/main.yml
```diff
@@ -67,9 +67,18 @@
...
   when: full_provision
 
-- name: Copy sysctl file for core dumps
-  copy: src=99-magma.conf dest=/etc/sysctl.d/99-magma.conf
-  when: full_provision
+- name: Set core pattern
+  sysctl:
+    name: kernel.core_pattern
+    value: '|/usr/local/bin/coredump core-%t-%e-%p'
+    state: present
```
pshelar: this already set in sysctl conf file, lets use it rather using sysclt cmd.

`sudo sysctl --system`
or
`sudo sysctl -p /etc/sysctl.d/99-magma.conf`

pshelar (changes requested):

lte/gateway/deploy/roles/magma/tasks/main.yml
themarwhal: ok let me try that

pshelar (approved):
LGTM

PR #5135, opened 2021-02-24
Make default drop flow stats reporting optional

koolzz:

## Summary
Some deployments don't care about this metric and it reduces the number of flows we install and the report size we send to sessiond

## Additional Information


koolzz:
Closing as this is not really needed

PR #5199, opened 2021-02-26
AGW: pipelined: qos: skip filter for APN ambr queue

pshelar:
QoS does not use apn root qos filter, avoid creating it.
This patch also increases FD limits on pipelineD service
this helps in scale test.

## Additional Information

### DIFF
--- a/lte/gateway/deploy/roles/magma/files/systemd/magma_pipelined.service
+++ b/lte/gateway/deploy/roles/magma/files/systemd/magma_pipelined.service
@@ -21,6 +21,9 @@ Wants=openvswitch-switch.service
 [Service]
 Type=simple
 PermissionsStartOnly=true
+LimitNOFILE=65536
+LimitNPROC=65536
+LimitSIGPENDING=65536
 EnvironmentFile=/etc/environment
 ExecStartPre=/usr/bin/ovs-vsctl set bridge gtp_br0 protocols=OpenFlow10,OpenFlow13,OpenFlow14 other-config:disable-in-band=true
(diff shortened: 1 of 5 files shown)


lte/gateway/python/magma/pipelined/qos/qos_meter_impl.py
```diff
@@ -72,7 +72,7 @@ def add_qos(self, _, qos_info: QosInfo, parent=None) -> int:
         LOG.debug("Adding meter_id %d", meter_id)
         return meter_id
 
-    def remove_qos(self, meter_id: int, d, recovery_mode=False):
+    def remove_qos(self, meter_id: int, d, recovery_mode=False, _=False):
```
koolzz, line 75: _=False -> _

lte/gateway/deploy/roles/magma/files/systemd/magma_pipelined.service
```diff
@@ -21,6 +21,9 @@ Wants=openvswitch-switch.service
 [Service]
 Type=simple
 PermissionsStartOnly=true
+LimitNOFILE=49000
```
koolzz: why specifically 49000? maybe just 50000

koolzz (approved):
lgtm

lte/gateway/python/magma/pipelined/qos/qos_meter_impl.py
pshelar: it needs default value, since recovery_mode param has it.

ardzoht (approved):

PR #5201, opened 2021-02-26
[fix] Update where we store Swagger spec

AndyLKhuu:

## Summary
Currently the Swagger UI is broken due to an issue with Swagger specs not properly copying over to the production image when not existing on the host image. This PR fixes this issue by moving the Swagger specs into a separate directory and copying it over from the build image to the production image.

## Test Plan
- [x] Manually checked that Swagger UI is now working properly on builds even without specs in host 

## Additional Information

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,7 @@ orc8r/cloud/test_certs/*.srl
 orc8r/cloud/test_certs/*.pfx
 
 # Ignore generated Swagger files
-orc8r/swagger/specs
+orc8r/cloud/swagger/specs
 
 # Ignore apt/build cache
 .cache/*
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,7 @@ orc8r/cloud/test_certs/*.srl
 orc8r/cloud/test_certs/*.pfx
 
 # Ignore generated Swagger files
-orc8r/cloud/configs/swagger_specs
+orc8r/swagger/specs
 
 # Ignore apt/build cache
 .cache/*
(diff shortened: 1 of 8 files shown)


orc8r/cloud/docker/controller/Dockerfile
```diff
@@ -108,6 +108,9 @@ COPY --from=builder /${SWAGGER_FILES}/v1/index.html /var/opt/magma/static/swagge
 COPY --from=builder /${SWAGGER_FILES}/v1/swagger.yml /var/opt/magma/static/swagger/v1/spec/swagger.yml
 COPY --from=builder /build/bin /var/opt/magma/bin
 
+## Swagger specs
+COPY --from=builder src/magma/orc8r/swagger /etc/magma/swagger
```
hcgatewood: This is also a build artifact. So just remove the comment and include it with the previous code paragarph

orc8r/lib/go/service/config/service_config.go
```diff
@@ -33,7 +33,7 @@ var (
 	configDir         = "/etc/magma/configs"
 	oldConfigDir      = "/etc/magma"
 	configOverrideDir = "/var/opt/magma/configs"
-	specDir           = "/etc/magma/configs/orc8r/swagger_specs"
+	specDir           = "/etc/magma/swagger/specs"
```
hcgatewood: Let's move all the Swagger-related logic out of lib/.../service_config.go and into the relevant cloud swagger pkg -- since specs aren't related to the config directory anymore

.gitignore
```diff
@@ -49,7 +49,7 @@ orc8r/cloud/test_certs/*.srl
 orc8r/cloud/test_certs/*.pfx
 
 # Ignore generated Swagger files
-orc8r/cloud/configs/swagger_specs
+orc8r/swagger/specs
```
hcgatewood: Let's actually do orc8r/cloud/swagger/specs, just to clarify that this is for cloud not gateway

hcgatewood (changes requested):
Thanks for fixing. Three small organizational fixes then let's merge the fix

hcgatewood (approved):
👍

PR #5205, opened 2021-02-26
Dockerfile for Ubuntu 20.04 capable of mme-build-from-source

electronjoe:
## Pinning of Versions

Relative to the existing Dockerfile, this dockerfile has fewer pinned versions / build from source dependencies.  I'm not strongly on one side of the camp or the other on this front - but some of our depedency pins have become super difficult to carry forward with modern distros.  I think exposing these dependencies and moving them forward or removing them is important - and unpinned dependencies can highlight that.  But we also don't want Docker builds breaking CI when a dependency moves. So I suspect we'll want a hybrid of prebuilt docker fetch to "pin" daily CI behavior, and ~nightly or weekly unpinned Dockerfile builds to see if any dependency movement is breaking us.

## Open Questions

- Relative to Dockerfile.Ubuntu18.04
  - Immediate objective is build and test - not execution
  - Trimmed the MME execution / config file / etc setup (but we should perhaps pull in)
- I discovered magma/third_party/build/bin
  - Wonder if I should be using these build scripts wherever possible?
  - I did end up using libfluid_build.sh
  - Came across this too late - got many artifacts building myself
- nlohmann-json-dev
  - Not available in ubuntu20.04 package repo - I migrated to nlohmann-json3-dev
- We should fixup MconfigLoader.cpp include path for json.hpp
  - `ln -s /usr/include/nlohmann/json.hpp /usr/include/json.hpp` fixup in image
- Can we remove the CMake repo additions `Add required package repository for CMake`
  - As it was insufficient for 20.04 and I ended up adding `deb http://security.ubuntu.com/ubuntu xenial-security main`
- Was my addition of `deb http://security.ubuntu.com/ubuntu xenial-security main` lacking pgp / verification? 

## Dependency Thoughts

- libfluid use
    - No maintenance for many years
    - There is a Russian Research group with a fork 30+ commits ahead (many fixes?)
 - We use Nettle 2.5 pinned
   -  Took Def hackery to build due to libc changes in libc of Ubuntu 20.04 - [Explanation](https://github.com/rdslw/openwrt/blob/e5d47f32131849a69a9267de51a30d6be1f0d0ac/tools/bison/patches/110-glibc-change-work-around.patch)
   - We should look at migrating to Nettle 3.5.1
     - See my attempt [Here](https://gist.github.com/electronjoe/a899e4bfbc2904cb353444386296c38e)
     - Need to fixup our API calls to migrate
- Libfolly
  - `Because folly does not provide any ABI compatibility guarantees from commit to commit, we generally recommend building folly as a static library.`
  - Not sure if our shared library use is of concern?

## Test Plan

```shell
cd lte/gateway/docker/mme/
docker build -t magma-mme-build -f Dockerfile.ubuntu20.04 ../../../../
docker run -i -t magma-mme-build:latest /bin/bash
cd magma/lte/gateway
make test_oai
make build_oai
```

### DIFF
--- a/lte/gateway/docker/mme/Dockerfile.ubuntu20.04
+++ b/lte/gateway/docker/mme/Dockerfile.ubuntu20.04
@@ -1,0 +1,221 @@
+################################################################
+# Builder Image (can also be used as developer's image)
+################################################################
+FROM ubuntu:focal as magma-mme-builder
+
+ARG GIT_PROXY
+ARG FEATURES=mme_oai
+ENV MAGMA_ROOT=/magma
+ENV BUILD_TYPE=Debug
+ENV C_BUILD=/build/c
+ENV TZ=Europe/Paris
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN mkdir -p $C_BUILD
+
+RUN ["/bin/bash", "-c", "echo \"Install general purpouse packages\" && \
+    apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get upgrade -y && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y gnupg wget software-properties-common autoconf automake \
+    libtool curl make g++ unzip git build-essential autoconf libtool pkg-config \
+    gcc-9 g++-9 apt-transport-https ca-certificates apt-utils vim redis-server tzdata \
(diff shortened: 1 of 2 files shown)


rdefosse:
MAGMA-OAI-TESTS

electronjoe:
Deprecated by merge of another PR containing dockerfile.

PR #5212, opened 2021-02-27
[subscriberdb] implement purgeUE

uri200:

## Summary

PurgeUE is not being implemented by subscriber db because it is not used on non federated case. But when testing federated case purgeUE was being called

Purge UE call is a 3gpp specific call to inform HSS that MME has dropped the information about this subscriber. In case of subscriberdb there is no action to be done, so we just return no error to this call

## Test Plan

make test_python

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/subscriberdb/protocols/s6a_proxy_servicer.py
+++ b/lte/gateway/python/magma/subscriberdb/protocols/s6a_proxy_servicer.py
@@ -19,6 +19,8 @@ from magma.subscriberdb.store.base import SubscriberNotFoundError
 
 from feg.protos import s6a_proxy_pb2, s6a_proxy_pb2_grpc
 
+from google.protobuf.json_format import MessageToJson
+
 
 class S6aProxyRpcServicer(s6a_proxy_pb2_grpc.S6aProxyServicer):
     """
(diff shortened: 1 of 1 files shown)


ardzoht (approved):

lte/gateway/python/magma/subscriberdb/protocols/s6a_proxy_servicer.py
```diff
@@ -127,6 +129,14 @@ def UpdateLocation(self, request, context):
 
         return ula
 
+    def PurgeUE(self, request, context):
+        self._print_grpc(request)
+        logging.warning(
+            "Purge request not implemented: {} {}".format(
```
ardzoht: minor nit: use lazy formatting for logging, lte-test is failing because of this

```
W1202: 135, 12: Use lazy % formatting in logging functions (logging-format-interpolation)
```

ardzoht (approved):
lgtm, please take a look at comment for logging

lte/gateway/python/magma/subscriberdb/protocols/s6a_proxy_servicer.py
uri200: done

PR #5219, opened 2021-03-01
[orc8r] Shift eventd handler attachment from orc8r to eventd

AndyLKhuu:

## Summary
Eventd handlers were being attach through the orchestrator service rather than by itself. This was mainly due to the fact that until recently eventd wasn't converted into a true service with it's own kubernetes labels. This PR migrates ownership of the attachment of eventd handlers back to eventd from orc8r.
Screen Shot 2021-03-02 at 1 15 18 PM

## Test Plan
- [x] Manually Spinning up Swagger UI and testing that eventd handler endpoints still work
- [x] Spin up service containers on Minikube and test that event endpoints are working

The 500 error is expected as elastic is not running.
Screen Shot 2021-03-02 at 2 22 22 PM

Double Checked to make sure Networks Path is still working.

Screen Shot 2021-03-02 at 5 43 45 PM

Screen Shot 2021-03-02 at 5 43 37 PM

 

## Additional Information


orc8r/cloud/go/services/eventd/obsidian/handlers/handlers.go
```diff
@@ -53,8 +56,45 @@ const (
...
+// GetObsidianHandlers returns all the obsidian handlers for eventd.
+func GetObsidianHandlers() []obsidian.Handler {
+	var ret []obsidian.Handler
+
+	// Elastic
+	elasticConfig, err := config.GetServiceConfig(orc8r.ModuleName, "elastic")
+	if err != nil {
+		ret = append(ret, obsidian.Handler{Path: LogSearchQueryPath, Methods: obsidian.GET, HandlerFunc: getInitErrorHandler(err)})
+		ret = append(ret, obsidian.Handler{Path: LogCountQueryPath, Methods: obsidian.GET, HandlerFunc: getInitErrorHandler(err)})
+		ret = append(ret, obsidian.Handler{Path: EventsPath, Methods: obsidian.GET, HandlerFunc: getInitErrorHandler(err)})
```
karthiksubraveti: minor nit: we can do an early return in this if block and unindent the else block

orc8r/cloud/go/services/eventd/obsidian/handlers/handlers.go
```diff
@@ -53,8 +56,45 @@ const (
...
+	} else {
+		elasticHost := elasticConfig.MustGetString("elasticHost")
+		elasticPort := elasticConfig.MustGetInt("elasticPort")
+
+		client, err := elastic.NewSimpleClient(elastic.SetURL(fmt.Sprintf("http://%s:%d", elasticHost, elasticPort)))
+		if err != nil {
+			ret = append(ret, obsidian.Handler{Path: LogSearchQueryPath, Methods: obsidian.GET, HandlerFunc: getInitErrorHandler(err)})
```
karthiksubraveti: minor nit: setting the error handling can be done in a loop, might be cleaner that way.

orc8r/cloud/go/services/eventd/obsidian/handlers/handlers.go
```diff
@@ -53,8 +56,45 @@ const (
...
 )
 
+// GetObsidianHandlers returns all the obsidian handlers for eventd.
+func GetObsidianHandlers() []obsidian.Handler {
+	var ret []obsidian.Handler
+
+	// Elastic
+	elasticConfig, err := config.GetServiceConfig(orc8r.ModuleName, "elastic")
+	if err != nil {
+		ret = append(ret, obsidian.Handler{Path: LogSearchQueryPath, Methods: obsidian.GET, HandlerFunc: getInitErrorHandler(err)})
```
karthiksubraveti: Also we should probably be setting error handlers for all the paths, I understand this is old code, but I guess this might not have been updated.

orc8r/cloud/go/services/eventd/obsidian/handlers/handlers.go
```diff
@@ -253,6 +293,12 @@ const (
 	pathParamEnd         = "end"
 )
 
+func getInitErrorHandler(err error) func(c echo.Context) error {
+	return func(c echo.Context) error {
+		return obsidian.HttpError(fmt.Errorf("initialization Error: %v", err), 500)
```
karthiksubraveti: minor nit: we should probably move our error codes to a constants file and use that instead of using 500 here.

karthiksubraveti (commented):
Most changes LGTM. Left a few comments.

orc8r/cloud/go/services/eventd/obsidian/handlers/handlers.go
```diff
@@ -53,8 +56,40 @@ const (
...
+		ret = append(ret, setInitErrorHandlers(err)...)
+		return ret
+	}
+
+	elasticHost := elasticConfig.MustGetString("elasticHost")
+	elasticPort := elasticConfig.MustGetInt("elasticPort")
+
+	client, err := elastic.NewSimpleClient(elastic.SetURL(fmt.Sprintf("http://%s:%d", elasticHost, elasticPort)))
+	if err != nil {
+		ret = append(ret, setInitErrorHandlers(err)...)
```
hcgatewood, line 81: Make this an early return instead of an if-else

hcgatewood (changes requested):
You'll also need to update the annotations in relevant service_registry.yml and values.yml files

Give it a sanity check on Minikube as well -- it's fine if it returns an error (depending on whether elastic is up in the cluster), but if it 404s that means the service mesh labels/annotations weren't correct

hcgatewood (approved):
👍

hcgatewood (changes requested):
Oops realized we also need to update chart version @AndyLKhuu

hcgatewood (approved):

PR #5225, opened 2021-03-01
[pipelined]Add rpc call future timeout

koolzz:

## Additional Information


pshelar (commented):
Lets also fix SetSMFSessions()

pshelar (approved):
otherwise looks good.

PR #5232, opened 2021-03-01
[cloud] [agw] AGW AMI deployment for Cloudstrapper

arunuke:
## Summary

AMI generation and in-region deployment of AGW AMIs on AWS.

## Test Plan

Tested and validated. Demo available at https://www.youtube.com/watch?v=sWZoopmEJrM&list=PLKqaoAnDyfgpCPTI9xDxxx2vuJsB1871N&index=9

## Additional Information

None


experimental/cloudstrapper/playbooks/roles/agw-infra/files/cfnMagmaAgwPublic.json
```diff
@@ -0,0 +1,55 @@
+{
+    "AWSTemplateFormatVersion": "2010-09-09",
+    "Parameters": {
+      "paramSubnetSgi": {
+        "Type": "String",
+	"Description": "SGi Subnet",
+	"Default": "subnet-0884f3b19999bf3ad"
```
mpgermano: For my understanding, are these defaults picked arbitrarily or is this set by AWS?

experimental/cloudstrapper/playbooks/roles/agw-infra/vars/varGatewayCupertinoAgwA.yaml
```diff
@@ -0,0 +1,10 @@
+---
+
+#Parameters for Cupertino Site AGW
+
+
+idAgw: AgwA
```
mpgermano, line 6: I don't understand the purpose of having AgwA and AgwB, could you explain this a bit?

experimental/cloudstrapper/playbooks/roles/agw-infra/vars/varGatewayTokyoAgwA.yaml
```diff
@@ -0,0 +1,10 @@
+---
+
+#Parameters for Cupertino Site AGW
```
mpgermano: Nit: incorrect name `Cupertino -> Tokyo`

experimental/cloudstrapper/playbooks/roles/agw-infra/vars/varGatewayTokyoAgwB.yaml
```diff
@@ -0,0 +1,10 @@
+---
+
+#Parameters for Cupertino Site AGW
```
mpgermano: Same comment here

experimental/cloudstrapper/playbooks/roles/agw-platform/files/control_proxy.j2
```diff
@@ -0,0 +1,6 @@
+cloud_address: "{{ cloudController }}"
```
mpgermano, line 1: This file is missing two parameters:

```
fluentd_address: "{{ fluentdController }}"
fluentd_port: 24224
```

experimental/cloudstrapper/playbooks/roles/agw-platform/tasks/attach-nic.yml
```diff
@@ -0,0 +1,47 @@
...
+
+- name: provision agw base
+  cloudformation:
+    stack_name: "stkAgwAnsible"
+    state: "present"
+    template: "roles/agw-infra/files/cfnMagmaAgwPublic.json"
+    template_parameters:
+      paramImageBase: "{{ imageDebian }}"
+
+- name: provision jump server
```
mpgermano: For my understanding, can you explain the purpose of the jump server?

experimental/cloudstrapper/playbooks/roles/agw-platform/vars/main.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - magma@lighttpd.service
+  - magma@mobilityd.service
+  - magma@redis.service
+
+agwCertsPath: /var/opt/magma/tmp/certs/
+agwConfigPath: /var/opt/magma/configs
+
+cloudController: "controller.{{ orc8rDomainName }}"
+bootstrapperController: "bootstrapper-controller.{{ orc8rDomainName }}"
+
```
mpgermano: Add `fluentdController` here

experimental/cloudstrapper/playbooks/roles/agw-platform/tasks/main.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - "{{ servicesMagma }}"
+  ignore_errors: yes
+
+- name: check services again
+  systemd: 
+    state: started
+    name: "{{ item }}"
+  become: true
+  with_items:
+    - "{{ servicesMagma }}"
```
mpgermano: Shouldn't this be `servicesMagmaAll`

experimental/cloudstrapper/playbooks/roles/agw-platform/vars/main.yaml
```diff
@@ -0,0 +1,24 @@
+---
+
+servicesMagma:
+  - magma@magmad.service
+
+servicesMagmaAll:
```
mpgermano, line 6: I'm concerned this could get stale anytime a new gateway service is added. Can we instead try to check service status with something like `magma@*`to avoid this.

experimental/cloudstrapper/playbooks/roles/build-platform/files/helm-publish.bash
```diff
@@ -0,0 +1,16 @@
+#arg1: gitusername arg2: gitpassword arg3: git publish branch 
+gitBranchCheckout=a7580153
```
mpgermano: Where does this commit id come from? Can we use a version branch here instead.

experimental/cloudstrapper/playbooks/roles/build-platform/tasks/build-magma.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+- name: preparing build
+  pause:
+    minutes: 5
+  tags: buildPrep
+
+- name: primary build to generate orc8r artifacts
+  command: "chdir={{ magmaBuildDir }} {{ dirShim }}/python3.7 build.py --all"
+
+- name: primary build to generate lte artifacts
```
mpgermano: Nit: `lte -> NMS`

experimental/cloudstrapper/playbooks/roles/build-platform/tasks/publish-helm-v14.yaml
```diff
@@ -0,0 +1,13 @@
+---
+
+#TODO: presumes helm14 directory is available on github. use gh to check and create
+#
+
+- name: publish helm charts
+  command: "/usr/bin/bash package.sh -d {{ buildDeploymentType }} "
```
mpgermano: We should consider backporting the package script to v1.3 so that we could collapse these two files. Maybe something we can look at in the future.

experimental/cloudstrapper/playbooks/roles/control/files/v1.3/orc8r-aws/eks.tf
```diff
@@ -0,0 +1,74 @@
+################################################################################
```
mpgermano: I think this whole section of versioned `.tf` files is unnecessary. We just need to provide a templated main.tf for v1.3 and v1.4 Terraform allows setting a reference to a github repo which we set for example like this
```
module orc8r {
  source = "github.com/magma/magma//orc8r/cloud/deploy/terraform/orc8r-aws?ref=v1.4"
  # ...
}

module orc8r-app {
  # ...
}
```
this will use the tf files from the repo rather than those locally.

experimental/cloudstrapper/playbooks/roles/control/templates/main.tf.1.3.org
```diff
@@ -0,0 +1,87 @@
+################################################################################
```
mpgermano: I don't see the point of these `.org` files. Could you explain why we need these?

experimental/cloudstrapper/playbooks/roles/control/templates/main.tf.1.3.org
```diff
@@ -0,0 +1,87 @@
...
+
+  eks_cluster_id = module.orc8r.eks_cluster_id
+
+  efs_file_system_id       = module.orc8r.efs_file_system_id
+  efs_provisioner_role_arn = module.orc8r.efs_provisioner_role_arn
+
+  elasticsearch_endpoint = module.orc8r.es_endpoint
+
+  orc8r_chart_version = "1.4.7"
+  orc8r_tag           = "1.0.1"
```
mpgermano: Tag here should be `1.3.0`

experimental/cloudstrapper/playbooks/roles/control/templates/main.tf
```diff
@@ -0,0 +1,89 @@
+################################################################################
```
mpgermano: Do we need this file. I think just the templated versions main.tf should do

mpgermano (changes requested):
Left a few questions and comments. Main thing is I think we can remove some redundancy in the terraform files added here. Also, I didn't see any of necessary security group rules needed for the AGW's (see https://docs.magmacore.org/docs/next/lte/ha_setup).

experimental/cloudstrapper/playbooks/roles/agw-infra/files/cfnMagmaAgwPublic.json
arunuke: These are default values that can be set by the user if doing repeated tests on the same Orchestrator deployment. I will look at removing defaults altogether so that they get picked up dynamically for every run (the code already does that)

experimental/cloudstrapper/playbooks/roles/agw-infra/vars/varGatewayCupertinoAgwA.yaml
arunuke: This is to demonstrate how to deploy multiple AGWs under the same orchestrator (not exclusively for the HA use-case alone). An example use-case would be stress testing the orchestrator with multiple gateways. 'AgwA' and 'AgwB' are used just as identifiers with some pre-baked configurations to help people get started.

experimental/cloudstrapper/playbooks/roles/agw-platform/tasks/attach-nic.yml
arunuke: This is vestige of an old version and I will delete this alongside a few other unused files. Thanks for catching this.

experimental/cloudstrapper/playbooks/roles/agw-infra/files/cfnMagmaAgwPublic.json
arunuke: The CloudFormation files are actually picked from the roles/cfn directory. Removing the "files" directory completely.

experimental/cloudstrapper/playbooks/roles/agw-infra/vars/varGatewayTokyoAgwA.yaml
arunuke: Fixed.

experimental/cloudstrapper/playbooks/roles/agw-infra/vars/varGatewayTokyoAgwB.yaml
arunuke: Fixed.

experimental/cloudstrapper/playbooks/roles/agw-platform/tasks/main.yaml
arunuke: Indeed :). Fixed.

experimental/cloudstrapper/playbooks/roles/build-platform/files/helm-publish.bash
arunuke: This is from the "build orchestrator" instructions from https://magma.github.io/magma/docs/orc8r/deploy_build. Given that this experimental feature will support only the currently available version and the upcoming version, the 1.3 code will get deprecated once 1.5 is released and removed.

experimental/cloudstrapper/playbooks/roles/agw-platform/vars/main.yaml
arunuke: Having a list of services specified in a variable allows us to check for practically any service on the gateway, even if it is not a magma service, but a native OS service (ex: firewalls). I agree that we do not have this use-case today, but this approach will allow us to be flexible on that front. WDYT?

experimental/cloudstrapper/playbooks/roles/build-platform/tasks/build-magma.yaml
arunuke: Fixed.

experimental/cloudstrapper/playbooks/roles/build-platform/tasks/publish-helm-v14.yaml
arunuke: My proposal is that we only support the "current" release (the stable one) and the "upcoming" release. This allows users to do three key use-cases.

a. Deploy a stable release
b. Deploy a new release
c. Upgrade from current release to new release.

WDYT?

experimental/cloudstrapper/playbooks/roles/control/files/v1.3/orc8r-aws/eks.tf
arunuke: I put in this file just to overcome some of the fixes that hadn't made it to the v1.3 release when I was building the tool. Once backported to 1.3, we can simply remove the call to patch-terraform.yaml altogether. For example, 1.4 does not have any patching requirement since the code works out of the box.

Since 1.3 would hang around only till 1.5, I will leave this as-is. Although a bit kludgy, it deploys 1.3 as expected and is used primarily for upgrade testing only at this time. 

WDYT?

experimental/cloudstrapper/playbooks/roles/control/templates/main.tf.1.3.org
arunuke: I left the .org files to indicate what a base version of the file looks like for the user to compare. I will remove these files.

experimental/cloudstrapper/playbooks/roles/control/templates/main.tf.1.3.org
arunuke: File removed.

experimental/cloudstrapper/playbooks/roles/control/templates/main.tf
arunuke: Agree. File removed.

experimental/cloudstrapper/playbooks/roles/agw-platform/files/control_proxy.j2
arunuke: Added.

experimental/cloudstrapper/playbooks/roles/agw-platform/vars/main.yaml
arunuke: Added.

experimental/cloudstrapper/playbooks/roles/build-platform/tasks/publish-helm-v14.yaml
mpgermano: Yeah I think this sounds reasonable to me.

experimental/cloudstrapper/playbooks/roles/agw-platform/vars/main.yaml
mpgermano: Yeah that's reasonable

experimental/cloudstrapper/playbooks/roles/control/files/v1.3/orc8r-aws/eks.tf
mpgermano: Yeah that makes sense. Can you let me know which specific changes aren't in v1.3 that are needed? I can work on backporting these.

Also, I think we should be able to remove all of the files in `experimental/cloudstrapper/playbooks/roles/control/files/v1.4/...` given that it's working without needing any changes.

mpgermano (commented):
Looks good. Just the one comment about removing the v1.4 tf files.

experimental/cloudstrapper/playbooks/roles/control/files/v1.3/orc8r-aws/eks.tf
arunuke: The fixes that need a backport to v1.3 are
1. orc8r-aws: Needs to specify cluster_version in eks.tf to 1.17 or higher.
2. orc8r-helm-aws: Needs to specify stable_helm_repo and incubator_helm_repo to charts.helm.sh

arunuke:
> Looks good. Just the one comment about removing the v1.4 tf files.

Removed v1.4 files since they are not applicable anymore. Added information on fixes that need to be backported to 1.3.

mpgermano (approved):

PR #5241, opened 2021-03-02
Merge Swagger Lib into Swagger Obsidian Service

AndyLKhuu:

## Summary
This PR coalesces the the Swagger library into Obsidian's Swagger directory. The main reason for this change is due to the fact that the Swagger Lib is only utilized by the Swagger UI components in one place. There's no need to separate these tightly coupled directories. This PR merges the contents of Swagger Lib into the obsidian service.

## Test Plan
- [x] Manual Inspection of Swagger UI working
 

## Additional Information


hcgatewood (changes requested):
I don't like the `mswagger` pkg -- let's just move `swagger/mswagger` contents directly into `swagger` pkg

If you think that makes the pkg too busy, an alternative option is to move the stuff directly operating on / merging swagger specs (basically the swagger_lib contents) into a `swagger/spec` pkg

AndyLKhuu:
> I don't like the `mswagger` pkg -- let's just move `swagger/mswagger` contents directly into `swagger` pkg
> 
> If you think that makes the pkg too busy, an alternative option is to move the stuff directly operating on / merging swagger specs (basically the swagger_lib contents) into a `swagger/spec` pkg

Sounds good, the main reason why I went for the mswagger directory in the first place was because the directory look extremely crowded. Adding a spec directory seems to help. I'll push an update shortly.

hcgatewood (approved):
👍

PR #5242, opened 2021-03-02
[feg] s6a_proxy modify Regional Zone Code from String to []byte

### LINKED ISSUE #4943: FEG support for TAC restriction

uri200:

## Summary

Changed format for Regional Sub Zone Code from string to []byte so MME can easily compare it with TAC comming from enB

## Additional Information


feg/protos/s6a_proxy.proto
```diff
@@ -168,7 +168,7 @@ message UpdateLocationAnswer {
 
     NetworkAccessMode network_access_mode = 8;
 
-    repeated string regional_subscription_zone_code = 9;
```
emakeev, line 171: Is it already being used by a deployed FeG/AGW? If yes, it may not be a safe change.
A backward compatible change would be adding a new binary field: repeated bytes regional_subscription_zone_code_b = 10;
and switch to using it over time by supporting both fields until all AGWs are upgraded

emakeev (approved):

PR #5247, opened 2021-03-02
[lte][agw] Adding write-through cache to RedisFlatDict, removing few locking on read mobilityd functions

### LINKED ISSUE #4830: [lte][agw] Improving mobilityd IP states storage maps

ardzoht:

## Summary

- This adds a writeback option to RedisFlatDict on py containers, which will maintain a local cache to read and write data, which will result in saving multiple redis read operations.
- Removing some locking on `get_sid_from_ip` and `get_ip_from_sid` since these are read operations which will not modify data
- Reformatting a few files on mobilityd

## Test Plan

- make integ_test
- make test
- testing allocate IP and release IP requests with `ghz` load testing tool 
image

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/mobilityd/ip_address_man.py
+++ b/lte/gateway/python/magma/mobilityd/ip_address_man.py
@@ -51,13 +51,11 @@ from typing import List, Optional, Tuple
 
 from lte.protos.mobilityd_pb2 import GWInfo, IPAddress
 
-from magma.mobilityd.ip_descriptor import IPState, IPType
+from magma.mobilityd.ip_descriptor import IPState
 from magma.mobilityd.metrics import (IP_ALLOCATED_TOTAL, IP_RELEASED_TOTAL)
-
 from .ip_allocator_base import DuplicateIPAssignmentError, IPAllocator, \
     IPNotInUseError, \
     MappingNotFoundError
-
 from .mobility_store import MobilityStore
 
 DEFAULT_IP_RECYCLE_INTERVAL = 15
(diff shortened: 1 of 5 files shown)


lte/gateway/python/magma/mobilityd/ip_address_man.py
```diff
@@ -333,21 +331,19 @@ def get_sid_ip_table(self) -> List[Tuple[str, ip_address]]:
...
     def get_ip_for_sid(self, sid: str) -> Optional[ip_address]:
         """ if ip is mapped to sid, return it, else return None """
-        with self._lock:
-            if sid in self._store.sid_ips_map:
-                if not self._store.sid_ips_map[sid]:
-                    raise AssertionError("Unexpected internal state")
-                else:
-                    return self._store.sid_ips_map[sid].ip
-            return None
+        if sid in self._store.sid_ips_map:
```
themarwhal: Could we have a case where there is modification between 
`if sid in self._store.sid_ips_map:` and `if not self._store.sid_ips_map[sid]:`? where sid no longer exists? 
Or do we only read from redis the first time and not the second?

orc8r/gateway/python/magma/common/redis/containers.py
```diff
@@ -125,9 +128,9 @@ def deserialize_key(serialized):
         return serialized.decode('utf-8')  # Redis returns bytes
 
     def __init__(
-        self, client, key, serialize, deserialize,
-        default_factory=None,
-        writeback=False,
+            self, client, key, serialize, deserialize,
+            default_factory=None,
+            writeback=False,
```
themarwhal: Can we add doc strings for default_factory and writeback?

orc8r/gateway/python/magma/common/redis/containers.py
```diff
@@ -247,6 +268,10 @@ def __getitem__(self, key: str) -> T:
         if ':' in key:
             raise ValueError("Key %s cannot contain ':' char" % key)
         composite_key = self._make_composite_key(key)
+
+        if self._writeback:
+            return self.cache[composite_key]
```
themarwhal: Does this throw a `KeyError` if the key does not exist?

orc8r/gateway/python/magma/common/redis/containers.py
```diff
@@ -265,6 +290,8 @@ def __setitem__(self, key: str, value: T) -> Any:
         version = self.get_version(key)
         serialized_value = self.serde.serialize(value, version + 1)
         composite_key = self._make_composite_key(key)
+        if self._writeback:
+            self.cache[composite_key] = value
         return self.redis.set(composite_key, serialized_value)
```
pshelar, line 298: since you are still setting the value in redis, this is write-through cache. am I missing someting?

ulaskozat:
are there any race conditions due to the removed locks (see the comment by @themarwhal)?

lte/gateway/python/magma/mobilityd/ip_address_man.py
ardzoht: Yeah I don't think we have a condition where we can hit this case, I updated the function to do the first check which will work the same.

ardzoht:
@ulaskozat There's no race condition on the removed locks on the functions, @themarwhal comments were related with two checks that were done on the same function, but there are no modification between the checks.

pshelar (approved):
LGTM

mpgermano (approved):
Looks good.

PR #5255, opened 2021-03-02
pipelined: classifier test failure #5239

prabinakpattnaik:
## Summary

The issue was reproduced with exception errors and two failed test cases while running the "test_classifier.py" when 
Port "g_*" and Interface "g_*" already attached to gtp_br0 bridge, like as below:

Bridge "gtp_br0"
        Controller "tcp:127.0.0.1:6633"
        Controller "tcp:127.0.0.1:6654"
        fail_mode: secure
        Port "g_913ca8c0"
            Interface "g_913ca8c0"
                type: gtp
                options: {key=flow, remote_ip="192.168.60.145"}

Steps for issue reproduction : 
1. Run the 4G testcase () Like : make integ_test TESTS=s1aptests/test_attach_ul_udp_data.py    
2. Once the testcase is finished stop the magma services.
3. Execute the testcase (test_classifier.py).
4. Observation is that the testcase fails. 

Reason being the integ_test does not cleans up the bridge and attached ports. Hence the port
numbering (for stored snapshot) mismatches.

For fix the exception, need to convert the htonl of Gnb_ip address before convert to port_name "g_*".

For test cases issue fix, The following approaches colud help to avoid the issues:
1) Need to restart the magma before running the test cases
2) Or need to put precondition if "g_*" port name is present then skip the test cases.
3) Or "ovs_multi_tunnel" flag set to false in "test_classifier.py" and "in_port" value will be "32768"

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/classifier.py
+++ b/lte/gateway/python/magma/pipelined/app/classifier.py
@@ -12,6 +12,7 @@ limitations under the License.
 """
 import subprocess
 import ipaddress
+import socket
 from collections import namedtuple
 from ryu.ofproto.ofproto_v1_4 import OFPP_LOCAL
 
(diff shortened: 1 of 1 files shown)


lte/gateway/python/magma/pipelined/app/classifier.py
```diff
@@ -12,6 +12,7 @@
 """
 import subprocess
 import ipaddress
```
ShubhamTatvamasi, line 14: Do we still need `import ipaddress` if we are not using it anywhere?

pshelar: it is still used after changes.

pshelar (approved):

PR #5262, opened 2021-03-02
[MME][Memory] Enable UBSAN for debug builds and resolve unaligned memory access issues

talkhasib:

## Summary

Enable UBSAN for debug builds (sanitize=undefined)
Once enabled, it uncovered many unaligned memory access problems.
- ZMQ received message needs to be copied to avoid unaligned memory access.
- Use memcpy() instead of casting memory addresses in few places
- Potential overflow when performing left shifts on bytes

sample logs
```
intertask_interface.c:407:48: runtime error: member access within misaligned address 0x629000163206 for type 'struct MessageDef', which requires 8 byte alignment
mme_app_main.c:78:11: runtime error: member access within misaligned address 0x629000163206 for type 'struct MessageDef', which requires 8 byte alignment
nas_message.c:729:9: runtime error: load of misaligned address 0x60200003c991 for type 'uint32_t', which requires 4 byte alignment
nas_stream_eia1.c:194:31: runtime error: load of misaligned address 0x603000ec0ef5 for type 'uint32_t', which requires 4 byte alignment
```

## Test Plan
fab integ_test

### DIFF
--- a/lte/gateway/c/oai/common/async_system.c
+++ b/lte/gateway/c/oai/common/async_system.c
@@ -56,10 +56,8 @@ static void async_system_exit(void);
 task_zmq_ctx_t async_system_task_zmq_ctx;
 
 static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
-  int rc              = 0;
-  zframe_t* msg_frame = zframe_recv(reader);
-  assert(msg_frame);
-  MessageDef* received_message_p = (MessageDef*) zframe_data(msg_frame);
+  int rc                         = 0;
+  MessageDef* received_message_p = receive_msg(reader);
 
   switch (ITTI_MSG_ID(received_message_p)) {
     case ASYNC_SYSTEM_COMMAND: {
@@ -86,7 +84,7 @@ static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
 
     case TERMINATE_MESSAGE: {
       itti_free_msg_content(received_message_p);
-      zframe_destroy(&msg_frame);
+      free(received_message_p);
       async_system_exit();
     } break;
 
@@ -98,7 +96,7 @@ static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
   }
 
   itti_free_msg_content(received_message_p);
-  zframe_destroy(&msg_frame);
+  free(received_message_p);
   return 0;
 }
 
--- a/lte/gateway/c/oai/common/common_defs.h
+++ b/lte/gateway/c/oai/common/common_defs.h
@@ -38,6 +38,7 @@
 #ifndef FILE_COMMON_DEFS_SEEN
 #define FILE_COMMON_DEFS_SEEN
 
+#include <string.h>  // memcpy
 #include <arpa/inet.h>
 //------------------------------------------------------------------------------
 #define STOLEN_REF
@@ -105,15 +106,18 @@ typedef enum {
   sIZE += sizeof(uint8_t)
 
 #define DECODE_U16(bUFFER, vALUE, sIZE)                                        \
-  vALUE = ntohs(*(uint16_t*) (bUFFER));                                        \
+  memcpy((unsigned char*) &vALUE, bUFFER, sizeof(uint16_t));                   \
+  vALUE = ntohs(vALUE);                                                        \
   sIZE += sizeof(uint16_t)
 
 #define DECODE_U24(bUFFER, vALUE, sIZE)                                        \
-  vALUE = ntohl(*(uint32_t*) (bUFFER)) >> 8;                                   \
+  memcpy((unsigned char*) &vALUE, bUFFER, sizeof(uint32_t));                   \
+  vALUE = ntohl(vALUE) >> 8;                                                   \
   sIZE += sizeof(uint8_t) + sizeof(uint16_t)
 
 #define DECODE_U32(bUFFER, vALUE, sIZE)                                        \
-  vALUE = ntohl(*(uint32_t*) (bUFFER));                                        \
+  memcpy((unsigned char*) &vALUE, bUFFER, sizeof(uint32_t));                   \
+  vALUE = ntohl(vALUE);                                                        \
   sIZE += sizeof(uint32_t)
 
 #if (BYTE_ORDER == LITTLE_ENDIAN)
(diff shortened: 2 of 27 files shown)


ssanadhya (approved):

lte/gateway/c/oai/common/common_defs.h
```diff
@@ -105,15 +106,18 @@ typedef enum {
...
 #define DECODE_U16(bUFFER, vALUE, sIZE)                                        \
-  vALUE = ntohs(*(uint16_t*) (bUFFER));                                        \
+  memcpy((unsigned char*) &vALUE, bUFFER, sizeof(uint16_t));                   \
+  vALUE = ntohs(vALUE);                                                        \
   sIZE += sizeof(uint16_t)
 
 #define DECODE_U24(bUFFER, vALUE, sIZE)                                        \
-  vALUE = ntohl(*(uint32_t*) (bUFFER)) >> 8;                                   \
+  memcpy((unsigned char*) &vALUE, bUFFER, sizeof(uint32_t));                   \
+  vALUE = ntohl(vALUE) >> 8;                                                   \
```
ulaskozat, line 115: This doesn't look correct to me if network byte order and host byte order are different. E.g.,
buffer -> byte-1, byte-2, byte-3, byte-4, byte-5, ....
Line 114: vALUE -> byte-1, byte-2, byte-3, byte-4
Line 115: nothl(vALUE) -> byte-4, byte-3, byte-2, byte-1  
                ntohl(vALUE) >> 8 -> byte-4, byte-3, byte-2

instead the target was to have byte-3, byte-2, byte-1. Do I see this wrong?

ulaskozat, line 115: should be like below:
```
memcpy((unsigned char*) &vALUE, bUFFER, 4);
vALUE = vALUE >> 8;                   
vALUE = ntohl(vALUE);
```

lte/gateway/c/oai/common/common_defs.h
```diff
@@ -131,16 +135,25 @@ typedef enum {
...
+    memcpy(buffer, (unsigned char*) &n_value, sizeof(uint16_t));               \
+    size += sizeof(uint16_t);                                                  \
+  } while (0)
 
 #define ENCODE_U24(buffer, value, size)                                        \
-  *(uint32_t*) (buffer) = htonl(value);                                        \
-  size += sizeof(uint8_t) + sizeof(uint16_t)
+  do {                                                                         \
+    uint32_t n_value = htonl(value);                                           \
+    memcpy(buffer, (unsigned char*) &n_value, sizeof(uint32_t));               \
```
ulaskozat, line 147: should not we just copy 3 bytes?

ulaskozat (approved):
LGTM

lte/gateway/c/oai/common/async_system.c
```diff
@@ -56,10 +56,8 @@ static void async_system_exit(void);
 task_zmq_ctx_t async_system_task_zmq_ctx;
 
 static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
-  int rc              = 0;
-  zframe_t* msg_frame = zframe_recv(reader);
-  assert(msg_frame);
-  MessageDef* received_message_p = (MessageDef*) zframe_data(msg_frame);
+  int rc                         = 0;
+  MessageDef* received_message_p = receive_msg(reader);
```
electronjoe, line 60: Do we understand why this buffer is unaligned? By fixing it here are we fixing it everywhere? Or is it possibly unaligned in a third party library etc? Wondering if this is hiding the root cause (maybe not, but asking the question).

lte/gateway/c/oai/common/async_system.c
```diff
@@ -86,7 +84,7 @@ static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
 
     case TERMINATE_MESSAGE: {
       itti_free_msg_content(received_message_p);
-      zframe_destroy(&msg_frame);
```
electronjoe, line 89: Here and elsewhere. Are we certain that zframe_destroy can be substituted with free() here? I think without a comment that's not clear to me. E.g. our Bstring library b_destroy actually frees two allocations - and if we moved to free() we would leak.

lte/gateway/c/oai/common/common_defs.h
```diff
@@ -105,15 +106,18 @@ typedef enum {
   sIZE += sizeof(uint8_t)
 
 #define DECODE_U16(bUFFER, vALUE, sIZE)                                        \
-  vALUE = ntohs(*(uint16_t*) (bUFFER));                                        \
+  memcpy((unsigned char*) &vALUE, bUFFER, sizeof(uint16_t));                   \
```
electronjoe, line 109: Same question as above - do we *expect* that these buffers might be unaligned? If so then no surprises here and great. In this particular fixup I could see it being less surprising - if this is say parsing a uint16_t that came after a type field that was a byte or whatever.

lte/gateway/c/oai/common/common_defs.h
electronjoe: This macro is a bit odd too - because as far as I can see in code search, it always receives a uint32_t and then packs all four bytes here. So it's really not an ENCODE_U24 its an ENCODE_U32... At least for the call sites I am spotting.

lte/gateway/c/oai/lib/secu/nas_stream_eia1.c
```diff
@@ -117,12 +118,13 @@ int nas_stream_encrypt_eia1(
...
   uint64_t M_D_2;
   int rem_bits;
-  uint32_t mask = 0;
-  uint32_t* message;
+  uint32_t mask     = 0;
+  uint32_t* message = (uint32_t*) malloc(
+      sizeof(uint32_t) * ((stream_cipher->blength >> 5) + 1));
+
+  /* copy message to avoid memory alignment issues */
+  memcpy(message, stream_cipher->message, ceil(stream_cipher->blength / 8.0));
```
electronjoe, line 126: Again, wondering if alignment issue here was expected or not?

electronjoe (commented):
As a top level concern, my experience with our CMake configs has me worried that other libraries that may be handling these buffers may not have UBSAN turned on - and it's possible that we might be fixing the issue in some MAGMA code but not detecting it at root cause / in other areas of the codebase.  CMake *badly* needs to be fixed up IMO. See questions in review!

rdefosse:
MAGMA-OAI-TESTS

lte/gateway/c/oai/common/async_system.c
talkhasib: ZMQ does not care about the data structure and treats it as a byte buffer. Thus, alignment is not an issue inside ZMQ.
When we cast back to the message structure at the receiving end. That caused the misalignment. Copying the data resolves this issue.

lte/gateway/c/oai/common/async_system.c
talkhasib: free does not replace zframe_destroy(). We still do a zframe_destroy()
 inside the receive_msg function after we copy the message body. We free the message that receive_msg() created.
itti_free_msg_content() handels freeing second level data allocations inside messages. Thats why we need to call both.

lte/gateway/c/oai/common/common_defs.h
talkhasib: Buffers are mainly packed data (IEs) received over the wire. Thus, no guarantee on alignment

lte/gateway/c/oai/common/common_defs.h
talkhasib: I noticed the oddness of this and kept the logic intact for the purpose of this PR

lte/gateway/c/oai/lib/secu/nas_stream_eia1.c
talkhasib: Same here, input is a stream of bytes that guarantees no alignment

lte/gateway/c/oai/common/common_defs.h
ulaskozat: It does that, but size is incremented by 3 bytes. So the last byte will be overwritten. For that reason I commented that we should make the copy for the 3 bytes as well to avoid any confusion and also avoid potential buffer overflow if this happens to be the last 3 bytes of space in the allocated buffer.

PR #5280, opened 2021-03-03
AGW: pipelined: introduce locking to avoid exception on TC queue removal.

pshelar:
pipelined executes datapath operations asynchronously using
multiple threads. This can results in duplicate request to
create/delete lower level datapath objects, this generally
works but at scale test exceptions can cause slow down. This
patch adds locking and current state check to avoid this issue.

This PR also handles case where APN AMBR is configured multiple
times in same session.

## Additional Information


koolzz (approved):
seems legit

PR #5291, opened 2021-03-03
DRAFT: AGW: fix pyroute2 dependency.

pshelar:
## Additional Information

### DIFF
--- a/lte/gateway/release/magma.lockfile
+++ b/lte/gateway/release/magma.lockfile
@@ -198,6 +198,12 @@
       "sysdep": "python3-pymemoize",
       "version": "1.0.2"
     },
+    "pyroute2": {
+      "root": true,
+      "source": "apt",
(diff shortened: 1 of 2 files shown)


lte/gateway/release/magma.lockfile
```diff
@@ -198,6 +198,12 @@
       "sysdep": "python3-pymemoize",
       "version": "1.0.2"
     },
+    "pyroute2": {
+      "root": true,
+      "source": "apt",
```
tmdzk, line 203: the source is pypi no?

tmdzk (changes requested):

lte/gateway/release/magma.lockfile
pshelar: This was auto generated. so I kept it.
`fab test package:git` is reverting it back to "apt"

PR #5294, opened 2021-03-04
[pipelined]Add rpc failure for uninitialized controllers

koolzz:

Respond with rpc failure when a controller is not yet initialized

## Additional Information


pshelar (approved):
Thanks!

PR #5306, opened 2021-03-04
[AGW][SessionD] Remove per-session counters on session termination

### LINKED ISSUE #5301: [AGW][Metrics][SessionD] Delete ue counter labels for terminating sessions
### Your Environment

- **Version:** master/v1.4
- **Affected Component:** Orchestrator, Access Gateway, or NMS
- **Affected Subcomponent:** Orchestrator pod/service or AGW service
- **Deployment Environment:** e.g. AWS (orc8r), local docker-compose (orc8r), bare-metal (AGW), Vagrant (AGW)

### Describe the Issue
The metrics stack is getting overloaded, as we create unique labels for each session that comes into the system but we never remove them when the session leaves. This leads to per-session metrics accumulating over time.

themarwhal:

## Summary
2 things:
1. Add a `RemoveCounter` method to the metric stack in orc8r/gateway/c/common.
2. On session termination, remove all counters that handle per-session information
   i. `ue_traffic`
  ii. `ue_dropped_usage`
  iii. `ue_reported_usage`

Removing the per-session metrics ensures, we don't gradually build up metrics of old sessions on the gateway. But for short-lived sessions (~few seconds), we might remove the metric too quickly for the metrics collector to report the usage.

I'm not sure if we should be using prometheus for data that we need to persist/rely on. So it would be good to migrate to a state based / event based approach.

## Test Plan
Run a s1ap test to attach and generate traffic.
During test:
```
(python) vagrant@magma-dev:~$ service303_cli.py metrics sessiond
name: "0"
help: ""
type: GAUGE
metric {
  gauge {
    value: 1614874631.0
  }
}

name: "3"
help: ""
type: GAUGE
metric {
  gauge {
    value: 44.085746504
  }
}

name: "1"
help: ""
type: GAUGE
metric {
  gauge {
    value: 21990594646016.0
  }
}

name: "2"
help: ""
type: GAUGE
metric {
  gauge {
    value: 59527168.0
  }
}

name: "ue_dropped_usage"
help: ""
type: COUNTER
metric {
  label {
    name: "IMSI"
    value: "IMSI001010036235177"
  }
  label {
    name: "apn"
    value: "magma.ipv4"
  }
  label {
    name: "direction"
    value: "down"
  }
  label {
    name: "msisdn"
    value: ""
  }
  counter {
    value: 0.0
  }
}
metric {
  label {
    name: "IMSI"
    value: "IMSI001010036235177"
  }
  label {
    name: "apn"
    value: "magma.ipv4"
  }
  label {
    name: "direction"
    value: "up"
  }
  label {
    name: "msisdn"
    value: ""
  }
  counter {
    value: 0.0
  }
}

name: "ue_reported_usage"
help: ""
type: COUNTER
metric {
  label {
    name: "IMSI"
    value: "IMSI001010036235177"
  }
  label {
    name: "apn"
    value: "magma.ipv4"
  }
  label {
    name: "direction"
    value: "down"
  }
  label {
    name: "msisdn"
    value: ""
  }
  counter {
    value: 2597.0
  }
}
metric {
  label {
    name: "IMSI"
    value: "IMSI001010036235177"
  }
  label {
    name: "apn"
    value: "magma.ipv4"
  }
  label {
    name: "direction"
    value: "up"
  }
  label {
    name: "msisdn"
    value: ""
  }
  counter {
    value: 20858.0
  }
}

name: "ue_traffic"
help: ""
type: COUNTER
metric {
  label {
    name: "IMSI"
    value: "IMSI001010036235177"
  }
  label {
    name: "direction"
    value: "down"
  }
  label {
    name: "session_id"
    value: "IMSI001010036235177-148766"
  }
  counter {
    value: 0.0
  }
}
metric {
  label {
    name: "IMSI"
    value: "IMSI001010036235177"
  }
  label {
    name: "direction"
    value: "up"
  }
  label {
    name: "session_id"
    value: "IMSI001010036235177-148766"
  }
  counter {
    value: 0.0
  }
}
```
After the test completes and terminates the session:
```
(python) vagrant@magma-dev:~$ service303_cli.py metrics sessiond
name: "0"
help: ""
type: GAUGE
metric {
  gauge {
    value: 1614874631.0
  }
}

name: "3"
help: ""
type: GAUGE
metric {
  gauge {
    value: 70.691450959
  }
}

name: "1"
help: ""
type: GAUGE
metric {
  gauge {
    value: 21990594711552.0
  }
}

name: "2"
help: ""
type: GAUGE
metric {
  gauge {
    value: 61468672.0
  }
}

name: "ue_dropped_usage"
help: ""
type: COUNTER

name: "ue_reported_usage"
help: ""
type: COUNTER

name: "ue_traffic"
help: ""
type: COUNTER
```
## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -1215,6 +1215,9 @@ void LocalEnforcer::complete_termination(
   reporter_->report_terminate_session(termination_req, logging_cb);
   events_reporter_->session_terminated(imsi, session);
 
+  // clear all metrics associated with this session
+  session->clear_session_metrics();
+
   // Delete the session from SessionMap
   session_uc.is_session_ended = true;
   session_map[imsi].erase(*session_it);
--- a/lte/gateway/c/session_manager/MeteringReporter.cpp
+++ b/lte/gateway/c/session_manager/MeteringReporter.cpp
@@ -49,27 +49,17 @@ void MeteringReporter::report_usage(
     total_rx += (double) credit_update.bucket_deltas[USED_RX];
   }
 
-  report_upload(imsi, session_id, total_tx);
-  report_download(imsi, session_id, total_rx);
+  report_traffic(imsi, session_id, DIRECTION_UP, total_tx);
+  report_traffic(imsi, session_id, DIRECTION_DOWN, total_rx);
 }
 
 void MeteringReporter::initialize_usage(
     const std::string& imsi, const std::string& session_id,
     SessionCredit::TotalCreditUsage usage) {
-  report_upload(imsi, session_id, usage.monitoring_tx + usage.charging_tx);
-  report_download(imsi, session_id, usage.monitoring_rx + usage.charging_rx);
-}
-
-void MeteringReporter::report_upload(
-    const std::string& imsi, const std::string& session_id,
-    double unreported_usage_bytes) {
-  report_traffic(imsi, session_id, DIRECTION_UP, unreported_usage_bytes);
-}
-
-void MeteringReporter::report_download(
-    const std::string& imsi, const std::string& session_id,
-    double unreported_usage_bytes) {
-  report_traffic(imsi, session_id, DIRECTION_DOWN, unreported_usage_bytes);
+  auto tx = usage.monitoring_tx + usage.charging_tx;
+  auto rx = usage.monitoring_rx + usage.charging_rx;
+  report_traffic(imsi, session_id, DIRECTION_UP, tx);
+  report_traffic(imsi, session_id, DIRECTION_DOWN, rx);
 }
 
 void MeteringReporter::report_traffic(
--- a/lte/gateway/c/session_manager/MeteringReporter.h
+++ b/lte/gateway/c/session_manager/MeteringReporter.h
@@ -42,20 +42,6 @@ class MeteringReporter {
 
  private:
   /**
-   * Report upload traffic usage for a session
-   */
-  void report_upload(
-      const std::string& imsi, const std::string& session_id,
-      double unreported_usage_bytes);
-
-  /**
-   * Report download traffic usage for a session
-   */
-  void report_download(
-      const std::string& imsi, const std::string& session_id,
-      double unreported_usage_bytes);
-
-  /**
    * Report traffic usage for a session
    */
   void report_traffic(
--- a/lte/gateway/c/session_manager/SessionState.cpp
+++ b/lte/gateway/c/session_manager/SessionState.cpp
@@ -30,15 +30,20 @@
 #include "DiameterCodes.h"
 
 namespace {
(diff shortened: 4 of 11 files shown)


lte/gateway/c/session_manager/MeteringReporter.cpp
```diff
@@ -49,27 +49,17 @@ void MeteringReporter::report_usage(
...
 }
 
 void MeteringReporter::initialize_usage(
     const std::string& imsi, const std::string& session_id,
     SessionCredit::TotalCreditUsage usage) {
-  report_upload(imsi, session_id, usage.monitoring_tx + usage.charging_tx);
-  report_download(imsi, session_id, usage.monitoring_rx + usage.charging_rx);
-}
-
-void MeteringReporter::report_upload(
```
themarwhal, line 63: these private methods don't add much value, so removing.

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1215,6 +1215,9 @@ void LocalEnforcer::complete_termination(
   reporter_->report_terminate_session(termination_req, logging_cb);
   events_reporter_->session_terminated(imsi, session);
 
+  // clear all metrics associated with this session
+  session->clear_session_metrics();
```
uri200, line 1219: should we do the same on session creation to cover the case when session is not terminated properly?

So when we do 
```
auto session_map     = session_store_.read_sessions({imsi});
```
If session_map is length 0, then trigger that function to make sure there is no left over for that imsi.

uri200 (approved):

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: What do you mean by `when session is not terminated properly` ?

mpgermano (approved):
Thanks for this. This should be a major improvement for metrics load

lte/gateway/c/session_manager/LocalEnforcer.cpp
uri200: Imagine we exit in line 1211. WE will never clear that metric. So it would be there forever (i guess reboot would clear it).

So maybe it would be good to make sure that when we start a session fresh, we do not have anything for that session in the metrics. It could be a simple way to reconcile both states for sure.

themarwhal: oooo I see. But if fail at line 1211, we don't delete the SessionState object. So there could be another opportunity for the state to be removed later on. 
I was thinking about cleaning up on Session creation as well. But since the metric specifies sessionID as one of its label, I can't easily remove it without changing our metric hashing implementation.  
So right now, the lifecycle of the metrics is tied to the SessionState object. If we have a trouble cleaning up the session in SessionD, the metric will never get cleared :/

PR #5307, opened 2021-03-04
[MME][ZMQ] Protect against multiple threads using the same context

talkhasib:

## Summary
Tasks interfacing to outside of MME create new threads when handling GRPC calls. Thus, zmq context is getting shared among multiple threads. This could create many race issues. The services in question are (SCTP, S6A, and SPGW) + Timers

This PR protects the zmq context with a mutex for exclusive access.

## Test Plan

fab integ_test


ulaskozat (approved):
LGTM

ulaskozat:
MAGMA-OAI-TESTS

PR #5319, opened 2021-03-05
[nms][tf] Add TF options to run Postgres with 2 logical DBs for orc8r/nms

andreilee:
## Summary

This pull request adds terraform options to choose whether the NMS DB is set up as MariaDB, Postgres, or both, to allow for data migration from MariaDB->Postgres.

To enable NMS to be migrated to use Postgres and share the same AWS RDS instance with orc8r, terraform changes are required. Two new terraform variables are added:
-`nms_using_mariadb`
-`nms_using_postgres`

Before migration, NMS will be using MariaDB, during migration, NMS will use both MariaDB/Postgres, and after migration, NMS will use Postgres only. NMS will share the same AWS RDS instance for Postgres as orc8r, but a separate logical DB will be setup for NMS, to prevent namespace issues.

To migrate, terraform should be configured to use both MariaDB and Postgres, and NMS should detect this automatically, and migrate data from MariaDB to Postgres (coming in a later PR). Terraform should be configured to use Postgres only, and `terraform apply` should be run again.

## Test Plan

To be tested in a test 1.4 deployment after landing.

## Additional Information

### DIFF
--- a/docs/docusaurus/versioned_docs/version-1.0.0/orc8r/deploy_terraform.md
+++ b/docs/docusaurus/versioned_docs/version-1.0.0/orc8r/deploy_terraform.md
@@ -44,6 +44,8 @@ keypair that you imported or created in the above step:
 $ cat vars.tfvars
 db_password = "foobar"
 nms_db_password = "foobar"
+nms_using_mariadb = true
+nms_using_postgres = false
 key_name = "my_key"
 ```
 
--- a/docs/readmes/orc8r/deploy_install.md
+++ b/docs/readmes/orc8r/deploy_install.md
@@ -126,6 +126,8 @@ new `main.tf` file. Follow the example Terraform root module at
 override the following parameters
 
 - `nms_db_password` must be at least 8 characters
+- `nms_using_mariadb` true/false, for migration off mariadb
+- `nms_using_postgres` true/false, for migration to postgres with orc8r
 - `orc8r_db_password` must be at least 8 characters
 - `orc8r_domain_name` your registered domain name
 - `docker_registry` registry containing desired Orchestrator containers
--- a/orc8r/cloud/deploy/terraform/orc8r-aws/db.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-aws/db.tf
@@ -32,6 +32,33 @@ resource "aws_db_instance" "default" {
   final_snapshot_identifier = "foo"
 }
 
+# Configure the MySQL provider based on the outcome of
+# creating the aws_db_instance.
+provider "postgresql" {
+  endpoint = "${aws_db_instance.default.endpoint}"
+
+  username = "${aws_db_instance.default.username}"
+  password = "${aws_db_instance.default.password}"
+
+  vpc_security_group_ids = "${aws_db_instance.default.vpc_security_group_ids}"
+
+  db_subnet_group_name = "${aws_db_instance.default.db_subnet_group_name}"
+
+  skip_final_snapshot = "${aws_db_instance.default.skip_final_snapshot}"
+}
+
+resource "postgresql_database" "nms" {
+  allocated_storage = var.nms_db_storage_gb
+
+  name     = var.nms_db_name
+
+  # we only need this as a placeholder value for `terraform destroy` to work,
+  # this won't actually create a final snapshot on destroy
+  final_snapshot_identifier = "nms-bar"
+
+  count = "${var.nms_using_postgres == true ? 1 : 0}"
+}
+
 resource "aws_db_instance" "nms" {
   identifier        = var.nms_db_identifier
   allocated_storage = var.nms_db_storage_gb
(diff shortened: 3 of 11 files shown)


orc8r/cloud/deploy/terraform/orc8r-aws/db.tf
```diff
@@ -32,6 +32,33 @@ resource "aws_db_instance" "default" {
...
+resource "postgresql_database" "nms" {
+  allocated_storage = var.nms_db_storage_gb
+
+  name     = var.nms_db_name
+
+  # we only need this as a placeholder value for `terraform destroy` to work,
+  # this won't actually create a final snapshot on destroy
+  final_snapshot_identifier = "nms-bar"
+
+  count = "${var.nms_using_postgres == true ? 1 : 0}"
```
karthiksubraveti, line 59: Will this deploy another db instance for NMS ?

docs/docusaurus/versioned_docs/version-1.0.0/orc8r/deploy_terraform.md
```diff
@@ -44,6 +44,8 @@ keypair that you imported or created in the above step:
 $ cat vars.tfvars
 db_password = "foobar"
 nms_db_password = "foobar"
+nms_using_mariadb = true
```
karthiksubraveti, line 47: Is this temporary ? Why do we need to do this for NMS alone ?

orc8r/cloud/deploy/terraform/orc8r-aws/db.tf
andreilee: This should create a second logical DB on the same AWS RDS instance.

docs/docusaurus/versioned_docs/version-1.0.0/orc8r/deploy_terraform.md
andreilee: This is temporary, and should be removed in 1.6. Only NMS is using MariaDB, and the terraform files are set up to name the MariaDB instance as "nms", so I think the name specifying NMS is natural here.

karthiksubraveti: If we are unifying this we might as well get rid of these differences ? Is the idea that in case we break something, this might give us a work around ?

andreilee: Both need to be deployed at some point to migrate the data through sequelize

karthiksubraveti: I don't feel that we have to pass this additional complexity during upgrade time and having to require user to later turn off the NMS DB. I feel it might be better to have a migration script prior to upgrade, enable us from moving NMS db tables to postgres.

orc8r/cloud/deploy/terraform/orc8r-aws/db.tf
karthiksubraveti: Just curious on why Orc8r doesn't require this ?

docs/docusaurus/versioned_docs/version-1.0.0/orc8r/deploy_terraform.md
hcgatewood, line 47: Correct me if I'm wrong, but I don't see why this needs to be backported to v1.0 docs

hcgatewood: Since we can assume the Postgres db already exists, I think it's safe to invert the normal order (tf then data migrations => data migrations then tf) for v1.5, for the benefit of just automatically removing the old NMS db in a single-pass tf apply

docs/readmes/orc8r/deploy_install.md
```diff
@@ -126,6 +126,8 @@ new `main.tf` file. Follow the example Terraform root module at
 override the following parameters
 
 - `nms_db_password` must be at least 8 characters
+- `nms_using_mariadb` true/false, for migration off mariadb
+- `nms_using_postgres` true/false, for migration to postgres with orc8r
```
hcgatewood, line 130: Re my other comments: I think we can get away with *not* adding these vars

orc8r/cloud/deploy/terraform/orc8r-helm-aws/examples/remote/main.tf
```diff
@@ -26,6 +26,13 @@ terraform {
     dynamodb_table = "my-dynamodb-table"
     region         = "us-west-2"
   }
+
+  required_providers {
+    postgresql = {
+      source = "cyrilgdn/postgresql"
```
hcgatewood, line 32: It wasn't immediately clear to me that this is "the" postgres provider. Can we add a comment that this is in fact the de facto postgres provider

hcgatewood (changes requested):
Excited for this functionality

I'm unconvinced that we need to have `nms_using_mariadb` and `nms_using_postgres` options, see my inline comment for why

P.S. I'm assuming this is handled but: we *don't* require Postgres for Orc8r deployments -- we also support maria. Can you add a note somewhere describing how this is being handled

andreilee:
@hcgatewood Abandoning this PR. Just going to use the same DB as orc8r. Still, thanks for the thoughtful review! (#5858 ) (https://github.com/magma/magma/tree/master/nms/app/packages/magmalte/scripts/fuji-upgrade)

PR #5320, opened 2021-03-05
[orc8r] Add LTE federated parameters to LTE mconfig and Federated Mapping definition

### LINKED ISSUE #4918: [Inbound Roaming][orc8r] Stream down roaming/switching related configuration to AGWs as part of Mconfig
Depends on: https://github.com/magma/magma/issues/4915, https://github.com/magma/magma/issues/4914

Once the above are resolve, we can stream down whatever configuration we end up with as part of the MME mconfig.

uri200:

## Summary

We have two types of lte networks: non federated and federated. mconfig builder is using LTE non federated serdes. So any parameter that is not common to both `lte non federated` and `lte federated` is not visible to the builder.

With this PR, the builder will try to parse that config file as a `lte federated` network and extract some of the fields in case of being a federated LTE network.

This PR also includes the definition in swagger of Federated Mode Mapping. That mapping  will help AGW to decide where to send the request based on IMSI range, PLMN and APN

The modes available will be 
- Federated OFF: where subscriberDb spgw_task and Policy store are used
- Federated local: where HSS (through feg)  and spgw_task (mme) are used
- Federated remote: where HSS (through feg) and s8_task are used)

Note MME tasks will implement the routing depending on those parameters. 

## Test Plan

./build -g
./build --test --up

## Additional Information


feg/cloud/go/go.mod
```diff
@@ -12,6 +12,7 @@
 module magma/feg/cloud/go
 
 replace (
+	magma/feg/cloud/go => ../../../feg/cloud/go
```
uri200, line 15: this replace is needed or some modules will not be found

feg/cloud/go/services/feg/obsidian/models/swagger.v1.yml
```diff
@@ -1562,3 +1564,53 @@ definitions:
       description:
         type: string
         x-nullable: false
+
+  federated_mode_map:
```
emakeev, line 1568: Can you please add some comments explaining the big picture here?

feg/cloud/go/services/feg/obsidian/models/swagger.v1.yml
```diff
@@ -1562,3 +1564,53 @@ definitions:
...
+
+  mode_map_item:
+    description: Item containing {mode, [plmnA, plmB], [imsi1, imsi2], [apnY, apnZ]}
+    type: object
+    properties:
+      mode:
+        type: string
+        enum:
+        - federated_off
+        - federated_local
```
emakeev: what are those? local/off/remote
and how federated can be off for federated network?

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
```diff
@@ -9200,6 +9216,40 @@ definitions:
...
+      plmn_ids:
+        items:
+          example: "123456"
+          maxLength: 6
+          minLength: 5
+          pattern: ^(\d{5,6})$
+          type: string
+          x-nullable: false
+        type: array
+    type: object
```
emakeev: Maybe makes sense to simplify it a little:
1) avoid arrays of keys (imsi_range, apn, plmn_id)
2) rename federated_off & federated_remote
3) get rid of federated_local, it's a default/legacy case anyway

emakeev (approved):
LGTM

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -18,6 +18,9 @@ import (
 	"fmt"
 	"sort"
 
+	feg "magma/feg/cloud/go/feg"
```
mpgermano, line 21: Importing from the feg to lte module is something we have purposely avoided as it creates a cyclic dependency between modules. Previously only feg could import from lte.

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -78,6 +81,12 @@ func (s *builderServicer) Build(ctx context.Context, request *builder_protos.Bui
 		return nil, err
 	}
 
+	FederatedNetworkConfigs, err := getFederatedNetworkConfigs(cellularNwConfig.FegNetworkID, request)
```
mpgermano: Nit: `FederatedNetworkConfigs -> federatedNetworkConfigs`

mpgermano (commented):
I think the main issue I have with this implementation is that it violates the dependency structure we have for orc8r modules. Previously, LTE never imported from the Feg module. I understand though, that there's not a great solution here otherwise.

Given that mconfiguration is a composite of outputs from different modules, a way to avoid the dependency change would be instead to add logic to the feg mconfig builder to handle specific configuration for `feg_lte` network types. In this way, lte returns as normal while the feg module is responsible for anything unique to a federated lte network. The issue here is that we don't support the case of two different mconfig Build requests (i.e. one for lte and one for feg) returning the same mconfig key (i.e. `mme`).

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -515,3 +525,30 @@ func getServiceAreaMaps(serviceAreaMaps map[string]lte_models.TacList) map[strin
 	}
 	return ret
 }
+
+// getFederatedNetworkConfigs in case this is a federated LTE networkm this function will try to parse out
+// feg_models.FederatedNetworkConfigs out of it
+func getFederatedNetworkConfigs(fegId lte_models.FegNetworkID, request *builder_protos.BuildRequest) (*feg_models.FederatedNetworkConfigs, error) {
+	if fegId == "" {
```
mpgermano, line 532: I think it would also be good as a sanity check to check the request.Network.Type to ensure it is feg_lte.

mpgermano (approved):
Followed up with @hcgatewood offline and we've decided that despite their cyclic dependency potential, there's not a better short term solution for now. We'll take the action to come up with a more clear plan here. For now to unblock this hi-pri work we can go ahead with this. Please address the few comments before landing.

uri200:
Hey @mpgermano  thanks for the comments. Let me add some context for other people reading this.

As you know, we have two different swagger definitions for lte network: `lte_network` and `feg_lte_network`. To me, those two types make sense because, even their similarities, those are different concepts that evolve independently. 

But, even that we have two types for those networks, mconfig builder only uses one the type `lte_network`. 

So far this has worked because `feg_lte_network` happened to have exactly the same keys as `lte_network`, so networks from both types were parseable using `lte_netowrk` models. Now that we need to define different atributes for `feg_lte_network`,  those atributes are not visible to the `lte_network` model anymore. 

Note those differences between `lte_network` and `feg_lte_network` will become wider now that we are going to have more federated use cases. So we can't keep pushing them on the feg models. Besides, those parameters don't belong to the feg, but to the AGW network (and not only the network but even to each  particular AGW gateway)

As a solutions I debated myself to do:

- Two separate mconfig builders, one for `lte_networks` and another for `feg_lte_networks`. But that would not solve the issue with the modules. Plus it would imply maintaining yet another mconfig builder.

- Another solution that crossed my mind was to add that config definition under `cellular`. Since cellular is in both swagger models it would appear in both and we could keep parsing with just lte_network model. Cons with that were
	- That config would make no sense for non federated network. 
	- Configuration I am adding it is not a `cellular` parameter.
	- We are just patching the need to parse two separate configurations. Next time we need to add a field just on `feg_lte_network`, we will pollute even more `lte_network` models. So, why to have 2 different models after all?

So to me the less intrusive and extendable solution is what you see on this PR: to inject the extra parameters and take the toll of adding those dependencies, that in the end sooner or later we will end up having. In fact CWF and FEG modules already have some LTE references

uri200:
Thanks @mpgermano @hcgatewood  for quick turnaround

PR #5329, opened 2021-03-05
[MME] Correct mis-sized allocation of grpc_service_data_t

electronjoe:
## Summary

Incorrect use of sizeof() reference here was probably guarded by the fact that grpc_service_data_t is strictly larger than the intended grpc_service_config struct (performance implications, not safety implications).

## Test Plan

Validated that the associated clang-tidy findings were resolved.

Additionally:

```
cd lte/gateway
make build_oai
```


rdefosse:
MAGMA-OAI-TESTS

talkhasib (approved):
LGTM

ssanadhya (approved):

PR #5330, opened 2021-03-05
[MME] Except some nas_precedures.c Clang-Tidy linting

electronjoe:
## Summary

These Clang-Tidy linting findings are false positives. The MME uses casting to provide struct inheritance over blocks of allocated memory.  Here we annotate these findings to except them from Linting.

Signed-off-by: Scott Harrison Moeller <smoeller@fb.com>

PR #5332, opened 2021-03-05
[MME] Except some nas_state_converter Clang-Tidy linting

electronjoe:
Summary

Due to the specific use of c-style inheritance and cross-struct-type
casting within the NAS - Clang-Tidy throws
clang-analyzer-unix.MallocSizeof warnings.  As these are intended we
annotate them for linter removal here.

Testing Plan

Confirmed Clang-Tidy findings were removed.

### DIFF
--- a/lte/gateway/c/oai/tasks/nas/nas_state_converter.cpp
+++ b/lte/gateway/c/oai/tasks/nas/nas_state_converter.cpp
@@ -1075,7 +1075,7 @@ void NasStateConverter::proto_to_emm_specific_proc(
     case oai::NasEmmProcWithType::kAttachProc: {
       OAILOG_DEBUG(LOG_MME_APP, "Reading attach proc from proto");
       state_emm_procedures->emm_specific_proc =
-          (nas_emm_specific_proc_t*) calloc(1, sizeof(nas_emm_attach_proc_t));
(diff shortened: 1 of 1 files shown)


electronjoe:
MAGMA-OAI-TESTS

ssanadhya (changes requested):
Same comment as #5330

electronjoe:
Thanks @ssanadhya!

Working as intended - this is how we are doing c inheritance. I will work at add Clang-Tidy annotations so this is ignored moving forward.

Niggling concern that these types may not technically be 100% `compatible types` due to differing tags - so just to drop that breadcrumb (don't think this is biting us):

```
6.2.7 Compatible type and composite type
1 Two types have compatible type if their types are the same. Additional rules for
determining whether two types are compatible are described in 6.7.2 for type specifiers,
in 6.7.3 for type qualifiers, and in 6.7.6 for declarators.55) Moreover, two structure,
union, or enumerated types declared in separate translation units are compatible if their
tags and members satisfy the following requirements: If one is declared with a tag, the
other shall be declared with the same tag...
```

Of C11 draft spec (all I can get for free online).

electronjoe:
This PR seemed like a reasonable place to move to a NOLINT annotation.

Please take a look @ssanadhya! This will prevent this warning in the future.

rdefosse:
MAGMA-OAI-TESTS

ssanadhya (approved):

PR #5337, opened 2021-03-05
Fix potential null pointer dereference in log.c

electronjoe:
## Summary

It's unclear to me whether is_async can change over the course of
execution in e.g. log_message_int_prefix_id or log_message_int.

But if this is possible (and static analysis wouldn't know if it
wasn't), then we can pass through the top check of g_oai_log.is_async
and assume it is false -> not initialize async behavior. Then hit an
error in e.g. the bvcformata call and goto error_event -> where it is
conceivable that is_async has toggled to true -> and we now de-reference
it's NULL-initialized value.

Here we add a guard against this sequence of events, however unlikely
given the use pattern of logging (I would note that the g_oai_log
structure is statically initialized, I'm not certain there are
guarantees on the ordering of its initialization and first function
calls?).

## Test Plan

```
cd lte/gateway
make build_oai
make test_oai
```

In addition to CI pipelines.

## Additional Information

### DIFF
--- a/lte/gateway/c/oai/common/log.c
+++ b/lte/gateway/c/oai/common/log.c
@@ -1403,6 +1403,10 @@ void log_message_int(
 error_event:
   if (!(g_oai_log.is_async)) {
     LOG_FREE_ITEM(sync_context_p);
+  } else if (async_context_p == NULL) {
+    // To guard against mutation of is_async during post-init
+    // runtime, though this should not be mutated.
+    return;
   } else {
     LOG_FREE_ITEM_ASYNC(*async_context_p);
   }
@@ -1477,7 +1481,9 @@ void log_message_int_prefix_id(
 error_event:
   if (!(g_oai_log.is_async)) {
     LOG_FREE_ITEM(sync_context_p);
-  } else {
+  } else if (async_context_p == NULL) {
+      return;
+  } else {  
     LOG_FREE_ITEM_ASYNC(*async_context_p);
   }
 }


electronjoe:
MAGMA-OAI-TESTS

lte/gateway/c/oai/common/log.c
```diff
@@ -1404,6 +1404,9 @@ void log_message_int(
   if (!(g_oai_log.is_async)) {
     _LOG_FREE_ITEM(sync_context_p);
   } else {
+    if (async_context_p == NULL) {
+      return;
+    }
```
talkhasib: We wont reach this point if async_context_p was null
Move to right after line 1381

lte/gateway/c/oai/common/log.c
```diff
@@ -1478,6 +1481,9 @@ void log_message_int_prefix_id(
...
```
talkhasib: same here, move to after line  1458

talkhasib (changes requested):

lte/gateway/c/oai/common/log.c
electronjoe: This is the claim of Clang-Tidy and the logic seems to hold - unless we make assumptions on immutability of g_oai_log.is_async (which Clang-Tidy is not allowed). 

- Line 1361
  - assume !g_oai_log.is_async
    - So we initialize sync_context_p etc
    - But then lets assume that one of the errors triggers `goto error_event` line 1403
- Line 1403
  - In the time between Line 1361 and 1403, g_oai_log.is_async toggles true
  - We dereference a null async_context_p

lte/gateway/c/oai/common/log.c
electronjoe: Same null-pointer access path as above comment, please take a look

lte/gateway/c/oai/common/log.c
ulaskozat: This is set during task initialization based on the THREAD_SAFE configuration in mme.conf.template. We can assume it is immutable during run time.

ulaskozat:
Can we resolve this and close or merge? @talkhasib @electronjoe

lte/gateway/c/oai/common/log.c
electronjoe: Our option is to either annotate this as expected behavior to remove the warning - or fix the issue from any future regression - seems to me we might as well put in the future-proof if statement instead of an annotation.

electronjoe:
@ulaskozat @talkhasib I would just commit the if conditions as that seems strictly better IMO than committing an annotation to silence the clang tidy warning (which is legitimate, even if presently the behavior is not possible due to how we currently employ logging changes).

ulaskozat (approved):
LGTM

lte/gateway/c/oai/common/log.c
talkhasib: Seeing "goto" gives me chills
Fine with moving ahead with this
Maybe change else to else if and add the check there

talkhasib (approved):

lte/gateway/c/oai/common/log.c
electronjoe: Yeah good point on else if -> I should not have nested this condition yet deeper in the conditional tree. I'll fixup that now.

rdefosse:
@electronjoe (scott)

It's weird. It says that the PR has merge conflicts but the OAI pipeline went though on the verify code stage. And try to build.

It should have stopped earlier.

Also the link to the Jenkins job is not present?

Did you forced-push quickly?
Raphael

electronjoe:
> has merge conflicts but the OAI pipeline went though on the verify code stage. And try to build.

@rdefosse  There was a merge conflict, and about an hour or so later I fixed the merge conflict and force pushed the fixup.

Maybe that timing broke something?

rdefosse:
@electronjoe 

Now it passes! Top

Issue #5222, opened 2021-03-01
[AGW] Our Cmake files need standardization / cleanup / linting


electronjoe:
## Problem

Many times in the last three weeks I've run into sometimes nuanced or nasty CMake errors in our existing Magma repo.

I recently spent an hour trying to uncomment two disabled MME unit tests - with compiler include errors - only to find a colleague had also attempted and aborted the same week.

I am also presently chasing an issue where our project builds fine in my local Dockerfile, but on CI fails with an error at linking time (unable to find the target library, defined in another CMakeLists.txt!).  Root cause isn't yet known but for example best practice seems to be providing full paths on target_link_libraries stanzas, but our Magma configs all use aliases.

Generally we appear to rely too frequently on looseness in the CMake definitions to compile by luck (or by a loophole of a particular Cmake version.. Ninja timing, etc).

This GH Issue should Collect resources / readings on best practices, and a list of cleanup actions to make the CMake situation better.  I'll pin them to the top-of-line Issue here (updating in place).

## Resources

- [Effective Modern CMake](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1)
- [Modern CMake is like inheritance](https://kubasejdak.com/modern-cmake-is-like-inheritance)

## Oustanding Cleanup Actions (to be forked into GH Issues if sizable)

- [Define Project Properties Globally](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#define-project-properties-globally)
  - E.g. move all flags, c++ std selection, etc - to global / top level
- Enable use of `find_*` for all libraries (including within our CMake tree)
  - See [Using a library defined in the same CMake tree should look the same as using an external library](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#using-a-library-defined-in-the-same-cmake-tree-should-look-the-same-as-using-an-external-library)

electronjoe:
I could collect a fairly large list of instances where our CMake use broadly is causing problems, I'm not sure it's helpful.

But as an example at this moment:

[...]

PR #5358, opened 2021-03-08
[AGW] Remove -Werror compiler flag everywhere, add -Wextra

electronjoe:
# Summary

Per  [This](https://github.com/magma/magma/discussions/5273) GH Discussion, here we drop the Werror compiler flag.  In its place, we offer #5404 which is capable of Github Pull Request GCC Warning annotations for code reviewers.

With `-Werror` gone, we also turn up the warnings by adding `-Wextra` and remove some suppressed warnings.

There are some places where `-Wextra` is noisy - e.g. functions that have unused parameters because they are callback driven and have to contain a specified shape.  We should be able to mitigate these warnings by appropriate modification of the unused parameter declaration. E.g. in C++ as advocated by [Herb Sutter](https://herbsutter.com/2009/10/18/mailbag-shutting-up-compiler-warnings/) and in C by omitting the name or void casting.

```
void foo(int /*bar_unused*/) {
    ...
}
```

or

```
void foo(int bar_unused) {
    // bar_unused is not needed for this function, but must exist to meet interface expectations
    (void)bar_unused;
    ...
}
```

### DIFF
--- a/lte/gateway/c/session_manager/CMakeLists.txt
+++ b/lte/gateway/c/session_manager/CMakeLists.txt
@@ -42,7 +42,7 @@ add_definitions(-DLOG_WITH_GLOG)
 
 message("Build type is ${CMAKE_BUILD_TYPE}")
 if (CMAKE_BUILD_TYPE STREQUAL "Debug")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wno-unused-function")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
 endif ()
 
 include_directories("${OUTPUT_DIR}")
--- a/lte/gateway/c/connection_tracker/CMakeLists.txt
+++ b/lte/gateway/c/connection_tracker/CMakeLists.txt
@@ -38,7 +38,7 @@ add_definitions(-DLOG_WITH_GLOG)
 
 message("Build type is ${CMAKE_BUILD_TYPE}")
 if (CMAKE_BUILD_TYPE STREQUAL "Debug")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wno-unused-function")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
 endif ()
 
 include_directories("${OUTPUT_DIR}")
--- a/lte/gateway/c/oai/CMakeLists.txt
+++ b/lte/gateway/c/oai/CMakeLists.txt
@@ -26,7 +26,7 @@ endif ()
 
 # TODO (amar) check all the flags used. Copied as is from OAI.
 set(CMAKE_C_FLAGS
-  "${CMAKE_C_FLAGS} ${C_FLAGS_PROCESSOR} -std=gnu99 -Wall -Werror -Wstrict-prototypes -fno-strict-aliasing -rdynamic -funroll-loops -Wno-packed-bitfield-compat -fPIC"
+  "${CMAKE_C_FLAGS} ${C_FLAGS_PROCESSOR} -std=gnu99 -Wall -Wextra -Wstrict-prototypes -fno-strict-aliasing -rdynamic -funroll-loops -fPIC"
 )
 
 # add autoTOOLS definitions that were maybe used


lte/gateway/c/session_manager/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ add_definitions(-DLOG_WITH_GLOG)
 
 message("Build type is ${CMAKE_BUILD_TYPE}")
 if (CMAKE_BUILD_TYPE STREQUAL "Debug")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wno-unused-function")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-unused-function")
```
uri200: Hey @electronjoe, just curious, but per your comment on one of the linked PRs, it was mentioned you observed issues with mconfig while building `test_session_manager` on your docker container. 

Are you observing the same on our agw Vagrant vm when you do:
```
vagrant@magma-dev:~/magma/lte/gateway$ make test_session_manager
```

If not, maybe the issue is different gcc versions?

rdefosse:
MAGMA-OAI-TESTS

lte/gateway/c/session_manager/CMakeLists.txt
electronjoe: Hi @uri200! Thanks for commenting.

> Hey @electronjoe, just curious, but per your comment on one of the linked PRs, it was mentioned you observed issues with mconfig while building `test_session_manager` on your docker container.
> 
> Are you observing the same on our agw Vagrant vm when you do:
> 
> ```
> vagrant@magma-dev:~/magma/lte/gateway$ make test_session_manager
> ```

Yes, In attempting to migrate to Ubuntu 20.04 I'm currently blocked by -Wdeprecated warnings on **protoc compiled protobuf outputs** that are converted to `-Werror`.  This does not occur in the Ubuntu 18.04 based vagrant VM - but note that we are migrating actively this quarter to Ubuntu 20.04. There is a Ubuntu 20.04 Vagrant floating around, I have not tracked it down yet to determine if and how this was addressed.

> If not, maybe the issue is different gcc versions?

This is gcc 9.3 (the baseline for Ubuntu 20.04) - but the issue is with the -Werror flag being applied to protobuf compiles (which is not the case in our older Ubunto 18.04 Vagrant VMs).  Ultimately this points to a problem with our CMake configurations  [See technical debt enumerated in #5222] and / or a problem with Cmake (less likely) - triggered by moving forward with the CMake bundled with Ubuntu 20.04.  Looking at how we compile these protos - it's not surprising to me that this is a tangle (compiler flags leaking / inconsistently applied across build targets and build environments).

Another option is to remove -Wdeprecated (which is what blocks the above - but would need additional fixups to allow Clang to build without `-Werror`).  Having `-Wdeprecated` and `-Werror` is super inconsistent.  What is the point of annotating deprecation if you block compile time until it's removed entirely from the codebase.

edit: I'm realizing - and you were probably indicating above - that GCC 9.3 may have added Deprecation warnings to `-Wall`, and that could be the origin of this particular issue.  We could add `-Wno-deprecated-declarations` to ignore deprecation warnings.  Though this seems to negate the value of having deprecation annotations.  And this doesn't address the Clang build issue (with 100's of new warnings that block builds).  Ultimately I just don't feel `-Werror` is healthy - and per some of my references in GH Discussion it seems that is not a unique view.

themarwhal (approved):

ssanadhya (approved):

lte/gateway/c/session_manager/CMakeLists.txt
uri200: @electronjoe thanks for the detailed description.
I agree with the -werror [article](https://embeddedartistry.com/blog/2017/05/22/werror-is-not-your-friend/#:~:text=In%20case%20you%20have%20been,behind%20enabling%20%2DWerror%20are%20benevolent.&text=%2DWerror%20creates%20a%20project%20dependency%20on%20a%20specific%20compiler%20version.) you mentioned.
My concerned was just for session_manager: if it worked so far and it was not giving us errors, it would be nice to make it work in Ubuntu focal. But maybe it doesn't make sense anymore

Will take a look at Focal VM and see if, at least for session_manager, we can keep it.

BTW, proto generation is kind of messy. Sometimes it is even preventing the IDE to use the autocomplete properly with the protos, which makes the developer experience very poor. So any change on that would be awesome!

electronjoe: @uri200 you prompted me to get GitHub GCC Warning annotations working for Pull Request code reviews.  I think this clears the way quite well for this PR to land, removing `-Werror`?

See PR #5404

electronjoe:
MAGMA-OAI-TESTS

koolzz (approved):

lte/gateway/c/session_manager/CMakeLists.txt
uri200: Sound like a good solution to unblock the build and still keep the warnings visibility👍

rdefosse (commented):
OK for me

PR #5368, opened 2021-03-08
[PipelineD][SessionD] Cleanup flows for non-existent sessions

themarwhal:

## Summary
On some failure cases, we can have flows leftover in PIpelineD that are not cleaned up. This PR adds an attempt to remove all flows that are no longer attached to an existing session.
This was already implemented for v1.4: https://github.com/magma/magma/pull/5055. Adding the change to master as well to be consistent. 

1. Add a `WILDCARD` type to `RequestOriginType` for when we do not want to specify Gx/Gy. I also added a corresponding change in PipelineD to handle deactivate flows + `WILDCARD` case. 
2. Added tunnel IDs to the RuleRecord protobuf. This PR does not add the code to add the tunnel IDs from PipelineD's end. 

## Test Plan
make precommit_sm

Running a modified s1ap test to reproduce the stale flow case. (debug log level)
```
Mar 09 01:12:57 magma-dev sessiond[7874]: I0309 01:12:57.544251  8124 LocalSessionManagerHandler.cpp:50] Aggregating 4 records
Mar 09 01:12:57 magma-dev sessiond[7874]: I0309 01:12:57.544564  7874 LocalEnforcer.cpp:273] Deactivating 2 flows that are attached to non-existent sessions
Mar 09 01:12:57 magma-dev sessiond[7874]: I0309 01:12:57.544582  7874 PipelinedClient.cpp:263] Deactivating 0 static rules and 0 dynamic rules and default drop flows for IMSI001010000000001, type=WILDCARD, ipv4=192.168.128.12, ipv6=
Mar 09 01:12:57 magma-dev sessiond[7874]: I0309 01:12:57.544694  7874 PipelinedClient.cpp:263] Deactivating 0 static rules and 0 dynamic rules and default drop flows for IMSI001010000000002, type=WILDCARD, ipv4=192.168.128.13, ipv6=
```
## Additional Information


uri200:
Hey @themarwhal I think TEIDs are not actively used by pipelined neither on activate or deactivate. Sessiond passes them because the idea is pipelined will use them in the near future to identify the flows (ip will not be used anymore)

Not sure if this applies to the PR, but if this PR depends on IP being sent, it should depend on teids too

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -234,14 +234,16 @@ void LocalEnforcer::aggregate_records(
   // Insert the IMSI+SessionID for sessions we received a rule record into a set
   // for easy access
   std::unordered_set<ImsiAndSessionID> sessions_with_reporting_flows;
+  // In some failure cases, PipelineD may still hold onto flows for sessions
+  // that do not exist in SessionD. In this case, send DeactivateFlowsRequest
+  ImsiAndIPSet dead_sessions_to_cleanup;
```
uri200: I think this should include teids or will not work in the future

All the separation of create session request into two was build so we could use teids instead of ip (soon ???, ever???)

themarwhal:
@uri200 @koolzz 
How does PipelineD identify flows to deactivate? Is IMSI+IPv4+IPv6 no longer sufficient nowadays? 
Since we remove all context from SessionD on termination, the only information SessionD has to clean up stale flows is the values given in the RuleRecord. (https://sourcegraph.com/github.com/magma/magma/-/blob/lte/protos/session_manager.proto#L24)
```
  string sid = 1;
  string rule_id = 2;
  string ue_ipv4 = 5;
  string ue_ipv6 = 6;
```

So if we need more information, then we need to either:
1. Add the necessary information to the RuleRecord
2. Maintain session context until all flows are removed -> hard to do atm without versioning

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -260,6 +262,25 @@ void LocalEnforcer::aggregate_records(
...
+              << " flows that are attached to non-existent sessions";
+  Teids empty_teids;
+  for (const ImsiAndIPs& info : dead_sessions_to_cleanup) {
+    pipelined_client_->deactivate_flows_for_rules_for_termination(
+        info.imsi, info.ipv4, info.ipv6, empty_teids, {}, {},
+        RequestOriginType::GX);
+    pipelined_client_->deactivate_flows_for_rules_for_termination(
+        info.imsi, info.ipv4, info.ipv6, empty_teids, {}, {},
+        RequestOriginType::GY);
+  }
```
uri200, line 281: is it possible that just sending one request with an empty `RequestOriginType` will remove both?

uri200 (changes requested):

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: @koolzz ? We could define a new origin type : ANY or something like that.

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: Yeah once I can get clarification on whether we can include Teids in the RuleRecord or not, I can make the change.

lte/gateway/c/session_manager/LocalEnforcer.cpp
koolzz: sure lets define ANY

lte/gateway/c/session_manager/LocalEnforcer.cpp
koolzz: yeah down the line we're switching from ip->teid. if its easy to include might as well just add teid now/or we can just patch it later

themarwhal: I can add a new field Teid in the rule record protobuf, but I'll leave it empty from PipelineDs end for now.

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: @koolzz Added a WILDCARD enum to specify both. Added the corresponding change on the PipelineD receiving side.

lte/gateway/c/session_manager/LocalEnforcer.h
```diff
@@ -39,6 +39,25 @@ using std::experimental::optional;
 
 typedef std::pair<std::string, std::string> ImsiAndSessionID;
 
+struct RuleRecord_equal {
+  bool operator()(const RuleRecord& l, const RuleRecord& r) const {
+    // TODO replace IP comparison to tunnelID comparison
```
themarwhal, line 44: Addressing these TODOs when we migrate should do the trick @uri200.

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -260,6 +262,23 @@ void LocalEnforcer::aggregate_records(
...
+}
+
+void LocalEnforcer::cleanup_dead_sessions(
+    const RuleRecordSet dead_sessions_to_cleanup) {
+  if (dead_sessions_to_cleanup.size() == 0) {
+    return;
+  }
+  MLOG(MINFO) << "Deactivating" << dead_sessions_to_cleanup.size()
+              << " flows that are attached to non-existent sessions";
+  Teids empty_teids;
```
uri200: Shouldn't we use the teids coming in the request instead?  Pipelined is going to ignore them for now. But at least we will have that done for the future

themarwhal: oh yeah good call :) one less thing to worry about.

uri200 (approved):

themarwhal:
MAGMA-OAI-TESTS

lte/gateway/c/session_manager/LocalEnforcer.h
```diff
@@ -39,6 +39,25 @@ using std::experimental::optional;
...
+    return l.sid() == r.sid() && l.ue_ipv4() == r.ue_ipv4() &&
+           l.ue_ipv6() == r.ue_ipv6();
+  }
+};
+struct RuleRecord_hash {
+  std::size_t operator()(const RuleRecord& el) const {
+    size_t h1 = std::hash<std::string>()(el.sid());
+    // TODO replace IP hash to tunnelID hash
+    size_t h2 = std::hash<std::string>()(el.ue_ipv4());
+    size_t h3 = std::hash<std::string>()(el.ue_ipv4());
```
uri200: shouldn't this be ipv6?

uri200 (changes requested):

lte/gateway/c/session_manager/LocalEnforcer.h
themarwhal: yea good catch resolved.

uri200 (approved):

koolzz (approved):
lgtm, should we test it on spirient before merging?

PR #5371, opened 2021-03-08
[AGW][MME] Remove IMSI SPGW state only when tunnel list is empty

ssanadhya:

## Summary

The `sgw_cm_remove_bearer_context_information` is called when a UE session is deleted or default bearer is deleted. In case of multiple PDNs per UE, this logic would lead to Redis state being deleted for the IMSI when default bearer for any PDN is deleted. This change fixes the logic to only delete state from Redis when the last bearer is removed for a given UE.

Also updated the `test_secondary_pdn_with_dedicated_bearer_multiple_services_restart.py` to use the latest API for sending ReAuthRequest.

## Test Plan

Regression testing: `make integ_test`
Functional testing
- `test_secondary_pdn_with_dedicated_bearer_multiple_services_restart.py`

TBD: Add new test case with delete bearer checks

### DIFF
--- a/lte/gateway/c/oai/tasks/mme_app/mme_app_context.c
+++ b/lte/gateway/c/oai/tasks/mme_app/mme_app_context.c
@@ -175,7 +175,7 @@ void mme_app_ue_sgs_context_free_content(
   }
   // Stop SGS EPS Detach indication timer if running
   if (sgs_context_p->ts8_timer.id != MME_APP_TIMER_INACTIVE_ID) {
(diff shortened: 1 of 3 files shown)


ssanadhya:
MAGMA-OAI-TESTS

ulaskozat (approved):
LGTM

ssanadhya:
MAGMA-OAI-TESTS

ardzoht (approved):
lgtm

PR #5373, opened 2021-03-08
[lte][agw] Adding sync of cache on RedisFlatDict init function

ardzoht:

## Summary

- On adding write-through cache to RedisFlatDict container, an issue is that on service restart, cache will be emptied out, causing a mismatch on data between redis and the cache-backed container, this PR pushes a function to sync the data on the init function

## Test Plan

- make integ_test
- Using s1ap tester (`s1aptests/test_attach_detach_with_mme_restart.py`) to ensure test passes and there's no ReleaseIPv4 address failure

## Additional Information

### DIFF
--- a/orc8r/gateway/python/magma/common/redis/containers.py
+++ b/orc8r/gateway/python/magma/common/redis/containers.py
@@ -217,6 +217,8 @@ class RedisFlatDict(MutableMapping[str, T]):
         self.serde = serde
         self.redis_type = serde.redis_type
         self.cache = {}
+        if self._writethrough:
+            self._sync_cache()
 
     def __len__(self) -> int:
         """Return the number of items in the dictionary."""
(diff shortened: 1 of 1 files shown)


orc8r/gateway/python/magma/common/redis/containers.py
```diff
@@ -422,5 +424,16 @@ def lock(self, key: str) -> Lock:
             strict=False,
         )
 
+    def _sync_cache(self):
+        """
+        Syncs write-through cache with redis data on store.
+        """
+        type_pattern = "*:" + self.redis_type
```
themarwhal: we should probably have a function called `get_type_pattern() -> str:` so that we make sure we're consistent everywhere.

themarwhal (approved):

mpgermano (approved):

hcgatewood (approved):

PR #5394, opened 2021-03-09
[feg] s8_proxy add rat types to proto definition

### LINKED ISSUE #4914: [Inbound Roaming][orc8r] Solidify routing configuration
- We currently have a PLMN based configuration for routing between different FeGs in the neutral host component.
- The goal of this task is to look into whether we need to expand the configuration to include APN and come up with a final model of the configuration.

uri200:

## Summary

Added rat types to protos definition so MME can transparently send the type. 

## Test Plan

/build -g
make precommit at feg

## Additional Information


feg/gateway/services/s8_proxy/servicers/ie_conversions.go
```diff
@@ -170,8 +170,28 @@ func getUserLocationIndication(mcc, mnc string, uli *protos.UserLocationInformat
 func getRatType(ratType protos.RATType) *ie.IE {
 	var rType uint8
 	switch ratType {
+	case protos.RATType_RESERVED:
```
themarwhal, line 173: Doesn't protobuf provide a function to convert between enums and its value? Or is the assignment slightly different from the actual value?

feg/protos/s8_proxy.proto
```diff
@@ -58,7 +58,17 @@ enum PDNType {
 }
 
 enum RATType {
```
themarwhal, line 60: We should converge this RatType with the one used in session_manager.proto? (I prefer converging to this one since it seems like we're mapping to the actual value)

themarwhal (approved):

feg/gateway/services/s8_proxy/servicers/ie_conversions.go
uri200: The reason for having all those (obvious) functions here is to decouple GRPC implementation of 3gpp.
Even the values are the same, they are not the same type. I felt we were kind of casting the value, which may lead to issues in the future in case of changes on grpc proto file.


Note i did that for selection mode type too where grpc indexes and 3gpp values were the same.

feg/protos/s8_proxy.proto
uri200: Well sessiond only supports those 3 types so adding the rest will not have any effect. 
Besides, Session_proxy should be the one translating those grpc values into 3gpp values. This way, as mentioned in the previous comment, we don't mix domains. 



```
enum RATType {
  TGPP_LTE = 0;  // will map to 06 (EUTRAN, 3GPP TS 29.274) for Gy, 1004 (EUTRAN, 3GPP TS 29.212) S6a, Gx
  TGPP_WLAN = 1; // will map to 03 (WLAN, 3GPP TS 29.274) for Gy, 0 (WLAN, 3GPP TS 29.212) S6a, Gx
  TGPP_NR = 2;
}
```

The reason I implemented them all in s8_proxy is that s8_proxy just translates whatever MME sends.

PR #5400, opened 2021-03-10
[orc8r] Create standalone specs at build time

AndyLKhuu:

## Summary
Per Service UI was not fully functional due to the complex level of dependencies that generated objects have on each other in our code base. This PR solves this problem by generating standalone specs for every service at build time and then serving the spec on Per Service UI requests.

## Test Plan
- [x] cd magma/orc8r/cloud/docker ./build.py -t
- [x] Manually iterating through Swagger UI to ensure per service UI works for all endpoints. 

## Additional Information


orc8r/cloud/go/obsidian/swagger/protos/swagger.proto
```diff
@@ -17,8 +17,11 @@ option go_package = "protos";
 
 // SwaggerSpec provides a Swagger specification of the service's REST API.
 service SwaggerSpec {
- // GetSpec returns the Swagger specification of the service's REST API.
- rpc GetSpec(GetSpecRequest) returns (GetSpecResponse) {}
+ // GetPartialSpec returns partial Swagger specification of the service's REST API.
+ rpc GetPartialSpec(GetSpecRequest) returns (GetSpecResponse) {}
```
hcgatewood: Each Get*Spec should have its own unique req+res message definitions

orc8r/cloud/go/tools/swaggergen/generate/generate.go
```diff
@@ -255,3 +301,33 @@ func readSwaggerSpec(filepath string) (MagmaSwaggerSpec, error) {
 	}
 	return spec, nil
 }
+
+// marshalToYAML marshals the passed Swagger spec to a YAML-formatted string.
+func marshalToYAML(spec MagmaSwaggerSpec) (string, error) {
```
hcgatewood: Make this a method on the object

orc8r/cloud/go/tools/swaggergen/main.go
```diff
@@ -100,6 +103,19 @@ func main() {
 		glog.Fatalf("Error parsing swagger spec dependency tree: %v\n", err)
 	}
 
+	if *generateStandAloneSpec && *targetFilepath != "swagger-common.yml" {
+		// Retrieve spec service name
+		splitPath := strings.Split(*targetFilepath, "/")
+		outFile := splitPath[len(splitPath)-4] + ".swagger.v1.yml"
+		outPath := filepath.Join(os.Getenv("MAGMA_ROOT"), "orc8r/cloud/swagger/specs/standalone", outFile)
```
hcgatewood: Use rootDir rather than env var

orc8r/cloud/go/tools/swaggergen/main.go
```diff
@@ -100,6 +103,19 @@ func main() {
...
+		err = generate.GenerateStandAloneSpec(*targetFilepath, specs, outPath)
+		if err != nil {
+			// Swallow GenerateStandAloneSpec error because build should
```
hcgatewood: In this case, we actually want to fatal. Since it's a tool, it's better to fail fast and loudly

orc8r/cloud/go/tools/swaggergen/main.go
```diff
@@ -100,6 +103,19 @@ func main() {
 		glog.Fatalf("Error parsing swagger spec dependency tree: %v\n", err)
 	}
 
+	if *generateStandAloneSpec && *targetFilepath != "swagger-common.yml" {
+		// Retrieve spec service name
+		splitPath := strings.Split(*targetFilepath, "/")
```
hcgatewood: https://golang.org/pkg/path/filepath/#Base

orc8r/cloud/go/tools/swaggergen/main.go
```diff
@@ -100,6 +103,19 @@ func main() {
...
+		outFile := splitPath[len(splitPath)-4] + ".swagger.v1.yml"
```
hcgatewood: I think the basename of the filepath is already in the correct format -- we want same format, just different directory (partial vs. standalone)

hcgatewood (commented):
Added some comments for the preliminary PR. Will provide code cleanliness comments tomorrow, though things look pretty solid

rdefosse:
MAGMA-OAI-TESTS

rdefosse:
Just using this PR to see if the OAI CI pipeline is properly bypassed on this PR

orc8r/cloud/go/tools/swaggergen/main.go
AndyLKhuu: It seems to me that all the targetFilePaths are actually only "swagger.v1.yml" besides "swagger-common.yml". This seems to reflect what is passed into the gen.go files of each service.

orc8r/cloud/go/obsidian/swagger/protos/swagger.proto
AndyLKhuu: Besides the benefit of just having separated definitions which may be useful for flexible development in the future, is there a good reason why using shared definitions is bad? Especially considering how simple and identical in functionality they are?

orc8r/cloud/go/obsidian/swagger/protos/swagger.proto
```diff
@@ -17,12 +17,21 @@ option go_package = "protos";
 
 // SwaggerSpec provides a Swagger specification of the service's REST API.
 service SwaggerSpec {
- // GetSpec returns the Swagger specification of the service's REST API.
- rpc GetSpec(GetSpecRequest) returns (GetSpecResponse) {}
+ // GetPartialSpec returns partial Swagger specification of the service's REST API.
+ rpc GetPartialSpec(PartialSpecRequest) returns (PartialSpecResponse) {}
```
hcgatewood: Nit: add a newline below, to separate each rpc definition

orc8r/cloud/go/obsidian/swagger/protos/swagger.proto
```diff
@@ -17,12 +17,21 @@ option go_package = "protos";
...
+ // GetStandaloneSpec returns a standalone Swagger specification of the
+ // service's REST API.
```
hcgatewood: Nit for this file: proto file indentation is two spaces

orc8r/cloud/go/obsidian/swagger/protos/swagger.proto
```diff
@@ -17,12 +17,21 @@ option go_package = "protos";
...
+ rpc GetStandaloneSpec(StandaloneSpecRequest) returns (StandaloneSpecResponse) {}
 }
 
-message GetSpecRequest {}
+message PartialSpecRequest {}
```
hcgatewood, line 28: Nit: group these by rpc, rather than by req/res

orc8r/cloud/go/obsidian/swagger/spec/combine.go
```diff
@@ -146,7 +146,7 @@ func makeAllYAMLReferencesLocal(yamlContents []string) []string {
 	// Match on any yml reference to file_name_here.foo.bar#/baz
 	// and change those references to #/baz (strip the prefix)
 	// e.g. $ref: 'foo_bar_baz.blah#/asdf' -> $ref: '#/asdf'
-	ymlRefRe := regexp.MustCompile(`(\$ref:\s*)['"].+(#/.+)['"]`)
+	ymlRefRe := regexp.MustCompile(`(\$ref:\s+)['"]?.*(#/[^"'\s]+)["']?`)
```
hcgatewood: Nit: put all single vs. double quotes in the same order. So change the first `['"]` to `["']`

orc8r/cloud/go/obsidian/swagger/poll.go
```diff
@@ -36,9 +36,9 @@ func GetCombinedSpec(yamlCommon string) (string, error) {
 
 	var yamlSpecs []string
 	for _, s := range servicers {
-		yamlSpec, err := s.GetSpec()
+		yamlSpec, err := s.GetPartialSpec()
 		if err != nil {
-			// Swallow GetSpec error because the polling should continue
+			// Swallow GetPartialSpec error because the polling should continue
```
hcgatewood: Nit: you can just say "Swallow error...", no need to include the name of the function

orc8r/cloud/go/obsidian/swagger/poll_test.go
```diff
@@ -85,7 +85,7 @@ func Test_GetCombinedSwaggerSpecs(t *testing.T) {
 	defer registry.RemoveServicesWithLabel(orc8r.SwaggerSpecLabel)
 
 	for i, s := range services {
-		registerServicer(t, s, tags[i])
+		registerServicer(t, s, tags[i], swagger_lib.TagDefinition{})
```
hcgatewood: Side note: is this PR just on an old commit? Cause didn't we remove the whole swagger_lib thing?

orc8r/cloud/go/obsidian/swagger/spec_servicer.go
```diff
@@ -26,37 +26,51 @@ import (
 )
 
 type specServicer struct {
-	spec string
+	partialSpec    string
+	standAloneSpec string
```
hcgatewood: Nit: `standAloneSpec` -> `standaloneSpec`

orc8r/cloud/go/obsidian/swagger/spec_servicer.go
```diff
@@ -26,37 +26,51 @@ import (
...
 func NewSpecServicerFromFile(service string) protos.SwaggerSpecServer {
 	service = strings.ToLower(service)
-	path := getSpecPath(service)
-	data, err := ioutil.ReadFile(path)
+	partialPath, standalonePath := getSpecPaths(service)
+	partial, err := ioutil.ReadFile(partialPath)
 	if err != nil {
-		// Swallowing ReadFile error because the service should
-		// continue to run even if it can't find its Swagger spec file.
+		// Swallowing ReadFile error because the service should continue to
```
hcgatewood: Nit: same as above, no need to specify the name of the function whose error you're swallowing

orc8r/cloud/go/tools/swaggergen/generate/generate.go
```diff
@@ -165,6 +167,50 @@ func GenerateModels(targetFilepath string, configFilepath string, rootDir string
 	return nil
 }
 
+// GenerateStandAloneSpec generates a standalone spec file for a particular
+// Swagger spec.
+func GenerateStandAloneSpec(targetFilePath string, specs map[string]MagmaSwaggerSpec, outPath string) error {
```
hcgatewood: Nit: StandAlone -> Standalone

hcgatewood: Nit: FilePath -> Filepath

orc8r/cloud/go/tools/swaggergen/generate/generate.go
```diff
@@ -165,6 +167,50 @@ func GenerateModels(targetFilepath string, configFilepath string, rootDir string
...
+			dependencies = append(dependencies, yamlSpec)
+		}
+	}
+
+	combined, warnings, err := spec.Combine(yamlCommon, dependencies)
+	if err != nil {
+		return err
+	}
+	if warnings != nil {
+		glog.Infof("Some Swagger spec traits were overwritten or unable to be read: %+v", warnings)
```
hcgatewood: Since this is a tool: we want to fail fast and loudly. So return an error here, and make sure it gets bubbled up into a fatal somewhere in the top of the callstac

orc8r/cloud/go/tools/swaggergen/generate/generate.go
```diff
@@ -242,6 +288,17 @@ func StripAndWriteSwaggerSpecs(specs map[string]MagmaSwaggerSpec) error {
 	return nil
 }
 
+// RetrieveServiceName parses the targetFilepath for its originating service.
+func RetrieveServiceName(targetFilePath string, rootDir string) (string, error) {
```
hcgatewood: This is actually returning the output filepath. So maybe GetStandaloneFilepath

orc8r/cloud/go/tools/swaggergen/generate/generate.go
```diff
@@ -242,6 +288,17 @@ func StripAndWriteSwaggerSpecs(specs map[string]MagmaSwaggerSpec) error {
...
+	absTargetFilepath, err := filepath.Abs(targetFilePath)
+	if err != nil {
+		return "", errors.Wrapf(err, "target filepath %s is invalid", targetFilePath)
+	}
+	splitPath := strings.Split(absTargetFilepath, "/")
```
hcgatewood: https://golang.org/pkg/path/filepath/#SplitList

orc8r/cloud/go/tools/swaggergen/generate/generate.go
```diff
@@ -242,6 +288,17 @@ func StripAndWriteSwaggerSpecs(specs map[string]MagmaSwaggerSpec) error {
...
+	outFile := splitPath[len(splitPath)-4] + ".swagger.v1.yml"
```
hcgatewood: This is too magic. Add a comment describing what/why you're grabbing that part of the filepath

orc8r/cloud/go/tools/swaggergen/generate/generate.go
hcgatewood: This is only being used in one spot. Consider making it a private function called from GenerateStandaloneSpec

orc8r/cloud/go/tools/swaggergen/generate/generate.go
```diff
@@ -242,6 +288,17 @@ func StripAndWriteSwaggerSpecs(specs map[string]MagmaSwaggerSpec) error {
 	return nil
 }
 
+// RetrieveServiceName parses the targetFilepath for its originating service.
+func RetrieveServiceName(targetFilePath string, rootDir string) (string, error) {
+	absTargetFilepath, err := filepath.Abs(targetFilePath)
+	if err != nil {
+		return "", errors.Wrapf(err, "target filepath %s is invalid", targetFilePath)
```
hcgatewood: "standalone output filepath %s is invalid"

orc8r/cloud/go/tools/swaggergen/generate/generate_test.go
```diff
@@ -47,6 +47,29 @@ func TestParseSwaggerDependencyTree_Cycle(t *testing.T) {
 	assert.Equal(t, expected, actual)
 }
 
+func TestGenerateStandaloneSpec(t *testing.T) {
+	goldenFilePath := "../testdata/out.yml.golden"
```
hcgatewood: `out` isn't descriptive enough to understand what it is just based on the name

orc8r/cloud/go/tools/swaggergen/generate/generate.go
```diff
@@ -165,6 +167,50 @@ func GenerateModels(targetFilepath string, configFilepath string, rootDir string
...
+	absTargetFilepath, err := filepath.Abs(targetFilePath)
+	if err != nil {
+		return errors.Wrapf(err, "target filepath %s is invalid", targetFilePath)
+	}
+
+	var dependencies []string
+	var yamlCommon string
+	for specFilePath, s := range specs {
+		if specFilePath != absTargetFilepath {
+			s.Paths = nil
```
hcgatewood: Add comment on why we're clearing these

orc8r/cloud/go/tools/swaggergen/generate/generate.go
```diff
@@ -165,6 +167,50 @@ func GenerateModels(targetFilepath string, configFilepath string, rootDir string
...
 
+// GenerateStandAloneSpec generates a standalone spec file for a particular
+// Swagger spec.
+func GenerateStandAloneSpec(targetFilePath string, specs map[string]MagmaSwaggerSpec, outPath string) error {
+	absTargetFilepath, err := filepath.Abs(targetFilePath)
+	if err != nil {
+		return errors.Wrapf(err, "target filepath %s is invalid", targetFilePath)
+	}
+
+	var dependencies []string
```
hcgatewood: This is actually yamlSpecs, not just dependencies, because it includes the target spec

orc8r/cloud/Makefile
```diff
@@ -102,7 +104,7 @@ gen: tools $(GEN_LIST)
 $(GEN_LIST): %_gen:
 	make -C $*/cloud/go gen
 
-fullgen: clean_gen gen swagger tidy
+fullgen: clean_gen swagger_directories gen swagger tidy
```
hcgatewood: swagger_directories actually is a requirement of gen now, so move it to `gen: tools swagger_directories $(GEN_LIST)`

hcgatewood (changes requested):
Looks good. Mostly small stuff. Lmk when updated and I'll approve

orc8r/cloud/go/tools/swaggergen/generate/generate.go
AndyLKhuu: The error is actually coming from reading the target file path and not the output directory. The output path is constructed after.

orc8r/cloud/go/tools/swaggergen/generate/generate.go
AndyLKhuu: I don't think SplitList wold be beneficial in this scenario since it works to split by List Separators (":") and not Path Separators ("/")

orc8r/cloud/Makefile
```diff
@@ -27,13 +27,15 @@ SWAGGER_V1_ROOT := $(SWAGGER_ROOT)/v1
 SWAGGER_V1_YML := $(SWAGGER_ROOT)/v1/swagger.yml
 SWAGGER_V1_CONFIG := $(MAGMA_ROOT)/orc8r/cloud/docker/controller/goswagger-config.yml
 SWAGGER_V1_SPECS_DIR := $(MAGMA_ROOT)/orc8r/cloud/swagger/specs
-SWAGGER_V1_COMMON_DIR := common
+SWAGGER_V1_PARTIAL_SPECS_DIR := $(SWAGGER_V1_SPECS_DIR)/partial
+SWAGGER_V1_COMMON_DIR := $(SWAGGER_V1_SPECS_DIR)/common
```
hcgatewood, line 31: Export the two new env vars you're creating

orc8r/cloud/Makefile
```diff
@@ -131,14 +133,17 @@ lint_tools:
...
-	combine_swagger --in=$(SWAGGER_V1_SPECS_DIR) --common=$(SWAGGER_V1_SPECS_DIR)/$(SWAGGER_V1_COMMON_DIR)/$(SWAGGER_COMMON) --out=$(SWAGGER_V1_YML)
+swagger: $(SWAGGER_LIST)
+	cp $(MAGMA_ROOT)/orc8r/cloud/go/models/swagger-common.yml $(SWAGGER_V1_COMMON_DIR)/$(SWAGGER_COMMON)
+	combine_swagger --in=$(SWAGGER_V1_PARTIAL_SPECS_DIR) --common=$(SWAGGER_V1_COMMON_DIR)/$(SWAGGER_COMMON) --out=$(SWAGGER_V1_YML)
 	swagger validate $(SWAGGER_V1_YML)
 
 swagger_directories:
 	rm -rf $(SWAGGER_V1_SPECS_DIR)
-	mkdir -p $(SWAGGER_V1_SPECS_DIR)/$(SWAGGER_V1_COMMON_DIR)
+	mkdir -p $(SWAGGER_V1_COMMON_DIR)
```
hcgatewood: You can do these in a single `mkdir -p dir1 dir2 dir3`

orc8r/cloud/go/obsidian/swagger/handlers/handlers_test.go
```diff
@@ -65,7 +65,7 @@ func Test_GenerateCombinedSpecHandler(t *testing.T) {
 	defer registry.RemoveServicesWithLabel(orc8r.SwaggerSpecLabel)
 
 	for i, s := range services {
-		registerServicer(t, s, tags[i])
+		registerServicer(t, s, tags[i], swagger_lib.TagDefinition{})
```
hcgatewood: Everywhere we're aliasing as `swagger_lib`, let's go ahead and convert to `spec` alias. You can add that change as part of this pr for expediency

orc8r/cloud/go/tools/swaggergen/generate/generate.go
```diff
@@ -165,6 +166,51 @@ func GenerateModels(targetFilepath string, configFilepath string, rootDir string
...
+	absTargetFilepath, err := filepath.Abs(targetFilepath)
+	if err != nil {
+		return errors.Wrapf(err, "target filepath %s is invalid", targetFilepath)
+	}
+
+	var yamlSpecs []string
+	var yamlCommon string
+	for specFilepath, s := range specs {
+		if specFilepath != absTargetFilepath {
+			// Clearing fields not utilized by target service
```
hcgatewood: This comment describes "what" -- but when we're adding an inline comment, we generally want to describe "why"

orc8r/cloud/go/tools/swaggergen/generate/generate.go
hcgatewood: Ah good catch. Okay then use strings.Split but with https://golang.org/pkg/os/#PathSeparator

orc8r/cloud/go/tools/swaggergen/generate/generate.go
```diff
@@ -242,6 +288,19 @@ func StripAndWriteSwaggerSpecs(specs map[string]MagmaSwaggerSpec) error {
...
+// GetStandaloneFilepath parses the targetFilepath for its originating service.
+func GetStandaloneFilepath(targetFilePath string, rootDir string) (string, error) {
+	absTargetFilepath, err := filepath.Abs(targetFilePath)
+	if err != nil {
+		return "", errors.Wrapf(err, "target filepath %s is invalid", targetFilePath)
+	}
+	// Iterate through the file path to retrieve the service name of
+	// the Swagger spec
+	splitPath := strings.Split(absTargetFilepath, "/")
+	outFile := splitPath[len(splitPath)-4] + ".swagger.v1.yml"
```
hcgatewood: Still missing comment on why `len(splitPath)-4` -- when I read this in isolation I think: "why 4 ??"

orc8r/cloud/go/tools/swaggergen/generate/generate_test.go
hcgatewood: Would still like this to be something like standalone.yml.golden

hcgatewood (approved):
Please address comments then msg to merge

hcgatewood (approved):
👍

mpgermano (approved):

PR #5401, opened 2021-03-10
[orcr8] Reverse enable field at FederatedModeMap

uri200:

## Summary

To avoid issues with MME mconfig parser, we will revert this condition. 

## Test Plan

`./build.py --test --up`

## Additional Information


feg/cloud/go/services/feg/obsidian/models/swagger.v1.yml
```diff
@@ -1569,12 +1569,12 @@ definitions:
     description: Mapping for PLMN, imsi ranges, vs its gateway mode
     type: object
     properties:
-      enabled:
-        description: If Enabled is false, mapping will not be applied
+      disabled:
```
themarwhal, line 1572: Having a disabled flag instead of enabled feels a bit counterintuitive to me. If we really want to send a flipped logic down to MME, we could flip the logic when we are translating to the Mconfig. 
Up to you though.

uri200: I agree. 

The reason I did this is mconfig at AGW doesn't show the field `enabled` if it is false. So I wanted to avoid something like what you see below: enabled is `false` but mconfig doesn't show that field. Yet, you still see the config.

Another way to do it is just not sending the mapping in case enabled is false. But then we have logic in the mconfig builder that depends on the content of the configuration itself. And i don't like that either.

mconfig should not be read by humans, so probably MME mconfig parser is able to see that `enabled` is `false` in that example and will ignore the conttent of `mapping`

So maybe this PR is not needed

```
            "federatedModeMap": {
                "mapping": [
                    {
                        "apn": "internet.com",
                        "imsiRange": "000000000000000:000019999999999",
                        "mode": "LOCAL_SUBSCRIBER",
                        "plmn": "123456"
                    }
                ]
            },
```

uri200: now that I am checking, I am not sure why enabled: false is not shown.... I will investigate that

(below you have false parameters from same mconfig file)
```
            "featureFlags": {
                "newfeature1": true,
                "newfeature2": false
            },
```

Issue #4866, opened 2021-02-10
[CI] Annotate GH Pull Requests with cranked up GCC / warning flags


electronjoe:
See the comprehensive list of GCC warning flags listed in the [cpp best practices](https://lefticus.gitbooks.io/cpp-best-practices/content/02-Use_the_Tools_Available.html) Gitbook.

This will drastically increase warning counts - and we should dedicate team effort (maybe clean-up days or some similar concept) to addressing the warnings being reported.

We should also track the warning count over time via CI tooling.

electronjoe:
@themarwhal noted that things like warning counts might be exposed in CI via CircleCI [Insights](https://circleci.com/blog/announcing-new-insights-endpoints-in-circleci-s-api-v2/) for example. This and other metrics.

PR #5404, opened 2021-03-10
[CI] Enable GCC Warning Annotations on GH Pull Requests

electronjoe:
# Summary

Here is an example GCC Warning Annotation directly on a Pull Request.

Screen Shot 2021-03-14 at 7 24 26 AM

Among several motivations for this change - I feel that by enabling GCC warning annotations we can now move forward with removal of `-Werror` compiler flags (#5358).

The approach of this PR can now be applied to also annotate Clang 11 warnings, Clang-Tidy findings, and Clang-Format findings (with some extra work required for each - but they will be much easier to add now).

This PR achieves the annotations through the following.

- Dockerfile capable of build and analysis of Magma c/c++ targets
  - Using Ubuntu 20.04
  - GCC 9.3 (default 20.04)
- A Github Workflow that emits Pull Request GCC Warning Annotations
  - That show up directly in-line during code review
  - Using [electronjoe/gcc-problem-matcher](https://github.com/electronjoe/gcc-problem-matcher)

# Some Details

## Annotation Count Limits

There are annotation limits in GitHub Actions:

- 10 warning annotations and 10 error annotations per step
- 50 annotations per job (sum of annotations from all the steps)
- 50 annotations per run (separate from the job annotations, these annotations aren’t created by users)

Given that our codebase now has 600 GCC 9.3 warnings (with `-Wextra` enabled), this PR had to filter them.  Annotations are filtered and emitted only for modified files.

** The implication is that only the first 10 Annotations will be emitted, for the modified files. **

This could at times miss things, until we burn down some of the GCC warnings being emitted.

## Determining changed files

In order to determine the set of files modified by the PR - we use the `get-changed-files@v2` github action.  The original implementation (jiterrbit/get-changed-files) has some weaknesses that result in action-terminating warnings under the Magma workflow (any time the forked-branch is not based upon origin-master).  Pragmatically, the warning emitted by jiterrbit's github action does not often impact the output results, so we use here a fork of get-changed-files by mmagician which disables the error / halt under this warning.

** The implication however is that the set of files filtered as having changes in a Pull Request could, in certain situations, not reflect reality due to the complexity of the merge situation. **

### DIFF
--- a/.github/workflows/gcc-problems.yml
+++ b/.github/workflows/gcc-problems.yml
@@ -1,0 +1,182 @@
+#######
+# This pipeline runs various static analysis (e.g. GCC warnings) against the c / c++ pull requests.
+#
+# TODO: Move Docker container caching / storage to a repository
+#
+# TODO: Reduce workflow description duplication across jobs
+#   - Option: by use of Workflow Templates per gcc-build-target (need to move docker build to another workflow and requires container repo)
+#   - Option: by improving our build system and enabling faster build-all-targets
+#######
+name: "GCC Warnings & Errors"
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+    paths:
+      - lte/gateway/c/**
+      - orc8r/gateway/c/**
+
+# See [Example Sharing Container Between Jobs](https://github.com/docker/build-push-action/issues/225)
+jobs:
+  gen_build_container:
+    runs-on: ubuntu-latest
+    steps:
+      - 
+        name: Check Out Repo 
+        uses: actions/checkout@v2
+      - 
+        name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      - 
+        name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+      - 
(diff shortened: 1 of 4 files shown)


electronjoe:
MAGMA-OAI-TESTS

themarwhal (approved):

uri200 (approved):

koolzz (approved):

rdefosse:
MAGMA-OAI-TESTS

tmdzk (approved):
LOVE IT! 

But please can you fix the few comments that Github Action posted on your PR? 😂

rdefosse (approved):
Good.

@electronjoe Scott

I have a question for the whole CI team.

On which resources these checks are running?

ardzoht (approved):
This is a great improvement, lgtm.

electronjoe:
> Good.
> 
> @electronjoe Scott
> 
> I have a question for the whole CI team.
> 
> On which resources these checks are running?

We spoke out of band, these GH Actions are executing on Github infrastructure. Today [CircleCI API to Github does not support Annotations](https://circleci.canny.io/cloud-feature-requests/p/see-specific-line-annotations-in-github-checks-when-pr-fails-lint-job).  One could write their own Annotation implementation against the GH annotation API - but that seemed highly non ideal.

It's good to be concerned over CI creep - and over how free GH Actions remain for us.  Should we need to migrate this away from GitHub Actions - the dockerfile and general approach are portable. We would need to author annotations capability in the target CI environment.

electronjoe:
This largely resolves GH Issue #4866

PR #5408, opened 2021-03-10
[AGW][Health] Update agw_health_cli + health service to handle error better

themarwhal:
## Summary
When `hssRelayEnabled` is missing from the mconfig, the health script fails with the following error: 
```
magma@agw2:~$ agw_health_cli.py
Access Gateway health summary
Error: 'hssRelayEnabled'
magma@agw2:~$
```
This PR fixes this problem by adding a default value=False.

Additionally, this PR fixes the incorrect coredump path as they are now generated into `/var/core/` not `/tmp/`.

## Test Plan
Run the agw_health_script

## Additional Information


ulaskozat (approved):
LGTM

ardzoht (approved):
lgtm

PR #5430, opened 2021-03-11
[orc8r][docs] Move state indexer docs to Docusaurus

hcgatewood:
## Summary

They're hidden away in a doc.go, and are useful/comprehensive enough to be on Docusaurus.

## Test Plan

n/a

## Additional Information


docs/readmes/orc8r/dev_indexers.md
```diff
@@ -0,0 +1,102 @@
...
+
+Orc8r stores reported gateway state as blobs, keyed by the primary key triplet `{network ID, state type, state key}`. We call this *primary state*. With this pattern, services can access reported state of a particular type with the state's key, which is often a well-known value like an IMSI or gateway hardware ID.
+
+However, sometimes services need to access a piece of state and all they know is the value of part of the blob. For example, finding the IMSI that owns an IP address.
+
+To support this functionality, Orc8r provides the state indexer pattern. State indexers receive reported state at the time of reporting, optionally transform it, then write it to their own storage under a new primary key of their choosing. We call this transformed state *derived state*.
+
+## Overview
+
+### State service
```
karthiksubraveti, line 23: Might be useful to add some details on clients of state service is in the our system and provide some context on why this is state reindexing is useful from product's perspective. Perhaps it let's us add REST API with different getters etc. 

Also just a minor note, from curiosity perspective, Is there a place where we can look for all state keys available in our system from the code(not from the tables themselves).

karthiksubraveti (approved):
Thanks for moving this to docusaurus. Changes LGTM. Left a few comments.

mpgermano (approved):

docs/readmes/orc8r/dev_indexers.md
hcgatewood: Unfortunately not at the moment. Next time I make a state change I'm planning to standardize things a bit so it's easier to find

PR #5434, opened 2021-03-11
AGW: pipelined: Set direction for inbound traffic,

pshelar:
in case of inbound roaming, the packet ingress on PGW GTP tunnel
port. This packets are marked as pass through packets. use that to 
to set direction bit for these packets.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/inout.py
+++ b/lte/gateway/python/magma/pipelined/app/inout.py
@@ -34,7 +34,7 @@ from magma.pipelined.bridge_util import BridgeTools, DatapathLookupError
 from magma.pipelined.openflow.magma_match import MagmaMatch
 from magma.pipelined.openflow.messages import MessageHub, MsgChannel
 from magma.pipelined.openflow.registers import load_direction, Direction, \
-    PASSTHROUGH_REG_VAL, TUN_PORT_REG, PROXY_TAG_TO_PROXY
+    PASSTHROUGH_REG_VAL, TUN_PORT_REG, PROXY_TAG_TO_PROXY, REG_ZERO_VAL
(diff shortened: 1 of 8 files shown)


lte/gateway/python/magma/pipelined/app/inout.py
```diff
@@ -363,14 +363,26 @@ def _get_default_ingress_flow_msgs(self, dp):
             )
 
         # set a direction bit for outgoing (pn -> inet) traffic for remaining traffic
-        match = MagmaMatch()
+        # Passthrough is zero for packets from eNodeB GTP tunnels
+        ps_match_out = MagmaMatch(passthrough=0)
```
koolzz: Maybe define const NO_PASSTHROUGH_REG_VAL = 0 in register.py?  or just use REG_ZERO_VAL

lte/gateway/python/magma/pipelined/app/inout.py
```diff
@@ -363,14 +363,26 @@ def _get_default_ingress_flow_msgs(self, dp):
...
         msgs.append(
-            flows.get_add_resubmit_next_service_flow_msg(dp, self._ingress_tbl_num,match,
+            flows.get_add_resubmit_next_service_flow_msg(dp, self._ingress_tbl_num, ps_match_out,
+                                                 actions=actions,
+                                                 priority=flows.MINIMUM_PRIORITY,
+                                                 resubmit_table=next_table)
+        )
+        # Passthrough is one for packets from remote PGW GTP tunnels, set direction
+        # flag to IN for such packets.
+        ps_match_in = MagmaMatch(passthrough=1)
```
koolzz: Use PASSTHROUGH_REG_VAL instead of 1

koolzz (approved):
minor nits, otherwise lgtm

PR #5443, opened 2021-03-11
Revert "[PipelineD][SessionD] Cleanup flows for non-existent sessions"

themarwhal:
Reverts magma/magma#5368


koolzz (approved):
Something in this pr removes flows way too quickly, when running stress test after attaches all flows were gone in a second(even before detaches started). 
Lets roll back and investigate

PR #5445, opened 2021-03-11
[orc8r] reverse condition to check network type on lte mconfig builder

uri200:

## Summary

It looks like some LTE networks at staging include `feg_netowrk_id` as part of its configuration. That is unexpected since those LTE network are not federated and shouldn't include that field. 

That field is causing the builder to think those are federated.

In order to avoid that, this PR reverse the way we are checking this is for sure a federated network. First it checks for the type and then it checks for the existence of the field

## Test Plan
```
./build -g
./build --test --up
```

## Additional Information

### DIFF
--- a/lte/cloud/go/services/lte/servicers/builder_servicer.go
+++ b/lte/cloud/go/services/lte/servicers/builder_servicer.go
@@ -81,7 +81,7 @@ func (s *builderServicer) Build(ctx context.Context, request *builder_protos.Bui
 		return nil, err
 	}
 
-	federatedNetworkConfigs, err := getFederatedNetworkConfigs(cellularNwConfig.FegNetworkID, request)
+	federatedNetworkConfigs, err := getFederatedNetworkConfigs(network.Type, cellularNwConfig.FegNetworkID, request)
 	if err != nil {
 		glog.Errorf("Failed to retrieve LTE_federated network config while building lte mconfig")
 		return nil, err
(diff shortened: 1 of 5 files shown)


lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -528,18 +528,21 @@ func getServiceAreaMaps(serviceAreaMaps map[string]lte_models.TacList) map[strin
...
-func getFederatedNetworkConfigs(fegId lte_models.FegNetworkID, request *builder_protos.BuildRequest) (*feg_models.FederatedNetworkConfigs, error) {
-	if fegId == "" {
+func getFederatedNetworkConfigs(networkType string, fegId lte_models.FegNetworkID, request *builder_protos.BuildRequest)(*feg_models.FederatedNetworkConfigs, error) {
+	if networkType != feg.FederatedLteNetworkType {
 		// this is a non federated network, return nothing
 		return nil, nil
 	}
+	if fegId == "" {
+		glog.Warning(fmt.Errorf("Cant get LTE Federated Configs since network is of type %s, " +
+			"but feg ID is not defined: %s. If this is non federated network, remove the `fed_netowrk id` field. " +
```
mpgermano: I think the section `If this is non federated network, remove the `fed_netowrk id` field. ` is unnecessary as we're already checking network type.

mpgermano (approved):

lte/cloud/go/services/lte/servicers/builder_servicer.go
uri200: You are right. That doesn't make sense anymore.
Let me simplify the message

PR #5452, opened 2021-03-12
[feg] s8_proxy use PdnAddressAllocation instead of string for CreateSessionResponse And adition of GtpError to Responses

uri200:

## Summary

Replaces string subscriber_ip for `PdnAddressAllocation paa` to be consistent with CreateSessionRequest 

Added some tests cases and fixed some missing `gtpv2` variables

Also added GtpError to the GRPC responses. s8_proxy will not error now when there is a GTP Error cause. Instead, it will forward the message to MME.

## Test Plan

make precommit at feg
teravm

## Additional Information


feg/gateway/services/s8_proxy/servicers/ie_conversions.go
```diff
@@ -173,25 +173,25 @@ func getRatType(ratType protos.RATType) *ie.IE {
...
+		rType = gtpv2.RATTypeVirtual
 	case protos.RATType_EUTRAN_NB_IOT:
-		rType = 8
+		rType = gtpv2.RATTypeEUTRANNBIoT
 	case protos.RATType_LTE_M:
-		rType = 9
+		rType = gtpv2.RATTypeLTEM
 	case protos.RATType_NR:
-		rType = 10
+		rType = gtpv2.RATTypeNR
```
uri200, line 194: Values are the same, but forgot to use the right type (from `gtpv2`)

feg/gateway/services/s8_proxy/servicers/proto_conversions.go
```diff
@@ -190,3 +190,27 @@ func handleFTEID(fteidIE *ie.IE) (*protos.Fteid, uint8, error) {
...
+		paa.Ipv6Address = paaIE.MustIPv6().String()
+	case gtpv2.PDNTypeIPv4v6:
+		pdnType = protos.PDNType_IPV4V6
+		paa.Ipv6Address = paaIE.MustIPv6().String()
+		paa.Ipv6Address = paaIE.MustIPv6().String()
+	case gtpv2.PDNTypeNonIP:
+		pdnType = protos.PDNType_NonIP
+	}
+	return &paa, pdnType, nil
+}
```
uri200, line 220: same as before, conversion could be direct since GRPC values and GTP values are the same. But I want to keep both implementation separated.

themarwhal (approved):

Issue #5412, opened 2021-03-10
Secure southbound interface with endpoint allowlist


hcgatewood:
Currently, all registered gateways have access to *every* Orc8r-internal gRPC endpoint. (Side note: in production, we block access to certifier service.)

Instead, transition to allowlist-style access enforcement at Orc8r proxy, similar to [this `nginx.conf`](https://github.com/magma/magma/blob/419cd8465bff5285a9d056fa05ec1f12b45384b6/cwf/gateway/integ_tests/nginx.conf#L56-L60).

PR #5457, opened 2021-03-12
[orc8r][lib] Security options for internal only orc8r services

emakeev:

## Summary

APIs to secure internal only orc8r services. 
Internal only services are orc9r services which can only be called by other orc8r services or obsidian, they should not
allow RPCs from Magma Gateways.

Currently, every orc8r intra-service RPC context contains a special Synthetic Orchestrator Certificate metadata, existing security middleware will block any RPC which doesn't have either Synthetic Orchestrator Certificate metadata or real Gateway Certificate Metadata.
When RPC reaches an orc8r service method, its context will either have Gateway Certificate Metadata or (in case of Synthetic Orchestrator Certificate) no certificate metadata at all. It should allow us to distinguish between Gateway originated RPC and orc8r service originated RPC without requiring additional configuration & deployment  overhead proposed in https://github.com/magma/magma/issues/5412

Developers can secure an internal only orc8r  service by using the provided NewIntraOrchestratorService() function instead of NewOrchestratorService() which allows remote GW RPCs.

If the service has mixed use (GW RPCs for some methods & internal only RPC for others), internal only service methods can be secured using provided  IsRemoteClient() function using the following pattern:

`	if protos.IsRemoteClient(ctx) {
		return nil, status.Error(codes.PermissionDenied, "Remote GetStates request")
	}
`
see state service changes as an example.

Several existing orc8r services (accessd, directoryd, certifier, configurator, state & service_registry) were modified to use the new approach. Additional existing services may need to follow.

Developer introducing new orc8r services should determine if Gateway service accessibility is required and use the introduced patterns accordingly.

## Test Plan
unit tests, additional staging testing to follow

## Additional Information

### DIFF
--- a/orc8r/cloud/go/service/middleware/unary/handler.go
+++ b/orc8r/cloud/go/service/middleware/unary/handler.go
@@ -102,6 +102,15 @@ func MiddlewareHandler(ctx context.Context, req interface{}, info *grpc.UnarySer
 	return
 }
 
+// IntraServiceHandler is an interceptor for internal only orc8r services
+// (services which are supposed to serve only internal orc8r clients)
+func IntraServiceHandler(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
+	if err = EnforceIntraCloudRPC(ctx, info); err != nil {
+		return resp, err
+	}
+	return callHandler(ctx, req, info, handler)
+}
+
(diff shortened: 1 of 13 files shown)


hcgatewood:
Closing to clear queue, since this is an explanatory PR. After some discussion, we're still leaning toward securing sb iface at the proxy. But will refer to this PR when we get to implementing (for v1.6).

PR #5303, opened 2021-03-04
[Bug_Fix] [MME_Crash] [Mobile_Reachability_Timer] Fixed MME crash during sanity while handling mobile reachability timer expiry

VinashakAnkitAman:
## Title
[Bug_Fix] [MME_Crash] [Mobile_Reachability_Timer] Fixed MME crash during sanity while handling mobile reachability timer expiry

## Summary
This PR is a fix for the flaky test case test_no_auth_response.py as part of issue #5053. While working with the PR #4798, the test case test_no_auth_response.py was always failing during sanity (make integ_test). 

The issue was observed with the mobile reachability timer for the test case s1aptests/test_idle_mode_with_mme_restart.py. During the MME restart for this test case, the mobile reachability timer id was not getting updated properly and because of this after MME restart, when it was handling service request, the stopping of timer was failing. The timer was still running with a wrong timer id and with the expiry of timer, MME was not able to find the context and getting crashed. 

This crash was always happening while running test case test_no_auth_response.py as the timer was expiring by that time, leaving the doubt over this test case while the issue was with the test case s1aptests/test_idle_mode_with_mme_restart.py. This PR fixes the issue.

This PR is re-created for the earlier PR #5227  to re-verify the failing CI.

## Test Plan
Verified with Sanity and individual test case executions
Verified with the changes of PR #4798
Verified with these non-sanity test cases: (Taken from comment by Shruti)
    s1aptests/test_ics_timer_expiry_with_mme_restart.py
    s1aptests/test_implicit_detach_timer_with_mme_restart.py
    s1aptests/test_mobile_reachability_timer_with_mme_restart.py

Signed-off-by: VinashakAnkitAman <ankit.aman@radisys.com>

PR #4765, opened 2021-02-04
[Converged-MME] Timers issue resolution.

Vyas20:
[Converged-MME][LTE] Timer 

## Summary

Started work on replacing timers in the existing MAGMA MME. This PR contains timer replacement in initial_context_setup_rsp_timer_expiry.

## Test Plan
Testing with S1-AP Tester and dsTester.

PR #5463, opened 2021-03-12
[AGW][MME] V1.4 ICS timer fix

ssanadhya:
## Summary

Backporting #5303 fails the `build_oai` target as the ICS timer fix from #4765 is not available on that branch. This change is not needed for master, but needs to be backported to v1.4 

## Test Plan

`make build_oai`


ssanadhya:
Not needed as relevant changes are in #5464

PR #5464, opened 2021-03-12
[AGW] [MME][v1.4] Fix ICS timer argument to mme_app_resume_timers

ssanadhya:

## Summary

Backporting #5303 fails the `build_oai` target as the ICS timer fix from #4765 is not available on that branch. This change is not needed for master, but needs to be backported to v1.4 

## Test Plan

`make build_oai`


ulaskozat (approved):
LGTM

PR #5468, opened 2021-03-12
Add namespace in kubectl commands

wallyrb:

## Summary

1. Adding namespace "orc8r" in kubectl commands for master documentation
2. In FAQ, fixing misspelling in the kubectl commands

## Test Plan

tested locally with create_docusaurus_website.sh

### DIFF
--- a/docs/readmes/faq/faq_magma.md
+++ b/docs/readmes/faq/faq_magma.md
@@ -84,9 +84,9 @@ This section lists some of the commonly asked questions related to Magma operati
 
 ## Orchestrator
 ### How can I check production pods in Orchestrator running on Kubernetes?
-  - **For all pods:** `kubectl -n magma get pods`.
-  - **Logs of a particular pod:** `kubectl -n magma logs -f <CONTROLLER PODNAME>`.
-  - **NOTE:** In above commands, please use the actual namespace after `-n` you have chosen to install orc8r in.
+  - **For all pods:** `kubectl --namespace orc8r get pods`.
+  - **Logs of a particular pod:** `kubectl --namespace orc8r logs -f <CONTROLLER PODNAME>`.
+  - **NOTE:** In above commands, please use the actual namespace after `--namespace` you have chosen to install orc8r in.
 
 ### How can I test Orchestrator connection with Gateway?
   - Login to Gateway, then run `sudo checkin_cli.py`.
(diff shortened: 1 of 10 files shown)


docs/readmes/faq/faq_magma.md
```diff
@@ -102,9 +102,9 @@ This section lists some of the commonly asked questions related to Magma operati
...
 
 ### How can I check the services running in Orchestrator?
-  - List the running pods with `kubectl -norc8r get pods`
+  - List the running pods with `kubectl -n orc8r get pods`
   - Grab the name of orc8r-controller pods, they are in the format `orc8r-controller-xxxxxxxxxx-yyyyy`
-  - Collect the state of services on each pod: `kubectl -norc8r exec orc8r-controller-xxxxxxxxxx-yyyyy  bash -- supervisorctl status`
+  - Collect the state of services on each pod: `kubectl -n orc8r exec orc8r-controller-xxxxxxxxxx-yyyyy  bash -- supervisorctl status`
 
 ### How to verify which services are running in Orc8r-controller?
   - **Get the controller pods:** `kubectl -n magma get pods`.
```
hcgatewood: Can we also fix this and the one directly below -- default namespace is orc8r not Magma. Also move the flag so that it's directly after `kubectl` -- so `kubectl -n orc8r exec`

hcgatewood (changes requested):
Hey Wally thanks for putting this up. Couple requested changes

- There are a couple places, on top of the ones I mentioned, where we're still using `magma` as the namespace. The default is actually `orc8r`, so let's change those to `orc8r`
- Some of the existing commands do include the namespace, but it's not the first flag. Let's move all these to `kubectl --namespace orc8r <rest of command>`
- On top of these previous points: rather than using `-n orc8r`, let's use `--namespace orc8r` since it's clearer what that means. So for everywhere in the docs that we currently use `kubectl -n orc8r`, let's move it to `kubectl --namespace orc8r`

Thanks Wally!

karthiksubraveti (approved):

wallyrb:
> Hey Wally thanks for putting this up. Couple requested changes
> 
> * There are a couple places, on top of the ones I mentioned, where we're still using `magma` as the namespace. The default is actually `orc8r`, so let's change those to `orc8r`
> * Some of the existing commands do include the namespace, but it's not the first flag. Let's move all these to `kubectl --namespace orc8r <rest of command>`
> * On top of these previous points: rather than using `-n orc8r`, let's use `--namespace orc8r` since it's clearer what that means. So for everywhere in the docs that we currently use `kubectl -n orc8r`, let's move it to `kubectl --namespace orc8r`
> 
> Thanks Wally!

Thanks for the comments @hcgatewood , I have updated the PR, let me know if you have any other suggestions. Thanks!

hcgatewood (approved):
Fantastic, thanks so much Wally

PR #5480, opened 2021-03-13
[NMS] Fix the sctpd alert

karthiksubraveti:
## Summary

Unexpected service restart metric doesn't get triggered if a service restarts too quickly. Changing the alert around service restart status as a workaround. 

## Additional Information


andreilee (changes requested):
Would this be better fixed in the implementation of `unexpected_service_restarts`? Are we certain that all service restarts should trigger this alert?

karthiksubraveti:
> Would this be better fixed in the implementation of `unexpected_service_restarts`? Are we certain that all service restarts should trigger this alert?

You are right. This would certainly be better if we fixed the underlying issue. I will discard this PR.

PR #5494, opened 2021-03-15
AGW: ubuntu: Fix linter checks.

pshelar:
1. redis: add default arg for containers get()
2. Fix import for Collections.

## Additional Information

### DIFF
--- a/orc8r/gateway/python/magma/common/redis/containers.py
+++ b/orc8r/gateway/python/magma/common/redis/containers.py
@@ -313,14 +313,14 @@ class RedisFlatDict(MutableMapping[str, T]):
             raise KeyError(composite_key)
         return deleted_count
 
-    def get(self, key: str) -> Optional[T]:
+    def get(self, key: str, default=None) -> Optional[T]:
         """Get ``d[key:type]`` from dictionary.
         Returns None if *key:type* is not in the map
         """
         try:
             return self.__getitem__(key)
         except (KeyError, ValueError):
-            return None
+            return default
 
     def clear(self) -> None:
         """
--- a/orc8r/gateway/python/magma/magmad/gateway_status.py
+++ b/orc8r/gateway/python/magma/magmad/gateway_status.py
@@ -19,7 +19,7 @@ import platform
 import time
 import netifaces
 from typing import NamedTuple, List, Any, Dict, Optional, Tuple
-from collections import KeysView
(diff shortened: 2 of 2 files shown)


hcgatewood (changes requested):
I'm not a fan of disabling linter checks. If it's a straightforward fix, let's go ahead and resolve the actual problem. If not, please also include a description of *why* we're disabling this check

Doing the latter will allow other developers to understand the context and intelligently decide whether to leave the skip or fix the underlying issue

pshelar:
> I'm not a fan of disabling linter checks. If it's a straightforward fix, let's go ahead and resolve the actual problem. If not, please also include a description of _why_ we're disabling this check
> 
> Doing the latter will allow other developers to understand the context and intelligently decide whether to leave the skip or fix the underlying issue

both issues are fixed.

orc8r/gateway/python/magma/common/redis/containers.py
```diff
@@ -313,14 +313,14 @@ def __delitem__(self, key: str) -> int:
             raise KeyError(composite_key)
         return deleted_count
 
-    def get(self, key: str) -> Optional[T]:
+    def get(self, key: str, default=None) -> Optional[T]:
         """Get ``d[key:type]`` from dictionary.
         Returns None if *key:type* is not in the map
```
hcgatewood, line 318: Fix this comment -- actually returns `default` if not found

orc8r/gateway/python/magma/common/redis/containers.py
```diff
@@ -313,14 +313,14 @@ def __delitem__(self, key: str) -> int:
...
         """
         try:
             return self.__getitem__(key)
         except (KeyError, ValueError):
-            return None
+            return default
```
hcgatewood, line 323: This is changing the contract of this function. As long as it's passing CI it should be fine, but considering AGW's low test coverage it may be worth double-checking usages and whether any of them are explicitly checking for KeyError etc

hcgatewood (approved):
Lgtm, thanks for fixing!

orc8r/gateway/python/magma/common/redis/containers.py
pshelar: How is it changing contract? This is adding optional parameter with same return value when arg is not specified.

PR #5495, opened 2021-03-15
AGW: pipelined: uplink-br0: avoid resetting SGi ip address

pshelar:
On service restart check if SGi port is part of uplink-br0
before resetting interface IP address.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/tests/test_uplink_bridge.py
+++ b/lte/gateway/python/magma/pipelined/tests/test_uplink_bridge.py
@@ -298,6 +298,8 @@ class UplinkBridgeWithNonNATTestVlan(unittest.TestCase):
                                      include_stats=False)
 
 
+@unittest.skip
+# this reset default GW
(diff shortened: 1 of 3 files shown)


lte/gateway/python/magma/pipelined/tests/test_uplink_bridge.py
```diff
@@ -298,6 +298,8 @@ def testFlowSnapshotMatch(self):
                                      include_stats=False)
 
 
+@unittest.skip
+# this reset default GW
```
koolzz, line 302: Why are we skipping these?

koolzz (approved):

lte/gateway/python/magma/pipelined/tests/test_uplink_bridge.py
pshelar: this resets default route in the VM and mess up connectivity.

pshelar: I need to update the test to mock the route updates.

PR #5497, opened 2021-03-15
AGW: Ubuntu upgrade proposal.

pshelar:
Attached is a proposal for transitioning to Ubuntu for the AGW deployments.

## Additional Information

### DIFF
--- a/docs/readmes/proposals/p013_Ubuntu_upgrade.md
+++ b/docs/readmes/proposals/p013_Ubuntu_upgrade.md
@@ -1,0 +1,91 @@
+---
+id: p013_Ubuntu_upgrade
+title: AGW Ubuntu upgrade
+hide_title: true
+---
+
+# Overview
+
+*Status: Accepted*\
+*Author: @pshelar*\
+*Last Updated: 03/15*\
+*Targeted Release: 1.5*\
+*Feedback requested from: @arunuke*
+
+This document cover AGW ubuntu upgrade path
+
+## Goal
+Add support for Ubuntu based AGW.
+
+## Introduction
+
+There are multiple reasons to update base OS from current Debian to ubuntu distribution
+
+1. Python 3.5 is EOL on September 13th, 2020. Current Debian does not support the latest python release. pip 21.0
+   dropped support for Python 3.5 in January 2021.
+2. Newer kernel is required for enabling next generation datapath performance.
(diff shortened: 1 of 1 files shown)


docs/readmes/proposals/p0010_Ubuntu_upgrade.md
```diff
@@ -0,0 +1,89 @@
...
+in smoother AGW updates for future releases.
+
+### *Debian based AGW 1.4 to Debian based AGW 1.5:*
+
+Current production deployment can use current Debian upgrade path for upgrading magma from 1.4 to 1.5. Existing workflow
+remains same.
+
+### *Debian based AGW 1.4 to Ubuntu based AGW 1.5:*
+
+This deployment involves two steps first, upgrading underlying OS, second step installing AGW packages.
```
arunuke: Would this be an upgrade of the OS or would it be a reinstall?

docs/readmes/proposals/p0010_Ubuntu_upgrade.md
```diff
@@ -0,0 +1,89 @@
...
+This type of deployment results in downtime.
+
+1. Bring down AGW services
+2. Update base OS on AGW from Debian to ubuntu
+3. Deploy AGW 1.5
+4. Configure AGW
+5. Start AGW services
+
+## Major changes for Ubuntu upgrade.
+1. Update Python based services to Python 3.8
```
arunuke: If an SI opts to use a commercially supported version of Ubuntu, will the Debian-to-Ubuntu be considered a supported host? Why would we not prefer a re-install for a clean slate?

arunuke (changes requested):

docs/readmes/proposals/p0010_Ubuntu_upgrade.md
pshelar: This is upgrade of OS.

docs/readmes/proposals/p0010_Ubuntu_upgrade.md
pshelar: I dont see much difference between upgrade of ubuntu from debian vs clean install of ubuntu once it is done.
operator can face issues when moving from debian -> ubuntu. so from that point of view. yes, starting with clean slate is preferred.
once the transition is done it would not matter.

arunuke (approved):
LGTM !!

PR #5518, opened 2021-03-16
correct links to images in feg setup guide

drbruced12:
Signed-off-by: Bruce Davie <3101026+drbruced12@users.noreply.github.com>

### DIFF
--- a/docs/readmes/feg/federated_FWA_setup_guide.md
+++ b/docs/readmes/feg/federated_FWA_setup_guide.md
@@ -23,13 +23,13 @@ When configuring an integration with LTE nodes, it is necessary to link these tw
 
 In the **Federated LTE** **Network**’s NMS page, the Federation config should be the **Federation** **Network**’s network ID.
 
-![NMS-FederatedLTE-Config.png](../assets/feg/NMS-FederatedLTE-Config.png)
+![NMS-FederatedLTE-Config.png](assets/feg/NMS-FederatedLTE-Config.png)
 
 ### Associating Federation network to a FederatedLTE network
 
 In order to complete the association, we also need to modify the **Federation Network**‘s federation configuration.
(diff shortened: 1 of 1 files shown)


uri200:
Hello @drbruced12,

I can seee the pictures with current path (`../assets`).

Can you take a look at the link below? 

https://github.com/magma/magma/blob/master/docs/readmes/feg/federated_FWA_setup_guide.md

This is how you should see it on master

image

drbruced12:
I agree, the ../assets approach works fine if you only want to read the docs on GitHub. However, when you build the docs using docusaurus, it breaks. So, for example, if you look at
https://docs.magmacore.org/docs/next/feg/federated_fwa_setup_guide

You can see that the images are missing. Conversely, a page like
https://docs.magmacore.org/docs/next/orc8r/architecture_overview

has images correctly displayed in the built version of the docs, but the images are missing when you look at it on GitHub.

Federated_fwa_setup_guide is the odd man out among all the docs. Hence my proposed change.

If you want to be able to see images both in the docusaurus site and on GitHub, I guess we need a more comprehensive review of how the images are being placed into a subdirectory. But I assume the primary goal is to have the docusaurus built version look correct.

Bruce

uri200:
Hey @drbruced12, you are right! Even at docusaurus repo they are not able to show that 

https://github.com/facebook/docusaurus/blob/ac28dc70917e2746b5229b1717f21dbe0acdba6e/website/docs/using-plugins.md

https://github.com/facebook/docusaurus/blob/311be6e71ae181178eb9dae453970d6f2e4dbd14/packages/docusaurus-mdx-

loader/src/remark/transformImage/__tests__/fixtures/img.md


 Good catch!

uri200 (approved):

PR #5526, opened 2021-03-16
[agw][new feature] Inbound roaming support

### LINKED ISSUE #4916: [Inbound Roaming][MME] s6a_task to support both SubscriberDB and FEG
For inbound roaming + SubscriberDB based model, we need to send the authentication request to different places based on the routing configuration.
For this, we need to initialize both clients on startup.

End goal:
- Add a new configuration flag to indicate the setup needs both FeG and SubscriberDB clients. (Can extend/modify the existing relay flag)
- Implementation in S6A task to use the configuration to initialize the clients. 
- Once the routing configuration is done, we will use it to switch between the two.

pruthvihebbani:
[agw][new feature] Inbound roaming support

## Summary

1. Added support to parse mme.conf file and store the configuration into mode_map_t in mme_config.h. This will be used by s6a and s8 tasks to route the message to appropriate node/interface
2. Added routing logic at s6a task which would route s6a messages tp wither FeG or subscriberdb based on the PLMN extracted from IMSI

## Test Plan

- Verified sanity
- Verified that s6a messages for UE with 001010 PLMN and mode set to LOCAL_SUBSCRIBER is routed to subscriberdb by executing test_attach_detach TC
- For UEs with PLMN 123456 and mode set to S8_SUBSCRIBER or SPGW_SUBSCRIBER, since FeG is not available in local setup verified that FeG grpc channel is created. Need to test this path on teravm setup


lte/gateway/c/oai/common/common_defs.h
```diff
@@ -100,6 +100,13 @@ typedef enum {
   TLV_FATAL_ERROR = TLV_VALUE_DOESNT_MATCH
 
 } error_code_e;
+
+typedef enum {
+  SPGW_SUBSCRIBER  = 0,  // default mode is HSS + spgw_task
```
ulaskozat, line 108: to be consistent `federated HSS`

lte/gateway/c/oai/common/common_defs.h
```diff
@@ -100,6 +100,13 @@ typedef enum {
...
+  LOCAL_SUBSCRIBER = 1,  // will use subscriberDb + policydb
+  S8_SUBSCRIBER    = 2,  // will use HSS (feg) + s8_task
```
ulaskozat: `federated HSS`

lte/gateway/c/oai/common/common_defs.h
```diff
@@ -100,6 +100,13 @@ typedef enum {
   TLV_FATAL_ERROR = TLV_VALUE_DOESNT_MATCH
 
 } error_code_e;
+
+typedef enum {
+  SPGW_SUBSCRIBER  = 0,  // default mode is HSS + spgw_task
+  LOCAL_SUBSCRIBER = 1,  // will use subscriberDb + policydb
```
ulaskozat: spgw_task rather than policydb?

lte/gateway/c/oai/lib/s6a_proxy/S6aClient.cpp
```diff
@@ -88,17 +90,64 @@ static bool read_mme_cloud_subscriberdb_enabled(void) {
...
+
+// Extract MCC and MNC from the imsi received and match with
+// configuration
+bool match_fed_mode_map(const char* imsi) {
+  uint8_t mcc_d1 = imsi[0] - '0';
+  uint8_t mcc_d2 = imsi[1] - '0';
+  uint8_t mcc_d3 = imsi[2] - '0';
+  uint8_t mnc_d1 = imsi[3] - '0';
+  uint8_t mnc_d2 = imsi[4] - '0';
+  uint8_t mnc_d3 = imsi[5] - '0';
```
ulaskozat, line 111: verify the range of digits.

lte/gateway/c/oai/common/common_defs.h
uri200: Not sure if those definitions are only for `c` but for `cpp` we could use values from protos. We should take advantage of the ENUM on `mconfigs.pb.h`. Otherwise if we ever change the proto values it will not match anymore. 

this is in `mconfigs.pb.h`.
```
enum ModeMapItem_FederatedMode {
  ModeMapItem_FederatedMode_SPGW_SUBSCRIBER  = 0,
  ModeMapItem_FederatedMode_LOCAL_SUBSCRIBER = 1,
  ModeMapItem_FederatedMode_S8_SUBSCRIBER    = 2,
  ModeMapItem_FederatedMode_ModeMapItem_FederatedMode_INT_MIN_SENTINEL_DO_NOT_USE_ =
      ::google::protobuf::kint32min,
  ModeMapItem_FederatedMode_ModeMapItem_FederatedMode_INT_MAX_SENTINEL_DO_NOT_USE_ =
      ::google::protobuf::kint32max
};
```


you can use it this way


```
#include "lte/protos/mconfig/mconfigs.pb.h"


 if (any_int_variable == magma::mconfig::ModeMapItem_FederatedMode_S8_SUBSCRIBER)
```

lte/gateway/c/oai/lib/s6a_proxy/S6aClient.cpp
```diff
@@ -88,17 +90,64 @@ static bool read_mme_cloud_subscriberdb_enabled(void) {
...
+         (mnc_d1 == mme_config.mode_map_config.mode_map[itr].plmn.mnc_digit1) &&
+         (mnc_d2 ==
+          mme_config.mode_map_config.mode_map[itr].plmn.mnc_digit2))) {
+      if (mme_config.mode_map_config.mode_map[itr].plmn.mnc_digit3 != 0xf) {
+        if (mnc_d3 !=
+            mme_config.mode_map_config.mode_map[itr].plmn.mnc_digit3) {
+          continue;
+        }
+      }
+      if ((mme_config.mode_map_config.mode_map[itr].mode == SPGW_SUBSCRIBER) ||
```
uri200:
```
 if (mme_config.mode_map_config.mode_map[itr].mode == magma::mconfig::ModeMapItem_FederatedMode_SPGW_SUBSCRIBER)
 ```

File: lte/gateway/c/oai/lib/s6a_proxy/S6aClient.cpp [hunk 6]
```diff
@@ -88,17 +90,64 @@ static bool read_mme_cloud_subscriberdb_enabled(void) {
...
-  // Create channel based on relay_enabled and cloud_subscriberdb_enabled
-  // flags. If relay_enabled is true, then create a channel towards the FeG.
+S6aClient& S6aClient::get_subdb_instance(bool enable_s6a_proxy_channel) {
+  static S6aClient subdb_instance(enable_s6a_proxy_channel);
+  return subdb_instance;
+}
+
+// Extract MCC and MNC from the imsi received and match with
+// configuration
+bool match_fed_mode_map(const char* imsi) {
```
uri200: wouldn't be easier to convert the imsi[0]:imsi[5] to a string and then compare that string to the string on the configuration? That conversion shouldn't cause much overhead to cpp

pruthvihebbani: AGW uses MCC and MNC as decimal digits which is the 3gpp specified format. After parsing configuration,  MCC and MNC are converted into decimal digits format. Hence this conversion is needed while extracting MCC and MNC from IMSI

uri200: Understood. 
I know this may not be possible since there is a lot of code in there, but in the future we may want to strip the 3gpp format and convert those values into ints or string so we can manage them properly in the code.

lte/gateway/c/oai/common/common_defs.h
```diff
@@ -100,6 +100,16 @@ typedef enum {
   TLV_FATAL_ERROR = TLV_VALUE_DOESNT_MATCH
 
 } error_code_e;
+
+/* This enum should match with the ModeMapItem_FederatedMode enum
+ * defined in mconfigs.pb.h
```
ssanadhya: As `mconfigs.pb.h` is generated from `mconfigs.proto`, better to mention that as reference in the code

lte/gateway/c/oai/lib/s6a_proxy/S6aClient.cpp
```diff
@@ -88,17 +90,63 @@ static bool read_mme_cloud_subscriberdb_enabled(void) {
   return mconfig.cloud_subscriberdb_enabled();
 }
 
-S6aClient& S6aClient::get_instance() {
-  static S6aClient client_instance;
-  return client_instance;
+S6aClient& S6aClient::get_s6a_proxy_instance(bool enable_s6a_proxy_channel) {
+  static S6aClient s6a_proxy_instance(enable_s6a_proxy_channel);
+  return s6a_proxy_instance;
```
ssanadhya: Why do we need `enable_s6a_proxy_channel` as a boolean argument here. Isn't it always true for s6a_proxy_instance?
```suggestion
S6aClient& S6aClient::get_s6a_proxy_instance() {
  static S6aClient s6a_proxy_instance(true);
  return s6a_proxy_instance;
```

lte/gateway/c/oai/lib/s6a_proxy/S6aClient.cpp
```diff
@@ -88,17 +90,63 @@ static bool read_mme_cloud_subscriberdb_enabled(void) {
...
 }
 
-S6aClient::S6aClient() {
-  // Create channel based on relay_enabled and cloud_subscriberdb_enabled
-  // flags. If relay_enabled is true, then create a channel towards the FeG.
+S6aClient& S6aClient::get_subdb_instance(bool enable_s6a_proxy_channel) {
+  static S6aClient subdb_instance(enable_s6a_proxy_channel);
+  return subdb_instance;
```
ssanadhya: Why do we need `enable_s6a_proxy_channel` as a boolean argument here. Isn't it always false for subdb_instance?
```suggestion
S6aClient& S6aClient::get_subdb_instance() {
  static S6aClient subdb_instance(false);
  return subdb_instance;
```

ssanadhya (changes requested):

ssanadhya (approved):

PR #5529, opened 2021-03-16
[CI] Enable c/c++ codecov support

### LINKED ISSUE #5043: [CI] Re-enable code coverage for c/c++, integrate with CI / PR
There was a time when it appears code coverage reporting was in place - as evident in the Makefiles of e.g. [lte/gateway](https://github.com/magma/magma/blob/v1.4/lte/gateway/Makefile#L195-L224) (thanks to @hcgatewood for what exists - which will make it easier to re-enable).

The key challenge today is that the AGW is not built on CircleCI machines, nor is it tested there (even unit tests, with exception of the Python tests).  Instead the job of build and test is pushed to remote machines using a [fab call](https://github.com/magma/magma/blob/v1.4/.circleci/config.yml#L285). So the challenge will be to update the fab call to collect coverage and then exfiltrate it to e.g. CodeCov.io (presently what we are migrating to).

electronjoe:
## Summary

Using the Docker container built for GCC Annotations in GH (#5404), build and run the following tests, then compute code coverage and push to codecov.io.

- test_oai
- test_session_manager

## Test Plan

See [Code Coverage Output](https://codecov.io/gh/magma/magma/tree/37ac447f941ee23df5b8c78bc5384d72b14a0549/lte) for this PR.

### DIFF
--- a/lte/gateway/Makefile
+++ b/lte/gateway/Makefile
@@ -179,8 +179,16 @@ test_oai: build_common ## Run all OAI-specific tests
 
 # Catch all for c service tests
 # This works with test_dpi and test_session_manager
-test_%: stop build_common
+test_%: build_common
 	$(call run_ctest, $(C_BUILD)/$*, $(GATEWAY_C_DIR)/$*, )
+
+coverage_oai: test_oai
+	lcov --capture --directory $(C_BUILD) --output-file /tmp/coverage_oai.info
+	rm -f `find $(C_BUILD) -name *.gcda` # Clean up any prior coverage data
+
+coverage_session_manager: test_session_manager
+	lcov --capture --directory $(C_BUILD) --output-file /tmp/coverage_session_manager.info
+	rm -f `find $(C_BUILD) -name *.gcda` # Clean up any prior coverage data
 
 # format and test c/session_manager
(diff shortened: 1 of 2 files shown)


lte/gateway/Makefile
```diff
@@ -179,66 +179,28 @@ test_oai: build_common ## Run all OAI-specific tests
 
 # Catch all for c service tests
 # This works with test_dpi and test_session_manager
-test_%: stop build_common
+test_%: build_common
```
electronjoe, line 182: Unclear to me what the purpose of the stop clauses was in our test environment. Can I remove? The `stop` and related makefile recipes use `sudo` which is not present in the docker environment.

pshelar: This done to avoid issues running pipelineD UT while pipelineD service is running.
so this needs to be done manually after this change.

pshelar (approved):
LGTM

tmdzk (approved):

electronjoe:
This PR resolved #5043

electronjoe:
Nice example of operation!
Screen Shot 2021-03-26 at 9 03 12 AM

PR #5534, opened 2021-03-16
[SessionD] Modify SessionD unit tests for better stability

themarwhal:

## Summary
The wait_for_setup... function relies on progressing the event loop a few times to make sure the initial setup calls are processed. On failure cases (especially in @electronjoe 's dockerized version), we've seen that this produces an unexpected expectation over-saturation failure as LocalEnforcer calls setup again. 

This is a minor change to expect at least one setup call for failure cases, since we can't guarantee how many.

#thanks @electronjoe for finding this!

## Test Plan
make precommit_sm

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/test/test_session_manager_handler.cpp
+++ b/lte/gateway/c/session_manager/test/test_session_manager_handler.cpp
@@ -101,9 +101,7 @@ class SessionManagerHandlerTest : public ::testing::Test {
     rule_store->insert_rule(rule);
   }
 
-  // This function should always be called at the beginning of the test to
-  // prevent unexpected SessionD <-> PipelineD syncing logic mid-test.
-  void send_empty_table_and_wait_for_setup(SetupFlowsResult_Result res) {
+  void send_empty_table() {
     RuleRecordTable empty_table;
     // epoch indicates the last PipelineD service start time
     empty_table.set_epoch(DEFAULT_PIPELINED_EPOCH);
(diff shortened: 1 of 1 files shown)


electronjoe (approved):

lte/gateway/c/session_manager/test/test_session_manager_handler.cpp
```diff
@@ -101,22 +101,26 @@ class SessionManagerHandlerTest : public ::testing::Test {
...
-  void send_empty_table_and_wait_for_setup(SetupFlowsResult_Result res) {
+  void send_empty_table() {
     RuleRecordTable empty_table;
     // epoch indicates the last PipelineD service start time
     empty_table.set_epoch(DEFAULT_PIPELINED_EPOCH);
     grpc::ServerContext context;
     session_manager->ReportRuleStats(
         &context, &empty_table,
         [this](grpc::Status status, Void response_out) {});
+  }
```
uri200, line 112: wouldnt make more sense to add the 

```
    evb->loopOnce();
    evb->loopOnce();
```
to this function `send_empty_table`?
I think the action that requires some event loops to happen is the `ReportRuleStats`

uri200 (approved):

lte/gateway/c/session_manager/test/test_session_manager_handler.cpp
themarwhal: The looping has to be done after the setup expectation. it is a bit ugly :/

PR #5542, opened 2021-03-16
[MME][Bug] On MME shutdown, stop GRPC service before releasing ZMQ ctx

talkhasib:

## Summary

When MME shutsdown, we destroy the grpc service ZMQ context before we stop the GRPC server.
This occasionally leads to processing a GRPC message while the ZMQ context is destroyed. Thus sending a ZMQ msg to NULL socket. There is an assert that catches the issue.

## Test Plan

Ran few integ tests


ulaskozat (approved):
LGTM

ulaskozat:
MAGMA-OAI-TESTS

ardzoht (approved):
lgtm

PR #5545, opened 2021-03-16
[AGW] Install script using shell in bash

tmdzk:

[AGW] Install script using shell in bash

## Summary

Using shell in bash cleaned up

## Test Plan
/shrug


pshelar (approved):
LGTM

PR #5551, opened 2021-03-17
[lte][agw] Datapath Actions Script

saurabhm3hra:
CLI wrapper around ovs-appctl ofproto/trace to display the Datapath actions of the supplied IMSI.
This will help to easily trace the packets coming in to the UE.
- If the packets are dropped, Datapath Actions will print drop
- If successful, Datapath Actions will print the tunnel id.

## Test Plan
Tested on AGW
image

### DIFF
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
@@ -171,7 +171,7 @@ variable "orc8r_deployment_type" {
 variable "orc8r_chart_version" {
   description = "Version of the core orchestrator Helm chart to install."
   type        = string
-  default     = "1.5.17"
+  default     = "1.5.18"
 }
 
 variable "cwf_orc8r_chart_version" {
--- a/orc8r/cloud/helm/orc8r/Chart.lock
+++ b/orc8r/cloud/helm/orc8r/Chart.lock
@@ -7,12 +7,12 @@ dependencies:
   version: 1.4.22
 - name: nms
   repository: ""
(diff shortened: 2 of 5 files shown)


lte/gateway/python/scripts/dp_actions_cli.py
```diff
@@ -0,0 +1,92 @@
...
+
+def output_datapath_actions(ue_ip: str):
+    """
+    Returns the Output of Datapath Actions based as per the supplied UE IP
+    """
+    service = MagmaService('pipelined', mconfigs_pb2.PipelineD())
+    if service.mconfig.nat_enabled:
+        in_port = 'local'
+    else:
+        in_port = '2'
```
pshelar: Lets use OVSDB to find port number, `sudo ovsdb-client dump Interface name ofport |grep patch-up`

lte/gateway/python/scripts/dp_actions_cli.py
```diff
@@ -0,0 +1,92 @@
...
+
+    cmd = ['sudo',
+    'ovs-appctl',
+    'ofproto/trace',
+    'gtp_br0',
+    'tcp,in_port='+ in_port +',ip_dst='+ ue_ip +',ip_src=8.8.8.8,tcp_src=3372,tcp_dst=80',
+    ]
```
pshelar: Can you also provide command line options for dst IP and dst port and protocol (tcp/udp/icmp)  so that user can check specific target.

pshelar (changes requested):
Thanks for working on this. This looks pretty close to complete.

PR #5062, opened 2021-02-20
[mme] Fixup missing includes in various headers

electronjoe:
## Summary

In attempting to build the MME by use of our docker container at `magma/lte/gateway/docker/mme/Dockerfile.ubuntu18.04`, I repeatedly ran into build failures due to missing header includes.  After some iteration, the following collection of fixups were necessary to achieve healthy build.

Nearly all were clearly missing headers and made sense.  The exception was unexplainable build failures of bytes_to_ie.c that only resolved after removal of commented-out header includes. I cannot explain that change, but on the other hand it's good cruft to remove regardless.

The build failures and missing includes further motivate Issue #4868.

## Test Plan

The following Docker Build now succeeds (where it did not before):

```shell
cd magma/lte/gateway/docker/mme
docker build -t magma_mme -f Dockerfile.ubuntu18.04 ../../../../
```

Signed-off-by: Scott Harrison Moeller <smoeller@fb.com>

PR #5562, opened 2021-03-17
[session_manager] Add missing headers and linking for GPR_ASSERT

electronjoe:
# Summary

GPR_ASSERT technically requires some header includes that were missing. Further, in my new docker developer environment, we failed linking for this target due to lack of link with -lgpr.

Working goes toward #5529.

### DIFF
--- a/lte/gateway/c/session_manager/SessionManagerServer.cpp
+++ b/lte/gateway/c/session_manager/SessionManagerServer.cpp
@@ -12,6 +12,9 @@
  */
 #include "SessionManagerServer.h"
 #include "magma_logging.h"
+#include <grpc/impl/codegen/port_platform.h>
+#include <stdarg.h>
+#include <stdlib.h>
 #include <chrono>
 #include <ctime>
 using grpc::Status;
--- a/lte/gateway/c/session_manager/CMakeLists.txt
+++ b/lte/gateway/c/session_manager/CMakeLists.txt
@@ -173,7 +173,7 @@ add_library(SESSION_MANAGER
 target_link_libraries(SESSION_MANAGER
   SERVICE303_LIB SERVICE_REGISTRY ASYNC_GRPC CONFIG POLICYDB EVENTD
   DATASTORE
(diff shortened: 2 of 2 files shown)


themarwhal (approved):

lte/gateway/c/session_manager/SessionManagerServer.cpp
```diff
@@ -12,6 +12,9 @@
  */
 #include "SessionManagerServer.h"
 #include "magma_logging.h"
+#include <grpc/impl/codegen/port_platform.h>
```
electronjoe, line 15: I was perhaps lazy on my include sequencing here. I'm not sure what our standards are - if any today - but I plan to fix this with IncludeWhatYouUse soon anyway?

uri200: so maybe this is an obvious question, but why were we able to build without those includes before?

Or is this some kind of issue happening with the new gcc version on docker?

uri200 (approved):

lte/gateway/c/session_manager/SessionManagerServer.cpp
electronjoe: These includes may not have been blockers for this building for me - the linker addition was a blocker.  Why this was linking without -lgpr I am unclear.

The new header includes may not be strictly necessary, we may be historically getting them through indirect means. But these are the headers advised for GPR_ASSERT in [their documentation](https://grpc.github.io/grpc/cpp/impl_2codegen_2log_8h.html).

We need to scrub all our includes with IncludeWhatYouUse (#5062) as today we rely on luck for many of our header imports (the ordering of compilation and header processing - which has no reason to remain stable across upgrades to CMake / GCC / Clang).

uri200: Thanks. I think order of compilation is helping. That may be why we have so many issues when we use IDE, but build still works

PR #5575, opened 2021-03-17
[SessionD] Activate all rules with PolicyRule definitions

### LINKED ISSUE #4090: [pipelined]Move to flat rate enforcement stats reporting

### LINKED ISSUE #4092: [Sessiond]Move to absolute value stats reporting

### **Definitions**

### Proposed Changes

### Timeline

### Proposed Changes

```
struct StatsPerPolicy {
   uint32_t current_version;
}
// PolicyID -> StatsPerPolicy
```

        `

### Timeline

### Proposed Changes

```
struct Usage {
    uint32_t rule_version;
};
struct RuleStats {
    uint64_t tx;
    uint64_t rx;
    uint64_t dropped_tx;
    uint64_t dropped_rx;
};
```

    1. Action: Do nothing
    2. Meaning:

themarwhal:
## Summary
The big change is to migrate all Setup / Activate flows calls to send policies by definition, not by name. This will reduce the need to look up ID->Policy definition in Redis on the PipelineD side.

In order to make this happen, I cleaned up `PipelinedClient`'s interface to only accept a vector of PolicyRules. (A combination of both static and dynamic rules). 

## Test Plan
CWF integ test: https://app.circleci.com/pipelines/github/magma/magma/17025/workflows/bcbd2d95-34dc-4b18-a7b3-ff944984ae8b/jobs/160111
LTE integ test: https://app.circleci.com/pipelines/github/magma/magma/17025/workflows/37d51a34-18f0-4a37-b725-4f1793fcbcc1/jobs/160110
make precommit_sm 

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -563,12 +563,11 @@ void LocalEnforcer::install_final_unit_action_flows(
 
 void LocalEnforcer::cancel_final_unit_action(
     const std::unique_ptr<SessionState>& session,
-    const std::vector<std::string>& restrict_rules,
+    std::vector<PolicyRule> gy_rules_to_deactivate,
     SessionStateUpdateCriteria& uc) {
   SessionState::SessionInfo info;
   session->get_session_info(info);
 
-  std::vector<PolicyRule> gy_rules_to_deactivate;
   for (const auto& rule : info.gy_dynamic_rules) {
     PolicyRule dy_rule;
     bool is_dynamic = session->remove_gy_dynamic_rule(rule.id(), &dy_rule, uc);
@@ -577,9 +576,9 @@ void LocalEnforcer::cancel_final_unit_action(
     }
   }
 
-  if (!gy_rules_to_deactivate.empty() || !restrict_rules.empty()) {
+  if (!gy_rules_to_deactivate.empty()) {
     pipelined_client_->deactivate_flows_for_rules(
-        info.imsi, info.ip_addr, info.ipv6_addr, info.teids, restrict_rules,
+        info.imsi, info.ip_addr, info.ipv6_addr, info.teids,
         gy_rules_to_deactivate, RequestOriginType::GY);
   }
 }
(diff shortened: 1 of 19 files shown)


lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -410,32 +408,17 @@ void LocalEnforcer::handle_force_termination_timeout(
...
-  if (!gy_rules.static_rules.empty() || !gy_rules.dynamic_rules.empty()) {
+      imsi, ip_addr, ipv6_addr, teids, RequestOriginType::GX);
+  if (!session->get_all_final_unit_rules().empty()) {
     pipelined_client_->deactivate_flows_for_rules_for_termination(
-        imsi, ip_addr, ipv6_addr, teids, gy_rules.static_rules,
-        gy_rules.dynamic_rules, RequestOriginType::GY);
+        imsi, ip_addr, ipv6_addr, teids, RequestOriginType::GY);
   }
+
+  session->remove_all_rules_for_termination(uc);
```
uri200, line 421: what if we add a comment here remining our future selfs that sending empty rules on `deactivate_flows_for_rules_for_termination` mean deactivation of all rules for that subscriber (static, dynamic and gy_dynamic)

lte/gateway/c/session_manager/RuleStore.cpp
```diff
@@ -137,6 +137,20 @@ bool PolicyRuleBiMap::get_rule(
   return true;
 }
 
+bool PolicyRuleBiMap::get_rules(
```
uri200: To differentiate between get rule name, and get its definition, wouldn't it be better `get_rules_definition`?  

Or maybe we already have a clear definition between functions that get rule name only and functions that gets its definition

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -856,10 +855,18 @@ void SessionState::get_session_info(SessionState::SessionInfo& info) {
...
+  gy_dynamic_rules_.get_rules(info.gy_dynamic_rules);
   info.static_rules = active_static_rules_;
-  info.ambr         = config_.get_apn_ambr();
+
+  for (const std::string& rule_id : active_static_rules_) {
+    PolicyRule rule;
+    if (static_rules_.get_rule(rule_id, &rule)) {
+      info.gx_static_rules.push_back(rule);
+    }
+  }
```
uri200, line 882: I don't remember all the usages of `get_session_info` but shouln't we keep the concept of static and dynamic on sessiond? and only remove it once we talk to pipelined?

uri200 (commented):
per what I understand, we keep the concept of dynamic and static at sessiond level, but we eliminate it from pipelined?

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: yea good call!

lte/gateway/c/session_manager/RuleStore.cpp
themarwhal: yeah to be extra clear I can rename it to `get_rules_by_ids`

lte/gateway/c/session_manager/SessionState.cpp
themarwhal: SessionInfo is only used in terms of communicating to PipelineD, so I think we're OK there. From a quick glance it is used in AAAClient, but we should migrate that to use `session->get_config` anyways.

themarwhal:
@uri200 Replying to your comment. Yes, we're not eliminating it from SessionD. We're just eliminating it from the calls to PipelineD. (This means modifying some structs and functions that are used to propagate the rule info to PipelineDClient)

uri200 (approved):

lte/protos/pipelined.proto
```diff
@@ -77,6 +77,7 @@ message RequestOriginType {
 message ActivateFlowsRequest {
   SubscriberID sid = 1;
   string ip_addr = 2; // Subscriber session ipv4 address
+  // Deprecated
   repeated string rule_ids = 3; // List of static rules obtained from PCRF
```
koolzz, line 81: should we rm `repeated string rule_ids = 3;` and add?
reserved 3;

koolzz (approved):
awesome stuff

lte/protos/pipelined.proto
themarwhal: isn't it still used in pipelined though?

koolzz: yup good point will change there

Issue #5383, opened 2021-03-09
[orc8r][helm] Refactor orc8r chart to use external-dns


mattymo:
### Describe the Issue

The current approach via terraform assumes precreated secrets. That isn't necessarily a bad thing, but it would be even better if these secrets could be generated by cert-manager, instead of shell scripts.

**Expected behavior**
A cloud-native approach should be possible so that helm install orc8r -f myvalues.yaml will cause cert-manager to create the necessary secrets for the certs that are needed for orc8r.

mattymo:
@ShubhamTatvamasi  I can't assign you, but I'm tagging you here

ShubhamTatvamasi:
> @ShubhamTatvamasi I can't assign you, but I'm tagging you here

Okay, Thanks @mattymo

karthiksubraveti:
@mattymo This sounds like a good issue to take on. 
Adding @arunuke for visibility here. @ShubhamTatvamasi were you interested in working on this. I can assign you this issue if you are interested. LMK.

ShubhamTatvamasi:
Yes, Please @karthiksubraveti. I would like to work on this.

karthiksubraveti:
@mattymo The issue title seems to be different from the description, can you change it accordingly ? Thanks

ShubhamTatvamasi:
@mattymo I have made this helm chart for generating secrets.
This is still a work in progress.
Can you please see if I am moving in the right direction?
https://github.com/ShubhamTatvamasi/magma-certs

karthiksubraveti:
cc @hcgatewood

hcgatewood:
Left some comments in Slack. Looks like we're making good progress. I think this will be a solid win for simplifying deploy+operate complexity

ShubhamTatvamasi:
@mattymo I have a question.

Right now we have 3 issuers for signing certificates:
```
➜  magma-certs git:(master) kubectl get issuers
NAME                            READY   AGE
magma-certs-certifier-issuer    True    40s
magma-certs-root-issuer         True    40s
magma-certs-selfsigned-issuer   True    40s
```

And in total 6 certificates:
```
➜  magma-certs git:(master) kubectl get certificate
NAME                                     READY   SECRET                           AGE
magma-certs-admin-operator-certificate   True    magma-certs-admin-operator-tls   4m57s
magma-certs-certifier-certificate        True    magma-certs-certifier-tls        4m57s
[...]

PR #5584, opened 2021-03-18
[orc8r][helm] Refactor orc8r chart with cert-manager

ShubhamTatvamasi:
## Summary

For Issue: https://github.com/magma/magma/issues/5383
With `certs` helm chart these certificates are auto-generated by cert-manager.

- admin-operator-tls
- certifier-tls
- controller-tls
- bootstrapper-tls
- fluentd-tls
- root-tls
- nms-tls

These new secrets are currently **not** used by other charts.

## Test Plan

Install prerequisite:
```bash
helm repo add jetstack https://charts.jetstack.io
helm repo update

helm install cert-manager jetstack/cert-manager \
  --create-namespace \
  --namespace cert-manager \
  --set installCRDs=true
```

Install certs:
```bash
cd $MAGMA_ROOT/orc8r/cloud/helm/orc8r/charts/certs
helm install certs .
```

## Additional Information


orc8r/cloud/helm/orc8r/values.yaml
```diff
@@ -52,6 +52,34 @@ secret:
...
+  create: false
+
+  dnsDomain: localhost
+
+  # Following durations is set on these certificates:
+  # - nms-certificate
+  # - controller-certificate
+  # - admin-operator-certificate
+  # 10 years = 87600h
+  duration: 87600h
```
hcgatewood: For here and other values.yml file: current default is 1 year, I think sticking to that default is sensible

CODEOWNERS
```diff
@@ -9,6 +9,7 @@ ci-scripts/ @rdefosse @tmdzk @mattymo @119Vik
 orc8r/ @mpgermano @hcgatewood @emakeev
 orc8r/tools/packer/ @tmdzk @mattymo @119Vik
 orc8r/cloud/deploy/bare-metal/ @tmdzk @mattymo @119Vik
+orc8r/cloud/helm/orc8r/charts/certs/ @ShubhamTatvamasi
```
hcgatewood: This change would need to be in a separate PR

Updating the codeowners is equivalent to updating the set of project maintainers. See https://www.magmacore.org/community/ for context on that process

orc8r/cloud/helm/orc8r/charts/certs/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+# Following durations is set on these certificates:
+# - nms-certificate
+# - controller-certificate
+# - admin-operator-certificate
+# 10 years = 87600h
+duration: 87600h
+
+adminOperator:
+  customIssuer:
```
hcgatewood, line 27: Here and for other values.yml: I think it makes sense to default custom issue to let's encrypt -- are there any downsides to this approach?

orc8r/cloud/helm/orc8r/charts/certs/templates/admin-operator-pkcs12-password.yaml
```diff
@@ -0,0 +1,19 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+{{- if .Values.create }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Release.Name }}-admin-operator-pkcs12-password
+data:
+  password: {{ .Values.adminOperator.pkcs12.password | b64enc | quote }}
```
hcgatewood: I think it makes sense to make this var required. Easiest way to make sure everyone uses "password" as their password is to make it the default

orc8r/cloud/helm/orc8r/charts/certs/templates/certifier-certificate.yaml
```diff
@@ -0,0 +1,27 @@
...
+kind: Certificate
+metadata:
+  name: {{ .Release.Name }}-certifier-certificate
+spec:
+  secretName: "{{ .Release.Name }}-certifier-tls"
+  commonName: "certifier.{{ required "dnsDomain must be provided" .Values.dnsDomain }}"
+  subject:
+    countries:
+      - US
+  duration: 87600h
```
hcgatewood: These hard-coded 10-year values should probably be variables. Related: what is the max duration supported for this value?

This is probably also a good point to consider what happens after they expire. The problem is we provision gateways with the specific root and fluentd crts, so when they expire under our current architecture, shit goes down. Specifically

- Root
    - All gateways drop connection to Orc8r, requiring reprovisioning of all gateways in network
        - Solution 1: support in-band mechanism for Orc8r sending new rootCA.pem to all gateways (messy/what if a gateway was disconnected from Orc8r for long period of time)
        - Solution 2: remove support for self-signed certs -- this way, gateways wouldn't need to have a rootCA.pem, instead relying on OS-level support
- Fluentd
    - Same story as root but just for log aggregation. Probably same solution too
- Certifier
    - Should be fine if the underlying keypair remains the same

hcgatewood (commented):
Excited for this. Couple general notes

- Is the general plan to create these secrets with new names, then gradually transition application code to consuming these cloud-native versions of the secrets?
- We should be able to rotate every secret except for
    - rootCA key+crt, since the rootCA.pem is manually provisioned onto each gateway... though see my note for how we could get around this
    - certifier key+crt, since that would effectively flush all admin_operator crts... though perhaps doing this infrequently but regularly would actually be a good idea? cc @karthiksubraveti @mpgermano 
- However: when a secret is rotated, we'd need to bounce the services that consume (the old version of) that secret. I think there should be an easy way to automate that, e.g. https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

CODEOWNERS
ShubhamTatvamasi: Okay, I will remove this for now.

orc8r/cloud/helm/orc8r/charts/certs/values.yaml
ShubhamTatvamasi: By default, most users use letsencrypt with **HTTP-01 challenge** which is okay for a single public domain.

But we have a wild card sub-domain for NMS which requires **DNS-01 challenge** for gathering certificate which requires updating DNS entries by cert-manager. And not all Domain providers are supported by cert-manager.

That's why I think we should leave it to the user for this setup.

orc8r/cloud/helm/orc8r/charts/certs/templates/admin-operator-pkcs12-password.yaml
ShubhamTatvamasi: Okay, Will update this

orc8r/cloud/helm/orc8r/charts/certs/templates/certifier-certificate.yaml
ShubhamTatvamasi: We don't have to worry about the expiration of these certificates. cert-manager will auto-renew those.
Even if the `secret` is deleted it will just create a new secret with a fresh certificate.

orc8r/cloud/helm/orc8r/values.yaml
ShubhamTatvamasi: It's 10 years

https://github.com/magma/magma/blob/master/orc8r/cloud/deploy/scripts/self_sign_certs.sh#L42

ShubhamTatvamasi:
@hcgatewood if we just set the low `duration`  time it will automatically renew these certificates.

We can also set `renewBefore: 12h` or any custom time to rotate it before expiration.

orc8r/cloud/helm/orc8r/charts/certs/templates/certifier-certificate.yaml
ShubhamTatvamasi: I don't know the max duration, but I was able to generate 100 years duration certificate.

orc8r/cloud/helm/orc8r/values.yaml
hcgatewood: Oh wow I 100% thought it was 1 year. Good to know

orc8r/cloud/helm/orc8r/charts/certs/templates/certifier-certificate.yaml
hcgatewood: So my concern isn't on the Orc8r side -- I agree that everything should work fine, from a cert perspective, on Orc8r side. The issue actually is we manually copy the root cert to every AGW that will connect to the Orc8r.

This means that once the *original* root cert expires, the gateways will immediately view Orc8r controller certs as invalid. And even though Orc8r has renewed its root cert, we don't currently have a mechanism to propagate that renewed cert down to each gateway

Let me know if I'm missing something

orc8r/cloud/helm/orc8r/values.yaml
hcgatewood: I'm fine with leaving default relatively high, but since auto-renew should work well, something like 1 or 3 years for these non-root certs also seems reasonable. Those original values were chosen under the expectation that it would be a hassle to manually renew

That aside, we do want to rotate the bootstrapper.key frequently, e.g. on a monthly or yearly basis, since it's being used relatively frequently (once every 4 days x number of gateways pointing to the Orc8r)

orc8r/cloud/helm/orc8r/charts/certs/values.yaml
hcgatewood: I see, thanks for providing context. I think that's a sensible decision then

orc8r/cloud/helm/orc8r/charts/certs/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Create orchestrator certs.
+create: true
+
+dnsDomain: localhost
```
karthiksubraveti: minor nit: Let's just call this domain_name, calling it dns domain sounds redundant.

orc8r/cloud/helm/orc8r/values.yaml
```diff
@@ -52,6 +52,34 @@ secret:
     orc8r: orc8r-secrets-configs-orc8r
   envdir: orc8r-secrets-envdir
 
+# certs sub-chart configuration.
```
karthiksubraveti, line 55: @hcgatewood This probably obsoletes the secret section above right ?

karthiksubraveti (commented):
@hcgatewood @ShubhamTatvamasi I'm still trying to understand how cert-manager works here. Can you help me understand what's the operator's flow here. When are the certs generated and how are they pushed to the individual pods ? 

Also we have a orc8r-secrets chart, Does that get obsoleted by this addition ?

@hcgatewood When I was reading through cert manager docs, It seems like our current gateway bootstrapping mechanism also can be replaced by this.  Is my understanding right ?

orc8r/cloud/helm/orc8r/values.yaml
arunuke: Agree with @hcgatewood on the need for rotation. The CA/Browser forum officially [adopted ](https://thehackernews.com/2020/09/ssl-tls-certificate-validity-398.html)398 days as the time limit for expiring security certificates. We should aspire to meet this goal with the option to automatically check and renew security certificates, which cannot happen with self-signed certificates since they do not share ancestry with any other certificates

orc8r/cloud/helm/orc8r/charts/certs/templates/certifier-certificate.yaml
arunuke: The possible resolution for this requires that only public CA signed certificates be used over self-signed (which cannot be automatically renewed). The classic flow would be to (a) Using an API, request for a public CA certificate that is used as the root to generate all the other keys. AWS Certificate Manager provides public CA certificates for free. (b) In the absence of an API based request or certificates already available, use public CA certificates from a local directory to generate all the other certificates (c) set expiry to 398 days for certificates (d) use cert manager to renew before expiry.

orc8r/cloud/helm/orc8r/values.yaml
ShubhamTatvamasi: I have added a new `bootstrapper-certificate`, which will create a new RSA key for us and also added `rotationPolicy: Always` to it.

orc8r/cloud/helm/orc8r/charts/certs/values.yaml
ShubhamTatvamasi: Updated the name to `domainName`

orc8r/cloud/helm/orc8r/values.yaml
ShubhamTatvamasi: No, we are not using any of these auto-generated secrets inside pods.

ShubhamTatvamasi:
@mattymo can you please also have a look at this PR?

mattymo (approved):

hcgatewood:
Mirroring my comments from Slack: I'm not ready to merge this PR until we have a clear story for how we're handling cloud-native Orc8r certs

This PR introduces useful new affordances for how Orc8r handles its certificates, but I think we'll miss out on concrete impact if we don't start with a clear high-level story for cloud-native Orc8r certs -- specifically since these changes will affect AGWs as well.

I'll write up a short GH Discussion proposal this week and send out a cal invite for us to align on how to proceed

hcgatewood (changes requested):
Clearing my review queue. See https://github.com/magma/magma/discussions/5728#discussioncomment-556134 for comments on next steps in getting this PR merged

ShubhamTatvamasi:
@hcgatewood I have added the following things:
1. let's encrypt issuers
1. rotation timings per secret

But the default is not set to let's encrypt issuer, because **HTTP-01 challenge** doesn't support wilcard certificates.

orc8r/cloud/helm/orc8r/charts/certs/templates/bootstrapper-certificate.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+{{- if .Values.create }}
+apiVersion: cert-manager.io/v1
+kind: Certificate
```
hcgatewood, line 14: Bootstrapper is just a private key -- no certificate necessary

(That said, if it's easier to have it be a cert instead of just a private key, that's fine. Just add a comment indicating as such.)

orc8r/cloud/helm/orc8r/charts/certs/templates/admin-operator-certificate.yaml
```diff
@@ -0,0 +1,37 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+{{- if .Values.create }}
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: {{ .Release.Name }}-admin-operator-certificate
```
hcgatewood, line 16: Minor nit: perhaps this is the right time to start calling these admin operator certs `default_admin_operator`. We support any number of admin operator certificates (i.e. certs giving admin permissions to the rest api), this is just the default one we create during provisioning

hcgatewood, line 16: Side note for a future PR: we'll still need to inject this default cert into the certifier service https://magma.github.io/magma/docs/next/orc8r/deploy_install#create-an-orchestrator-admin-user

Though perhaps we can invert this requirement -- instead of it being a manual step during installation, the certifier service can instead proactively check for the default-admin-operator-tls secret and pull it in if it exists

orc8r/cloud/helm/orc8r/charts/certs/values.yaml
```diff
@@ -0,0 +1,63 @@
...
+  # letsencrypt-prod-issuer
+  # letsencrypt-staging-issuer
+  # NOTE: this won't work with wildcard domain.
+  customIssuer:
+
+controller:
+  # 10 year = 87600h
+  duration: 87600h
+  renewBefore: 24h
+  customIssuer:
```
hcgatewood, line 46: Let's also add the note here about using the provided letsencrypt issuers

hcgatewood (changes requested):
Looking good, I think we're almost there. Couple top-level things

### ACME challenge type

I don't understand the reasoning for supporting HTTP01 -- my cursory understanding is that it won't work at all for NMS (?).

To me, the goal here is to provide a *default that just works*. Supporting DNS01 via Let's Encrypt allows us just that -- a default that "just works". Let me know if I'm misunderstanding something.

So then we have the following deployment types

- (a) ACME (managed) (default)
	- Default via Let's Encrypt
	- Also support for non-default providers via Helm values
- (b) Self-signed (managed)
- (c) Traditional (unmanaged)
	- This is the current mechanism
	- Upload arbitrary secrets via AWS KMS, manually inject into the cluster via Terraform

I don't think (a) works unless we provide DNS01 challenge type for our Let's Encrypt provider. Looks like (b) works great given what you've written, and of course (c) works bc it's our existing functionality.

### Configuring deployment type

I think keeping these helm charts configurable, as you have them now, is the right choice. Then, when we update Terraform with the option to deploy cert-manager, we can provide better guidance on selecting between the 3 above-mentioned deployment types.

hcgatewood:
From offline sync

- For now, we will default to self-signed issuers and forego support for ACME-based issuers
	- HTTP01: NMS requires wildcard certs, and we don't want to create the footgun of requiring a kubectl command in order to create a new NMS tenant
	- DNS01: Orc8r doesn't currently manage its DNS records. @ShubhamTatvamasi will take on writing up a proposal for remedying this, deprioritized till after we integrate this existing PR with Orc8r/NMS as a whole
- This restricts us to the following deployment types: managed self-signed and unmanaged traditional. This will still be a win, however, because (a) for self-signed deployments, cert management will be more straightforward, and (b) this PR paves the way for, once we wrangle the DNS management story, having fully-managed+fully-trusted certs generated via DNS01 method

orc8r/cloud/helm/orc8r/charts/certs/values.yaml
ShubhamTatvamasi: Removed letsencrypt issuers as per our discussion.

orc8r/cloud/helm/orc8r/charts/certs/templates/bootstrapper-certificate.yaml
ShubhamTatvamasi: I have updated the same on `values.yaml` file.

orc8r/cloud/helm/orc8r/charts/certs/templates/admin-operator-certificate.yaml
ShubhamTatvamasi: For now let it be as it is, so we don't get confused and also it's compatible with previous versions.

ShubhamTatvamasi: I will look into it next once this PR has merged.

orc8r/cloud/helm/orc8r/charts/certs/templates/controller-certificate.yaml
```diff
@@ -0,0 +1,34 @@
...
+  renewBefore: {{ .Values.controller.renewBefore }}
+  issuerRef:
+{{- if .Values.controller.customIssuer }}
+    name: {{ .Values.controller.customIssuer }}
+{{- else }}
+    name: {{ .Release.Name }}-root-issuer
+{{- end }}
+  dnsNames:
+  - "*.{{ required "domainName must be provided" .Values.domainName }}"
+  - "*.nms.{{ required "domainName must be provided" .Values.domainName }}"
```
hcgatewood, line 33: Why are we having controller.crt and nms.crt each cover both controller and nms domains?

hcgatewood (approved):
Lgtm. One small q then let's merge

orc8r/cloud/helm/orc8r/charts/certs/templates/controller-certificate.yaml
ShubhamTatvamasi: Keep both for backwards compatibilities.

Issue #5490, opened 2021-03-15
[AGW] Silence extraneous GCC -Wunused-parameter warnings


electronjoe:
## Situation

Now that we have turned up warnings with `-Wextra` (see #5358), there are many GCC warnings of type `-Wunused-parameter` around functions with unused parameters.  A subset (perhaps the vast majority) of these functions are callback functions with a specified interface, so the unused parameter is necessary.

Per [this discussion on Herb Sutter's Blog](https://herbsutter.com/2009/10/18/mailbag-shutting-up-compiler-warnings/) it seems the best solution in C land for this issue is to add a (void) cast of the unused variable at the top of the function, like follows.

```c
int emm_proc_attach_complete(mme_ue_s1ap_id_t ue_id, const_bstring esm_msg_pP,
                             int emm_cause,
                             const nas_message_decode_status_t status) {
  // Function with emm_cause never referenced.
...
```

Becomes

```
int emm_proc_attach_complete(mme_ue_s1ap_id_t ue_id, const_bstring esm_msg_pP,
                             int emm_cause,
                             const nas_message_decode_status_t status) {
  (void)emm_cause; // -Wunused-parameter, callback interface required
...
```

I have tested this with GCC 9.3 and it successfully removes our `-Wunused-parameter` warnings.

## Action

**We should comb through the codebase and for all expected unused parameters, apply this modification**.

The following are per-file counts of the warning emission for the MME (build_oai) target. Derived from [a complete ouput of pertinent warnings](https://gist.github.com/electronjoe/7806a2305b22b1f212e87f46e3dc63e2) (with recent but undocumented commit - apologies).

```shell
cd magma/lte/gateway
make build_oai | grep "Wunused-parameter" | tee build.log
cat build.log  | cut -d " " -f 1 | cut -d ":" -f 1 | sort | uniq -c
      2 /magma/lte/gateway/c/oai/common/async_system.c
      1 /magma/lte/gateway/c/oai/common/backtrace.c
      6 /magma/lte/gateway/c/oai/common/log.c
      1 /magma/lte/gateway/c/oai/common/pid_file.c
      8 /magma/lte/gateway/c/oai/common/shared_ts_log.c
     14 /magma/lte/gateway/c/oai/lib/3gpp/3gpp_24.008_common_ies.c
      3 /magma/lte/gateway/c/oai/lib/3gpp/3gpp_24.008_gmm_ies.c
[...]

PR #5594, opened 2021-03-18
[common] Annotate unused parameters

electronjoe:
Working towards #5490


electronjoe:
This approach is deprecated and we want to move to `__attribute__((unused))` Instead, see referenced GH Issue. I will likely fixup this PR.

PR #5602, opened 2021-03-18
[orchestrator][metricsd] remove exporter queuing

Scott8440:

## Summary
Simple refactor of the remote_exporter to reduce complexity. We had originally built this exporter to queue up datapoints that were submitted, then periodically push them out. This was probably because it's what we did for the FB internal datastore (we needed to optimize it for fewer, larger pushes). With our existing and upcoming supported datastores we don't need to do that, so removing the queuing makes things a lot simpler.

The removal of queuing in the exporter meant I needed to add a tiny bit of queueing in the cloud gatherer specifically, since the way it works it would cause a submit for every single datapoint which felt non-ideal. In the future I think the way that collector works could be improved (e.g. don't use a channel) but that's out of the scope of this refactor I think.

## Test Plan
Unit tests

## Additional Information

### DIFF
--- a/orc8r/cloud/go/services/metricsd/servicers/servicer.go
+++ b/orc8r/cloud/go/services/metricsd/servicers/servicer.go
@@ -29,6 +29,10 @@ import (
 	"golang.org/x/net/context"
 )
 
+const (
+	cloudMetricQueueSize = 100
+)
+
 // MetricsControllerServer implements a handler to the gRPC server run by the
 // Metrics Controller. It can register instances of the Exporter interface for
 // writing to storage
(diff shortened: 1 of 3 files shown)


karthiksubraveti:
@Scott8440 I believe XWF might still be doing the pushes to ODS. (cc @mpgermano ??)

Scott8440:
> @Scott8440 I believe XWF might still be doing the pushes to ODS. (cc @mpgermano ??)

Sorry should have been more clear. The ODS exporter in `fbinternal` still does queuing. Back when we made the prometheus exporter we copied that design, but it doesn't actually make any sense to do that with prometheus/non-ods stores.

Scott8440:
Closing for stats, please commandeer and finish when you get to it :)

PR #5605, opened 2021-03-18
[lte_integ_test] force reprovision on integ_test

uri200:

## Summary

When we run lte_integ_test we normally use `fab integ_test:destroy_vm=false`. That allow us to reuse the VMs. However that skips vagrant provisioning that can cause some changes not to be added to the VM

This PR forces the re-provision in case `destroy_vm=false`. This does not have any impact when we run the regular command that CI runs `fab integ_test`

Note we have added yet another flag to disable that too (disable of reprovison can be done using `provsion_vm=False`)

## Test Plan
```
fab integ_test:destroy_vm=false
fab integ_test:destroy_vm=false,provision_vm=false
```

## Additional Information

### DIFF
--- a/lte/gateway/fabfile.py
+++ b/lte/gateway/fabfile.py
@@ -188,7 +188,7 @@ def s1ap_setup_cloud():
 
 
 def integ_test(gateway_host=None, test_host=None, trf_host=None,
-               destroy_vm="True"):
+               destroy_vm='True', provision_vm='True'):
     """
     Run the integration tests. This defaults to running on local vagrant
     machines, but can also be pointed to an arbitrary host (e.g. amazon) by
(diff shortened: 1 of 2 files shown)


lte/gateway/fabfile.py
```diff
@@ -188,7 +189,7 @@ def s1ap_setup_cloud():
 
 
 def integ_test(gateway_host=None, test_host=None, trf_host=None,
-               destroy_vm="True"):
+               destroy_vm="True", provision_magma="True"):
```
ardzoht:
```suggestion
               destroy_vm="True", force_provision="True"):
```

lte/gateway/fabfile.py
```diff
@@ -208,10 +209,16 @@ def integ_test(gateway_host=None, test_host=None, trf_host=None,
     """
 
     destroy_vm = bool(strtobool(destroy_vm))
+    provision_magma = bool(strtobool(provision_magma))
+
+    if not destroy_vm and provision_magma:
+        # this will only run in case we don't destroy the VM
+        os.system("vagrant provision magma")
```
ardzoht: I would use `local("vagrant provision magma")` this way you don't have to include `os` module

ardzoht (changes requested):

lte/gateway/fabfile.py
uri200: ohh was looking for that!!
Thanks!

lte/gateway/fabfile.py
```diff
@@ -188,7 +188,7 @@ def s1ap_setup_cloud():
 
 
 def integ_test(gateway_host=None, test_host=None, trf_host=None,
-               destroy_vm="True"):
+               destroy_vm='True', provision_magma='True'):
```
ardzoht: why not `provision_vm`? seems better naming for the param (similar to `destroy_vm`)

ardzoht (approved):

themarwhal (approved):
lgtm!

hcgatewood (approved):

tmdzk (approved):

PR #5606, opened 2021-03-18
[docs]Changes to avoid builds when only docs folders are committed

ranj8work:
## Summary
Changes to avoid build if only changes are made in docs/ folder

## Test Plan

**CircleCI build on making changes only to the docs folder:**
https://app.circleci.com/pipelines/github/magma/magma?branch=docs_build_test

## Additional Information

- [NO ] This change is backwards-breaking

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,7 @@ orbs:
                   fi
                 done
                 echo "No changes in [$paths]"
-                circleci step halt
+                circleci-agent step halt
 
 commands:
   install-go:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -83,14 +83,25 @@ orbs:
                   exit 0
                 fi
 
-                paths=".circleci circleci <<parameters.paths>>"
+                paths="circleci docs <<parameters.paths>>"
```
tmdzk, line 86: The solution is just to add docs here and it should be fine. 
Please let `.circleci` It is a valid repo that we need to loop through

tmdzk (changes requested):

PR #5612, opened 2021-03-18
[AGW] Change log message due to lack of time sync; impacts bootstrapping.

sudhikan:
Changes log message for when bootstrap fails due to incorrect system time.

Signed-off-by: Sudhi Kandi <sudhikan@fb.com>

## Summary

In the lab we tried bringing up a gateway who's time was at least 8 hours off. As such we were seeing the following error. The error wasn't quite clear that the system time was off, after a little bit of debugging in looking at the bootstrapping code, the issue became apparent. Feel that the log messaging could have been better to point the user to the root cause faster. 

```
INFO:root:StateReporting (Checkin) failure threshold met, remediating...
Mar 18 10:20:24 debian magmad[4175]: ERROR:root:Error making SSL connection: No such file or directory, [Errno 2] No such file or directory
Mar 18 10:20:24 debian magmad[4175]: ERROR:root:Bootstrapping due to invalid cert
Mar 18 10:20:24 debian magmad[4175]: ERROR:root:Received a not-yet-valid certificate from: 2021-03-18 17:20:21.685505
```

Adds a log message that hints at checking system time if bootstrap fails due to notBefore time happens after current system time. 

## Test Plan
NA; log message change. 

## Additional Information

### DIFF
--- a/orc8r/gateway/python/magma/magmad/bootstrap_manager.py
+++ b/orc8r/gateway/python/magma/magmad/bootstrap_manager.py
@@ -371,7 +371,7 @@ class BootstrapManager(SDWatchdogTask):
         not_before = cert.not_before.ToDatetime()
         if now < not_before:
             logging.error(
-                'Received a not-yet-valid certificate from: %s', not_before)
+                'Current system time indicates certificate received is not yet valid (notBefore: %s). Consider checking NTP.', not_before)
             return False
 
         not_after = cert.not_after.ToDatetime()


orc8r/gateway/python/magma/magmad/bootstrap_manager.py
```diff
@@ -371,7 +371,7 @@ def _is_valid_certificate(self, cert):
         not_before = cert.not_before.ToDatetime()
         if now < not_before:
             logging.error(
-                'Received a not-yet-valid certificate from: %s', not_before)
+                'Cert notBefore time occurs after the current system time; check ntp? notBefore time is %s', not_before)
```
hcgatewood: `'Current system time indicates certificate received from %s is not yet valid. Consider checking NTP.'`

hcgatewood (approved):
👍

orc8r/gateway/python/magma/magmad/bootstrap_manager.py
sudhikan: thanks for the rephrase. this reads better :) will update.

hcgatewood:
Force-merge to override broken CI check. Change is small and unrelated.

PR #5618, opened 2021-03-19
[cloud] [orc8r] Building artifacts for Cloudstrapper

arunuke:

## Summary

Building all orchestrator artifacts for Cloudstrapper

## Test Plan

Build screenshots added. Orchestrator docker images tagged and github repo uploaded with helm files.

Screen Shot 2021-03-19 at 12 49 02 AM (3)

## Additional Information

None.


experimental/cloudstrapper/playbooks/build-configure.yaml
```diff
@@ -0,0 +1,21 @@
+---
+
+- hosts: "{{ buildnode }}"
+  roles:
+    - { role: build-platform, tags: platform }
+  vars:
+    awsTargetRegion: "{{ awsOrc8rRegion }}"
```
karthiksubraveti, line 7: one minor comment. I see use of both camel casing and underscores. If we can use one convention that will be great. Personally i feel using underscores for variable naming suits ansible and python.

karthiksubraveti (approved):

experimental/cloudstrapper/playbooks/build-configure.yaml
arunuke: Agree with the overall idea of using a consistent naming formation. However, in this case, I have used underscores where the underlying tool (in this case ansible), backend application (a shell script that use environment variables) or general platform related conventions (linux environment variables).

PR #5628, opened 2021-03-19
[CI] Turn up Dockerfile Linting in GH PR

electronjoe:
[Hadolint](https://github.com/hadolint/hadolint) is an industry leading Dockerfile linting tool written in Haskell.  This GH action uses [reviewdog](https://github.com/reviewdog/reviewdog) for Github Actions to annotate our Dockerfiles with linting warnings from Hadolint - in Pull Requests containing Dockerfiles.

Hadolint also lints shell commands thanks to the fantastic [Shellcheck](https://github.com/koalaman/shellcheck) linting tool for the shell.

The sample findings on my existing Dockerfile are very helpful, this seems worth having available to us on PRs.  @rdefosse if you find it annoying, or disagree with its advice, we can certainly whitelist some directories so they aren't linted - if you like. Let me know.

Presently I have it linting any file that is touched. We can alternatively lint only modified lines (one line change). Up to us.

### DIFF
--- a/.github/workflows/hadolint.yml
+++ b/.github/workflows/hadolint.yml
@@ -1,0 +1,17 @@
+name: Dockerfile Linting
+on: [pull_request]
+jobs:
+  hadolint:
+    name: Dockerfile Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v1
+      - name: hadolint
+        uses: reviewdog/action-hadolint@v1
+        with:
+          github_token: ${{ secrets.github_token }}
+          reporter: github-pr-review # Default is github-pr-check
+          # Ignore DL3005-"Do not use apt-get upgrade or dist-upgrade"
+          hadolint_ignore: DL3005
+          filter_mode: file


electronjoe:
This Dockerfile linting utility seems quite useful - and took only a moment to activate.

Time for me to apply some fixups to my Dockerfile! LOL

tmdzk (approved):
Approved: but let's disable 
```
[hadolint] reported by reviewdog 🐶 Use $(...) notation instead of legacy backticked `...`.
```

electronjoe:
> Approved: but let's disable
> 
> ```
> [hadolint] reported by reviewdog 🐶 Use $(...) notation instead of legacy backticked `...`.
> ```

I looked into the backticked operator - and we are in fact using it for the `nproc` evaluations. Do you still want it disabled?

rdefosse:
@electronjoe Hello Scott,

This is nice.

2 questions/comments:

- about the backticked and nproc: is there any alternative? also could we tweak the linter syntax checks?
- Why is it complaining about `apt-get upgrade`? In my experience, if you don't do that, you may face issues at some time of your development. Or is is done multiple time in the dockerfile?

Raphael

rdefosse:
@electronjoe 

Who are codeowners on the `.github` folder? It is not specified in the `CODEOWNERS.md` file and based on the number of requested reviewers, anyone can contribute to it!

electronjoe:
> @electronjoe
> 
> Who are codeowners on the `.github` folder? It is not specified in the `CODEOWNERS.md` file and based on the number of requested reviewers, anyone can contribute to it!

Good point. We should add owners, presumably we can mimic the CODEOWNERS from .circleci? @tmdzk - thoughts?

electronjoe:
> @electronjoe Hello Scott,
> 
> This is nice.
> 
> 2 questions/comments:
> 
> * about the backticked and nproc: is there any alternative? also could we tweak the linter syntax checks?
> * Why is it complaining about `apt-get upgrade`? In my experience, if you don't do that, you may face issues at some time of your development. Or is is done multiple time in the dockerfile?
> 
> Raphael

On the backtick linting - we can turn off the particular linting warning if we want to maintain back-tick (instead of using $()?).  Do we prefer the old style for readability? Or..

As to the `apt-get upgrade` use, it sounds like Docker warns that this is a dangerous (fragile? not sure) operation.  For example, in their [Dockerfile Best Practices Doc](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run).

> Avoid RUN apt-get upgrade and dist-upgrade, as many of the “essential” packages from the parent images cannot upgrade inside an unprivileged container. If a package contained in the parent image is out-of-date, contact its maintainers. If you know there is a particular package, foo, that needs to be updated, use apt-get install -y foo to update automatically.

[Edit]: It also sounds like generally they recommend pinning versions if they matter to us, and `apt-get upgrade` doesn't pin the upgrade versions (which presumably we need some specific version, if we require the upgrade request?).  So their recommendation of apt-get install -y foo=version is probably what we should aim for?

rdefosse:
> As to the apt-get upgrade use, it sounds like Docker warns that this is a dangerous (fragile? not sure) operation. For example, in their Dockerfile Best Practices Doc.

Yes I understand the point. But I have an example where I was given a Dockerfile without it and my Ubuntu base image was like 6 months newer that the one used by the dockerfile author and it could not build an image.

For me, this is good if you never change the base image. But for reproducibility, you may face the issue I mentioned.

rdefosse:
> They recommend pinning versions if they matter to us

For me (and that is also true for the packages we are installing from source), this is quite dangerous. You may not get security updates if you do so.

electronjoe:
> > As to the apt-get upgrade use, it sounds like Docker warns that this is a dangerous (fragile? not sure) operation. For example, in their Dockerfile Best Practices Doc.
> 
> Yes I understand the point. But I have an example where I was given a Dockerfile without it and my Ubuntu base image was like 6 months newer that the one used by the dockerfile author and it could not build an image.
> 
> For me, this is good if you never change the base image. But for reproducibility, you may face the issue I mentioned.

I think they recommend pinning the base image as well - for this reason (they do, I got linter warnings when using :latest).

> For me (and that is also true for the packages we are installing from source), this is quite dangerous. You may not get security updates if you do so.

I think for this - security updates, fixes to dependencies, etc - this needs to be addressed with CI process?  E.g. we want a Dockerfile that has everything pinned so it's reproducible.  But then we want ~weekly runs with :latest on everything - and if that passes test suites then we migrate to those new versions.  Granted this relies on people making those promotions (or automation to do so being authored). And that can be an issue =D

[edit] Also, thanks for the discussion @rdefosse, this is helpful for me.

electronjoe:
## On `apt-get upgrade`

@rdefosse thanks for poking at this topic for me! I found a [very good article](https://pythonspeed.com/articles/security-updates-in-docker/) this weekend on `apt-get upgrade` and how the advice to avoid it by Docker Best practices and Hadolint are actually bad advice (per your rationale).  You are right, no surprise I suppose. I tend to trust best practices published by tool providers but this sounds like it's being removed as it's bad advice!

## On Pinning Versions

I'm still not totally sure on this one. Our existing dockerfiles pin many libraries - but not all dependencies.  Is the practice applied today (e.g. Dockerfile.ubuntu18.04) to pin only what won't work unless pinned? And leave everything else unpinned? How often does that break your builds & tests?

Thanks!

rdefosse:
@electronjoe 

scott i think we misunderstood each other.

can you please make a comment in `lte/gateway/docker/mme/Dockerfile.rhel8` so I can see on that file?

Raphael

electronjoe:
@rdefosse done! Re-running checks now.

rdefosse:
@electronjoe 

OK. Thanks. Pretty much the same kind of linter issues.

Regards
Raphael

rdefosse (approved):

PR #5629, opened 2021-03-19
[CI] only run c/c++ build and test when appropriate

electronjoe:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,8 @@ aliases:
           - develop
   - &lte_build_verify
     paths: "orc8r lte"
+  - &c_cpp_build_verify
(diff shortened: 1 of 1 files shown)


electronjoe:
My remaining question is how CodeCov.io will handle this - for PRs that don't touch c/c++ targets. Will  it ~assume no change to test coverage even without a report? We will see.

tmdzk (approved):

tmdzk:
Oh auto-merge was enable. Let me know if it CodeCov handles that. 
cc @electronjoe

electronjoe:
> Oh auto-merge was enable. Let me know if it CodeCov handles that.
> cc @electronjoe

Yeah no worries, we'll see how CodeCov handles it by experimentation =D Not hard to revert if we need to.

PR #5646, opened 2021-03-22
AGW: mobilityD: handle missing gw config.

pshelar:
It is possible to have AGW without any default route.
Following patch handle such case.

## Additional Information


ardzoht (approved):
lgtm

PR #5648, opened 2021-03-22
[CI] enhance oai pipeline on secutity

rdefosse:
## Summary

Based on a discussion I had with Timothee, I am fetching the `python` scripts from a trusted source and not from the pull request context.

It will prevent any unknown contributor to mess with them

I've also added an automatic docker clean-up procedure when the build fails.

Finally I fixed the warnings/errors counting mechanism. Currently for the `oai-mme` component, it was reporting no warnings. Not correct since the removal of `-Werror` flag.

## Test Plan

This version of pipeline is already in use. Let see how it behaves a bit.

## Additional Information

### DIFF
--- a/ci-scripts/JenkinsFile-OAI-Container-GitHub
+++ b/ci-scripts/JenkinsFile-OAI-Container-GitHub
@@ -99,7 +99,7 @@ pipeline {
               // Check if the pull request has files that will impact MME behavior
               // If so, we will run the OAI pipeline
               // For security reasons, we retrieve from a trusted forked repository
-              sh 'wget https://raw.githubusercontent.com/rdefosse/magma/trusted-oai-ci-pipeline/ci-scripts/check_pr_modified_files_for_oai_pipeline.py -o ci-scripts/ci-check_pr_modified_files_for_oai_pipeline.py'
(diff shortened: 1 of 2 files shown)


ci-scripts/JenkinsFile-OAI-Container-GitHub
```diff
@@ -98,7 +98,9 @@ pipeline {
             try {
               // Check if the pull request has files that will impact MME behavior
               // If so, we will run the OAI pipeline
-              sh 'python3 ci-scripts/check_pr_modified_files_for_oai_pipeline.py'
+              // For security reasons, we retrieve from a trusted forked repository
+              sh 'wget https://raw.githubusercontent.com/rdefosse/magma/trusted-oai-ci-pipeline/ci-scripts/check_pr_modified_files_for_oai_pipeline.py -o ci-scripts/ci-check_pr_modified_files_for_oai_pipeline.py'
```
tmdzk: The overall looks good however I think the source should be user agnostic. (we have a dedicated magma repo for that)

tmdzk (changes requested):

ci-scripts/JenkinsFile-OAI-Container-GitHub
rdefosse: @tmdzk 

I can pass both (`userspace` and `branch-name`) as parameters.

Is it OK with you?

Raphael

rdefosse: Then once this PR is merged, I can set the parameters to (`magma` and `master`).

119Vik (approved):

tmdzk (approved):

PR #5655, opened 2021-03-22
[MME] Migrate assertion to __builtin_trap()

### LINKED ISSUE #5357: [MME] Address any pertinent clang compiler warnings in MME
## Summary

Now that Clang build is working (thanks to a number of minor fixup PRs), Clang with -Wall has brought to light a number of new Compiler Warnings.  We should work through the warnings.

The `-Wunknown-warning-option` and `-Wunused-command-line-argument` warnings are due to CMake config targeting gcc. Others are potentially more interesting.

```shell
root@8830d9627e79:/magma/lte/gateway# grep "\[-W" clang-build.log | awk -F'[][]' '{print $2}' | sort | uniq -c
      1 -Wconstant-conversion
      6 -Wdeprecated-declarations
    182 -Wenum-conversion
    144 -Wextern-c-compat
      2 -Winconsistent-missing-override
     10 -Winitializer-overrides
      8 -Wnon-c-typedef-for-linkage
    353 -Wnull-dereference
     51 -Wparentheses-equality
      4 -Wpointer-bool-conversion
      4 -Wtautological-constant-out-of-range-compare
      2 -Wtautological-overlap-compare
      1 -Wtautological-pointer-compare
    324 -Wtypedef-redefinition
   1967 -Wunknown-warning-option
   1916 -Wunused-command-line-argument
```

For tracking of this effort's sub-Issues:

- [ ] -Wextern-c-compat - Issue #5143

electronjoe:
Clang warns that the existing assertion approach is not guaranteed to result in SEGFAULT for all compilers or systems.  Clang advises moving to __builtin_trap() as a modernization.

Test Plan

- Run end-to-end integration tests
- Force an AssertFatal and validate GDB halts appropriately

### DIFF
--- a/lte/gateway/c/oai/common/assertions.h
+++ b/lte/gateway/c/oai/common/assertions.h
@@ -49,19 +49,22 @@
   } while (0)
 
 #ifndef __clang_analyzer__
-#define _Assert_SegFault_                                                      \
+// TODO: explore use of portable github.com/scottt/debugbreak.
+// In the event that __builtin_trap behavior is not desirable across platforms
+// and compilers.
+#define _Assert_Builtin_Trap_                                                  \
   do {                                                                         \
     fprintf(                                                                   \
         stderr,                                                                \
-        "\n Will Intentionaly raise SEGFAULT to be catched by GDB!\n");        \
+        "\n Will use __builtin_trap() to be caught by GDB!\n");                \
     display_backtrace();                                                       \
     fflush(stdout);                                                            \
     fflush(stderr);                                                            \
-    *(int*) 0 = 0;                                                             \
+    __builtin_trap();                                                          \
     exit(EXIT_FAILURE);                                                        \
   } while (0)
 #else
-#define _Assert_SegFault_ _Assert_Exit_
+#define _Assert_Builtin_Trap_ _Assert_Exit_
 #endif
 
 #define _Assert_(cOND, aCTION, fORMAT, aRGS...)                                \
@@ -80,7 +83,7 @@
   } while (0)
 
 #if DEBUG_IS_ON
-#define _ASSERT_FINAL_ _Assert_SegFault_
+#define _ASSERT_FINAL_ _Assert_Builtin_Trap_
 #else
 #define _ASSERT_FINAL_ _Assert_Exit_
 #endif


themarwhal:
@ulaskozat  / @ardzoht could we possibly review/merge this? The current approach and the new approach end up with a very similar looking coredump, but the buildin_trap results in a MUCH nicer Sentry stacktrace.(In the current approach, it doesn't show up.)

I've added a `DevAssert(false);` manually to trigger the assert to test.

GDB backtrace with the current approach:
Screen Shot 2021-04-21 at 3 22 40 PM

GDB backtrace with __builtin_trap:
Screen Shot 2021-04-21 at 3 27 50 PM

Sentry stacktrace with __buildin_trap:
Screen Shot 2021-04-21 at 3 28 25 PM
Screen Shot 2021-04-21 at 3 28 42 PM

Sentry would be much more useful if we can catch all the AssertFatals. :)

ulaskozat:
@themarwhal why is this in draft mode?

themarwhal:
@ulaskozat I've asked @electronjoe  to put it up for review :) I think he hadn't tested it when he wrote the PR, but I did yesterday.

ardzoht (approved):
lgtm, just one followup question, does this results in a non-zero exit status for MME? asking because health service will look for these to detect crash loops / failures on the service

themarwhal:
@ardzoht 
It fails with error code 4
```
magma@mme.service: Main process exited, code=dumped, status=4/ILL
```

electronjoe:
Just saw this was moving forward. I'm happy that Marie was able to test it.

From GCC Docs (clang mimics):

> __builtin_trap function causes the program to exit abnormally. GCC implements this function by using a target-dependent mechanism (such as intentionally executing an illegal instruction) or by calling abort. The mechanism used may vary from release to release so you should not rely on any particular implementation.

SO yes, should result in error code (as Marie sees).

I guess the remaining question is what ARM support looks like for this?  It's GCC and Clang supported.

I'm finding ARM patches for __builtin_trap().

> ARM, __builtin_trap() generates a call to abort()

From a [library for ARM + many other architectures](https://github.com/scottt/debugbreak) to support debug_break().

PR #5677, opened 2021-03-23
[feg] Add ability to use orc8r with S6a_cli

uri200:

## Summary

Add capability to run the client from AGW through orcr8r

## Test Plan

make precommit
used in terravm

## Additional Information

### DIFF
--- a/feg/gateway/services/s8_proxy/client_api.go
+++ b/feg/gateway/services/s8_proxy/client_api.go
@@ -17,11 +17,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"magma/feg/cloud/go/protos"
 	"magma/feg/gateway/registry"
+	"magma/orc8r/lib/go/util"
 
 	"github.com/golang/glog"
+	"google.golang.org/grpc"
 )
 
 type s8ProxyClient struct {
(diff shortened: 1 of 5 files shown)


feg/gateway/services/s8_proxy/client_api.go
```diff
@@ -17,9 +17,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
+
+	"google.golang.org/grpc"
```
themarwhal: I think the grpc include should be below glog

themarwhal (approved):

feg/gateway/services/s8_proxy/client_api.go
uri200: ops. Thanks!

feg/gateway/tools/s6a_cli/main.go
```diff
@@ -180,6 +182,7 @@ func air(cmd *commands.Command, args []string) int {
 	var cli s6aCli
 	var peerAddr string
 	if len(s6aAddr) > 0 || useMconfig { // use direct HSS connection if address is provided
+		fmt.Println("Using builtin S8_proxy")
```
emakeev: S6a_proxy...

feg/gateway/services/s8_proxy/client_api.go
```diff
@@ -62,7 +65,15 @@ func SendEcho(req *protos.EchoRequest) (*protos.EchoResponse, error) {
 }
 
 func getS8ProxyClient() (*s8ProxyClient, error) {
-	conn, err := registry.GetConnection(registry.S8_PROXY)
+	var (
+		conn *grpc.ClientConn
+		err  error
+	)
+	if util.GetEnvBool("USE_REMOTE_S8_PROXY") {
```
emakeev, line 72: You may also use util.GetEnvBool("USE_REMOTE_S8_PROXY", true) if you want it default to remote

emakeev (approved):

feg/gateway/services/s8_proxy/client_api.go
uri200: But then when `USE_REMOTE_S8_PROXY` is missing it will be defaulted to true. 
Isn't it kind of counter intuitive?

themarwhal (approved):

mpgermano (approved):

PR #5683, opened 2021-03-23
[MME][RHEL][OAI] Better work on building for RHEL8

rdefosse:
## Summary

I did not like the fact that we were patching `lte/gateway/c/oai/oai_mme/CMakeLists.txt` to append a few packages

This morning with Mohit, we tried to append them anyway but the `agw_of` variant on the Debian MAGMA-dev VM would not link

So I took another approach and tried to detect we were on `RHEL` OS and then added the missing packages

`cmake` does not provide this out of the box, hence the analysis of `/etc/os-release`

## Test Plan

```bash
sudo podman build --no-cache --target magma-mme --tag magma-mme:test --ignorefile lte/gateway/docker/mme/.dockerignore  --file ci-scripts/docker/Dockerfile.mme.ci.rhel8 .
```

Worked fine

as well with Ubuntu18 dockerfile.

## Additional Information


lte/gateway/c/oai/oai_mme/CMakeLists.txt
```diff
@@ -62,3 +62,19 @@ endif ( NOT EMBEDDED_SGW )
...
+      if("${_var}" MATCHES "^(ID)=(.*)$")
+        set(os_name "${CMAKE_MATCH_2}")
+      endif()
+    endforeach()
+    if ("${os_name}" STREQUAL "\"rhel\"")
+      target_link_libraries(mme
+        gpr yaml z cares boost_program_options boost_filesystem boost_system boost_regex double-conversion address_sorting gflags unwind iberty dl)
+    endif()
+  endif()
+endif ( NOT EMBEDDED_SGW AND NOT S6A_OVER_GRPC )
```
lionelgo, line 80: you mean here that all other non MME S11/S6a deployments won't build on RHEL. 
It is the case now I imagine.
We could detect the OS as you do and not bother with EMBEDDED_SGW, S6A_OVER_GRPC ?

rdefosse: I wanted to be the less intrusive possible.

Currently adding these "missing" libraries is only needed for that variant.

rdefosse:
RestartIntegrationTests

119Vik (approved):

tmdzk (approved):

ardzoht (approved):
lgtm

mattymo (approved):

PR #5689, opened 2021-03-23
[pipelined]Use Rule Version Provided by SessionD

koolzz:

## Summary

This PR adds logic to use the rule version provided by sessiond in the flow rules. Currently I'm reusing the old version dict for rule removal logic but In the future we can get rid of that completely.

## Test Plan

make test

lte/cwf integ tests

## Additional Information


themarwhal (approved):
feg changes lgtm

pshelar (approved):

themarwhal (approved):

uri200 (approved):

PR #5698, opened 2021-03-23
[LTE][integtests] run timeout in the foreground for it to receive CTRL+C

talkhasib:

## Summary

We are currently unable to kill a running integtest using CTRL+C since timeout command places itself in its own process group. Thus not receiving the signal from the parent shell.

Run timeout in the foreground for it to receive the signal from the shell

## Test Plan
Run integtest and kill it using CTRL+C


ssanadhya:
@talkhasib , can you also test the 'fab integ_tests' job to ensure this does not affect CI?

ulaskozat (approved):
LGTM

talkhasib:
Tested with "fab integtest". No issues.

ssanadhya (approved):

PR #5699, opened 2021-03-23
[PipelineD] Add policy+version pairs to ActivateFlows and DeactivateFlows requests

### LINKED ISSUE #4092: [Sessiond]Move to absolute value stats reporting

### LINKED ISSUE #4599: Move rule version generation to Sessiond

themarwhal:

## Summary
Initial changes to include version information to Activate/Deactivate requests.

## Test Plan
no test

## Additional Information


lte/protos/pipelined.proto
```diff
@@ -74,34 +74,45 @@ message RequestOriginType {
...
-  repeated PolicyRule dynamic_rules = 4;  // List of dynamic rules obtained from PCRF
+  repeated PolicyRule dynamic_rules = 4;  // List of dynamic rules obtained from PCRF TODO deprecate
   RequestOriginType request_origin = 5; // Activate flow source (GX/GY)
   magma.lte.AggregatedMaximumBitrate apn_ambr = 6; // APN AMBR info, passed at the time of session creation
   bytes ipv6_addr = 7; // Subscriber session ipv6 address
   bytes msisdn = 8; // From SubscriberDB or HSS
   uint32 uplink_tunnel = 9; // Subscriber uplink tunnel
   uint32 downlink_tunnel = 10; // Subscriber downlink tunnel
-
+  repeated PolicyAndVersion rules = 11; // Policy + version pairs
```
koolzz: Let's name this field policies (that way the indexing looks better policy.rule)

lte/protos/pipelined.proto
```diff
@@ -74,34 +74,45 @@ message RequestOriginType {
...
   reserved 3;
 }
 
+message PolicyIDAndVersion {
```
koolzz: Already defined above? Also should we name this something like VersionedPolicy, having 'and' in the name sounds a bit off to me

koolzz (approved):

lte/protos/pipelined.proto
themarwhal: ah sounds good. I'll rename
PolicyAndVersion -> VersionedPolicy
PolicyIDAndVersion -> VersionedPolicyID ?

PR #5702, opened 2021-03-23
[SessionD] Refactor PipelineDClient to take in RulesToProcess struct into requests

### LINKED ISSUE #4092: [Sessiond]Move to absolute value stats reporting

### LINKED ISSUE #4599: Move rule version generation to Sessiond

themarwhal:
## Summary
Another PR in preparation for adding rule version generation to Activate/Deactivate flows request. (See here: https://github.com/magma/magma/pull/5699)
Eventually, the `RulesToProcess` struct will have
1. `std::vector<PolicyRule> rules`
2. `std::vector<uint32_t> versions`

This will help in passing the Policy+version combination into PipelineDClient->PipelineD.

## Test Plan
unit tests + integ

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -542,16 +542,19 @@ void LocalEnforcer::install_final_unit_action_flows(
       const auto& rule = create_redirect_rule(action);
       // check if the rule has been installed already.
       if (!session->is_gy_dynamic_rule_installed(rule.id())) {
+        RulesToProcess to_process;
+        to_process.rules = std::vector<PolicyRule>{rule};
         pipelined_client_->add_gy_final_action_flow(
-            imsi, ip_addr, ipv6_addr, teids, msisdn, {rule});
+            imsi, ip_addr, ipv6_addr, teids, msisdn, to_process);
         session->insert_gy_dynamic_rule(rule, lifetime, session_uc);
       }
       return;
     }
     case RESTRICT_ACCESS: {
+      RulesToProcess to_process;
+      to_process.rules = action->get_restrict_rules();
       pipelined_client_->add_gy_final_action_flow(
-          imsi, ip_addr, ipv6_addr, teids, msisdn,
-          action->get_restrict_rules());
+          imsi, ip_addr, ipv6_addr, teids, msisdn, to_process);
       return;
     }
     default:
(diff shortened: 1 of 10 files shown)


lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -323,18 +323,19 @@ void LocalEnforcer::execute_actions(
...
-  const auto ambr                         = action_p->get_ambr();
-  const std::vector<PolicyRule> rule_defs = action_p->get_rule_definitions();
+  const std::string imsi      = action_p->get_imsi(),
+                    msisdn    = action_p->get_msisdn(),
+                    ip_addr   = action_p->get_ip_addr(),
+                    ipv6_addr = action_p->get_ipv6_addr();
+  const Teids teids           = action_p->get_teids();
+  const auto ambr             = action_p->get_ambr();
+  RulesToProcess to_process;
+  to_process.rules = action_p->get_rule_definitions();
```
uri200, line 333: will `ServiceAction` be updated to use RulesToProcess too?

Or in other words, who will set the versions of those rules? 
Is this versioning only needed to make sure sessiond and pipelined are in sync? Or is policyd also impacted by that versioning?
(sorry if you have already described this somewhere else )

themarwhal: I think I can leave ServiceAction alone as processing that information is handled in `LocalEnforcer`. When it activates/deactivates rules in `SessionState`, we will fetch / populate the rule version.

The rule versioning will be owned by SessionD. So whenever it activates / deactivates a rule, SessionD will keep track of the updated version number. PipelineD will report the version attached to rule stats as well. 

For example if ruleX is installed with version 1 and then removed with version 2. If PipelineD keeps sending stats for ruleX at version 1, we know the deactivation didn't go through properly.
So this versioning exchange will help us keep track of the state syncs better. 

This is for `PipelineD` and `SessionD` only.

uri200: Versioning make sense. 

We just have to make sure we don't fix pipelined issues at sessiond level

uri200 (approved):

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: Yeah totally agree.

PR #5730, opened 2021-03-25
[MME]  Free cid in mme_app_edns_emulation

### LINKED ISSUE #5333: [MME] Address Clang-Tidy clang-analyzer-unix.Malloc findings
This GH Issue is to investigate the clang-analyzer-unix.Malloc findings from Clang-Tidy. Analysis was run on Magma @ [6c13c49](https://github.com/magma/magma/commit/9cc4be1e362e4cd0171be2bfcb47a054d6c13c49).

These are all of the shape `warning: Potential leak to memory pointed to by...`.

This is a result of #5302 execution of Clang-Tidy.

To keep track of the findings and status of each, here is a checklist which will tie to PRs:

- [ ] bstrlib.c:2465
- [ ] bstrlib.c:3161
- [ ] obj_hashtable.c:1018 [#5802]
- [ ] obj_hashtable.c:1053 [#5802]
- [ ] obj_hashtable_uint64.c:998 [#5801]
- [ ] mme_app_edns_emulation.c:63 [#5730]
- [ ] mme_app_state_manager.cpp:235
- [ ] s1ap_mme_handlers.c:1271

As these findings tend to be long (the static analyzer indicates the precise path through the codebase required for the leak) I will point at full logs and not quote them here.

Full logs from the above are available in [This](https://gist.github.com/electronjoe/987d20e3e730d0a22ce184644c70b449) Github Gist.

anushadatar:
## Summary
clang-tidy currently shows a warning in the file
mme_app_edns_emulation.c due to an allocated cid
value that is never freed. Free that value prior
to returning.

Signed-off-by: Anusha Datar <anushadatar@gmail.com>

## Test Plan

```
cd lte/gateway
make build_oai
```
And CI Pipelines.

### DIFF
--- a/lte/gateway/c/oai/tasks/mme_app/mme_app_edns_emulation.c
+++ b/lte/gateway/c/oai/tasks/mme_app/mme_app_edns_emulation.c
@@ -57,8 +57,12 @@ int mme_app_edns_add_sgw_entry(bstring id, struct in_addr in_addr) {
 
       hashtable_rc_t rc =
           obj_hashtable_insert(g_e_dns_entries, cid, strlen(cid), data);
-      if (HASH_TABLE_OK == rc) return RETURNok;
+      if (HASH_TABLE_OK == rc) {
+        free(cid);
+        return RETURNok;
+      }
     }
+    free(cid);
   }
   return RETURNerror;
 }


lte/gateway/c/oai/tasks/mme_app/mme_app_edns_emulation.c
```diff
@@ -60,6 +60,7 @@ int mme_app_edns_add_sgw_entry(bstring id, struct in_addr in_addr) {
       if (HASH_TABLE_OK == rc) return RETURNok;
     }
   }
+  free(cid);
```
electronjoe: Awesome! Finding real memory leaks and fixing them!

## Minor Change Request

My recommendation on this would be to move the free() up to just before line 60. The rationale is that when others come along and modify this function, they could add additional return statements that short-circuit your free (not seeing it).  I think it's best to free the resource as soon as you are done with it.

## Better Yet - But not this PR? File a GH Issue for discussion?

We might consider a broader campaign to improve the safety of our allocations like this one (where the resource should clearly be freed when context exits).  A better solution in this type of context would even be to consider a RAII style approach in C.  It appears GCC (and maybe Clang but undocumented) support the `__cleanup__` variable attribute. The major negative here is that no such capability is defined by the c standard, so this will be compiler-specific which is generally a cautious thing to introduce.  If we did this, it would look something like this.

```c
/ *
  * Some shared utility header
  * /
void clean_up_char_array(char* to_be_freed)
{
  free(to_be_freed);
}

/ *
  * Use in this file or others like it
  * /
// No matter what happens, when cid goes out of context we call 
char* cid __attribute__ ((__cleanup__(clean_up_char_array))) = calloc(1, blength(id) + 1) ;
if (stuff){
  // do things
  return;
}
return;
```

anushadatar: I'll definitely apply the change you mentioned, but I think it makes sense to move the RAII point into a larger discussion because I feel unsure about the implications associated with the compiler specificity of using GCC. Should I file the GitHub issue associated with this - if you have broader vision and context and want to do it, go for it, but I also can.

electronjoe: I spoke with some of the team about the `__cleanup__` RAII option - and @amarpad brought up the excellent point that we are going to be supporting ARM more and more. Looking into it, I [could not find support for __cleanup__ by the armcc compiler](https://developer.arm.com/documentation/dui0472/k/Compiler-specific-Features/Function-attributes?lang=en). I do see `__cleanup__` used in linux's `systemd`, which surprises me (how can this not support ARM?).  But perhaps that is a [rogue linux agent](https://lwn.net/Articles/589433/).  Sadly, I think we have to avoid __cleanup__ use.  Another option I'm finding is that some have authored smart-pointer like libraries for c. But that's another discussion =D Lets drop the GH Issue request and just move the free here, then it LGTM and you can convert to a PR (remove draft)!

electronjoe: As to testing this PR, I realize you have Vagrant issues that are blocking you. There was a regression on our vagrant based E2E testing this week - or a couple regressions. Rebasing to master might fix your issue?  I'm testing that now myself for my PRs.  If we cannot get this working for you in reasonable time, we could rely on the OAI CI test (MAGMA-OAI-TEST in CI report) coverage which seems reasonable to me for this PR (and not run local Vagrant tests).

electronjoe: Vagrant VM is still broken due to migration to Ubuntu 20.04 effort - you should omit that test and rely entirely on CI tests for this PR - you have my OK.

anushadatar: Sounds good, I'll make the change later this morning. Is there anything I need to do to omit the local vagrant tests within the context of this PR?

anushadatar:
@electronjoe  Made the change requested, thanks for the feedback. Please let me know if there's anything else that needs to happen here wrt RAII or anything else on this PR.

electronjoe (approved):
Let's convert this from a draft, and I'll get code owners to review for commit!

anushadatar:
> Let's convert this from a draft, and I'll get code owners to review for commit!

Marked as ready for review, thanks for all of your help!

lte/gateway/c/oai/tasks/mme_app/mme_app_edns_emulation.c
```diff
@@ -57,6 +57,7 @@ int mme_app_edns_add_sgw_entry(bstring id, struct in_addr in_addr) {
 
       hashtable_rc_t rc =
           obj_hashtable_insert(g_e_dns_entries, cid, strlen(cid), data);
+      free(cid);
```
ulaskozat: what happens when data is null?

ulaskozat:
This part of code will be removed, it is legacy and not utilized in our current call flows.

lte/gateway/c/oai/tasks/mme_app/mme_app_edns_emulation.c
anushadatar: That's an important catch,  moved `free` to line 62.

anushadatar:
> This part of code will be removed, it is legacy and not utilized in our current call flows.
Mot sure what y'alls philosophy is in terms of deciding whether or not to update legacy code or to wait for larger changes to delete this; feel free to let me know if i should abandon this PR.

electronjoe:
> This part of code will be removed, it is legacy and not utilized in our current call flows.

Can we enumerate what is to be deleted and delete it all now in this PR?

If not I think we merge the fix?

lte/gateway/c/oai/tasks/mme_app/mme_app_edns_emulation.c
electronjoe: Whoops! Thanks @ulaskozat for catching that!

lte/gateway/c/oai/tasks/mme_app/mme_app_edns_emulation.c
```diff
@@ -59,6 +59,7 @@ int mme_app_edns_add_sgw_entry(bstring id, struct in_addr in_addr) {
           obj_hashtable_insert(g_e_dns_entries, cid, strlen(cid), data);
       if (HASH_TABLE_OK == rc) return RETURNok;
```
ulaskozat: we still need to free before returning okay.

ulaskozat (changes requested):
We can merge it and handle clean up of dead code in another PR. however the current PR has still an issue.

ardzoht:
@anushadatar What is the plan for this PR?

anushadatar:
> @anushadatar What is the plan for this PR?

Sorry, dropped the ball on this. Will add the extra `free` later today and update this accordingly.

lte/gateway/c/oai/tasks/mme_app/mme_app_edns_emulation.c
anushadatar: Freed before returning okay, thanks for the note!

ulaskozat (approved):
LGTM

ulaskozat:
For required units tests to pass, you will need to rebase your PR.

anushadatar:
Rebased! Thanks for all of your help, and please let me know if there is anything else I need to do.

ulaskozat:
Enabled auto-merge, if not merged due to some requirement not passing, ping me...

electronjoe:
Congrats on your merge @anushadatar! Thanks for the review @ulaskozat!

PR #5731, opened 2021-03-25
[agw][fix for PLMN restriction] PLMN restriction handle mnc digit 2

pruthvihebbani:
[agw][fix for PLMN restriction] PLMN restriction handle mnc digit 2

## Summary

Add support to handle PLMN with 2 digit MNC. If the configuration/filter has a 2 digit MNC, then validate only first 5 digits of IMSI instead of 6 digits to check for PLMN restriction.

## Test Plan

- Verified s1sim sanity
- Verified test_attach_restricted_plmn.py TC


ulaskozat (approved):
LGTM

PR #5733, opened 2021-03-25
[docs] Reorganize docusaurus by topic

hcgatewood:
## Summary

As part of making the Docusaurus more approachable, we're flipping the hierarchy of topics to feature "what I want to do" vs. "what component". Ref: https://fb.quip.com/RMQ1AVZtcaP9

Sidebar changes:

Screen Shot 2021-03-25 at 12 50 38 AM
Screen Shot 2021-03-25 at 12 50 46 AM
Screen Shot 2021-03-25 at 12 50 53 AM
Screen Shot 2021-03-25 at 12 50 59 AM
Screen Shot 2021-03-25 at 12 51 14 AM

## Test Plan

- [x] Manual inspection

## Additional Information


themarwhal (approved):
looks much cleaner! thanks!

docs/readmes/lte/deploy_config_agw.md
```diff
@@ -18,11 +19,11 @@ You will need to set up a super-user in a valid NMS Organization in order to
 use the NMS. See "[NMS Multitenancy](https://magma.github.io/magma/docs/nms/nms_organizations)"
 to complete this step.
 
-You also should have completed all the steps in "[Access Gateway Setup (On Bare Metal)](https://magma.github.io/magma/docs/lte/setup_deb)".
+You also should have completed all the steps in "[Install Access Gateway (On Bare Metal)](https://magma.github.io/magma/docs/lte/deploy_install)".
 For this part, we strongly recommend that you SSH into the AGW box from a host
 machine instead of using the AGW directly.
 
-In this latest version, Magma Access Gateway no longer has a hardwired default Access Point Name (APN). Therefore, each UE must have a subscription profile that includes at least one APN to be able to attach to the network. Please follow the instructions in "[APN Configuration](config_apn.md)".
+In this latest version, Magma Access Gateway no longer has a hardwired default Access Point Name (APN). Therefore, each UE must have a subscription profile that includes at least one APN to be able to attach to the network. Please follow the instructions in "[APN Configuration](deploy_config_apn.md)".
```
themarwhal, line 26: I think the version we made the switch was 1.3? cc @ulaskozat 
Maybe we can explicitly mention the version.

docs/docusaurus/create_docusaurus_website.sh
```diff
@@ -38,5 +38,5 @@ echo 'If you want to follow the build logs, run docker-compose logs -f docusauru
 spin
 echo 'Navigate to http://localhost:3000/ to see the docs.'
 
-open 'http://localhost:3000/' || true
+open 'http://localhost:3000/docs/next/basics/introduction.html' || true
```
andreilee, line 41: Is there a way to always just point to the newest release? I don't think we should show the docs from `master` by default, that's only useful for developers, but not users

andreilee (approved):
I like this new structure better!

uri200 (approved):

docs/docusaurus/i18n/en.json
```diff
@@ -936,16 +951,24 @@
       "Community": "Community"
     },
     "categories": {
+      "Getting Started": "Getting Started",
+      "Tutorials": "Tutorials",
+      "Usage": "Usage",
+      "Architecture": "Architecture",
+      "Guide: Deploy": "Guide: Deploy",
```
karthiksubraveti: I dislike seeing "Guide:" on the sidebars, It might be better for us to either call this 'Deploy', 'Architecture', 'Usage' itself or have a subsection under guide. Since we may have more guides in the future, I would probably go with latter.

docs/README.md
```diff
@@ -12,9 +12,10 @@ organized into a Docusaurus site for easy consumption.
 [Docusaurus](https://docusaurus.io/) is a framework for static site generation,
 focusing on documentation-based sites.
 
-The `sidebars.json` and `siteConfig.js` are the main entrypoints for updating
-our documentation site. The former determines which README docs are displayed,
-and the latter sets site-wide config values.
+The [`sidebars.json`](https://v1.docusaurus.io/docs/en/navigation) and
+[`siteConfig.js`](https://v1.docusaurus.io/docs/en/site-config) are the main
+entrypoints for updating our documentation site. The former determines which
+README docs are displayed, and the latter sets site-wide config values.
```
karthiksubraveti, line 18: minor nit: Can you also add the naming convention around the what the ID value should be for each document.

karthiksubraveti (approved):
Thanks for making this change. It looks great. Left a few comments.

docs/docusaurus/i18n/en.json
hcgatewood: The latter was the original intent. Unfortunately Docusaurus doesn't support two-tiered subcategorization, so we can't do e.g. Guides > Deploy > Orchestrator. I think for now we can just remove the `Guide: ` and, if the number of guides gets too large, we can figure out a solution then

docs/docusaurus/create_docusaurus_website.sh
hcgatewood: `/` points to newest release. My assumption, and experience, is that when you're editing the docs, you're editing and viewing the "next" docs. That's why I've included this change

PR #5744, opened 2021-03-25
AGW: deploy OVS 2.14 on magma-dev VM

pshelar:
Use prebuilt OVS packages for ubuntu.

## Additional Information

### DIFF
--- a/lte/gateway/deploy/roles/magma/tasks/main.yml
+++ b/lte/gateway/deploy/roles/magma/tasks/main.yml
@@ -231,6 +231,14 @@
       # install openvswitch
       - openvswitch-switch
       - openvswitch-test
+      - libopenvswitch
+      - openvswitch-datapath-dkms
+      - openvswitch-common
+      - libopenvswitch-dev
+      - python3-openvswitch
+      - openvswitch-doc
+      - openvswitch-pki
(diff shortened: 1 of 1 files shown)


lte/gateway/deploy/roles/magma/tasks/main.yml
```diff
@@ -231,6 +231,14 @@
       # install openvswitch
       - openvswitch-switch
       - openvswitch-test
+      - libopenvswitch
```
ardzoht, line 234: It would be good to verify this with building local package, to confirm there's no dependency issue?

ardzoht (approved):
lgtm

tmdzk (changes requested):
This role is for local VM. It's pointing to an internal registry that don't have any OVS packages.. Which means It will take the community one and as we didn't fix the versions It will install the 2.15 one. 

2 Solutions: 
- host ovs packages in that internal registry
- Fix version of OVS

pshelar:
facebookconnectivity has OVS 2.14 packages. I have already verified these patches. Once external repo is up and running we can upload patches OVS packages in that repo, so that deployments would not hit community repo.

tmdzk (approved):

PR #5745, opened 2021-03-25
AGW: deploy envoy from magma repo.

pshelar:
Repo URL: https://getenvoy.io/gpg is not reliable, Use
package cached in magma repo.

## Additional Information

### DIFF
--- a/lte/gateway/deploy/magma_dev_focal.yml
+++ b/lte/gateway/deploy/magma_dev_focal.yml
@@ -48,6 +48,5 @@
         magma_repo: /home/{{ ansible_user }}/magma-packages
         magma_deps: /home/{{ ansible_user }}/magma-deps
     - role: fluent_bit
-    - role: envoy
     - role: pyvenv
     - role: focal_hacks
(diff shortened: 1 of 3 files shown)


ardzoht (approved):
lgtm

lte/gateway/deploy/magma_dev_focal.yml
```diff
@@ -48,6 +48,5 @@
         magma_repo: /home/{{ ansible_user }}/magma-packages
         magma_deps: /home/{{ ansible_user }}/magma-deps
     - role: fluent_bit
-    - role: envoy
```
tmdzk, line 51: If you remove it from there It won't be install in the devs VM.

tmdzk (changes requested):
Envoy had an issue with their registry but it's fixed now. 

2 solutions: 
- we change the dev role to point to our selfhost version of envoy
- we keep it that way and trust it will never happen again.

tmdzk (approved):

lte/gateway/deploy/magma_dev_focal.yml
pshelar: @tmdzk the magma role would install it from our repo.

PR #5778, opened 2021-03-26
[SessionD] Light ServiceAction refactor to use RulesToProcess struct

### LINKED ISSUE #4599: Move rule version generation to Sessiond

themarwhal:
## Summary
Similar to https://github.com/magma/magma/pull/5702

Essentially, I want to use RulesToProcess to do transfer all Policy+Version propagation. For this I need to migrate ServiceAction as well. :'( 
Although I think this refactor made some logic around ServiceAction cleaner.

Some big changes
1. Split `get_charging_updates` into multiple functions for each switch case. This function was really long and covers all the service action types. (Activate/Redirect/Restrict/Terminate.. etc.)
    -  get_update_for_continue_service for doing the logic to construct a charging update request
    - `fill_service_action_for_activate`, `fill_service_action_for_restrict`, `fill_service_action_for_redirect` for each of the cases
2. Introduce a new `Types.h` to add shared types. The dependency chain of type includes were getting out of hand so introducing a new file to store shared structs/types.

## Test Plan
Run sessiond unit tests
CWF Integration test to make sure GY credit activation / suspension is still ok

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/SessionState.cpp
+++ b/lte/gateway/c/session_manager/SessionState.cpp
@@ -46,7 +46,6 @@ using magma::service303::increment_counter;
 using magma::service303::remove_counter;
 
 namespace magma {
-
 std::unique_ptr<SessionState> SessionState::unmarshal(
     const StoredSessionState& marshaled, StaticRuleStore& rule_store) {
   return std::make_unique<SessionState>(marshaled, rule_store);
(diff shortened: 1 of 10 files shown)


themarwhal:
RestartCWAGTests

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -1679,22 +1643,126 @@ void SessionState::get_charging_updates(
...
+  key.set_credit_usage(&usage);
+
+  auto request = make_credit_usage_update_req(usage);
+  request_number_++;
+  session_uc.request_number_increment++;
+  return request;
+}
+
+void SessionState::fill_service_action_for_activate(
+    std::unique_ptr<ServiceAction>& action_p, const CreditKey& key) {
```
uri200, line 1686: not sure if there is a convention here, but since `action_p` is the object modified/output argument, shouldn't we put it as the last argument?

lte/gateway/c/session_manager/Types.h
```diff
@@ -0,0 +1,190 @@
...
+typedef std::unordered_map<PolicyID, uint32_t, PolicyIDHash> BearerIDByPolicyID;
+
+struct RulesToProcess {
+  // If this vector is set, then it has PolicyRule definitions for both static
+  // and dynamic rules
+  std::vector<PolicyRule> rules;
+  bool empty() const;
+};
+
+}  // namespace magma
```
uri200, line 190: 👍

uri200 (approved):

lte/gateway/c/session_manager/SessionState.cpp
themarwhal: yeah actually, maybe it might be good to just attach these functions to the Action object.... Will fix in one of the stacked changes to avoid a nasty rebase.

PR #5779, opened 2021-03-26
[SessionD] Add a new field to track per-rule versions

### LINKED ISSUE #4599: Move rule version generation to Sessiond

themarwhal:
## Summary
1. Cleanup rule manipulation logic in SessionState. We should only use the deactivet_* activate_* rule functions if we want to activate/deactivate rules. Otherwise, we should work with the object directly. 
2. Add a new PolicyAndStats map to track per-policy versioning. Eventually, this will store flatrate values when we migrate to flat rate reporting from PipelineD. 

## Test Plan
make precommit_sm
CWF integration test

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/RuleStore.cpp
+++ b/lte/gateway/c/session_manager/RuleStore.cpp
@@ -194,14 +194,15 @@ bool PolicyRuleBiMap::get_monitoring_key_for_rule_id(
     const std::string& rule_id, std::string* monitoring_key) {
   std::lock_guard<std::mutex> lock(map_mutex_);
   auto it = rules_by_rule_id_.find(rule_id);
-  if (it == rules_by_rule_id_.end()) {
+  if (it == rules_by_rule_id_.end() ||
+      !should_track_monitoring_key(it->second->tracking_type())) {
     return false;
   }
-  if (should_track_monitoring_key(it->second->tracking_type())) {
+  // nullptr means the caller does not care about retrieving the value
+  if (monitoring_key != nullptr) {
     monitoring_key->assign(it->second->monitoring_key());
-    return true;
   }
-  return false;
+  return true;
 }
 
 bool PolicyRuleBiMap::get_rule_ids_for_charging_key(
--- a/lte/gateway/c/session_manager/SessionState.cpp
+++ b/lte/gateway/c/session_manager/SessionState.cpp
@@ -46,6 +46,12 @@ using magma::service303::increment_counter;
 using magma::service303::remove_counter;
 
 namespace magma {
+
+template<class T>
+void remove_from_vec_by_value(std::vector<T>& vec, T value) {
+  vec.erase(std::remove(vec.begin(), vec.end(), value), vec.end());
+}
+
 std::unique_ptr<SessionState> SessionState::unmarshal(
     const StoredSessionState& marshaled, StaticRuleStore& rule_store) {
   return std::make_unique<SessionState>(marshaled, rule_store);
@@ -113,7 +119,7 @@ StoredSessionState SessionState::marshal() {
   for (auto& it : rule_lifetimes_) {
     marshaled.rule_lifetimes[it.first] = it.second;
   }
-
+  marshaled.policy_version_and_stats = policy_version_and_stats_;
   return marshaled;
 }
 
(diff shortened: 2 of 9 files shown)


lte/gateway/c/session_manager/Types.h
```diff
@@ -187,4 +188,12 @@ struct RulesToProcess {
   bool empty() const;
 };
 
+struct StatsPerPolicy {
+  // The version maintained by SessionD for this rule
+  uint32_t current_version;
+  // The last reported version from PipelineD
+  uint32_t last_reported_version;
```
themarwhal, line 195: `last_reported_version` : Will be set when we receive stats from PipelineD.

lte/gateway/c/session_manager/RuleStore.cpp
```diff
@@ -197,11 +197,13 @@ bool PolicyRuleBiMap::get_monitoring_key_for_rule_id(
   if (it == rules_by_rule_id_.end()) {
     return false;
   }
-  if (should_track_monitoring_key(it->second->tracking_type())) {
+  if (!should_track_monitoring_key(it->second->tracking_type())) {
+    return false;
+  }
+  if (monitoring_key != nullptr) {
```
uri200, line 202: i see that when `monitoring_key==nul` we still return true, is that correct?

wouldn't it be better to do
```
  if (monitoring_key != nullptr || it == rules_by_rule_id_.end() || monitoring_key != nullptr) {
    return false;
  }
```
(maybe remove `monitoring_key` from there if it is really man to return true when null)

uri200: do you know what is the case to use `get_monitoring_key_for_rule_id` with a null monitoring_key?

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -350,114 +356,71 @@ bool SessionState::apply_update_criteria(SessionStateUpdateCriteria& uc) {
...
     if (is_static_rule_installed(rule_id)) {
-      deactivate_static_rule(rule_id, _);
-    } else if (is_static_rule_scheduled(rule_id)) {
-      install_scheduled_static_rule(rule_id, _);
-      deactivate_static_rule(rule_id, _);
-    } else {
-      MLOG(MERROR) << "Failed to merge: " << session_id_
-                   << " because static rule already uninstalled: " << rule_id;
-      return false;
+      remove_from_vec_by_value<std::string>(active_static_rules_, rule_id);
```
uri200, line 368: i read the summary, but why is the reason we can't use deactivate herer?

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -350,114 +356,71 @@ bool SessionState::apply_update_criteria(SessionStateUpdateCriteria& uc) {
...
   for (const auto& rule_id : uc.static_rules_to_uninstall) {
     if (is_static_rule_installed(rule_id)) {
-      deactivate_static_rule(rule_id, _);
-    } else if (is_static_rule_scheduled(rule_id)) {
-      install_scheduled_static_rule(rule_id, _);
-      deactivate_static_rule(rule_id, _);
-    } else {
-      MLOG(MERROR) << "Failed to merge: " << session_id_
-                   << " because static rule already uninstalled: " << rule_id;
-      return false;
```
uri200, line 363: Thanks! never understood why we were exiting the whole thing if there were an error on a specific field here.

uri200 (approved):

lte/gateway/c/session_manager/SessionState.cpp
themarwhal: I wanted to separate out data structure manipulation from actually `activating`/ `deactivating` a rule. `activating`/ `deactivating` would entail some sort of version update. Whereas if we're just translating between StoredSessionState <-> SessionState, we're just copying information rather than updating some state.

It's just confusing because we have to keep translating between structs. I'll add some extra documentation on the relevant functions to make that clear.

lte/gateway/c/session_manager/RuleStore.cpp
themarwhal: Yes the monitoring_key being null or not doesn't affect the return value. It just indicates whether we should copy the monitoring key into the pointer or not. 

It's kind of like how we have an interface for `remove_rule` with a PolicyRule pointer. It's not meant for success/failure indication, but it offers a way to fetch the definition at the same time as removing the rule.

Yeah would be good to consolidate the two failure cases as you said. Will apply the change. :)

lte/gateway/c/session_manager/SessionState.cpp
themarwhal: Actually will update the next PR with documentation, since this PR does not include the version update change yet :)

PR #5787, opened 2021-03-29
[SessionD] Update rule versions on install/uninstalling rules 

themarwhal:
## Summary
This PR adds new logic to: 
1. Every time we enable / disable a rule (gx/gy static/dynamic), we update the version corresponding to the rule. 
2. This value is propagated down to PipelineD as part of the `Activate/DeactivateFlowsRequest`.
3. In case of where all rules are removed by providing an empty DeactivateRequest, `PipelienD` will have to figure out the correct version.

### New / Modified interfaces: 
* `get_current_rule_version` : given a rule_id, get the current version
* `insert_gy_dynamic_rule` -> `insert_gy_rule` : gy static rules were currently not tracked, so expanding the interface to also track static rules as well
* `insert_dynamic_rule`/`activate_static_rule`/`insert_gy_rule` : now returns the updated rule version after inserting a rule
* `remove_dynamic_rule`/`remove_gy_rule`/`deactivate_static_rule` : now returns a optional<uint32_t> that includes the updated version after removing a rules. If the rule removal failed, it will return {} which is a false-y value

`RulesToProcess` now has two fields and a new append method. This struct is now used for ALL interactions where we need to propagate rule definitions to `PipelineDClient`
```
struct RulesToProcess {
  // If this vector is set, then it has PolicyRule definitions for both static
  // and dynamic rules
  std::vector<PolicyRule> rules;
  std::vector<uint32_t> versions;
  bool empty() const;
  void append_versioned_policy(PolicyRule rule, uint32_t version);
};
```

## Test Plan
CWF integration test locally
SessionD unit tests

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -587,13 +587,20 @@ void LocalEnforcer::schedule_static_rule_activation(
                          << "during installation of static rule " << rule_id;
           return;
         }
+        PolicyRule rule;
+        if (!rule_store_->get_rule(rule_id, &rule)) {
+          MLOG(MWARNING) << "Could not find static rules definition for "
+                         << rule_id;
+          return;
+        }
+
         auto& session = **session_it;
         auto& uc      = session_update[imsi][session_id];
 
         std::time_t current_time = time(nullptr);
         // don't install the rule if the current time is out of lifetime
-        if (session->should_rule_be_active(rule_id, current_time)) {
-          session->deactivate_scheduled_static_rule(rule_id, uc);
+        if (!session->should_rule_be_active(rule_id, current_time)) {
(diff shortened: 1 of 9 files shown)


uri200:
```In case of where all rules are removed by providing an empty DeactivateRequest, PipelienD will have to figure out the correct version.```

Do you know why pipelined may care about hte version when deleting?

themarwhal:
> `In case of where all rules are removed by providing an empty DeactivateRequest, PipelienD will have to figure out the correct version.`
> 
> Do you know why pipelined may care about hte version when deleting?

@uri200 Yeah I'm not sure if there are cases where the versions are relevant when all rules are being deleted. I just meant that that is the only case where SessionD doesn't explicitly specify the new version. So in case that becomes relevant, we'll need to do something extra there.

uri200 (approved):

PR #5788, opened 2021-03-29
[CI] Add CodeCov.io flags to disambiguate reports

electronjoe:
## Summary

We are observing erratic coverage reporting as CodeCov.io wants every git hash being reported to have complete coverage statistics - which we are not doing for latency / CPU / cost reasons (e.g. if a PR does not touch c/c++ targets we are omtting the hour+ of testing and upload of coverage).  This PR attempts to use per-coverage-domain codecov.io flags to alleviate the issue (why do they call them flags and not tags! ARG).

The strategy is to apply a codecov.io project flag per CircleCI pipeline stage that pushes a coverage report - to ensure each can be run in isolation of any others.

## Test Plan

Test-ability of CircleCI / CodeCov configuration does not seem great - particular for the multi-target PR such as this.  I think given the current broken codecov behavior, we might as well merge this and then iterate.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -512,6 +512,7 @@ jobs:
             python3 build.py --coverage
       - codecov/upload:
           file: /home/circleci/project/orc8r/cloud/coverage/all.gocov
+          flags: cloud_lint
       - magma_slack_notify
 
   # Fail if checked-in generated code doesn't match output from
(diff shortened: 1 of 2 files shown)


tmdzk (approved):

electronjoe:
Looks like it's working - I am seeing CodeCov per-project coverage stats that seem sane. Lets commit!

PR #5793, opened 2021-03-29
Revert "[CODEOWNERS] Point orc8r codeowners to orc8r-approvers team"

themarwhal:
Reverts magma/magma#5791
Didn't work for some reason. Reverting for now.


themarwhal:
Using admin to revert to make sure orc8r reviews can happen.

PR #5801, opened 2021-03-30
[lib][agw] Free keysp value in obj_hashtable_uint64

### LINKED ISSUE #5333: [MME] Address Clang-Tidy clang-analyzer-unix.Malloc findings

anushadatar:
## Summary

clang-tidy currently shows a warning in the file
obj_hashtable_uint64.c due to an allocated keysP
value that is never freed. Free that value prior
to returning if that value is populated by the
original calloc() call.

Signed-off-by: Anusha Datar anushadatar@gmail.com

## Test Plan

```
cd lte/gateway
make build_oai
```
And CI Pipelines.

### DIFF
--- a/lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c
+++ b/lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c
@@ -994,6 +994,7 @@ hashtable_rc_t obj_hashtable_uint64_get_keys(
       }
     }
 
+    free(keysP);
     PRINT_HASHTABLE(hashtblP, "return OK\n");
     return HASH_TABLE_OK;
   }


lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c
```diff
@@ -994,6 +994,7 @@ hashtable_rc_t obj_hashtable_uint64_get_keys(
       }
     }
 
+    free(keysP);
```
ssanadhya, line 997: This function call is supposed to return the list of keys in a hashtable. If we free it here, then the function will always return an empty list. This memory should be freed by the caller.

ssanadhya (changes requested):

electronjoe:
Progress! Thanks @anushadatar!  And thanks @ssanadhya for the quick review.

@anushadatar in many places within the Magma c codebase, pointed values **can be but are not necessarily** an indication of a return value - because the code often returns only int values for function values, any more sophisticated outputs from a function come from the pointed-to parameters. Unfortunately, unlike Google c++ codebase you have experienced, immutable function parameters are sometimes also passed by pointer instead of const reference (because c has no const reference concept) - making it unclear what function parameters are inputs, what are outputs.

For some of these memory leaks, you have to review the detailed Clang-Tidy findings [regenerate or review [This Gist](https://gist.github.com/electronjoe/987d20e3e730d0a22ce184644c70b449)].

Below is the snippet for this hash table finding.  This is effectively a theorem prover output, providing you a path explanation for the memory leak and therefor hinting at how one would resolve it. Sometimes its easier to read from the tail back up to the start of the leak.

```
/magma/lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c:998:12: warning: Potential leak of memory pointed to by 'keysP' [clang-analyzer-unix.Malloc]
    return HASH_TABLE_OK;
           ^
/magma/lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c:981:7: note: Assuming 'hashtblP' is not equal to NULL
  if (hashtblP == NULL) {
      ^
/magma/lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c:981:3: note: Taking false branch
  if (hashtblP == NULL) {
  ^
/magma/lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c:987:12: note: Memory is allocated
  keysP  = calloc(hashtblP->num_elements, sizeof(void*));
           ^
/magma/lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c:989:7: note: Assuming 'keysP' is non-null
  if (keysP) {
      ^
/magma/lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c:989:3: note: Taking true branch
  if (keysP) {
  ^
/magma/lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c:990:17: note: Assuming 'n' is >= field 'size'
    for (n = 0; n < hashtblP->size; ++n) {
                ^
/magma/lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c:990:5: note: Loop condition is false. Execution continues on line 998
    for (n = 0; n < hashtblP->size; ++n) {
    ^
/magma/lte/gateway/c/oai/lib/hashtable/obj_hashtable_uint64.c:998:12: note: Potential leak of memory pointed to by 'keysP'
    return HASH_TABLE_OK;
```

electronjoe:
Reading through the path taken by Clang-Tidy, and reviewing this portion of the codebase, I see the following.

From what I can tell (no comments but looking at code behavior):

- It appears that our hash table is a [Chained Hash Table with Linked Lists](https://en.wikipedia.org/wiki/Hash_table#Separate_chaining_with_linked_lists)
- hash table `->size` is the number of hash bins (each then containing a linked list)
- hash table `->num_elements` is the number of populated keys across all bin-linked-lists

What I *think* Clang-Tidy is telling us is that the mixture of `->size` and `->num_elements` in `obj_hashtable_uint64_get_keys` below can result in a memory leak if `hashtblP->num_elements` is non-zero and yet `hashtblP->size` is zero (so e.g. `sizeP` never increments to indicate to caller the correct size of the allocated `keysP`.

1. It should not be possible for `size == 0 && num_elements != 0` under correct implementation
    1. Therefore I think this leak should not occur in practice
    1. But we should address this clang-tidy warning in some manner

Options for removal of the warning (if the above assessment is correct):

- An ignore annotation for clang-tidy as a comment on the ?calloc? `// NOLINT(clang-analyzer-unix.Malloc)`
- A guard (maybe early in the function) checking that size != 0

I prefer the latter as it enables Clang-Tidy to continue watching for future leaks as this function is modified. What are your thoughts @ulaskozat?

```c
hashtable_rc_t obj_hashtable_uint64_get_keys(
    const obj_hash_table_uint64_t* const hashtblP, void** keysP,
    unsigned int* sizeP) {
  size_t n                     = 0;
  obj_hash_node_uint64_t* node = NULL;
  obj_hash_node_uint64_t* next = NULL;
  if (hashtblP == NULL) {
    keysP = NULL;
    return HASH_TABLE_BAD_PARAMETER_HASHTABLE;
  }
  *sizeP = 0;
  keysP  = calloc(hashtblP->num_elements, sizeof(void*));
  if (keysP) {
    for (n = 0; n < hashtblP->size; ++n) {
      for (node = hashtblP->nodes[n]; node; node = next) {
        keysP[*sizeP++] = node->key;
        next            = node->next;
      }
    }

    PRINT_HASHTABLE(hashtblP, "return OK\n");
    return HASH_TABLE_OK;
  }
  PRINT_HASHTABLE(hashtblP, "return SYSTEM_ERROR\n");
  return HASH_TABLE_SYSTEM_ERROR;
}
```

ulaskozat:
closing this PR. how to handle clang false alarm can be discussed as a github discussion.

PR #5827, opened 2021-03-30
AGW: MME: add mac header to L3 packets

pshelar:
This makes OVS happily remove the L2 header on egress and avoid
crash.

## Additional Information


ulaskozat (approved):
LGTM

PR #5836, opened 2021-03-30
[lte]Changes to fix code that is build error prone

ranj8work:
Signed-off-by: Ranjini Iyer <ranj8work@gmail.com>

## Summary
Fixed code which could potentially cause build failures on master

## Test Plan
Testing link is pasted below:

https://app.circleci.com/pipelines/github/magma/magma?branch=build_errors_fixes

## Additional Information

- [NO ] This change is backwards-breaking

### DIFF
--- a/circleci/fabfile.py
+++ b/circleci/fabfile.py
@@ -218,6 +218,12 @@ def _run_remote_lte_integ_test(repo: str, magma_root: str):
         # Copy to the directory CircleCI expects
         local('sudo mkdir -p /tmp/test-results/')
         local('sudo mv test-results/* /tmp/test-results/')
+        local('if [ "$(ls -A test-results)" ]; then \n'
+              '   echo "test-results directory NOT empty. Moving files..." \n'
+              '   sudo mv test-results/* /tmp/test-results \n'
+              'else \n'
+              '   echo "test-results directory is empty. Nothing to move." \n'
+              'fi')
 
         # On failure, transfer logs from all 3 VMs and copy to the log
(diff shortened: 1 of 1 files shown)


circleci/fabfile.py
```diff
@@ -218,6 +218,12 @@ def _run_remote_lte_integ_test(repo: str, magma_root: str):
         # Copy to the directory CircleCI expects
         local('sudo mkdir -p /tmp/test-results/')
         local('sudo mv test-results/* /tmp/test-results/')
```
themarwhal, line 220: Do you mean to replace this line with the new one?

ranj8work: Yes @themarwhal. I thought @tmdzk was reviewing it but if you have questions, I am happy to answer it for you. Thank you for taking a look.

circleci/fabfile.py
```diff
@@ -218,6 +218,12 @@ def _run_remote_lte_integ_test(repo: str, magma_root: str):
...
+        local('if [ "$(ls -A test-results)" ]; then \n'
```
tmdzk, line 221: Let's do the logic in python.

tmdzk (changes requested):
The root cause of the CI failure is deeper than that but that's a good catch let's make those changes. (if the folder is empty it means tests were not triggered properly and CI job failed to run)

tmdzk (changes requested):
see my comment about making the logic in python.

ekowtaylor:
@ranj8work , @tmdzk what are the plans to close this?

PR #5851, opened 2021-03-31
[CI] Fail gcc-problems checker if builds fail

electronjoe:
## Summary

Capture the build status for OAI, session_manager when doing GCC warning annotations - return failure status if build fails.

This can then be used as a Required check by Github for a PR Merge.

Making the status of these stages blocking for Github Merge requires UI involvement by an admin, separate from this PR.

## Test

Pushed a [master-breaking OAI change](https://github.com/magma/magma/actions/runs/705306327) and observed the check failure for the following job.

- `build_oai`
- or `build_session_manager`


themarwhal (approved):
lgtm!!

themarwhal:
hmmmmmmm looks like the c++ unit test checks are mandatory even if they didn't run :'( pls rebase and push again or just override.

electronjoe:
> hmmmmmmm looks like the c++ unit test checks are mandatory even if they didn't run :'( pls rebase and push again or just override.

I'll rebase and run again.

PR #5864, opened 2021-04-01
[MME] Save UE Uplink Tunnel in OVS Register

koolzz:
Saving this reg to simplify a bunch of logic in the datapath.

## Additional Information


koolzz:
On second thought I might not even even need this, closing for now...

Maybe for ipv6? will get back to this

PR #5871, opened 2021-04-01
[Docs] Update LTE architecture overview diagram

electronjoe:
Looking for feedback and iteration on any missing edges or components / groupings.

Recommend you toggle "Display the rich diff" so that it renders closer to final doc appearance (e.g. below).

```
   ┏━━━━━━━━━━━━━━┓           ┏━━━━━━━━━━━━━━┓
   ┃ Magma Orc8r  ┃───────────┃  Magma Feg   ┃
   ┗━━━━━━━━━━━━━━┛           ┗━━━━━━━━━━━━━━┛
           ▲      ▲
           │      │
     ┏━━━━━┻━━━━━━┻━━━━━┓
     ┃  control_proxy   ┃
     ┗━━━━━┳━━━━━━┳━━━━━┛
           │      └────────────────────────────────────────────────┐
           └──────────────────────────────────┐                    │
                                              │                    │
┏Access Control & ┓         ┏━━━━Session━━━━┓ │  ┏━━━━Device━━━━━┓ │┏━━━Optional━━━━┓
┃   Management    ┃         ┃   & Policy    ┃ │  ┃  Management   ┃ │┃   Services    ┃
┃   ┌───────────┐ ┃         ┃  Management   ┃ │  ┃ ┌───────────┐ ┃ │┃ ┌────────────┐┃
┃   │    MME    │ ┃         ┃ ┌───────────┐ ┃ └──╋─│  magmad   │ ┃ │┃ │  monitord  │┃
┃┌─▶│ (accessd) │◀╋───────┐ ┃ │ policydb  │◀╋┐   ┃ └───────────┘ ┃ │┃ └────────────┘┃
┃│  └───────────┘ ┃       │ ┃ └───────────┘ ┃│   ┃ ┌───────────┐ ┃ │┃ ┌────────────┐┃
┃│  │     │     ──╋────┐  │ ┃ ┌───────────┐ ┃│   ┃ │   dnsd    │ ┃ │┃ │ redirectd  │┃
┃│  │     └───────╋──┐ │  └─╋▶│ sessiond  │◀╋┘   ┃ └───────────┘ ┃ │┃ └────────────┘┃
┃│  └─────────────╋┐ │ │    ┃ └───────────┘ ┃    ┃ ┌───────────┐ ┃ │┃ ┌────────────┐┃
┃│                ┃│ │ │    ┗━━━━━━━▲━━━━━━━┛    ┃ │  health   │ ┃ │┃ │ conntrackd │┃
┃│  ┌───────────┐ ┃│ │ │            │            ┃ └───────────┘ ┃ │┃ └────────────┘┃
┃└─▶│   sctpd   │ ┃│ │ │            │            ┃ ┌───────────┐ ┃ │┃ ┌────────────┐┃
┃   └───────────┘ ┃│ │ │            │            ┃ │  enodebd  │ ┃ └╋─│td-agent-bit│┃
┃   ┌───────────┐ ┃│ │ │            └───────┐    ┃ └───────────┘ ┃  ┃ └────────────┘┃
┃   │   smsd    │ ┃│ │ │                    │    ┗━━━━━━━━━━━━━━━┛  ┃        ▲      ┃
┃   └───────────┘ ┃│ │ │  Packet Processing │                       ┗━━━━━━━━╋━━━━━━┛
┃                 ┃│ │ │  ┃   Pipeline    ┃ │                                │
┗━━━━━━━━━━━━━━━━━┛│ │ │  ┃ ┌───────────┐ ┃ │                                │
                   │ │ │  ┃ │ pipelined │◀╋─┘                                │
 ┏━━Subscriber ━━┓ │ │ │  ┃ └───────────┘ ┃                                  │
 ┃  Management   ┃ │ │ │  ┃       ▲       ┃  ┏━━Telemetry, Logging, Debug━━━┓│
 ┃ ┌───────────┐ ┃ │ │ │  ┃       │       ┃  ┃  ┌───────────┐┌───────────┐  ┃│
 ┃ │directoryd │◀╋─┘ │ │  ┃ ┌───────────┐ ┃  ┃  │  ctraced  ││  eventd   │──╋┘
 ┃ └───────────┘ ┃   │ │  ┃ │   dpid    │ ┃  ┃  └───────────┘└───────────┘  ┃
 ┃ ┌───────────┐ ┃   │ │  ┃ └───────────┘ ┃  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
 ┃ │ mobilityd │◀╋───┘ │  ┗━━━━━━━━━━━━━━━┛
 ┃ └───────────┘ ┃     │
 ┃┌────────────┐ ┃     │
 ┃│subscriberdb│◀╋─────┘
 ┃└────────────┘ ┃
 ┗━━━━━━━━━━━━━━━┛
```

For comparison, here is what this looks like in my [Excalidraw](https://excalidraw.com/) reproduction. If we are going to settle on one or a couple diagramming tools, which doesn't sound like an awful thing to try to do - I'll move that to a new GH Proposal outside this PR (and presumably hold this PR until that's settled?).

lte_architecture


hcgatewood (commented):
I'm unsure about the text drawings. Excalidraw is my go-to for dev-level documentation https://excalidraw.com/, and we can save the Excalidraw asset file directly in `assets/`

As far as the actual content: I'm not 100% sure for gw side, but my understanding is that Service303 is a generic interface that each service implements individually. If that's the case, it doesn't make sense to put bounding boxes around a set of services with the label service303

pshelar:
DPId is closer to datapath component that should belong to packet processing block.

electronjoe:
> DPId is closer to datapath component that should belong to packet processing block.

Thank you! I'll move it now.

electronjoe:
> I'm unsure about the text drawings. Excalidraw is my go-to for dev-level documentation https://excalidraw.com/, and we can save the Excalidraw asset file directly in `assets/`

One can save the Monodraw assets directly in assets as well (I intended to fixup this PR to do so).

> As far as the actual content: I'm not 100% sure for gw side, but my understanding is that Service303 is a generic interface that each service implements individually. If that's the case, it doesn't make sense to put bounding boxes around a set of services with the label service303

So it's more the API back to Orc8r is driven by some service303 standard?

themarwhal:
We also have `redirectd` which is an optional service that helps `pipelined` with data redirections? (@koolzz will know more)

Does `td-agent-bit` belong in the `logging` section? It handles all log forwarding. 

If there was a graphical way to show that most services have the Service303 interface, that might be better than the boxes. Essentially it is an interface for `MagmaD` to query periodically.

electronjoe:
> Does `td-agent-bit` belong in the `logging` section? It handles all log forwarding.

td-agent-bit sounds like an optional service, at least that is the [claim of our documentation](https://github.com/magma/magma/blob/master/docs/readmes/lte/architecture_overview.md#td-agent-bit)? It certainly is logging related though - so I'm not sure how to differentiate. I thought it might be particularly important to let people know that the log exfiltration is not mandatory.

ardzoht:
For `enodebd`, even though is stated as an optional service for unmanaged eNodebs, the service will still run for stats reporting for both managed and unmanaged devices, I feel it would make more sense to put it in the Device Management section.

docs/readmes/lte/architecture_overview.md
```diff
@@ -29,7 +29,50 @@ Note: The *mme* service will be renamed soon to *accessd*, as its main purpose i
...
+     ┗━━━━━━━━━━━━━━┛           ┗━━━━━━━━━━━━━━┛
+             ▲      ▲
+             │      │
+       ┏━━━━━┻━━━━━━┻━━━━━┓
+╔══════╣  control_proxy   ╠═══════════════════════════════════════════Access Gateway (AGW)
+║      ┗━━━━━┳━━━━━━┳━━━━━┛                                                              ║
+║            │      └────────────────────────────────────────────────┐                   ║
+║            └──────────────────────────────────┐                    │                   ║
+║                                               │                    │                   ║
+║ ┏━━━━Magma EPC━━━━┓         ┏━━━━Session━━━━┓ │  ┏━━━━Device━━━━━┓ │┏━━━Optional━━━━┓  ║
```
ulaskozat: EPC is the whole thing not only access side. Instead simply put Access Control & Management. Can you write (accessd) in parenthesis next to MME?

docs/readmes/lte/architecture_overview.md
```diff
@@ -29,7 +29,50 @@ Note: The *mme* service will be renamed soon to *accessd*, as its main purpose i
...
+║ ┃   ┌───────────┐ ┃         ┃   & Policy    ┃ │  ┃  Management   ┃ │┃   Services    ┃  ║
+║ ┃┌─▶│    MME    │◀╋───────┐ ┃  Management   ┃ │  ┃ ┌───────────┐ ┃ │┃ ┌────────────┐┃  ║
+║ ┃│  └───────────┘ ┃       │ ┃ ┌───────────┐ ┃ └──╋─│  magmad   │ ┃ │┃ │  monitord  │┃  ║
+║ ┃│  │     │     ──╋────┐  │ ┃ │ policydb  │◀╋┐   ┃ └───────────┘ ┃ │┃ └────────────┘┃  ║
+║ ┃│  │     │       ┃    │  │ ┃ └───────────┘ ┃│   ┃ ┌───────────┐ ┃ │┃ ┌───────────┐ ┃  ║
```
ulaskozat: enodebd is also part of device management.

ulaskozat: I would not categorize directoryd and mobilityd as Subscriber Management. Don't have a good alternative name to consider.

ulaskozat (changes requested):
Left some comments, thanks for doing this...

koolzz:
> We also have `redirectd` which is an optional service that helps `pipelined` with data redirections? (@koolzz will know more)
> 
> Does `td-agent-bit` belong in the `logging` section? It handles all log forwarding.
> 
> If there was a graphical way to show that most services have the Service303 interface, that might be better than the boxes. Essentially it is an interface for `MagmaD` to query periodically.

Yeah redirectd runs a flask server that we turn on if we need to enable subscriber policies with redirection (maybe just put it into optional)?

Additionally we also have conntrackd which looks at linux conntrack and sends of packets to trigger datapath ipfix exports. Technically this is not actual packet processing loop though so we can put into optional as well. 

Can you also add an edge from dpi->pipelined we have grpc calls there

docs/readmes/lte/architecture_overview.md
electronjoe: > I would not categorize directoryd and mobilityd as Subscriber Management. Don't have a good alternative name to consider.

Should I just remove the subscriber Management grouping box and leave all three components `directoryd`, `mobilityd` and `subscriberdb` free-floating?

hcgatewood (approved):
Lgtm. Can you include the asset for the diagrams as well, to support future edits

From offline: choosing to not standardize on a diagramming tool at this point. Devs are free to use whatever tool they prefer 🙂

docs/readmes/lte/architecture_overview.md
ulaskozat: I would prefer them floating.

ulaskozat (approved):
Minus the subscriber management group, it looks good.

docs/readmes/lte/architecture_overview.md
electronjoe: I'll float them! Thanks for the feedback @ulaskozat - I think I had auto-merge on so I'll fix this now in a follow up PR.

electronjoe: https://github.com/magma/magma/pull/6123

PR #5868, opened 2021-04-01
[lte][agw] Add hash-based validation to reduce redis writes operations on MME

ardzoht:
## Summary

- Adds hash computation based on contents of serialized proto message for redis_client for task state, ue context and s1ap imsi map values
- Checks if last written proto hash value is different from next to be written, if it's equal, skips write operation
- Remove read state proto to check last version on write, instead keep track of it on `state_manager` as member value

## Test Plan

- make integ test as sanity and continuous attach test
- tested on spirent as performance and load test validation 

## Additional Information

- [ ] This change is backwards-breaking

PR #5888, opened 2021-04-02
[lte][agw] Eliminate some of the unnecessary state syncs on s1ap

ulaskozat:
## Summary

Each message handling by default triggers state synchronization service on S1AP task. This PR eliminates such syncs for the subset of S1AP handled messages that does not make any state modifications. This is complementary to the PR #5868. 

## Test Plan

Integration tests for regression.
Spirent tests for performance. Attached below 3 iterations with 500 UEs, 5 UE/sec attach rate, 500kbps per UE traffic.

Screen Shot 2021-04-01 at 11 16 34 PM

## Additional Information


lte/gateway/c/oai/tasks/s1ap/s1ap_mme.c
```diff
@@ -112,12 +112,16 @@ static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
   state                          = get_s1ap_state(false);
   AssertFatal(state != NULL, "failed to retrieve s1ap state (was null)");
 
+  bool is_state_same = false;
```
ardzoht, line 115: minor nit: I feel `has_state_changed` is a better name for this, and it should reduce the number of the value being set to true

ardzoht (approved):
minor comment, overall lgtm

lte/gateway/c/oai/tasks/s1ap/s1ap_mme.c
ulaskozat: @ardzoht I want to have the default behavior to sync the state and explicitly call out the ones that do not modify the state (at all or most of the time). This will be more robust against new additions to the code base (e.g., s1 mobility).

PR #5905, opened 2021-04-03
[lte][agw] Bug fix in handling ULA message

ulaskozat:
## Summary

On loading tests, this particular bug is observed. ULA message is received after timer has expired and while a detach process is ongoing, the attach procedure continues. The bug fix logic is simple: when ULA message is received, the timer should be still running. If it is not, then it indicates that ULA failure is handled already and this message can be ignored.

## Test Plan

Integration tests and loading tests on spirent.

## Additional Information


ardzoht (approved):
lgtm

PR #5911, opened 2021-04-05
[CI][AGW]Force setuptools version

tmdzk:

[CI][AGW]Force setuptools version

## Summary

If we let the latest setuptools version being installed it require to upgrade python which we are not ready yet, fixing the setuptools version 

## Test Plan

lte-test pass


themarwhal:
Approving to unblock CI.

themarwhal (approved):

PR #5921, opened 2021-04-05
[lte][agw] Bug fix in parsing cause type

ulaskozat:
## Summary

Parsing of the failure cause_type and cause_value were not done correctly in handling context modification failure.

## Test Plan

Sanity integration tests.

## Additional Information


ardzoht (approved):
woah, good catch, lgtm!

PR #5928, opened 2021-04-06
[orc8r][configurator] Fix non-scalable entity get-all

hcgatewood:
### Context

Under non-trivial but expected load, the "get all subscribers" codepath is causing timeouts due to the DB request taking 10s of seconds -- timing out when it passes 60s (due to Orc8r-default 60s gRPC timeout).

This was caused by configurator constructing get-all entity reads to Postgres as `select * from ... where key=$1 or key=$2 or key=$3 or ...`, for O(n) keys.

### Fix bottleneck

This PR updates configurator to make the request with O(1) included values, by reusing the existing query but joining on the assoc table. This should dramatically reduce lock contention, which should resolve this scale bottleneck.

### Additional changes

*ACLs.* We also remove ACL functionality from configurator. This functionality is unused, and, since configurator is a complicated service, removing ACLs makes its non-trivial SQL statements easier to understand+manage.

*Fix tests.* Attempted to make as minimal changes as possible to the tests, especially the integ test. They should be 95%+ identical in terms of functionality tested.

### Performance improvements

Local (dev machine) benchmark, see test plan for details. tl;dr

- Reduce worst-case e2e request duration 6x+ (60s+ to 9s)
- Reduce worst-case DB transaction duration 25x (35s to 1.4s)
- Remove catastrophic lock contention causing e2e request duration timeout after 60s

Screen Shot 2021-04-08 at 2 01 28 AM
Screen Shot 2021-04-08 at 2 08 55 AM

We'll still need to validate whether this fix is sufficient in a real deployment, but I'm assuming it will be sufficient for 15-20k subscribers and a small but non-trivial number of gateways.

We may still need the subscriber cache singleton service to support 400-500 gateways, which is our GA target. The configurator logic can probably be further optimized, but (a) unclear if it would be sufficient for 500 gateways (scale test will tell) and (b) this is a lot of wasted CPU cycles anyway, since it's largely redundant data requests+transformations.

### Correctness

- [x] Mostly minimal changes to existing tests
- [x] Manual inspection

### Performance

Use the following script to locally simulate load on gateways requesting subscribers. Previously, this would cause timeouts (60s+). After, every request completes in under 1s.

```sh
# Load simulation

$ cat ./get_all_updates.sh

#!/usr/bin/env bash

seq 300 | parallel -j 40 ./get_updates.bash

$ cat ./get_updates.sh

#!/usr/bin/env bash

SECONDS=0
grpcurl \
    -insecure \
    -key /tmp/magma_protos/gateway.key \
    -cert /tmp/magma_protos/gateway.crt \
    -authority streamer-controller.magma.test \
    -protoset ~/Desktop/magma_protos/out.protoset \
    -d '{"gatewayId": "1eec0f55-e576-44ce-a5b5-2148ba410646", "stream_name": "subscriberdb"}' \
    localhost:7443 \
    magma.orc8r.Streamer/GetUpdates \
    > /dev/null

echo "$1 took ${SECONDS}s"

$ time ./get_all_updates.sh  # before

...
79 took 56s
73 took 59s
77 took 60s
ERROR:
  Code: Aborted
  Message: error while streaming updates: rpc error: code = DeadlineExceeded desc = context deadline exceeded
82 took 61s
86 took 60s
85 took 60s
ERROR:
  Code: Aborted
  Message: error while streaming updates: rpc error: code = DeadlineExceeded desc = context deadline exceeded
87 took 60s
ERROR:
  Code: Aborted
  Message: error while streaming updates: rpc error: code = DeadlineExceeded desc = context deadline exceeded
94 took 57s
91 took 59s
90 took 60s
ERROR:
  Code: Aborted
  Message: error while streaming updates: rpc error: code = DeadlineExceeded desc = context deadline exceeded
92 took 60s
ERROR:
  Code: Aborted
  Message: error while streaming updates: rpc error: code = DeadlineExceeded desc = context deadline exceeded
112 took 27s
111 took 29s
109 took 33s
...
./get_all_updates.bash ... 4:36.11 total

$ time ./get_all_updates.sh  # after

...
221 took 7s
217 took 7s
213 took 8s
224 took 5s
222 took 7s
226 took 6s
223 took 8s
225 took 7s
230 took 6s
...
./get_all_updates.bash ... 51.034 total

```

## Additional Information

    - Since configurator's ACL functionality was unused, this change is *not* backwards-breaking


docs/readmes/orc8r/dev_security.md
```diff
@@ -170,18 +170,24 @@ openssl s_client \
 Debug southbound interface (gRPC)
 
 ```bash
+# Emulate a full gateway bootstrap
+#
+# Places generated session secrets at /tmp/magma_protos/gateway.{key,crt}
+${MAGMA_ROOT}/orc8r/tools/scripts/bootstrap.bash YOUR_PROVISIONED_GATEWAY_HWID
```
karthiksubraveti, line 176: Nice!! I can add this to the orcl tool as a part of post deployment verification.

orc8r/cloud/docker/docker-compose.override.yml
```diff
@@ -33,5 +33,10 @@ services:
       POSTGRES_USER: magma_test
       POSTGRES_PASSWORD: magma_test
       POSTGRES_DB: magma_test
-    command: ["postgres", "-c", "log_statement=all"]
+    command:
```
karthiksubraveti, line 36: why ? former version seemed more readable imo :-)

orc8r/cloud/docker/docker-compose.yml
```diff
@@ -44,6 +44,9 @@ services:
 
   postgres:
     image: postgres
+    # Default is 64mb
```
karthiksubraveti, line 47: Why this change ? Is this the default in our production setup ?

orc8r/cloud/go/services/configurator/storage/sql.go
```diff
@@ -574,78 +549,57 @@ func (store *sqlConfiguratorStorage) UpdateEntity(networkID string, update Entit
...
 	// To make testing easier, we'll order the returned entities by TK
-	retEnts := funk.Map(internalGraph.entsByPk, func(_ string, ent *NetworkEntity) *NetworkEntity { return ent }).([]*NetworkEntity)
-	retRoots := funk.Map(rootPks, func(pk string) *EntityID { return &EntityID{Type: entTksByPk[pk].Type, Key: entTksByPk[pk].Key} }).([]*EntityID)
-	sort.Slice(retEnts, func(i, j int) bool {
-		return storage.IsTKLessThan(retEnts[i].GetTypeAndKey(), retEnts[j].GetTypeAndKey())
-	})
-	sort.Slice(retRoots, func(i, j int) bool {
-		return storage.IsTKLessThan(retRoots[i].ToTypeAndKey(), retRoots[j].ToTypeAndKey())
-	})
+	entsByPK := internalGraph.entsByTK.ByPK()
```
karthiksubraveti, line 598: Can this operation be pretty expensive for entity with large number of entity associations ? If it is not for correctness and just for testing, should we wrap this under a flag ?

orc8r/cloud/go/services/configurator/storage/sql_entity_graph_helpers.go
```diff
@@ -23,53 +23,54 @@ import (
...
 // component search on it, and relabel components if a partition is detected.
 // entToUpdateOut is an output parameter
-func (store *sqlConfiguratorStorage) fixGraph(networkID string, graphID string, entToUpdateOut *entWithPk) error {
+func (store *sqlConfiguratorStorage) fixGraph(networkID string, graphID string, entToUpdateOut *NetworkEntity) error {
 	internalGraph, err := store.loadGraphInternal(networkID, graphID, EntityLoadCriteria{})
 	if err != nil {
 		return errors.Wrap(err, "failed to load graph of updated entity")
 	}
-	if funk.IsEmpty(internalGraph.entsByPk) {
+	if funk.IsEmpty(internalGraph.entsByTK) {
```
karthiksubraveti, line 60: If this was just changing the attributes names, looks like there was a typo here.
entsByTK  -> entsByPK

orc8r/cloud/go/services/configurator/storage/sql.go
```diff
@@ -431,58 +394,67 @@ func (store *sqlConfiguratorStorage) UpdateNetworks(updates []NetworkUpdateCrite
...
+		}
+		for _, assoc := range assocs {
+			// Assoc may be from a not-loaded ent
+			e, ok := entsByTK[assoc.fromTK]
+			if ok {
+				e.Associations = append(e.Associations, assoc.getToID())
+			}
+		}
+		for _, ent := range entsByTK {
+			SortIDs(ent.Associations) // for deterministic return
```
karthiksubraveti, line 420: If we load this with an ORDER criteria won't this be deterministic ? Can we avoid this additional sorting here ?

orc8r/cloud/go/services/configurator/storage/sql.go
```diff
@@ -431,58 +394,67 @@ func (store *sqlConfiguratorStorage) UpdateNetworks(updates []NetworkUpdateCrite
...
+		}
+		for _, parentAssoc := range parentAssocs {
+			// Assoc may be to a not-loaded ent
+			e, ok := entsByTK[parentAssoc.toTK]
+			if ok {
+				e.ParentAssociations = append(e.ParentAssociations, parentAssoc.getFromID())
+			}
+		}
+		for _, ent := range entsByTK {
+			SortIDs(ent.ParentAssociations) // for deterministic return
```
karthiksubraveti, line 437: Same comment as before ?

orc8r/cloud/go/services/configurator/storage/sql.go
```diff
@@ -491,52 +463,55 @@ func (store *sqlConfiguratorStorage) CreateEntity(networkID string, entity Netwo
...
 	if err != nil {
 		return NetworkEntity{}, err
 	}
-	createdEntWithPk.GraphID = newGraphID
+	createdEnt.GraphID = newGraphID
 
 	// If we were given duplicate edges, get rid of those
-	if !funk.IsEmpty(createdEntWithPk.Associations) {
-		createdEntWithPk.Associations = funk.Chain(createdEntWithPk.Associations).
+	if funk.NotEmpty(createdEnt.Associations) {
```
karthiksubraveti, line 505: is this a common case or is this a workaround for handling buggy requests, Should we print warning when we see this ?

orc8r/cloud/go/services/configurator/storage/sql_entity_graph_helpers.go
```diff
@@ -23,53 +23,54 @@ import (
...
+// This function will NOT fill loadEntities with associations.
 func (store *sqlConfiguratorStorage) loadGraphInternal(networkID string, graphID string, criteria EntityLoadCriteria) (internalEntityGraph, error) {
 	loadFilter := EntityLoadFilter{GraphID: &wrappers.StringValue{Value: graphID}}
-	entsByPk, err := store.loadFromEntitiesTable(networkID, loadFilter, criteria)
+	ents, err := store.loadEntities(networkID, loadFilter, criteria)
 	if err != nil {
 		return internalEntityGraph{}, errors.Wrap(err, "failed to load entities for graph")
 	}
-	if funk.IsEmpty(entsByPk) {
+	if funk.IsEmpty(ents) {
```
karthiksubraveti, line 38: minor nit: just curious on why we use funk isEmpty instead of checking ents length ? I see this pattern being quite common across this codebase. When i looked at funk library, I see that it uses reflection to check appropriate object. Won't we be more efficient checking for length rather than using funk here ?

karthiksubraveti (commented):
Left a few initial comments.

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -27,147 +28,195 @@ import (
...
-	// [[ ORDER BY ent.key ASC LIMIT page_size ]]
-	selectBuilder := store.builder.Select(getLoadEntitiesColumns(criteria)...).
-		From(fmt.Sprintf("%s AS ent", entityTable))
+	// [[ JOIN (on cfg_assocs and cfg_entities to get child/parent assocs) ]]
+	// [[ WHERE ent.network_id = $network_filter AND ent.key = $key_filter AND ent.type = $type_filter AND ent.key > $page_token ]]
+	// ORDER BY ent.key
+	// LIMIT $page_size ;
+
+	// Helpers
+	entCol := func(c string) string {
```
karthiksubraveti, line 120: Older code seems to be ordering ASC, we aren't doing that now. Is this intentional ?

mpgermano: +1 for this question

hcgatewood: ASC is default, both for postgres and maria, so it's unnecessary

mpgermano (approved):
This looks good to me. Do we need to do any migration related to the removal of ACLs?

orc8r/cloud/go/services/configurator/storage/sql.go
mpgermano: This is necessary as it takes into account both type and key whereas the order only orders by key. For multi-type loads this ensures deterministic ordering

hcgatewood:
> This looks good to me. Do we need to do any migration related to the removal of ACLs?

No migration necessary for correctness. Also I don't think any ACLs were ever created, so it would just be a cleanup of the table names. I think it's fine to leave that out

orc8r/cloud/docker/docker-compose.override.yml
hcgatewood: Makes it easier to comment in/out, add additional statements, etc

orc8r/cloud/docker/docker-compose.yml
hcgatewood: In prod we use a managed db, so it's irrelevant. For dev, the db was crashing when I added the 15k subscribers because it was out of memory. This fixes by increasing the shared memory, allowing db to hold all subs

orc8r/cloud/go/services/configurator/storage/sql.go
hcgatewood: From offline: flag is good idea, will defer to subsequent PR though to minimize number of required changes to tests

orc8r/cloud/go/services/configurator/storage/sql_entity_graph_helpers.go
hcgatewood: - PK := primary key, a unique uuid generated specifically for this ent
- TK := type+key, a logical identifier with semantic meaning to Orc8r

orc8r/cloud/go/services/configurator/storage/sql.go
hcgatewood: We could probably move this to db, with some maneuvering. Will defer to subsequent PR / after we have scale test framework ready

orc8r/cloud/go/services/configurator/storage/sql.go
hcgatewood: Not a common case, might be worth printing a warning / handling this earlier in the fn

orc8r/cloud/go/services/configurator/storage/sql_entity_graph_helpers.go
hcgatewood: This is from Jacky. I went back and forth on this, but have tentatively landed on "IsEmpty" and "NotEmpty" are more readable. Once we have scale test framework we can benchmark how slow it is. If it's non-trivially slower, we can remove their usages

But I'm expecting it to be not too slow, especially since it's not in a tight loop. So tentatively in favor of the added readability

PR #5939, opened 2021-04-06
[DevExp] Migrate container to clang-format 7 for backwards compatibility

electronjoe:
## Test Plan

Loaded branch in GH Code Spaces, modify a file in `lte/gateway/c/oai/` with incorrect formatting for clang-format. Then confirmed that the following fixed-up ONLY that modified file and no others.

```
cd lte/gateway/
make build_oai
```


themarwhal (approved):
:D thanks!

tmdzk (approved):

PR #5995, opened 2021-04-07
[orc8r][nprobe] implement events collection

### LINKED ISSUE #4722: [X2][orc8r] Implement LI Event Collection, Storage, Retrieval

### LINKED ISSUE #4732: [X2][orc8r] Support configuration retrieval from intercept service

ymasmoudi:

## Summary

NProbe is a service that will provide LI capabilities. It will collect the required events,  format records and export them to a remote destination.
This PR implements the first part of this service providing the required events collect from fluentd.

## Test Plan

Unit test cases for event collection 
```
ymasmoudi-mbp:nprobe ymasmoudi$ go test ./...
?   	magma/lte/cloud/go/services/nprobe	[no test files]
ok  	magma/lte/cloud/go/services/nprobe/collector	(cached)
?   	magma/lte/cloud/go/services/nprobe/nprobe	[no test files]
?   	magma/lte/cloud/go/services/nprobe/nprobe_manager	[no test files]
ok  	magma/lte/cloud/go/services/nprobe/obsidian/handlers	4.746s
?   	magma/lte/cloud/go/services/nprobe/obsidian/models	[no test files]
```
Local Orc8r
```
nprobe stderr | E0407 10:41:00.736572     110 nprobe_manager.go:113] Process task for targetID IMSI001010000000001
nprobe stderr | E0407 10:41:00.744917     110 nprobe_manager.go:91] Collected 15 events for targetID IMSI001010000000001
```


lte/cloud/go/services/nprobe/collector/collector.go
```diff
@@ -0,0 +1,156 @@
...
+		tags:      tags,
+		timestamp: timestamp,
+	}
+}
+
+// EventsCollector wraps an elastic client, query multiple streams per
+// subscriber and retrieve all events since last timestamp
+type EventsCollector struct {
+	ElasticClient *elastic.Client
+}
```
andreilee: This functionality should sit in `eventd`, and be exposed through a gRPC interface

lte/cloud/go/services/nprobe/const.go
```diff
@@ -0,0 +1,45 @@
...
+	ESStreamSessionD = "sessiond"
+
+	AttachSuccess        = "attach_success"
+	DetachSuccess        = "detach_success"
+	SessionCreated       = "session_created"
+	SessionUpdated       = "session_updated"
+	SessionTerminated    = "session_terminated"
+	S1SetupSuccess       = "s1_setup_success"
+	SessionCreateFailure = "session_create_failure"
+)
```
andreilee, line 27: Put this stuff in a `nprobe.yml`

andreilee (dismissed):
**Edit (04/13/21):**
Reduce the code duplication by keeping a lot of this functionality under `orc8r/cloud/go/services/eventd`. Expose functionality through something a package that can be imported by `nprobe`. See the comment by @mpgermano 

**Old:**
> A lot of this responsibility should sit in the `eventd` orc8r service, and be exposed through gRPC

lte/cloud/go/services/nprobe/collector/collector.go
mpgermano: I think my main concern here is that we're duplicating a decent amount of code from eventd. I think we should look at more effectively exposing a client and the necessary struct types into a package (`e.g. eventd_client`) in eventd and then import those so that we can avoid this duplication.

I'm not sure that it's worthwhile to have both HTTP and gRPC implementations in eventd that wrap elasticsearch querying. As a similar example, if an orchestrator service (i.e. `analytics`) needs to query metrics, it makes more sense for that service just to use a a prometheus client, rather than adding a gRPC servicer to metricsd which would wrap the prometheus client. I feel similarly here.

andreilee: Okay, let's go with that

mpgermano (approved):

lte/cloud/configs/nprobe.yml
```diff
@@ -9,3 +9,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+updateIntervalSecs: 60
+backoffIntervalSecs: 360
```
andreilee: - Switch the names of these to snake_case to match conventions of other `yml` files
- Add some documentation comments on what these intervals are used for

andreilee (approved):

PR #6002, opened 2021-04-07
[NMS] Fix gateway HA pool config

HannaFar:

## Summary

Hide gateway id option if already associated with other gateway pool.
Add gateway primary/secondary id select placeholder.

Capture d’écran 2021-04-14 à 14 57 21

Capture d’écran 2021-04-14 à 14 57 31

## Additional Information

### DIFF
--- a/nms/app/packages/magmalte/app/views/equipment/GatewayPoolGatewaysEdit.js
+++ b/nms/app/packages/magmalte/app/views/equipment/GatewayPoolGatewaysEdit.js
@@ -13,7 +13,6 @@
  * @flow strict-local
  * @format
  */
-/*[object Object]*/
 import type {GatewayPoolEditProps} from './GatewayPoolEdit';
 
 import AddCircleOutline from '@material-ui/icons/AddCircleOutline';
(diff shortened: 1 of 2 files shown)


nms/app/packages/magmalte/app/views/equipment/GatewayPoolGatewaysEdit.js
```diff
@@ -49,22 +49,36 @@ export default function GatewayEdit(props: GatewayPoolEditProps) {
...
+  // gateways already associated to gateway pools
+  const [gwPoolIds, setGwPoolIds] = useState(
+    Object.keys(ctx.state)
+      .map(poolId => {
+        return ctx.state[poolId].gatewayPoolRecords.map(
+          record => record.gateway_id,
+        );
+      })
+      .flat(),
+  );
```
andreilee: Thanks for adding the comment here, it's helpful. Please rename this variable though, it's not a `gateway pool id`, it's a `gateway id for gateway in a pool`
```suggestion
  // gateways already associated to gateway pools
  const [pooledGwIds, setPooledGwds] = useState(
    Object.keys(ctx.state)
      .map(poolId => {
        return ctx.state[poolId].gatewayPoolRecords.map(
          record => record.gateway_id,
        );
      })
      .flat(),
  );
```

nms/app/packages/magmalte/app/views/equipment/GatewayPoolGatewaysEdit.js
```diff
@@ -133,9 +150,9 @@ export default function GatewayEdit(props: GatewayPoolEditProps) {
                     variant={'outlined'}
                     displayEmpty={true}
                     value={gw.gateway_id}
-                    onChange={({target}) =>
-                      handleGwIdChange(target.value, index)
-                    }
+                    onChange={({target}) => {
+                      handleGwIdChange(target.value, index, gw.gateway_id);
+                    }}
```
andreilee: For this dropdown menu, add a placeholder value of `Unpooled Gateway ID` or other appropriate value to communicate that only gatewayIDs that aren't part of a gateway pool can be selected.

Also, only display the selectable gateway IDs

andreilee (dismissed):

nms/app/packages/magmalte/app/views/equipment/GatewayPoolGatewaysEdit.js
HannaFar: Not sure I understand 😄 
Right now I'm displaying all the gateway ids, but if a gateway id is associated to another gateway pool the option is disabled. So the only gateway IDs that the user can select are gateway IDs that aren't part of a gateway pool.

andreilee: So, rather than disabling some of the options, just not showing them.

https://user-images.githubusercontent.com/804385/114628666-1523c980-9c6c-11eb-820d-3a798114d1a8.mov

So what I mean by 'placeholder value' would correspond to 'None' in the small recording above.

andreilee (approved):

PR #6008, opened 2021-04-07
[lte][subscriberdb] Add subscriberdb client to to gateway for paginated streaming support

mpgermano:

## Summary

This PR adds support for paginated streaming of subscribers from the orchestrator to the gateway. This PR
is intended to fix problems we have encountered with streaming all subscribers registered in a network in a
single RPC call. The fundamental functionality of the subscriber process is unchanged.

The main idea is that we will continuously loop and call the ListSubscribers endpoint until we
receive an empty page token, which signifies that we have received that last page of subscribers. We then use the aggregated list of subscribers to perform the existing operations on the subscriberdb store.

There also is new configuration to allow for modification of the page size requested as well as the frequency
that streaming is completed. Previously this was set to 60s.

## Test Plan

Added unit tests.
Tested locally with gateway and orchestrator


lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -0,0 +1,162 @@
...
+    DEFAULT_SUBSCRIBERDB_SYNC_INTERVAL = 180
+    DEFAULT_SUBSCRIBER_PAGE_SIZE = 250
+    SUBSCRIBERDB_REQUEST_TIMEOUT = 10
+
+    def __init__(self,
+                 service: MagmaService,
+                 store: SqliteStore,
+                 grpc_client_manager: GRPCClientManager):
+
+        sync_interval = service.config.get(
```
ardzoht: Let's do the config retrieval on main function and pass it on the constructor instead of the whole service param

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -0,0 +1,162 @@
...
+            'subscriberdb_sync_interval',
+            self.DEFAULT_SUBSCRIBERDB_SYNC_INTERVAL)
+
+        super().__init__(
+            sync_interval,
+            service.loop
```
ardzoht: same for this, let's just pass the loop property as param for the constructor

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -0,0 +1,162 @@
...
+        # cloud streams down subscriber IDs without 'IMSI' prepended,
+        # subscriberdb will try to delete all of the subscribers from MME every
+        # time it streams from cloud because the set membership will fail
+        # when comparing '12345' to 'IMSI12345'.
+        new_sub_ids = set(
+            map(
+                lambda s: 'IMSI' + s if not s.startswith('IMSI') else s,
+                new_sub_ids,
+            ),
+        )
```
ardzoht:
```suggestion
        new_sub_ids = set(
            'IMSI' + s if not s.startswith('IMSI') else s for s in new_sub_ids
        )
```

ardzoht (changes requested):

lte/gateway/python/magma/subscriberdb/main.py
```diff
@@ -55,11 +56,20 @@ def main():
 
     # Start a background thread to stream updates from the cloud
     if service.config['enable_streaming']:
-        callback = SubscriberDBStreamerCallback(store, service.loop)
-        stream = StreamerClient({"subscriberdb": callback}, service.loop)
-        stream.start()
+        grpc_client_manager = GRPCClientManager(
+            service_name="subscriberdb",
+            service_stub=SubscriberDBCloudStub,
+            max_client_reuse=60,
```
karthiksubraveti, line 65: If it prevents bugs, we can set it to 0, if we are  getting subscriber state once in 3 minutes getting a new connection doesn't sound expensive.

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -0,0 +1,162 @@
...
+
+        # grpc_client_manager to manage grpc client recycling
+        self._grpc_client_manager = grpc_client_manager
+
+    async def _run(self) -> None:
+        subscribers = await self._get_subscribers()
+        if subscribers is not None:
+            self._process_subscribers(subscribers)
+
+    async def _get_subscribers(self) -> [SubscriberData]:
```
karthiksubraveti, line 84: I wonder if we can add some metrics around this piece including total time to sync etc which we might want to use later in case we want more visibility around this one.

karthiksubraveti (commented):
Most changes LGTM. Thanks for making these changes.

ulaskozat (approved):
LGTM

mpgermano:
1. Address comments
2. Add latency and request success/failure metrics

Still working on fixing the linting

ardzoht (approved):

lte/gateway/configs/subscriberdb.yml
```diff
@@ -16,6 +16,12 @@
 # Append GRPC content to the log
 print_grpc_payload: False
 
+# Size of subscriber pages returned from cloud
+subscriber_page_size: 250
```
hcgatewood: I feel like we can bump this up to at least 3k. I'd be in favor of even 5k

For 15k subscribers, 250=>60 round trips. While 3k=>5 round trips

lte/gateway/python/magma/subscriberdb/main.py
```diff
@@ -13,19 +13,23 @@
...
 from .protocols.diameter.server import S6aServer
 from .rpc_servicer import SubscriberDBRpcServicer
 from .subscription_profile import get_default_sub_profile
-from .streamer_callback import SubscriberDBStreamerCallback
 from .store.sqlite import SqliteStore
 from .protocols.s6a_proxy_servicer import S6aProxyRpcServicer
 from lte.protos.mconfig import mconfigs_pb2
+from lte.protos.subscriberdb_pb2_grpc import SubscriberDBCloudStub
+
+DEFAULT_SUBSCRIBERDB_SYNC_INTERVAL = 180
```
hcgatewood: I don't like having config defaults. If it's unavoidable then fine, but I think hard-coding defaults makes things brittle and causes unexpected values. Instead, we can just provide the default value through the yml file

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -0,0 +1,167 @@
...
+
+class SubscriberDBCloudClient(SDWatchdogTask):
+    """
+    SubscriberDBCloudClient for requesting subscribers from Orchestrator.
+
+    SubscriberDBCloudClient calls the Orchestrator's SubscriberDBCloud service
+    to fetch pages of subscribers stored in Orchestrator and updates the local
+    subscriberdb store with the changes.
+    """
+    DEFAULT_SUBSCRIBERDB_SYNC_INTERVAL = 180
```
hcgatewood: Aside from not liking defaults: looks like we're defining these defaults twice -- once in main and once in client

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -0,0 +1,167 @@
...
+            loop
+        )
+        self._loop = loop
+        self._subscriber_page_size = subscriber_page_size
+        self._store = store
+
+        # grpc_client_manager to manage grpc client recycling
+        self._grpc_client_manager = grpc_client_manager
+
+    async def _run(self) -> None:
```
hcgatewood, line 79: Where are we enforcing the sync_interval ?

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -0,0 +1,167 @@
...
+                              err.details())
+                time_elapsed = datetime.datetime.now() - sync_start
+                SUBSCRIBER_SYNC_LATENCY.observe(
+                    time_elapsed.total_seconds() * 1000)
+                SUBSCRIBER_SYNC_FAILURE_TOTAL.inc()
+                return None
+
+        return subscribers
+
+    def _process_subscribers(self, subscribers: SubscriberData) -> None:
```
hcgatewood, line 128: Maybe `_install_subscribers`, either is fine

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -0,0 +1,167 @@
...
+        Compares current subscriber ids and new subscriber ids list
+        just streamed from the cloud to figure out the deleted subscribers.
+        Then send grpc DeleteSubscriber request to mme to detach all the
+        deleted subscribers.
+        :param old_sub_ids: a list of old subscriber ids in the store.
+        :param new_sub_ids: a list of new active subscriber ids
+                just streamed from the cloud
+        :return: n/a
+        """
+        # THIS IS A HACK UNTIL WE FIX THIS ON CLOUD
```
hcgatewood, line 159: I think this is fixed now on cloud? Correct me if I'm wrong

https://github.com/magma/magma/blob/95ded6cf88eb6dbd55bfe383cdb98e8fcd7bf6a0/lte/cloud/go/services/policydb/obsidian/models/swagger.v1.yml#L842-L846

hcgatewood (changes requested):
Couple things, mostly concerned about `subscriber_page_size`

lte/gateway/python/magma/subscriberdb/client.py
mpgermano: It's enforced via SDWatchdog and initialized from `super().__init__(
            sync_interval,`

lte/gateway/python/magma/subscriberdb/client.py
mpgermano: Any chance we have legacy subscribers that would still be streamed without this?

lte/gateway/python/magma/subscriberdb/client.py
mpgermano: Leaving as `process` as we delete some and install others

hcgatewood (approved):

PR #6009, opened 2021-04-07
[Gateway] Fix gateway install script so it can deploy specific version of AGW

119Vik:
## Summary
To deploy a specific version of FeG  one should define  **IMAGE_VERSION** in **.env** file as 
```
IMAGE_VERSION="<image-tag>|<magma-tag>"
```
**install_gateway.sh** partially handles this trick but it doe not share any results of **IMAGE_VERSION** parsing with docker-compose or **recreate_services.sh**. So both are failing cause they read the original **IMAGE_VERSION** from **.env**.
Everything works well with default values (which are the magma master branch and the latest image tag) and does not work if we try to redeploy non latest version.

## Test Plan

Deploy FeG with default and version-specific values.

### DIFF
--- a/orc8r/tools/docker/install_gateway.sh
+++ b/orc8r/tools/docker/install_gateway.sh
@@ -13,13 +13,15 @@
 
 # This script is intended to install a docker-based gateway deployment
 
+# Both ENV vars are moved in to .env file:
+# GIT_HASH="v1.3.3"
+# IMAGE_VERSION="docker-tag-1.3.3"
 set -e
 
 CWAG="cwag"
 FEG="feg"
 XWF="xwf"
 INSTALL_DIR="/tmp/magmagw_install"
-GIT_HASH="master"
 
 # TODO: Update docker-compose to stable version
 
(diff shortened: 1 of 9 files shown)


orc8r/tools/docker/install_gateway.sh
```diff
@@ -13,13 +13,15 @@
 
 # This script is intended to install a docker-based gateway deployment
 
+# Both ENV vars are moved in to .env file:
+# GIT_HASH="v1.3.3"
```
themarwhal, line 17: Does GIT_HASH not contain a hash value? Is it more of a release tag name?

tmdzk (approved):

orc8r/tools/docker/install_gateway.sh
119Vik: This env var is used later in the script as 
` git -C $INSTALL_DIR/magma checkout "$GIT_HASH"`
so it's basically anything you can git checkout :
- commit ID
- tag
- branch

themarwhal (approved):

PR #6011, opened 2021-04-07
[DevExp] Add dockerfile health tester

electronjoe:
## Summary

Once this GH Action has been in place for awhile, we can turn up `Required` status for the following and prevent Dockerfile breakages such as those I committed in #5939.

- build_ubuntu20_04_dockerfile
- build_devcontainer_dockerfile

## Take Note of Caching Change

I am attempting to change the caching behavior for dockerx builds in GH.  Please take a look. The intent here is to re-use the buildx cache so long as the dockerfiles repo-wide are unchanged.  We could additionally insert date strings here to force rollover on some weekly cadence...

```yaml
        with:
          path: /tmp/.buildx-cache
          key: ${{ runner.os }}-buildx-${{ hashFiles( '**/Dockerfile*' ) }}
```

## Test

Cherry-picked my master-breaking PR #5939 and observed failure to build one of our two docker containers tested in this PR.

Screen Shot 2021-04-08 at 7 40 07 AM

### DIFF
--- a/.github/workflows/dockerfile-health.yml
+++ b/.github/workflows/dockerfile-health.yml
@@ -1,0 +1,99 @@
+---
+name: "Dockerfile Health"
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+    paths:
+      - .devcontainer/Dockerfile
+      - .github/workflows/dockerfile-health.yml
+      - lte/gateway/docker/mme/Dockerfile.ubuntu20.04
+
+jobs:
+  build_ubuntu20_04_dockerfile:
+    env:
+      DOCKERFILE: lte/gateway/docker/mme/Dockerfile.ubuntu20.04
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Check Out Repo
+        uses: actions/checkout@v2
+      -
+        name: Docker meta
+        id: meta
+        uses: crazy-max/ghaction-docker-meta@v2
+        with:
+          # list of Docker images to use as base name for tags
+          images: |
+            magma/mme_builder_focal
+          # generate Docker tags based on the following events/attributes
+          tags: |
+            type=sha
+      -
+        name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      -
+        name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ hashFiles( '**/Dockerfile*' ) }}
+      -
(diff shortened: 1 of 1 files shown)


.github/workflows/dockerfile-health.yml
```diff
@@ -0,0 +1,75 @@
...
+          key: ${{ runner.os }}-buildx-${{ hashFiles( '**/Dockerfile*' ) }}
+      -
+        name: Docker Build
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ${{ env.DOCKERFILE }}
+          push: false
+          tags: magma/mme_builder:latest
```
tmdzk: I know it's only a docker build test, but as a good practise let's double tag with latest and version (git-hash0:8)

tmdzk (commented):
Is there a way to bundle it into a function? both jobs have exactly the same steps, just different builld env/dockerfile.

.github/workflows/dockerfile-health.yml
electronjoe: I'll explore function support - I did a brief google-fu earlier and didn't find but it must support better re-use. I'll look harder this time =D

themarwhal (approved):
lgtm! :)

.github/workflows/dockerfile-health.yml
```diff
@@ -0,0 +1,75 @@
...
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+    paths:
+      - .devcontainer/Dockerfile
+      - .github/workflows/dockerfile-health.yml
+      - lte/gateway/docker/mme/Dockerfile.ubuntu20.04
```
themarwhal, line 12: Should we do the same for the ubuntu18? dockerfile? Or even better, can we converge to the 20.04 one?

electronjoe: I was going to get this in place and confirm it's good - and then ask @rdefosse if he wants me to cover the ubuntu 18 dockerfile. I suspect he already has coverage over in the MAGMA-OAI-MME land?  As to convergence, that would depend upon the OAI use cases for the Dockerfile.ubuntu18.04 builds.

.github/workflows/dockerfile-health.yml
electronjoe: Huh. Not great. Appears my early google-fu was correct and GH Actions don't yet support any form of re-use?
https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851

tmdzk: Should we start bundling them in our own Github Action marketplace "orbs"?

PR #6015, opened 2021-04-07
[lte][agw] Bug fix in handling ICS failure

ulaskozat:
## Summary

ICS failure when UE is not in registered state should lead to an implicit detach procedure. The current code was instead checking if the UE was in deregistered state leading to stale entries for UEs that have no pending timers and are between deregistered and registered states.

## Test Plan

Integ tests.
Spirent loading tests that induce ICS failures and create stale entries.

## Additional Information

### DIFF
--- a/lte/gateway/c/oai/tasks/mme_app/mme_app_bearer.c
+++ b/lte/gateway/c/oai/tasks/mme_app/mme_app_bearer.c
@@ -1970,7 +1970,7 @@ void mme_app_handle_initial_context_setup_failure(
      * the UE context.
      */
     ue_context_p->ue_context_rel_cause = S1AP_INITIAL_CONTEXT_SETUP_FAILED;
-    if (ue_context_p->mm_state == UE_UNREGISTERED) {
+    if (ue_context_p->mm_state != UE_REGISTERED) {
       // Initiate Implicit Detach for the UE
       nas_proc_implicit_detach_ue_ind(ue_context_p->mme_ue_s1ap_id);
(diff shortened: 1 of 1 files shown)


ardzoht (approved):

ulaskozat:
RestartIntegTests

lte/gateway/c/oai/tasks/mme_app/mme_app_bearer.c
```diff
@@ -1970,7 +1970,7 @@ void mme_app_handle_initial_context_setup_failure(
      * the UE context.
      */
     ue_context_p->ue_context_rel_cause = S1AP_INITIAL_CONTEXT_SETUP_FAILED;
-    if (ue_context_p->mm_state == UE_UNREGISTERED) {
+    if (ue_context_p->mm_state != UE_REGISTERED) {
```
rsarwad, line 1973: mm_state_t used within structure, ue_mm_context_s has only two states. i.e as defined below
 79 
 typedef enum {
 80   UE_UNREGISTERED = 0,
 81   UE_REGISTERED,
 82 } mm_state_t;
 
 Both checks results into same results.

ulaskozat: Ahh good point, I was confusing this with the FSM state.

ulaskozat:
I am closing this PR as I was confusing the FSM states with the EMM state. PR does not change the effective behavior.

PR #6016, opened 2021-04-08
[Pipelined]Fix IPv6 Restart Flow Recovery

koolzz:

## Summary

Discovered that we don't properly recover ipv6 flows. This pr fixes that

## Additional Information


pshelar:
I guess there is no test to validate IPv6 recovery, can we add it?

koolzz:
> I guess there is no test to validate IPv6 recovery, can we add it?

Added

koolzz:
> I guess there is no test to validate IPv6 recovery, can we add it?

added

pshelar (approved):
Thanks!

PR #6017, opened 2021-04-08
[orc8r] Adding api to provide total number of subscribers

### LINKED ISSUE #5892: [Orc8r] Additional api endpoints for supporting subscriber paging
### Your Environment

- **Version:**
- **Affected Component:** Orchestrator, Access Gateway, or NMS

### Describe the Issue

Need for 3 additional API endpoints to support subscriber paging
- subscriber state paging endpoint
- total subscriber count endpoint
- total subscriber state count endpoint

karthiksubraveti:
## Summary
Added total count attribute to existing subscriber v2 api

## Test Plan
Changed existing test to check for total count. 

## Additional Information


lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -210,6 +210,19 @@ func listSubscribersV2Handler(c echo.Context) error {
...
+	loadCriteria := configurator.EntityLoadCriteria{}
+	count, err := configurator.CountAllEntitiesOfType(
+		networkID,
+		lte.SubscriberEntityType,
+		loadCriteria,
+		serdes.Entity)
+	if err != nil {
+		return c.JSON(http.StatusOK, nil)
+	}
+	totalCount := int64(count)
```
andreilee:
```suggestion
	totalCount := swag.int64(count)
```

orc8r/cloud/go/services/configurator/client_api.go
```diff
@@ -688,6 +688,28 @@ func LoadAllEntitiesOfType(networkID string, entityType string, criteria EntityL
 	return ret, res.NextPageToken, nil
 }
 
+// CountAllEntitiesOfType provides total count of entities of this type
+func CountAllEntitiesOfType(networkID string, entityType string, criteria EntityLoadCriteria, serdes serde.Registry) (uint64, error) {
```
hcgatewood: "CountEntitiesOfType"

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -210,6 +210,17 @@ func listSubscribersV2Handler(c echo.Context) error {
 
 	var pageSize uint64 = 0
 	var err error
+
+	// get total number of subscribers
+	loadCriteria := configurator.EntityLoadCriteria{}
+	count, err := configurator.CountAllEntitiesOfType(
```
hcgatewood: Let's not make a configurator call here, since we may be making an early return if ParamMSISDN or ParamIP are set. Move the configurator call below those

lte/cloud/go/services/subscriberdb/obsidian/models/swagger.v1.yml
```diff
@@ -373,7 +373,13 @@ definitions:
     required:
       - next_page_token
       - subscribers
+      - total_count
```
hcgatewood, line 376: I'm fine with the way we're doing this count (making it a separate call to configurator) for the sake of simplicity of implementation. That said, if we do keep it as a separate configurator call, this field needs to be `estimated_total_count`, to indicate that it's not a reliable count of exactly how many subscribers there are

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -59,6 +59,55 @@ func (store *sqlConfiguratorStorage) loadFromEntitiesTable(networkID string, fil
 	return entsByPk, nil
 }
 
+func (store *sqlConfiguratorStorage) countEntitiesTable(networkID string, filter EntityLoadFilter, criteria EntityLoadCriteria) (uint64, error) {
```
hcgatewood: "countEntities"

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -59,6 +59,55 @@ func (store *sqlConfiguratorStorage) loadFromEntitiesTable(networkID string, fil
...
+		return 0, err
+	}
+	var count uint64
+	if err = selectBuilder.RunWith(store.tx).Scan(&count); err != nil {
+		return 0, err
+	}
+	return count, nil
+}
+
+func (store *sqlConfiguratorStorage) getEntitiesCountBuilder(networkID string, filter EntityLoadFilter, criteria EntityLoadCriteria) (sq.SelectBuilder, error) {
```
hcgatewood: This is duplicating a lot of logic. It's basically the same as the normal builder, but with "select count" instead of "select rows". I think there's a way to work those together

That said, wait for this PR until we get the configurator changes merged. Cause it will change things enough that you'll need to rework just a bit

hcgatewood (changes requested):
Let's wait for https://github.com/magma/magma/pull/5928 to merge this -- so we can reuse the query builder, just with "select count" instead"

lte/cloud/go/services/subscriberdb/obsidian/models/swagger.v1.yml
karthiksubraveti: I don't think we need to be extremely precise in naming here. Calling this estimated_total_count maybe more confusing rather than clarifying the implementation detail involved here. Since it is more an informational field in UI, I would prefer ignoring this detail and just calling this total_count.

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -129,7 +142,12 @@ func (store *sqlConfiguratorStorage) getBuilder(networkID string, filter EntityL
...
-	cols := getLoadEntitiesCols(criteria)
+	var cols []string
+	if loadTyp == countEntities {
+		cols = []string{"COUNT(1)"}
+	} else {
+		cols = getLoadEntitiesCols(criteria)
+	}
 	// If we're loading assocs: don't need to load everything
 	if loadTyp == loadChildren || loadTyp == loadParents {
 		cols = getLoadAssocCols()
```
hcgatewood: Make this one big switch on loadTyp -- right now it's two separate if statements

lte/cloud/go/services/subscriberdb/obsidian/models/swagger.v1.yml
```diff
@@ -373,7 +373,13 @@ definitions:
...
     properties:
+      total_count:
+        type: integer
+        format: int64
+        description: total number of subscriber entries
```
hcgatewood: Fine with leaving it as `total_count`, but then let's note the fact that it's an estimation, or best-effort count, in the description

lte/cloud/go/services/subscriberdb/obsidian/models/swagger.v1.yml
```diff
@@ -373,7 +373,13 @@ definitions:
     required:
       - next_page_token
       - subscribers
+      - total_count
     properties:
+      total_count:
```
hcgatewood, line 378: Since total_count is a required property, make it `x-nullable: false`

hcgatewood (approved):
Lgtm, please handle 3x notes before merging

PR #6025, opened 2021-04-08
[5G: MVC][component: agw][type: enhancement] Periodic registration

hkshiv1:
## Summary

Handling Periodic registration updating request from UE. Handled the request and sends accept message with new GUTI

## Test Plan

5G integration tests passed, Periodic_Registration_Update.xml used in DS-tester. 

## Additional Information

### DIFF
--- a/lte/gateway/c/oai/tasks/amf/amf_recv.cpp
+++ b/lte/gateway/c/oai/tasks/amf/amf_recv.cpp
@@ -77,7 +77,7 @@ int amf_handle_registration_request(
 
   ue_m5gmm_context_s* ue_context = amf_ue_context_exists_amf_ue_ngap_id(ue_id);
   if (ue_context == NULL) {
-     OAILOG_INFO(LOG_AMF_APP, "AMF_TEST: ue_context is NULL\n");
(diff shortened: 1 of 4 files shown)


lte/gateway/c/oai/tasks/amf/amf_recv.cpp
```diff
@@ -71,51 +74,143 @@ int amf_handle_registration_request(
...
-        MOBILE_IDENTITY_PROTECTION_SCHEME_NULL) {
-      /*
-       * Extract the SUPI or IMSI from SUCI as scheme output is not encrypted
-       */
-      params->imsi = new (imsi_t);
+
+  ue_m5gmm_context_s* ue_context = amf_ue_context_exists_amf_ue_ngap_id(ue_id);
+  if (ue_context == NULL) {
+    OAILOG_INFO(LOG_AMF_APP, "AMF_TEST: ue_context is NULL\n");
+  }
```
ulaskozat, line 81: This does not return an error and if null, it will lead to segmentation fault in the other sections of this function as you are accessing ue_context.

PR #6035, opened 2021-04-08
[SessionD] Remove non-versioned policies from Activate/Deactivate req…

themarwhal:
…uests

## Summary
- Remove the old policy fields in Activate/DeactivateFlows request as we have already transitioned to the version policy fields on PipelineD's end. (https://github.com/magma/magma/pull/5689)

## Test Plan
make precommit_sm
lte integ test

## Additional Information


themarwhal:
cc @koolzz 
After this lands, we can mark the proto fields as deprecated?

koolzz:
> cc @koolzz
> After this lands, we can mark the proto fields as deprecated?

actually, after this lands that is https://github.com/magma/magma/pull/6037

uri200 (approved):

PR #6039, opened 2021-04-08
[NMS] Add subscriber state view

HannaFar:
## Summary
Add subscriber state view

Capture d’écran 2021-04-13 à 17 32 10

Capture d’écran 2021-04-13 à 17 32 21

## Additional Information

### DIFF
--- a/nms/app/packages/magmalte/app/state/lte/SubscriberState.js
+++ b/nms/app/packages/magmalte/app/state/lte/SubscriberState.js
@@ -26,9 +26,11 @@ import type {
 
 import MagmaV1API from '@fbcnms/magma-api/client/WebClient';
 
-import {getLabelUnit} from '../../views/subscriber/SubscriberUtils';
-
-const MAX_PAGE_ROW_COUNT = 1000;
+import {
+  DEFAULT_PAGE_SIZE,
+  MAX_PAGE_ROW_COUNT,
+  getLabelUnit,
+} from '../../views/subscriber/SubscriberUtils';
 
 type FetchProps = {
   enqueueSnackbar?: (msg: string, cfg: {}) => ?(string | number),
@@ -63,7 +65,7 @@ export async function FetchSubscribers(props: FetchProps) {
     try {
       return await MagmaV1API.getLteByNetworkIdSubscribersV2({
         networkId,
-        pageSize: pageSize ?? 10,
+        pageSize: pageSize ?? DEFAULT_PAGE_SIZE,
         pageToken: token ?? '',
       });
     } catch (e) {
@@ -147,9 +149,12 @@ export default async function InitSubscriberState(
     setSessionState,
     enqueueSnackbar,
   } = props;
-  const subscribers = await FetchSubscribers({networkId, enqueueSnackbar});
-  if (subscribers) {
-    setSubscriberMap(subscribers.subscribers);
+  const susbcriberResponse = await FetchSubscribers({
+    networkId,
+    enqueueSnackbar,
+  });
+  if (susbcriberResponse) {
+    setSubscriberMap(susbcriberResponse.subscribers);
   }
 
   const state = await FetchSubscriberState({networkId, enqueueSnackbar});
@@ -175,10 +180,7 @@ type SubscriberStateProps = {
   setSessionState: ({[string]: subscriber_state}) => void,
   key: string,
   value?: mutable_subscriber,
-  newState?: {
-    state: {[string]: subscriber},
-    sessionState: {[string]: subscriber_state},
-  },
+  newState?: {[string]: subscriber},
 };
 
 export async function setSubscriberState(props: SubscriberStateProps) {
@@ -186,14 +188,12 @@ export async function setSubscriberState(props: SubscriberStateProps) {
     networkId,
     subscriberMap,
     setSubscriberMap,
-    setSessionState,
     key,
     value,
     newState,
   } = props;
   if (newState) {
-    setSessionState(newState.sessionState);
-    setSubscriberMap(newState.state);
+    setSubscriberMap(newState);
     return;
   }
   if (value != null) {
@@ -229,11 +229,13 @@ export async function setSubscriberState(props: SubscriberStateProps) {
   }
 }
 
-export function getSubscriberGatewayMap(subscribers: {[string]: subscriber}) {
+export function getGatewaySubscriberMap(sessions: {
+  [string]: subscriber_state,
+}) {
   const gatewayMap = {};
-  Object.keys(subscribers).forEach(id => {
-    const subscriber = subscribers[id];
-    const gwHardwareId = subscriber?.state?.directory?.location_history?.[0];
+  Object.keys(sessions).forEach(id => {
+    const subscriber = sessions[id];
+    const gwHardwareId = subscriber?.directory?.location_history?.[0];
     if (
       gwHardwareId !== null &&
       gwHardwareId !== undefined &&
@@ -299,10 +301,7 @@ export async function handleSubscriberQuery(props: SubscriberQueryType) {
         }
         setTokenList([...newTokenList]);
         // set subscriber state with current subscriber rows.
-        ctx.setState?.('', undefined, {
-          state: subscriberResponse.subscribers,
-          sessionState: {},
-        });
+        ctx.setState?.('', undefined, subscriberResponse.subscribers);
       }
       const tableData: Array<SubscriberRowType> = subscriberResponse
         ? Object.keys(subscriberResponse.subscribers).map((imsi: string) => {
--- a/nms/app/packages/magmalte/app/views/subscriber/SubscriberOverview.js
+++ b/nms/app/packages/magmalte/app/views/subscriber/SubscriberOverview.js
@@ -13,17 +13,7 @@
  * @flow strict-local
  * @format
  */
-import type {ActionQuery} from '../../components/ActionTable';
-import type {WithAlert} from '@fbcnms/ui/components/Alert/withAlert';
-import type {
-  mutable_subscriber,
-  subscriber,
-  subscriber_state,
-} from '@fbcnms/magma-api';
-
-import ActionTable from '../../components/ActionTable';
 import AddSubscriberButton from './SubscriberAddDialog';
-import CardTitleRow from '../../components/layout/CardTitleRow';
 import Dialog from '@material-ui/core/Dialog';
 import DialogContent from '@material-ui/core/DialogContent';
 import DialogTitle from '@material-ui/core/DialogTitle';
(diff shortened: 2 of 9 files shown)


nms/app/packages/magmalte/app/state/lte/SubscriberState.js
```diff
@@ -140,7 +149,7 @@ export default async function InitSubscriberState(
   } = props;
   const subscribers = await FetchSubscribers({networkId, enqueueSnackbar});
```
karthiksubraveti: minor nit: let's rename this as susbcriberResponse

nms/app/packages/magmalte/app/views/subscriber/SubscriberStateTable.js
```diff
@@ -0,0 +1,268 @@
...
+  return (
+    <>
+      <div>
+        <JsonDialog open={jsonDialog} onClose={onClose} imsi={currRow.imsi} />
+        <ActionTable
+          data={tableData}
+          columns={tableColumns}
+          handleCurrRow={(row: SubscriberRowType) => setCurrRow(row)}
+          menuItems={
+            networkCtx.networkType === FEG_LTE
```
karthiksubraveti: You can remove this FEG_LTE check in this table and keep only the View JSON option in this table. Meanwhile, the View, Edit and Remove options can be present in the Subscriber table.

nms/app/packages/magmalte/app/views/subscriber/SubscriberOverview.js
```diff
@@ -254,223 +213,46 @@ function SubscriberInternal(props: WithAlert) {
...
+          value={currTabIndex}
+          onChange={(_, newIndex: number) => setCurrTabIndex(newIndex)}
+          variant="fullWidth">
+          {subscribersTabList.map((k: string, idx: number) => {
+            return (
+              <SubscriberTab key={idx} label={k} className={classes.tab} />
+            );
+          })}
+        </SubscriberTabs>
+        {currTabIndex === 0 && <Subscriber columns={tableColumns} />}
```
karthiksubraveti: Can you add check here to display subscriber table only for LTE networks and not for FEG_LTE networks

nms/app/packages/magmalte/app/views/subscriber/SubscriberOverview.js
```diff
@@ -114,81 +108,56 @@ type SubscriberRowType = {
...
   const networkCtx = useContext(NetworkContext);
-  // $FlowIgnore
-  const subscriberMap: {[string]: subscriber} = state.state;
-  // $FlowIgnore
-  const sessionState: {[string]: subscriber_state} = state.sessionState;
-  const subscriberMetrics = ctx.metrics;
-  const [jsonDialog, setJsonDialog] = useState(false);
+  const ctx = useContext(SubscriberContext);
+  const subscriberMap = ctx.state;
+  const subscribersTabList: Array<string> = ['Susbcribers', 'Subscriber State'];
```
karthiksubraveti: Thinking about this further, These should be named as 
"subscribers" -> "Config" 
"subscriber state" -> "Sessions"

If it isn't too much work, It might be nicer if we can perhaps add tabs to TopBar itself I think. Sorry for suggesting this to be similar to existing policy tabs. I was thinking this way because there is quite a bit of real estate there and we can add sessions and other subscriber bulk related views there.

nms/app/packages/magmalte/app/views/subscriber/SubscriberOverview.js
```diff
@@ -114,81 +108,56 @@ type SubscriberRowType = {
...
+  const classes = useStyles();
+  const [currTabIndex, setCurrTabIndex] = useState<number>(0);
   const networkCtx = useContext(NetworkContext);
-  // $FlowIgnore
-  const subscriberMap: {[string]: subscriber} = state.state;
-  // $FlowIgnore
-  const sessionState: {[string]: subscriber_state} = state.sessionState;
-  const subscriberMetrics = ctx.metrics;
-  const [jsonDialog, setJsonDialog] = useState(false);
+  const ctx = useContext(SubscriberContext);
```
karthiksubraveti: Is the plan to add useRefreshingContext in another PR ?

nms/app/packages/magmalte/app/views/subscriber/SubscriberStateTable.js
karthiksubraveti: We can only have view JSON option here

nms/app/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -0,0 +1,150 @@
...
+  const [currRow, setCurrRow] = useState<SubscriberRowType>({});
+  const networkId: string = nullthrows(match.params.networkId);
+  const enqueueSnackbar = useEnqueueSnackbar();
+  const ctx = useContext(SubscriberContext);
+  const networkCtx = useContext(NetworkContext);
+  const subscriberMetrics = ctx.metrics;
+  const [jsonDialog, setJsonDialog] = useState(false);
+  const [pageSize, setPageSize] = useState(10);
+  // first token (page 1) is an empty string
+  const [tokenList, setTokenList] = useState(['']);
```
karthiksubraveti, line 64: I'm apprehensive to using tokenList given that the tokens may be stale over time. We can just keep track of next page and previous page, perhaps when we refresh, we can clear the token list I guess

nms/app/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -0,0 +1,150 @@
...
+                              },
+                            );
+                          }
+                        });
+                    },
+                  },
+                ]
+          }
+          onChangeRowsPerPage={newPageSize => {
+            setPageSize(newPageSize);
```
karthiksubraveti: If it is easier not to give pageSize option, we can get rid of that. We can set it to say standard 25 subscribers per page and not have option to modify it. I'm not sure if people will modify that often

karthiksubraveti (commented):
Thanks for the changes. Left a few comments.

nms/app/packages/magmalte/app/views/subscriber/SubscriberOverview.js
HannaFar: Oh ok you're right. This should not be too much work !

nms/app/packages/magmalte/app/views/subscriber/SubscriberTable.js
HannaFar: If we're on page 4, we would keep track of previous page token (page 3) and next page token (page 5). We would not be able to go back to page 2

nms/app/packages/magmalte/app/views/subscriber/SubscriberOverview.js
HannaFar: Yes! I'll add auto refresh for subscriber state

nms/app/packages/magmalte/app/views/subscriber/SubscriberUtils.js
```diff
@@ -35,3 +35,7 @@ export function getPromValue(resp: promql_return_object) {
     .filter(Boolean);
   return respArr && respArr.length ? respArr[0] : 0;
 }
+
+export const DEFAULT_PAGE_SIZE = 25;
+
+export const MAX_PAGE_ROW_COUNT = 1000;
```
andreilee: - [ ] Add comments describing in greater detail what the constants are for

- [ ] Also group these constants with the others at the top of this file

Some ambiguity that can pop up for example for `DEFAULT_PAGE_SIZE` is what the units are, whether it's a row count, or rendering size, or something else. This is information that somebody reading the code can figure out by looking at the usages of this constant, but I very much prefer to have this sort of thing stated upfront with the definitions.

nms/app/packages/magmalte/app/state/lte/SubscriberState.js
```diff
@@ -229,11 +229,13 @@ export async function setSubscriberState(props: SubscriberStateProps) {
   }
 }
 
-export function getSubscriberGatewayMap(subscribers: {[string]: subscriber}) {
+export function getSubscriberGatewayMap(sessions: {
+  [string]: subscriber_state,
+}) {
```
andreilee, line 234: Add a return type annotation

andreilee: Also rename this to `getGatewaySubscriberMap`, it's confusing that in `LteContext.js`, `gwSubscriberMap` is getting set to the output of this.

nms/app/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -0,0 +1,204 @@
...
+  activeApns?: string,
+  ipAddresses?: string,
+  activeSessions?: number,
+  service: string,
+  currentUsage: string,
+  dailyAvg: string,
+  lastReportedTime: Date | string,
+};
+
+function Subscribers(props: WithAlert) {
```
andreilee: Rename this to `SubscribersTable` while we're doing this

nms/app/packages/magmalte/app/views/subscriber/SubscriberStateTable.js
```diff
@@ -0,0 +1,275 @@
...
+        activeSessions: activeSessions,
+        ipAddress: ipAddresses.length > 0 ? ipAddresses.join() : '-',
+        dailyAvg: metrics?.dailyAvg ?? '0',
+        lastReportedTime:
+          subscriberInfo.monitoring?.icmp?.last_reported_time === 0
+            ? new Date(subscriberInfo.monitoring?.icmp?.last_reported_time)
+            : '-',
+      };
+    },
+  );
```
andreilee, line 170: Split this off into a separate function

nms/app/packages/magmalte/app/views/subscriber/SubscriberStateTable.js
```diff
@@ -0,0 +1,275 @@
...
+
+  const tableColumns = [
+    {
+      title: 'Name',
+      field: 'name',
+    },
+    {
+      title: 'IMSI',
+      field: 'imsi',
+      render: currRow => {
```
andreilee, line 180: Split this off into a function as well

nms/app/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -0,0 +1,204 @@
...
+      field: 'name',
+    },
+    {
+      title: 'IMSI',
+      field: 'imsi',
+      render: currRow => {
+        const subscriberConfig = subscriberMap[currRow.imsi];
+        const imsi = currRow.imsi;
+        return subscriberConfig ? (
+          <Link
```
andreilee: Split this into a separate function

andreilee (changes requested):
Small nits. Looks good otherwise. Something that I think we should move away from as we get the chance is to reduce the size of code in the function React components, so some of my comments are about that.

nms/app/packages/magmalte/app/views/subscriber/SubscriberUtils.js
HannaFar: The constants at the top of the file are used in the function at the top of the file. I think it could be confusing to add DEFAULT_PAGE_SIZE and MAX_PAGE_ROW_COUNT at the top too

nms/app/packages/magmalte/app/views/subscriber/SubscriberStateTable.js
```diff
@@ -0,0 +1,263 @@
...
+        const subscriberConfig = subscriberMap[currRow.imsi];
```
karthiksubraveti, line 181: Hey @HannaFar , given that we don't have subscriberMap populated with all IMSIs, won't subscriberConfig be empty here ? We can perhaps avoid displaying it as a link when a user attempts views the subscriber state table.

nms/app/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -0,0 +1,186 @@
...
+import NetworkContext from '../../components/context/NetworkContext';
+import React from 'react';
+import SubscriberContext from '../../components/context/SubscriberContext';
+import nullthrows from '@fbcnms/util/nullthrows';
+import withAlert from '@fbcnms/ui/components/Alert/withAlert';
+
+import {DEFAULT_PAGE_SIZE} from '../../views/subscriber/SubscriberUtils';
+import {JsonDialog} from './SubscriberOverview';
+import {RenderLink} from './SubscriberOverview';
+import {handleSubscriberQuery} from '../../state/lte/SubscriberState';
```
karthiksubraveti, line 29: One thing which we might have to fix in future PR is to enable search through the subscriber table. The query can contain more than pageSize. So we might have to invoke other queries on subscriber API involving searching through IMSI or MSISDN

andreilee (approved):

PR #6046, opened 2021-04-08
AGW: MME: GTP: add S8 TEID args

pshelar:
Add missing TEIDs to program flow from eNB to PGW and vice-versa.
now the API takes in in and out TEID for S1 (eNB) tunnel as well as
S8 (PGW) tunnel.

## Additional Information

### DIFF
--- a/lte/gateway/c/oai/lib/openflow/controller/ControllerEvents.cpp
+++ b/lte/gateway/c/oai/lib/openflow/controller/ControllerEvents.cpp
@@ -144,6 +144,8 @@ AddGTPTunnelEvent::AddGTPTunnelEvent(
       pgw_ip_(INADDR_ZERO),
       in_tei_(in_tei),
       out_tei_(out_tei),
+      pgw_in_tei_(0),
+      pgw_out_tei_(0),
       imsi_(imsi),
       dl_flow_valid_(false),
       dl_flow_(),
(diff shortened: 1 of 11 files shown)


lte/gateway/c/oai/lib/openflow/controller/ControllerEvents.cpp
```diff
@@ -173,14 +177,17 @@ AddGTPTunnelEvent::AddGTPTunnelEvent(
 AddGTPTunnelEvent::AddGTPTunnelEvent(
     const struct in_addr ue_ip, struct in6_addr* ue_ipv6, int vlan,
     const struct in_addr enb_ip, const struct in_addr pgw_ip,
-    const uint32_t in_tei, const uint32_t out_tei, const char* imsi,
+    const uint32_t in_tei, const uint32_t out_tei,
```
ulaskozat: to be consistent across the board, shall we change these as enb_in_tei and enb_out_tei? to be completely agnostic, might be better to replace 'enb' with 'node1' and pgs with 'node2' across the board.

pshelar: I would like to make naming change in separate PR. since we also need better names for the APIs.
Combining all these changes will make review harder.

ulaskozat (approved):
LGTM as per the discussion.

PR #6048, opened 2021-04-08
[Sentry][AGW] Handle env var lookup error better in sentry init

themarwhal:

## Summary
- `os.environ['COMMIT_HASH']` can throw an error when the environment doesn't exist so using `os.getenv` instead.
- Addressing all lint warnings from the file along the way

## Test Plan
Ran through the case where the environment doesn't exist locally

## Additional Information


andreilee (approved):
Could you move a lot of the hardcoded strings and values into constants?

PR #6086, opened 2021-04-12
[nms] Fix infinite 302 for default generated admin users

### LINKED ISSUE #5759: [NMS] organization home page not opening
### Describe the Issue
Brought up NMS instance and created a new user and seeing organization page redirect to
`https://magma-test.localhost/[`

Screen Shot 2021-03-26 at 5 01 03 AM

**Expected behavior**

**Screenshots**

**Additional context**

andreilee:

*Note: will look into a follow-up fix on `@fbcnms/sequelize-models`*

## Summary

Fix an issue that would cause logged-in NMS users to be infinitely redirected via HTTP `302` responses to path `/[`.

In the case of the development docker setup, this would end up at `http://magma-test.localhost:8081/[`, if the user initially tried to visit `http://magma-test.localhost:8081/`. This issue is not present for all requests to the NMS. For example, visiting `http://magma-test.localhost:8081/nms` directly would avoid this redirect issue.

On investigation, it's been discovered that the default routing set up for the NMS through [Express.js](https://expressjs.com/) will redirect the user to whatever 'tab' or feature has been enabled for them. See [here](https://github.com/magma/magma/blob/master/nms/app/packages/magmalte/server/main/routes.js#L119). Typically, what this means in practice is that if a user requests a page such as `http://magma-test.localhost:8081/does_not_exist` or any other page which should `404`, our routing is set up to redirect the user to `http://magma-test.localhost:8081/admin/networks`, for example, depending on what 'tabs' the user has set up. See [here](https://github.com/facebookincubator/fbc-js-core/blob/master/fbcnms-packages/fbcnms-sequelize-models/models/user.js#L61) for the definition of 'tabs' in the `@fbcnms/sequelize-models` dependency.

What's causing this issue is that for the default admin users created through the yarn command `setAdminPassword` ([Link](https://github.com/magma/magma/blob/master/nms/app/packages/magmalte/scripts/setPassword.js) to the script it calls), the rows in Postgres are mis-typed.

**Before Fix**
```
nms=# select tabs from "Users";
 tabs
------
 "[]"
 "[]"
(2 rows)
```

The tabs column is expected to be of JSON type. When our routing redirects users to the first 'tab' they have enabled, this ends up redirecting the user to tab `/[`. A regular `tabs` value would be something like `['nms', 'admin']`.

**After Fix**
```
nms=# select tabs from "Users";
 tabs
------
 []
 []
(2 rows)
```

## Test Plan

Checked 'tabs' column after the fix for a default admin user. Also tried visiting `http://magma-test.localhost:8081/` and verified that I was redirected to `http://magma-test.localhost:8081/nms` or something else that actually worked.

## Additional Information

### DIFF
--- a/nms/app/packages/magmalte/scripts/setPassword.js
+++ b/nms/app/packages/magmalte/scripts/setPassword.js
@@ -50,6 +50,7 @@ async function createUser(userObject: UserObject) {
     password: passwordHash,
     role,
     networkIDs: [],
+    tabs: ['nms'],
     organization: org.name,
     readOnly: false,
   });


nms/app/packages/magmalte/scripts/setPassword.js
```diff
@@ -50,6 +50,7 @@ async function createUser(userObject: UserObject) {
     password: passwordHash,
     role,
     networkIDs: [],
+    tabs: [],
```
karthiksubraveti: Can we set the default tab with 'nms'. Since currently the code is primarily driven by magma usecase, no point in setting it empty. Also does this work with mariadb ? How did this work earlier ?

andreilee: So I looked into this a bit more - it's defaulting to the tabs of the organization here. This also does work with mariadb. Prior to this change, in MariaDB, the 'tabs' value would be NULL, and the user would get redirected to the tabs specified in the organization

karthiksubraveti (approved):

PR #6105, opened 2021-04-12
AGW Physical install script for Ubuntu20.

jeremy-donson:
Signed-off-by: Jeremy Donson <jdonson@fb.com>

[lte][agw] Changeset
Physical AGW on Ubuntu20

## Summary

We still have some tweaks, but this is nearing code complete.

    Please take a moment to read through the Magma project's
    - Contributing conventions: https://docs.magmacore.org/docs/next/contributing/contribute_conventions

## Test Plan

Tested this firsthand on Physical AGW running Ubuntu20 Server.
## Additional Information

This change will impact our public magma docs.

### DIFF
--- a/lte/gateway/release/agw_setup_ubuntu.sh
+++ b/lte/gateway/release/agw_setup_ubuntu.sh
@@ -1,0 +1,36 @@
+
+
+apt update
+
+# rename interfaces
+sed -i 's/enp0s3/eth0/g' /etc/netplan/50-cloud-init.yaml
+sed -i 's/GRUB_CMDLINE_LINUX=""/GRUB_CMDLINE_LINUX="net.ifnames=0 biosdevname=0"/g' /etc/default/grub
+grub-mkconfig -o /boot/grub/grub.cfg
+
+# device management
+apt install ifupdown
+
+echo "auto eth0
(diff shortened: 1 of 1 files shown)


lte/gateway/release/agw_setup_ubuntu.sh
```diff
@@ -0,0 +1,36 @@
+
+
+apt update
+
+# rename interfaces
```
tmdzk, line 5: Before renaming all interfaces let's make sure they are not already setup properly

lte/gateway/release/agw_setup_ubuntu.sh
```diff
@@ -0,0 +1,36 @@
...
+sed -i 's/enp0s3/eth0/g' /etc/netplan/50-cloud-init.yaml
+sed -i 's/GRUB_CMDLINE_LINUX=""/GRUB_CMDLINE_LINUX="net.ifnames=0 biosdevname=0"/g' /etc/default/grub
```
tmdzk, line 7: I think we don't need that anymore with netplan. But not 100% sure.

tmdzk (commented):
Looks good just some minor comments

PR #6120, opened 2021-04-13
[LoadTest][SessionD] Add a load testing script for SessionD

### LINKED ISSUE #5012: [Reliability] Benchmarking performance between stateful and stateless modes
We need a concrete way to compare the performance on releases / versions of the platform using stateful and stateless modes, we can use scale testing on Spirent setups to get results of:

- Attach rate of UEs per second
- Total UEs attached successfully
- Total UEs detached successfully
- Time of attach process per UE

### LINKED ISSUE #5696: Sessiond Performance Profiling Tests
Add and automate rpc scaling/performance tests for sessiond.

themarwhal:
## Summary
Initial script to enable loadtesting for SessionD.
Can't enable this in CI just yet, because this script needs to be run with SessionD built with BUILD_TYPE=RelWithDebInfo

## Test Plan
Ran script

## Additional Information

### DIFF
--- a/lte/gateway/python/load_tests/loadtest_sessiond.py
+++ b/lte/gateway/python/load_tests/loadtest_sessiond.py
@@ -1,0 +1,166 @@
+"""
+Copyright 2020 The Magma Authors.
+
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import argparse
+import json
+from typing import List
+
+from google.protobuf import json_format
+from load_tests.common import (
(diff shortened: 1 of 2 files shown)


lte/gateway/python/load_tests/loadtest_sessiond.py
```diff
@@ -0,0 +1,188 @@
...
+    file_name = ''
+    if req_type == 'CreateSession':
+        file_name = 'create_session_data.json'
+    elif req_type == 'EndSession':
+        file_name = 'end_session_data.json'
+    cmd_list = [
+        'ghz',
+        '--insecure', '--proto', proto_path, '--total',
+        str(num_reqs), '--call', req_name, '-D', file_name,
+        '-O', 'html', '-o', 'output.html', '0.0.0.0:50065',
```
ardzoht: can we update this to generate json output? Ideally this should be saved as `result_<name_of_loadtest>.json` (on a dir like `/var/tmp/`) so when we upload to ghz-web these are identified properly

ardzoht (changes requested):

ardzoht (approved):

PR #6127, opened 2021-04-13
[docusaurus] Add help message to delete node_modules in case of error

uri200:

## Summary

Sometimes when running docusaurus with `create_docusaurus_website.sh` the container may be stuck do to a yarn exiting prematurely. The cause may be on an out-of-date content on `node_modules`. To fix that, we just need to `rm -Rf node_modules` and run again the script to refresh all the files.

To avoid adding more automation that can slow down when we use docusaurus test VM, we will add that information on the error message so the user can take an action.

Other changes:
- Timeout increase
- Remove of PWD on docker-compose volume path.

## Test Plan

`./create_docusaurus_website.sh`

## Additional Information

### DIFF
--- a/docs/docusaurus/create_docusaurus_website.sh
+++ b/docs/docusaurus/create_docusaurus_website.sh
@@ -17,12 +17,14 @@ function exit_timeout() {
   docker-compose logs docusaurus
   echo ''
   echo "Timed out after ${1}s waiting for Docusaurus container to build. See logs above for more info."
+  echo "Possible remedies:"
+  echo '  - Remove node_modules directory (rm -rf node_modules) and try again.'
   exit 1
 }
 
 # spin until localhost:3000 returns HTTP code 200.
 function spin() {
-  maxsec=120
+  maxsec=300
   spin='-\|/'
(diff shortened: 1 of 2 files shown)


docs/docusaurus/create_docusaurus_website.sh
```diff
@@ -17,12 +17,14 @@ function exit_timeout() {
   docker-compose logs docusaurus
   echo ''
   echo "Timed out after ${1}s waiting for Docusaurus container to build. See logs above for more info."
+  echo "Possible remedies:"
+  echo "  - Remove node_modules directory (rm -Rf node_modules) and try again."
```
hcgatewood: Make this single quotes for defense-in-depth since there's an `rm -rf` inside

hcgatewood: Tiny nit: `rm -rf` instead of `rm -Rf` cause I had to look up `-R`, didn't realize it was equivalent

hcgatewood (approved):
Two nits

Thanks for adding!

PR #6138, opened 2021-04-14
[feg] s8_proxy fix PCO issues and remove some grpc deprecated fields

uri200:

## Summary
This PR includes
- Remove some deprecated fields not used from the protos anymore (will be removed from protos before making this feature official)
- Replacing some functions
- Bump go-gtp to 0.7.21

## Test Plan

Make precommit

## Additional Information

### DIFF
--- a/feg/gateway/services/s8_proxy/servicers/proto_conversions.go
+++ b/feg/gateway/services/s8_proxy/servicers/proto_conversions.go
@@ -47,12 +47,10 @@ func parseCreateSessionResponse(msg message.Message) (csRes *protos.CreateSessio
 
 	// get PDN Allocation from PGW
 	if paaIE := csResGtp.PAA; paaIE != nil {
-		paa, pdnType, err2 := handlePDNAddressAllocation(paaIE)
-		if err2 != nil {
+		csRes.Paa, csRes.PdnType, err = handlePDNAddressAllocation(paaIE)
+		if err != nil {
 			return
 		}
-		csRes.Paa = paa
-		csRes.PdnType = pdnType
 	} else {
 		csRes.GtpError = errorIeMissing(ie.PDNAddressAllocation)
 		return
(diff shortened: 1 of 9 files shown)


feg/gateway/services/s8_proxy/servicers/proto_conversions.go
```diff
@@ -47,52 +47,34 @@ func parseCreateSessionResponse(msg message.Message) (csRes *protos.CreateSessio
...
 	}
 
 	// Pgw control plane fteid
 	if pgwCFteidIE := csResGtp.PGWS5S8FTEIDC; pgwCFteidIE != nil {
-		pgwCFteid, _, err2 := handleFTEID(pgwCFteidIE)
-		if err2 != nil {
-			err = fmt.Errorf("Couldn't get PGW control plane FTEID: %s ", err2)
+		csRes.CPgwFteid, _, err = handleFTEID(pgwCFteidIE)
+		if err != nil {
+			err = fmt.Errorf("Couldn't get PGW control plane FTEID: %s ", err)
```
themarwhal, line 65: What is this line for if we exit the function?

feg/gateway/services/s8_proxy/servicers/proto_conversions.go
```diff
@@ -47,52 +47,34 @@ func parseCreateSessionResponse(msg message.Message) (csRes *protos.CreateSessio
...
 	}
 
 	// Protocol Configuration Options (PCO) optional
 	if pgwPcoIE := csResGtp.PCO; pgwPcoIE != nil {
-		pgwPcoField, err2 := pgwPcoIE.ProtocolConfigurationOptions()
-		if err2 != nil {
-			err2 = fmt.Errorf("Couldn't get PGW  Protocol Configuration Options: %s ", err2)
+		csRes.ProtocolConfigurationOptions, err = handlePCO(pgwPcoIE)
+		if err != nil {
+			err = fmt.Errorf("Couldn't get Protocol Configuration Options: %s ", err)
```
themarwhal, line 78: Same here

themarwhal (approved):

feg/gateway/services/s8_proxy/servicers/proto_conversions.go
uri200: Yes that part can be confusing but it was the best way to reduce lots of `ifs` and repeated code 

If you check the signature of the function, it includes variables csRes and err

So basically when we return, we are in fact returning `csRes` and `err` variable. 
```
// parseCreateSessionResponse parses a gtp message into a CreateSessionResponsePgw. In case
// the message is proper it also returns the session. In case it there is an error it returns
// the cause of error
func parseCreateSessionResponse(msg message.Message) (csRes *protos.CreateSessionResponsePgw, err error) {
	//glog.V(4).Infof("Received Create Session Response (gtp):\n%s", message.Prettify(msg))
	csResGtp := msg.(*message.CreateSessionResponse)
	glog.V(2).Infof("Received Create Session Response (gtp):\n%s", csResGtp.String())
	csRes = &protos.CreateSessionResponsePgw{}

```

themarwhal: oo I see! Any reason why we don't explicitly return the variables?

uri200: It made more sense when I was coding this and I was adding back and forth return parameters. Basically I was able to add/remove a parameter without having to change every single return.

But now that is more stable, maybe it makes more sense to just have 

```
return cRes, err
```
 in every single return

themarwhal: Yeah I think it would avoid confusion to explicitly return :) 
We point to this doc in our golang contributor guide too. https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

PR #6150, opened 2021-04-15
[NMS] Enable audit logging view in NMS by default

karthiksubraveti:
## Summary
NMS already has audit log capability. We don't enable by default.  Also we only dump last 20 entries. 
- Modified that to dump last 5k entries and always display audit log entries in admin page
- Modified audit log to display user email instead of userID
- Modified AuditLog component to use ActionTable instead of native table and get features like filtering for free. 

## Test Plan
Tested this manually
Screen Shot 2021-04-15 at 9 20 41 AM

Screen Shot 2021-04-15 at 9 12 44 AM

## Additional Information

### DIFF
--- a/nms/app/packages/magmalte/app/components/admin/AuditLog.js
+++ b/nms/app/packages/magmalte/app/components/admin/AuditLog.js
@@ -13,157 +13,132 @@
  * @flow
  * @format
  */
-
-import Button from '@fbcnms/ui/components/design-system/Button';
+import ActionTable from '../ActionTable';
+import CardTitleRow from '../../components/layout/CardTitleRow';
 import DeviceStatusCircle from '@fbcnms/ui/components/icons/DeviceStatusCircle';
 import Dialog from '@material-ui/core/Dialog';
-import DialogActions from '@material-ui/core/DialogActions';
 import DialogContent from '@material-ui/core/DialogContent';
 import DialogTitle from '@material-ui/core/DialogTitle';
-import IconButton from '@material-ui/core/IconButton';
-import JSONTree from 'react-json-tree';
+import Grid from '@material-ui/core/Grid';
+import ListAltIcon from '@material-ui/icons/ListAlt';
 import LoadingFiller from '@fbcnms/ui/components/LoadingFiller';
-import MoreVert from '@material-ui/icons/MoreVert';
-import NestedRouteLink from '@fbcnms/ui/components/NestedRouteLink';
-import Paper from '@material-ui/core/Paper';
 import React from 'react';
-import Table from '@material-ui/core/Table';
-import TableBody from '@material-ui/core/TableBody';
-import TableCell from '@material-ui/core/TableCell';
-import TableHead from '@material-ui/core/TableHead';
-import TableRow from '@material-ui/core/TableRow';
-import Text from '../../theme/design-system/Text';
-import {colors} from '../../theme/default';
+import ReactJson from 'react-json-view';
 
-import {Route} from 'react-router-dom';
 import {makeStyles} from '@material-ui/styles';
 import {useAxios} from '@fbcnms/ui/hooks';
-import {useRouter} from '@fbcnms/ui/hooks';
+import {useState} from 'react';
 
 const useStyles = makeStyles(theme => ({
-  actionsCell: {
-    textAlign: 'right',
-  },
-  actionsColumn: {
-    width: '160px',
-  },
-  header: {
-    margin: '10px',
-    display: 'flex',
-    justifyContent: 'space-between',
-  },
-  paper: {
-    margin: theme.spacing(3),
-  },
-  iconButton: {
-    color: colors.primary.brightGray,
-    padding: '5px',
+  dashboardRoot: {
+    margin: theme.spacing(5),
   },
 }));
 
-const THEME = {
-  scheme: 'monokai',
-  base00: '#272822',
-  base01: '#383830',
-  base02: '#49483e',
-  base03: '#75715e',
-  base04: '#a59f85',
-  base05: '#f8f8f2',
-  base06: '#f5f4f1',
-  base07: '#f9f8f5',
-  base08: '#f92672',
-  base09: '#fd971f',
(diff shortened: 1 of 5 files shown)


nms/app/packages/magmalte/server/apicontroller/auditLoggingDecorator.js
```diff
@@ -94,6 +82,46 @@ const PATHS: Array<{
     path: '/magma/v1/networks/:networkId',
     resolver: (_, params) => [params[1], 'network'],
   },
+  {
+    path: '/magma/v1/feg_lte/:networkId',
+    resolver: (_, params) => [params[1], 'FEG LTE network'],
```
andreilee: `feg_lte` should be described as `Federated LTE`

andreilee: I have similar feedback about the other strings in this file

nms/app/packages/magmalte/app/components/admin/AuditLog.js
```diff
@@ -13,157 +13,119 @@
...
           />
-        )}
-      />
-    </>
+        </Grid>
+      </Grid>
+    </div>
   );
 }
+type DialogProps = {
```
andreilee, line 118: nit: newline above this

nms/app/packages/magmalte/app/components/admin/AuditLog.js
```diff
@@ -13,157 +13,119 @@
...
+export type AuditLogRowType = {
+  id: string,
+  actingUserEmail: string,
+  mutationType: string,
+  objectType: string,
+  objectId: string,
+  mutationData: {},
 };
 
 export default function () {
```
andreilee: Could you add a name to this, eg. `AuditLog`, and another line at the bottom of the file:

```
export default AuditLog;
```

andreilee: Also a small docstring for the component would be helpful as well

nms/app/packages/magmalte/app/components/admin/AuditLog.js
```diff
@@ -13,157 +13,119 @@
...
+                    />
+                    {currRow.status}
+                  </>
+                ),
+              },
+            ]}
+            options={{
+              actionsColumnIndex: -1,
+              pageSize: 20,
+              pageSizeOptions: [20],
```
andreilee: I think we can re-use and move the constants here:
```
// default subscriber count in get subscriber query
export const DEFAULT_PAGE_SIZE = 25;

// max subscribers in array
export const MAX_PAGE_ROW_COUNT = 1000;
```

https://github.com/magma/magma/blob/master/nms/app/packages/magmalte/app/views/subscriber/SubscriberUtils.js#L39-L43

nms/app/packages/magmalte/app/components/admin/AuditLog.js
```diff
@@ -13,157 +13,119 @@
...
+              {title: 'Object Type', field: 'objectType'},
+              {title: 'Object ID', field: 'objectId'},
+              {
+                title: 'Status',
+                field: 'status',
+                render: currRow => (
+                  <>
+                    <DeviceStatusCircle
+                      isGrey={false}
+                      isActive={currRow.status === 'SUCCESS'}
```
andreilee, line 90: Should probably group the `'SUCCESS'` with the same string in `auditLoggingDecorator.js`, into a single constant, and import it here

nms/app/packages/magmalte/app/components/admin/AuditLog.js
```diff
@@ -13,157 +13,119 @@
...
-  base0A: '#f4bf75',
-  base0B: '#a6e22e',
-  base0C: '#a1efe4',
-  base0D: '#66d9ef',
-  base0E: '#ae81ff',
-  base0F: '#cc6633',
+export type AuditLogRowType = {
+  id: string,
+  actingUserEmail: string,
+  mutationType: string,
```
andreilee, line 41: Since there's a small limited set of values that are valid here, I would use js flow type annotations here to specify

andreilee (commented):
I haven't looked through this completely yet

nms/app/packages/magmalte/app/components/admin/AuditLog.js
karthiksubraveti: I don't understand this comment. Can you clarify ?

nms/app/packages/magmalte/app/components/admin/AuditLog.js
karthiksubraveti: I can move this as a constant outside this code, however this can stay local. Every component can define its own PAGE SIZE. It is more dependent on the view I would think.

nms/app/packages/magmalte/app/components/admin/AuditLog.js
karthiksubraveti: I think that is a good practice, However in this case, I think 'SUCCESS' or 'FAILURE' is well understood. If this is important, then we should create a constant.js which can be imported across the board and we just use those constants across the board. To make this change for this case alone, I don't believe is useful. It might be more efficient to just grep for commonly used strings and create a single util js and move it there.

nms/app/packages/magmalte/app/components/admin/AuditLog.js
karthiksubraveti: Same comment as I made below. For now, I prefer to leave it here and will add a cleanup PR where we can move all constants to a common place and import it. I think doing it piecemeal isn't useful. It is just going to move things around.

nms/app/packages/magmalte/app/components/admin/AuditLog.js
andreilee: Oh since there's only three values possible of
```
CREATE
UPDATE
DELETE
```
for `mutationType`, this can be an enum of sorts.

andreilee (approved):

Issue #5477, opened 2021-03-13
[NMS] Read only user not support in NMS


karthiksubraveti:
when the Role of NMS user is changed to Read only User with read only permission to specific network , still the user is able to make changes (add/delete) info.

-- 

This is a known issue and we aren't currently supporting user roles yet.

PR #6156, opened 2021-04-15
[NMS] Add checks to ensure mutation is only allowed for non readonly users

karthiksubraveti:

## Summary
Add checks to ensure read only user cannot make mutations. 

## Test Plan
Enhanced existing test to include this filter check
Screen Shot 2021-04-15 at 10 54 04 AM

## Additional Information

### DIFF
--- a/nms/app/packages/magmalte/server/apicontroller/__tests__/proxy-test.js
+++ b/nms/app/packages/magmalte/server/apicontroller/__tests__/proxy-test.js
@@ -14,10 +14,10 @@
  * @format
  */
 
-import {networkIdFilter, networksResponseDecorator} from '../routes';
+import {apiFilter, networksResponseDecorator} from '../routes';
 
 // $FlowIgnore Ignoring error for tests.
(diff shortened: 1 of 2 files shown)


andreilee (approved):

andreilee:
Approving this PR, but I have concerns that this can lead to some confusion for read only users. Are all our edit/delete flows able to handle this 403 in a graceful way that communicates to the user what they can't do?

karthiksubraveti:
> Approving this PR, but I have concerns that this can lead to some confusion for read only users. Are all our edit/delete flows able to handle this 403 in a graceful way that communicates to the user what they can't do?

It will handle like handling any other error in the response. This above screenshot misses the notification which was displayed in the form. However that being said, your concern is right on providing more clarity on what a user can do and not do. I have added this PR to support from the correctness perspective. However I'm hoping when we look into adding RBAC support we can make larger changes in this area.

karthiksubraveti:
Closes #5477

PR #6176, opened 2021-04-16
DO NOT COMMIT - Demonstration of cpplint on large cpp commit

electronjoe:


electronjoe:
Oof, this PR stack with a git merge is not one I want to deal with for linting proof of concept

PR #6177, opened 2021-04-16
DO NOT COMMIT Pr gh cpplint demo

electronjoe:


orc8r/gateway/c/common/service303/MagmaService.h
```diff
@@ -12,6 +12,8 @@
  */
 #pragma once
 
+// DO NOT COMMIT TEST
+
```
themarwhal, line 16: so we use c++11? I keep getting confused as to whether we use 11/14

electronjoe: cpplint is unaware of our compiler flags, it's pure compilation unit pattern matching without makefile analysis.  So how I read this warning was that std::chrono is a c++11 release library.  But looking into it it appears to be c++11/20... so I dunno - maybe this error message is just stale since the new c++20 chrono mods. I would ignore the c++11 part and just read it as "unallowed library in google". As they use absl::time (we should probably too - e.g. compatibility with gRPC).

PR #6190, opened 2021-04-16
Fix minor bug in precommit.py for python lint/format

themarwhal:

## Summary
The current script terminates as soon as `add-trailing-comma` modifies one file. This is because for some reason the script returns 1 when that happens. This can be changed by passing in the `--exit-zero-even-if-changed`.

## Test Plan
Run format on multiple files

## Additional Information


ardzoht (approved):
lgtm

Issue #6130, opened 2021-04-14
CreateSessionUtil imports issue for pipelined_cli


panyogesh:
Issue :
pipelined_cli.py : from magma.pipelined.tests.app.ng_set_session_msg import CreateSessionUtil
                                 ImportError: No module named 'magma.pipelined.tests'

Cause : For fab builds the tests directory is not present hence the CreateSessionUtil() is not present hence the error.

ekowtaylor:
Stub, new more info to debug further

PR #6191, opened 2021-04-16
CreateSessionUtil imports issue for pipelined_cli #6130

prabinakpattnaik:

## Summary

ng_set_session_msg.py script is changed the place from tests/app to python/scripts.
Bare metal build is not included the tests directory.

## Test Plan
Executed the use-case using python/scripts/pipelined_cli.py.


lte/gateway/python/scripts/pipelined_cli.py
```diff
@@ -48,11 +48,10 @@
 from magma.pipelined.policy_converters import convert_ipv4_str_to_ip_proto
 from magma.pipelined.qos.common import QosManager
 from magma.pipelined.service_manager import Tables
-from magma.pipelined.tests.app.ng_set_session_msg import CreateSessionUtil
 from magma.subscriberdb.sid import SIDUtils
+from ng_set_session_msg import CreateSessionUtil
```
ardzoht: Not sure if `lte/gateway/python/scripts` would be the best place to have these on here, this looks pretty specific to pipelined since it was on pipelined/tests? If there's no clear path to move this one here, something we could try is maybe define a `scripts/helpers` to put util related files on here? cc @koolzz

ardzoht (changes requested):

lte/gateway/python/scripts/pipelined_cli.py
koolzz: Yeah I agree, I don't think we currently have a good dir for this. scripts/helpers sounds fine to me

prabinakpattnaik: The "ng_set_session_msg" python file is used in pipeleind_cli.py and tests/test_ng_servicer_session.py. But for bare metal build "pipeliend/tests" directory is not included. Previously ng_set_session_msg file was kept on pipelined/tests. So CreateSessionUtil imports issue getting when executed pipelined_cli.py.

Now we moved to "pipeliend/tests" to "python/scripts" for "ng_set_session_msg" python file.

Yes I agree, it should be defined in meaningful path. I will do create a new directory "scripts/helpers" for same.

koolzz (approved):

ardzoht (approved):

PR #6227, opened 2021-04-19
AGW: sysctl: workaround to apply sysctl config.

pshelar:
service procps is suppose to read config from '/etc/sysctl.d/'
and apply it at startup. But it is not applying it due to known bug.
Restarting this service seems to apply all configuration.
Following patch add this step to sctpd service start.

## Additional Information

### DIFF
--- a/lte/gateway/python/scripts/config_stateless_agw.py
+++ b/lte/gateway/python/scripts/config_stateless_agw.py
@@ -165,6 +165,8 @@ def ovs_reset_bridges():
 
 
 def sctpd_pre_start():
+    subprocess.Popen("service procps restart".split())
+
     if check_stateless_services() == return_codes.STATEFUL:
         # switching from stateless to stateful
         print("AGW is stateful, nothing to be done")


lte/gateway/python/scripts/config_stateless_agw.py
```diff
@@ -165,6 +165,8 @@ def ovs_reset_bridges():
 
 
 def sctpd_pre_start():
+    subprocess.Popen("service procps restart".split())
```
ardzoht, line 168: should this be a blocking call? currently it will call the service restart and continue with the other process, is this the intended behavior?

ardzoht (approved):
lgtm, please take a look at question about Popen call

lte/gateway/python/scripts/config_stateless_agw.py
pshelar: yes, thats fine. it does not need to block sctpd service start.

themarwhal:
Can you provide a link to the "known bug"? I'm curious.

pshelar:
> Can you provide a link to the "known bug"? I'm curious.

I missed this comment, I will send another PR.

PR #6228, opened 2021-04-19
[lte][agw] Add missing parameter in log message

ulaskozat:

## Summary

Added missing parameter in log message format.

## Test Plan

Build with no error.

## Additional Information


ardzoht (approved):
lgtm

lionelgo (approved):
I would also push the idea to use MME_UE_S1AP_ID_FMT defined in common/common_types.h instead of %u,  and change it to PRIu32

ulaskozat:
> I would also push the idea to use MME_UE_S1AP_ID_FMT defined in common/common_types.h instead of %u, and change it to PRIu32

I missed this comment. I can send a clean up PR in a separate one...

PR #6230, opened 2021-04-19
[lte][agw] Keep track of UE handover state

shaddi:
## Summary

During handover procedures, we need to keep track of certain state related to source and target eNB across messages. This adds this state to the `ue_description`.

Signed-off-by: Shaddi Hasan <shasan@fb.com>

## Test Plan

- Local testing (2x Baicells eNB + Pixel 3)
- TeraVM + Spirent


lte/gateway/c/oai/include/s1ap_types.h
```diff
@@ -79,6 +80,14 @@ enum s1_ue_state_s {
                         /// UE context Release Complete
 };
 
+typedef struct handover_state_s {
+  mme_ue_s1ap_id_t mme_ue_s1ap_id;
+  uint32_t source_enb_id;
+  uint32_t target_enb_id;
+  enb_ue_s1ap_id_t
+      target_enb_ue_s1ap_id : 24;  ///< Unique UE id over eNB (24 bits wide)
+} handover_state_t;
```
ardzoht: To keep the state naming convention, can we update this struct name to `s1ap_handover_state`? And also for its protobuf message types as well

lte/gateway/c/oai/include/s1ap_types.h
```diff
@@ -29,6 +29,7 @@ struct enb_description_s;
 
 #define S1AP_TIMER_INACTIVE_ID (-1)
 #define S1AP_UE_CONTEXT_REL_COMP_TIMER 1  // in seconds
+#define S1AP_UE_HANDOVER_NULL_ENB 0
```
ardzoht: what is the usage for this? I don't see this being used at the moment, also if this can be updated to reflect the usage (e.g `S1AP_UE_HANDOVER_NULL_ENB_ID`?)

ardzoht (approved):
left 2 naming convention nits, overall lgtm

lte/gateway/c/oai/include/s1ap_types.h
shaddi: Good catch, no longer needed.

PR #6231, opened 2021-04-19
[Enhanced_Testing] Added check to verify if requested eps attach type is matching with accepted result

VinashakAnkitAman:
## Title
[Enhanced_Testing] Added check to verify if requested eps attach type is matching with accepted result

## Summary
This PR adds verification for accepted attach result to check if it matches the request EPS attach type.

## Test Plan
Verified with sanity


ulaskozat (approved):
LGTM

Issue #5672, opened 2021-03-23
[mme + subscriberdb] Add 5G NSA IEs in Update Location Request/Answer


ssanadhya:
Add the following **information elements** to the GRPC proto messages between MME and SubscriberDB:
1. [Update Location Request](https://github.com/magma/magma/blob/master/feg/protos/s6a_proxy.proto#L132) from MME to Subscriberdb: Dual-Registration-5G-Indicator 
2. [Update Location Answer](https://github.com/magma/magma/blob/master/feg/protos/s6a_proxy.proto#L146) from Subscriberdb to MME over GRPC: 
- NR as Secondary RAT
- Extend the [AggregateMaximumBitrate](https://github.com/magma/magma/blob/master/feg/protos/s6a_proxy.proto#L203) message to include an enum for specifying the units of maximum bitrate in `bps`(4G) or `kbps`(5G-NSA)

Add **logic in MME** to include the IE as part of ULR and process the IEs in ULA

Add **logic in SubscriberDB** to parse the `Dual-Registration-5G-Indicator` IE and respond with the correct values for `NR as Secondary RAT` and `AggregateMaximumBitrate` units

Ref: [LTE 5G Dual connectivity attach call flow](http://www.eventhelix.com/5G/non-standalone-access-en-dc/en-dc-secondary-node-addition-core-network-details.pdf)

uri200:
Hey @ssanadhya I may be missing some context here, but is this feature trying to implement dual connectivity? 

If so, per the  [LTE 5G Dual connectivity attach call flow](http://www.eventhelix.com/5G/non-standalone-access-en-dc/en-dc-secondary-node-addition-core-network-details.pdf) doc I see that we need to send `Feature-List-ID-2 {NR as Secondary RAT}` on the ULA. 

![image](https://user-images.githubusercontent.com/16157139/113225944-cfb2d580-9243-11eb-9854-a6b6879feea3.png)

Table 7.3.10/2 at [3GPP TS 29.272](https://www.etsi.org/deliver/etsi_ts/129200_129299/129272/16.03.00_60/ts_129272v160300p.pdf)



while dual registration 5g indications says 

![image](https://user-images.githubusercontent.com/16157139/113226214-75fedb00-9244-11eb-9f88-045a272684f1.png)

Vyas20:
Hey @uri200 
Yes, this feature is for Dual connectivity implementation.
[...]

PR #6234, opened 2021-04-19
[feg] s6a_proxy add Feature List ID 2 and DualRegistration_5GIndicator

### LINKED ISSUE #5673: [feg] Addition of 5G NSA IEs in Update Location Request/Answer between MME and Feg
Extend the handling of 5G NSA IEs in ULA/ULR described in https://github.com/magma/magma/issues/5672 to S6a proxy in Feg

uri200:

## Summary

Added `DualRegistration_5GIndicator` and `Feature List ID 2` with `NrAsSecondaryRat` to support 5G NSA https://github.com/magma/magma/issues/5672 

## Test Plan

make precommit at feg

## Additional Information


feg/gateway/services/s6a_proxy/servicers/s6a_definitions.go
```diff
@@ -379,4 +380,5 @@ type ULR struct {
 	VisitedPLMNID    datatype.Unsigned32       `avp:"Visited-PLMN-Id"`
 	RATType          datatype.Unsigned32       `avp:"RAT-Type"`
 	ULRFlags         datatype.Unsigned32       `avp:"ULR-Flags"`
+	SupportedFeatures           []SupportedFeatures         `avp:"Supported-Features"`
```
themarwhal: maybe run go fmt? the spacing looks weird

feg/gateway/services/s6a_proxy/servicers/ul.go
```diff
@@ -55,6 +64,27 @@ func (s *s6aProxy) sendULR(sid string, req *protos.UpdateLocationRequest, retryC
...
+		ulrFlags = ulrFlags | 1<<8
+	}
+	return ulrFlags
+}
+
+// createFeatureListID2 creates the Feature-List-ID 2 based on TS 29.272
+func createFeatureListID2(id2 *protos.FeatureListId2) int {
+	res := 0
+	if id2.NrAsSecondaryRat {
+		// set bit 27 (TS 29.272 Nr As Secondary Rat)
```
themarwhal, line 89: You could use a bitvector type to clean these things up if you are up for it.

themarwhal (approved):
few comments, but looks good in general

feg/gateway/services/s6a_proxy/servicers/s6a_definitions.go
uri200: yep, I think I forgot to amend after running make pre commit 😅

feg/gateway/services/s6a_proxy/servicers/ul.go
uri200: I was not able to find a bitvector on the golang standard modules. Is there any I am missing?

I also didn't want to hardcode any structure because the length of bits may change in the future. And also because s6a_proxy has other booleans to signal bits. 

I think that even this requires more code, it is easier to add more flags and see the specific bit positions they are added.

themarwhal: Gotcha! sounds good then

feg/gateway/services/s6a_proxy/servicers/s6a_proxy.go
```diff
@@ -35,7 +35,7 @@ import (
 
 const (
 	ULR_RAT_TYPE     = 1004
-	ULR_FLAGS        = 1<<1 | 1<<5
+	ULR_FLAGS        = 1<<1 | 1<<5 // 29.272 Table 7.3.7/1: ULR-Flags (S6a/S6d-Indicator, Initial-AttachIndicator)
```
emakeev: 0b100010 could be easier to read - https://play.golang.org/p/oSavbH6XtbO

uri200: I agree. I thought shifting bits was the `magma standard`
I will modify it

feg/gateway/services/s6a_proxy/servicers/ul.go
```diff
@@ -158,3 +189,17 @@ func (s *s6aProxy) UpdateLocationImpl(req *protos.UpdateLocationRequest) (*proto
 	}
 	return res, err
 }
+
+// getFeatureListID2 creates the Feature-List-ID 2 based on TS 29.272
+func getFeatureListID2(supportedFeatures []SupportedFeatures) *protos.FeatureListId2 {
+	protoFeatureList := &protos.FeatureListId2{}
```
emakeev, line 205: if FeatureListId2 is an optional field, it may make sense to not to set it at all - return nil.

feg/gateway/services/s6a_proxy/servicers/ul.go
```diff
@@ -158,3 +189,17 @@ func (s *s6aProxy) UpdateLocationImpl(req *protos.UpdateLocationRequest) (*proto
...
+	for _, features := range supportedFeatures {
+		if features.FeatureListID == 2 {
+			// get bit 27 (TS 29.272 Nr As Secondary Rat)
+			if features.FeatureList&(1<<27) != 0 {
+				protoFeatureList.NrAsSecondaryRat = true
```
emakeev, line 210: return &protos.FeatureListId2{NrAsSecondaryRat: true}

feg/gateway/services/s6a_proxy/servicers/ul.go
```diff
@@ -158,3 +189,17 @@ func (s *s6aProxy) UpdateLocationImpl(req *protos.UpdateLocationRequest) (*proto
...
+			}
+		}
+	}
+	return protoFeatureList
```
emakeev, line 214: return nil

feg/gateway/services/s6a_proxy/servicers/ul.go
```diff
@@ -55,6 +64,27 @@ func (s *s6aProxy) sendULR(sid string, req *protos.UpdateLocationRequest, retryC
 	return err
 }
 
+// createULR_Flags creates ULR Flags based on TS 29.272
+func createULR_Flags(req *protos.UpdateLocationRequest) int {
+	// ULR_FLAGS contains defaults flags for ULR-Flags
+	ulrFlags := ULR_FLAGS
+	if req.DualRegistration_5GIndicator {
```
emakeev: you can use req.GetDualRegistration_5GIndicator() instead, then you don't have to check for req == nil

feg/gateway/services/s6a_proxy/servicers/ul.go
```diff
@@ -55,6 +64,27 @@ func (s *s6aProxy) sendULR(sid string, req *protos.UpdateLocationRequest, retryC
...
+// createFeatureListID2 creates the Feature-List-ID 2 based on TS 29.272
+func createFeatureListID2(id2 *protos.FeatureListId2) int {
+	res := 0
+	if id2 != nil && id2.NrAsSecondaryRat {
+		// set bit 27 (TS 29.272 Nr As Secondary Rat)
+		res = res | 1<<27
+	}
+	return res
+}
+
```
emakeev, line 94: maybe: 
  
```
  func createFeatureListID2(id2 *protos.FeatureListId2) (ret int) {
	if id2 == nil { return }
	switch {
         case id2.NrAsSecondaryRat:
		// set bit 27 (TS 29.272 Nr As Secondary Rat)
		res = res | Bit27 // or whatever the appropriate name would be
        // // some future case -
        // case id2.SomeFutureFlag:
        //       res = res | SomeFutureBit
         }
         return
  }

```

feg/gateway/services/s6a_proxy/servicers/ul.go
```diff
@@ -44,9 +44,18 @@ func (s *s6aProxy) sendULR(sid string, req *protos.UpdateLocationRequest, retryC
 	m.NewAVP(avp.UserName, avp.Mbit, 0, datatype.UTF8String(req.UserName))
 	m.NewAVP(avp.AuthSessionState, avp.Mbit, 0, datatype.Enumerated(1))
 	m.NewAVP(avp.RATType, avp.Mbit, uint32(diameter.Vendor3GPP), datatype.Enumerated(ULR_RAT_TYPE))
-	m.NewAVP(avp.ULRFlags, avp.Vbit|avp.Mbit, uint32(diameter.Vendor3GPP), datatype.Unsigned32(ULR_FLAGS))
+	m.NewAVP(avp.ULRFlags, avp.Vbit|avp.Mbit, uint32(diameter.Vendor3GPP), datatype.Unsigned32(createULR_Flags(req)))
 	m.NewAVP(avp.VisitedPLMNID, avp.Vbit|avp.Mbit, diameter.Vendor3GPP, datatype.OctetString(req.VisitedPlmn))
 
+	// Supported feature-List-id-2
```
emakeev, line 50: would it make sense to not to add SupportedFeatures AVP at all if req.FeatureListId_2 is not set, for backward compatibility?

`if req.GetFeatureListId_2() != nil { ...`

emakeev (approved):
it could improve readability if you define consts for all the bitflags used.
then, it'll read like: res = res | Bit27 // or whatever the appropriate name would be

You can do something like this:

```
type URLFlag int

const (
    FlagBit1 URLFlag = 1 << iota
    FlagBit2
    FlagBit3
    _
    FlagBit5
    FlagBit27 = 1 << 27
)

```

feg/gateway/services/s6a_proxy/servicers/ul.go
uri200: My idea was this code to be easily extensible once we have more flags. There are 28 other potential flags we will be using, specially with 5g.

If I change the code to return in case of finding bit 27, the day we add a new flag, we will have to modify the algorithm back to its current implementation.

Also, I thought it was better to return an empty slice instead of a nil to avoid nil pointer panics?

feg/gateway/services/s6a_proxy/servicers/ul.go
uri200: Ok with that example, now I get your other comment with the `type URLFlag int`

And it looks nice that way. Will adopt it

uri200:
@emakeev per your suggestion I created FlagBits. To use with new bits we just need to add a new flag

Changes are on the second commit on this PR

feg/gateway/services/s6a_proxy/servicers/ul.go
uri200: now I am not adding the AVP if `FeatureListId_2` is not set

PR #6278, opened 2021-04-21
[orc8r][configurator] Tidy up loadType logic

hcgatewood:
## Summary

Clarify addSuffix functionality

## Test Plan

- [x] Pass CI

## Additional Information


karthiksubraveti (approved):
I had left it that way originally because I felt the cases were redundant. That being said, code is more readable here. Thanks for the change.

PR #6281, opened 2021-04-21
[lte][agw] Add HO Cancel and Failure procedures

shaddi:
# Summary

Adds the HandoverCancel, HandoverCancelAcknowledge, HandoverFailure, and HandoverPreprarationFailure procedures. These cover all the (expected-in-the-spec) negative cases for S1 handover.

Signed-off-by: Shaddi Hasan <shasan@fb.com>

## Test Plan

- 2x Baicells eNB + Pixel 3
- TeraVM

### DIFF
--- a/lte/gateway/c/oai/tasks/s1ap/s1ap_mme_handlers.c
+++ b/lte/gateway/c/oai/tasks/s1ap/s1ap_mme_handlers.c
@@ -115,10 +115,10 @@ bool is_all_erabId_same(S1ap_PathSwitchRequest_t* container);
 s1ap_message_handler_t message_handlers[][3] = {
     {s1ap_mme_handle_handover_required, 0, 0}, /* HandoverPreparation */
     {0, s1ap_mme_handle_handover_request_ack,
-     0},                                     /* HandoverResourceAllocation */
+     s1ap_mme_handle_handover_failure},      /* HandoverResourceAllocation */
     {s1ap_mme_handle_handover_notify, 0, 0}, /* HandoverNotification */
     {s1ap_mme_handle_path_switch_request, 0, 0}, /* PathSwitchRequest */
-    {0, 0, 0},                                   /* HandoverCancel */
+    {s1ap_mme_handle_handover_cancel, 0, 0},     /* HandoverCancel */
     {0, s1ap_mme_handle_erab_setup_response,
      s1ap_mme_handle_erab_setup_failure},      /* E_RABSetup */
     {0, 0, 0},                                 /* E_RABModify */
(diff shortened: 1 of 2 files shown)


lte/gateway/c/oai/tasks/s1ap/s1ap_mme_handlers.c
```diff
@@ -2059,6 +2059,326 @@ int s1ap_mme_handle_handover_request_ack(
...
+    case S1ap_Cause_PR_misc:
+      cause_value = ie->value.choice.Cause.choice.misc;
+      break;
+
+    default:
+      OAILOG_ERROR(
+          LOG_S1AP, "HANDOVER FAILURE with Invalid Cause_Type = %d\n",
+          cause_type);
+      OAILOG_FUNC_RETURN(LOG_S1AP, RETURNerror);
+  }
```
ulaskozat: let's consider factoring out this part eventually as we have used this pattern other places.

lte/gateway/c/oai/tasks/s1ap/s1ap_mme_handlers.c
```diff
@@ -2059,6 +2059,326 @@ int s1ap_mme_handle_handover_request_ack(
...
+
+  if (ue_ref_p->s1_ue_state == S1AP_UE_HANDOVER) {
+    // this effectively cancels the HandoverPreparation proecedure as we
+    // only send a HandoverCommand if the UE is in the S1AP_UE_HANDOVER
+    // state.
+    ue_ref_p->s1_ue_state         = S1AP_UE_CONNECTED;
+    ue_ref_p->s1ap_handover_state = (struct s1ap_handover_state_s){0};
+  } else {
+    // Not a failure, but nothing for us to do.
```
ulaskozat, line 2172: do we need to continue with the processing of the message? What is the side effect or how do we end up in this state?

lte/gateway/c/oai/tasks/s1ap/s1ap_mme_handlers.c
```diff
@@ -2059,6 +2059,326 @@ int s1ap_mme_handle_handover_request_ack(
...
+  // Grab the Cause Type and Cause Value
+  S1AP_FIND_PROTOCOLIE_BY_ID(
+      S1ap_HandoverCancelIEs_t, ie, container, S1ap_ProtocolIE_ID_id_Cause,
+      true);
+  if (ie) {
+    cause_type = ie->value.choice.Cause.present;
+  } else {
+    OAILOG_FUNC_RETURN(LOG_S1AP, RETURNerror);
+  }
+  switch (cause_type) {
```
ulaskozat: considering this is already reused in this PR twice, actually it is better to factor out the setting of the cause_value.

ulaskozat (approved):
Approving with few nits.

lte/gateway/c/oai/tasks/s1ap/s1ap_mme_handlers.c
shaddi: This is about cleaning up the state on the source eNB. We can get into this situation if:

1. Source eNB sends a HandoverRequired to the MME
2. The MME sends the HandoverRequest to the Target eNB
3. The Source eNB sends a HandoverCancel (for some reason) to the MME, which is processed.
4. The Target eNB (for whatever reason) separately rejects the handover, generating a failure.

It'd be reasonable to assume that most eNBs, after sending a cancel, will clean up their own state and ignore the HandoverPreparationFailure message we send, but my read of the spec suggests that we should still send the unsuccessful outcome message to properly complete the procedure just in case.

I am sure there are other situations where the UE may change state (e.g., power off) during the handover procedure as well. In all cases, we should still notify the source eNB that the target eNB did not accept the handover.

ulaskozat: sounds good.

PR #6292, opened 2021-04-21
[Orc8r] Improve orcl cleanup and verify functionality

karthiksubraveti:

## Summary
- Decouple Orc8r verification from terraform state inspection. We currently depend on terraform state only for getting the namespace. Modified to obtain it from user option (with default as orc8r)
- Enable orc8r cleanup without terraform state inspection. Provide default values when not provided
- Removed aws route53 and elb ansible plugins and used aws cli through ansible shell. Using aws cli is more reliable and consistent.
- Removed security groups cyclic dependency by purging rules prior to deletion

## Test Plan
Manually ran orcl and verified the behavior

## Additional Information


hcgatewood (approved):
Fix lints

PR #5105, opened 2021-02-24
[metricsd] Increase test coverage of obsidian handlers

Scott8440:
Signed-off-by: Scott8440 <scott8440@gmail.com>

## Summary

Add tests for all prometheus obsidian handlers in metricsd. Some small fixes/improvements to error messages were found during testing.

## Test Plan

it's unit tests

PR #6304, opened 2021-04-22
[orc8r] Fix get-all for prometheus alerts in REST API

andreilee:

## Summary

Behavior was changed in #5105

This is a fix for the alerts REST API in orchestrator. Get-all behavior was broken accidentally in #5105 and was preventing NMS alerts page from working correctly. This fix changes the API behavior and allows for an empty `alert_name` to specify a get-all again.

## Test Plan

Verified through swagger API:

Screen Shot 2021-04-22 at 10 40 18 AM

Also verified that `Sync Predefined Rules` for alerts on the NMS works. Results after syncing on a new network:

Screen Shot 2021-04-22 at 10 38 44 AM

## Additional Information


karthiksubraveti (commented):
Do we know when the alert name is empty ? Are we passing having invalid rules in NMS ?

karthiksubraveti (approved):
ignore comment. changes lgtm

PR #6307, opened 2021-04-22
[DO NOT MERGE][AGW][MME][RHEL8] Fix event size

rdefosse:

## Summary

This is a problem we at OAI have been aware for quite a long time. Just making it upstream.

Let make an example:

```c
#include <iostream>
#include <netinet/sctp.h>

int main() {
  struct sctp_event_subscribe event;
  std::cout << "sizeof(event) = "  << sizeof(event) << std::endl;
  return 0;
}
```

Inside my Ubuntu18 container:

```bash
# ./event-size.exe 
sizeof(event) = 10
# dpkg --list | grep sctp
ii  libsctp-dev:amd64                      1.0.17+dfsg-2                       amd64        user-space access to Linux kernel SCTP - development files
ii  libsctp1:amd64                         1.0.17+dfsg-2                       amd64        user-space access to Linux kernel SCTP - shared library
```

Inside my RHEL8 container:

```bash
# ./event-size.exe 
sizeof(event) = 14
# yum list installed | grep sctp
lksctp-tools.x86_64                           1.0.18-3.el8                             @rhel-8-for-x86_64-baseos-rpms           
lksctp-tools-devel.x86_64                     1.0.18-3.el8                             @rhel-8-for-x86_64-baseos-rpms           
```

Here comes the reason for the value `10`.

## Test Plan

I was able to test w/ and w/o the patch locally. 

- Without patch --> no HSS connection, no eNB connection
- With patch, everything is OK.

This patch is required for me to move forward on CI tasks (as I mentioned yesterday in the `devops` meeting.

## Additional Information

### DIFF
--- a/lte/gateway/c/sctpd/util.cpp
+++ b/lte/gateway/c/sctpd/util.cpp
@@ -102,8 +102,12 @@ int set_sctp_opts(
   event.sctp_association_event = on;
   event.sctp_shutdown_event    = on;
   event.sctp_data_io_event     = on;
+  int event_size               = sizeof(event);
+  if (event_size > 10) {
+    event_size = 10;
+  }
 
-  if (setsockopt(sd, IPPROTO_SCTP, SCTP_EVENTS, &event, sizeof(event)) < 0) {
+  if (setsockopt(sd, IPPROTO_SCTP, SCTP_EVENTS, &event, event_size) < 0) {
     MLOG_perror("setsockopt");
     return -1;
   }
--- a/ci-scripts/docker/Dockerfile.mme.ci.rhel8
+++ b/ci-scripts/docker/Dockerfile.mme.ci.rhel8
@@ -36,7 +36,7 @@ RUN yum update -y && \
       libubsan \
       libasan \
       psmisc \
-      psmisc \
+      tcpdump \
       openssl \
       net-tools \
       tzdata && \
--- a/lte/gateway/docker/mme/Dockerfile.rhel8
+++ b/lte/gateway/docker/mme/Dockerfile.rhel8
@@ -456,7 +456,7 @@ RUN yum update -y && \
       libubsan \
       libasan \
       psmisc \
-      psmisc \
+      tcpdump \
       openssl \
       net-tools \
       tzdata && \


lionelgo (approved):

lte/gateway/c/sctpd/util.cpp
```diff
@@ -102,8 +102,12 @@ int set_sctp_opts(
   event.sctp_association_event = on;
   event.sctp_shutdown_event    = on;
   event.sctp_data_io_event     = on;
+  int event_size               = sizeof(event);
+  if (event_size > 10) {
```
ulaskozat, line 106: This looks too hacky... Do we know why sizeof generates a wrong value?

electronjoe: Agree that this is unexpected and a root cause would be good to know

electronjoe: What kernel versions are running (`uname -a` if possible!) in each?

electronjoe: I'm a bit confused actually - as linux kernel does include a definition of `sctp_event_subscribe` - is that not the source of this struct? Meanwhile I unpacked `lksctp-tools-devel.x86_64` and found that `<netinet/sctp.h>` contains no struct definition for `sctp_event_subscribe`.

```
➜  lksctp-tools-devel-1.0.18-3.el8.x86_64 grep -R sctp_event_subscribe .
grep: ./usr/lib64/lksctp-tools/libwithsctp.so: No such file or directory
grep: ./usr/lib64/libsctp.so: No such file or directory
./usr/share/lksctp-tools/sctp_test.c:	struct sctp_event_subscribe subscribe;
./usr/share/lksctp-tools/sctp_darn.c:	struct sctp_event_subscribe subscribe;
```

Even though the package does supply netinet/sctp.h:

```
grep sctp_event_subscribe usr/include/netinet/sctp.h
```

electronjoe: FWIW I dug into RHEL8 (kernel 4.18 and they claim upstream, so I assume [this is the defn of sctp_event_subscribe](https://sourcegraph.com/github.com/torvalds/linux@94710ca/-/blob/include/uapi/linux/sctp.h#L582:8), and also ubuntu 18.04 which appears to be kernel 5.3 though upstream less clear, so I looked at [this defn of sctp_event_subscribe](https://sourcegraph.com/github.com/torvalds/linux@4d856f7/-/blob/include/uapi/linux/sctp.h#L591:8)).  And in both cases, the struct has size 13 bytes...

electronjoe: @rdefosse for visibility

rdefosse:
@electronjoe @ulaskozat 

We had a discussion about this w/ @lionelgo this afternoon.

It's a bit more complex to explain than I thought.

Regards
Raphael

rdefosse:
@lionelgo @electronjoe 

I will close this PR.

We need to come back to something else later.

Raphael

PR #6332, opened 2021-04-23
[NMS] Cleanup subscriber export logic

karthiksubraveti:

## Summary
Subscriber export was failing because we were trying to join 'state' attribute in subscriber entity instead of subscriber_config.lte entity. Cleaned up this logic a bit. 

## Test Plan
Was able to verify the behavior in staging.

## Additional Information


HannaFar (approved):
LGTM

PR #6335, opened 2021-04-23
AGW: pipelined: QoS: Fix max bandwidth calculation.

### LINKED ISSUE #6289: [agw][pipelined] QoS Enforcement is not working with pyroute2 enabled
### Your Environment

- **Version:** master
- **Affected Subcomponent:** Pipelined
- **Deployment Environment:**  Vagrant (AGW)

### Describe the Issue

In the default setting of `enable_pyroute2: true` in pipelined.yml file, QoS enforcement is failing. When the setting is changed to `enable_pyroute2: false`, QoS enforcement is working as expected. 

**To Reproduce**
1. Bring up AGW test set up with magma, magma_test and magma_trfserver VMs. 
2. Disable tcp checksum on all three machines via running the command `disable-tcp-checksumming`.
3. Make sure that magma services are running on magma VM and traffic server is started on magma_trfserver VM.
4. Run the integration test with `make integ_test TESTS=s1aptests/test_attach_detach_setsessionrules_tcp_data.py` under directory `magma/lte/gateway/python/integ_tests` on magma_test VM.

Test fails on asserting the achieved rate (that it should be less than the MBR set in the test case). Also check the TC rules and packet marking rules using. Observe that TC rules are wrong.

```
echo 'DUMPING ETH1'
sudo tc -s -d class show dev eth1
sudo tc -s -d filter show dev eth1

echo 'DUMPING ETH2'
sudo tc -s -d class show dev eth2
sudo tc -s -d filter show dev eth2

echo 'Dumping flows'
sudo ~/build/python/bin/pipelined_cli.py debug display_flows | grep 'PKT_MARK'
```

Change the setting in pipelined.yml file and disable pyroute, restart magma services, rerun the test on step 4. Observe that the test passes this time.

**Expected behavior**
Test should pass whether or not the flag `enable_pyroute2` set as true or false.

pshelar:
Pyroute2 API expects max-bandwidth parameter in bytes rather than
bit rate.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/qos/tc_ops_pyroute2.py
+++ b/lte/gateway/python/magma/pipelined/qos/tc_ops_pyroute2.py
@@ -37,9 +37,24 @@ class TcOpsPyRoute2(TcOpsBase):
 
     def create_htb(self, iface: str, qid: str, max_bw: int, rate: str,
                    parent_qid: str = None) -> int:
-        LOG.debug("Create HTB iface %s qid %s", iface, qid)
+        """
+        Create HTB class for a UE session.
 
+        Args:
+            iface: Egress interface name.
+            qid: qid number.
+            max_bw: ceiling in bits per sec.
+            rate: rate limiting.
+            parent_qid: HTB parent queue.
+
+        Returns:
+            zero on success.
+        """
+
+        LOG.debug("Create HTB iface %s qid %s max_bw %s rate %s", iface, qid, max_bw, rate)
         try:
+            # API needs ceiling in bytes per sec.
+            max_bw = max_bw / 8
             if_index = self._get_if_index(iface)
             htb_queue = QUEUE_PREFIX + qid
             ret = self._ipr.tc("add-class", "htb", if_index,
@@ -53,6 +68,15 @@ class TcOpsPyRoute2(TcOpsBase):
         return 0
 
     def del_htb(self, iface: str, qid: str) -> int:
+        """
+        Delete given queue from HTB classed
+
+        Args:
+            iface: interface name
+            qid: queue-id of the HTB class
+
+        Returns:
+        """
         LOG.debug("Delete HTB iface %s qid %s", iface, qid)
 
         try:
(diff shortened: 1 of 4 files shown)


lte/gateway/python/magma/pipelined/qos/tc_ops_pyroute2.py
```diff
@@ -37,9 +37,10 @@ def __init__(self):
 
     def create_htb(self, iface: str, qid: str, max_bw: int, rate: str,
                    parent_qid: str = None) -> int:
-        LOG.debug("Create HTB iface %s qid %s", iface, qid)
+        LOG.debug("Create HTB iface %s qid %s max_bw %s rate %s", iface, qid, max_bw, rate)
 
         try:
+            max_bw = max_bw / 8
```
koolzz, line 57: Maybe add a note that api needs bytes

koolzz (approved):

lte/gateway/python/magma/pipelined/qos/tc_ops_pyroute2.py
```diff
@@ -37,9 +37,23 @@ def __init__(self):
...
+            rate: rate limiting.
+            parent_qid: HTB parent queue.
+
+        Returns:
+            zero on success.
+        """
+
+        LOG.debug("Create HTB iface %s qid %s max_bw %s rate %s", iface, qid, max_bw, rate)
         try:
+            # API needs ceiling in KBytes per sec.
```
ulaskozat: KBytes or Bytes?

ulaskozat (approved):
There are some legit style guide comments. Otherwise looks good.

lte/gateway/python/magma/pipelined/qos/tc_ops_pyroute2.py
pshelar: right, its byte

ulaskozat:
@pshelar this should be also backported to 1.4 right?

themarwhal:
Removed the tag so it can be tracked here : https://github.com/magma/magma/issues/6827

PR #6349, opened 2021-04-23
[CWAG] Fix PipelineD build failure for CWAG

themarwhal:
Signed-off-by: Marie Bremner <marwhal@fb.com>

## Summary
the CWAG pipelined uses a focal image. https://github.com/magma/magma/issues/6152 split up python dependencies between focal and debian, which broke some python dependencies.

Also modifying the fabfile to only force provision if it is the first time accessing the VM. (was slowing down my testing) + ran ran formatter. (Will address other lints in a separate PR)

I think there's a better way to fix this, but since CWAG/XWF has been broken for ~2 weeks and we want a clean 1.5 release, I'm going to go with this approach unless someone suggests otherwise.

## Test Plan
`docker-compose build pipelined` in both `cwf/gateway/docker` and `xwf/docker` work
running integration tests...
[X] passes locally (DONE 40 tests, 11 skipped in 542.147s)
[ ] passes on circleci
[ ] passes on jenkins

## Additional Information

- [ ] This change is backwards-breaking

PR #6353, opened 2021-04-23
Clean up fabfile and add retry logic to cwf tests

themarwhal:

## Summary
* Addressing oriols comment from https://github.com/magma/magma/pull/6349#discussion_r619389802
  * force_provision -> provision_vm
* Also add the flag `--rerun-fails=1` to re-run any failed tests once. I'm hoping this will help with some flaky tests.  

## Test Plan
run integ test locally

## Additional Information

### DIFF
--- a/cwf/gateway/fabfile.py
+++ b/cwf/gateway/fabfile.py
@@ -11,10 +11,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 import sys
-sys.path.append('../../orc8r')
 import time
 from enum import Enum
-import os
+
 from fabric.api import (
     cd,
     env,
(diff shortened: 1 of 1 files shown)


cwf/gateway/fabfile.py
```diff
@@ -465,6 +465,8 @@ def _run_integ_tests(
 
     # QOS take a while to run. Increasing the timeout to 50m
     go_test_cmd = "gotestsum --format=standard-verbose "
+    # Retry once on failure
+    go_test_cmd += "--rerun-fails=1 --packages='./...' "
```
uri200, line 469: what is this?

uri200 (approved):

cwf/gateway/fabfile.py
themarwhal: it re-runs any failed tests

themarwhal: *at most once. so if it's --rerun-fails=5 it'll re-run until success at most 5 times

** it'll also only run tests that failed not the whole suite

themarwhal: @uri200 ^^

uri200: ya, not sure is that is very good... we shouldnt be failing
But i guess we dont have any other option :(

themarwhal: @uri200 yeah.. not ideal
This will only help with flaky tests on CI. But if we want to make integ tests mandatory, this will help.

PR #6364, opened 2021-04-23
[PipelineD][SessionD] Cleanup flows for non-existent sessions

themarwhal:
## Summary
Basically the same thing as this [PR](https://github.com/magma/magma/pull/5368/) I had to revert on master. (It broke some testing workflows for the PipelineD folks)

The main difference between this PR and the previous one is that this adds a configuration `cleanup_all_dangling_flows` that can be set to false. I've marked in the config file that it should only be used for testing and this is enabled by default.

Some new things:
* new config `cleanup_all_dangling_flows` if this is set to false, the clean up logic will be removed
* by default, when sessiond receives a rule with IMSI+ip that it doesn't recognize, it will send a deactivate flows request for that rule (TODO: migrate to teid based approach)
* new origin type "WILDTYPE" to remove flows from both GX and GY tables

## Test Plan
make precommit_sm
tested with s1ap tests to make sure flows get cleaned up
run cwf integ test

## Additional Information


uri200 (approved):

koolzz (approved):
Thanks for the debug flag!

ardzoht (approved):
lgtm

PR #6288, opened 2021-04-21
[feg][CI] Check FEG crash looping

uri200:
Signed-off-by: Oriol Batalla <obatalla@fb.com>

## Summary

This PR adds some endpoints to Makefile to deal with start, stop and health check of the processes

Note if you are running this on MAC you will have to comment the `LOG_DRIVER=journald` form the .env file at `~/magma/feg/gateway/docker`

Health Check will run docker-compose up -d and check for few seconds if any of the containers restarts

This PR also adds this check on CI in every PR on as part of the `feg-build` task

## Test Plan
`ci/circleci: feg-build`  will pass once this https://github.com/magma/magma/issues/6282 is fixed

Output of running the command manually
```
╰─ make docker_health_check
cd /Users/obatalla/magma/feg/gateway/docker; docker-compose up -d
WARNING: The LOG_DRIVER variable is not set. Defaulting to a blank string.
WARNING: The USE_GY_FOR_AUTH_ONLY variable is not set. Defaulting to a blank string.
WARNING: The GY_SUPPORTED_VENDOR_IDS variable is not set. Defaulting to a blank string.
WARNING: The GY_SERVICE_CONTEXT_ID variable is not set. Defaulting to a blank string.
aaa_server is up-to-date
eventd is up-to-date
s8_proxy is up-to-date
td-agent-bit is up-to-date
feg_hello is up-to-date
eap_aka is up-to-date
hss is up-to-date
Recreating redis ...
eap_sim is up-to-date
s6a_proxy is up-to-date
session_proxy is up-to-date
test is up-to-date
health is up-to-date
csfb is up-to-date
swx_proxy is up-to-date
radiusd is up-to-date
control_proxy is up-to-date
Recreating redis ... done
cd /Users/obatalla/magma/feg/gateway/docker/tools/; ./health_check_docker_ps.sh /Users/obatalla/magma/feg/gateway/docker
aaa_server      envdir /var/opt/magma/envd ...   Up
control_proxy   /bin/bash -c /usr/local/bi ...   Up
csfb            envdir /var/opt/magma/envd ...   Up
eap_aka         envdir /var/opt/magma/envd ...   Up
eap_sim         envdir /var/opt/magma/envd ...   Up
eventd          python3.8 -m magma.eventd.main   Up
feg_hello       envdir /var/opt/magma/envd ...   Up
health          envdir /var/opt/magma/envd ...   Up
hss             envdir /var/opt/magma/envd ...   Up
[...]

PR #6370, opened 2021-04-23
refactor: Add feg GitHub Actions Workflow 

jwagantall:
-Add initial feg GHA workflow.

These workflow is migrated from CircleCI.

## Additional Information

### DIFF
--- a/.github/workflows/feg-workflow.yml
+++ b/.github/workflows/feg-workflow.yml
@@ -1,0 +1,174 @@
+---
+name: feg-workflow
+
+# yamllint disable-line rule:truthy
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'orc8r/**'
+      - 'lte/**'
+      - 'feg/**'
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'orc8r/**'
+      - 'lte/**'
+      - 'feg/**'
+jobs:
+  feg-lint:
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on  # yamllint disable-line rule:truthy
+      MAGMA_ROOT: "${{ github.workspace }}"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.15.7'
+      - run: go version
+      - name: Run golang_before_install.sh script
+        run: |
+              ./circleci/golang_before_install.sh
+      - name: Run go mod download with retry
+        uses: nick-invision/retry@v2
+        with:
+          command: cd ${MAGMA_ROOT}/cwf/gateway && go mod download
+          timeout_minutes: 10
+      - name: Go lint code
+        run: |
+              cd ${MAGMA_ROOT}/feg/gateway
+              make -C ${MAGMA_ROOT}/feg/gateway lint
+      - name: Generate test coverage
+        run: |
+              cd ${MAGMA_ROOT}/feg/gateway
+              make -C ${MAGMA_ROOT}/feg/gateway cover
+      - uses: codecov/codecov-action@v1
+        with:
+          files: '${{ env.MAGMA_ROOT}}/feg/gateway/coverage/feg.gocov'
+          flags: feg-lint
+      # Notify ci channel when failing
+      # Plugin info: https://github.com/marketplace/actions/slack-notify
+      - name: Notify failure to slack
+        if: failure() && github.ref == 'refs/heads/master'
+        uses: rtCamp/action-slack-notify@v2.0.2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_CI }}
+          SLACK_TITLE: "Github action feg-lint update failed"
+          SLACK_USERNAME: "Feg workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+  feg-precommit:
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on  # yamllint disable-line rule:truthy
+      MAGMA_ROOT: "${{ github.workspace }}"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.15.7'
+      - run: go version
+      - name: Run golang_before_install.sh script
+        run: ./circleci/golang_before_install.sh
+      - name: Run go mod download with retry
+        uses: nick-invision/retry@v2
+        with:
+          command: cd ${MAGMA_ROOT}/cwf/gateway && go mod download
+          timeout_minutes: 10
+      - name: make feg precommit
+        run: |
+            go get gotest.tools/gotestsum
+            cd ${MAGMA_ROOT}/feg/gateway
+            make -C ${MAGMA_ROOT}/feg/gateway precommit
+      - name: Upload Test Results
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: Unit Test Results
+          path: "/tmp/test-results"
+      # Notify ci channel when failing
+      # Plugin info: https://github.com/marketplace/actions/slack-notify
+      - name: Notify failure to slack
+        if: failure() && github.ref == 'refs/heads/master'
+        uses: rtCamp/action-slack-notify@v2.0.2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_CI }}
+          SLACK_TITLE: "Github action feg-precommit update failed"
+          SLACK_USERNAME: "Feg workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+  feg-build:
+    needs: feg-precommit
+    runs-on: ubuntu-latest
+    env:
+      MAGMA_ROOT: "${{ github.workspace }}"
+      DOCKER_REGISTRY: "${{ secrets.DOCKER_REGISTRY }}"
+      DOCKER_USERNAME: "${{ secrets.DOCKER_USERNAME }}"
+      DOCKER_PASSWORD: "${{ secrets.DOCKER_PASSWORD }}"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
+      - name: generate test certs and snowflake
+        run: |
+            # TODO add rootCA.pem and snowflake files in the ubuntu-1604:201903-01 image
+            # create directories
+            cd ${MAGMA_ROOT} && mkdir -p .cache/test_certs/ && mkdir -p .cache/feg/
+            # create test certs
+            cd ${MAGMA_ROOT}/.cache/test_certs/
+            openssl genrsa -out rootCA.key 2048
+            openssl req -x509 -new -nodes -key rootCA.key -sha256 -days 365000 -out rootCA.pem -subj "/C=US/CN=rootca.magma.test"
+            # create snowflake
+            cd ${MAGMA_ROOT}/.cache/feg/ && touch snowflake
+      - name: Build docker images
+        env:
+          DOCKER_REGISTRY: feg_
+        run: |
+            cd ${MAGMA_ROOT}/feg/gateway/docker
+            python3 build.py
+      - name: run docker containers and check health
+        env:
+          DOCKER_REGISTRY: feg_
+        run: |
+            cd ${MAGMA_ROOT}/feg/gateway/docker
+            python3 build.py -e
+      - name: Tag and push to Docker Registry
+        if: github.ref == 'refs/heads/master'
+        # yamllint disable rule:line-length
+        run: |
+            ./ci-scripts/tag-push-docker.sh --images 'gateway_go|gateway_python' --tag "${GITHUB_SHA:0:8}" --tag-latest true --project feg
+      - name: Tag and push to Jfrog Registry
+        if: github.ref == 'refs/heads/master'
+        env:
+          DOCKER_REGISTRY: "${{ secrets.JFROG_DOCKER_ORC8R_REGISTRY }}"
+          DOCKER_USERNAME: "${{ secrets.JFROG_USERNAME }}"
+          DOCKER_PASSWORD: "${{ secrets.JFROG_PASSWORD }}"
+        run: |
+            ./ci-scripts/tag-push-docker.sh --images 'gateway_go|gateway_python' --tag "${GITHUB_SHA:0:8}" --tag-latest true --project feg
+      # Notify ci channel when failing
+      # Plugin info: https://github.com/marketplace/actions/slack-notify
+      # yamllint enable
+      - name: Notify failure to slack
+        if: failure() && github.ref == 'refs/heads/master'
+        uses: rtCamp/action-slack-notify@v2.0.2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_CI }}
+          SLACK_TITLE: "Github action feg-precommit update failed"
+          SLACK_USERNAME: "Feg workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+      # Notify ci channel when push succeeds
+      - name: Notify success to slack
+        if: success() && github.ref == 'refs/heads/master'
+        uses: rtCamp/action-slack-notify@v2.0.2
+        env:
+          SLACK_CHANNEL: ${{ secrets.SLACK_CHANNEL_ARTIFACTS }}
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_OSS }}
+          SLACK_TITLE: "*FeG Artifact Has Been Published*"
+          SLACK_USERNAME: "Feg workflow"
+          SLACK_ICON_EMOJI: ":heavy_check_mark:"
+          SLACK_COLOR: "#FF0000"
--- a/.github/workflows/unit-test-workflow.yml
+++ b/.github/workflows/unit-test-workflow.yml
@@ -2,7 +2,9 @@
 name: Cloud Unit Test Results
 on:  # yamllint disable-line rule:truthy
   workflow_run:
-    workflows: ["cloud-workflow"]
+    workflows:
+      - cloud-workflow
+      - feg-workflow
     types:
       - completed
 
@@ -29,5 +31,6 @@ jobs:
       - name: Publish Unit Test Results
         uses: EnricoMi/publish-unit-test-result-action@v1
         with:
+          check_name: ${{ github.event.workflow.name }}
           commit: ${{ github.event.workflow_run.head_sha }}
           files: "artifacts/**/*.xml"


.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,70 @@
...
+        timeout_minutes: 10
+    - name: make feg precommit
+      run: |
+          cd ${MAGMA_ROOT}/feg/gateway
+          make -C ${MAGMA_ROOT}/feg/gateway precommit
+#    - magma_slack_notify
+
+########## FEG-BUILD ##########
+
+  feg-build:
```
uri200, line 103: note this flow will change on Circle CI

https://github.com/magma/magma/pull/6288/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R715

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,70 @@
...
+
+    needs: feg-precommit
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+#      - build/determinator:
+#          paths: "cwf/k8s"
```
bramwelt: Github Actions support limiting workflows to changes to [specific files](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths) on pushes and pull requests (though not tags).

I'm not certain if both these jobs are supposed to be triggered when changes happen to files under `cwf/k8s`, or if only _this_ job should be triggered. If it's the latter, then you may need to look into using a [label action](https://github.com/marketplace/actions/labeler) to signify a PR made changes to a file, followed by an action that's triggered on those [labels](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#label).

.github/workflows/feg-workflow.yml
jwagantall: Thanks for this. Let me update my PR

.github/workflows/feg-workflow.yml
jwagantall: Thanks Trevor!.. I just double checked and all feg jobs should be triggered for paths under *federated_build_verify

In that case, should I still look into the label action?

bramwelt: If it's all the jobs, then you should be able to trigger on the [paths](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths) and not worry about labeling.

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,179 @@
...
+########## FEG-BUILD ##########
+
+  feg-build:
+
+    needs: feg-precommit
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+#      - build/determinator:
```
tmdzk: Can you remobe that?

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,179 @@
+name: feg-workflow
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
tmdzk, line 15: can we add that: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,179 @@
...
+            if [ ! -z $PROJECT ]; then
+              IMAGE_TOSEARCH="${PROJECT}_${IMAGE}"
+            fi
+            IMAGE_ID=$(docker images "$IMAGE_TOSEARCH:latest" --format "{{.ID}}")
+            tag_and_push "$TAG"
+            if [ "$TAG_LATEST" = true ]; then
+              tag_and_push "latest"
+            fi
+          done
+    - name: tag-push-docker
```
tmdzk: redundant no?

tmdzk (changes requested):

tmdzk:
Also can you commit something to re-trigger the DCO check?

.github/workflows/feg-workflow.yml
jwagantall: Yea,, could be redundant but the tag and push is done in two different registries. 
This is where I wish Anchors could help out so we don't have to re-write code :(

jwagantall (commented):
@tmdzk  thanks for your comments. I am updating this workflow now since it was missing some of the parts with regards to the build determinator for example. 

Let me update it again and it should trigger the DCO

jwagantall:
@tmdzk , on regards to your question about the redundancy with the docker push and tag, the original workflow does a push using the docker registry and a 2nd push to the Jfrog registry. This is why I am adding the 2 pushes. This is an example I was having issues with becuase we can't use anchors, we need to copy the whole step twice (one per registry)

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,137 @@
...
+            ./ci-scripts/tag-push-docker.sh -i 'gateway_go|gateway_python' -t '${GITHUB_SHA:0:8}' -tl false -p feg
+      - name: persist-githash-version - Set prefix
+        run: |
+            echo "file_prefix=feg" >> $GITHUB_ENV
+      - name: persist-githash-version - Create version file
+        run: |
+            cd ${MAGMA_ROOT}/circleci
+            mkdir -p versions
+            echo "${GITHUB_SHA:0:8}" > versions/<< ${file_prefix} >>_version
+      # Notify ci channel when failing
```
quentinDERORY, line 152: There is one step missing where you notify the ci-artifact slack channel that the artifact was correctly published

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,137 @@
...
+            # create snowflake
+            cd ${MAGMA_ROOT}/.cache/feg/ && touch snowflake
+      - name: Export Docker registry
+        run: |
+            cd ${MAGMA_ROOT}/feg/gateway/docker
+            echo "DOCKER_REGISTRY=feg_ docker-compose build --parallel" >> $GITHUB_ENV
+      - name: run docker containers and check health
+        run: |
+            cd ${MAGMA_ROOT}/feg/gateway
+            echo "DOCKER_REGISTRY=feg_ make -C ${MAGMA_ROOT}/feg/gateway docker_healthcheck" >> $GITHUB_ENV
```
quentinDERORY: Looks like it shoudl be this instead
`DOCKER_REGISTRY=feg_ make -C ${MAGMA_ROOT}/feg/gateway docker_healthcheck`

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,137 @@
...
+            # create test certs
+            cd ${MAGMA_ROOT}/.cache/test_certs/
+            openssl genrsa -out rootCA.key 2048
+            openssl req -x509 -new -nodes -key rootCA.key -sha256 -days 365000 -out rootCA.pem -subj "/C=US/CN=rootca.magma.test"
+            # create snowflake
+            cd ${MAGMA_ROOT}/.cache/feg/ && touch snowflake
+      - name: Export Docker registry
+        run: |
+            cd ${MAGMA_ROOT}/feg/gateway/docker
+            echo "DOCKER_REGISTRY=feg_ docker-compose build --parallel" >> $GITHUB_ENV
```
quentinDERORY: Looks like it should be this instead 
`DOCKER_REGISTRY=feg_ docker-compose build --parallel`

quentinDERORY (changes requested):
See my comments about the missing slack notifications and feg build steps

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,137 @@
...
+      - name: run docker containers and check health
+        run: |
+            cd ${MAGMA_ROOT}/feg/gateway
+            echo "DOCKER_REGISTRY=feg_ make -C ${MAGMA_ROOT}/feg/gateway docker_healthcheck" >> $GITHUB_ENV
+      - name: Tag and push to Docker Registry
+        run: |
+            ./ci-scripts/tag-push-docker.sh -i 'gateway_go|gateway_python' -t '${GITHUB_SHA:0:8}' -tl false -p feg
+      - name: persist-githash-version - Set prefix
+        run: |
+            echo "file_prefix=feg" >> $GITHUB_ENV
```
quentinDERORY: don't need one step to export file_prefix, just use it direclty below

.github/workflows/feg-workflow.yml
jwagantall: Updated to the latest master https://github.com/magma/magma/pull/7836/files

.github/workflows/feg-workflow.yml
jwagantall: fixed

.github/workflows/feg-workflow.yml
jwagantall: Updated to the latest master https://github.com/magma/magma/pull/7836/files

.github/workflows/feg-workflow.yml
jwagantall: added

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,148 @@
...
+            # create test certs
+            cd ${MAGMA_ROOT}/.cache/test_certs/
+            openssl genrsa -out rootCA.key 2048
+            openssl req -x509 -new -nodes -key rootCA.key -sha256 -days 365000 -out rootCA.pem -subj "/C=US/CN=rootca.magma.test"
+            # create snowflake
+            cd ${MAGMA_ROOT}/.cache/feg/ && touch snowflake
+      - name: Export Docker registry
+        run: |
+            cd ${MAGMA_ROOT}/feg/gateway/docker
+            echo "DOCKER_REGISTRY=feg_ python3 build.py" >> $GITHUB_ENV
```
quentinDERORY: This should be this command.

```
      - name: Export Docker registry
        env:
          DOCKER_REGISTRY: feg_
        run: |
            cd ${MAGMA_ROOT}/feg/gateway/docker
            python3 build.py
```

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,148 @@
...
+      - name: run docker containers and check health
+        run: |
+            cd ${MAGMA_ROOT}/feg/gateway/docker
+            echo "DOCKER_REGISTRY=feg_ python3 build.py -e" >> $GITHUB_ENV
```
quentinDERORY: This should be this command.

```
      - name: Export Docker registry
        env:
          DOCKER_REGISTRY: feg_
        run: |
            cd ${MAGMA_ROOT}/feg/gateway/docker
            python3 build.py -e
```

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,148 @@
...
+      - name: Tag and push to Docker Registry
+        run: |
+            ./ci-scripts/tag-push-docker.sh -i 'gateway_go|gateway_python' -t '${GITHUB_SHA:0:8}' -tl false -p feg
```
quentinDERORY: We have to call this script twice one with the default docker and one with the jfrog docker credentials.
Use step specific environment variable to propagate them

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,148 @@
...
+      - name: persist-githash-version - Create version file
```
quentinDERORY: There is no need to persist the githash version, you need to include the version in the successful slack message.
See here for content of message: https://github.com/magma/magma/blob/ee7c301e4216a61ca32e0ed701194a265c317913/.circleci/config.yml#L503

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,148 @@
...
+        uses: rtCamp/action-slack-notify@v2.0.2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_CI }}
+          SLACK_TITLE: "Github action feg-precommit update failed"
+          SLACK_USERNAME: "Feg workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+  feg-build:
+    needs: feg-precommit
+    runs-on: ubuntu-latest
```
quentinDERORY, line 105: add this to avoid uploading builds from PRs: `if: github.ref == 'refs/heads/master'`

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,148 @@
...
+          SLACK_TITLE: "Github action feg-precommit update failed"
+          SLACK_USERNAME: "Feg workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+      # Notify ci channel when push succeeds
+      - name: Notify success to slack
+        if: success() && github.ref == 'refs/heads/master'
+        uses: rtCamp/action-slack-notify@v2.0.2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_CI }}
```
quentinDERORY: You must specify this WEBHOOK secret: SLACK_WEBHOOK_OSS
We are publishing successful artifact upload to ci-artifact channel

jwagantall: Ah!.. thank you! let me fix that.

.github/workflows/feg-workflow.yml
jwagantall: fixed

.github/workflows/feg-workflow.yml
jwagantall: fixed

jwagantall:
@quentinDERORY . .I might need your help on the attachment part of the Slack message.. do you have an example I can see?

Also,, i will modify the triggers for now so I can see this workflow running. I will remove the paths but I will make sure I add them later when the workflow is ready

jwagantall:
@quentinDERORY .. finally! 
here is a better run: https://github.com/magma/magma/runs/3005544313?check_suite_focus=true

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,162 @@
...
+      - name: Tag and push to Docker Registry
+        run: |
+            ./ci-scripts/tag-push-docker.sh -i 'gateway_go|gateway_python' -t '${GITHUB_SHA:0:8}' -tl false -p feg
+      - name: Tag and push to Jfrog Registry
+        env:
+          DOCKER_REGISTRY: "${{ secrets.JFROG_DOCKER_ORC8R_REGISTRY }}"
+          DOCKER_USERNAME: "${{ secrets.JFROG_USERNAME }}"
+          DOCKER_PASSWORD: "${{ secrets.JFROG_PASSWORD }}"
+        run: |
+            ./ci-scripts/tag-push-docker.sh -i 'gateway_go|gateway_python' -t '${GITHUB_SHA:0:8}' -tl false -p feg
```
quentinDERORY: Use long options name like --tag-latest true instead of just -tl
And actually switch the tag-latest to true

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,162 @@
...
+            # TODO add rootCA.pem and snowflake files in the ubuntu-1604:201903-01 image
+            # create directories
+            cd ${MAGMA_ROOT} && mkdir -p .cache/test_certs/ && mkdir -p .cache/feg/
+            # create test certs
+            cd ${MAGMA_ROOT}/.cache/test_certs/
+            openssl genrsa -out rootCA.key 2048
+            openssl req -x509 -new -nodes -key rootCA.key -sha256 -days 365000 -out rootCA.pem -subj "/C=US/CN=rootca.magma.test"
+            # create snowflake
+            cd ${MAGMA_ROOT}/.cache/feg/ && touch snowflake
+      - name: Export Docker registry
```
quentinDERORY: This step is currently called 'Build docker', maybe keep it that way or name it 'Build docker images'

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,162 @@
...
+        uses: rtCamp/action-slack-notify@v2.0.2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_CI }}
+          SLACK_TITLE: "Github action feg-precommit update failed"
+          SLACK_USERNAME: "Feg workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+      # Notify ci channel when push succeeds
+      - name: Notify success to slack
+        if: success() && github.ref == 'refs/heads/master'
```
quentinDERORY, line 166: don't need to specify `github.ref == 'refs/heads/master'` as the job will only run on master

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,162 @@
...
+        env:
+          DOCKER_REGISTRY: "${{ secrets.JFROG_DOCKER_ORC8R_REGISTRY }}"
+          DOCKER_USERNAME: "${{ secrets.JFROG_USERNAME }}"
+          DOCKER_PASSWORD: "${{ secrets.JFROG_PASSWORD }}"
+        run: |
+            ./ci-scripts/tag-push-docker.sh -i 'gateway_go|gateway_python' -t '${GITHUB_SHA:0:8}' -tl false -p feg
+      # Notify ci channel when failing
+      # Plugin info: https://github.com/marketplace/actions/slack-notify
+      - name: Notify failure to slack
+        if: failure() && github.ref == 'refs/heads/master'
```
quentinDERORY, line 156: same thing for `github.ref == 'refs/heads/master'`

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,162 @@
...
+      - name: make feg precommit
+        run: |
+            go get gotest.tools/gotestsum
+            cd ${MAGMA_ROOT}/feg/gateway
+            make -C ${MAGMA_ROOT}/feg/gateway precommit
+      - name: Publish Unit Test Results
+        uses: EnricoMi/publish-unit-test-result-action@v1
+        if: always()
+        with:
+          files: /tmp/test-results
```
quentinDERORY: Add a relevant name to the artifact like feg-precommit-test-results

.github/workflows/feg-workflow.yml
jwagantall: fixed

.github/workflows/feg-workflow.yml
jwagantall: fixed

.github/workflows/feg-workflow.yml
jwagantall: fixed

.github/workflows/feg-workflow.yml
jwagantall: fixed

.github/workflows/feg-workflow.yml
jwagantall: fixed

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,164 @@
...
+      - name: make feg precommit
+        run: |
+            go get gotest.tools/gotestsum
+            cd ${MAGMA_ROOT}/feg/gateway
+            make -C ${MAGMA_ROOT}/feg/gateway precommit
+      - name: Publish Unit Test Results
+        uses: EnricoMi/publish-unit-test-result-action@v1
+        if: always()
+        with:
+          files: /tmp/feg-precommit-test-results
```
quentinDERORY: Can you add explicit name to the artifact uploaded

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,164 @@
...
+      - name: run docker containers and check health
+        env:
+          DOCKER_REGISTRY: feg_
+        run: |
+            cd ${MAGMA_ROOT}/feg/gateway/docker
+            python3 build.py -e
+      - name: Tag and push to Docker Registry
+        # yamllint disable rule:line-length
+        run: |
+            ./ci-scripts/tag-push-docker.sh --images 'gateway_go|gateway_python' --tag '${GITHUB_SHA:0:8}' --tag-latest false --project feg
```
quentinDERORY: tag latest to true here too

jwagantall: sure!

.github/workflows/feg-workflow.yml
quentinDERORY: Sorry to go back and forth on this but could we trigger the build part with pull request event but just not the push and notification.
TLDR: put the if statement on the tag&push and notifications steps rather than here

quentinDERORY:
For the upload test result workflow (.github/workflows/unit-test-workflow.yml):
Add a name to the uploaded test results so we can differentiate between feg and cloud test results.
ie add the `check_name` option to the https://github.com/EnricoMi/publish-unit-test-result-action module and use the github variable  `github.event.workflow.name` .
So it looks like `feg unit tests` or 'cloud unit tests'.

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,169 @@
...
+        run: |
+            go get gotest.tools/gotestsum
+            cd ${MAGMA_ROOT}/feg/gateway
+            make -C ${MAGMA_ROOT}/feg/gateway precommit
+      - name: Upload Test Results
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: Unit Test Results
+          path: "/tmp/feg-precommit-test-results"
```
quentinDERORY: The correct path is `/tmp/test-results`

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,169 @@
...
+      - name: run docker containers and check health
+        env:
+          DOCKER_REGISTRY: feg_
+        run: |
+            cd ${MAGMA_ROOT}/feg/gateway/docker
+            python3 build.py -e
+      - name: Tag and push to Docker Registry
+        # yamllint disable rule:line-length
+        run: |
+            ./ci-scripts/tag-push-docker.sh --images 'gateway_go|gateway_python' --tag '${GITHUB_SHA:0:8}' --tag-latest true --project feg
```
quentinDERORY: use double quotes to enclose GITHUB_SHA

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,169 @@
...
+      - name: Tag and push to Jfrog Registry
+        env:
+          DOCKER_REGISTRY: "${{ secrets.JFROG_DOCKER_ORC8R_REGISTRY }}"
+          DOCKER_USERNAME: "${{ secrets.JFROG_USERNAME }}"
+          DOCKER_PASSWORD: "${{ secrets.JFROG_PASSWORD }}"
+        run: |
+            ./ci-scripts/tag-push-docker.sh --images 'gateway_go|gateway_python' --tag '${GITHUB_SHA:0:8}' --tag-latest true --project feg
```
quentinDERORY: same here

.github/workflows/feg-workflow.yml
jwagantall: fixed

.github/workflows/feg-workflow.yml
jwagantall: fixed

.github/workflows/feg-workflow.yml
jwagantall: fixed

.github/workflows/feg-workflow.yml
jwagantall: I agree.. like we did in the other workflow... thank you!

quentinDERORY: You can remove the if statement from there now

.github/workflows/feg-workflow.yml
```diff
@@ -0,0 +1,170 @@
...
+          SLACK_TITLE: "Github action feg-precommit update failed"
+          SLACK_USERNAME: "Feg workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+      # Notify ci channel when push succeeds
+      - name: Notify success to slack
+        if: success() && github.ref == 'refs/heads/master'
+        uses: rtCamp/action-slack-notify@v2.0.2
+        env:
+          SLACK_CHANNEL: ${SLACK_CHANNEL_ARTIFACTS}
```
quentinDERORY: should be ${{ secrets. SLACK_CHANNEL_ARTIFACTS }} 
Can you correct it in the cloud workflow that is already merged.

.github/workflows/unit-test-workflow.yml
```diff
@@ -29,5 +31,6 @@ jobs:
       - name: Publish Unit Test Results
         uses: EnricoMi/publish-unit-test-result-action@v1
         with:
+          check_name: github.event.workflow.name
```
quentinDERORY: Can you put it between ${{ }} , just to be consistent with the line underneath

quentinDERORY (approved):

PR #6371, opened 2021-04-24
refactor: Add cwf GitHub Actions Workflow

jwagantall:
-Add initial cwf-operator GHA workflow.

These workflow is migrated from CircleCI.

## Additional Information

### DIFF
--- a/.github/workflows/cwf-operator.yml
+++ b/.github/workflows/cwf-operator.yml
@@ -1,0 +1,98 @@
+---
+name: cwf-operator
+
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - master
+    paths:
+      - 'cwf/**'
+      - 'k8s/**'
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'cwf/**'
+      - 'k8s/**'
+jobs:
+  cwf-operator-precommit:
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on  # yamllint disable-line rule:truthy
+      MAGMA_ROOT: "${{ github.workspace }}"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.15.7'
+      - name: Run golang_before_install.sh script
+        run: ./circleci/golang_before_install.sh
+      - name: Run go mod download with retry
+        uses: nick-invision/retry@v2
+        with:
+          command: cd ${MAGMA_ROOT}/cwf/k8s/cwf_operator && go mod download
+          timeout_minutes: 10
+      - name: Run precommit
+        run: |
+            cd ${MAGMA_ROOT}/cwf/k8s/cwf_operator
+            make -C ${MAGMA_ROOT}/cwf/k8s/cwf_operator precommit
+      # Notify ci channel when failing
+      # Plugin info: https://github.com/marketplace/actions/slack-notify
+      - name: Notify failure to slack
+        if: failure() && github.ref == 'refs/heads/master'
+        uses: rtCamp/action-slack-notify@v2.0.2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_CI }}
+          SLACK_TITLE: "Github action cwf-operator-precommit update failed"
+          SLACK_USERNAME: "Cwf workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+  cwf-operator-build:
+    needs: cwf-operator-precommit
+    runs-on: ubuntu-latest
+    env:
+      MAGMA_ROOT: "${{ github.workspace }}"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run docker compose build
+        env:
+          DOCKER_REGISTRY: cwf_
+        run: |
+             cd ${MAGMA_ROOT}/cwf/k8s/cwf_operator/docker
+             DOCKER_REGISTRY=cwf_ docker-compose build --parallel
+      - name: Tag and push to Docker Registry
+        if: github.ref == 'refs/heads/master'
+        # yamllint disable rule:line-length
+        run: |
+            ./ci-scripts/tag-push-docker.sh --images 'operator' --tag "${GITHUB_SHA:0:8}" --tag-latest true --project cwf
+      - name: Tag and push to Jfrog Registry
+        if: github.ref == 'refs/heads/master'
+        env:
+          DOCKER_REGISTRY: "${{ secrets.JFROG_DOCKER_ORC8R_REGISTRY }}"
+          DOCKER_USERNAME: "${{ secrets.JFROG_USERNAME }}"
+          DOCKER_PASSWORD: "${{ secrets.JFROG_PASSWORD }}"
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/cwf-operator.yml
```diff
@@ -0,0 +1,75 @@
...
+      run: cd ${MAGMA_ROOT}/cwf/k8s/cwf_operator
+    - name: Run go mod download with retry
+      uses: nick-invision/retry@v2
+      with:
+        command: cd ${MAGMA_ROOT}/cwf/k8s/cwf_operator && go mod download
+        timeout_minutes: 10
+    - name: Run precommit
+      run: |
+          make -C ${MAGMA_ROOT}/cwf/k8s/cwf_operator precommit
+#    - magma_slack_notify
```
bramwelt: Currently Circle-CI will post a notification to Slack on failed builds on master. Because this workflow runs on both PRs and merges, you'll want to restrict the action to only trigger when those two conditions are true (refs == master && job failed).

There is the option of setting up Slack integration with the [Github app](https://github.com/integrations/slack#configuration), but from what I could see from the configuration, there's no way to configure it to notify on changes from checks.

Since there's _several_ actions for sending [Slack notifications](https://github.com/marketplace?type=actions&query=Slack), I'd suggest going with one of the top supported/starred ones: [rtCamp/action-slack-notify](https://github.com/marketplace/actions/slack-notify) or [8398a7/notify-slack](https://github.com/marketplace/actions/action-slack). From what I can see they all work essentially the same way, requiring a `SLACK_WEBHOOK` secret and some optional configuration for how messages should be formatted.

Here's an example using the `rtCamp` action:
```
- name: Slack Notification
  uses: rtCamp/action-slack-notify@v2
  env:
      SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
  if: (failure() || cancelled()) && ${{ github.ref == 'refs/heads/main' }}
 ```
 
Notifications also happen when new artifacts are created after a merge to master, which either of the slack notification actions I linked should support, though it looks like the one from 8398a7 provides a bit more flexibility on the message format.

File: .github/workflows/cwf-operator.yml [hunk 2]
```diff
@@ -0,0 +1,75 @@
+name: cwf-operator
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches: [ master ]
```
bramwelt: For consistency, can this be changed to
```
branches:
  - master
```

.github/workflows/cwf-operator.yml
```diff
@@ -0,0 +1,75 @@
...
+           cd ${MAGMA_ROOT}/cwf/k8s/cwf_operator/docker
+           docker-compose up -d
+    - name: Run docker compose build
+      run: |
+           cd ${MAGMA_ROOT}/cwf/k8s/cwf_operator/docker
+           DOCKER_REGISTRY=cwf_ docker-compose build --parallel
+#      - tag-push-docker:
+#          project: cwf
+#          images: "operator"
+#          registry: $DOCKER_MAGMA_REGISTRY
```
bramwelt: Even if this is waiting on secrets, it should still be filled out with the `run` command to tag and push the containers.

bramwelt (changes requested):

.github/workflows/cwf-operator.yml
jwagantall: addressed

.github/workflows/cwf-operator.yml
jwagantall: You are right... will do that. 
Hopefully anchors can be a thing soon... this particular block is aprox 30 lines of code :/

.github/workflows/cwf-operator.yml
```diff
@@ -0,0 +1,100 @@
...
+           cd ${MAGMA_ROOT}/cwf/k8s/cwf_operator/docker
+           DOCKER_REGISTRY=cwf_ docker-compose build --parallel
+    - name: Login to Docker registry
+      uses: docker/login-action@v1
+      with:
+          registry: ${{ DOCKER_MAGMA_REGISTRY }}
+          username: ${{ USERNAME }}
+          password: ${{ PASSWORD }}
+    - name: Build and push
+      uses: docker/build-push-action@v2
```
jwagantall: Not sure if this is going to work.. I will be able to test it once we have the secrets in place but I am thinking if we can only use the push part of this action and not the build since we are already doing the build in line 64

If not, I will need to figure how to configure this action to run the docker compose command like we need to.

.github/workflows/cwf-operator.yml
```diff
@@ -0,0 +1,100 @@
...
+           cd ${MAGMA_ROOT}/cwf/k8s/cwf_operator/docker
+           docker-compose up -d
+    - name: Run docker compose build
+      run: |
+           cd ${MAGMA_ROOT}/cwf/k8s/cwf_operator/docker
+           DOCKER_REGISTRY=cwf_ docker-compose build --parallel
+    - name: Login to Docker registry
+      uses: docker/login-action@v1
+      with:
+          registry: ${{ DOCKER_MAGMA_REGISTRY }}
```
fabiopalumbo: usually is ${{ secrets.DOCKER_MAGMA_REGISTRY }}

jwagantall: Thanks!.. let me fix this..

.github/workflows/cwf-operator.yml
```diff
@@ -0,0 +1,93 @@
...
+    needs: cwf-operator-precommit
+    runs-on: ubuntu-latest
+    env:
+      MAGMA_ROOT: "${{ github.workspace }}"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run docker compose build
+        run: |
+             cd ${MAGMA_ROOT}/cwf/k8s/cwf_operator/docker
+             DOCKER_REGISTRY=cwf_ docker-compose build --parallel
```
quentinDERORY, line 62: Can you export DOCKER_REGISTRY as a step environment variable?

jwagantall: oops. i forgot.. thanks!

quentinDERORY: can you delete it from the run command and add the environment variable only for the step

jwagantall: sure!

.github/workflows/cwf-operator.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Notify success to slack
+        if: success() && github.ref == 'refs/heads/master'
+        uses: rtCamp/action-slack-notify@v2.0.2
+        env:
+          SLACK_CHANNEL: ${{ secrets.SLACK_CHANNEL_ARTIFACTS }}
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_OSS }}
+          SLACK_TITLE: "*Cwf Artifact Has Been Published*"
+          SLACK_USERNAME: "Cwf workflow"
+          SLACK_ICON_EMOJI: ":heavy_check_mark:"
+          SLACK_COLOR: "#FF0000"
```
quentinDERORY: Can this be green and not red

.github/workflows/cwf-operator.yml
quentinDERORY: Can you remove the export from the run command?

.github/workflows/cwf-operator.yml
jwagantall: sure!.. i have also added the paths too so we can be ready to merge

quentinDERORY (approved):

PR #6376, opened 2021-04-24
[feg] fix dropped test error

alrs:
Signed-off-by: Lars Lehtonen <lars.lehtonen@gmail.com>

## Summary

This fixes a dropped test error, and uses `testify/require` to ensure the test stops immediately on failure.

## Test Plan

This fixes an existing unit test. I ran `go test ./...` locally.

### DIFF
--- a/feg/gateway/diameter/connection_manager_test.go
+++ b/feg/gateway/diameter/connection_manager_test.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


uri200:
Thanks @alrs for your contribution! 
In case you wanted to run all format tools and test locally you can do this
```
cd  ~/magma/fe/gateway/docker
docker-compose build test 
docker-compose exec test bash
#inside the container on /magma/feg/gateway run the tests
make precommit
``

uri200 (approved):

PR #6393, opened 2021-04-26
[Orc8r] Use gotestsum to run tests to generate summary XML

### LINKED ISSUE #6389: [Orc8r][UnitTests] Upload test summaries in `cloud-test` step
POC: @themarwhal (GH) Marie Bremner (Slack)

We have roughly ~100 tests running in the orc8r `cloud-test` precommit check in CI. It would be good to start uploading test summaries to CI so that flaky/failing/slow tests can be identified more quickly.

[gotestsum](https://github.com/gotestyourself/gotestsum) is a tool we use in the CWF integration test to generate XML files that can be uploaded + understood by CI. 

Suggested Steps:
* Figure out how Orc8r go unit tests are run and replace `go test ./...` with something like `gotestsum --format=standard-verbose --packages='./...'`. 
* Once the tests are running properly, add this flag to add the summary generation. `--junitfile <PATH>`
* In CircleCI, upload the test summary by using this command: `store_test_results`

Useful references:
* https://circleci.com/blog/level-up-go-test-with-gotestsum/

themarwhal:

## Summary
This PR adds all necessary changes to produce test summaries on cloud test so it can be uploaded to CircleCI. This will allow CircleCI to display the failure cases nicely and do analytics over time. (Junit format is widely accepted, so this should work if/when we migrate to Jenkins/GH action)

In summary, the benefits are
* Generating and uploading test summaries that is easily understood by CI platforms so we can monitor test health/runtime over time. 
  * With CircleCI, this is automatically done with CircleCI insights
  * With GithubAction, we might have to figure out a good platform to upload these summaries so we can get something like CircleCI insights
* [gotestsum](https://github.com/gotestyourself/gotestsum) produces a cleaner test output :) It also has a few other configurations that are nice to use. 

## Test Plan
Test locally
Screen Shot 2021-04-27 at 11 04 05 AM

CI success case:
Screen Shot 2021-04-27 at 11 23 30 AM

CI failure case:
Screen Shot 2021-04-27 at 11 34 44 AM

## Additional Information


uri200 (approved):

hcgatewood (approved):
Ah this is huge, thanks Marie. Couple notes

- From offline: I don't anticipate this causing any issues with the shared IntelliJ run configs. If it does, I can fix the run configs later
- Does GH Actions support this type of functionality? I'll be sad to lose it when we transition our CircleCI tests over to GH Actions
- In terms of functionality -- can you share a link with the enumerated benefits of this change

orc8r/cloud/docker/build.py
```diff
@@ -99,7 +99,7 @@ def main() -> None:
     elif args.tests:
         _run(['up', '-d', 'postgres_test'])
         _run(['build', 'test'])
-        _run(['run', '--rm', 'test', 'make test'])
+        _run(['run', '--rm'] + _get_mnt_vols(mods) + ['test', 'make test'])
```
hcgatewood: Okay actually one thing -- we're intentionally *not* mounting any volumes here, because it breaks the "clean room" effect. If you want to just exfiltrate the junit files, mount that directory specifically

hcgatewood (changes requested):

orc8r/cloud/docker/build.py
themarwhal: good point. Will fix :)

themarwhal:
> Ah this is huge, thanks Marie. Couple notes
> 
> * From offline: I don't anticipate this causing any issues with the shared IntelliJ run configs. If it does, I can fix the run configs later
> * Does GH Actions support this type of functionality? I'll be sad to lose it when we transition our CircleCI tests over to GH Actions
> * In terms of functionality -- can you share a link with the enumerated benefits of this change

In terms of CircleCI+Insights, here is a nice article I read when I initially found out about this feature: https://circleci.com/blog/level-up-go-test-with-gotestsum/ . 
It can highlight most frequently failing tests like this:
Screen Shot 2021-04-27 at 1 02 22 PM
Also a table of slowest running tests:
Screen Shot 2021-04-27 at 1 09 27 PM


In terms of GitHub action, looks like there's one action that can use this info to annotate PRs (https://github.com/marketplace/actions/junit-report-action). Not sure if I love the annotations, but it could be something. There might be other places where we could upload the summaries to though. Would have to look into it.

In summary, the benefits are
* Generating and uploading test summaries that is easily understood by CI platforms so we can monitor test health/runtime over time. 
  * With CircleCI, this is automatically done with CircleCI insights
  * With GithubAction, we might have to figure out a good platform to upload these summaries so we can get something like CircleCI insights
* [gotestsum](https://github.com/gotestyourself/gotestsum) produces a cleaner test output :) It also has a few other configurations that are nice to use.

hcgatewood (approved):
💯

tmdzk (approved):

PR #6412, opened 2021-04-27
[Sentry][Native] Update version 0.4.8 -> 0.4.9

themarwhal:

## Summary
Update Sentry Native version from 0.4.8 -> 0.4.9

## Test Plan
sentry still builds
MME/SessionD still builds and runs with Sentry

## Additional Information

### DIFF
--- a/lte/gateway/deploy/roles/dev_common/tasks/main.yml
+++ b/lte/gateway/deploy/roles/dev_common/tasks/main.yml
@@ -273,7 +273,7 @@
 
 - name: Install Sentry Native SDK
   vars:
-    sentry_native_version: "0.4.8"
+    sentry_native_version: "0.4.9"
   get_url:
     url: "https://github.com/getsentry/sentry-native/releases/download/{{ sentry_native_version }}/sentry-native.zip"
     dest: "{{ tmp_directory }}/sentry-native.zip"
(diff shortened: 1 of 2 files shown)


lte/gateway/deploy/roles/dev_common/tasks/main.yml
```diff
@@ -272,7 +272,7 @@
 
 - name: Install Sentry Native SDK
   vars:
-    sentry_native_version: "0.4.8"
+    sentry_native_version: "0.4.9"
```
pshelar, line 276: currently we have separate ways to install sentry in production and dev environment. It would be nice to use same mechanism for both. Can you add it ad TODO.

pshelar (approved):

lte/gateway/deploy/roles/dev_common/tasks/main.yml
themarwhal: how do we install sentry for prod? @pshelar 
Yeah will add a todo

pshelar: package is built using: https://github.com/magma/magma/blob/master/third_party/build/bin/sentry-native-build.sh
This is deployed in production. we could deploy same in vagrant VM.

ekowtaylor:
Once this PR has been approved and passing all CI checks, use the ready2merge label to indicate the maintainers can merge your PR.

PR #6417, opened 2021-04-27
[lte][agw] Updating state conversion on SPGW to be aligned with state_converter interface

ardzoht:

## Summary

- Current usage of SPGW UE state relies on 2 hashtables (`state_imsi_ht` and `state_ue_ht`), the naming for these hashtables is confusing as they are being used for inverse purposes (which is very confusing to read, and in unit testing the interface of the state manager becomes difficult to be aligned with). 
- This PR updates `state_imsi_ht` to `state_teid_ht` (as it's actually keyed by TEID), and `state_ue_ht` is used for IMSI keyed UE context.
- This also updates `spgw_state_converter` to remove unused `spgw_state` in different functions, and uses `spgw_ue_context` as `state_manager` interface defines it.
- Removes dead code used on `sgw_context_manager.c`

## Test Plan

- make integ_test
- ensure all sanity testing passes 

## Additional Information

### DIFF
--- a/lte/gateway/c/oai/tasks/sgw/sgw_context_manager.c
+++ b/lte/gateway/c/oai/tasks/sgw/sgw_context_manager.c
@@ -26,7 +26,6 @@
 
 #include <stdio.h>
 #include <stdlib.h>
-#include <stdint.h>
 #include <stdbool.h>
 #include <inttypes.h>
 
@@ -38,15 +37,6 @@
 #include "common_defs.h"
 #include "log.h"
 #include "sgw_context_manager.h"
-
-//-----------------------------------------------------------------------------
-void sgw_display_s11teid2mme(mme_sgw_tunnel_t* mme_sgw_tunnel)
-//-----------------------------------------------------------------------------
-{
-  OAILOG_DEBUG(
-      LOG_SPGW_APP, "| " TEID_FMT "\t<------------->\t" TEID_FMT "\n",
-      mme_sgw_tunnel->remote_teid, mme_sgw_tunnel->local_teid);
-}
 
 //-----------------------------------------------------------------------------
 void sgw_display_sgw_eps_bearer_context(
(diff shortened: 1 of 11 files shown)


lte/gateway/c/oai/tasks/sgw/sgw_task.c
```diff
@@ -172,6 +177,7 @@ static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
       int32_t rc = sgw_handle_ip_allocation_rsp(
           spgw_state, &received_message_p->ittiMsg.ip_allocation_response,
           imsi64);
+      put_spgw_state();
```
ulaskozat: are we saving much by pushing this inside case statements? maybe adopt an approach similar to s1ap task with a flag set as true and if there is no need to update state set it as false in few of these cases?

ardzoht: this change slipped on this PR, I am planning to do this on a different PR to take care of this similarly to the s1ap task pattern

ardzoht: I will revert this change

lte/gateway/c/oai/tasks/sgw/sgw_context_manager.c
```diff
@@ -354,28 +340,26 @@ int sgw_cm_remove_eps_bearer_entry(
 s_plus_p_gw_eps_bearer_context_information_t* sgw_cm_get_spgw_context(
     teid_t teid) {
   s_plus_p_gw_eps_bearer_context_information_t* spgw_bearer_context_info = NULL;
-  hash_table_ts_t* state_imsi_ht = get_spgw_ue_state();
+  hash_table_ts_t* state_teid_ht = get_spgw_teid_state();
```
themarwhal, line 343: slightly unrelated to this PR but
```
lte/gateway/c/oai/tasks/sgw/sgw_context_manager.c#L326

unused parameter 'sgw_pdn_connection' [-Wunused-parameter]
```
is the lint accurate?

ardzoht: yes, it's legit, I will take care of that on a separate PR, thanks for flagging!

ulaskozat (approved):
LGTM

PR #6427, opened 2021-04-27
[AGW][UPGRADE] Upgrade Script

tmdzk:

[AGW][UPGRADE] Upgrade Script

## Summary

We have multiple agw already deployed in the field, those are mapping to old registries. 
The old registries had their own file in `/etc/apt/sources.list.d/`, however now with the new registry we only need one `artifactory.magmacore.org` that will host all debian/docker/helm

This script upgrade to latest magma and make sure to remove all mapping to the old registries. 

## Test Plan

Lab testing

### DIFF
--- a/lte/gateway/release/upgrade_magma.sh
+++ b/lte/gateway/release/upgrade_magma.sh
@@ -1,0 +1,34 @@
+WHOAMI=$(whoami)
+MAGMA_VERSION="1.5.0"
+# Default is focal
+OS_VERSION="focal"
+
+echo "Checking if the script has been executed by root user"
+if [ "$WHOAMI" != "root" ]; then
+  echo "You're executing the script as $WHOAMI instead of root.. exiting"
+  exit 1
+fi
+
+while true; do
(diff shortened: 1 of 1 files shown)


themarwhal (approved):
do we need the shebang as the lint suggests?

lte/gateway/release/upgrade_magma.sh
```diff
@@ -0,0 +1,34 @@
...
+  OS_VERSION="stretch"
+fi
+
+# We have changed the name too many time we have to wipe all versions
+rm -rf /etc/apt/sources.list.d/*
+
+wget https://artifactory.magmacore.org:443/artifactory/api/gpg/key/public -O /tmp/public
+apt-key add /tmp/public
+
+echo "deb https://artifactory.magmacore.org/artifactory/debian $OS_VERSION-$MAGMA_VERSION main" >> /etc/apt/sources.list.d/magma
```
sudhikan: I think it needs to be `>> /etc/apt/sources.list.d/magma.list` without the suffix I was seeing an error on my machine.

PR #6444, opened 2021-04-28
Fix nil exception in TestGyReAuth

themarwhal:

## Summary
Seeing that the reauth test gets a nil exception when the reauth response is never received. (https://app.circleci.com/pipelines/github/magma/magma/20924/workflows/b812ebf9-baac-4e40-9066-1ff403b7f787/jobs/217185/tests#failed-test-0)
This leads to an ugly test output, so adding a check to only assert on the value if raa is not nil. (The test will still fail if it is nil)

Additionally, the re-run logic doesn't get applied when there is an exception as well.
```
ERROR rerun aborted because previous run had a suspected panic and some test may not have run
```

I also fixed the golanglint issues seen in the integration tests directory
```
root@docker-desktop:/magma/cwf/gateway/integ_tests# golangci-lint run
rule_manager.go:162: File is not `gofmt`-ed with `-s` (gofmt)
	for imsi, _ := range rulesIDsByIMSI {
test_runner.go:252: File is not `gofmt`-ed with `-s` (gofmt)
	for imsi, _ := range tr.imsis {
assertions.go:36:34: S1039: unnecessary use of fmt.Sprintf (gosimple)
	assert.NotNil(tr.t, eapMessage, fmt.Sprintf("EAP Message from authentication is nil"))
	                                ^
assertions.go:37:76: S1039: unnecessary use of fmt.Sprintf (gosimple)
	assert.True(tr.t, reflect.DeepEqual(int(eapMessage[0]), eap.SuccessCode), fmt.Sprintf("UE Authentication did not return success"))
	                                                                          ^
assertions.go:47:34: S1039: unnecessary use of fmt.Sprintf (gosimple)
	assert.NotNil(tr.t, eapMessage, fmt.Sprintf("EAP Message from authentication is nil"))
	                                ^
assertions.go:74:28: S1002: should omit comparison to bool constant, can be simplified to `!reflect.DeepEqual(int(eapMessage[0]), eap.SuccessCode)` (gosimple)
			if eapMessage == nil || reflect.DeepEqual(int(eapMessage[0]), eap.SuccessCode) == false {
			                        ^
rule_manager.go:148:3: S1011: should replace loop with `rules = append(rules, accountRules.StaticRuleNames...)` (gosimple)
		for _, ruleID := range accountRules.StaticRuleNames {
		^
service_wrappers.go:543:14: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
			Addr:     fmt.Sprintf(address),
			          ^
root@docker-desktop:/magma/cwf/gateway/integ_tests#
```

## Test Plan
CWF integration test 

## Additional Information

### DIFF
--- a/cwf/gateway/go.sum
+++ b/cwf/gateway/go.sum
@@ -379,12 +379,14 @@ github.com/mailru/easyjson v0.0.0-20190626092158-b2ccc519800e/go.mod h1:C1wdFJiN
 github.com/mattn/go-colorable v0.0.9/go.mod h1:9vuHe8Xs5qXnSaW/c/ABM9alt+Vo+STaOChaDxuIBZU=
 github.com/mattn/go-colorable v0.1.2 h1:/bC9yWikZXAL9uJdulbSfyVNIR3n3trXl+v8+1sx8mU=
 github.com/mattn/go-colorable v0.1.2/go.mod h1:U0ppj6V5qS13XJ6of8GYAs25YV2eR4EVcfRqFIhoBtE=
+github.com/mattn/go-colorable v0.1.8 h1:c1ghPdyEDarC70ftn0y+A/Ee++9zz8ljHG1b13eJ0s8=
 github.com/mattn/go-colorable v0.1.8/go.mod h1:u6P/XSegPjTcexA+o6vUJrdnUu04hMope9wVRipJSqc=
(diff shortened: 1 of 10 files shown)


emakeev (approved):
LGTM

uri200:
If you can, run `golangci-lint run` one more time to see if all lint issues are gone. For me while doing feg golangci-lint was not providing all the issues on one go

uri200 (approved):

PR #6361, opened 2021-04-23
Add log forwarding to fluentd configuration to Federation gateway deployment

119Vik:
Signed-off-by: Vitalii Kostenko <vitalii@freedomfi.com>

## Summary

At the moment in the scope of FEG deployment, we have a fluent-bit container which only processing eventD messages however it's preconfigured to forward system logs and docker logs. This PR updates several files to ensure that syslog and docker logs may be forwarded to fluentD by adding the correct log_driver type.

## Test Plan

deploy FEG and check that docker logs and system logs are going to ELK

## Additional Information

- in the .env file we should use LOG_DRIVER=fluentd
- docker-compose 1.29.0 + required

PR #6457, opened 2021-04-28
bump docker compose to 1.29.1 cause it supports better container dependencies definitions

119Vik:
Signed-off-by: Vitalii Kostenko <vitalii@freedomfi.com>

## Summary

Starting from v1.29.0  docker-compose now supports better  depends_on definition. It allows us to use initContainer like definitions until we migrate all the workloads to k8s. FEG and CWAG are good examples of such workloads.

Release notes https://docs.docker.com/compose/release-notes/.
Examples of using such extended depends_on instruction: https://github.com/magma/magma/pull/6361.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,7 +70,7 @@ orbs:
           compose_version:
             description: Docker-compose version to install
             type: string
-            default: "1.25.4"
+            default: "1.29.1"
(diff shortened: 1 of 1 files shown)


tmdzk (approved):
LGTM 

CI is green so let's go

PR #6462, opened 2021-04-28
[lte][agw] Reduce SPGW state writes if state has not changed

### LINKED ISSUE #5693: MME Redis - Write Optimization
Improve serde speeds.
Reduce the frequency of writes to a bare minimum value.

ardzoht:
## Summary

- This PR removes the need to write `SPGW` state to redis if the state content has not been changed after processing the message, SPGW UE state will still be written since pretty much every processed message updates SPGW UE context.
- Some clang-format changes for `Attach.c` and `mme_app_bearer.c`

## Test Plan

- make integ_test as sanity
- Tested with `test_continuous_random_attach.py` s1ap test, and compared the number of times the spgw state is updated:
```
Continuous random attach test (200 UEs) (5 runs)

*master*         *This revision*
313 writes        198 writes
287                   190
337			195
352			195
338			195
```

## Additional Information


ulaskozat (approved):
LGTM

PR #6515, opened 2021-04-30
feat(orc8r): add terraform support of nprobe service

### LINKED ISSUE #4726: [X2] Support nprobe service in terraform

ymasmoudi:

## Summary

This PR is the last piece in X2 interface implementation. It adds the following changes,

1/ Renamed service config items to match other services
2/ Fix timestamp encoding for both asn1 content and X2 header following partner testing
3/ Make LIID mandatory for X2 asn1 content
4/ Updated unit tests

NProbe default service config is overridden by /var/opt/magma/configs/orc8r/nprobe.yml 

NOTE: nprobe.pem/key must be generated for existing orc8r deployments and loaded in secrets manager.
You can follow these steps,
```
${MAGMA_ROOT}/orc8r/cloud/deploy/scripts/create_application_certs.sh yourdomain.com 
terraform taint module.orc8r-app.null_resource.orc8r_seed_secrets
terraform apply -target=module.orc8r-app.null_resource.orc8r_seed_secrets
```

## Test Plan

Remote server collector is available here,
```
https://github.com/ymasmoudi/nprobe-collector
```

Generated certs for nprobe service and collector using
```
~ ./make_certs.sh -d
```

Provision a task through swagger for a speicific IMSI then generate mme/sessiond event for the subscriber.

Examples of output records are available
https://github.com/ymasmoudi/nprobe-collector/tree/master/examples

Unit Test
```
ymasmoudi-mbp:nprobe ymasmoudi$ go test ./...
?   	magma/lte/cloud/go/services/nprobe	[no test files]
ok  	magma/lte/cloud/go/services/nprobe/encoding	0.813s
?   	magma/lte/cloud/go/services/nprobe/exporter	[no test files]
?   	magma/lte/cloud/go/services/nprobe/nprobe	[no test files]
?   	magma/lte/cloud/go/services/nprobe/nprobe_manager	[no test files]
ok  	magma/lte/cloud/go/services/nprobe/obsidian/handlers	(cached)
?   	magma/lte/cloud/go/services/nprobe/obsidian/models	[no test files]
ok  	magma/lte/cloud/go/services/nprobe/storage	(cached)
```


themarwhal:
Hey @ymasmoudi! There has been recent changes in CODEOWNERS file so please rebase onto master before you request for reviews! Thanks :)

ekowtaylor:
@magma/approvers-orc8r this PR is pending a review.

emakeev:
It may need to be rebased/fixed - there is a conflict

ymasmoudi:
> It may need to be rebased/fixed - there is a conflict

rebased and fixed the conflict.

ekowtaylor:
@hcgatewood , @koolzz this PR is ready for review

electronjoe:
I've asked @hcgatewood and @mstre123 to coordinate feedback on this PR.

orc8r/cloud/helm/orc8r/charts/secrets/values.yaml
```diff
@@ -55,6 +63,10 @@ secret:
         elasticPort: 9200
       analytics.yml: |-
         analysisSchedule: 3
+      nprobe.yml: |-
+        operatorID: 1
+        updateIntervalSecs: 60
+        backoffIntervalSecs: 360
```
hcgatewood: Use the `nprobe.yml` file to set these default values. Only include values here that *must* be set (i.e. are ~unique per deployment / a default doesn't make sense)

That said, if you're providing defaults for all these values in the tf, then don't include the configs here at all

orc8r/cloud/helm/orc8r/charts/secrets/Chart.yaml
```diff
@@ -13,7 +13,7 @@ apiVersion: v1
 appVersion: "1.0"
 description: A Helm chart to create magma orchestrator secrets
 name: secrets
-version: 0.1.10
```
hcgatewood: Throughout: you'll likely have to update the versions based on latest master. Do that last, once we get through the rest of the review process

orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
```diff
@@ -445,4 +445,26 @@ variable "cloudwatch_exporter_enabled" {
...
+variable "nprobe_delivery_server" {
+  description = "Remote LIMS address where nprobe service exports records."
+  type        = string
+  default     = "10.10.0.2:6666"
+}
+
+variable "nprobe_skip_verify_server" {
+  description = "allows nprobe service to skip tls server verification when exporting records."
+  type        = bool
+  default     = true
```
hcgatewood: Why default true? skip_verify=true should only be used for local calls or tests

orc8r/cloud/deploy/scripts/create_application_certs.sh
```diff
@@ -53,6 +53,13 @@ echo "######################"
 openssl genrsa -out fluentd.key 2048
 openssl req -x509 -new -nodes -key fluentd.key -sha256 -days 3650 -out fluentd.pem -subj "/C=US/CN=fluentd.$domain"
 
+echo ""
+echo "######################"
+echo "Creating nprobe certs"
+echo "######################"
+openssl genrsa -out nprobe.key 2048
+openssl req -x509 -new -nodes -key nprobe.key -sha256 -days 3650 -out nprobe.pem -subj "/C=US/CN=nprobe.$domain"
```
hcgatewood: Are the nprobe certs provided externally, or are they always generated? If they're always generated, we don't need to use a new keypair, we can just reuse the existing `controller.{key,crt}` keypair

lte/cloud/helm/lte-orc8r/values.yaml
```diff
@@ -137,3 +137,13 @@ smsd:
...
+
+nprobe:
+  service:
+    labels:
+      orc8r.io/obsidian_handlers: "true"
+      orc8r.io/swagger_spec: "true"
+    annotations:
+      orc8r.io/obsidian_handlers_path_prefixes: >
+        /magma/v1/lte/:network_id/network_probe/tasks,
+        /magma/v1/lte/:network_id/network_probe/destinations,
```
hcgatewood: This routing is based on prefixes, so just one prefix is sufficient
```suggestion
      orc8r.io/obsidian_handlers_path_prefixes: >
        /magma/v1/lte/:network_id/network_probe,
```

lte/cloud/go/services/nprobe/encoding/utils.go
```diff
@@ -35,14 +38,102 @@ func convertUint32ToBytes(v uint32) []byte {
...
+		firstDigit := int(imei[idx])
+		secondDigit := int(imei[idx+1])
+		i := 3 - idx/2
+		if idx >= 8 {
+			i = 7 - idx/2 + 3
+		}
+		encodedIMEI[i] = byte((firstDigit << 4) + secondDigit)
+	}
+
+	// encode spare
```
hcgatewood: Nit: sentence-case your comments

lte/cloud/go/services/nprobe/encoding/record_test.go
```diff
@@ -29,47 +30,41 @@ type timeTest struct {
 
 var (
 	generalizedTimeTestData = []timeTest{
-		{"2021-02-18T05:13:26.019519+00:00", true, "010000000ed7bfee760129d6180000"},
+		{"2021-02-18T05:13:26.019519+00:00", true, "32303231303231383035313332362e303139"},
 	}
 
 	// This is a minimal bearer activation record that should be parsed correctly.
 	encodedRecord = []byte{
```
hcgatewood: Here and elsewhere: for big values like this, instead of storing this value inline, move it to a a base64-encoded [golden file](https://softwareengineering.stackexchange.com/questions/358786/what-is-golden-files)

lte/cloud/go/services/nprobe/config.go
```diff
@@ -31,15 +32,15 @@ const (
...
-	ExporterCrtFile      string `yaml:"exporter_crt"`
+	UpdateIntervalSecs  uint32 `yaml:"updateIntervalSecs"`
+	BackOffIntervalSecs uint32 `yaml:"backoffIntervalSecs"`
+	MaxExportRetries    uint32 `yaml:"maxExportRetries"`
+	OperatorID          uint32 `yaml:"operatorID"`
+
+	DeliveryServer   string `yaml:"deliveryServer"`
+	ExporterKey      string `yaml:"exporterKey"`
+	ExporterCrt      string `yaml:"exporterCrt"`
+	SkipVerifyServer bool   `yaml:"skipVerifyServer"`
```
hcgatewood: Replicate the nprobe.yml comments to this Config struct as well

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -588,8 +588,13 @@ func getNetworkProbeConfig(networkID string) ([]*lte_mconfig.NProbeTask, *lte_mc
 
 	for _, ent := range ents {
 		task := (&nprobe_models.NetworkProbeTask{}).FromBackendModels(ent)
-		npTasks = append(npTasks, nprobe_models.ToMConfigNProbeTask(task))
+		if task.TaskDetails.DeliveryType ==
+			nprobe_models.NetworkProbeTaskDetailsDeliveryTypeEventsOnly {
```
hcgatewood:
```suggestion
		if task.TaskDetails.DeliveryType == nprobe_models.NetworkProbeTaskDetailsDeliveryTypeEventsOnly {
```

lte/cloud/configs/nprobe.yml
```diff
@@ -10,20 +10,21 @@
...
-# exporter_crt provides the absolute path to exporter tls certificate.
-# skip_verify_server enables exporter to skip server tls certificate verifications.
+# operatorID represents the mobile operator identifier
+# updateIntervalSecs sets the priodic time between runs in seconds
+# backoffIntervalSecs sets the backoff time when remote records collector is not available
+# maxExportRetries sets the number of retries when exporting a record
+# deliveryServer defines the address of the remote server collecting records
+# exporterKey provides the absolute path to exporter tls private key
+# exporterCrt provides the absolute path to exporter tls certificate
+# verifyServer enables exporter to skip server tls certificate verifications
```
hcgatewood: Move all these docstring comments directly above their key-value pair -- otherwise we have to hunt down-and-back to see what each comment refers to

lte/cloud/configs/nprobe.yml
```diff
@@ -10,20 +10,21 @@
...
-# operator_id represents the mobile operator identifier
-# update_interval_secs sets the priodic time between runs in seconds.
-# backoff_interval_secs sets the backoff time when remote records collector is not
-# available.
-# delivery_function_address defines the address of the remote server collecting records.
-# exporter_key provides the absolute path to exporter tls private key.
-# exporter_crt provides the absolute path to exporter tls certificate.
-# skip_verify_server enables exporter to skip server tls certificate verifications.
+# operatorID represents the mobile operator identifier
+# updateIntervalSecs sets the priodic time between runs in seconds
```
hcgatewood:
```suggestion
# updateIntervalSecs sets the periodic time between runs in seconds
```

lte/cloud/configs/nprobe.yml
```diff
@@ -10,20 +10,21 @@
...
+# backoffIntervalSecs sets the backoff time when remote records collector is not available
+# maxExportRetries sets the number of retries when exporting a record
+# deliveryServer defines the address of the remote server collecting records
+# exporterKey provides the absolute path to exporter tls private key
+# exporterCrt provides the absolute path to exporter tls certificate
```
hcgatewood: Indicate that these are for a tls client cert (?)

hcgatewood (changes requested):
Couple requests, mainly around configuration handling, skip_verify=true, and whether we need to create a new keypair for this purpose

P.S. generally speaking, prefer to split application code changes (`*.go`) into separate PRs from infrastructure changes (`*.tf`, `*.yaml`, etc) for readability and reviewability

lte/cloud/go/services/nprobe/encoding/record_test.go
ymasmoudi: I moved both tests to golden files

orc8r/cloud/deploy/scripts/create_application_certs.sh
ymasmoudi: nprobe.{key, pem} are client certificates and will provided by LI vendors. 
During the setup, I generate these to allow the setup to go through.

ymasmoudi:
@hcgatewood I fixed your review comments.

lte/cloud/helm/lte-orc8r/templates/nprobe.service.yml
```diff
@@ -0,0 +1,36 @@
+{{/*
```
hcgatewood: Make all these helm templates `.yaml` extension for consistency

lte/cloud/go/services/nprobe/encoding/asn1_test.go
```diff
@@ -24,41 +23,40 @@ import (
 var (
 	// This is a minimal bearer activation record that should be parsed correctly.
 	encodedPayload = []byte{
```
hcgatewood: Ideally this payload would come from a goldenfile as well

lte/cloud/go/services/nprobe/config.go
```diff
@@ -31,15 +32,23 @@ const (
...
+	MaxExportRetries uint32 `yaml:"maxExportRetries"`
+
+	// deliveryServer defines the address of the remote server collecting records
+	DeliveryServer string `yaml:"deliveryServer"`
+	// exporterKey provides the absolute path to exporter tls client private key
+	ExporterKey string `yaml:"exporterKey"`
+	// exporterCrt provides the absolute path to exporter tls client certificate
+	ExporterCrt string `yaml:"exporterCrt"`
+	// skipVerifyServer enables exporter to skip server tls certificate verifications
+	SkipVerifyServer bool `yaml:"skipVerifyServer"`
```
hcgatewood: Docstring comment should match the casing -- so e.g. `// DeliveryServer defines...`

orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
```diff
@@ -452,3 +452,25 @@ variable "cloudwatch_exporter_enabled" {
...
+  description = "Remote LIMS address where nprobe service exports records."
+  type        = string
+  default     = "10.10.0.2:6666"
+}
+
+variable "nprobe_skip_verify_server" {
+  description = "allows nprobe service to skip tls server verification when exporting records."
+  type        = bool
+  default     = false
+}
```
hcgatewood: In my mind, the nprobe architecture is doing too much via *configuration* (infrastructure-level) rather than *in-application* (rest api, nms)

Advantages of the in-application style
- Network operators can make changes without having to redeploy the entire Orc8r
- Nprobe configuration can be scoped to a particular network, allowing multiple operators (sharing the multi-tenant Orc8r) to independently make use of the nprobe functionality

hcgatewood (changes requested):
Couple more nits then I'll be content

Main thing is: can you, here or in DM, describe to me why we're going with such a configuration-heavy approach? I find it incongruent with the general Orc8r principles. Was it an intentional design decision, or due to time constraints?

orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
ymasmoudi: During design sessions, It was agreed to keep destinations provisioning through APIs (which is part of 3gpp std btw) and multi-tenancy support to a second release!

Anyhow, I remove the above config from terraform by moving operator_id to task definition and retrieving destination server from swagger.  
TLS server verification is set back to false to allow the incoming interop testing for now.

lte/cloud/helm/lte-orc8r/values.yaml
```diff
@@ -142,3 +142,12 @@ subscriberdb_cache:
   service:
     labels: {}
     annotations: {}
+
+nprobe:
```
hcgatewood, line 146: Replicate these values to https://github.com/magma/magma/blob/20eb7d7fa5bd136d9af5f27df3455e22e92d9fd0/lte/cloud/configs/service_registry.yml so they'll work in local dev (Docker) setup

orc8r/cloud/deploy/scripts/create_application_certs.sh
hcgatewood: So, if the certs are going to be provided by operators, then that means, eventually, we'll want multi-tenant support

Which leads me to the idea of uploading these certs per-network via Swagger. Which would then require that we gracefully handle the certs not existing

Refocusing on this PR then: I think it would be cleaner to just not generate these certs (since, if I'm reading correctly, they're "throwaway" certs that will never be used ?). The nprobe codepath should handle the certs not existing

Let me know if I'm misunderstanding. My reason for pushing back on this is we're actively trying to *decrease* the number of certs we have to track, as cert issues is a perennial SEV rc. So actually adding new certs needs an air-tight rationale

hcgatewood (changes requested):

orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
hcgatewood: @ymasmoudi I wasn't part of the design process as you know, I've been asked to fill in on reviewing these PRs. I think we can come to a compromise between our goals (yours being landing this functionality, mine being keeping Orc8r complexity manageable and removing unnecessary externalities) 🙂 

For the config changes
- From skimming the PR, I don't see where the operator_id is being set -- can you point me in the right direction?
- I strongly think leaving TLS server verification to false is a bad idea. If you accidentally forget to switch it back to true, or even if partners move forward with the feature with an "outdated" Orc8r version, they'll be broadcasting LI data (?) to the world. I can't imagine any feature velocity would be worth that negative security externality

Let me know if you want to schedule a sync to synchronously move through my concerns. I think we're almost there with getting this merged 🙂

hcgatewood:
See https://github.com/magma/magma/pull/6515#discussion_r662578045 for the bulk of my comments

orc8r/cloud/deploy/scripts/create_application_certs.sh
ymasmoudi: Moving the config including certs to swagger will not make the solution multi-tenant. As you know, a single admin_operator cert will provide access to all tenants data. Thus, we will not be able to expose X1 in multi-tenant environment anyway.

In addition, having a certificate with private key accessible through swagger without any access control due to current limitation seems to me a bigger security concern!

Anyway, your requested changes were added.

orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
ymasmoudi: Since disabling tls server verification is a big concern, I wonder why it is disabled by default in many places in orc8r and feg. 
In addition, all our deployments uses create_application_certs.sh to generate self-signed certs. This script does not use any SNI extension. Thus forcing all integrations with orc8r to disable tls server verification!

Finally, thanks for taking over this review. I'm just not sure if a PR review is the correct place to change a project scope agreed with your team.

Anyhow, all your requested changes were added.

lte/cloud/go/services/nprobe/obsidian/models/swagger.v1.yml
```diff
@@ -259,6 +259,12 @@ definitions:
         format: date-time
         example: 2020-03-11T00:36:59.65Z
         description: The timestamp in ISO 8601 format
+      operator_id:
```
ymasmoudi, line 262: operator_id is moved here

orc8r/cloud/deploy/scripts/create_application_certs.sh
hcgatewood: Agreed that moving that part of the config to swagger doesn't achieve multi-tenant solution -- rather, pointing to the idea that, since we'll have to move it to swagger eventually (unless I'm misunderstanding), may as well do it now

Agreed that it's not ideal that a private key is accessible through swagger given our lack of fine-grained access controls. However, (a) that will be fixed on-demand (when partners request it we'll prioritize the fix), and (b) the current pattern of Orc8r usage is admin_operator is only ever used by NMS or admins who already have full access to the Orc8r infrastructure -- meaning moving the configs to swagger isn't any less secure under current patterns

Thanks for making the change 🙂

orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
hcgatewood: For disabling tls server verification: I can't speak for feg, but for orc8r if you follow the thread of each instance of "no tls", it's never enabled by default (as best I can tell)

For self-signed certs -- the server's TLS cert's root is rootCA.pem, which we do in fact add to each gateway so that they can securely connect to orc8rs that use self-signed certs

Anyways, I agree that these requested changes would ideally have come earlier in the design process. I wasn't part of that process, and was asked to help at the last-mile, so that's why we're running with this suboptimal pattern

Thanks for the changes, lgtm 🙂

hcgatewood (approved):
Lgtm, thanks for the flexibility in getting this merged 🙂

PR #6557, opened 2021-04-30
Include OVS RYU Patch into Build Script

koolzz:

## Additional Information


pshelar:
How are we going to package it for prod AGW.

koolzz:
> How are we going to package it for prod AGW.

not sure what you mean, this is modifying the build script to include it so it will be in production deployments. Locally its already there in the ansible playbook

pshelar (approved):

uri200 (approved):

tmdzk (approved):

PR #6611, opened 2021-05-01
AGW: deploy: post install script cleanups

pshelar:
Following patch adds script download to install script to
avoid manually downloading the post deploy script.
It also fixes file permissions of all bash scripts.

## Test Plan
Ran the command on local setup.
I will try install script on virtual AGW

## Additional Information


ardzoht (approved):
lgtm

PR #6638, opened 2021-05-03
Add TEID to PolicyBearer binding response

themarwhal:
Signed-off-by: Marie Bremner <marwhal@fb.com>

## Summary
* PipelineD requires per-bearer TEIDs when activating flows. 
* @ulaskozat will add changes to send the TEIDs to SessionD
* @themarwhal will add changes to propagate the per-bearer TEIDs for policies that have dedicated bearers

## Test Plan
no test 

## Additional Information

- [ ] This change is backwards-breaking

PR #6690, opened 2021-05-03
<feat:>[SessionD]: Store teids from PolicyBearerBindingRequest

### LINKED ISSUE #6754: [SessionD] Send down Teids that correspond to the serving bearer to PipelineD

themarwhal:

## Summary
* In order to migrate to TEID based PipelineD interactions, we will need to send down teids that are consistent with the bearer each policy is served by.
* https://github.com/magma/magma/pull/6638 added dedicated bearer teids to `PolicyBearerBindingRequest`
* This PR will store the TEID values into the bearerIDsByPolicyID map that already exists

## Test Plan
unit tests
should not affect existing flows

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/SessionState.cpp
+++ b/lte/gateway/c/session_manager/SessionState.cpp
@@ -1985,7 +1985,10 @@ void SessionState::bind_policy_to_bearer(
   }
   MLOG(MINFO) << session_id_ << " now has policy " << rule_id
               << " tied to bearerID " << request.bearer_id();
-  bearer_id_by_policy_[PolicyID(*policy_type, rule_id)] = request.bearer_id();
+  BearerIDAndTeid brearer_id_and_teid;
+  brearer_id_and_teid.bearer_id                         = request.bearer_id();
+  brearer_id_and_teid.teids                             = request.teids();
+  bearer_id_by_policy_[PolicyID(*policy_type, rule_id)] = brearer_id_and_teid;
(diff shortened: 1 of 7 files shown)


uri200:
this is because the different dedicated bearers were identified by their own IP and now we need to identify them with TEIDs, right?

koolzz:
> this is because the different dedicated bearers were identified by their own IP and now we need to identify them with TEIDs, right?

correct

uri200:
But after checking the code, I don't see any reference to the IP before. So how was this working before? Did we store the specific bearer IP somewhere else I may have missed?

koolzz:
> 

Actually yeah correction, we were sending the IP of the main bearer. This solution should be cleaner as we'll know exactly which bearer is reporting

uri200:
Ok so you were guessing using the IP and the specific bearer what flow sessiond was referring to.

Now, since pipelined/ovs uses TEID we have a unique way to identify the flow. So you just need the tunnel id. 

Btw, what tunnel id will you be using, UL or DL or both?

koolzz:
> Ok so you were guessing using the IP and the specific bearer what flow sessiond was referring to.
> 
> Now, since pipelined/ovs uses TEID we have a unique way to identify the flow. So you just need the tunnel id.
> 
> Btw, what tunnel id will you be using, UL or DL or both?

Yeah.
UL will be used as a flow key, but DL is required for other stuff( like ipv6)

uri200 (approved):

PR #6692, opened 2021-05-04
refactor: Clean up process_rules_to_install to prepare for separating out default and dedicated bearer policies

### LINKED ISSUE #6754: [SessionD] Send down Teids that correspond to the serving bearer to PipelineD

themarwhal:

## Summary
In order to fully migrate to Teid based flow/policy identification as opposed to IP, we will need to separate out policies that are served by the default bearer from ones that are served by dedicated bearers. (Each bearer has a unique set of teids)

For policies that are served by dedicated bearers, we will need to delay the flow activation until we receive the matching teids from MME. 

This PR is just a cosmetic change to prepare for the actual change described above. 
More specifically, I cleaned up `process_rules_to_install` in `LocalEnforcer` by splitting the function into two functions (`process_static_rule_installs` and `process_dynamic_rule_installs` in `SessionState`).

## Test Plan
existing unit tests
CI integration tests

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -835,11 +835,11 @@ void LocalEnforcer::filter_rule_installs(
 }
 
 void LocalEnforcer::handle_session_activate_rule_updates(
-    const std::string& imsi, SessionState& session_state,
+    const std::string& imsi, SessionState& session,
     const CreateSessionResponse& response,
     std::unordered_set<uint32_t>& charging_credits_received) {
-  RulesToProcess rules_to_activate;
-  RulesToProcess rules_to_deactivate;
+  RulesToProcess to_activate;
+  RulesToProcess to_deactivate;
   std::vector<StaticRuleInstall> static_rule_installs =
       to_vec(response.static_rules());
   std::vector<DynamicRuleInstall> dynamic_rule_installs =
@@ -849,24 +849,24 @@ void LocalEnforcer::handle_session_activate_rule_updates(
 
   SessionStateUpdateCriteria uc;  // TODO remove unused UC
   process_rules_to_install(
-      session_state, imsi, static_rule_installs, dynamic_rule_installs,
-      rules_to_activate, rules_to_deactivate, uc);
+      session, imsi, static_rule_installs, dynamic_rule_installs, &to_activate,
+      &to_deactivate, &uc);
 
   // activate_flows_for_rules() should be called even if there is no rule
   // to activate, because pipelined activates a "drop all packet" rule
   // when no rule is provided as the parameter.
-  const auto& config = session_state.get_config();
+  const SessionConfig& config = session.get_config();
   propagate_rule_updates_to_pipelined(
-      imsi, config, rules_to_activate, rules_to_deactivate, true);
+      imsi, config, to_activate, to_deactivate, true);
 
   if (config.common_context.rat_type() == TGPP_LTE) {
-    auto bearer_updates = session_state.get_dedicated_bearer_updates(
-        rules_to_activate, rules_to_deactivate, uc);
+    BearerUpdate bearer_updates =
+        session.get_dedicated_bearer_updates(to_activate, to_deactivate, uc);
     if (bearer_updates.needs_creation) {
       // If a bearer creation is needed, we need to delay this by a few seconds
       // so that the attach fully completes before.
       schedule_session_init_dedicated_bearer_creations(
-          imsi, session_state.get_session_id(), bearer_updates);
+          imsi, session.get_session_id(), bearer_updates);
     }
   }
 }
@@ -1367,20 +1367,18 @@ void LocalEnforcer::update_monitoring_credits_and_rules(
     session->receive_monitor(usage_monitor_resp, uc);
     session->set_tgpp_context(usage_monitor_resp.tgpp_ctx(), uc);
 
-    RulesToProcess rules_to_activate;
-    RulesToProcess rules_to_deactivate;
+    RulesToProcess to_activate, to_deactivate;
 
     process_rules_to_remove(
-        imsi, session, usage_monitor_resp.rules_to_remove(),
-        rules_to_deactivate, uc);
+        imsi, session, usage_monitor_resp.rules_to_remove(), to_deactivate, uc);
 
     process_rules_to_install(
         *session, imsi, to_vec(usage_monitor_resp.static_rules_to_install()),
-        to_vec(usage_monitor_resp.dynamic_rules_to_install()),
-        rules_to_activate, rules_to_deactivate, uc);
+        to_vec(usage_monitor_resp.dynamic_rules_to_install()), &to_activate,
+        &to_deactivate, &uc);
 
     propagate_rule_updates_to_pipelined(
-        imsi, config, rules_to_activate, rules_to_deactivate, false);
+        imsi, config, to_activate, to_deactivate, false);
 
     if (terminate_on_wallet_exhaust() && is_wallet_exhausted(*session)) {
       actions.sessions_to_terminate.insert(ImsiAndSessionID(imsi, session_id));
@@ -1402,8 +1400,8 @@ void LocalEnforcer::update_monitoring_credits_and_rules(
     }
 
     if (config.common_context.rat_type() == TGPP_LTE) {
-      const auto update = session->get_dedicated_bearer_updates(
-          rules_to_activate, rules_to_deactivate, uc);
+      const BearerUpdate update =
+          session->get_dedicated_bearer_updates(to_activate, to_deactivate, uc);
       propagate_bearer_updates_to_mme(update);
     }
   }
@@ -1505,8 +1503,8 @@ void LocalEnforcer::handle_set_session_rules(
     RuleSetBySubscriber rule_set_by_sub(rules_per_sub);
 
     for (const auto& session : session_it->second) {
-      RulesToProcess rules_to_activate;
-      RulesToProcess rules_to_deactivate;
+      RulesToProcess to_activate;
+      RulesToProcess to_deactivate;
       const auto& config = session->get_config();
 
       const auto& apn = config.common_context.apn();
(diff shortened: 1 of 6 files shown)


lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -1062,6 +1062,92 @@ bool SessionState::deactivate_scheduled_static_rule(
   return true;
 }
 
+void SessionState::process_rule_installs(
+    std::vector<StaticRuleInstall> static_rule_installs,
+    std::vector<DynamicRuleInstall> dynamic_rule_installs,
+    RulesToProcess& to_activate, RulesToProcess& to_deactivate,
+    RulesToSchedule& to_schedule, SessionStateUpdateCriteria& session_uc) {
+  std::time_t current_time = std::time(nullptr);
+  std::string ip_addr      = config_.common_context.ue_ipv4();
```
amarpad: where is this used?

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1748,72 +1748,30 @@ void LocalEnforcer::process_rules_to_install(
...
-    }
+  RulesToSchedule to_schedules;
+  session.process_rule_installs(
+      static_rule_installs, dynamic_rule_installs, rules_to_activate,
+      rules_to_deactivate, to_schedules, uc);
 
-    RuleLifetime lifetime(rule_install);
-    if (lifetime.activation_time > current_time) {
-      session.schedule_static_rule(id, lifetime, uc);
+  const std::string session_id = session.get_session_id();
```
amarpad: stylistic: I think you can catch this in const std::string &? the tmp return value will be bound to the lifetime of session_id. Though I believe modern compilers do this for you automatically, not sure.

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1748,72 +1748,30 @@ void LocalEnforcer::process_rules_to_install(
...
     }
-  }
-
-  for (auto& rule_install : dynamic_rule_installs) {
-    PolicyRule dynamic_rule = rule_install.policy_rule();
-    auto rule_id            = dynamic_rule.id();
-    RuleLifetime lifetime(rule_install);
-    if (lifetime.activation_time > current_time) {
-      session.schedule_dynamic_rule(dynamic_rule, lifetime, uc);
+    if (p_type == STATIC && p_action == ACTIVATE) {
```
amarpad: Is this meant to be DYNAMIC?

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -1062,6 +1062,92 @@ bool SessionState::deactivate_scheduled_static_rule(
   return true;
 }
 
+void SessionState::process_rule_installs(
+    std::vector<StaticRuleInstall> static_rule_installs,
```
amarpad: We probably shouldn't copy a vector by value no?

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -1062,6 +1062,92 @@ bool SessionState::deactivate_scheduled_static_rule(
...
+    RulesToSchedule& to_schedule, SessionStateUpdateCriteria& session_uc) {
+  std::time_t current_time = std::time(nullptr);
+  std::string ip_addr      = config_.common_context.ue_ipv4();
+  std::string ipv6_addr    = config_.common_context.ue_ipv6();
+  for (const auto& rule_install : static_rule_installs) {
+    const std::string& rule_id = rule_install.rule_id();
+    RuleLifetime lifetime(rule_install);
+    if (is_static_rule_installed(rule_id)) {
+      // Session proxy may ask for duplicate rule installs.
+      // Ignore them here.
```
amarpad, line 1098: Do we want to log debug or no?

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -1062,6 +1062,92 @@ bool SessionState::deactivate_scheduled_static_rule(
...
+          RuleToSchedule(STATIC, rule_id, ACTIVATE, lifetime.activation_time));
+    }
+    // Schedule deactivation time in the future
+    if (lifetime.should_schedule_deactivation(current_time)) {
+      to_schedule.push_back(RuleToSchedule(
+          STATIC, rule_id, DEACTIVATE, lifetime.deactivation_time));
+    }
+  }
+
+  for (auto& rule_install : dynamic_rule_installs) {
```
amarpad: why can't this be a different function from the above one, none of this loop applies to the previous loop?

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -1062,6 +1062,92 @@ bool SessionState::deactivate_scheduled_static_rule(
...
+    if (lifetime.exceeded_lifetime(current_time)) {
+      optional<uint32_t> op_version =
+          remove_dynamic_rule(rule_id, nullptr, session_uc);
+      if (op_version) {
+        to_deactivate.append_versioned_policy(dynamic_rule, *op_version);
+      }
+      continue;
+    }
+    // If the rule should be active now, install
+    if (lifetime.is_within_lifetime(current_time)) {
```
amarpad, line 1162: You could probably fold both the loops into a single function too potentially

lte/gateway/c/session_manager/SessionState.h
```diff
@@ -572,6 +572,26 @@ class SessionState {
...
+   * PCF/PCRF/PolicyDB
+   * @param dynamic_rule_installs DynamicRuleInstall received from
+   * PCF/PCRF/PolicyDB
+   * @param rules_to_activate output parameter
+   * @param rules_to_deactivate output parameter
+   * @param rules_to_schedule output parameter
+   * @param session_uc
+   */
+  void process_rule_installs(
+      std::vector<StaticRuleInstall> static_rule_installs,
```
amarpad: const REF&?

amarpad (changes requested):
Some drive by comments

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: Yeah I think better practice to use const ref

lte/gateway/c/session_manager/SessionState.cpp
themarwhal: probably not used anywhere. I was moving this function from LocalEnforcer into SessionState so still cleaning up

themarwhal:
@amarpad this is still a draft PR so it's not really complete yet :)

lte/gateway/c/session_manager/SessionState.cpp
amarpad: sorry if I jumped the review gun

lte/gateway/c/session_manager/SessionState.cpp
themarwhal: yeah I think it would be cleaner to just separate them out into `process_static_rule_installs` and `process_dynamic_rule_installs`

themarwhal:
lte-test is broken to an unrelated issue :/

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -1083,6 +1085,98 @@ bool SessionState::deactivate_scheduled_static_rule(
...
+    PolicyRule static_rule;
+    if (!static_rules_.get_rule(rule_id, &static_rule)) {
+      MLOG(MERROR) << "static rule " << rule_id
+                   << " is not found, skipping install...";
+      continue;
+    }
+
+    // If the rule should be deactivated already, deactivate just in case and
+    // continue
+    if (lifetime.exceeded_lifetime(current_time)) {
```
amarpad, line 1113: This seems exactly identical between static and dynamic, you could fold the code into a single function if you feel up to it.

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1505,8 +1503,8 @@ void LocalEnforcer::handle_set_session_rules(
     RuleSetBySubscriber rule_set_by_sub(rules_per_sub);
 
     for (const auto& session : session_it->second) {
-      RulesToProcess rules_to_activate;
-      RulesToProcess rules_to_deactivate;
+      RulesToProcess to_activate;
```
amarpad, line 1506: Do we want to do variable instantiation within a loop?

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -849,24 +849,24 @@ void LocalEnforcer::handle_session_activate_rule_updates(
...
-      imsi, config, rules_to_activate, rules_to_deactivate, true);
+      imsi, config, to_activate, to_deactivate, true);
 
   if (config.common_context.rat_type() == TGPP_LTE) {
-    auto bearer_updates = session_state.get_dedicated_bearer_updates(
-        rules_to_activate, rules_to_deactivate, uc);
+    BearerUpdate bearer_updates =
+        session.get_dedicated_bearer_updates(to_activate, to_deactivate, uc);
     if (bearer_updates.needs_creation) {
       // If a bearer creation is needed, we need to delay this by a few seconds
```
amarpad, line 866: Can you educate me on this comment a bit? This seems like a complicated state dependency

themarwhal:
bad rebase/merge patching into a separate branch https://github.com/magma/magma/pull/6814

amarpad (commented):
Sorry forgot to publish the review, feel free to disregard, just nits

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: For how the function is structured now, yes. It processes it and makes a request to PipelineD/MME in lines 1524-1529. (Might be good to pull out the loop body into a function though)

lte/gateway/c/session_manager/SessionState.cpp
themarwhal: Yeah I am tempted to do this, I think if we just pull in the static/dynamic difference into SessionState the two cases collapse nicely. (But that is a slightly big change that touches a lot of functions.) So maybe in a separate PR :D

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: We added this logic because until recently we installed all policies into PipelineD/create dedicated bearers when the Session was created. (This is before `LocalCreateSessionResponse` is returned to MME) So in order to get around having to wait until MME is ready to accept dedicated bearer requests, we had to wait a few seconds.

Although, I think now that we don't do any rule processing until we receive session activation request from MME (after the create session exchange completes), it might be safe to send the dedicated bearer request directly there. 

(CC @ulaskozat)

PR #6694, opened 2021-05-04
[feg] s8_proxy create bearer protos

uri200:

## Summary

This PR adds the endpoints on s8_proxy protos for create bearer request.

We will keep the responder service at s8_proxy.proto file because it uses the same messages as CreateSession. 
Moving the responder definition to the lte/protos folder forces us to duplicate the structures which can end up diverging in the future (and we don't want that)

## Test Plan

make precommit both ortr8 and feg

## Additional Information


hcgatewood (commented):
Can you mark this as a draft PR while it's WIP 🙂

uri200:
@hcgatewood @themarwhal what is the logic here?  https://github.com/magma/magma/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+wip

themarwhal:
> @hcgatewood @themarwhal what is the logic here? [magma/magma/pulls (is:open sort:updated-desc wip)](https://github.com/magma/magma/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+wip)

I prefer using draft PRs instead of marking it with "WIP". Maybe this is something to write down into a doc somewhere. (Maybe it already is... :p )

uri200:
I agree. But it would be good we all follow the same guideline. 

Maytbe it is a good question for a poll ;)

lte/protos/s8_service.proto.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
```
emakeev: lte/protos/s8_service.proto.proto => lte/protos/s8_service.proto

feg/protos/s8_proxy.proto
```diff
@@ -168,9 +170,12 @@ message DeleteSessionResponsePgw{
     GtpError gtp_error = 1;
 }
 
+
```
emakeev: between LN 156 & 157 above - a new CreateSessionResponsePgw field: string teid_key = 9;

feg/protos/s8_proxy.proto
```diff
@@ -13,6 +13,8 @@ limitations under the License.
 
 syntax = "proto3";
 
+import "lte/protos/oai/spgw_state.proto";
```
emakeev, line 16: such inter module dependencies may be tricky to manage/maintain. Is there a way to  redefine TrafficFlowTemplate here or maybe put them in orc8r/protos?

uri200: We can redefine just copying all the objects and nested object in this file.

But my concern is that then MME will have to get those feg.TFT objects and parse them out into the same lte.TFT objects. And this will work now, but the day we add a field in one and forget in the other one this will break.


Not sure if you will have an answer to this question, but after having tried different combinations my question is: why protos needed be defined per gateway instead of defining all of them in a separate independent module? Protos are interfaces between gateways so they don't really belong to a gateway. Or in other words, why protos were designed to be separated if they can share so many objects?

emakeev: well, some modules may have nothing to do with other module protos, but it obviously not the case with LTE, FEG & CWF

feg/protos/s8_proxy.proto
```diff
@@ -168,6 +175,28 @@ message DeleteSessionResponsePgw{
     GtpError gtp_error = 1;
 }
 
+message CreateBearerRequestPgw {
+    uint32 linked_bearer_id = 1;
+    ProtocolConfigurationOptions protocol_configuration_options = 2;
+    BearerContext bearer_context = 3;
+    // PGW-FQ-CSID-1 ???
+    bytes indication_flag = 5;
```
emakeev: 3 & 5 - is it intentional?

feg/protos/s8_proxy.proto
```diff
@@ -168,6 +175,28 @@ message DeleteSessionResponsePgw{
...
+    string pgwAddrs = 1;        // Ip:port of pgw to send the request. If empty
+                                // s8_proxy will use value in its config
+    uint32 c_pgw_teid = 2;
+    ServingNetwork serving_network = 3;
+
+    uint32 cause = 6;
+    BearerContext bearer_context = 7;
+    // SGW-FQ-CSID ???
+    ProtocolConfigurationOptions protocol_configuration_options = 9;
+    TimeZone time_zone = 11;
```
emakeev: 7, 9, 11, 15 ?

uri200: fixed

emakeev (approved):

PR #6695, opened 2021-05-04
AGW: Non-NAT: configure SGi as a router

pshelar:
Non-NAT Bridged mode configuration can be complicated in production
environment. Following patch adds support for adding SGi as
a router along with bridged mode for UE traffic.

## Test Plan
`make test`
Ran scale test in lab.

## Additional Information


lte/gateway/python/magma/pipelined/tests/test_uplink_bridge.py
```diff
@@ -505,7 +505,7 @@ def testFlowSnapshotMatch(self):
         self.assertIn(cls.SGi_GW, get_iface_gw_ipv4(cls.UPLINK_BRIDGE),
                       "gw not found")
 
-
+@unittest.skip
```
koolzz, line 508: Why is it off?

koolzz (approved):

lte/gateway/python/magma/pipelined/tests/test_uplink_bridge.py
pshelar: This changes resolve conf on the AGW, this is already skipped. I just moved skip to class to avoid test setup.

PR #6768, opened 2021-05-05
devex: Try fix PR HelloBot

hcgatewood:
## Summary

Seems like all existing solutions are ~broken and this is the preferred way to handle. But needs to be merged because this runs as a "trusted" Action based on the base's code. So, #test_in_prod

## Additional Information


themarwhal (approved):

hcgatewood:
Force-merge to override unrelated flaky CI

PR #6769, opened 2021-05-05
refactor: Clean up process_rules_to_install to prepare for separating out default and dedicated bearer policies

### LINKED ISSUE #6754: [SessionD] Send down Teids that correspond to the serving bearer to PipelineD

themarwhal:
## Summary
Another Teid related refactor.. The following will only apply to LTE sessions.

Each policy in a session is served by either the default bearer or dedicated bearer. 
```
default   bearer  <default teid pair>    -- serves --> policyA, policyB
dedicated bearer1 <dedicated teid pair1> -- serves --> policyC
dedicated bearer2 <dedicated teid pair2> -- serves --> policyD
```
This means that when we activate policyA, we want to send down the default teid pair. Similarly for PolicyD, we want to send down dedicated teid pair2. 
In order to make sense out of this, I'm modifying the PipelineDClient interface slightly so that we send separate requests for different teid values.

## The big changes
`create_activate_req` -> `make_activate_req_by_teid` : This function will now return a map of Teid -> ActivateFlows. It will go through all the rules passed in and bundle them based on the teids. If RulesToProcess is empty, it will simple create an ActivateFlowsRequest with default teid. (This is consistent with the current behavior.)

`create_deactivate_req` -> `make_deactivate_req_by_teid` : This function will now return a map of Teid -> DeactivateFlows. It will go through all the rules passed in and bundle them based on the teids. If RulesToProcess is empty, it will simple create an DectivateFlowsRequest with default teid. (This is consistent with the current behavior.)

`RuleToProcess` now includes a Teid field. This value currently always contains the default teid. (Again, this is consistent with the current behavior.) Once we start receiving dedicated bearer teids from MME, we can fill it out accordingly. 

`deactivate_flows_for_rules_for_termination` now takes in a vector of teids. (At the moment, we only pass in the default teid, so this is the same as the current behavior.) Once we get teids from MME, we can send DeactivateFlowsRequest for all IMSI+teids that are associated with the session.

## Test Plan
CI
sessiond unit tests

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/PipelinedClient.cpp
+++ b/lte/gateway/c/session_manager/PipelinedClient.cpp
@@ -50,10 +50,9 @@ magma::SessionSet create_session_set_req(
   return req;
 }
 
-magma::DeactivateFlowsRequest create_deactivate_req(
+magma::DeactivateFlowsRequest make_deactivate_req(
     const std::string& imsi, const std::string& ip_addr,
     const std::string& ipv6_addr, const magma::Teids teids,
-    const magma::RulesToProcess& to_process,
     const magma::RequestOriginType_OriginType origin_type,
     const bool remove_default_drop_rules) {
   magma::DeactivateFlowsRequest req;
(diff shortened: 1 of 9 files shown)


lte/gateway/c/session_manager/PipelinedClient.cpp
```diff
@@ -64,21 +63,59 @@ magma::DeactivateFlowsRequest create_deactivate_req(
...
+    const magma::RulesToProcess& to_process,
+    const magma::RequestOriginType_OriginType origin_type,
+    const bool remove_default_drop_rules) {
+  magma::DeactivateReqByTeids deactivate_req_by_teids;
+  if (to_process.empty()) {
+    // Send an empty request with the default teid
+    deactivate_req_by_teids[default_teids] = make_deactivate_req(
+        imsi, ip_addr, ipv6_addr, default_teids, origin_type,
+        remove_default_drop_rules);
+    return deactivate_req_by_teids;
```
uri200, line 95: will this option deactivate all default and non default bearers. 
meaming, will pipelined look for all the flows related to that imsi, no matter teid?. Shouldn't you include any dedicated teid?

lte/gateway/c/session_manager/PipelinedClient.cpp
```diff
@@ -91,13 +128,52 @@ magma::ActivateFlowsRequest create_activate_req(
...
+    const optional<magma::AggregatedMaximumBitrate>& ambr,
+    const magma::RulesToProcess& to_process,
+    const magma::RequestOriginType_OriginType origin_type) {
+  magma::ActivateReqByTeids activate_req_by_teids;
+  if (to_process.empty()) {
+    // Send an empty request with the default teid
+    activate_req_by_teids[default_teids] = make_activate_req(
+        imsi, ip_addr, ipv6_addr, default_teids, msisdn, ambr, origin_type);
+    return activate_req_by_teids;
+  }
```
uri200, line 162: if we dont have rules to process, why do we want to activate?

lte/gateway/c/session_manager/PipelinedClient.cpp
```diff
@@ -249,18 +324,18 @@ void AsyncPipelinedClient::set_upf_session(
...
+  for (const Teids& t : teids) {
+    MLOG(MDEBUG) << "Deactivating all rules and default drop flows "
+                 << "for " << imsi << ", ipv4: " << ip_addr
+                 << ", ipv6: " << ipv6_addr << ", agw teid: " << t.agw_teid()
+                 << ", enb teid: " << t.enb_teid() << ", origin_type: "
+                 << request_origin_type_to_str(origin_type);
+    DeactivateFlowsRequest req =
+        make_deactivate_req(imsi, ip_addr, ipv6_addr, t, origin_type, true);
+    deactivate_flows(req);
+  }
```
uri200, line 338: why don't pipelined do the same as we do with the case of empty rules? If teid is empty, then deactivate all for that IMSI

uri200 (approved):

lte/gateway/c/session_manager/PipelinedClient.cpp
themarwhal: no for removing all default and non default bearers, we have to specify the specific teids. (PipelineD doesn't have a concept of default vs dedicated, unfortunately)
I just included the clause to stay consistent with the current behavior. But in reality, for termination we use `deactivate_flows_for_rules_for_termination`, which takes in a vector of teids so we can make a request for all of them.

lte/gateway/c/session_manager/PipelinedClient.cpp
themarwhal: I think we rarely hit this case, but even if there are no policies we need to install the drop all rule. So for session creation/activation, even in the rare case we have no policies, we need to send a request.

lte/gateway/c/session_manager/PipelinedClient.cpp
themarwhal: This doesn't work because there could be multiple sessions for that IMSI. :/ We can't accidentally remove flows for a separate session for the same UE. (CC @koolzz)

lte/gateway/c/session_manager/PipelinedClient.cpp
themarwhal: cc @koolzz for confirmation :D

koolzz: yeah thats correct :) we attach but drop all ue traffic in that case

PR #6778, opened 2021-05-05
[docs] Add Resources folder with PCAPs and Useful Links sections

uri200:

## Summary

Add a new section called `Resources` where to drop useful links, files, or information for Magma users.

Added PCAPs with GX/GY and inbound roaming examples

image

## Test Plan
run docusaurus locally

## Additional Information


emakeev (approved):
This is a great idea, we are always searching for pcaps & protocol examples for unit tests & implementation specifics.

docs/readmes/resources/pcap.md
```diff
@@ -0,0 +1,39 @@
...
+id: pcap
+title: PCAP Collection
+hide_title: true
+---
+
+*Last Updated: 4/20/2021*
+
+# PCAP collection
+
+This is a collection of `.pcap` files that can be open using [Wireshark](https://www.wireshark.org/). Some of those PCAP collect data
```
themarwhal: `Some of those PCAP collect data` seems like an incomplete sentence :D

hcgatewood (changes requested):
Love the additions, thanks @uri200! Two requests

- We're trying to keep the number of top-level sections as small as possible. For the pcaps etc, please put those into a subsection under Technical Reference. For the links, please put them somewhere in the intro/getting started section
- Is the inline HTML necessary in the Markdown files? e.g. the `<br />`

docs/readmes/resources/pcap.md
uri200: good catch! it was hidden on my IDE far far in the right

uri200:
> Love the additions, thanks @uri200! Two requests
> 
> * We're trying to keep the number of top-level sections as small as possible. For the pcaps etc, please put those into a subsection under Technical Reference. For the links, please put them somewhere in the intro/getting started section
> * Is the inline HTML necessary in the Markdown files? e.g. the `<br />`

Hey @hcgatewood, initially I had it in Technical Resources, then I moved it to How Tos. After discussing with @talkhasib we thought it didn't belong to any of the categories displayed. I am fine with any location. Should we move `Technical Reference` and `Usage` under `Resources`? 

For the `<br />` the reason is I wanted to have the filename separated from the description. Adding that as a title didn't satisfied it since it was leaving a big gap. And no without the `<br />`, the same line was used for both link and description. See below the different options. I am open to chose between the current or any of those 2 if you like them better.


image


image

hcgatewood:
So the idea is to separate "Usage" as basic information for operators, and "Technical Reference" as more in-depth information for SIs, VARs, and devs. So I would say this info would fit best under `Technical Reference > General` subsection

For the Markdown -- just do something like this, maybe even as a list

- [Example 1](example.com). Foo bar.
- [Example 2](example.com). Apple banana.

uri200:
> So the idea is to separate "Usage" as basic information for operators, and "Technical Reference" as more in-depth information for SIs, VARs, and devs. So I would say this info would fit best under `Technical Reference > General` subsection
> 
> For the Markdown -- just do something like this, maybe even as a list
> 
> * [Example 1](example.com). Foo bar.
> * [Example 2](example.com). Apple banana.

Techincal Reference/General sounds good to me if @talkhasib  doesn't have any problem with it.

uri200:
Found a better way for the entries
image

talkhasib:
under Reference/General sounds good

uri200:
Done, added it to `Technical Reference` section. See picture on the description of the PR

themarwhal (approved):

docs/readmes/resources/pcap.md
```diff
@@ -0,0 +1,55 @@
+---
+id: pcap
+title: PCAP Collection
+hide_title: true
+---
+
+*Last Updated: 4/20/2021*
```
hcgatewood: For both files: these timestamps aren't necessary / won't get updated when people make changes. Docusaurus has a feature to add these automatically, so look into that if you're interested in providing this affordance

docs/docusaurus/sidebars.json
```diff
@@ -181,6 +182,14 @@
       "contributing/contribute_codeowners"
     ],
     "Technical Reference": [
+      {
+        "type": "subcategory",
+        "label": "General",
+        "ids": [
+          "resources/pcap",
```
hcgatewood: For both ids: prefix them both with either `dev_` or `ref_` for namespacing purposes

docs/readmes/resources/useful_links.md
```diff
@@ -0,0 +1,14 @@
...
+hide_title: true
+---
+
+*Last Updated: 4/20/2021*
+
+# Useful Links
+
+This is a list of links that may be useful to understand Magma
+
+- [Netmanias](https://www.netmanias.com/)
```
hcgatewood: Add some minor info about what these links contain

hcgatewood (approved):
Thanks for adding. Couple small comments before merge

PR #6816, opened 2021-05-07
AGW: pipelineD: set first IP address from UE-IP-Block to gtp-br0

### LINKED ISSUE #3829: modify gtp_br0 address when UE IP block is updated
### Your Environment

- **Version:** 1.3.1-1605904866-7933c828
- **Affected Component:** AGW, NMS
- **Affected Subcomponent:** OVS Bridge (gtp_br0) and mobilityd
- **Deployment Environment:** AWS (orc8r), bare-metal (AGW)

### Describe the Issue

Today UE IP block is set to 192.168.128.0/24 and the first IP in this range is used for the gtp_br0 interface on the access gateway. In order to expand to more than 244 UEs (first 10 in the range are reserved), this UE CIDR block needs to be expanded to a larger subnet. 

gtp_br0 interface address is assigned statically (192.168.128.1) and because the first 10 IPs in the UE block are ignored by mobilityd, this IP is always available for the AGW's gtp_br0 interface. This is not the case if this IP range were to expand/change.

If the UE block is expanded to a larger space (e.g. 192.168.0.0/16), there is a very high likelihood that the gtp_br0 address can be allocated by mobilityd to a UE. In this scenario, return traffic for the UE will be black-holed locally.

Furthermore, if for some reason the new CIDR block is entirely different than the 192.168 space, the gtp_br0 interface would have to be updated accordingly. 

**To Reproduce**
NA

**Expected behavior**
* gtp_br0 address should be dynamically selected based on the UE IP range or manually (and explicitly) set gtp_br0 IP from the NMS. 

**Screenshots**
NA

**Additional context**
NA

pshelar:
Today the IP address is hardcoded in config file. This patch sets the
IP address from configured UE IP block.
This safely allow any CIDR as a UE block.

## Test Plan
`make test`
sanity integ tests.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/inout.py
+++ b/lte/gateway/python/magma/pipelined/app/inout.py
@@ -12,6 +12,8 @@ limitations under the License.
 """
 import ipaddress
 import threading
+import subprocess
+import sys
 from collections import namedtuple
 
 from lte.protos.mobilityd_pb2 import IPAddress
@@ -34,6 +36,7 @@ from magma.pipelined.openflow.registers import (
     Direction,
     load_direction,
 )
+
 from ryu.controller import ofp_event
 from ryu.controller.handler import MAIN_DISPATCHER, set_ev_cls
 from ryu.lib import hub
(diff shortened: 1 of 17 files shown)


lte/gateway/python/magma/pipelined/app/inout.py
```diff
@@ -545,6 +554,40 @@ def _handle_barrier(self, ev):
...
+    def _set_lte_bridge_ip_address(self):
+        if self.config.setup_type != 'LTE':
+            return
+
+        if self.config.ue_ip_block is None:
+            return
+
+        try:
+            first_ip = BridgeTools.get_ip_addr_gtp_br(self.config.ue_ip_block)
+            network_mask = self.config.ue_ip_block.split('/')[1]
```
ulaskozat, line 566: Does this preclude to have multiple IP blocks?

pshelar: Not really.
But I am not sure if we have fully validated support for multiple IP blocks. It is planned for 1.6.

pshelar:
This needs changes after AON restructuring. closing for now.

PR #6828, opened 2021-05-07
devops: Turn off Golang dependabot

hcgatewood:
## Summary

Dependabot PRs will always fail insync-checkin. Until we determine a solution for that (not high-pri), turning off to avoid noise.

## Additional Information


themarwhal (approved):
lgtm

electronjoe (approved):

Issue #6829, opened 2021-05-07
[MME][S8_task] Use a random assignment for sgw control plane TEID


uri200:
In order to avoid collisions between agw teids, we need to assign sgw control plane randomly.

uri200:
closing this issue since it is duplicated

https://github.com/magma/magma/issues/6887

PR #6830, opened 2021-05-07
[mme] Modify directoryd api to include fields

### LINKED ISSUE #6886: [Orc8r] Add control plane SGW Teid to directoryd on AGW
Since Create Bearer Request only has SGW control plane TEID, we need a way to find from FEG what AGW is serving an specific TEID.

For that reason we will add TEID to directoryd so FEG can do a query to find out AGW HwId.

Tasks to do are:
- Add a call to directoryd AGW api to add fields to the record
- Add new table on directoryd to match Teid to AGW_HwId
- Add Create Bearer Request to feg relay using TEID as a key to find serving AGW

uri200:

## Summary

This PRs adds the ability on directoryd client at MME to add store fields 

This results on directoryd entries like
```
{
  "identifiers": {
    "sgw_c_teid": "1"
  },
  "location_history": [
    "d35eb697-ad83-4dee-a049-6e15273dd0ee"
  ],
  "py/object": "magma.directoryd.rpc_servicer.DirectoryRecord"
}
```

Right now this is just an easy way to have teid available on directoryd. This way we can see the feasibility of finding a subscriber based on TEID from the FEG. 

As you can see the record has teid and location_history (AGW HwID). So the FEG should be able to find the AGW just with a TEID.

The idea is to add the `sgw_c_teid` every time we create session for a subscriber.
The whole record will be deleted on delete_session.

The entries at `sgw_c_teid` will be modified every time there is a addition or deletion of default bearer (this is still pending to find out where in the code place it)

Note right now AGW assigns TEID starting from 1. We will change this to a random assignment at https://github.com/magma/magma/issues/6829

## Test Plan

make test
## Additional Information

### DIFF
--- a/lte/gateway/c/oai/lib/directoryd/GatewayDirectorydClient.cpp
+++ b/lte/gateway/c/oai/lib/directoryd/GatewayDirectorydClient.cpp
@@ -57,22 +57,26 @@ GatewayDirectoryServiceClient::GatewayDirectoryServiceClient() {
 bool GatewayDirectoryServiceClient::UpdateRecord(
     const std::string& id, const std::string& location,
     std::function<void(Status, Void)> callback) {
-  GatewayDirectoryServiceClient& client = get_instance();
-
   UpdateRecordRequest request;
-  Void response;
-
   request.set_id(id);
   request.set_location(location);
+  return GatewayDirectoryServiceClient::updateRecordImpl(request, callback);
+}
 
-  // No values for fields param in UpdateRecord
-  auto fields = request.fields();
-  std::unordered_map<std::string, std::string> fields_map;
+bool GatewayDirectoryServiceClient::UpdateRecordField(
+    const std::string& id, const std::string& field_key,
+    const std::string& field_value,
+    std::function<void(Status, Void)> callback) {
+  UpdateRecordRequest request;
+  request.set_id(id);
+  auto update_fields = request.mutable_fields();
+  update_fields->insert({field_key, field_value});
+  return GatewayDirectoryServiceClient::updateRecordImpl(request, callback);
+}
 
-  Map<std::string, std::string> proto_fields(
-      fields_map.begin(), fields_map.end());
-  fields = proto_fields;
-
+bool GatewayDirectoryServiceClient::updateRecordImpl(
+    UpdateRecordRequest& request, std::function<void(Status, Void)> callback) {
+  GatewayDirectoryServiceClient& client = get_instance();
   // Create a raw response pointer that stores a callback to be called when the
   // gRPC call is answered
   auto local_response =
--- a/lte/gateway/c/oai/lib/directoryd/directoryd.h
+++ b/lte/gateway/c/oai/lib/directoryd/directoryd.h
@@ -36,6 +36,8 @@ bool directoryd_remove_location(char* imsi);
 
 bool directoryd_update_location(char* imsi, char* location);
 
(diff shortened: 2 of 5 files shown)


lte/gateway/c/oai/tasks/sgw_s8/sgw_s8_handlers.c
```diff
@@ -506,6 +508,13 @@ void sgw_s8_handle_create_session_response(
     sgw_remove_sgw_bearer_context_information(
         sgw_state, session_rsp_p->context_teid, imsi64);
   }
+
+  char imsi_str[IMSI_BCD_DIGITS_MAX + 1];
+  IMSI64_TO_STRING(imsi64, (char*) imsi_str, IMSI_BCD_DIGITS_MAX);
+  char teidString[16];
+  sprintf(teidString, "%u", session_rsp_p->context_teid);
+  directoryd_update_field(imsi_str, "sgw_c_teid", teidString);
```
rsarwad: Need to update only for success case.

uri200: yep. I put it here as an example. Not sure where to put it, either s8_task or mme_task

What we need is to get all the available Teids and put them in a comma separated string. So directoryd record would be like 

```
{
  "identifiers": {
    "sgw_c_teid": "1467, 3688, 31093"
  },
  "location_history": [
    "d35eb697-ad83-4dee-a049-6e15273dd0ee"
  ],
  "py/object": "magma.directoryd.rpc_servicer.DirectoryRecord"
}
```

where `sgw_c_teid` is all the default bearer teids. for that subscriber

uri200: for now we can keep it here so we can test it in teravm using FEG. 
Once it is landed we can move around this statement and place it in a proper place

ulaskozat:
@uri200  can you provide more context on when this value should be stored, how it will be used and when it gets deleted?

uri200:
> @uri200 can you provide more context on when this value should be stored, how it will be used and when it gets deleted?

Just added some more context to the pr

lte/gateway/c/oai/lib/directoryd/GatewayDirectorydClient.cpp
```diff
@@ -57,22 +57,26 @@ GatewayDirectoryServiceClient::GatewayDirectoryServiceClient() {
...
     const std::string& id, const std::string& location,
     std::function<void(Status, Void)> callback) {
-  GatewayDirectoryServiceClient& client = get_instance();
-
   UpdateRecordRequest request;
-  Void response;
-
   request.set_id(id);
   request.set_location(location);
+  return GatewayDirectoryServiceClient::updateRecordImpl(request, callback);
```
ardzoht, line 63: nit: remove the `GatewayDirectoryServiceClient` context here? since the function it's on the same namespace

lte/gateway/c/oai/lib/directoryd/directoryd.h
```diff
@@ -36,6 +36,8 @@ bool directoryd_remove_location(char* imsi);
 
 bool directoryd_update_location(char* imsi, char* location);
 
+bool directoryd_update_field(char* imsi, char* key, char* value);
```
ardzoht: should this be named `directoryd_update_record_field`? seems more aligned with what the function does and with the function name on GatewayDirectorydClient

ardzoht (approved):
lgtm

PR #6831, opened 2021-05-07
[NMS] Add  gateway YML configs editor 

HannaFar:

## Summary

Add  gateway YML configs editor 
 
## Additional Information

### DIFF
--- a/nms/app/packages/magmalte/app/views/equipment/GatewayYMLConfig.js
+++ b/nms/app/packages/magmalte/app/views/equipment/GatewayYMLConfig.js
@@ -1,0 +1,184 @@
+/**
+ * Copyright 2020 The Magma Authors.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @flow
+ * @format
+ */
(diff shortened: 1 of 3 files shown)


nms/app/packages/magmalte/app/views/equipment/GatewayYMLConfig.js
```diff
@@ -0,0 +1,170 @@
...
+        await MagmaV1API.postNetworksByNetworkIdGatewaysByGatewayIdCommandGeneric(
+          {
+            networkId,
+            gatewayId,
+            parameters: {
+              command: 'bash',
+              params: {
+                // $FlowIssue[incompatible-call]
+                shell_params: [
+                  `-c 'printf -- "${config}" > /etc/magma/test.yml'`,
```
andreilee: Avoid single quotes `'` like the plague.

https://www.gnu.org/software/bash/manual/html_node/Single-Quotes.html

```suggestion
                  `"cat >/etc/magma/test.yml <<EOL ${config} EOL "`,
```

With the suggested change, you're never enclosing strings in single quotes.

andreilee (changes requested):
Make the suggested change to the generic bash command that gets run so that single quotes in YML files will work.

Make sure that the service page displays something if the NMS can't connect to the gateway on the other end.

Also make sure that when the save operation succeeds, the user gets some sort of snackbar notification

andreilee (approved):
Thanks for working on this. This will be a great MVP and will move us towards a Magma where operators can do everything through the NMS. Something to note is that `RUNNING_SERVICES` includes services that do not run on either the CWAG or FeG. This YML configs editor is useful for those gateway types as well. This is something that we can fix later, but please annotate `RUNNING_SERVICES` constant with the note that it is a list of services that runs on the LTE AGW.

andreilee:
Please fix the eslint issues before merging

Issue #6636, opened 2021-05-03
[Orc8r] Add test utilities to be used in cluster tests


karthiksubraveti:

karthiksubraveti:
With inputs containing kubectl, api endpoints and certs. Need to add test utilities which can enable Orc8r tests to 
- invoke Orc8r APIs
- helper utilities to enable adding networks, gateways, subscribers etc. 
- upgrading or downgrading helm charts from test setup/teardown

hcgatewood:
@karthiksubraveti the only difficult thing here is our Helm charts are actually a bit unwieldy. I think for now let's just have orcl install "latest" Helm charts and not mess with it.

Later we can add support for messing with Helm charts directly from tests, but I probably want to clean up our charts first

hcgatewood:
@karthiksubraveti the only difficult thing here is our Helm charts are actually a bit unwieldy. I think for now let's just have orcl install "latest" Helm charts and not mess with it.

Later we can add support for messing with Helm charts directly from tests, but I probably want to clean up our charts first

PR #6858, opened 2021-05-11
orc8r: Generate northbound API stubs

hcgatewood:
## Summary

This PR generates, and keeps checked-in + updated, Golang stubs for the Orc8r northbound interface (as exposed by obsidian).

It's a large PR, but 99+% of it is generated code. See the gen.go and makefile changes for the substance.

The Golang client stubs will be used by our Orc8r scale tests to register gateways, add subscribers, scrape metrics, etc.

## Test Plan

Will test during scale test impl. Quick glance looks like it's feasible to use mutual TLS with these stubs.

## Additional Information


themarwhal (approved):

mpgermano (approved):

karthiksubraveti:
closes #6636

PR #6864, opened 2021-05-11
feat: [Sentry][SessionD] Add hwid tag and move to a separate file

themarwhal:

## Summary
Add hwid as a tag so events can be differentiated properly
Also experimenting with adding a transaction tag... might revert if it's not helpful.

## Test Plan
testing manually 
Screen Shot 2021-05-11 at 11 38 05 AM

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalSessionManagerHandler.cpp
+++ b/lte/gateway/c/session_manager/LocalSessionManagerHandler.cpp
@@ -22,6 +22,7 @@
 #include "GrpcMagmaUtils.h"
 #include "LocalSessionManagerHandler.h"
 #include "magma_logging.h"
+#include "SentryWrappers.h"
 #include "Utilities.h"
 
 using grpc::Status;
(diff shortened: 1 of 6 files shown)


uri200 (approved):

lte/gateway/c/session_manager/LocalSessionManagerHandler.cpp
```diff
@@ -46,6 +47,7 @@ LocalSessionManagerHandlerImpl::LocalSessionManagerHandlerImpl(
 void LocalSessionManagerHandlerImpl::ReportRuleStats(
     ServerContext* context, const RuleRecordTable* request,
     std::function<void(Status, Void)> response_callback) {
+  set_sentry_transaction("ReportRuleStats");
```
uri200, line 50: so besides crashes you want to report transactions?

Isn't it the same as the metrics/events we already track?

themarwhal: The transaction isn't an event on its own. It's more like a tag. So if we get a crash while handling ReportRuleStats, it'll just say it was during the ReportRuleStats transaction.

PR #6873, opened 2021-05-12
orc8r: Support subscriber bulk-upload

hcgatewood:
## Summary

Efficient subscriber bulk-upload is a GA requirement. I was going to have my intern land this impact, but it's blocking scale test framework work, so I took a pit stop to land it.

This endpoint is placed under `subscribers_v2`, and supports uploading any number of subscribers -- so it can fully replace the existing subscriber post endpoint.

Needed to upgrade the `client_max_body_size` for the northbound/REST API interface in nginx conf. Set it to the max value for simplicity. For reference, the (barebones) set of 20k subs I uploaded were ~5mb. Eventually the NMS will need to support chunking the bulk create requests, but for now we can probably get away with a single request.

Note: NMS's current bulk-upload seems to overwrite any existing subs. This new endpoint *does not overwrite* and will throw an error if an IMSI already exists. @karthiksubraveti @andreilee please weigh in on whether this is desirable. For reference, it will list the set of already-existing subscribers in the error message.

## Test Plan

- [x] New tests
- [x] Manual: bulk-provision 20k subs via Swagger UI took ~30 seconds on my dev machine

Screen Shot 2021-05-11 at 6 29 56 PM

## Additional Information


themarwhal (approved):
lgtm!

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -19,6 +19,9 @@ import (
 	"regexp"
 	"strconv"
 
+	"github.com/hashicorp/go-multierror"
```
mpgermano, line 22: Wrong import ordering unless we've recently changed this?

mpgermano (approved):
LGTM

andreilee:
@hcgatewood Throwing an error if an IMSI already exists is fine

andreilee:
Bulk edit endpoint would be nice too. Is that easy to do?

PR #6893, opened 2021-05-12
Add sentry release github action

themarwhal:

## Summary
See doc: https://github.com/marketplace/actions/sentry-release
This will help us pinpoint which commits introduce specific errors.

## Test Plan
test on master???
Will revert if it is broken

## Additional Information

### DIFF
--- a/.github/workflows/sentry-release.yml
+++ b/.github/workflows/sentry-release.yml
@@ -1,0 +1,22 @@
+---
+name: Sentry release on master commits
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches: [master]
+jobs:
+  sentry-release:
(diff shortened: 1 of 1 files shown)


electronjoe (approved):
There are a couple YAMLLint comments to address! Otherwise LGTM.  The "truthy value should be..." one is unavoidable so I annotate it away (see other GH actions for example).

themarwhal:
> There are a couple YAMLLint comments to address! Otherwise LGTM. The "truthy value should be..." one is unavoidable so I annotate it away (see other GH actions for example).

Yeah wasn't sure how to fix that one, but I fixed the other space one.
Edit: fixed all lint

PR #6897, opened 2021-05-13
[feg] Basic FeG accounting service

emakeev:

Basic FeG accounting service to work with external ANSC contract/accounting

## Summary

A new FeG cloud independant accounting service to collect optional accounting info & forward it to an external augmented network SC contract accounting service.

## Test Plan

added a unit test with mock ANSC service.

## Additional Information


feg/cloud/go/services/basic_acct/servicers/service.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+// Set Config updates current server configs
+func (s *BaseAccService) SetConfig(cfg *Config) {
+	if s != nil && cfg != nil {
+		s.Lock()
+		s.cfg = cfg
+		s.Unlock()
+	}
+}
```
uri200, line 86: why do we need locks to set and read the config? Is this config changing over time. Are we going to read it while is being written?

feg/cloud/go/services/basic_acct/servicers/service.go
```diff
@@ -0,0 +1,232 @@
...
+func NewBaseAcctService() *BaseAccService {
+	cfg := &Config{
+		RemoteAddr:   *remoteAddrFlag,
+		RootCaCert:   *caCertFlag,
+		ClientCrt:    *clientCertFlag,
+		ClientCrtKey: *clientKeyFlag,
+		NoTls:        *notlsFlag,
+		Insecure:     *insecureFlag,
+	}
+	_, _, err := config.GetStructuredServiceConfig(feg.ModuleName, basic_acct.ServiceName, cfg)
```
uri200, line 98: i have seen `GetStructuredServiceConfig` before in or8r. I understand that the config struct has to have the same field names and types as the yaml config fie (?)

feg/cloud/go/services/basic_acct/servicers/participants.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+const (
+	// Minimal length of PLMNID
+	MinPlmnIdLen = 5
+	MaxPlmnIdLen = 6
+)
+
+// RetrieveParticipants retrieves GW Network, FeG Network & GW logical ID from the caller's context
+func RetrieveParticipants(ctx context.Context, session *protos.AcctSession) (provider, consumer, gw string, err error) {
```
uri200, line 39: is this function the reason why removed some fields from the protos?

By any chance, Is the bootstraper the one responsible on setting up this identity? Or is it sent in every single request?
http://github.com/magma/magma/blob/b8ac7839b660ac38a5a7949bc6bf031d5182f6d6/orc8r/gateway/go/services/bootstrapper/service/bootstrapper.go#L308-L308

feg/cloud/go/services/basic_acct/servicers/service.go
```diff
@@ -0,0 +1,232 @@
...
+
+	resp, err := client.Start(outCtx, anSession)
+	if err != nil {
+		return nil, err
+	}
+	return &protos.AcctSessionResp{
+		ReportingAdvisory: &protos.AcctSessionResp_ReportLimits{
+			OctetsIn:       resp.GetReportingAdvisory().GetOctetsIn(),
+			OctetsOut:      resp.GetReportingAdvisory().GetOctetsOut(),
+			ElapsedTimeSec: resp.GetReportingAdvisory().GetElapsedTimeSec(),
```
uri200, line 135: is this the total quota given to the user?

feg/cloud/go/services/basic_acct/servicers/service.go
emakeev: Yes, the yaml config parsing is permissive though, if parameter names match, it reads them, if not - it should ignore them.

feg/cloud/go/services/basic_acct/servicers/service.go
emakeev: Currently, configs are initialized on init and it means - the service would need to be restarted on cfg changes, but there is Get & Set APIs and they require synchronization and I plan to make changes dynamic, not requiring restart in the future.

feg/cloud/go/services/basic_acct/servicers/participants.go
emakeev: This one starts the process, but the cert SN and its GW identity association are created by the cloud bootstrapper service & this mapping only exists on the cloud, GW itself has no control over it.

feg/cloud/go/services/basic_acct/servicers/service.go
emakeev: It's mostly for the future use since neither CWAG nor AGW currently have means of enforcing it. When session is created, the smart contract accounting service can request that the update for this session should be issued as soon as new usage reaches either  OctetsIn, OctetsOut or after ElapsedTimeSec from previous update, whatever comes first.

feg/cloud/go/services/basic_acct/servicers/participants.go
uri200: ooh so the way it is identified is through the cert SN👍

uri200 (approved):

PR #2718, opened 2020-09-11
[SessionD] Schedule bearer creation for Session Create

themarwhal:
Signed-off-by: Marie Bremner <marwhal@fb.com>

## Summary
This PR adds a change to delay dedicated bearer creation by a configured amount of time for session creation. This is needed since, we need to send the bearer request after UE attach is completed.

When we process the create session response, if a bearer creation is needed we schedule the request to be sent in the event loop. When the function is called, we double check that the request is still valid before making the request. The conditions we check are:
1. The session is exists
2. The session is active
3. The rule exists in the session
4. A bearer creation is still needed -> meaning that the rule still has a qos / is not already tied to a bearer etc.

This amount is configured in `sessiond.yml` under : `bearer_creation_delay_on_session_init`. 

## Test Plan
Additional SessionD unit tests to cover the edge cases
SessionD unit tests + S1AP tests
Also did some manual testing by modifying the dynamic rule PolicyDB returns on SessionCreation to include a QoS field and running an `attach-detach` test. I've set ` bearer_creation_delay_on_session_init: 1000`
```
def get_allow_all_policy_rule(
    subscriber_id: str,
    apn: str,
) -> PolicyRule:
    policy_id = _get_allow_all_rule_id(subscriber_id, apn)
    return PolicyRule(
        id=policy_id,
        priority=2,
        flow_list=_get_allow_all_flows(),
        tracking_type=PolicyRule.TrackingType.Value("NO_TRACKING"),
        qos = FlowQos(
            qci=5,  # qci value [1 to 9]
            max_req_bw_ul=10000000,  # MAX bw Uplink
            max_req_bw_dl=15000000,  # MAX bw Downlink
            gbr_ul=1000000,  # GBR Uplink
            gbr_dl=2000000,  # GBR Downlink
            arp=QosArp(
                priority_level=15,  # ARP priority
                pre_capability=1,  # pre-emption capability
                pre_vulnerability=1,  # pre-emption vulnerability
            ),
        ),
    )
```
Here is the logs from SessionD, the important lines are "Scheduling a bearer creation request for newly created IMSI001010000000001-623138 in 1000 ms" @ Sep 11 12:38:48 and "Creating dedicated bearers for IMSI001010000000001, 192.168.128.13, rules={ allowlist_sid-IMSI001010000000001-magma.ipv4 }" @ Sep 11 12:38:49.

[...]

PR #6898, opened 2021-05-13
refactor: remove bearer scheduling for create session

themarwhal:

## Summary
A few months ago, we made a change in SessionD to not process any rule installs for new sessions until we receive `UpdateTunnelIds` from MME. This means that by the time we install rules and potentially make dedicated bearer requests to MME, the default bearer is already created and ready. (We know this because the `UpdateTunnelIds` request from MME contains the teid for the default bearer.)
PR that introduced the delay: https://github.com/magma/magma/pull/2718

When we initially added logic to optionally send dedicated bearer requests upon processing the initial policies, we did not have this two phase session creation. In order to get around the fact that MME is not prepared to create dedicated bearers at the time when `CreateSession` request/response is made, we added few seconds of delay before sending the dedicated bearer request. 

As described above, we do not need this delay anymore since we don't process any rules until we receive `UpdateTunnelIds`.

This PR removes the config value / logic relating to the dedicated bearer creation delay for initial policy processing. 

## Test Plan
unit tests
CI / integration test

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -32,10 +32,9 @@
 #include "Utilities.h"
 
 namespace magma {
-uint32_t LocalEnforcer::BEARER_CREATION_DELAY_ON_SESSION_INIT = 2000;
-uint32_t LocalEnforcer::REDIRECT_FLOW_PRIORITY                = 2000;
-bool LocalEnforcer::SEND_ACCESS_TIMEZONE                      = false;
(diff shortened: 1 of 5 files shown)


uri200:
so basically that delay was to give MME some time to get ready, specially with OVS. Now that happens before UpdateTunnelId.

themarwhal:
> so basically that delay was to give MME some time to get ready, specially with OVS. Now that happens before UpdateTunnelId.

yes exactly! We were trying to get around MME not being ready by putting in a delay before, but now we wait for `UpdateTunnelId` which guarantees that MME is ready.

uri200 (approved):

ulaskozat (approved):
LGTM

PR #6901, opened 2021-05-13
[feg] Simplify basic acct session

emakeev:

## Summary

Remove user_network_id, serving_network_id, serving_gateway_id - they can be discovered via caller CTX; Add serving_apn.

## Test Plan
build.py -g; Unit tests

## Additional Information


feg/protos/basic_acct.proto
```diff
@@ -49,16 +49,12 @@ message AcctSession {
...
-    // ID of the user network (MNO, Identity Provider, etc.)
-    string user_network_id = 5;
     // unique session ID
-    string session_id = 6;
-    // ID of the service provider network
-    string serving_network_id = 7;
+    string session_id = 5;
     // ID of the service provider gateway (AP, AGW, enodeb, etc.)
-    string serving_gateway_id = 8;
+    string serving_apn = 6;
```
electronjoe, line 55: If serving_apn is a rename of the existing session_id field, then this update is "safe". But if it's a new field entirely it should avoid re-use of the existing tag (6) and should instead use a new and not-previously-allocated tag.  This is best practice for backwards compatibility, even if you believe or know that in this case all producers and consumers of the FEG API will be synchronized (not sure if that is true).

feg/protos/basic_acct.proto
```diff
@@ -49,16 +49,12 @@ message AcctSession {
...
     // available QoS at the caller's site
-    AcctQoS    available_qos = 9;
+    AcctQoS available_qos = 7;
```
electronjoe, line 57: Regardless of whether this API is guaranteed to be updated uniformly and instantly across our ecosystem - you should avoid changing tag identifiers (here from 9 to 7) as this is not backwards compatible and it's hard to review these changes without system context. Tag identifiers do not have to be in-order within the proto definition. Unless you intend for the new available_qos to be a new field with non-backward-compatibility, leave it at tag 9?

feg/protos/basic_acct.proto
```diff
@@ -49,16 +49,12 @@ message AcctSession {
         bytes  hardware_addr = 3;
         string name = 4;
     }
-    // ID of the user network (MNO, Identity Provider, etc.)
-    string user_network_id = 5;
```
electronjoe, line 53: Instead of deleting the tag in a PR that deprecates it's use, opt to either use `string user_network_id = 5 [deprecated=true];` (if you want deprecation to be soft / warning based in compilers) - or remove the line entirely but add the tag to the reserve list:
```
message AcctSession {
    reserved 5,other_tags_if_any;

    oneof user { ...
```

The reserved annotations inform future modifications of the protobuf definition - that client or server code out there in the wild may already have this tag associated with some now-unused data field... and it should be avoided for safety and backwards compatibility.

electronjoe: Some more details on the nomenclature and impact are available at https://developers.google.com/protocol-buffers/docs/proto

uri200: Hey @electronjoe , I think this is an experimental service that hasn't been used yet.

On services that are being used we don't remove old values, we just leave them there. But maybe for the future it is a good practice to, at least, use `[deprecated]`

uri200 (approved):
@emakeev does it mean the client adds `user_network_id, serving_network_id, serving_gateway_id` to the request?

feg/protos/basic_acct.proto
emakeev: This is a brand new proto, there is no service using it yet. All changes are safe for now.

feg/protos/basic_acct.proto
emakeev: It doesn't have to be backward compatible yet.

feg/protos/basic_acct.proto
emakeev: We generally know how to preserve backward compatibility of protos, but in this particular case, it's not necessary, this is still a proto definition state, no services or APIs using this proto exist yet. So, there is no reason to pollute the design with unnecessary, depreciated fields.

emakeev:
@uri200 user_network_id, serving_network_id, serving_gateway_id can be inferred from the client (GW) context on the service side (see https://github.com/magma/magma/pull/6897 for example). In any case a typical GW (AGW, CWAG, etc.) is not aware of either what network it belongs to or wich federation its network belongs to, so - it won't be able to provide most of those values.

PR #6912, opened 2021-05-13
clean: clean up unused functions and param types in SessionCredit

themarwhal:

## Summary
Recently realized that there were a few unused functions that live in SessionCredit so cleaning those up.

### Functions Removed
* `set_grant_tracking_type`
* `set_received_granted_units`
* `compute_reporting_limit`
* `apply_reporting_limits`

### Modified funtions / variables
* We have some functions that do *not* modify state so marking those as consts
* The general guidelines for C++ is to have output parameters as pointers, not references. So modifying the type there to follow google standards. This way it's clear from looking at the signature that the variable is to be modified. 
  * Although a bigger TODO here would be to get rid of `SessionCreditUpdateCriteria` entirely. (but that is for some other time)
* rename all `SessionCreditUpdateCriteria` to be `credit_uc`
* I also moved some functions that are only used internally in SessionCredit to be marked as private

### Other
* I moved `Usage` and `TotalCreditUsage` to `Types.h` since they are used in a lot of different places. (And `TotalCreditUsage` doesn't really belong in SessionCredit anymore since it's only used outside of the class)

## Test Plan
unit tests

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/ChargingGrant.cpp
+++ b/lte/gateway/c/session_manager/ChargingGrant.cpp
@@ -83,7 +83,7 @@ CreditValidity ChargingGrant::is_valid_credit_response(
       !gsu.total().is_valid() && !gsu.rx().is_valid() && !gsu.tx().is_valid();
   if (gsu_all_invalid) {
     if (update.credit().is_final() || credit_validity == TRANSIENT_ERROR) {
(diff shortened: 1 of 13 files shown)


uri200 (approved):
those functions were used by the previous algorithm.
Lets remember this PR just in case we need them back

PR #6914, opened 2021-05-13
feg: cwf: Fix feg_lte missing serdes

hcgatewood:
## Summary

FeG was including serdes from *lte service* rather than all serdes in *lte module*. Side note: we have to do per-service like this otherwise we run into merge conflicts later up in the hierarchy.

Reported issue:

```
$ curl -X PUT "https://api.example.com/magma/v1/feg_lte/feg_lte/subscriber_config" -H "accept: application/json" -H "content-type: application/json" -d "{ \"network_wide_base_names\": [ \"base_1\" ], \"network_wide_rule_names\": [ \"omni_1\" ]}"

{
  "message": "failed to marshal update: failed to serialize config network_subscriber_config: no serde in registry for type network_subscriber_config"
}
```

## Test Plan

Added regression test -- failed before fix, passes after.

## Additional Information

### DIFF
--- a/cwf/cloud/go/serdes/serdes.go
+++ b/cwf/cloud/go/serdes/serdes.go
@@ -17,6 +17,8 @@ import (
 	"magma/cwf/cloud/go/cwf"
 	"magma/cwf/cloud/go/services/cwf/obsidian/models"
 	feg_models "magma/feg/cloud/go/services/feg/obsidian/models"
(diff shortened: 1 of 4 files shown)


uri200 (approved):

emakeev:
[cwf][feg] in the title, to be consistent?

themarwhal (approved):

Issue #6870, opened 2021-05-11
[feg][orc8r] Use of Network Level Federation Configs is broken


emakeev:
On Federation network, it was possible to specify configuration on Network level ONLY and get this config distributed to the network's Federated Gateways.
It is an important feature since FeG networks may have active and standby FeGs and their configuration often identical, so network level config is the most appropriate way to configure them.
If somebody wants to overwrite the configs for specific gateway - they can specify GW level config and it'll take precedence over network one for this GW.
Now, after some of the changes, neither swagger API nor NMS allows to NOT to provide federation GW configs.
It means that I had to copy/paste Network FeG configs to every Federation Gateway that I added to the network.
It's obviously error prone and not documented or recommended in our docs (even if you are trying to create some empty federation GW configs and then delete them later, REST API will return an error & won't do it).

**Expected behavior**

An operator/admin should be able to add Federation Gateway to the network without explicit federation configs & the gateway should be provided with network level configs. This functionality was supported and was broken recently.

PR #6925, opened 2021-05-13
[feg] Issue 6870 fix

emakeev:

## Summary

A possible fix fo https://github.com/magma/magma/issues/6870. This change makes federation section optional for Federation Gateways.

## Test Plan

unit tests; test on staging: TBD

## Additional Information

### DIFF
--- a/feg/cloud/go/services/feg/obsidian/models/mutable_federation_gateway_swaggergen.go
+++ b/feg/cloud/go/services/feg/obsidian/models/mutable_federation_gateway_swaggergen.go
@@ -28,8 +28,7 @@ type MutableFederationGateway struct {
 	Device *models5.GatewayDevice `json:"device"`
 
 	// federation
-	// Required: true
-	Federation *GatewayFederationConfigs `json:"federation"`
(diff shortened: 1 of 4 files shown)


tmdzk (approved):

themarwhal (approved):
lgtm 
the network federation config is mandatory right?
So if the gateway one is empty it's overridden by the network one?

emakeev:
> lgtm
> the network federation config is mandatory right?
> So if the gateway one is empty it's overridden by the network one?

The FeG GW config is initially set to its network federation configs and then, if GW federation section is not empty, the federation configs are completely overwritten with the GW's version. I didn't change the network config requirements, they should remain required.

uri200:
will this not hide some fields on the examples?

themarwhal:
@emakeev just letting you know the insync check is failing :)

emakeev:
> @emakeev just letting you know the insync check is failing :)

Thanks. Will rebase & fix.

emakeev:
> will this not hide some fields on the examples?
It might, but using per GW configuration should be an exception, in most cases - federation is configured on the network level.

PR #6861, opened 2021-05-11
Modified code to update sgw_c_teid to directoryd for multi pdns

rsarwad:
[lte][agw]:  Modified code to update sgw_c_teid to directoryd for multi pdns

## Summary
Modified code to update sgw_c_teid to directoryd for multi pdns
## Test Plan
1. Attached roaming UE in Tera vm
2. Executed s1ap sanity test suite
3. Executed multi-pdn test case and verified with log that sgw_c_teid will be updated with sgw's control plane teid

PR #6928, opened 2021-05-14
[or8r][directoryd] Add a TeidHwId mapping to directoryd

uri200:

## Summary

When core sends Create Bearer Request, Sgw Control Plane TEID is the only field to identify the subscriber on that request.

For that reason we need to maintain a secondary state on directoryd that maps Teid to HwId so we can find what AGW is serving that subscriber.

Note this is very similar to the SessionId to IMSI map that we already have in directoryd,

Directoryd records belonging to roaming subscribers will include a new parameter under the key `sgw_c_teid`. The addition of that parameter will be in this PR https://github.com/magma/magma/pull/6861. The directoryd record will look like this:

```
{
  "identifiers": {
    "sgw_c_teid": "1"
  },
  "location_history": [
    "d35eb697-ad83-4dee-a049-6e15273dd0ee"
  ],
  "py/object": "magma.directoryd.rpc_servicer.DirectoryRecord"
}
```

## Test Plan

Make precommit

## Additional Information


electronjoe:
Nice tests!

themarwhal:
Does teid map uniquely to a gw? Or is it a combination of session id and teid?

uri200:
> Does teid map uniquely to a gw? Or is it a combination of session id and teid?

Collisions has been a debate the last two weeks with other members of the team.  

Unfortunately PGW sees magma as a unique SGW, so ideally we should maintain a service to provide teids per network and PLMN served. 

However, that would break the concept the cloud should not have any specific state from our subscribers. For that reason, for now, we are going to change the sgw C TEID assignment MME does from sequential to random. This way the collision of teid betweern AGW is reduced. We have calculated that with 20k Users (all in roaming), the collision rate would be around 4.5%. But that collision drops to 0.1% when we have lower that to 2k subscribers.

Since collisions only matter for roaming subscribers and roaming subscribers are temporal, for now we will rely on statistics. 

In the future we may need a central service, but not just for TEIDs but for many other indicators that are still currently relying on random processes. 


image

themarwhal:
> > Does teid map uniquely to a gw? Or is it a combination of session id and teid?
> 
> Collisions has been a debate the last two weeks with other members of the team.
> 
> Unfortunately PGW sees magma as a unique SGW, so ideally we should maintain a service to provide teids per network and PLMN served.
> 
> However, that would break the concept the cloud should not have any specific state from our subscribers. For that reason, for now, we are going to change the sgw C TEID assignment MME does from sequential to random. This way the collision of teid betweern AGW is reduced. We have calculated that with 20k Users (all in roaming), the collision rate would be around 4.5%. But that collision drops to 0.1% when we have lower that to 2k subscribers.
> 
> Since collisions only matter for roaming subscribers and roaming subscribers are temporal, for now we will rely on statistics.
> 
> In the future we may need a central service, but not just for TEIDs but for many other indicators that are still currently relying on random processes.
> 
> image

oooh interesting. Thanks for the detailed info.

emakeev:
What if you add a timestamp to the teid -> hwid record, then in the MapSgwCTeidToHWID() implementation, you'll fail if 1) one of the tieidIDs is already in use, 2) is not mapping to the same hwid & 3) has a freshness better then some preconfigured threshold (too new to overwrite).
If a calling GW would receive such an error with indication which randomly selected TEIDs, it'll re-generate those conflicting TEIDs and reissue the MapSgwCTeidToHWID() request, you can also limit max number of retries to avoid spamming the service and add a "force overwrite" flag for the GW to use if all retries fail. This way, you may avoid collisions completely.

uri200:
> What if you add a timestamp to the teid -> hwid record, then in the MapSgwCTeidToHWID() implementation, you'll fail if 1) one of the tieidIDs is already in use, 2) is not mapping to the same hwid & 3) has a freshness better then some preconfigured threshold (too new to overwrite).
> If a calling GW would receive such an error with indication which randomly selected TEIDs, it'll re-generate those conflicting TEIDs and reissue the MapSgwCTeidToHWID() request, you can also limit max number of retries to avoid spamming the service and add a "force overwrite" flag for the GW to use if all retries fail. This way, you may avoid collisions completely.

Maybe I didn't understand it well, but note that directoryd on the cloud is synch by default every minute (on teravm i have it to 4 seconds). The synchronization happens through `state` service between local directoryd records and orcr8r directoryd. Per my understanding there is no direct interaction between MME and orc8r directoryd. 

So TEID assignment comes first at MME and after few seconds map is created. We can't use directoryd as a central way to check if teid is being used

Per my understanding syncronization is one way: from gateway to Or8cr, but not the other way around. So Teid->HwId map is not shared back with the gateways (but maybe that could be a solution to have some kind of way to check if a teid is being used) 

I think we should plan sometime in the future to have a centric ID allocator in the or8r that would provide IDs, allocating ranges of ids per gateway, network, or per any other kind of grouping.  also sessiond and other ids that are currently assigned using rand functions. We could handle there issues like stale ids or stateless.

emakeev (approved):
To the previous comment: it is also possible to use directoryd as a remote orc8r service. In this scenario MME will call directoryd RPC on the cloud instead of adding this teid locally to be synced with cloud state later. Since it'll be unary call, directoryd can provide immediate response if MME selected TEID is not unique & MME can regenerate conflicting TEID. 

Just a suggestion to avoid relatively high collision probability, LGTM othervice.

Issue #6904, opened 2021-05-13
[MME] Fix missing Cmake linking for Common library dependencies


electronjoe:
## Summary

I was struggling to get a new unit test to build, with dependency on magma's oai/common.  After much exploration, I determined that in fact the Common CMake project has dependencies on external libraries that are not specified in the linkage directives for the CMake project.  It appears others have worked around this (perhaps unknowingly) by adding these link directives to their tests and libraries, that depend on our oai/common.

Adding the missing dependencies to the common project's `target_link_libraries` has repaired my unit test - we should apply this fix AND go find all places where others have worked around this hole, so we stop copying linkage directives that are not necessary in other CMake projects...

electronjoe:
Related #5222

Issue #6350, opened 2021-04-23
[AGW] Move Folly dependency to static linking


electronjoe:
## Summary

We presently dynamically link against Folly in our runtime environments.  This presents at least a couple issues.

1. Folly explicitly indicates one should statically link
    1. Folly provides no API stability guarantees commit to commit
1. Folly is some odd 150 MB shared library
    1. This is unfortunate for containerized environments
    2. Static linking would trim a vast majority of this sizing

## Approach

This is not the only external library dependency that we might want or need to pull into our build.  We probably need to evaluate a strategy for this in our version control system (e.g. Git submodules, or use of a build system that supports fetching dependencies during build).

PR #6936, opened 2021-05-14
[MME] Clean up assorted library linking

electronjoe:
libcrypto, prometheus-cpp and yaml-cpp should be linked via the Common's CmakeList, as that is the true dependency, not mme_app. This PR moves these linkage directives. While here, I was unsure of why a crypt dependency existed - and removal had no impact on build success. So I remove it.

This works towards #6904.

### DIFF
--- a/lte/gateway/c/oai/common/CMakeLists.txt
+++ b/lte/gateway/c/oai/common/CMakeLists.txt
@@ -68,10 +68,15 @@ set(COMMON_SRC
 add_subdirectory(glogwrapper)
 add_subdirectory(redis_utils)
 
+pkg_search_module(CRYPTO libcrypto REQUIRED)
+
 add_library(COMMON ${COMMON_SRC})
 target_link_libraries (COMMON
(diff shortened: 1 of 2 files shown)


electronjoe:
These OAI RHEL build errors make me think that the RHEL Folly version might be a problem.

See #6350

RHEL is [using](https://sourcegraph.com/github.com/magma/magma/-/blob/lte/gateway/docker/mme/Dockerfile.rhel8?subtree=true#L310:21) a [2020 git commit](https://github.com/facebook/folly/commit/efe2962d54734d9e899f592dfaa68055e7752858) of Folly, Magma is generally using a mixture of 2018 git hash [Here](https://sourcegraph.com/github.com/magma/magma/-/blob/third_party/build/bin/folly_build.sh#L98) and newer versions for Dockerfile based testing...

We should really be static linking Folly as Facebook advises...

electronjoe:
Unable to debug RHEL issues - at this time.

PR #6979, opened 2021-05-18
[MME][DOCKER][RHEL8] Only using cmake3

rdefosse:

## Summary

Since last night, `cmake3` is not implicitly installed on `RHEL8` Ubi images.
I also forced all commands to be using that version (vs `cmake`)

## Test Plan

I did a manual build.

## Additional Information


themarwhal (approved):
lgtm

PR #6982, opened 2021-05-18
Remove startup sentry message from MME/SessionD

themarwhal:

## Summary
Sentry charges by the number of events, so sometimes it's not great to send an event every time the service starts up. This event was useful at first, but getting rid of them as it doesn't provide useful information.

## Test Plan
tested locally 

## Additional Information


uri200 (approved):

ulaskozat (approved):
LGTM

PR #6992, opened 2021-05-18
[circleCi] test to check proto determinators

uri200:

## Summary

This is a test to see if this pr causes issues due to missing protos on OIA build process

## Additional Information


uri200:
Fails as expected 

image

PR #6995, opened 2021-05-19
Fix ubuntu IPFIX RYU Patch

koolzz:

## Summary
Conntrack new match wasn't working properly so replacing with solution proposed by @pshelar by matching on ct_mart and setting it on conntrack commit

## Additional Information


pshelar (approved):
LGTM

tmdzk (approved):

PR #7008, opened 2021-05-19
[orc8r] Fix hard coded namespace for fluentd configmap

rmeleromira:
## Summary

When deploying 1.4, you get the following error for the fluentd-daemon.configmap.yaml template and the resource fails to create. 

```
error: the namespace from the provided object "kube-system" does not match the namespace "magma". You must pass '--namespace=kube-system' to perform this operation.
```

## Test Plan

I replaced the hard coded  `kube-system` with the correct namespace variable and re-ran the helm templating and the error went away and the resource wass created 

```
root@ctl01:~/helmchart# kubectl -n magma get cm orc8r-fluentd-es-configs
NAME                       DATA   AGE
orc8r-fluentd-es-configs   4      14m
```

### DIFF
--- a/orc8r/cloud/helm/orc8r/charts/logging/templates/fluentd-daemon.configmap.yaml
+++ b/orc8r/cloud/helm/orc8r/charts/logging/templates/fluentd-daemon.configmap.yaml
@@ -2,7 +2,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: {{ .Release.Name }}-fluentd-es-configs
-  namespace: kube-system
+  namespace: {{ .Release.Namespace }}
   labels:
     addonmanager.kubernetes.io/mode: Reconcile
 data:


rmeleromira:
Resubmitted with signed code commit

hcgatewood (changes requested):
As far as I can tell, we intentionally put this configmap in the `kube-system` namespace so that e.g. it can grab k8s-"internal" logs

Please provide additional information in the PR description on why you believe we should move this to the Magma-specific namespace 🙂

rmeleromira:
You're totally right that I'm not sure why kube-system was used here.

What I think is happening is that since the namespace here is hardcoded to kube-system, when you run a helm install command specifying the namespace, you get this error from helm, and it doesn't create the resource.

```
helm install -n magma orc8r etc etc
```

I found a related bugfix for helm covering this exact issue. It's possible my helm client at that time didn't have the fix. I don't have enough cycles to reproduce that environment atm.

https://github.com/jetstack/cert-manager/issues/1744
https://github.com/helm/helm/pull/6046

Probably a helm bug then? Since it should be able to create items in other namespaces. Close it out?

PR #7009, opened 2021-05-19
Fix S1 test default rule verificaton

koolzz:

## Summary

After I increased the ipfix flow priority the test fails, reducing the tbl0 priority0 flow count verification to 1

## Additional Information

### DIFF
--- a/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
+++ b/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
@@ -583,7 +583,7 @@ class S1ApUtil(object):
         print("Checking if all uplink/downlink flows were deleted")
         dpath = get_datapath()
         flows = get_flows(
-            dpath, {"table_id": self.SPGW_TABLE, "priority": 0}
+            dpath, {"table_id": self.SPGW_TABLE}
         )
         assert(
             len(flows) == 2), "There should only be 2 default table 0 flows"


lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -586,7 +586,7 @@ def verify_flow_rules_deletion(self):
             dpath, {"table_id": self.SPGW_TABLE, "priority": 0}
```
koolzz: this`verify_flow_rules_deletion` function seems odd, I'm assuming we're verifying that all rules were removed? Should we remove the priority match here completely then?

Or we could potentially query rules of different priorities (assert that no UE rules are left)
@ulaskozat

koolzz: Lets merge this PR to unblock CI though

ulaskozat: let's merge it and can discuss after unblocked for the best way forward.

ulaskozat (approved):
LGTM

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
ulaskozat: but yes, why do we check against priority 0? @pruthvihebbani

ulaskozat: @koolzz on table zero no rules but the default rule(s) should remain. removing "priority":0 in the match criteria should fix the logic.

koolzz: sure nuked the priority match lets see

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -583,7 +583,7 @@ def verify_flow_rules_deletion(self):
         print("Checking if all uplink/downlink flows were deleted")
         dpath = get_datapath()
         flows = get_flows(
-            dpath, {"table_id": self.SPGW_TABLE, "priority": 0}
+            dpath, {"table_id": self.SPGW_TABLE}
         )
         assert(
             len(flows) == 2), "There should only be 2 default table 0 flows"
```
ulaskozat, line 589: this should be 1 after ipfix changes, no?

koolzz: hm no, the ipfix flow is still there, just with higher pri

ulaskozat: ah, okay. for a moment I thought they were removed as well.

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
pruthvihebbani: I took this piece of code from test_attach_detach_rar_tcp_data.py TC assuming that flows with priority 0 are the default rules which are there even when the TCs are not running

ulaskozat: @pruthvihebbani raised and merged PR #7013.

themarwhal:
Removing backport as 1.5 doesn't have this function: `verify_flow_rules_deletion`

PR #7046, opened 2021-05-21
Modified sgw_s8 code to generate random control plane teid 

### LINKED ISSUE #6887: [MME] random sgw control plane TEID assignment
For Inbound roaming PGW will be using teid. as an anchor to find the subscriber. For that reason we need to find a way to provide unique TEIDs across AGW that provide service to the same roaming PLMN. Since we can't have a central service on orc8r to maintain that, we will use randomness to provide certain collision avoidance. 

Since TEIDs are used across the system  either for inbound and local roaming, the assignment for any subscriber (inbound or local) will have to be done with same random function to prevent collisions. That means s8_task and spgw_task will have to make sure they do not use same teids

The random generator should have a time based seed and every time we request for a new random number, we should check if that TEID is being used anywhere in the system. Meaning we need to check if sgw control plane teid is being used for any local, or inbound subscribers. If the teid is being used, we will ask random generator to provide another number and check it again

Note that for the stateless function, once we use random assignment we will not need to save the last sgw cp teid anymore.

rsarwad:
[lte][agw]: Modified sgw_s8 code to generate random control plane teid 
## Summary
Modified sgw_s8 code to generate random control plane teid that shall be used for both s11  and s8 interface to uniquely identify default bearer associated with PDN session,

The PR shall address the issue, https://github.com/magma/magma/issues/6887
## Test Plan
Executed s1ap test suite
Executed roaming test cases


uri200 (commented):
do we have to remove teid from stateless functionality?

Currently sgw_s8 task has the framework to support stateless but it doesn't have support to store in-memory contexts to Redis DB.
The reason why tunnel id is removed from sgw_state_t is because, previously this parameter was used as initial parameter for generating teid in fun, sgw_get_new_S11_tunnel_id(&state->tunnel_id);
But now teid generation doesn't depend on "state->tunnel_id", so removed it.

uri200 (approved):

lte/gateway/c/oai/tasks/sgw_s8/sgw_s8_handlers.c
```diff
@@ -1108,3 +1109,19 @@ void sgw_s8_handle_release_access_bearers_request(
...
 }
+
+/* Generates random control plane teid and is used for both s11 and
+ * s8 interface to uniquely identify sgw_s8_context for pdn session
+ */
+static teid_t sgw_s8_generate_new_cp_teid(void) {
+  OAILOG_FUNC_IN(LOG_SGW_S8);
+  sgw_eps_bearer_context_information_t* sgw_context_p = NULL;
+  teid_t teid                                         = INVALID_TEID;
+  // note srand with seed is init at main
```
ulaskozat: init -> initialized

lte/gateway/c/oai/tasks/sgw_s8/sgw_s8_handlers.c
```diff
@@ -1108,3 +1109,19 @@ void sgw_s8_handle_release_access_bearers_request(
   }
   OAILOG_FUNC_OUT(LOG_SGW_S8);
 }
+
+/* Generates random control plane teid and is used for both s11 and
+ * s8 interface to uniquely identify sgw_s8_context for pdn session
+ */
+static teid_t sgw_s8_generate_new_cp_teid(void) {
```
ulaskozat, line 1116: if this is generic id generation service, then let's make name generic e.g., `sgw_generate_new_cp_teid`.

ulaskozat (changes requested):
you need to adopt the same random assignment for sgw task.

lte/gateway/c/oai/tasks/sgw_s8/sgw_s8_handlers.c
```diff
@@ -1108,3 +1109,19 @@ void sgw_s8_handle_release_access_bearers_request(
...
+  OAILOG_FUNC_IN(LOG_SGW_S8);
+  sgw_eps_bearer_context_information_t* sgw_context_p = NULL;
+  teid_t teid                                         = INVALID_TEID;
+  // note srand with seed is init at main
+  do {
+    teid          = (teid_t) rand();
+    sgw_context_p = sgw_get_sgw_eps_bearer_context(teid);
+  } while (sgw_context_p);
+
```
ulaskozat, line 1125: does this ensure that TEID is unique across S8 and spgw tasks?

rsarwad: No. This logic creates unique control plane tied in sgw_s8 task. The same teid is assigned to both s11 and s8 interface as it maps to single PDN session. 
But we don't need to have unique tied across s8 and spgw tasks. Because MME may be connected with multiple SGWs but we don't  need have unique control plane teids across SGWs.
In mme, sgw_s11_cp_teid is updated per PDN, so I think it should be fine
Me and @uri200  had discussion and Oriol modified inbound_roaming_02.ntl script to run both roaming and non-roaming subscribers and I generated mme binary with same teid assigned to both S8 task and spgw task. 
It worked fine.

A unique sgw_s11_teid is required in SGW to identify procedure belongs to specific PDN and later to identify specific bearer, we use the ebi.
Is there any other requirement?

lte/gateway/c/oai/tasks/sgw_s8/sgw_s8_handlers.c
rsarwad: This function generates control plane teid for sgw_s8 task only. So kept the name as sgw_s8_generate_new_cp_teid

lte/gateway/c/oai/tasks/sgw_s8/sgw_s8_handlers.c
ulaskozat: Okay good points.

ulaskozat (approved):
LGTM. We should revisit the naming conventions.

PR #7067, opened 2021-05-22
[feg] S8_proxy add C agw Teid to Create Bearer Request

### LINKED ISSUE #7104: [orc8r] modify FEG rely api to support CreateBearerRequest from PGW

uri200:

## Summary

Added missing Control Plane AGW TEID to the Create Bearer Request

## Test Plan

make precommit at feg
./build.py -g at orc8

## Additional Information


themarwhal (approved):

feg/protos/s8_proxy.proto
```diff
@@ -176,10 +176,11 @@ message DeleteSessionResponsePgw{
 }
 
 message CreateBearerRequestPgw {
-    uint32 linked_bearer_id = 1;
-    ProtocolConfigurationOptions protocol_configuration_options = 2;
-    BearerContext bearer_context = 3;
-    bytes indication_flag = 4;
+    uint32 c_agw_teid = 1;
```
themarwhal, line 179: Why add it as field 1? Shouldn't we just add as field 5? Or is this proto not used anywhere in prod.

uri200: This is not being used yet 
Otherwise i would add it as 5

Btw, i wanted it to be the first because that is the only identifier to know what subscriber the request is for.

PR #7107, opened 2021-05-24
[orc8r] Add certificate lifespan metric and alert

### LINKED ISSUE #6969: Add certificate expiry alerts in orc8r

andreilee:
## Summary

Adds an Orc8r metric, `cert_life_hours`, which counts the remaining lifespan of a certificate used on the orchestrator. Specifically, this checks the certificate's `NotAfter` property, and compares it to the current time.

This remaining lifespan is published for the following certificate files on the Orc8r:
- `rootCA.pem`
- `admin_operator.pem`
- `controller.crt`
- `certifier.pem`
- `fluentd.pem`

The responsibility of getting and publishing this metric is placed in the `certifier` service, which currently has some responsibilities around cert management. The `analytics` service will periodically query the `certifier` service for the cert lifespan metric.

## Test Plan

Tested manually using a docker setup for Orc8r. With additional logs whenever the `analytics` service queries `certifier`, the following logs were reported:

```
controller_1     | certifier stderr | I0524 01:00:00.363915     113 lifespan.go:57] Calculate Cert Lifespan
controller_1     | certifier stderr | I0524 01:00:00.366429     113 lifespan.go:98] Calculated metric for rootCA.pem
controller_1     | certifier stderr | I0524 01:00:00.366465     113 lifespan.go:99] value:2.562047e+06 metricName:"cert_life_hours" labels:<key:"certName" value:"rootCA.pem" > labels:<key:"class" value:"reliability" >
controller_1     | certifier stderr | I0524 01:00:00.366531     113 lifespan.go:75] value:2.562047e+06 metricName:"cert_life_hours" labels:<key:"certName" value:"rootCA.pem" > labels:<key:"class" value:"reliability" >
controller_1     | certifier stderr | I0524 01:00:00.368093     113 lifespan.go:98] Calculated metric for admin_operator.pem
controller_1     | certifier stderr | I0524 01:00:00.368123     113 lifespan.go:99] value:1774 metricName:"cert_life_hours" labels:<key:"certName" value:"admin_operator.pem" > labels:<key:"class" value:"reliability" >
controller_1     | certifier stderr | I0524 01:00:00.368148     113 lifespan.go:75] value:1774 metricName:"cert_life_hours" labels:<key:"certName" value:"admin_operator.pem" > labels:<key:"class" value:"reliability" >
controller_1     | certifier stderr | I0524 01:00:00.368892     113 lifespan.go:98] Calculated metric for controller.crt
controller_1     | certifier stderr | I0524 01:00:00.368901     113 lifespan.go:99] value:866495 metricName:"cert_life_hours" labels:<key:"certName" value:"controller.crt" > labels:<key:"class" value:"reliability" >
controller_1     | certifier stderr | I0524 01:00:00.368921     113 lifespan.go:75] value:866495 metricName:"cert_life_hours" labels:<key:"certName" value:"controller.crt" > labels:<key:"class" value:"reliability" >
controller_1     | certifier stderr | I0524 01:00:00.370022     113 lifespan.go:98] Calculated metric for certifier.pem
controller_1     | certifier stderr | I0524 01:00:00.370055     113 lifespan.go:99] value:2.562047e+06 metricName:"cert_life_hours" labels:<key:"certName" value:"certifier.pem" > labels:<key:"class" value:"reliability" >
```

Synced pre-defined alerts with development setup:
Screen Shot 2021-05-24 at 10 59 30 PM

## Additional Information

### DIFF
--- a/nms/app/packages/magmalte/alerts/lteAlerts.js
+++ b/nms/app/packages/magmalte/alerts/lteAlerts.js
@@ -20,6 +20,14 @@ export default function getLteAlerts(
   networkID: string,
 ): {[string]: prom_alert_config} {
   return {
+    'Certificate Expiring Soon': {
+      alert: 'Certificate Expiring Soon',
+      expr: `cert_expires_in_hours > 720`,
+      labels: {severity: 'major'},
+      annotations: {
+        description: `Alerts when certificate necessary for Orc8r function is expiring soon`,
+      },
+    },
     'Service Restart Alert': {
       alert: 'Service Restart Alert',
       expr: `increase(service_restart_status{networkID=~"${networkID}"}[5m]) > 0`,
--- a/orc8r/cloud/configs/certifier.yml
+++ b/orc8r/cloud/configs/certifier.yml
@@ -9,3 +9,27 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# All orchestratorCerts are assumed to live in this directory
+certsDirectory: "/var/opt/magma/certs/"
+
+# These certs will be alerted on when near expiry, so only include
+# certs that are necessary for Orc8r function
+orchestratorCerts:
+  - "rootCA.pem"
+  - "admin_operator.pem"
+  - "controller.crt"
+  - "certifier.pem"
+  - "fluentd.pem"
+
+analytics:
+  # Metrics in this certifier configuration should strictly be generic in
+  # nature independent of the type of deployment. It is to be also free of any
+  # user level metrics
+  metrics:
+    # cert metrics
+    cert_expires_in_hours:
+      register: true
+      export: false
+      labels:
+        class: reliability
--- a/orc8r/cloud/go/services/analytics/analyzer.go
+++ b/orc8r/cloud/go/services/analytics/analyzer.go
@@ -77,7 +77,7 @@ func (a *PrometheusAnalyzer) Schedule() error {
 
 // Analyze methods runs through collectors and exports their metrics
 func (a *PrometheusAnalyzer) Analyze() {
-	glog.Info("Running  Analyze")
+	glog.V(2).Info("Running Analyze")
 	collectorClients, err := getRemoteCollectors()
 	if err != nil {
 		glog.Infof("err %v failed to get remote collectors", err)
--- a/orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
+++ b/orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2020 The Magma Authors.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package calculations
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"io/ioutil"
+	"log"
+	"math"
+	"strings"
+	"time"
+
+	"magma/orc8r/cloud/go/services/analytics/calculations"
+	"magma/orc8r/cloud/go/services/analytics/protos"
+	"magma/orc8r/cloud/go/services/analytics/query_api"
+	"magma/orc8r/lib/go/metrics"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type CertLifespanCalculation struct {
+	CertsDirectory string
+	Certs          []string
+	calculations.BaseCalculation
+}
+
+func (x *CertLifespanCalculation) Calculate(prometheusClient query_api.PrometheusAPI) ([]*protos.CalculationResult, error) {
+	glog.V(2).Infof("Calculating %s", metrics.CertExpiresInHoursMetric)
+	var results []*protos.CalculationResult
+
+	metricConfig, ok := x.AnalyticsConfig.Metrics[metrics.CertExpiresInHoursMetric]
+	if !ok {
+		glog.Errorf("%s metric not found in metric config", metrics.CertExpiresInHoursMetric)
+		return results, nil
+	}
+
+	for _, certName := range x.Certs {
+		result, err := calculateCertLifespanHours(x.CertsDirectory, certName, metricConfig.Labels)
+		if err != nil {
+			glog.Errorf("Could not get lifespan for cert %s: %+v", certName, err)
+			continue
+		}
+		results = append(results, result)
+	}
+
+	return results, nil
+}
+
+func calculateCertLifespanHours(certsDirectory string, certName string, metricConfigLabels map[string]string) (*protos.CalculationResult, error) {
+	dat, err := getCert(certsDirectory + certName)
+	if err != nil {
+		return nil, err
+	}
+	cert, err := x509.ParseCertificate(dat)
+	if err != nil {
+		return nil, err
+	}
+	// Hours remaining
+	hoursLeft := math.Floor(time.Until(cert.NotAfter).Hours())
+	labels := prometheus.Labels{
+		metrics.CertNameLabel: certName,
+	}
+	labels = calculations.CombineLabels(labels, metricConfigLabels)
+	result := calculations.NewResult(hoursLeft, metrics.CertExpiresInHoursMetric, labels)
+	glog.V(2).Infof("Calculated metric %s for %s: %f", metrics.CertExpiresInHoursMetric, certName, hoursLeft)
+	return result, nil
+}
+
+func getCert(certPath string) ([]byte, error) {
+	dat, err := ioutil.ReadFile(certPath)
+	if err != nil {
+		return nil, err
+	}
+	if strings.HasSuffix(certPath, ".pem") || strings.HasSuffix(certPath, ".crt") {
+		block, _ := pem.Decode(dat)
+		if block == nil || block.Type != "PUBLIC KEY" {
+			log.Fatal("failed to decode PEM block containing public key")
+		}
+		return block.Bytes, nil
+	}
+	return dat, nil
+}
--- a/orc8r/cloud/go/services/certifier/certifier/main.go
+++ b/orc8r/cloud/go/services/certifier/certifier/main.go
@@ -21,7 +21,10 @@ import (
 	"magma/orc8r/cloud/go/blobstore"
 	"magma/orc8r/cloud/go/orc8r"
 	"magma/orc8r/cloud/go/service"
+	"magma/orc8r/cloud/go/services/analytics"
+	analytics_protos "magma/orc8r/cloud/go/services/analytics/protos"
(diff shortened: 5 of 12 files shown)


orc8r/cloud/configs/certifier.yml
```diff
@@ -9,3 +9,15 @@
...
 # limitations under the License.
+
+analytics:
+  # Metrics in this Orchestrator configuration should strictly be generic in
+  # nature independent of the type of deployment. It is to be also free of any
+  # user level metrics
+  metrics:
+    # cert metrics
+    cert_life_hours:
+      register: false
```
karthiksubraveti: You want to set the register to true and export to false. Register causes the metric to be registered with prometheus. Export causes the metric to be exported to external metric collectors

orc8r/cloud/configs/certifier.yml
```diff
@@ -9,3 +9,15 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+analytics:
+  # Metrics in this Orchestrator configuration should strictly be generic in
+  # nature independent of the type of deployment. It is to be also free of any
+  # user level metrics
+  metrics:
+    # cert metrics
+    cert_life_hours:
```
karthiksubraveti: Will prefer the naming as follows -> cert_life_hours -> cert_expires_in_seconds or cert_expires_in_hours

orc8r/cloud/go/services/analytics/analyzer.go
```diff
@@ -77,7 +77,7 @@ func (a *PrometheusAnalyzer) Schedule() error {
 
 // Analyze methods runs through collectors and exports their metrics
 func (a *PrometheusAnalyzer) Analyze() {
-	glog.Info("Running  Analyze")
+	glog.Info("Running Analyze")
```
karthiksubraveti: I added it, you can reduce the verbosity here.

orc8r/cloud/go/services/certifier/certifier/main.go
```diff
@@ -83,6 +87,20 @@ func main() {
 		caMap[protos.CertType_VPN] = &servicers.CAInfo{Cert: vpnCert, PrivKey: vpnPrivKey}
 	}
 
+	var serviceConfig certifier.Config
+	_, _, err = config.GetStructuredServiceConfig(orc8r.ModuleName, certifier.ServiceName, &serviceConfig)
+	if err != nil {
+		glog.Errorf("err %v failed parsing the config file: skipping CollectorServicer creation ", err)
```
karthiksubraveti: Currently let's make this glog fatal and move the else out of the conditional. When it is desirable to make this optional, we can make this change accordingly

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type CertLifespanCalculation struct {
+	calculations.BaseCalculation
+}
+
+const (
+	DefaultCertsDirectory = "/var/opt/magma/certs/"
```
karthiksubraveti: let's move all these to config file if possible.

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
```diff
@@ -0,0 +1,105 @@
...
+	labels := prometheus.Labels{
+		metrics.CertNameLabel: certName,
+	}
+	labels = calculations.CombineLabels(labels, metricConfigLabels)
+	result := calculations.NewResult(hoursLeft, metrics.CertLifeHoursMetric, labels)
+	glog.V(2).Infof("Calculated metric %s for %s: %f", metrics.CertLifeHoursMetric, certName, hoursLeft)
+	return result, nil
+}
+
+func getCertBytes(certName string) ([]byte, error) {
```
karthiksubraveti: minor nit: Will prefer calling this getCertBytes -> getCert

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
```diff
@@ -0,0 +1,105 @@
...
+	dat, err := ioutil.ReadFile(DefaultCertsDirectory + certName)
+	if err != nil {
+		return nil, err
+	}
+	if strings.HasSuffix(certName, ".pem") || strings.HasSuffix(certName, ".crt") {
```
karthiksubraveti: If file name isn't pem or crt should we fail here ?

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
```diff
@@ -0,0 +1,105 @@
...
+		block, rest := pem.Decode(dat)
+		if len(rest) > 0 {
```
karthiksubraveti: Should we check for block being nil here? Also when rest is > 0 why is this an error case? Just curious

karthiksubraveti (commented):
Left a few comments. Thanks for the changes.

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
```diff
@@ -0,0 +1,105 @@
...
+
+	var results []*protos.CalculationResult
+
+	metricConfig, ok := x.AnalyticsConfig.Metrics[metrics.CertLifeHoursMetric]
+	if !ok {
+		glog.Errorf("%s metric not found in metric config", metrics.CertLifeHoursMetric)
+		return results, nil
+	}
+
+	var certs = [...]string{RootCert, AdminCert, ControllerCert, CertifierCert, FluentdCert}
```
hcgatewood: No need for an array literal here, just use a slice
```suggestion
	var certs = []string{RootCert, AdminCert, ControllerCert, CertifierCert, FluentdCert}
```

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
```diff
@@ -0,0 +1,105 @@
...
+
+	for _, certName := range certs {
+		result, err := calculateCertLifespanHours(certName, metricConfig.Labels)
+		if err != nil {
+			glog.Errorf("Could not get lifespan for cert %s: %s", certName, err)
```
hcgatewood: Report errors as `%+v`
```suggestion
			glog.Errorf("Could not get lifespan for cert %s: %+v", certName, err)
```

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
```diff
@@ -0,0 +1,105 @@
...
+func calculateCertLifespanHours(certName string, metricConfigLabels map[string]string) (*protos.CalculationResult, error) {
+	dat, err := getCertBytes(certName)
+	if err != nil {
+		return nil, err
+	}
+	cert, err := x509.ParseCertificate(dat)
+	if err != nil {
+		return nil, err
+	}
+	// Hours left
```
hcgatewood:
```suggestion
	// Hours remaining
```

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
```diff
@@ -0,0 +1,105 @@
...
+	hoursLeft := math.Floor(cert.NotAfter.Sub(time.Now()).Hours())
```
hcgatewood: Nit: maybe just convert this to int for simplicity

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
hcgatewood: Either way, use an early return rather than nested if block

hcgatewood (dismissed):
Looks good, thanks for adding. Couple small changes on top of Karthik's

orc8r/cloud/go/services/certifier/certifier/main.go
andreilee: The thought here was that this is only used for metrics, so if this fails for some reason, the entire `certifier` service shouldn't fail to run

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
andreilee: `rest` is whatever can't get decoded properly by the `pem` package. So if it's len>0 that's equivalent to an error, IMO.

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
andreilee: pem and crt files need to get decoded, but other files can be directly parsed as a certificate through the x509 package

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
andreilee: sure

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
andreilee: It needs to be cast back to `float64` after, so I don't think that's a win in this case

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
andreilee: sure

orc8r/cloud/go/services/certifier/certifier/main.go
karthiksubraveti: Given that we are collecting these metrics unconditionally and we always expect this file to be around, I think adding this assertion should help us catch any issues early on. If we return with an error, we might fail silently and never realize that the alert metrics never got registered. I'm favor of failing hard than failing silently.

andreilee: We might need to discuss this as a team, since we're removing assertFatal in MME, which is going in the opposite direction

andreilee: IMO we need the equivalent of a "File an issue in Github if this ever happens"

karthiksubraveti: I think MME is fundamentally different problem than orc8r. We can certainly discuss this. Please initiate the conversation.

nms/app/packages/magmalte/alerts/lteAlerts.js
```diff
@@ -20,6 +20,13 @@ export default function getLteAlerts(
   networkID: string,
 ): {[string]: prom_alert_config} {
   return {
+    'Certificate Expiring Soon': {
+      alert: 'Certificate Expiring Soon',
+      expr: `cert_expires_in_hours > 720`,
+      annotations: {
```
karthiksubraveti, line 27: This is missing severity.

orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go
karthiksubraveti: In this documentation and following example - https://golang.org/pkg/encoding/pem/#Decode.
 it is mentioned the "Decode will find the next PEM formatted block (certificate, private key etc) in the input. It returns that block and the remainder of the input. If no PEM data is found, p is nil and the whole of the input is returned in rest". I wasn't able to find where rest will be returned if it can't be decoded properly. Let me know if i have this mixed up. I feel its better to use the following check over the checking for rest's length.
```
	if block == nil || block.Type != "PUBLIC KEY" {
		log.Fatal("failed to decode PEM block containing public key")
	}
```

karthiksubraveti (approved):
Changes LGTM

PR #7125, opened 2021-05-25
feat(agw): Added stateless support feature for ueip_imsi map

### LINKED ISSUE #6333: [MME] Create  Redis table to store  ip, IMSI and type of subscriber
Right now MME needs to call `get_subscriber_id_from_ipv4` that calls mobilityd http://github.com/magma/magma/blob/master/lte/gateway/c/oai/tasks/sgw/sgw_paging.c#L36-L36 
every time there is a paging request which creates a new GRPC call just to extract the IMSI

Also for inbound roaming users, this is not working because we don't store the IP, so we can't find them.

To be able to find those users we need a Redis table that stores the mapping of imsi, ipv4, ipv6 and type of subscriber (s11 or s8).  The key will be IP (either ipv4 or ipv6)

This table will be used to find out the IMSI of subscribers when paging is needed. 

Note that this table will may be updated (created entry and delete entry) by two different tasks, so maybe locks are required to avoid write access at the same time. To avoid having those locks lets try to implement the functions that are going to add/delete entries on that table in mme task. Reading from redis is not a problem and can be done by sp task and sp_s8 task

rsarwad:
feat(agw): Added stateless support for ueip_imsi map

## Summary
1. Added stateless support for ueip_imsi map
2. Modified state_cli.py script to parse mme_ueip_imsi_map
3. Modified config_stateless_agw.py script to clear contents of mme_ueip_imsi_map
4. Added new test case, test_paging_after_mme_restart.py to validate the resumption of mme_ueip_imsi_map after mme has restarted and added to sanity test suite
5. Modified s1ap_utils.py to get the remaining entries after test case is completed.

## Test Plan


lte/gateway/c/oai/tasks/mme_app/mme_app_state_manager.cpp
```diff
@@ -222,6 +228,7 @@ void MmeNasStateManager::free_state() {
     return;
   }
   clear_mme_nas_hashtables();
+  clear_mme_ueip_imsi_map();
```
uri200: when is `free_state` executed on mme? I mean, in what situation de we need to clear all those tables?

lte/gateway/c/oai/tasks/mme_app/mme_app_ip_imsi.cpp
```diff
@@ -15,24 +15,26 @@ limitations under the License.
...
-Ipv4Map ipv4map;
+#include "mme_app_ueip_imsi_map.h"
+#include "mme_app_state_manager.h"
 
-void initialize_ipv4_map() {
-  OAILOG_FUNC_IN(LOG_MME_APP);
-  ipv4map = Ipv4Map{};
-  OAILOG_FUNC_OUT(LOG_MME_APP);
-}
+using namespace std;
```
ulaskozat: don't use such pattern. explicitly use `std::` in scoping.

lte/gateway/c/oai/tasks/mme_app/mme_app_ip_imsi.cpp
```diff
@@ -46,18 +48,26 @@ void mme_app_log_ipv4_imsi_map() {
...
+  }
   char ipv4[INET_ADDRSTRLEN] = {0};
   inet_ntop(AF_INET, (void*) &ipv4_addr, ipv4, INET_ADDRSTRLEN);
-  auto itr = ipv4map.find(ipv4);
-  if (itr == ipv4map.end()) {
-    vector<uint64_t> vec = {imsi64};
-    ipv4map[ipv4]        = vec;
+  auto itr_map = ueip_imsi_map->find(ipv4);
+  if (itr_map == ueip_imsi_map->end()) {
+    vector<uint64_t> vec   = {imsi64};
```
ulaskozat: don't use like this, instead explicitly call std::vector.  Also have `#include <vector>`.

lte/gateway/c/oai/tasks/mme_app/mme_app_ip_imsi.cpp
```diff
@@ -15,24 +15,26 @@ limitations under the License.
...
-  ipv4map = Ipv4Map{};
-  OAILOG_FUNC_OUT(LOG_MME_APP);
-}
+using namespace std;
 
 void mme_app_log_ipv4_imsi_map() {
   OAILOG_FUNC_IN(LOG_MME_APP);
-  for (auto itr = ipv4map.begin(); itr != ipv4map.end(); ++itr) {
-    for (auto it_vec = itr->second.begin(); it_vec != itr->second.end();
+  UeIpImsiMap* ueip_imsi_map =
```
ulaskozat: use `UeIpImsiMap&` instead of pointer and update the method signature accordingly.

lte/gateway/c/oai/tasks/mme_app/mme_app_ip_imsi.cpp
```diff
@@ -46,18 +48,26 @@ void mme_app_log_ipv4_imsi_map() {
...
+  if (itr_map == ueip_imsi_map->end()) {
+    vector<uint64_t> vec   = {imsi64};
+    (*ueip_imsi_map)[ipv4] = vec;
     OAILOG_DEBUG_UE(LOG_MME_APP, imsi64, "Inserting ue_ip:%x \n", ipv4_addr);
   } else {
     OAILOG_DEBUG_UE(
         LOG_MME_APP, imsi64, "Inserting imsi for existing ue_ip:%x \n",
         ipv4_addr);
-    ipv4map[ipv4].push_back(imsi64);
+    (*ueip_imsi_map)[ipv4].push_back(imsi64);
```
ulaskozat: see my comment above about changing the type from pointer to reference.

lte/gateway/c/oai/tasks/mme_app/mme_app_ip_imsi.cpp
```diff
@@ -68,19 +78,28 @@ int mme_app_insert_ue_ipv4_addr(uint32_t ipv4_addr, imsi64_t imsi64) {
  */
 int mme_app_get_imsi_from_ipv4(uint32_t ipv4_addr, imsi64_t** imsi_list) {
   OAILOG_FUNC_IN(LOG_MME_APP);
-  int num_imsis              = 0;
+  UeIpImsiMap* ueip_imsi_map =
```
ulaskozat: same comment as before.

lte/gateway/c/oai/tasks/mme_app/mme_app_ip_imsi.cpp
```diff
@@ -91,26 +110,37 @@ int mme_app_get_imsi_from_ipv4(uint32_t ipv4_addr, imsi64_t** imsi_list) {
 
 void mme_app_remove_ue_ipv4_addr(uint32_t ipv4_addr, imsi64_t imsi64) {
   OAILOG_FUNC_IN(LOG_MME_APP);
+  UeIpImsiMap* ueip_imsi_map =
```
ulaskozat: same comment.

lte/gateway/c/oai/tasks/mme_app/mme_app_main.c
```diff
@@ -460,6 +460,7 @@ static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
 
   put_mme_nas_state();
   put_mme_ue_state(mme_app_desc_p, imsi64);
+  put_mme_ueip_imsi_map();
```
ulaskozat: you need to rebase as these parts have changed.

lte/gateway/c/oai/tasks/mme_app/mme_app_state_converter.cpp
```diff
@@ -819,5 +820,34 @@ void MmeNasStateConverter::proto_to_ue(
     const oai::UeContext& ue_ctxt_proto, ue_mm_context_t* ue_ctxt) {
   proto_to_ue_mm_context(ue_ctxt_proto, ue_ctxt);
 }
+
+void MmeNasStateConverter::mme_app_ueip_imsi_map_to_proto(
+    const UeIpImsiMap* ueip_map, oai::MmeUeIpImsiMap* ueip_proto) {
+  auto proto_map = ueip_proto->mutable_mme_ueip_imsi_map();
+  for (auto itr = ueip_map->begin(); itr != ueip_map->end(); ++itr) {
```
ulaskozat: use `auto &` everywhere it makes sense unless you need a copy.

lte/gateway/c/oai/tasks/mme_app/mme_app_state_manager.cpp
```diff
@@ -259,5 +266,54 @@ int MmeNasStateManager::read_ue_state_from_db() {
...
+  }
+
+  oai::MmeUeIpImsiMap ueip_proto = oai::MmeUeIpImsiMap();
+  MmeNasStateConverter::mme_app_ueip_imsi_map_to_proto(
+      ueip_imsi_map, &ueip_proto);
+  std::string proto_msg;
+  redis_client->serialize(ueip_proto, proto_msg);
+  std::size_t new_hash = std::hash<std::string>{}(proto_msg);
+
+  // s1ap_imsi_map is not state service synced, so version will not be updated
```
ulaskozat: fix the comment

ulaskozat:
Do we really want to use protobuf here for serialization while we plan to get rid of it completely? can't we use the same marshal/unmarshal pattern as in sessiond?  @ardzoht @themarwhal what do you think?

ulaskozat (changes requested):
Left some comments. I think we should not replicate the pattern we used for other MME states here.

lte/gateway/c/oai/tasks/mme_app/mme_app_main.c
ulaskozat: we should not use this pattern at all. let's directly write to store when actually it is written.

lte/gateway/c/oai/tasks/mme_app/mme_app_state_manager.cpp
```diff
@@ -259,5 +266,54 @@ int MmeNasStateManager::read_ue_state_from_db() {
...
+  if (new_hash != this->ueip_imsi_map_hash) {
+    redis_client->write_proto_str(MME_UEIP_IMSI_MAP_NAME, proto_msg, 0);
+    this->ueip_imsi_map_hash = new_hash;
+  }
```
ulaskozat: why serialize the whole map? we can explicitly use the IP address as part of the key and retrieve/write the value (i.e., vector of IMSIs).

ardzoht:
> Do we really want to use protobuf here for serialization while we plan to get rid of it completely? can't we use the same marshal/unmarshal pattern as in sessiond? @ardzoht @themarwhal what do you think?

We need protobuf (or at least a serialization protocol - flatbuffers e.g.) as we need something to parse the C structs. Sessiond uses folly dynamic objects (@uri200 @themarwhal can keep me truthful here) which makes it easier to ser/deserialize to JSON, I think that is out of scope of this change.  We should however not write the imsi map on every message, let's only serialize and write whenever there is a change to the map.

lte/gateway/c/oai/tasks/mme_app/mme_app_state_converter.cpp
ardzoht: same here, let's use for : each loop traversal here, and as @ulaskozat suggested, use `const auto&` as much as possible

lte/gateway/c/oai/tasks/mme_app/mme_app_ip_imsi.cpp
```diff
@@ -15,24 +15,26 @@ limitations under the License.
...
   OAILOG_FUNC_IN(LOG_MME_APP);
-  for (auto itr = ipv4map.begin(); itr != ipv4map.end(); ++itr) {
-    for (auto it_vec = itr->second.begin(); it_vec != itr->second.end();
+  UeIpImsiMap* ueip_imsi_map =
+      magma::lte::MmeNasStateManager::getInstance().get_mme_ueip_imsi_map();
+  if (!ueip_imsi_map) {
+    OAILOG_ERROR(LOG_MME_APP, "ueip_imsi_map is NULL \n");
+    OAILOG_FUNC_OUT(LOG_MME_APP);
+  }
+  for (auto itr_map = ueip_imsi_map->begin(); itr_map != ueip_imsi_map->end();
```
ardzoht: I'd encourage for : each range based loop here, you can access each entry of the map and it will contain `entry.first` and `entry.second` to access each key and value of the map

lte/gateway/c/oai/tasks/mme_app/mme_app_ueip_imsi_map.h
```diff
@@ -0,0 +1,22 @@
...
+
+#pragma once
+
+#include <bits/stdc++.h>
+#include <unordered_map>
+#include <iostream>
+#include <string>
+#include <vector>
+
+typedef std::unordered_map<std::string, std::vector<uint64_t>> UeIpImsiMap;
```
ardzoht: Add comment here of what this represents, what is the key and what value it maps to

lte/gateway/c/oai/tasks/mme_app/mme_app_ip_imsi.h
```diff
@@ -20,7 +20,6 @@ extern "C" {
 #include "log.h"
 #include "common_defs.h"
 #include "common_types.h"
-void initialize_ipv4_map(void);
 int mme_app_insert_ue_ipv4_addr(uint32_t ipv4_addr, imsi64_t imsi64);
```
ardzoht: please add documentation to each of these public functions

ardzoht (changes requested):

uri200:
> Do we really want to use protobuf here for serialization while we plan to get rid of it completely? can't we use the same marshal/unmarshal pattern as in sessiond? @ardzoht @themarwhal what do you think?

I think I am missing something here: In sessiond we use folly + protos to serialize: 

We build a marshaled key:value structure which we serialize into JSON using folly (see [here](http://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/RedisStoreClient.cpp#L199-L199))

Every key on this marshaled structure is hardcoded (see [here](http://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/StoredState.cpp#L496-L496)). Values are basic types (or folly arrays). But there are some of those values that are the result of a proto serialization (see [here](http://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/StoredState.cpp#L513-L513)).

The reason to use proto serialization is that it abstracts changes on the proto definition from the serialization. In other words, if you change the proto, serialization is changed automatically too. Having to do manual serialization on every single change on session definition was not scalable in sessiond. 

In fact,in my opinion, this `marshaled` object is nothing else than some kind of proto, but without all the advantages proto provides.


Going back to MME, do we want to introduce folly to MME to avoid protos?

If we don't want to use neither protos nor folly, the only way to serialize/deserialize the vector of IPs is manually creating a coma separated string (and its counterpart parsing function). This may work for this small hashmap, but we shouln't be doing this for bigger objects.   

So we will need to use some kind of structure to serialize, and protos are well understood, supported and scale well.

rsarwad:
> > Do we really want to use protobuf here for serialization while we plan to get rid of it completely? can't we use the same marshal/unmarshal pattern as in sessiond? @ardzoht @themarwhal what do you think?
> 
> We need protobuf (or at least a serialization protocol - flatbuffers e.g.) as we need something to parse the C structs. Sessiond uses folly dynamic objects (@uri200 @themarwhal can keep me truthful here) which makes it easier to ser/deserialize to JSON, I think that is out of scope of this change. We should however not write the imsi map on every message, let's only serialize and write whenever there is a change to the map.

Hi @ardzoht Content of map is written to DB only if there is change in the map. This is taken care by condition, https://github.com/magma/magma/pull/7125/files#diff-f57ac4be9f583acbf410f4697861afa94f12d4137b9f8d8167e86b3b1338eb12R301

ardzoht:
> > > Do we really want to use protobuf here for serialization while we plan to get rid of it completely? can't we use the same marshal/unmarshal pattern as in sessiond? @ardzoht @themarwhal what do you think?
> > 
> > 
> > We need protobuf (or at least a serialization protocol - flatbuffers e.g.) as we need something to parse the C structs. Sessiond uses folly dynamic objects (@uri200 @themarwhal can keep me truthful here) which makes it easier to ser/deserialize to JSON, I think that is out of scope of this change. We should however not write the imsi map on every message, let's only serialize and write whenever there is a change to the map.
> 
> Hi @ardzoht Content of map is written to DB only if there is change in the map. This is taken care by condition, https://github.com/magma/magma/pull/7125/files#diff-f57ac4be9f583acbf410f4697861afa94f12d4137b9f8d8167e86b3b1338eb12R301

@rsarwad This condition will save the write operation, but it will still require serialization to calculate hash value of the map content, thus the need to reduce the number of calls to `put_ueip_imsi_map`

rsarwad:
> > > > Do we really want to use protobuf here for serialization while we plan to get rid of it completely? can't we use the same marshal/unmarshal pattern as in sessiond? @ardzoht @themarwhal what do you think?
> > > 
> > > 
> > > We need protobuf (or at least a serialization protocol - flatbuffers e.g.) as we need something to parse the C structs. Sessiond uses folly dynamic objects (@uri200 @themarwhal can keep me truthful here) which makes it easier to ser/deserialize to JSON, I think that is out of scope of this change. We should however not write the imsi map on every message, let's only serialize and write whenever there is a change to the map.
> > 
> > 
> > Hi @ardzoht Content of map is written to DB only if there is change in the map. This is taken care by condition, https://github.com/magma/magma/pull/7125/files#diff-f57ac4be9f583acbf410f4697861afa94f12d4137b9f8d8167e86b3b1338eb12R301
> 
> @rsarwad This condition will save the write operation, but it will still require serialization to calculate hash value of the map content, thus the need to reduce the number of calls to `put_ueip_imsi_map`

Ya right. Let me call put_ueip_imsi_map() on any modification to ueip_imsi map.

lte/gateway/c/oai/tasks/mme_app/mme_app_ip_imsi.cpp
rsarwad: Sure I shall remove "using namespace std" and use "std::vector"
#include <vector> is included in file "mme_app_ueip_imsi_map.h", which is included here. So explicitly we don't need to include <vector> in this file.

lte/gateway/c/oai/tasks/mme_app/mme_app_state_manager.cpp
rsarwad: On mme exit or on mme restart,  mme clears all the in-memory contexts. free_state() is invoked to clear in memory contexts

lte/gateway/c/oai/tasks/mme_app/mme_app_state_converter.cpp
rsarwad: Thanks @ulaskozat and @ardzoht for suggestions and happy to learn new concepts

lte/gateway/c/oai/tasks/mme_app/mme_app_state_manager.cpp
rsarwad: Version is not updated, so added this comment. Could you please elaborate on required comment

lte/gateway/c/oai/tasks/mme_app/mme_app_main.c
rsarwad: Corrected code such that when in-memory map is modified, it's written DB at the same time

lte/gateway/c/oai/tasks/mme_app/mme_app_ip_imsi.h
rsarwad: Added description in file, mme_app_ip_imsi.cpp
https://github.com/magma/magma/pull/7125/files#diff-1b5fb3ec21f41be0470796c1df743a332a0af3dd3f77d85a324a3303efca6acbR44
Do we need provide description in both files, mme_app_ip_imsi.h and mme_app_ip_imsi.cpp?

ardzoht: missed the docs on the other one, thanks!

lte/gateway/c/oai/tasks/mme_app/mme_app_state_manager.cpp
```diff
@@ -259,5 +264,39 @@ int MmeNasStateManager::read_ue_state_from_db() {
...
+    return;
+  }
+
+  oai::MmeUeIpImsiMap ueip_proto = oai::MmeUeIpImsiMap();
+  MmeNasStateConverter::mme_app_ueip_imsi_map_to_proto(
+      ueip_imsi_map, &ueip_proto);
+  std::string proto_msg;
+  redis_client->serialize(ueip_proto, proto_msg);
+
+  // s1ap_imsi_map is not state service synced, so version will not be updated
```
ardzoht:
```suggestion
  // ueip_imsi_map is not state service synced, so version will not be updated
```

rsarwad:
> > Do we really want to use protobuf here for serialization while we plan to get rid of it completely? can't we use the same marshal/unmarshal pattern as in sessiond? @ardzoht @themarwhal what do you think?
> 
> I think I am missing something here: In sessiond we use folly + protos to serialize:
> 
> We build a marshaled key:value structure which we serialize into JSON using folly (see [here](https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/RedisStoreClient.cpp#L199-L199))
> 
> Every key on this marshaled structure is hardcoded (see [here](https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/StoredState.cpp#L496-L496)). Values are basic types (or folly arrays). But there are some of those values that are the result of a proto serialization (see [here](https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/StoredState.cpp#L513-L513)).
> 
> The reason to use proto serialization is that it abstracts changes on the proto definition from the serialization. In other words, if you change the proto, serialization is changed automatically too. Having to do manual serialization on every single change on session definition was not scalable in sessiond.
> 
> In fact,in my opinion, this `marshaled` object is nothing else than some kind of proto, but without all the advantages proto provides.
> 
> Going back to MME, do we want to introduce folly to MME to avoid protos?
> 
> If we don't want to use neither protos nor folly, the only way to serialize/deserialize the vector of IPs is manually creating a coma separated string (and its counterpart parsing function). This may work for this small hashmap, but we shouln't be doing this for bigger objects.
> 
> So we will need to use some kind of structure to serialize, and protos are well understood, supported and scale well.

From discussion, I understand that we will be keeping ueip_imsi map as part of proto file. That shall be updated to DB only on modification.

lte/gateway/c/oai/tasks/mme_app/mme_app_bearer.c
```diff
@@ -905,10 +905,20 @@ void mme_app_handle_delete_session_rsp(
     ue_context_p->nb_active_pdn_contexts -= 1;
   }
 
+  pdn_cid_t pid =
+      ue_context_p->bearer_contexts[EBI_TO_INDEX(delete_sess_resp_pP->lbi)]
+          ->pdn_cx_id;
+  pdn_context_t* pdn_context = ue_context_p->pdn_contexts[pid];
+  if ((pdn_context) && (pdn_context->paa.ipv4_address.s_addr)) {
+    OAILOG_DEBUG(
```
ardzoht: let's use OAILOG_DEBUG_UE here

ardzoht (commented):
left one more minor comment, it would be good before merging to double check that this does not regress performance, @uri200 can we run spirent testing to confirm this?

rdefosse:
Hello @rsarwad 

When running on the OAI pipeline, all DsTester runs look very similar to be a fluke:

On the 1st test we try to attach/detach 10 UEs  --> most runs only 6 attach

Same thing for the NSA test (1 over 3 attachments).

Is your PR having a performance impact?

Raphael

rsarwad:
> Hello @rsarwad
> 
> When running on the OAI pipeline, all DsTester runs look very similar to be a fluke:
> 
> On the 1st test we try to attach/detach 10 UEs --> most runs only 6 attach
> 
> Same thing for the NSA test (1 over 3 attachments).
> 
> Is your PR having a performance impact?
> 
> Raphael

Hi Raphael, thank you for noticing. 
@ardzoht Can I execute spirent test cases

rdefosse:
@rsarwad 

the run on the OAI pipeline `https://jenkins-oai.eurecom.fr/job/MAGMA-MME-production/3430/` is a false positive

One GitHub service was down and as a result, the OAI pipeline was in bypass mode.

Please re-trigger it again.

Sorry for the inconvenience
Raphael

rsarwad:
> @rsarwad
> 
> the run on the OAI pipeline `https://jenkins-oai.eurecom.fr/job/MAGMA-MME-production/3430/` is a false positive
> 
> One GitHub service was down and as a result, the OAI pipeline was in bypass mode.
> 
> Please re-trigger it again.
> 
> Sorry for the inconvenience
> Raphael

No Problem Raphael. Rebased to latest master codebase

rdefosse:
@rsarwad 

Thank you so much for resynching. 

It is still a problem `https://jenkins-oai.eurecom.fr/job/MAGMA-MME-production/3441/` 

I think I need to run this test manually

Raphael

rdefosse:
@rsarwad @lionelgo 

We are in `STATEFUL` mode.

I had to increase some timer to get the 10 UEs to attach.

But the detach sequence is going wrong.

[PR-7125-mme_check_run.zip](https://github.com/magma/magma/files/6616882/PR-7125-mme_check_run.zip)

themarwhal:
Hi @rsarwad, please mind the `Python Format Check` Github action.

rdefosse:
@rsarwad 

```bash
LeakSanitizer:DEADLYSIGNAL
==3917==ERROR: LeakSanitizer: SEGV on unknown address 0x000000000044 (pc 0x557ca3f58965 bp 0x7ff45beda8b0 sp 0x7ff45beda870 T15)
==3917==The signal is caused by a READ memory access.
==3917==Hint: address points to the zero page.
    #0 0x557ca3f58964 in mme_app_handle_delete_session_rsp /magma/lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c:910
    #1 0x557ca3f56f67 in handle_message /magma/lte/gateway/c/core/oai/tasks/mme_app/mme_app_main.c:189
    #2 0x7ff46bc72885 in zloop_start (/usr/lib/x86_64-linux-gnu/libczmq.so.4+0x27885)
    #3 0x557ca3f56c42 in mme_app_thread /magma/lte/gateway/c/core/oai/tasks/mme_app/mme_app_main.c:523
    #4 0x7ff46e6296da in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76da)
    #5 0x7ff46a73271e in __clone (/lib/x86_64-linux-gnu/libc.so.6+0x12171e)

LeakSanitizer can not provide additional info.
SUMMARY: LeakSanitizer: SEGV /magma/lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c:910 in mme_app_handle_delete_session_rsp
```

When you detach on `S11` with OAI SPGW-CUPS.

@lionelgo can give more details.

Raphael

lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c
```diff
@@ -905,10 +905,21 @@ void mme_app_handle_delete_session_rsp(
     ue_context_p->nb_active_pdn_contexts -= 1;
   }
 
+  pdn_cid_t pid =
+      ue_context_p->bearer_contexts[EBI_TO_INDEX(delete_sess_resp_pP->lbi)]
```
lionelgo, line 907: This linked bearer ID is not in 3GPP GTPV2C DELETE_SESSION_RESPONSE message, so this will make the CUPS variant segv.

lionelgo (changes requested):
The linked bearer ID is not in 3GPP GTPV2C DELETE_SESSION_RESPONSE message, so this will make the CUPS variant segv.

lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c
rsarwad: Hi @lionelgo if linked bearer ID is not in 3GPP GTPV2C DELETE_SESSION_RESPONSE message, then how to find the pdn context for this delete session procedure?

rsarwad: Hi @lionelgo  delete_sess_resp_pP->lbi was already being used, how it was not causing the issue for the CUPS variant?

ulaskozat: @lionelgo indeed `delete_sess_resp_pP->lbi` is used in several places in the mme code, are these code segments never executed when CUPS is used?

@rsarwad @lionelgo shall we simply sandwich these code segments between `#if EMBEDDED_SGW` and `#endif`?

ulaskozat: @rsarwad @lionelgo since this map is really used by the paging logic that is internal to Magma only and not used by OAI call flows, it should be safe to do what I suggest I think. @lionelgo could you please confirm?

ulaskozat: but I am also surprised that before these changes on line 909 within the if clause `(ue_context_p->emm_context.new_attach_info)`, the same lines of codes do exist.

rsarwad: Hi @ulaskozat @lionelgo , delete_sess_resp_pP->lbi is also used for PDN disconnect scenario.https://github.com/magma/magma/blob/master/lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c#L934

rsarwad: > @lionelgo indeed `delete_sess_resp_pP->lbi` is used in several places in the mme code, are these code segments never executed when CUPS is used?
> 
> @rsarwad @lionelgo shall we simply sandwich these code segments between `#if EMBEDDED_SGW` and `#endif`?

@ulaskozat I am thinking of two ways; 
1.Moving this snippet of code as part removing PDN context; And writing to DB should be called on every message transaction as it done for UE context and shall write to DB only if content of map is modified. Currently we only write to DB while modifying to DB. 
 In case if we are not calling writing to DB on every message transaction. What's happening is; on mme restart; mme calls free_state() which calls mme_app_ue_context_free_content() which eventually calls mme_app_free_pdn_context(); their removes the ue_ip from map; since content of map is changed; it writes to DB with ue_ip removed. 
 2. We need to define two pdn removal functions; in one function_1(): Only in-memory contents are freed, called as part mme restart  and in function_2(), along with removal of in-memory contexts, we shall write to DB also

ardzoht (changes requested):
Requesting changes to clear review queue

lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c
lionelgo: > Hi @lionelgo delete_sess_resp_pP->lbi was already being used, how it was not causing the issue for the CUPS variant?

Good question

ulaskozat: @lionelgo can we have a conclusion on this?

rdefosse:
@rsarwad 

Eurecom network crashed during the week-end.

A lot of PR were not run.

Can you please re-sync w/ `master` to re-start the CI pipelines?

Sorry for the inconvenience.

Raphael

lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c
ardzoht: @rsarwad I think point 1 introduces dependency between functions trying to maintain the state, if we can separate these logic into two functions (point 2) that is more clear for state changes, plus helps testability

rsarwad: Hi @ulaskozat, Could you confirm for making these changes

ulaskozat: I am not sure if I understood the benefits of these options. Isn't the issue here that we use lbi and OAI code does not use it? cc @ardzoht @lionelgo @rsarwad

rsarwad: In approach-2, mme shall remove ue_ip from ueip_imsi map, once mme deletes the PDN context.
 mme_app_free_pdn_context() is getting called when mme restarts, as part of clearing in-memeory contexts. In mme_app_remove_ue_ipv4_addr() ue_ip is removed from map and is also updated to DB. So mme_app_remove_ue_ipv4_addr() is called within mme_app_free_pdn_context(); ue_ip will not be written to DB. 
 
So thought of defining two functions to clear  PDN contexts; in function_1(): Only in-memory contents are freed, called as part mme restart and in function_2(), along with removal of in-memory contexts, we shall write to DB also.

ulaskozat: when mme is restarted, all in memory context is already gone and gets populated from Redis. I am not sure what clean up you are referring to here when mme restarts.

ardzoht: @rsarwad point is different from lbi discussion, it is about freeing in-memory state on mme exit, @rsarwad can you specify when this situation happens `mme_app_remove_ue_ipv4_addr() ue_ip is removed from map and is also updated to DB`? It is not clear to me from your description

rsarwad: On mme restart, mme_app_exit() is invoked, then it invokes clear_mme_nas_state(). Backtrace would something looks like, 
mme_app_exit()
clear_mme_nas_state()
 MmeNasStateManager::getInstance().free_state()
 clear_mme_nas_hashtables()
 
 While clearing hahtable entries; mme_app_state_free_ue_context() gets invoked which eventually call mme_app_free_pdn_context();

In mme_app_free_pdn_context(), mme_app_remove_ue_ipv4_addr() was called to remove ue_ip from map and was also updated to DB.
So added code to remove ue_ip from map on reception of Delete session response message; Based on lbi, ue_ip is fetched
To resolve the lbi dependency, I thought use different logic to remove ue_ip from and also update the DB.
During mme restart, let mme call mme_app_free_pdn_context() that includes only clearing in-memory contexts and shall update the DB.

ulaskozat: `MmeNasStateManager::getInstance().free_state()` does not need to free ueip_imsi_map as it is not a dynamically created data structure at the level we are using. Class destructor itself would take care of cleaning up the dynamic memory created under the hood when we come out of scope. Why we should worry about removing IP addresses in the map when the process is exiting?

ulaskozat: map::clear() method can be used if/when needed.

rsarwad: Yes we don't need to clear map contents as it doesn't have dynamic contexts. I shall call mme_app_remove_ue_ipv4_addr() within pdn_connectivity_delete(). So we don't have to rely on lbi within delete session response and on mme restart, map is not modified.

ardzoht:
@rsarwad There are some merge conflicting files going on, could you please rebase on top of latest master so I can give a final re-pass on this? Thanks!

ardzoht (changes requested):
Requesting changes to clear the review queue

rsarwad:
> Requesting changes to clear the review queue

Could you let me know the steps to clear the review queue

ardzoht:
> > Requesting changes to clear the review queue
> 
> Could you let me know the steps to clear the review queue

Sorry, was just specifying that I was requesting changes due to the merge conflicts, can you resolve these so I can take another pass ? Thanks

ulaskozat (approved):
LGTM

ulaskozat:
@ardzoht @lionelgo could you please have a final review on this and stamp?

lte/gateway/c/core/oai/tasks/mme_app/mme_app_ip_imsi.cpp
```diff
@@ -15,50 +15,48 @@ limitations under the License.
...
 /* Description: ue_ip address is allocated by either roaming PGWs or mobilityd
- * So there is possibility to allocate same ue ip address for the UEs.
+ * So there is possibility to allocate same ue ip address for different UEs.
  * So defining ue_ip_imsi map with key as ue_ip and value as list of imsis
  * having same ue_ip
  */
 int mme_app_insert_ue_ipv4_addr(uint32_t ipv4_addr, imsi64_t imsi64) {
   OAILOG_FUNC_IN(LOG_MME_APP);
+  UeIpImsiMap& ueip_imsi_map =
+      magma::lte::MmeNasStateManager::getInstance().get_mme_ueip_imsi_map();
```
ardzoht: minor nit: add `using magma::lte::MmeNasStateManager` as this class is used multiple times on this file

lte/gateway/python/scripts/state_cli.py
```diff
@@ -100,6 +104,7 @@ class StateCLI(object):
         'mme_nas_state': MmeNasState,
         'spgw_state': SpgwState,
         's1ap_state': S1apState,
+        'mme_ueip_imsi_map': MmeUeIpImsiMap,
```
ardzoht, line 107: Thanks for this! 👍

ardzoht (approved):
left one minor nit, overall LGTM

lte/gateway/c/core/oai/tasks/mme_app/mme_app_ip_imsi.cpp
rsarwad: Incorporated minor comment

uri200 (approved):

lte/gateway/python/integ_tests/s1aptests/test_paging_after_mme_restart.py
```diff
@@ -0,0 +1,139 @@
...
+        self.assertEqual(
+            response.msg_type, s1ap_types.tfwCmd.UE_CTX_REL_IND.value,
+        )
+        wait_time = 0.3
+        time.sleep(wait_time)
+        print('************************* Restarting MME service on', 'gateway')
+        self._s1ap_wrapper.magmad_util.restart_services(['mme'])
+
+        wait_time = 20
+        for j in range(wait_time):
```
lionelgo: Why not a reverse range to display remaining time to wait ? range(wait_time, 0, -1)

lionelgo (approved):

PR #7130, opened 2021-05-25
Add basic syntax check for codeowners changes

themarwhal:

## Summary
- Useful syntax check that should be run for all codeowners changes since they can't be tested before committing
- The tool used: https://github.com/mszostok/codeowners-validator
- I want to enable the owners check eventually, which checks that all users/teams specified belong in the org, but could not get permissions working for some reason. (Will bucket it for now)

## Additional Information


themarwhal:
wip

PR #7142, opened 2021-05-25
reduce noisy error level log in sessiond

themarwhal:

## Summary
I'm seeing tons of noisy error level logs like these:
```
May 21 16:52:17 phy-u2 sessiond[1067372]: I0521 16:52:17.786420 1067372 SessionState.cpp:473] Reported rule (internal_default_drop_flow_rule) not found, ignoring
May 21 16:52:17 phy-u2 sessiond[1067372]: I0521 16:52:17.786453 1067372 SessionState.cpp:473] Reported rule (internal_default_drop_flow_rule) not found, ignoring
May 21 16:52:17 phy-u2 sessiond[1067372]: I0521 16:52:17.786485 1067372 SessionState.cpp:473] Reported rule (internal_default_drop_flow_rule) not found, ignoring
May 21 16:52:17 phy-u2 sessiond[1067372]: I0521 16:52:17.786516 1067372 SessionState.cpp:473] Reported rule (internal_default_drop_flow_rule) not found, ignoring
```
Logging about the drop flow is not very useful, so adding a condition to only log if it's not that rule.

## Test Plan
make precommit_sm

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/ChargingGrant.cpp
+++ b/lte/gateway/c/session_manager/ChargingGrant.cpp
@@ -389,7 +389,7 @@ static RedirectInformation_AddressType address_type_converter(
     case RedirectServer_RedirectAddressType_SIP_URI:
       return RedirectInformation_AddressType_SIP_URI;
     default:
-      MLOG(MERROR) << "Unknown redirect address type!";
+      MLOG(MWARNING) << "Unknown redirect address type!";
       return RedirectInformation_AddressType_IPv4;
   }
 }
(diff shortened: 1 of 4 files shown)


uri200:
As noisy log, we still have the policydb synch message coming every 2 seconds I think.

uri200 (approved):

themarwhal:
> As noisy log, we still have the policydb synch message coming every 2 seconds I think.

right we can get rid of that here too.

PR #7145, opened 2021-05-25
docusaurus v2 migration

shanku9:
Signed-off-by: shanku9 <ankursh@fb.com>

## Additional Information


hcgatewood (changes requested):
Let's take another pass at this before I dig in and review. Couple notes

- Do we need an entirely new directory? Theoretically the `docs/readmes` should be identical, or at least like 90% the same. Why not
```
docs/
    readmes/
    docusaurus/
    docusaurus_v2/
```

- In any case, if we do keep the separate directory, (a) why is the `docs/` directory changed at all and (b) name the new directory `docs_v2` and move around its subdirectories to mirror the `docs` directory -- namely, there should only be `readmes/` and `docusaurus/` top-level files/directories, with everything docusurus-specific encapsulated under `docusaurus/`
- Add a readme so I know how to go through and spin up the site locally, a summary of the differences from v1, etc
- What's with all these "FWA" docs?

Screen Shot 2021-06-03 at 3 18 48 PM

- Something seems very wrong with the versioned docs. For reference, the current versioned docs has like 30-100 docs per version, while the v2 seems redundant and has like 400+ per version. Related: previously, we *did not* copy assets between versions. Is this a required change in v2? Reason: they're big objects that basically never change, and will bloat our repo

Screen Shot 2021-06-03 at 3 22 07 PM

themarwhal (changes requested):
What's the status on this? Requesting changes to clear my queue

hcgatewood:
Closing for staleness, epic is still being tracked

PR #7146, opened 2021-05-25
Remove dead service mesh code

wangyyt1013:
Signed-off-by: Yuanyuting Wang <yuanyutingwang@fb.com>

## Summary

Remove Indexer.GetIndexerInfo rpc and related objects / function implementations
Remove reporter_hwid message field from IndexRequest proto
Remove StreamProvider.GetStreamName function definition and related implementations, and adjust related test cases to bypass this change

## Test Plan

Run all existing test cases; all tests should pass


orc8r/cloud/go/services/state/protos/indexer.proto
```diff
@@ -34,23 +31,11 @@ service Indexer {
...
-  repeated string state_types = 2;
-}
-
 message IndexRequest {
   // states to reindex
   repeated magma.orc8r.State states = 1;
   // network_id of the states
   string network_id = 2;
-  // reporter_hwid is the hwid of the gateway that reported the states
-  string reporter_hwid = 3;
```
uri200, line 53: Not sure if this apply for this service, but we normally don't remove fields on proto files, but deprecate it. 

This is to avoid that other gateways may be using a different version of the proto

```
message Foo {
 string old_field = 1 [deprecated=true];
}
```

hcgatewood: This is good general advice. In this case though since this is an Orc8r-internal service, and we don't store these protos, it's safe to remove the fields 🙂 @wangyyt1013

hcgatewood (changes requested):
Looks good. Go ahead and remove the `reporter_hwid` field, then I'll be ready to merge

orc8r/cloud/go/services/state/protos/indexer.proto
wangyyt1013: sounds good, just removed the field!

hcgatewood (approved):
💯

themarwhal (approved):
lgtm!

PR #7148, opened 2021-05-25
Fix loading all subscriber state

wangyyt1013:
Signed-off-by: Yuanyuting Wang <yuanyutingwang@fb.com>

## Summary

1. Add step in [SearchStates](https://github.com/magma/magma/blob/0a16e586054d0a416bbd9a2d2c547fffd3e610c6/orc8r/cloud/go/services/state/client_api.go#L80-L107) of `subscriberdb` to filter for states with exact match of the IMSI prefix (if specified) 
2. Add regression test to highlight this functionality

## Test Plan

1. Run the existing test suites; all tests should pass
*Note*: augmented test case `TestGetSubscriberByExactIMSI` is located in [the handlers_test.go for the subscriberdb service](https://github.com/magma/magma/blob/0a16e586054d0a416bbd9a2d2c547fffd3e610c6/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go).

### DIFF
--- a/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
+++ b/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
@@ -94,7 +94,6 @@ func GetHandlers() []obsidian.Handler {
 
 const (
 	mobilitydStateExpectedMatchCount = 2
-	imsiExpectedMatchCount           = 2
 )
 
 var (
@@ -102,8 +101,6 @@ var (
 	// Mobilityd states are keyed as <IMSI>.<APN>.
 	mobilitydStateKeyRe = regexp.MustCompile(`^(?P<imsi>IMSI\d+)\..+$`)
 
-	// imsiKeyRe captures the IMSI portion of any IMSI-prefixed ID strings
-	imsiKeyRe                    = regexp.MustCompile(`^(?P<imsi>IMSI\d+)\.?.*$`)
 	apnPolicyProfileLoadCriteria = configurator.EntityLoadCriteria{LoadAssocsFromThis: true, LoadAssocsToThis: true}
 )
 
@@ -410,7 +407,7 @@ func getSubscriberStateHandler(c echo.Context) error {
 		return nerr
 	}
 
-	states, err := getStateExactIMSIPrefix(networkID, allSubscriberStateTypes, nil, &subscriberID, serdes.State)
+	states, err := getStatesForIMSIs(networkID, allSubscriberStateTypes, &subscriberID, serdes.State)
 	if err != nil {
 		return makeErr(err)
 	}
@@ -543,25 +540,30 @@ func makeSubscriberStateHandler(desiredState string) echo.HandlerFunc {
 	}
 }
 
-func getStateExactIMSIPrefix(networkID string, typeFilter []string, keyFilter []string, keyPrefix *string, serdes serde.Registry) (state_types.StatesByID, error) {
-	// post-filter functionality to only return states with the exact IMSI prefix as specified
-	states, err := state.SearchStates(networkID, typeFilter, keyFilter, keyPrefix, serdes)
+func getStatesForIMSIs(networkID string, typeFilter []string, keyPrefix *string, serdes serde.Registry) (state_types.StatesByID, error) {
+	// Post-filter functionality to only return states with the exact IMSI
+	// prefix as specified
+	states, err := state.SearchStates(networkID, typeFilter, nil, keyPrefix, serdes)
 	if err != nil {
 		return nil, err
 	}
 
-	if !funk.IsEmpty(keyPrefix) {
-		for stateID := range states {
-			matches := imsiKeyRe.FindStringSubmatch(stateID.DeviceID)
-			if len(matches) != imsiExpectedMatchCount {
-				glog.Errorf("state device ID %s did not match IMSI-prefixed regex", stateID.DeviceID)
+	for stateID := range states {
+		imsi := stateID.DeviceID
+		if stateID.Type == lte.MobilitydStateType {
+			matches := mobilitydStateKeyRe.FindStringSubmatch(stateID.DeviceID)
+			if len(matches) != mobilitydStateExpectedMatchCount {
+				glog.Infof("state device ID '%s' did not match IMSI-prefixed regex", stateID.DeviceID)
 				continue
 			}
-			if matches[1] != *keyPrefix {
-				delete(states, stateID)
-			}
+			imsi = matches[1]
+		}
+
+		if imsi != *keyPrefix {
+			delete(states, stateID)
 		}
 	}
+
 	return states, nil
 }
 
@@ -647,7 +649,7 @@ func loadSubscriber(networkID, key string) (*subscribermodels.Subscriber, error)
 		return nil, err
 	}
 
-	states, err := getStateExactIMSIPrefix(networkID, allSubscriberStateTypes, nil, &key, serdes.State)
+	states, err := getStatesForIMSIs(networkID, allSubscriberStateTypes, &key, serdes.State)
 	if err != nil {
 		return nil, err
 	}
@@ -866,10 +868,13 @@ func deleteSubscriber(networkID, key string) error {
 	return nil
 }
 
+// loadAllStatesForIMSIs loads all states whose IMSI prefix is contained in the
+// IMSI array passed in as argument. If the imsis argument is nil, the function
+// loads all states given the networkID, regardless of their IMSIs.
 func loadAllStatesForIMSIs(networkID string, imsis []string) (map[string]state_types.StatesByID, error) {
-	imsiMap := make(map[string]int)
-	for i, v := range imsis {
-		imsiMap[v] = i
+	validIMSIs := map[string]struct{}{}
+	for _, v := range imsis {
+		validIMSIs[v] = struct{}{}
 	}
 
 	imsiKeyStates, err := state.SearchStates(networkID, subscriberStateTypesKeyedByIMSI, imsis, nil, serdes.State)
@@ -893,7 +898,7 @@ func loadAllStatesForIMSIs(networkID string, imsis []string) (map[string]state_t
 				glog.Errorf("mobilityd state composite ID %s did not match regex", sidKey)
 				continue
 			}
-			if _, exists := imsiMap[matches[1]]; !exists && len(imsiMap) != 0 {
+			if !shouldLoadState(validIMSIs, matches[1]) {
 				continue
 			}
 			sidKey = matches[1]
@@ -906,6 +911,14 @@ func loadAllStatesForIMSIs(networkID string, imsis []string) (map[string]state_t
 	}
 
 	return statesBySid, nil
+}
+
+func shouldLoadState(validIMSIs map[string]struct{}, imsi string) bool {
+	if len(validIMSIs) == 0 {
+		return true
+	}
+	_, valid := validIMSIs[imsi]
+	return valid
 }
 
 func makeSubscriberState(subscriberID string, states state_types.StatesByID) *subscribermodels.SubscriberState {
--- a/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
+++ b/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
@@ -407,7 +407,7 @@ func getSubscriberStateHandler(c echo.Context) error {
 		return nerr
 	}
 
-	states, err := getStatesForIMSIs(networkID, allSubscriberStateTypes, &subscriberID, serdes.State)
+	states, err := getStatesForIMSIs(networkID, allSubscriberStateTypes, subscriberID, serdes.State)
 	if err != nil {
 		return makeErr(err)
 	}
@@ -540,10 +540,10 @@ func makeSubscriberStateHandler(desiredState string) echo.HandlerFunc {
 	}
 }
 
-func getStatesForIMSIs(networkID string, typeFilter []string, keyPrefix *string, serdes serde.Registry) (state_types.StatesByID, error) {
+func getStatesForIMSIs(networkID string, typeFilter []string, keyPrefix string, serdes serde.Registry) (state_types.StatesByID, error) {
 	// Post-filter functionality to only return states with the exact IMSI
 	// prefix as specified
-	states, err := state.SearchStates(networkID, typeFilter, nil, keyPrefix, serdes)
+	states, err := state.SearchStates(networkID, typeFilter, nil, &keyPrefix, serdes)
 	if err != nil {
 		return nil, err
 	}
@@ -553,13 +553,13 @@ func getStatesForIMSIs(networkID string, typeFilter []string, keyPrefix *string,
 		if stateID.Type == lte.MobilitydStateType {
 			matches := mobilitydStateKeyRe.FindStringSubmatch(stateID.DeviceID)
 			if len(matches) != mobilitydStateExpectedMatchCount {
-				glog.Infof("state device ID '%s' did not match IMSI-prefixed regex", stateID.DeviceID)
+				glog.Infof("state device ID '%s' with type '%s' did not match IMSI-prefixed regex", stateID.DeviceID, stateID.Type)
 				continue
 			}
 			imsi = matches[1]
 		}
 
-		if imsi != *keyPrefix {
+		if imsi != keyPrefix {
 			delete(states, stateID)
 		}
 	}
@@ -649,7 +649,7 @@ func loadSubscriber(networkID, key string) (*subscribermodels.Subscriber, error)
 		return nil, err
 	}
 
-	states, err := getStatesForIMSIs(networkID, allSubscriberStateTypes, &key, serdes.State)
+	states, err := getStatesForIMSIs(networkID, allSubscriberStateTypes, key, serdes.State)
 	if err != nil {
 		return nil, err
 	}
@@ -869,12 +869,21 @@ func deleteSubscriber(networkID, key string) error {
 }
 
 // loadAllStatesForIMSIs loads all states whose IMSI prefix is contained in the
-// IMSI array passed in as argument. If the imsis argument is nil, the function
-// loads all states given the networkID, regardless of their IMSIs.
+// IMSI array passed in as argument. If passed IMSIs is nil,
+// loads states for all IMSIs in the network.
 func loadAllStatesForIMSIs(networkID string, imsis []string) (map[string]state_types.StatesByID, error) {
-	validIMSIs := map[string]struct{}{}
+	requestedIMSIs := map[string]struct{}{}
 	for _, v := range imsis {
-		validIMSIs[v] = struct{}{}
+		requestedIMSIs[v] = struct{}{}
+	}
+
+	shouldLoadState := func(validIMSIs map[string]struct{}, imsi string) bool {
+		if len(validIMSIs) == 0 {
+			// load all states regardless of their IMSIs if requested IMSIs is nil,
+			return true
+		}
+		_, valid := validIMSIs[imsi]
+		return valid
 	}
 
 	imsiKeyStates, err := state.SearchStates(networkID, subscriberStateTypesKeyedByIMSI, imsis, nil, serdes.State)
@@ -898,7 +907,7 @@ func loadAllStatesForIMSIs(networkID string, imsis []string) (map[string]state_t
 				glog.Errorf("mobilityd state composite ID %s did not match regex", sidKey)
 				continue
 			}
-			if !shouldLoadState(validIMSIs, matches[1]) {
+			if !shouldLoadState(requestedIMSIs, matches[1]) {
 				continue
 			}
 			sidKey = matches[1]
@@ -911,14 +920,6 @@ func loadAllStatesForIMSIs(networkID string, imsis []string) (map[string]state_t
 	}
 
 	return statesBySid, nil
-}
-
-func shouldLoadState(validIMSIs map[string]struct{}, imsi string) bool {
-	if len(validIMSIs) == 0 {
-		return true
-	}
-	_, valid := validIMSIs[imsi]
-	return valid
 }
 
 func makeSubscriberState(subscriberID string, states state_types.StatesByID) *subscribermodels.SubscriberState {
--- a/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
+++ b/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
@@ -19,6 +19,8 @@ import (
 	"regexp"
 	"strconv"
 
+	"magma/orc8r/cloud/go/serde"
+
 	"github.com/hashicorp/go-multierror"
 	"github.com/thoas/go-funk"
 
@@ -92,6 +94,7 @@ func GetHandlers() []obsidian.Handler {
 
 const (
 	mobilitydStateExpectedMatchCount = 2
+	imsiExpectedMatchCount = 2
 )
 
 var (
@@ -99,6 +102,8 @@ var (
 	// Mobilityd states are keyed as <IMSI>.<APN>.
 	mobilitydStateKeyRe = regexp.MustCompile(`^(?P<imsi>IMSI\d+)\..+$`)
 
+	// imsiKeyRe captures the IMSI portion of any IMSI-prefixed ID strings
+	imsiKeyRe = regexp.MustCompile(`^(?P<imsi>IMSI\d+)\.?.*$`)
 	apnPolicyProfileLoadCriteria = configurator.EntityLoadCriteria{LoadAssocsFromThis: true, LoadAssocsToThis: true}
 )
 
@@ -405,7 +410,7 @@ func getSubscriberStateHandler(c echo.Context) error {
 		return nerr
 	}
 
-	states, err := state.SearchStates(networkID, allSubscriberStateTypes, nil, &subscriberID, serdes.State)
+	states, err := getStateExactIMSIPrefix(networkID, allSubscriberStateTypes, nil, &subscriberID, serdes.State)
 	if err != nil {
 		return makeErr(err)
 	}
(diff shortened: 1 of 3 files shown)


lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,166 @@ func TestGetSubscriber(t *testing.T) {
...
+		"n1",
+		[]configurator.NetworkEntity{
+			{Type: lte.APNEntityType, Key: apn1},
+			{Type: lte.APNEntityType, Key: apn2},
+		},
+		serdes.Entity,
+	)
+	assert.NoError(t, err)
+
+	// No sub profile configured, we should return "default"
```
hcgatewood:
```suggestion
	// Create two subscribers, one a prefix of the other
```

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,166 @@ func TestGetSubscriber(t *testing.T) {
...
+	stateTestInit.StartTestService(t)
+	err := configurator.CreateNetwork(configurator.Network{ID: "n1"}, serdes.Network)
+	assert.NoError(t, err)
+
+	e := echo.New()
+	testURLRoot := "/magma/v1/lte/:network_id/subscribers/:subscriber_id"
+	handlers := handlers.GetHandlers()
+	getSubscriber := tests.GetHandlerByPathAndMethod(t, handlers, testURLRoot, obsidian.GET).HandlerFunc
+
+	// preseed 2 apns
```
hcgatewood: Only need to create 1 apn here
```suggestion
	// Pre: create an APN
```

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,166 @@ func TestGetSubscriber(t *testing.T) {
...
+	apn1, apn2 := "foo", "bar"
+	_, err = configurator.CreateEntities(
```
hcgatewood, line 1065: Create the apn through the rest api rather than through configurator

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,166 @@ func TestGetSubscriber(t *testing.T) {
...
+		[]configurator.NetworkEntity{
+			{Type: lte.APNEntityType, Key: apn1},
+			{Type: lte.APNEntityType, Key: apn2},
+		},
+		serdes.Entity,
+	)
+	assert.NoError(t, err)
+
+	// No sub profile configured, we should return "default"
+	_, err = configurator.CreateEntities(
```
hcgatewood: Create the subscribers through the rest api rather than through configurator

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,166 @@ func TestGetSubscriber(t *testing.T) {
...
+		"n1",
+		[]configurator.NetworkEntity{
+			{
+				Type: lte.SubscriberEntityType, Key: "IMSI123",
+				Name: "Jane Doe",
+				Config: &subscriberModels.SubscriberConfig{
```
hcgatewood: Store these subscriber configs as variables, so you can reference them later in the test

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,166 @@ func TestGetSubscriber(t *testing.T) {
...
+
+	icmpStatus := &subscriberModels.IcmpStatus{LatencyMs: f32Ptr(12.34)}
+	ctx := test_utils.GetContextWithCertificate(t, "hw1")
+	test_utils.ReportState(t, ctx, lte.ICMPStateType, "IMSI123", icmpStatus, serdes.State)
+	mmeState := state.ArbitraryJSON{"mme": "foo"}
+	test_utils.ReportState(t, ctx, lte.MMEStateType, "IMSI123", &mmeState, serdes.State)
+	spgwState := state.ArbitraryJSON{"spgw": "foo"}
+	test_utils.ReportState(t, ctx, lte.SPGWStateType, "IMSI12345", &spgwState, serdes.State)
+	s1apState := state.ArbitraryJSON{"s1ap": "foo"}
+	test_utils.ReportState(t, ctx, lte.S1APStateType, "IMSI12345", &s1apState, serdes.State)
```
hcgatewood: Just report mme state for both, should be sufficient to validate the change in functionality

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,166 @@ func TestGetSubscriber(t *testing.T) {
...
+
+	// should only report states for IMSI123, not IMSI12345
+	tc = tests.Test{
+		Method:         "GET",
+		URL:            testURLRoot,
+		Handler:        getSubscriber,
+		ParamNames:     []string{"network_id", "subscriber_id"},
+		ParamValues:    []string{"n1", "IMSI123"},
+		ExpectedStatus: 200,
+		ExpectedResult: &subscriberModels.Subscriber{
```
hcgatewood: Side note: you can use https://github.com/magma/magma/blob/4886b0f0c5082668a339187030d5f4e0e6899aa6/lte/cloud/go/services/subscriberdb/obsidian/models/conversion.go#L36 to convert MutableSubscriber (the input to the post rest api) to Subscriber (the output of the get rest api)

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,166 @@ func TestGetSubscriber(t *testing.T) {
 	tests.RunUnitTest(t, e, tc)
 }
 
+func TestGetSubscriberByExactIMSI(t *testing.T) {
```
hcgatewood, line 1050: Add a comment above this test explaining why we added this

orc8r/cloud/go/services/state/client_api.go
```diff
@@ -103,7 +112,27 @@ func SearchStates(networkID string, typeFilter []string, keyFilter []string, key
...
+				glog.Errorf("state device ID %s did not match IMSI-prefixed regex", stateID.DeviceID)
+				continue
+			}
+			if matches[1] != *keyPrefix {
+				delete(states, stateID)
+			}
+		}
+	}
+
+	return states, nil
```
hcgatewood: This is the wrong place for the post-filtering. That is, the state service is doing it's job by, in a domain-agnostic way, providing key-prefix search functionality. So it's actually in the subscriberdb handlers (e.g. https://github.com/magma/magma/blob/23d8f7e37e9d66446aaa4f1b16b67f3f4f2b99ea/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go#L408) that we need to handle post-filtering.

We call `state.SearchStates` in a couple places in that handlers file, and we need to (a) test each spot and (b) filter/fix each of those locations where it's currently broken

Feel free to send me a message if you want more info

hcgatewood (changes requested):
Main things
1. Only use rest api for creating entities, don't use configurator directly
2. Move the post-filter step to the subscriberdb rest api handler file

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
wangyyt1013: The current version of code makes use of the `newMutableSubscriber` helper function, which requires two APNs and is used by other tests as well. For simplicity's sake, will probably stick to having two APNs @hcgatewood

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
wangyyt1013: Looked through the REST apis for this service as well as the other test cases, can't seem to find an appropriate api for creating an APN & the other test cases are consistently using the `configurator` for this purpose. Wondering if I should just stick to using the `configurator` then? @hcgatewood

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -862,6 +893,9 @@ func loadAllStatesForIMSIs(networkID string, imsis []string) (map[string]state_t
 				glog.Errorf("mobilityd state composite ID %s did not match regex", sidKey)
 				continue
 			}
+			if _, exists := imsiMap[matches[1]]; !exists && len(imsiMap) != 0 {
```
themarwhal: More of a question then a review comment :) 
what is the clause `len(imsiMap) != 0` for?

wangyyt1013: Hi Marie! Here's my thought process: the function `loadAllStatesForIMSIs` is used with the `imsis` field empty in `listSubscriberStateHandler`, in which case the expected behavior is to load the states for all subscribers, regardless of imsi. Therefore, when the `imsis` arg is an empty string, there's no need for `loadAllStatesForIMSIs` to filter out states of subscribers whose keys are not contained in `imsis` (which is empty). @themarwhal

hcgatewood: So yes, basically the `imsis` arg to this fn can be empty, in which case we search for "all" IMSIs. Let's handle this a little cleaner though

1. Document that functionality in the function docstring
2. Rather than "empty", indicate `imsis` should be `nil` to search all
3. Create an inline helper function, e.g. `shouldLoadIMSI`, that encapsulates this line (the map check and whether `imsis` is nil)

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -840,6 +867,11 @@ func deleteSubscriber(networkID, key string) error {
 }
 
 func loadAllStatesForIMSIs(networkID string, imsis []string) (map[string]state_types.StatesByID, error) {
+	imsiMap := make(map[string]int)
```
hcgatewood: Don't use hungarian notation https://docs.quantifiedcode.com/python-anti-patterns/readability/putting_type_information_in_a_variable_name.html -- instead, just call this e.g. `imsisExist`

hcgatewood: Use the literal declaration, and use the go-idiosyncrasy `struct{}` value
```suggestion
	imsisExist := &map[string]struct{}{}
```

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -538,6 +543,28 @@ func makeSubscriberStateHandler(desiredState string) echo.HandlerFunc {
 	}
 }
 
+func getStateExactIMSIPrefix(networkID string, typeFilter []string, keyFilter []string, keyPrefix *string, serdes serde.Registry) (state_types.StatesByID, error) {
```
hcgatewood: "getStatesForIMSIs", and you can remove the args that non of the callers use

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -538,6 +543,28 @@ func makeSubscriberStateHandler(desiredState string) echo.HandlerFunc {
...
+	// post-filter functionality to only return states with the exact IMSI prefix as specified
```
hcgatewood: Sentence-case your comments

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -538,6 +543,28 @@ func makeSubscriberStateHandler(desiredState string) echo.HandlerFunc {
...
+	states, err := state.SearchStates(networkID, typeFilter, keyFilter, keyPrefix, serdes)
+	if err != nil {
+		return nil, err
+	}
+
+	if !funk.IsEmpty(keyPrefix) {
```
hcgatewood: This function is only ever used for the purpose of using keyPrefix, so rework this so that you don't need this check

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -538,6 +543,28 @@ func makeSubscriberStateHandler(desiredState string) echo.HandlerFunc {
...
+		for stateID := range states {
+			matches := imsiKeyRe.FindStringSubmatch(stateID.DeviceID)
+			if len(matches) != imsiExpectedMatchCount {
+				glog.Errorf("state device ID %s did not match IMSI-prefixed regex", stateID.DeviceID)
```
hcgatewood: This can be induced by a misbehaving gateway. So it's not a "platform" error. So just do glog.Infof https://magma.github.io/magma/docs/next/contributing/contribute_conventions#go

hcgatewood:
```suggestion
				glog.Infof("state device ID '%s' did not match IMSI-prefixed regex", stateID.DeviceID)
```

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -538,6 +543,28 @@ func makeSubscriberStateHandler(desiredState string) echo.HandlerFunc {
...
+func getStateExactIMSIPrefix(networkID string, typeFilter []string, keyFilter []string, keyPrefix *string, serdes serde.Registry) (state_types.StatesByID, error) {
+	// post-filter functionality to only return states with the exact IMSI prefix as specified
+	states, err := state.SearchStates(networkID, typeFilter, keyFilter, keyPrefix, serdes)
+	if err != nil {
+		return nil, err
+	}
+
+	if !funk.IsEmpty(keyPrefix) {
+		for stateID := range states {
+			matches := imsiKeyRe.FindStringSubmatch(stateID.DeviceID)
```
hcgatewood: This is making an assumption about the format of the keys, which happens to be true generally, but isn't guaranteed -- the format of the keys is specified on a per-state-type basis. So you'll need to follow the pattern elsewhere in this file of checking the state's type for e.g. mobilityd type

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
hcgatewood: Yeah this is fine. There is a way, but this is easier and unrelated to the functionality under test

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,109 @@ func TestGetSubscriber(t *testing.T) {
...
+			// LastReportedTime is calculated in milliseconds
+			LastReportedTime: frozenClock * 1000,
+			LatencyMs:        f32Ptr(12.34),
+		},
+	}
+	sub1_with_state.State = &subscriberModels.SubscriberState{
+		Mme: mmeState1,
+	}
+
+	// should only report states for IMSI1234567890, not IMSI123456789000
```
hcgatewood: Here and elsewhere: sentence-case your comments

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,109 @@ func TestGetSubscriber(t *testing.T) {
...
+	// sub1 and sub2 differ in their mme states
+	mmeState1 := state.ArbitraryJSON{"mme": "foo"}
+	test_utils.ReportState(t, ctx, lte.MMEStateType, "IMSI1234567890", &mmeState1, serdes.State)
+	mmeState2 := state.ArbitraryJSON{"mme": "foo"}
+	test_utils.ReportState(t, ctx, lte.MMEStateType, "IMSI123456789000", &mmeState2, serdes.State)
+
+	sub1_with_state := sub1.ToSubscriber()
+	sub1_with_state.Monitoring = &subscriberModels.SubscriberStatus{
+		Icmp: &subscriberModels.IcmpStatus{
+			// LastReportedTime is calculated in milliseconds
```
hcgatewood: This comment isn't relevant, since we don't actually care about the semantic content of the reported state here

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -92,13 +94,16 @@ func GetHandlers() []obsidian.Handler {
...
+	imsiExpectedMatchCount           = 2
 )
 
 var (
 	// mobilitydStateKeyRe captures the IMSI portion of mobilityd state keys.
 	// Mobilityd states are keyed as <IMSI>.<APN>.
 	mobilitydStateKeyRe = regexp.MustCompile(`^(?P<imsi>IMSI\d+)\..+$`)
 
+	// imsiKeyRe captures the IMSI portion of any IMSI-prefixed ID strings
+	imsiKeyRe                    = regexp.MustCompile(`^(?P<imsi>IMSI\d+)\.?.*$`)
```
hcgatewood: As described below: you'll probably end up wanting to just use `mobilitydStateKeyRe` and can remove these new vars

hcgatewood (changes requested):
Okay looking good in terms of functionality and location in the call stack. Let's clean up the style a bit then I think we'll be ready 🙂

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -862,6 +898,9 @@ func loadAllStatesForIMSIs(networkID string, imsis []string) (map[string]state_t
 				glog.Errorf("mobilityd state composite ID %s did not match regex", sidKey)
 				continue
 			}
+			if !shouldLoadState(validIMSIs, matches[1]) {
```
themarwhal: 👍

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -538,6 +540,33 @@ func makeSubscriberStateHandler(desiredState string) echo.HandlerFunc {
...
+	if err != nil {
+		return nil, err
+	}
+
+	for stateID := range states {
+		imsi := stateID.DeviceID
+		if stateID.Type == lte.MobilitydStateType {
+			matches := mobilitydStateKeyRe.FindStringSubmatch(stateID.DeviceID)
+			if len(matches) != mobilitydStateExpectedMatchCount {
+				glog.Infof("state device ID '%s' did not match IMSI-prefixed regex", stateID.DeviceID)
```
themarwhal: If we want to log something here, might be useful to log the state type as well.

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -538,6 +540,33 @@ func makeSubscriberStateHandler(desiredState string) echo.HandlerFunc {
 	}
 }
 
+func getStatesForIMSIs(networkID string, typeFilter []string, keyPrefix *string, serdes serde.Registry) (state_types.StatesByID, error) {
```
hcgatewood: keyPrefix can just be a string, since we're always passing it in

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
hcgatewood: ^

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -839,7 +868,15 @@ func deleteSubscriber(networkID, key string) error {
 	return nil
 }
 
+// loadAllStatesForIMSIs loads all states whose IMSI prefix is contained in the
+// IMSI array passed in as argument. If the imsis argument is nil, the function
+// loads all states given the networkID, regardless of their IMSIs.
```
hcgatewood:
```suggestion
// loadAllStatesForIMSIs loads all states whose IMSI prefix is contained in the
// IMSI array passed in as argument. If passed IMSIs is nil,
// loads states for all IMSIs in the network.
```

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -839,7 +868,15 @@ func deleteSubscriber(networkID, key string) error {
...
 func loadAllStatesForIMSIs(networkID string, imsis []string) (map[string]state_types.StatesByID, error) {
+	validIMSIs := map[string]struct{}{}
```
hcgatewood:
```suggestion
	requestedIMSIs := map[string]struct{}{}
```

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -874,6 +913,14 @@ func loadAllStatesForIMSIs(networkID string, imsis []string) (map[string]state_t
 	return statesBySid, nil
 }
 
+func shouldLoadState(validIMSIs map[string]struct{}, imsi string) bool {
```
hcgatewood: Nit: define this function as an anonymous function within the caller's body (assign it to a variable within `loadAllStatesForIMSIs`). Because
- It's only relevant there
- We're only defining it for clarity's purpose
- It can use variables from the caller's scope

Ref: https://gobyexample.com/closures

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,108 @@ func TestGetSubscriber(t *testing.T) {
...
+	tests.RunUnitTest(t, e, tc)
+
+	// Now create some AGW-reported state for the subscribers
+	// First we need to register a gateway which can report state
+	_, err = configurator.CreateEntity(
+		"n1",
+		configurator.NetworkEntity{Type: orc8r.MagmadGatewayType, Key: "g1", Config: &models.MagmadGatewayConfigs{}, PhysicalID: "hw1"},
+		serdes.Entity,
+	)
+	assert.NoError(t, err)
```
hcgatewood: Use this convenience fn instead https://github.com/magma/magma/blob/dc72fd4842888768f0e2974b964d99d873cc4f23/orc8r/cloud/go/services/configurator/test_utils/utils.go#L33-L35

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1044,108 @@ func TestGetSubscriber(t *testing.T) {
...
+	sub1_with_state := sub1.ToSubscriber()
+	sub1_with_state.Monitoring = &subscriberModels.SubscriberStatus{
+		Icmp: &subscriberModels.IcmpStatus{
+			LastReportedTime: frozenClock * 1000,
+			LatencyMs:        f32Ptr(12.34),
+		},
+	}
+	sub1_with_state.State = &subscriberModels.SubscriberState{
+		Mme: mmeState1,
+	}
```
hcgatewood: - Don't use icmpStatus state at all, get rid of that and just report the mme state
- Call it `sub1` -- also, generally, use camelCase rather than snake_case

hcgatewood (changes requested):
Looking good, couple more things

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
wangyyt1013: Sounds good! I renamed it to `sub1ExpectedResult` instead so as there's no confusion about `sub1` being first a `MutableSubscriber` and then `Subscriber`, if that sounds ok to you? @hcgatewood

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -538,6 +540,33 @@ func makeSubscriberStateHandler(desiredState string) echo.HandlerFunc {
...
 
+func getStatesForIMSIs(networkID string, typeFilter []string, keyPrefix string, serdes serde.Registry) (state_types.StatesByID, error) {
+	// Post-filter functionality to only return states with the exact IMSI
+	// prefix as specified
+	states, err := state.SearchStates(networkID, typeFilter, nil, &keyPrefix, serdes)
+	if err != nil {
+		return nil, err
+	}
+
+	for stateID := range states {
```
hcgatewood:
```suggestion
	states, err := state.SearchStates(networkID, typeFilter, nil, &keyPrefix, serdes)
	if err != nil {
		return nil, err
	}

	// Returned states contain matches by prefix, so filter out non-exact matches
	for stateID := range states {
```

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -839,7 +868,24 @@ func deleteSubscriber(networkID, key string) error {
...
+	}
+
+	shouldLoadState := func(validIMSIs map[string]struct{}, imsi string) bool {
+		if len(validIMSIs) == 0 {
+			// load all states regardless of their IMSIs if requested IMSIs is nil,
+			return true
+		}
+		_, valid := validIMSIs[imsi]
+		return valid
+	}
```
hcgatewood: Use `requestedIMSIs` directly as a closure, rather than passing it in as an argument -- i.e. `shouldLoadState` should only take 1 argument

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1045,92 @@ func TestGetSubscriber(t *testing.T) {
 	tests.RunUnitTest(t, e, tc)
 }
 
+func TestGetSubscriberByExactIMSI(t *testing.T) {
+	// Regression test to ensure we are loading states with the exact same IMSI
+	// key as the subscriber
```
hcgatewood: Move this to the func docstring -- `TestGetSubscriberByExactIMSI is a regression test to ...`

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -1044,6 +1045,92 @@ func TestGetSubscriber(t *testing.T) {
...
+	ctx := test_utils.GetContextWithCertificate(t, "hw1")
+
+	// Sub1 and sub2 differ in their mme states
+	mmeState1 := state.ArbitraryJSON{"mme": "foo"}
+	test_utils.ReportState(t, ctx, lte.MMEStateType, "IMSI1234567890", &mmeState1, serdes.State)
+	mmeState2 := state.ArbitraryJSON{"mme": "fee"}
+	test_utils.ReportState(t, ctx, lte.MMEStateType, "IMSI123456789000", &mmeState2, serdes.State)
+
+	sub1ExpectedResult := sub1.ToSubscriber()
+	sub1ExpectedResult.Monitoring = &subscriberModels.SubscriberStatus{}
```
hcgatewood, line 1115: Do you need to set this (empty value)?

hcgatewood (changes requested):
Looks good, couple aesthetic items then lgtm

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
wangyyt1013: yep! this is to match the result to what's actually returned from the api call @hcgatewood

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -839,7 +866,24 @@ func deleteSubscriber(networkID, key string) error {
...
+		requestedIMSIs[v] = struct{}{}
+	}
+
+	shouldLoadState := func(imsi string) bool {
+		if len(requestedIMSIs) == 0 {
+			// load all states regardless of their IMSIs if requested IMSIs is nil,
+			return true
+		}
+		_, valid := requestedIMSIs[imsi]
+		return valid
```
hcgatewood: Generally just use `ok` for this var

hcgatewood (approved):
Nice and concise 🙂

PR #7149, opened 2021-05-25
AGW: datapath: Fix iptable rule config.

pshelar:
Rather than adding rule on every restart, check for rule.
this avoids accumulation of iptable rules on every service restart.

## Test Plan
validated on lab AGW.

## Additional Information


koolzz (approved):
lgtm

ulaskozat (approved):
LGTM

PR #7154, opened 2021-05-26
commit feg generated changes

themarwhal:

## Summary
run make gen in feg test container

TODO: feg generated files are not covered in the insync check. We should probably fix this at some point

## Test Plan
ci

## Additional Information


uri200:
not sure if this will fix the problem. I think there is an issue since we had to add FEG into LTE module at mconfig builder. We may need to look deeper

uri200 (approved):

themarwhal:
> not sure if this will fix the problem. I think there is an issue since we had to add FEG into LTE module at mconfig builder. We may need to look deeper

ohhh interesing.... maybe there is a version discrepency in the proto generation?

uri200:
I think the problem is on how we build the protos. Per my understanding t we wanted to keep somehow LTE and FEG separated. But IMO that is not possible on the long run. Protos shouldn't belong to a gateway, but to its interfaces. Since AGW and FEG talk each other that means they have a common interface between each other. So we should expect to have common objects. 

To me all protos should be part of an independent module so messages could be shared by any GRPC service. This way, for example we would have a single definition of an APN, not 3.

PR #7163, opened 2021-05-26
make ipfix optional in sessiond

themarwhal:

## Summary
we recently made a change to make ipfix reporting enabled for both lte and cwf agws, but it turns out we don't always need this feature to be enabled. 
if ipfix is not enabled on the pipelined side, this should be a no-op. But since we've been seeing performance issues in pipelined lately, it is best to have an option to turn off the extra set of GRPC calls.

default settings are:
cwf -> ipfix enabled
everything else (lte) -> ipfix disabled
This is consistent with the corresponding pipelined config

Updated startup logs:
```
sessiond_main.cpp:244] Started EventD response thread
sessiond_main.cpp:271] Started SPGW response thread
sessiond_main.cpp:172] Successfully connected to Redis
sessiond_main.cpp:141] ==== Constants/Configs loaded from sessiond.yml ====
sessiond_main.cpp:142] USAGE_REPORTING_THRESHOLD: 0.8
sessiond_main.cpp:144] TERMINATE_SERVICE_WHEN_QUOTA_EXHAUSTED: 1
sessiond_main.cpp:146] DEFAULT_REQUESTED_UNITS: 200000
sessiond_main.cpp:148] SEND_ACCESS_TIMEZONE: 1
sessiond_main.cpp:150] CLEANUP_DANGLING_FLOWS: 1
sessiond_main.cpp:152] SEND_IPFIX: 0
sessiond_main.cpp:153] ==== Constants/Configs loaded from sessiond.yml ====
LocalEnforcer.cpp:1981] Access timezone is UTC +0 minutes
sessiond_main.cpp:313] Started SessionD restart handler thread
sessiond_main.cpp:337] Added LocalSessionManagerAsyncService to service's server
sessiond_main.cpp:339] Added SessionProxyResponderAsyncService to service's server
sessiond_main.cpp:364] Initialized SetMessageManagerHandler
sessiond_main.cpp:370] Added SessionProxyResponderAsyncService to service's server
sessiond_main.cpp:406] Started proxy service thread
sessiond_main.cpp:401] Started local service thread
SessionManagerServer.cpp:84] Initializing new call data for SetAmfSessionContext
```

## Test Plan
ci
unit test
cwf integ test

## Additional Information

### DIFF
--- a/cwf/gateway/configs/sessiond.yml
+++ b/cwf/gateway/configs/sessiond.yml
@@ -51,3 +51,7 @@ sessions_table: sessiond:sessions
 # Sets the amount of octets that will be requested on the CCR-I for credit if
 # credit is given. If unset, defaults to 200000
 default_requested_units: 200000
+
+# set to true if subscriber information export over IPFIX needs to be enabled
+# in order to enable ipfix, make sure to add ipfix as a static service in pipelined.yml
+enable_ipfix: true
--- a/cwf/gateway/integ_tests/sessiond.yml
+++ b/cwf/gateway/integ_tests/sessiond.yml
@@ -42,3 +42,7 @@ sessions_table: sessiond:sessions
 # Sets the amount of octets that will be requested on the CCR-I for credit if
 # credit is given. If unset, defaults to 200000
 default_requested_units: 200000
+
+# set to true if subscriber information export over IPFIX needs to be enabled
+# in order to enable ipfix, make sure to add ipfix as a static service in pipelined.yml
+enable_ipfix: true
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -34,6 +34,7 @@
 namespace magma {
 bool LocalEnforcer::SEND_ACCESS_TIMEZONE   = false;
 bool LocalEnforcer::CLEANUP_DANGLING_FLOWS = true;
+bool LocalEnforcer::SEND_IPFIX             = true;
 
 using google::protobuf::RepeatedPtrField;
 
@@ -1818,6 +1819,10 @@ void LocalEnforcer::create_bearer(
 void LocalEnforcer::update_ipfix_flow(
     const std::string& imsi, const SessionConfig& config,
     const uint64_t pdp_start_time) {
+  if (!LocalEnforcer::SEND_IPFIX) {
+    return;
+  }
+
   MLOG(MDEBUG) << "Updating IPFIX flow for subscriber " << imsi;
   SubscriberID sid;
   sid.set_id(imsi);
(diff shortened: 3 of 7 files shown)


koolzz (approved):
awesome

ardzoht (approved):
lgtm

uri200:
Is that really saving sessiond anything? I am asking because having a setting spread across systems makes its configuration kind of difficult. you can forget to enable it in one of the systems. If sessiond had a way to know if pipelind has it enabled then it would be different.

What would happen if pipelined has it true and sessiond false? Or reverse

themarwhal:
> Is that really saving sessiond anything? I am asking because having a setting spread across systems makes its configuration kind of difficult. you can forget to enable it in one of the systems. If sessiond had a way to know if pipelind has it enabled then it would be different.
> 
> What would happen if pipelined has it true and sessiond false? Or reverse

I don't think it'll save sessiond anything and that is not the intention. We're trying to manage load in pipelined at this point. 
I think your point is valid about the split config though, maybe there is a way to signify from PipelineD to SessionD during the setup stage about this config. I thought about doing both via mconfig, but from the way it is configured in pipelined.yml, it didn't seem worth the trouble. (@koolzz thoughts?)

As far as what happens when the configs are inconsistent:
* sessiond has ipfix enabled, pipelined has ipfix disabled -> this is the default setting for all lte agws rn. the request from sessiond just gets ignored.
* sessiond has ipfix enabled, pipelined has ipfix enabled -> this is the default setting for all cwf agws. this is the happy case.
* sessiond has ipfix disabled, pipelined has ipfix enabled -> probably resulting in the feature not working properly
* sessiond has ipfix disabled, pipelined has ipfix disabled -> also a happy case

koolzz:
> > Is that really saving sessiond anything? I am asking because having a setting spread across systems makes its configuration kind of difficult. you can forget to enable it in one of the systems. If sessiond had a way to know if pipelind has it enabled then it would be different.
> > What would happen if pipelined has it true and sessiond false? Or reverse
> 
> I don't think it'll save sessiond anything and that is not the intention. We're trying to manage load in pipelined at this point.
> I think your point is valid about the split config though, maybe there is a way to signify from PipelineD to SessionD during the setup stage about this config. I thought about doing both via mconfig, but from the way it is configured in pipelined.yml, it didn't seem worth the trouble. (@koolzz thoughts?)
> 
> As far as what happens when the configs are inconsistent:
> 
> * sessiond has ipfix enabled, pipelined has ipfix disabled -> this is the default setting for all lte agws rn. the request from sessiond just gets ignored.
> * sessiond has ipfix enabled, pipelined has ipfix enabled -> this is the default setting for all cwf agws. this is the happy case.
> * sessiond has ipfix disabled, pipelined has ipfix enabled -> probably resulting in the feature not working properly
> * sessiond has ipfix disabled, pipelined has ipfix disabled -> also a happy case

Yeah mconfig sounds good, we already have something for specifying ipfix targets. I think its a clean up task though I'd rather we just have it in .yml for now to reduce pipelined stress for 1.6

uri200 (approved):

PR #7164, opened 2021-05-26
Improve Enforcement Stats Poll Time Calculation

koolzz:

## Summary

 - Adds a timestamp logic to pipelined to poll only after the report was sent succesfully to sessiond. This helps when pipelined gets overloaded during high ue attach rates(and can't keep up with poll/report cycle)
 - Removes the yield logic and runs a loop for stats processing (this seems to improve perf slightly)

## Test Plan

```
(python) vagrant@magma-dev:~/magma/lte/gateway/python/scripts$ pipelined_cli.py enforcement stress_test_grpc --attaches_per_sec 50 --num_of_ues 5000 --time_between_detach 10 --test_iterations 1 --disable_qos 
/home/vagrant/build/python/lib/python3.5/site-packages/scapy/config.py:411: CryptographyDeprecationWarning: Python 3.5 support will be dropped in the next release of cryptography. Please upgrade your Python.
  import cryptography
WARNING: DO NOT USE ON PRODUCTION SETUPS
Attach every ~0.02 seconds
QOS Disabled
Starting iteration 0 of attach/detach requests
Starting attaches
Finished 5000 attaches in 109.917671 seconds
Actual attach rate = 45 UEs per sec
Starting detaches
Finished 5000 detaches in 108.484732 seconds
Actual detach rate = {0} UEs per sec 46

```
## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/enforcement_stats.py
+++ b/lte/gateway/python/magma/pipelined/app/enforcement_stats.py
@@ -12,6 +12,7 @@ limitations under the License.
 """
 import os
 from collections import defaultdict
+from datetime import datetime, timedelta
 
 from lte.protos.pipelined_pb2 import RuleModResult
 from lte.protos.policydb_pb2 import FlowDescription
@@ -59,6 +60,8 @@ class EnforcementStatsController(PolicyMixin, RestartMixin, MagmaController):
     SESSIOND_RPC_TIMEOUT = 10
     # 0xffffffffffffffff is reserved in openflow
     DEFAULT_FLOW_COOKIE = 0xfffffffffffffffe
+    INIT_SLEEP_TIME = 3
+    MAX_DELAY_INTERVALS = 20
 
     _CONTEXTS = {
         'dpset': dpset.DPSet,
@@ -85,6 +88,8 @@ class EnforcementStatsController(PolicyMixin, RestartMixin, MagmaController):
             kwargs['config']['enforcement']['default_drop_flow_name']
         self.flow_stats_thread = hub.spawn(self._monitor, poll_interval)
         self._print_grpc_payload = os.environ.get('MAGMA_PRINT_GRPC_PAYLOAD')
+        self._last_poll_time = datetime.now()
+        self._last_report_timestamp = datetime.now()
         if self._print_grpc_payload is None:
             self._print_grpc_payload = \
                 kwargs['config'].get('magma_print_grpc_payload', False)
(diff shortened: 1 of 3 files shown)


lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -347,20 +363,14 @@ def _handle_flow_stats(self, stats_msgs):
 
         self.logger.debug("Processing %s stats responses", len(stats_msgs))
         # Aggregate flows into rule records
-        current_usage = defaultdict(RuleRecord)
         for flow_stats in stats_msgs:
```
koolzz, line 375: ah I need to merge all the stats_msgs into one single dict. fixing

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -289,14 +293,26 @@ def _monitor(self, poll_interval):
...
-                    self._poll_stats(datapath)
-                else:
-                    # Still send an empty report -> needed for pipelined setup
-                    self._report_usage({})
             hub.sleep(poll_interval)
+            for _, datapath in self.dpset.get_all():
+                now = datetime.now()
+                # If we're struggling to report give it some time to recover
+                if self._last_poll_time > self._last_report_timestamp:
+                    self.logger.error('Previous update running late, sleeping')
```
pshelar: should this be at ERROR level? how this msg would help operator or us to fix or diagnose this issue?

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -289,14 +293,26 @@ def _monitor(self, poll_interval):
...
+                    self._last_report_timestamp = now
```
pshelar, line 315: why report timestamp is update even when this function is not reporting statss.

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -289,14 +293,26 @@ def _monitor(self, poll_interval):
...
+                    hub.sleep(poll_interval)
```
pshelar: in this case stat monitoring thread can sleep for longer duration, since the thread just woke up after poll_interval time.
can we make it dynamic ? we could look at avg time required for last polling function and guess when it would done.
or  simply reduce the poll_interval by half.

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -347,20 +363,17 @@ def _handle_flow_stats(self, stats_msgs):
...
-                    return
+            aggregated_msgs += flow_stats
+
+        self.logger.debug("Processing stats of %d flows", len(aggregated_msgs))
+        try:
+            current_usage = self._get_usage_from_flow_stat(aggregated_msgs)
+        except ConnectionError:
+            self.logger.error('Failed processing stats, redis unavailable')
+            self.unhandled_stats_msgs.append(stats_msgs)
+            return
```
pshelar, line 384: can you explain why this refactoring change is done in this commit ?

pshelar (commented):
thanks for tracking this down.

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
koolzz: yeah good point, maybe info?

pshelar: I am fine with info level, but it needs more info to debug it further. like timestamp values, amount of time spent in polling.
we could also throw exception to restart pipelineD if it is stuck long enough.

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
koolzz: Because this is an edge where we didn't get any reports for a while and we try to fix that by sleeping 2times. updating the report timestamp in case that one never gets updated so we don't get stuck forever

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
koolzz: I've seen the yield behave slightly worse that just looping through the function

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
koolzz: It's sleeping twice already because the beginning of the func will also sleep

koolzz: I think making this dynamic is overkill, we should take that when we get to smart polling

pshelar: I am fine have static sleep time, can we reduce it to 50% of poll interval so avoid longer delay in stats poll.

koolzz: Sure

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
koolzz: Basically, if we don't update the timestamp and for some reason we never reach the future callback we would never poll again

pshelar (approved):
LGTM

PR #7172, opened 2021-05-27
[orc8r][helm]Update helm chart version for feg-orc8r

emakeev:

Update helm chart version for feg-orc8r

## Additional Information


themarwhal (approved):

karthiksubraveti (changes requested):
Orc8r's chart has no dependency with feg-orc8r. We need to only bump the feg-orc8r chart.

PR #7187, opened 2021-05-27
[orc8r] Refactor swagger spec filepath handling

### LINKED ISSUE #7112: Fix Test_GetCommonSpec reading from /etc
`Test_GetCommonSpec` reads from `/etc`, which breaks when run on host (if `/etc/magma` doesn't exist). Instead, slightly rework code to better encapsulate the default config directory, and in the tests pass in a generated tmp directory.

markjen:
fixes #7112 

## Summary

- Allow injection of base dir path when reading swagger spec files
- Update test cases to run without requiring /etc/magma to exist and be writable

## Test Plan

- Unit tests for specs.go
- Updated/fixed unit tests for package


markjen:
mocked out swagger.Specs, added test coverage, rebased

orc8r/cloud/go/obsidian/swagger/poll.go
```diff
@@ -68,13 +66,3 @@ func GetServiceSpec(service string) (string, error) {
 
 	return yamlSpec, nil
 }
-
-// GetCommonSpec returns the Swagger common spec as a YAML string.
-func GetCommonSpec() (string, error) {
-	data, err := ioutil.ReadFile(commonSpecPath)
```
hcgatewood, line 74: Also remove `commonSpecPath` definition

orc8r/cloud/go/obsidian/swagger/spec_servicer.go
```diff
@@ -30,32 +26,34 @@ type specServicer struct {
...
+			err)
 	}
-	standalone, err := ioutil.ReadFile(standalonePath)
+	standalone, err := specs.GetStandaloneSpec(service)
 	if err != nil {
-		// Swallowing ReadFile error because the service should continue to
-		// run even if it can't find its standalone Swagger spec file.
-		glog.Errorf("Error retrieving Swagger Spec of service %s: %+v", service, err)
-		return NewSpecServicer("", "")
+		glog.Errorf(
```
hcgatewood: Here and elsewhere: don't line-break these. We're fine with 120+ char lines, and even past that just wrap as non-aggressively as possible

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -15,43 +15,53 @@ package swagger_test
...
 
 	"magma/orc8r/cloud/go/obsidian/swagger"
+	"magma/orc8r/cloud/go/obsidian/swagger/mocks"
 	"magma/orc8r/cloud/go/obsidian/swagger/protos"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestSpecServicer_NewSpecServicerFromFile(t *testing.T) {
+	// missing default dir / spec files should not panic
```
hcgatewood: Sentence-case all comments

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -15,43 +15,53 @@ package swagger_test
...
+	assert.NotNil(t, swagger.NewSpecServicerFromFile("foo"))
```
hcgatewood: Nit: I prefer `err := ...` then `assert.NotNil(t, err)`, just for readability

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -15,43 +15,53 @@ package swagger_test
...
+
+func TestSpecServicer_NewSpecServicerFromSpecs(t *testing.T) {
+	d, err := ioutil.TempDir("", "")
+	assert.NoError(t, err)
+	defer os.RemoveAll(d)
+
+	partialDir := filepath.Join(d, "partial")
+	standaloneDir := filepath.Join(d, "standalone")
+	assert.NoError(t, os.MkdirAll(partialDir, os.ModePerm))
+	assert.NoError(t, os.MkdirAll(standaloneDir, os.ModePerm))
```
hcgatewood: :+1:

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -15,43 +15,53 @@ package swagger_test
...
-	assert.NoError(t, err)
+	assert.NoError(
+		t,
+		ioutil.WriteFile(
+			tmpStandaloneSpecPath, []byte(testStandaloneFileContents), 0644))
 
 	// Success
-	servicer := swagger.NewSpecServicerFromFile("test_spec_servicer")
+	servicer := swagger.NewSpecServicerFromSpecs(
+		swagger.NewSpecs(d), "test_spec_servicer")
```
hcgatewood: Copying from above: don't wrap these lines. Also, go ahead and put the "actions" on separate lines, that way the asserts can be simple as `assert.NoError(t, err)`

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -66,6 +76,66 @@ func TestSpecServicer_NewSpecServicerFromFile(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Empty(t, partialRes.SwaggerSpec)
+
+	standaloneRes, err := servicer.GetStandaloneSpec(
+		context.Background(), &protos.StandaloneSpecRequest{})
+	assert.NoError(t, err)
+	assert.Equal(t, standaloneSpec, standaloneRes.SwaggerSpec)
+}
+
+func TestSpecServicer_NewSpecServicerFromSpecsStandaloneErr(t *testing.T) {
```
hcgatewood, line 98: Looks like your additions remove the need for the two `TestSpecsServicer_*` below, so can remove them

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -66,6 +76,66 @@ func TestSpecServicer_NewSpecServicerFromFile(t *testing.T) {
...
+	service := "another service"
+
+	standaloneErrMock := &mocks.Specs{}
+	partialSpec := "partial spec"
+	standaloneErrMock.On(
+		"GetPartialSpec", service,
```
hcgatewood: Similar to above: for constructing these mocks, break lines following this pattern https://github.com/magma/magma/blob/60df24fd4470e09dd02ab03c5429946010ad61eb/orc8r/cloud/go/blobstore/sql_test.go#L122-L124

orc8r/cloud/go/obsidian/swagger/specs.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"strings"
+)
+
+// readEntireFileString wraps ioutil.ReadFile and returns a string iff the
+// entire file can be read without error.
+func readEntireFileString(path string) (string, error) {
```
hcgatewood: "readFile"

hcgatewood: Place helper fns at the end -- start with, roughly speaking, the most important info at top of the file

orc8r/cloud/go/obsidian/swagger/specs.go
```diff
@@ -0,0 +1,88 @@
...
+const (
+	partial    specType = "partial"
+	standalone specType = "standalone"
+)
+
+func (t specType) path(base string, service string) string {
+	return filepath.Join(
+		base,
+		string(t),
+		fmt.Sprintf("%s.swagger.v1.yml", strings.ToLower(service)))
```
hcgatewood: Don't line break. If you do need to though, move the final closing paren to its own line

orc8r/cloud/go/obsidian/swagger/specs.go
```diff
@@ -0,0 +1,88 @@
...
+	GetPartialSpec(service string) (string, error)
+	GetStandaloneSpec(service string) (string, error)
+}
+
+// fsSpecs implements Specs via the filesystem at a specific base dir.
+type fsSpecs struct {
+	baseDir string
+}
+
+var _ Specs = (*fsSpecs)(nil)
```
hcgatewood: Don't think this is necessary, since we have `NewSpecs` below

orc8r/cloud/go/obsidian/swagger/specs.go
```diff
@@ -0,0 +1,88 @@
...
+	return readEntireFileString(partial.path(f.baseDir, service))
+}
+
+// GetStandaloneSpec returns ./standalone/$service.swagger.v1.yml
+func (f *fsSpecs) GetStandaloneSpec(service string) (string, error) {
+	return readEntireFileString(standalone.path(f.baseDir, service))
+}
+
+// NewSpecs returns Specs with the specified base path.
+func NewSpecs(baseDir string) Specs {
```
hcgatewood: "NewFSSpecs"

hcgatewood: Add a docstring note to use DefaultSpecs except with particular reason

orc8r/cloud/go/obsidian/swagger/specs.go
```diff
@@ -0,0 +1,88 @@
...
+	return &fsSpecs{baseDir: baseDir}
+}
+
+const defaultSpecBasePath = "/etc/magma/swagger/specs"
+
+// DefaultSpecs returns Specs with the default production base path.
+func DefaultSpecs() Specs {
```
hcgatewood: Verbify: "GetDefaultSpecs"

orc8r/cloud/go/obsidian/swagger/mocks/Specs.go
```diff
@@ -0,0 +1,73 @@
+// Code generated by mockery v1.0.0. DO NOT EDIT.
```
hcgatewood: Use a gen.go file to automatically generate this, e.g. https://github.com/magma/magma/blob/14449aa94ef81b7d259f63c33de4aa41bb1f3481/orc8r/cloud/go/blobstore/gen.go#L14

hcgatewood (changes requested):
Shape of the change looks great, thanks for adding! Couple aesthetic requests then I'll be ready to merge 🙂

orc8r/cloud/go/obsidian/swagger/specs.go
markjen: yes, although it's possible `NewFSSpecs` could be changed later (and the future author would most likely not add a type assert in at that time). do we generally prefer to have type assertions?

orc8r/cloud/go/obsidian/swagger/specs.go
markjen: this fn is intentionally named because it operates differently than ioutil.ReadFile (otherwise it shouldn't exist :) )

happy to change it and rely on the docstring, but that requires the reader to go to definition

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
markjen: Hmmm I'm not sure what you mean by "actions"

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
markjen: In this particular case, the return value is a `protos.SwaggerSpecServer` so `assert.NoError` cannot be used.higher

In general, `err :=` at larger-than-necessary scope can cause variable shadowing issues. I would recommend considering inlining err when possible. (Even in cases where there is not another use of `err` in scope at the moment, future maintainers may run into the issue)

orc8r/cloud/go/obsidian/swagger/spec_servicer.go
markjen: Curiously, there appears to be a convention around 80 col wrap on comments. One might consider that the format of source code is also for human consumption, which might lead us to consider the same 80 col wrap for both?

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
markjen: While the new tests are covering at the moment, the target unit of the new tests is `NewSpecServicerFromSpecs` and the target of the old tests is `NewSpecServicer`. It would be more complete to leave them all. Also, this ensures coverage is preserved even if the implementation of `NewSpecServicerFromSpecs` changes to not call `NewSpecServicer` internally.

markjen (commented):
Thanks for the review and comments! I've updated the diff to address feedback

orc8r/cloud/go/module.mk
```diff
@@ -29,6 +29,9 @@ fmt::
 	gofmt -s -w .
 
 gen::
+	# obsidian/swagger/gen.go depends on obsidian/swagger/protos/gen.go, so it
+	# must be generated first.
+	go generate obsidian/swagger/protos
```
markjen: @hcgatewood after orc8r/cloud/go/obsidian/swagger/gen.go was added, `make fullgen` in `orc8r/cloud` started failing because `magma/orc8r/cloud/go/obsidian/swagger` depends on generated .pb.go files from `magma/orc8r/cloud/go/obsidian/swagger/protos` (and I think `go generate ./...` runs in lexicographic order).

Not sure if we've run into this anywhere else before; let me know if there's a better way to get this dependency sorted properly. Note: if preferred, I think I can change line 35 to just have this package listed before `./...` with the same comment.

markjen: oops, I had to move this to [Makefile](https://github.com/magma/magma/commit/5b363f0ca16d83e3fbdd98be0c31087f52b381fd#diff-6407cdf0d027d391dd9c70c218d9ee3a8728895ac8e6029e41906afa5a81b809R13) since module.mk is shared.

orc8r/cloud/go/Makefile
```diff
@@ -9,6 +9,12 @@ TOOL_DEPS:= \
 	github.com/vektra/mockery/cmd/mockery \
 	github.com/wadey/gocovmerge \
 	gotest.tools/gotestsum
+
+gen::
+	# obsidian/swagger/gen.go depends on obsidian/swagger/protos/gen.go, so it
+	# must be generated first.
+	go generate ./obsidian/swagger/protos
+
```
hcgatewood: I don't think this change is necessary, please correct me if I'm wrong!

You can verify with `cd ${MAGMA_ROOT}/orc8r/cloud/docker && ./build.py --generate`

orc8r/cloud/go/obsidian/swagger/spec_servicer.go
```diff
@@ -30,32 +26,28 @@ type specServicer struct {
...
-func NewSpecServicerFromFile(service string) protos.SwaggerSpecServer {
-	service = strings.ToLower(service)
-	partialPath, standalonePath := getSpecPaths(service)
-	partial, err := ioutil.ReadFile(partialPath)
+// NewSpecServicer constructs a spec servicer.
+func NewSpecServicer(partialSpec string, standaloneSpec string) protos.SwaggerSpecServer {
+	return &specServicer{partialSpec: partialSpec, standaloneSpec: standaloneSpec}
+}
+
+func NewSpecServicerFromSpecs(specs Specs, service string) protos.SwaggerSpecServer {
```
hcgatewood: Nit: add docstring

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
hcgatewood: Generally my approach to testing non-critical functionality is to be lightweight / ready for change. But I'm fine with leaving it as well, agreed that it does provide a bit of extra completeness

orc8r/cloud/go/obsidian/swagger/specs.go
hcgatewood: I go back and forth. So far I've preferred to do the type assertion in the `New*` functions and call it a day. I think for uniformity let's just go with that, but I'm generally open to being convinced

orc8r/cloud/go/obsidian/swagger/specs.go
```diff
@@ -0,0 +1,87 @@
...
+	standalone specType = "standalone"
+)
+
+func (t specType) path(base string, service string) string {
+	return filepath.Join(base, string(t), fmt.Sprintf("%s.swagger.v1.yml", strings.ToLower(service)))
+}
+
+// readEntireFileString wraps ioutil.ReadFile and returns a string iff the
+// entire file can be read without error.
+func readEntireFileString(path string) (string, error) {
```
hcgatewood: Overall it's a nit, this fn name is fine, I just prefer to have overly simple+concise names to get the gist, then use docstring for higher-resolution understanding

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -66,6 +70,50 @@ func TestSpecServicer_NewSpecServicerFromFile(t *testing.T) {
...
+
+func TestSpecServicer_NewSpecServicerFromSpecsStandaloneErr(t *testing.T) {
+	service := "another service"
+
+	standaloneErrMock := &mocks.Specs{}
+	partialSpec := "partial spec"
+	standaloneErrMock.On("GetPartialSpec", service).
+		Return(partialSpec, errors.New("partial err"))
+	standaloneErrMock.On("GetStandaloneSpec", service).
+		Return("", nil)
```
hcgatewood: Here and above: I would go ahead and just not line break on these, they're short enough

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
```diff
@@ -15,43 +15,47 @@ package swagger_test
...
-
-	err := os.MkdirAll(partialDir, os.ModePerm)
-	assert.NoError(t, err)
-	err = os.MkdirAll(standaloneDir, os.ModePerm)
-	assert.NoError(t, err)
 
 	tmpPartialSpecPath := filepath.Join(partialDir, testFile)
-	err = ioutil.WriteFile(tmpPartialSpecPath, []byte(testPartialFileContents), 0644)
-	assert.NoError(t, err)
+	assert.NoError(t, ioutil.WriteFile(tmpPartialSpecPath, []byte(testPartialFileContents), 0644))
```
hcgatewood: The reason I don't like doing it like this is it "reads" in a surprising way -- you start by reading that this is an assert line, which you think is nullipotent, but then come to find out, later in the line, that the line actually has a side effect / is taking an action

To me, it reads more naturally to follow this pattern

```suggestion
	err = ioutil.WriteFile(tmpPartialSpecPath, []byte(testPartialFileContents), 0644)
	assert.NoError(t, err)
```

hcgatewood (changes requested):
Will respond to other comments here in a sec

Lgtm, just want clarification on the Makefile change (don't think we need it but lmk), then msg me and we can merge

orc8r/cloud/go/obsidian/swagger/spec_servicer.go
hcgatewood: So (a) the doc comments get ingested by external tools and could end up anywhere and (b) they will naturally end up a lot longer than 80/120 chars if we don't hard-break

To me the goal is to get a natural / easy-to-read flow. And line-breaking code decreases readability imo, while line-breaking docs doesn't have the same effect because a "line" is not a semantically meaningful designation in the latter

But ultimately, the Go convention (e.g. go internal code) is ~120 chars for code, 80 chars for docs

orc8r/cloud/go/obsidian/swagger/spec_servicer_test.go
hcgatewood: We end up with this shadowing problem either way considering the `val, err := ...` pattern, so I don't view that as a strong reason. My nit here is basically just from my other comment about the "readability" of the line, where I think it flows better to do `action \n assert` rather than `assert(action)`, if that makes sense

orc8r/cloud/go/Makefile
markjen: For posterity: after discussion Hunter, we moved the swagger.Specs interface swagger/spec.Loader, which also removed the generate dependency on missing pb.go files during a full regen.

hcgatewood (approved):
Clean and super helpful. Thanks @markjen!

PR #7190, opened 2021-05-28
[Orc8r] APN data models refactoring

karthiksubraveti:

## Summary

Added new network wide APNConfig and gateway specific APNGatewayConfig data models.

## Test Plan
`./build.py -t` runs fine


uri200:
Note we have APN defined in multiple services (see s6a_proxy.proto). Some services only use few of its parameters so they don't even define an APN object (see session_manager.proto). It may be confusing that some services use a total different model. Can this refactor be applied to all gateways and services?

karthiksubraveti:
> Note we have APN defined in multiple services (see s6a_proxy.proto). Some services only use few of its parameters so they don't even define an APN object (see session_manager.proto). It may be confusing that some services use a total different model. Can this refactor be applied to all gateways and services?

cc @uri200 This mainly affects the apn gateway config models we have defined. The s6a models should still be good. I don't see that providing any gateway specific configuration down. 
This is the proposal around this change https://github.com/magma/magma/pull/7191

However I see your point in using common models across the board here. We can evolve these changes over time to ensure feg and LTE uses or derives from common models. However for now the focus is on LTE aspect. If this change is small enough and doesn't impact FEG, then I can make changes there as well.

lte/protos/subscriberdb.proto
```diff
@@ -102,50 +103,9 @@ enum AccessNetworkIdentifier {
...
-  }
-  // resource is gateway-specific guidance for serving the APN
-  APNResource resource = 7;
-}
-
-message AggregatedMaximumBitrate {
-  // Maximum uplink bitrate
-  uint32 max_bandwidth_ul = 1;
-  // Maximum downlink bitrate
-  uint32 max_bandwidth_dl = 2;
```
pshelar, line 148: I thought we were going to keep this during the transition period to avoid breaking existing APIs.

karthiksubraveti: I haven't removed it. Just moved it to apn.proto file.

lte/protos/apn.proto
```diff
@@ -0,0 +1,128 @@
...
+    bool preemption_vulnerability = 4;
+}
+
+// APN authorized bitrate
+message AggregatedMaximumBitrate {
+  // Maximum uplink bitrate
+  uint32 max_bandwidth_ul = 1;
+  // Maximum downlink bitrate
+  uint32 max_bandwidth_dl = 2;
+}
```
ulaskozat: see the changes done in terms of units in other protos (e.g., https://github.com/magma/magma/blob/91bd361fa53cf6e079300624b9dcda2bd5f38e02/lte/protos/subscriberdb.proto#L144)

ulaskozat: please adopt the same...

pshelar (approved):
can you also fix import in loadtest_mobilityd.py.
otherwise looks good.

themarwhal (approved):

koolzz (approved):

uri200 (approved):

ulaskozat (approved):
LGTM. Please run integ tests as well before merging as I do not see that they are running by the CI.

PR #7196, opened 2021-05-28
[nms]Made Gateways register only unregistered eNodeBs

ato-dani:
## Summary
In the GatewayDetailConfigEdit.js file, I made the ran edit component to filter out an already registered eNodeB's under a different gateway. This way, when trying to register eNodeBs, it now shows only unregistered eNodeBs. Therefore it makes sure that an eNodeB can not be registered to more than one Gateway.

## Test Plan
Tested it by creating eNodeBs and access gateways and registering the eNodeBs to some access gateway and seeing what happens when trying to register the same eNodeB to another gateway. Video is provided. Also, I ran the previous tests that were written and made sure it is compatible.

## Additional Information

unregistered_eNodeB_test_part_1
unregistered_eNodeB_test_part_2


nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
```diff
@@ -808,7 +808,18 @@ export function RanEdit(props: Props) {
       setError(e.response?.data?.message ?? e.message);
     }
   };
-
+  const isEnodebUnregistered = (enb: enodeb, currentGateway: lte_gateway) => {
```
andreilee: Fix the flow error on this line. Asides from running flow on your machine, you can see the result of the Github check running `nms-flow-test`

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
```diff
@@ -855,15 +866,28 @@ export function RanEdit(props: Props) {
...
-                    />
-                  </MenuItem>
-                ))}
+                Object.keys(enbsCtx.state.enbInfo).map(enbSerial => {
+                  if (
+                    isEnodebUnregistered(
+                      enbsCtx.state.enbInfo[enbSerial].enb,
+                      props.gateway,
+                    )
+                  ) {
```
andreilee, line 878: Is this getting called on every render? It's a bit expensive to compute. See [useMemo](https://reactjs.org/docs/hooks-reference.html#usememo) hook

andreilee (changes requested):
Thanks for submitting this pull request! There are a lot of minor usability issues that need investment, so work like this is always appreciated.

The PR title should start with the workstream. Don't worry about including the specific file in the title.
`[nms] Made gateways register only unregistered eNodeBs`

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
ato-dani: I have updated it to use useCallback hook. Sorry, for the late reply. It's because I was learning the useMemo and useCallback hook and trying different alternatives like saving the enodeb state and filtering that which would have made it faster, but the saved enodeb state would not be visible when we change the access gateways. Any

andreilee (approved):

andreilee:
You should close this pull request

PR #7208, opened 2021-05-31
Magma 5 g

Kaleem-Wavelabs:
## Additional Information


feg/protos/s6a_proxy.proto
```diff
@@ -105,6 +105,8 @@ message AuthenticationInformationAnswer {
         bytes xres = 2;
         bytes autn = 3;
         bytes kasme = 4;
+        bytes ck = 5;
+        bytes ik = 6;
```
emakeev, line 109: According to https://www.etsi.org/deliver/etsi_ts/129200_129299/129272/15.10.00_60/ts_129272v151000p.pdf
CK & IK are not a part of E-UTRAN-Vector, they are in UTRAN-Vector (see below) & you can use already defined UTRANVector below.

emakeev (changes requested):
Could you please remove changes to s6a_proxy.proto, they may not be needed (see inline comment).
I'd also suggest splitting this PR & at least have separate PRs for proto changes

.circleci/config.yml
```diff
@@ -17,6 +17,10 @@ aliases:
     paths: "orc8r lte"
   - &c_cpp_build_verify
     paths: "orc8r/gateway/c orc8r/protos lte/gateway/c lte/protos"
+  - &session_manager_build_verify
```
ulaskozat, line 20: why does this PR touch circleci directory?

feg/protos/s6a_proxy.proto
ulaskozat: Actually, we need a bit more than reusing UTRANVector. I suggest adding an NGRANVector and return kausf (rather than kasme). This would require adding a new field `num_requested_ngran_vectors` to Authentication Information Request so that s6a proxy on subscriberdb knows that it needs to generate `kausf` and not `kasme` for the response. The logic to generate `kausf` needs to be added to subscriberdb. cc @emakeev @panyogesh

emakeev: what RFC/TS defines NGRANVector? S6a proxy converts gRPC fields to diameter AVPs and forwards requests to HSS. Just adding fields to protos won't be enough, we'd need to add the support for them in s6a_proxy & most likely - go-diameter. S6a proxy is dedicated to one specific protocol S6a (29.272), I checked up to v16.6 and I don't see  any mentioning NGRANVector there. s6a_proxy does not generate anything, it just passes through vectors returned by HSS & in general, doesn't depend on or even know of subscriberdb existence.
It's possible that we may need a new service/RPC instead. Currently subscriberdb & s6a_proxy are sharing/implementing the same service and clients decide which one to use: https://github.com/magma/magma/blob/master/lte/gateway/c/oai/lib/s6a_proxy/S6aClient.cpp#L170-L176
But, if protocols are diverging, we may want to consider a new service definition for subscriberdb.

ulaskozat: @emakeev these are 5G extensions there is no HSS etc. Why should proto extensions break FEG?

emakeev: @ulaskozat, they shouldn't break FeG/s6a_proxy, we can add optional fields to AuthenticationInformationRequest/Answer they just won't be supported by FeG. It could be confusing to have 5G extensions defined in s6a_proxy.proto that's why I am suggesting to  consider different service or RPC

ulaskozat: @emakeev I understand the concern. I will ask the team to implement a 5G specific RPC call that can be extended independently.

orc8r/cloud/deploy/bare-metal-ansible/deploy.sh
```diff
@@ -20,7 +20,8 @@ cd kubespray
 git fetch --all
 git checkout $KUBESPRAY_RELEASE
 # Copy ``inventory/sample`` as ``inventory/$CLUSTER_NAME``
-cp -rnp inventory/sample inventory/$CLUSTER_NAME
+mkdir -p ../inventory/$CLUSTER_NAME
+cp -rnp inventory/sample ../inventory/$CLUSTER_NAME
```
ulaskozat, line 24: why are these touched?

lte/gateway/c/session_manager/EnumToString.cpp
```diff
@@ -213,4 +215,22 @@ std::string event_trigger_to_str(EventTrigger event_trigger) {
   }
 }
 
+std::string request_origin_type_to_str(
```
themarwhal, line 218: Is this a new change? I see it already on master: https://sourcegraph.com/github.com/magma/magma/-/blob/lte/gateway/c/session_manager/EnumToString.cpp

themarwhal (changes requested):
This PR seems to contain some changes that already exist on master. Can you rebase and try to apply the changes again?

feg/protos/s6a_proxy.proto
uri200: Not sure if you refer to this, but note we already have some 5g fields on s6a_proxy [here](http://github.com/magma/magma/blob/7441d3f1467d142e0f01f101c0e5c0a7a1228592/feg/protos/s6a_proxy.proto#L146-L146) and [here](http://github.com/magma/magma/blob/7441d3f1467d142e0f01f101c0e5c0a7a1228592/feg/protos/s6a_proxy.proto#L281-L281)


Those are on [3gpp 29.272](https://www.etsi.org/deliver/etsi_ts/129200_129299/129272/16.06.00_60/ts_129272v160600p.pdf)

lte/gateway/c/session_manager/test/test_local_enforcer.cpp
```diff
@@ -2686,15 +2704,9 @@ TEST_F(LocalEnforcerTest, test_final_unit_redirect_activation_and_termination) {
...
-          IMSI1, ip_addr, ipv6_addr, CheckTeids(teids), RequestOriginType::GX));
-  // the request should has no rules so PipelineD deletes all rules
-  EXPECT_CALL(
-      *pipelined_client,
-      deactivate_flows_for_rules_for_termination(
-          IMSI1, ip_addr, ipv6_addr, CheckTeids(teids), RequestOriginType::GY))
-      .Times(1);
+      *pipelined_client, deactivate_flows_for_rules_for_termination(
+                             IMSI1, ip_addr, ipv6_addr, CheckTeids(teids),
+                             RequestOriginType::WILDCARD));
```
uri200, line 2709: why are we using wildcard? We do want to check there is one of eacn (Gx and Gy)

feg/protos/s6a_proxy.proto
ulaskozat: @uri200 this is a draft PR. Note that 5G fields you are referring to on s6a are for NSA that uses 4G core.

ekowtaylor:
@Kaleem-Wavelabs, this PR has now gone stale. To maintain a healthy codebase, it will be automatically closed unless there are plans to rebase and merge by close of week

ekowtaylor:
@Kaleem-Wavelabs, this PR has now gone stale. To maintain a healthy codebase, it will be automatically closed unless there are plans to rebase and merge by close of week

panyogesh:
Closing the pull request as no longer needed.

PR #7259, opened 2021-06-01
Fix clearing of PCO attributes to preserve const correctness

amarpad:
Move clearing of PCO to caller to preseve const correctness of
attribtes

Signed-off-by: Amar Padmanabhan <amarpadmanabhan@fb.com>

## Additional Information


ulaskozat (approved):
LGTM

PR #7269, opened 2021-06-02
[AGW] Fix parsing errors caused from scraping node exporter metrics.

karthiksubraveti:

## Summary

node exporter provides metrics with 5 value tuples. 
For e.g.
Sample(name='', labels={'device': 'eth0'}, value=74652.0, timestamp=None, exemplar=None)
Magmad parses metrics expecting only name, labels, value to be present. Added an additional *_ to throw away additional tuple values here. 
Also removed an unnecessary exception and let "unknown" metric types be parsed as untyped metrics instead. 

Additionally also ran precommit script to format the existing metrics_collector code
## Test Plan
Once this issue was fixed. I was able to export these metrics to prometheus. 
Screen Shot 2021-06-02 at 7 07 30 AM

## Additional Information

### DIFF
--- a/orc8r/gateway/python/magma/magmad/metrics_collector.py
+++ b/orc8r/gateway/python/magma/magmad/metrics_collector.py
@@ -32,8 +32,12 @@ from prometheus_client.parser import text_string_to_metric_families
 
 # ScrapeTarget Holds information required to scrape and process metrics from a
 # prometheus target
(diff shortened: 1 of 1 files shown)


themarwhal (approved):

hcgatewood (approved):
Stamping. @themarwhal is working on the autoformatting noise

PR #7281, opened 2021-06-02
[agw][ami] Clean agw ami and add optional ssh key

### LINKED ISSUE #7201: AGW Image Enrichment

quentinDERORY:
## Summary

AGW image enrichment:
- Clean instance before generating the ami: snowflake directory and agw challenge  key
- Add AWS KMS key when specified to agw ami
- Add ssh key when specified to agw ami

## Test Plan

Integrated playbook call to existing ci pipeline.
Pipeline worked fine and ssh in the instance to verify the presence of the key


experimental/cloudstrapper/README.md
```diff
@@ -234,7 +234,7 @@
 
     agw-configure: configures the AGW to include controller information Command:
     ```
-    ansible-playbook agw-configure.yaml -i <DynamicInventoryFile> -e "agw=tag_Name_<GatewayId>" -e "dirLocalInventory=<WORK_DIR>" -u ubuntu
+    ansible-playbook agw-configure.yaml -i <DynamicInventoryFile> -e "agw=tag_Name_<GatewayId>" -e "dirLocalInventory=<WORK_DIR>" -u ubuntu [ -e KMSKeyID=<KEY_ID_TO_ADD_TO_SSH_AUTHORIZED> -e sshKey=<PATH-TO-PUBLIC-KEY-TO-ADD>]
```
arunuke, line 237: Perhaps adding a reference to tag platform if no keys need to be added?

experimental/cloudstrapper/playbooks/ami-configure.yaml
```diff
@@ -2,7 +2,7 @@
 - hosts: "{{ aminode }}"
   roles:
     - magma-agw-1.5
-    - {role: clean-authorized-keys, tags: clearSSHKeys}
```
arunuke, line 5: I think there is a use-case where we do need to clean keys (for the base AMI generation). Can we still keep this one as well and have customers choose the option based on tags?

arunuke: This can be resolved. I see you have covered it in clean-ami below.

arunuke (changes requested):
LGTM for the most part. Some changes to documentation to indicate whether customer wants to clean up keys only or wants to add keys would be great. For the standard AMI generation used by our devops chain, we will clean up everything. For a customer related build, keys can be added. If there are tags that are described these use-cases, we should be good.

experimental/cloudstrapper/README.md
quentinDERORY: By default it will use the platform tag and not run the add keys if none are defined. I added a tag to use skip-tag in our devops usecase.

arunuke: Good point and I agree. Resolving this.

arunuke (approved):
LGTM. Please proceed with merge.

tmdzk (approved):

PR #7299, opened 2021-06-03
add a Github Action to check for any leftover python formatting 

themarwhal:

## Summary
Few things:
1. refactor precommit.py/Dockerfile script so that the docker command can be run from anywhere in the codebase. (It previously had to be run in `lte/gateway/python`.) This adds a new dependency where the script requires the MAGMA_ROOT environment variable to be set. But I put in an additional warning log to indicate this and most dev setups have this env var already.
2. add a GitHub action to run the same set of formatting commands and fail if any changes are needed. This is non-required for now, but once I get around to fixing up the relevant python files, we can make this mandatory. (Note that the check currently excludes PipelineD as there is a lot of pending formatting changes there)

## Test Plan
Ran the check with multiple failing python files
Shouldn't greatly affect anything as this check is not mandatory yet.
(What it looks like when there is pending formatting change)
Screen Shot 2021-06-03 at 11 50 15 AM

## Additional Information

### DIFF
--- a/.github/workflows/python-formatting-check.yml
+++ b/.github/workflows/python-formatting-check.yml
@@ -1,0 +1,54 @@
+---
+name: Python Format Check
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+    paths:  # order matters here, always put includes then excludes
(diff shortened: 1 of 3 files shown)


ardzoht (approved):
lgtm

tmdzk (approved):

PR #7262, opened 2021-06-01
Create orc8r go library code to take deterministic digest of slice of unordered protobufs

wangyyt1013:
## Summary

1. Create new library package `mproto` in orc8r cloud
2. Create protobuf messages `ProtoSlice` for storing individual protobuf messages to be encoded, and `TestData` and `TestDataCollection` for testing purposes
3. Create function `EncodeProtosDeterministic` for encoding a slice of protobuf messages (uniquely indexed) using the current protobuf serialization implementation `encodeProtoDeterministic`
4. Create test cases (including a golden file test) that check if `EncodeProtosDeterministic` is consistently performing deterministic serialization, considering [common sources of protobuf undeterminism](https://gist.github.com/kchristidis/39c8b310fd9da43d515c4394c3cd9510)

## Test Plan

Run all existing Orc8r test cases; all should pass.
*Note*: The augmented test cases (2) can be found in `orc8r/cloud/go/mproto/mproto_test.go`.

## Additional Information

- [ ] This change is backwards-breaking

Signed-off-by: Yuanyuting Wang <yw3241@columbia.edu>

PR #7302, opened 2021-06-03
Update Orc8r’s subscriberdb servicer to support flat subscriber digests

wangyyt1013:

## Summary

1. Modify protos for `subscriberdb` servicer to include digest/no-update short-circuiting logic-related fields in `ListSubscribersRequest` and `ListSubscribersResponse`
2. Modify `ListSubscribers` rpc implementation to include logic for comparing gateway and cloud digests to decide whether to send over updated subscriber data
3. Add test case for communicating with the servicer using digests

## Test Plan

1. Run all existing tests. All should pass
2. Add new test case in `subscriberdb_servicer_test.go` that tests the `subscriberdb` cloud servicer with digests. This test should also pass

## Additional Information

- This PR is based on changed introduced in #7262


lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,13 +94,73 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
 		NextPageToken: nextToken,
+		Digest:        *digest,
+		NoUpdates:     noUpdates,
 	}
 	return listRes, nil
 }
 
+// getDigest loads all subscribers registered on the current network, and returns
+// a deterministic subscriber flat digest.
+func getDigest(gateway *protos.Identity_Gateway, apnsByName map[string]*lte_models.ApnConfiguration, apnResourcesByAPN lte_models.ApnResources) (*string, error) {
```
emakeev: This line seems to be too long even for Go's generous 120 chars convention.
Also, strings in Go are immutable & are 'pointers' already, like slices. they are passed/returned by values.

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,13 +94,73 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+	networkID := gateway.NetworkId
```
emakeev: use gateway.GetNetworkId() or check for gateway != nil

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,13 +94,73 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+
+		index := subProto.Sid.Id
+		subProtosById[index] = subProto
+	}
+
+	digest, err := mproto.MarshalManyDeterministic(subProtosById)
+	if err != nil {
+		return nil, err
+	}
+	digestB64 := b64.StdEncoding.EncodeToString(digest)
```
emakeev: I wonder if you even need an extra overhead of base64 encoding & extra memory allocation instead of just declaring & passing digests as []byte?
Just change 'string digest = 3;' to 'bytes digest = 3;' in your proto definition?

wangyyt1013: Thanks for the suggestion @emakeev, I'm currently using `string` simply because it appeared to be more intuitive to me & it's tentatively thus outlined in [the subscriber scaling project plan](https://magma.github.io/magma/docs/next/proposals/p010_subscriber_scaling#%EF%B8%8F-objective-3-reduce-network-pressure-at-scale), but I'll discuss this with @hcgatewood & keep you updated!

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -15,8 +15,13 @@ package servicers
 
 import (
 	"context"
+	b64 "encoding/base64"
```
hcgatewood: Don't alias unless necessary

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -52,8 +57,8 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
 	if !gateway.Registered() {
 		return nil, status.Errorf(codes.PermissionDenied, "gateway is not registered")
 	}
-	networkID := gateway.NetworkId
-	gatewayID := gateway.LogicalId
+	networkID := gateway.GetNetworkId()
+	gatewayID := gateway.GetLogicalId()
```
hcgatewood: Why the change? Generally don't use "guarded" getters unless you expect the object (gateway) to possibly be nil

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,13 +94,77 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
 		subProto.NetworkId = &protos.NetworkID{Id: networkID}
 		subProtos = append(subProtos, subProto)
 	}
+
+	// Short-circuit subscriber data transmission by comparing cloud & gateway digests.
```
hcgatewood: Nit: prefer "and" over using ampersands. Rationale is (a) it looks like the bitwise-and character, and (b) stdlib Go doesn't use it almost at all
```suggestion
	// Short-circuit subscriber data transmission by comparing cloud and gateway digests.
```

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,13 +94,77 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+	// If digest generation fails, the error is swallowed to not affect the main functionality.
+	//
+	// Note: currently the logic only applies to a request for the first page to avoid concurrency issues;
+	// more to be implemented with gateway subscriber cache that will canonize cloud digests.
```
hcgatewood: No need for this note in the doc comment, can move it to the relevant spot in the impl if you want to keep it for now

That said, since this is in-progress, not a huge need to document the transient note

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,13 +94,77 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+		digest, err = getDigest(gateway, apnsByName, apnResourcesByAPN)
+		if err != nil {
+			glog.Errorf("Generating digest for subscribers in network of gateway %s failed", gatewayID)
+		} else {
+			noUpdates = req.PreviousDigest == digest
+			if noUpdates {
+				subProtos = []*lte_protos.SubscriberData{}
+			}
+		}
+	}
```
hcgatewood: Move this to its own function

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,13 +94,77 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
 		NextPageToken: nextToken,
+		Digest:        digest,
+		NoUpdates:     noUpdates,
 	}
 	return listRes, nil
 }
 
+// getDigest loads all subscribers registered on the current network, and returns
+// a deterministic subscriber flat digest.
+func getDigest(
```
hcgatewood, line 125: Place this functionality behind a feature flag. That is, you want to be able to enable it for validation purposes, but we don't actually want to enable it in prod yet (since we're loading *all* subs on every first-page request)

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,13 +94,77 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+	gateway *protos.Identity_Gateway,
+	apnsByName map[string]*lte_models.ApnConfiguration,
+	apnResourcesByAPN lte_models.ApnResources,
+) (string, error) {
+	networkID := gateway.GetNetworkId()
```
hcgatewood: Again, don't guard the getter

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,13 +94,77 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+	lc := configurator.EntityLoadCriteria{
+		PageSize:           0,
+		PageToken:          "",
+		LoadConfig:         true,
+		LoadAssocsToThis:   true,
+		LoadAssocsFromThis: true,
+	}
```
hcgatewood: Side note: using PageSize of 0 only loads up to a max page size of subs (like 10k or something). So the proper way to do this is to iterate till the returned PageToken is empty

That said, this is fine for now, since we're only doing it this way for local validation purposes (i.e. behind a feature flag)

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,13 +94,77 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+		subProtosById[index] = subProto
+	}
+
+	digest, err := mproto.MarshalManyDeterministic(subProtosById)
+	if err != nil {
+		return "", err
+	}
+	digestB64 := b64.StdEncoding.EncodeToString(digest)
+
+	return digestB64, nil
```
hcgatewood: Ah, something I realized we forgot to do in the previous PRs -- we want this library code to return a *digest*, so the hash of this encoded value. Cause right now the encoded value is O(n) (for n := num subscribers), while the digest would be O(1) (constant length)

Here's the example -- converting O(n) bytes to O(1) bytes using md5.Sum https://github.com/magma/magma/blob/509e5b4bcdb68b0021044fb1dc8cf47554446b8f/orc8r/lib/go/protos/mconfig_digest.go#L24-L35

And here's the example proto defns

https://github.com/magma/magma/blob/509e5b4bcdb68b0021044fb1dc8cf47554446b8f/orc8r/protos/mconfig.proto#L34-L45

lte/protos/subscriberdb.proto
```diff
@@ -257,11 +257,20 @@ message ListSubscribersRequest {
   uint32 page_size = 1;
   // page_token is a serialized entity page token for paginated loads.
   string page_token = 2;
+  // previous_digest is the digest received from the most-recent request to
+  // ListSubscribers.
+  string previous_digest = 3;
```
hcgatewood: Just call it "digest"

lte/protos/subscriberdb.proto
```diff
@@ -257,11 +257,20 @@ message ListSubscribersRequest {
...
 message ListSubscribersResponse {
   repeated SubscriberData subscribers = 1;
   // next_page_token is a serialized entity page token for subsequent paginated
   // loads.
   string next_page_token = 2;
+  // digest is the server-defined, deterministic digest of the full set of
+  // returned subscriber objects.
+  string digest = 3;
+  // no_updates is true if client's existing subscribers match those on the
+  // server.
```
hcgatewood: Add additional note that `subscribers` will be empty when `no_updates` is true

hcgatewood (changes requested):

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
hcgatewood: The digest is meant to be ~4 bytes, and we're just getting/setting, not building. So using string is probably easier. (Though if we were sending something longer, or building the digest, bytes probably makes more sense.)

Ref for how we do/did it for mconfigs https://github.com/magma/magma/blob/509e5b4bcdb68b0021044fb1dc8cf47554446b8f/orc8r/protos/mconfig.proto#L34-L45

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
wangyyt1013: This practice of guarding against the gateway being `nil` has been recommended by @emakeev in a [previous review](https://github.com/magma/magma/pull/7302#discussion_r645828623); wondering what is the downside to using a guarded function like this?

lte/cloud/go/services/subscriberdb/config.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/golang/glog"
+)
+
+type Config struct {
+	// FlatDigestEnabled is a feature flag for the flat digest functionality.
+	FlatDigestEnabled bool `yaml:"flatDigestEnabled"`
+}
+
+func GetServiceConfig() Config {
```
hcgatewood: Since we're fataling here on failure, prefix fn name with `Must`

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
hcgatewood: Gateway won't ever be nil here -- and if it is, it's a bug and we'll fix it

IMO it's about communicating to future readers your assumptions about `gateway` -- when you use guarded getters, you communicate that `gateway` is expected to possibly be nil. But it's not, we expect it to always be non-nil here

Unfortunately Go doesn't handle nil-ness very well -- languages like Rust build nil-ness into the static type system and it makes this type of thing much easier. That said, Go's method is more concise, so let's stick with the conciseness for now -- we don't expect `gateway` to be nil ever, so communicate that properly by not using the guarded getter

hcgatewood: To follow up -- if `gateway` is nil (somehow), then the rest of the function runs with empty string for networkID and gatewayID, which is useless and potentially dangerous. So realistically the solution would be just checking whether `gateway == nil`. But, we can skip that cause we "know" it's not, based on the impl of the fn that creates it

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,11 +95,96 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
 		NextPageToken: nextToken,
 	}
-	return listRes, nil
+
+	return listResWithDigest(s.flatDigestEnabled, listRes, req, gateway, apnsByName, apnResourcesByAPN), nil
+}
+
+// listResWithDigest returns a ListSubscribersResponse with the fields augmented
+// according to the flat digest logic.
+func listResWithDigest(
```
hcgatewood: There's a less messy way of doing this -- just have a fn that returns (digest, no_updates) or something like that (and takes into account the feature flag -- which you can get from the method receiver rather than as an arg). Then can place those values into the proto from the upper fn

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -89,11 +95,96 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+
+			index := subProto.Sid.Id
+			subProtosById[index] = subProto
+		}
+
+		if nextPagetoken == "" {
+			break
+		}
+		curPageToken = nextPagetoken
+	}
```
hcgatewood: This logic is redundant to the per-page logic, coalesce it into a fn (which we'll then also use in the subscriberdb_cache service)

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer_test.go
```diff
@@ -41,7 +41,7 @@ func TestSubscriberdbCloudServicer(t *testing.T) {
 	lte_test_init.StartTestService(t)
 	configurator_test_init.StartTestService(t)
 
-	servicer := servicers.NewSubscriberdbServicer()
+	servicer := servicers.NewSubscriberdbServicer(false)
```
hcgatewood: Nit: passing unadorned booleans is considered an anti-pattern, especially to `new` functions. Consider passing e.g. `SubscriberdbServicerConfig` struct whose only field is that feature flag

Actually, can probably just pass the `Config` object directly that you defined to hold the feature flags. Abstractly it's a bit incongruent but I think it works. Either way, up to you

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer_test.go
```diff
@@ -295,8 +295,116 @@ func TestSubscriberdbCloudServicer(t *testing.T) {
 	assert.Equal(t, expectedToken, res.NextPageToken)
 }
 
+func TestSubscriberdbCloudServicerWithDigest(t *testing.T) {
```
hcgatewood: "Cloud" is redundant since this is `lte/cloud/go/...`

lte/protos/subscriberdb.proto
```diff
@@ -252,16 +252,30 @@ service SubscriberDBCloud {
   rpc ListSubscribers (ListSubscribersRequest) returns (ListSubscribersResponse) {}
 }
 
+message SubscriberDigest {
```
hcgatewood:
```suggestion
message SubscribersDigest {
```

lte/protos/subscriberdb.proto
```diff
@@ -252,16 +252,30 @@ service SubscriberDBCloud {
...
 message ListSubscribersResponse {
   repeated SubscriberData subscribers = 1;
   // next_page_token is a serialized entity page token for subsequent paginated
   // loads.
   string next_page_token = 2;
+  // digest is the server-defined, deterministic digest of the full set of
+  // returned subscriber objects.
+  SubscriberDigest digest = 3;
+  // no_updates is true if client's existing subscribers match those on the
+  // server. The subscribers array will be empty if no_updates = true.
```
hcgatewood:
```suggestion
  // server. If true, the subscribers array will be empty.
```

orc8r/cloud/go/mproto/mproto.go
```diff
@@ -14,14 +14,31 @@
...
+func HashManyDeterministic(protosByID map[string]proto.Message) (string, error) {
+	digest, err := marshalManyDeterministic(protosByID)
+	if err != nil {
+		return "", err
+	}
+	// Convert to a constant-length hash of the encoded value.
+	sum := md5.Sum(digest)
+	digestHash := base64.StdEncoding.EncodeToString(sum[:])
+
+	return digestHash, nil
```
hcgatewood:
```suggestion
	digest := base64.StdEncoding.EncodeToString(sum[:])

	return digest, nil
```

orc8r/cloud/go/mproto/mproto.go
```diff
@@ -14,14 +14,31 @@
...
+	"encoding/base64"
+
 	"magma/orc8r/cloud/go/mproto/protos"
 
 	"github.com/golang/protobuf/proto"
 )
 
-// MarshalManyDeterministic deterministically encodes a slice of protobuf messages,
+// HashManyDeterministic takes a collection of proto messages and returns a deterministic
+// base64-encoded md5 hash of the messages.
```
hcgatewood, line 26: Nit for uniformity
```suggestion
// base64-encoded md5 digest of the messages.
```

orc8r/cloud/go/mproto/testdata/determinisic-digest.b64.golden
```diff
@@ -1 +1 @@
-CpIBCgJjMRKLAQoCYzESCQoFMTIzNDUQChoSCgUxMjM0NRIJCgUxMjM0NRAKGhIKBTIzNDU2EgkKBTIzNDU2EA8aEgoFMzQ1NjcSCQoFMzQ1NjcQFBoSCgU0NTY3OBIJCgU0NTY3OBAZIgkKBTY3ODkwECMiCQoFNTY3ODkQHiIJCgU0NTY3OBAZIgkKBTM0NTY3EBQKcgoCYzISbAoCYzISCQoFMjM0NTYQDxoSCgUzNDU2NxIJCgUzNDU2NxAUGhIKBTQ1Njc4EgkKBTQ1Njc4EBkaEgoFNTY3ODkSCQoFNTY3ODkQHiIJCgU0NTY3OBAZIgkKBTM0NTY3EBQiCQoFMjM0NTYQDwpTCgJjMxJNCgJjMxIJCgUzNDU2NxAUGhIKBTU2Nzg5EgkKBTU2Nzg5EB4aEgoFNjc4OTASCQoFNjc4OTAQIyIJCgUyMzQ1NhAPIgkKBTEyMzQ1EAo=
\ No newline at end of file
+uG3FbxZiriDDbA2oz+UFjw==
```
hcgatewood, line 1: 👍

hcgatewood (changes requested):

orc8r/cloud/go/mproto/mproto.go
wangyyt1013: Changed the initial variable name to `digestBytes` to avoid redeclaring the same variable as a different type @hcgatewood

orc8r/cloud/go/mproto/mproto.go
```diff
@@ -14,14 +14,30 @@
...
+func HashManyDeterministic(protosByID map[string]proto.Message) (string, error) {
+	digestBytes, err := marshalManyDeterministic(protosByID)
+	if err != nil {
+		return "", err
+	}
+	// Convert to a constant-length hash of the encoded value.
+	sum := md5.Sum(digestBytes)
```
hcgatewood: "digestBytes" isn't a digest yet -- it's just the marshaled bytes. So can call it "marshaled" or "bytes"

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -80,20 +75,107 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
 		return nil, err
 	}
 
+	subProtos, nextToken, err := loadSubProtosPage(req.PageSize, req.PageToken, networkID, apnsByName, apnResourcesByAPN)
```
hcgatewood, line 78: 👍

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -80,20 +75,107 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+	if err != nil {
+		return nil, err
+	}
+
+	digest, noUpdates := s.getDigestAndNoUpdates(req, networkID, apnsByName, apnResourcesByAPN)
+	if noUpdates {
+		subProtos = []*lte_protos.SubscriberData{}
+	}
```
hcgatewood: 👍

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -80,20 +75,107 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+		Subscribers:   subProtos,
+		NextPageToken: nextToken,
+		Digest:        digest,
+		NoUpdates:     noUpdates,
+	}
+
+	return listRes, nil
+}
+
+func (s *subscriberdbServicer) getDigestAndNoUpdates(
```
hcgatewood: Nit: I don't like names with "and" in them. Just add the return values to a docstring for the method

```suggestion
func (s *subscriberdbServicer) getDigestInfo(
```

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -80,20 +75,107 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+	req *lte_protos.ListSubscribersRequest,
+	networkID string,
+	apnsByName map[string]*lte_models.ApnConfiguration,
+	apnResourcesByAPN lte_models.ApnResources,
+) (*lte_protos.SubscribersDigest, bool) {
+
+	digest, noUpdates := req.Digest, false
```
hcgatewood:
```suggestion
) (*lte_protos.SubscribersDigest, bool) {
	digest, noUpdates := req.Digest, false
```

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -80,20 +75,107 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+	// This functionality is currently placed behind a feature flag.
+	if s.flatDigestEnabled && req.PageToken == "" {
+		digestString, err := getDigest(networkID, apnsByName, apnResourcesByAPN)
```
hcgatewood:
```suggestion
		digest, err := getDigest(networkID, apnsByName, apnResourcesByAPN)
```

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -80,20 +75,107 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+		// If digest generation fails, the error is swallowed to not affect the main functionality.
+		if err != nil {
+			glog.Errorf("Generating digest for subscribers in network of gateway %s failed", networkID)
+		} else {
+			digest = &lte_protos.SubscribersDigest{Md5Base64Digest: digestString}
```
hcgatewood:
```suggestion
			digestProto = &lte_protos.SubscribersDigest{Md5Base64Digest: digestString}
```

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -80,20 +75,107 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+func (s *subscriberdbServicer) getDigestAndNoUpdates(
+	req *lte_protos.ListSubscribersRequest,
+	networkID string,
+	apnsByName map[string]*lte_models.ApnConfiguration,
+	apnResourcesByAPN lte_models.ApnResources,
+) (*lte_protos.SubscribersDigest, bool) {
+
+	digest, noUpdates := req.Digest, false
+	// This functionality is currently placed behind a feature flag.
+	if s.flatDigestEnabled && req.PageToken == "" {
```
hcgatewood: Flip this if -- instead of nesting the functionality inside the if block, use an early return

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -80,20 +75,107 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+			digest = &lte_protos.SubscribersDigest{Md5Base64Digest: digestString}
+			noUpdates = req.Digest.Md5Base64Digest == digestString
+		}
+	}
+
+	return digest, noUpdates
+}
+
+// getDigest loads all subscribers registered on the current network, and returns
+// a deterministic subscriber flat digest.
```
hcgatewood:
```suggestion
// getDigest returns a deterministic digest of all subscribers in the network.
// Digest is generated by loading all subscribers in the network.
```

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -80,20 +75,107 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+	subProtosById := map[string]proto.Message{}
+	curToken := ""
+	for {
+		subProtos, nextToken, err := loadSubProtosPage(0, curToken, networkID, apnsByName, apnResourcesByAPN)
+		if err != nil {
+			return "", err
+		}
+		for _, subProto := range subProtos {
+			index := subProto.Sid.Id
+			subProtosById[index] = subProto
```
hcgatewood: Coalesce these lines

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -80,20 +75,107 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
+// getDigest loads all subscribers registered on the current network, and returns
+// a deterministic subscriber flat digest.
+func getDigest(
+	networkID string,
+	apnsByName map[string]*lte_models.ApnConfiguration,
+	apnResourcesByAPN lte_models.ApnResources,
+) (string, error) {
+	subProtosById := map[string]proto.Message{}
+	curToken := ""
+	for {
```
hcgatewood: Nit: generally speaking I prefer not using while loops, as the exit condition is implicit / requires reading the whole block to understand

That said, I get that this is basically a "do-for" loop, so this isn't a bad way of handling it. Maybe try a `for curToken != "" || isFirstLoop)`

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer_test.go
```diff
@@ -295,8 +296,116 @@ func TestSubscriberdbCloudServicer(t *testing.T) {
 	assert.Equal(t, expectedToken, res.NextPageToken)
 }
 
+func TestSubscriberdbServicerWithDigest(t *testing.T) {
+	lte_test_init.StartTestService(t)
+	configurator_test_init.StartTestService(t)
+
+	// Create servicer with flat digest feature flag turned on.
```
hcgatewood: Nit: be consistent about whether you place a period at the end of your comments

The general rule of thumb in Orc8r code is "terminate with period" for docstrings and "don't terminate with period" for inline comments

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer_test.go
```diff
@@ -295,8 +296,116 @@ func TestSubscriberdbCloudServicer(t *testing.T) {
...
+		{
+			Sid:        &lte_protos.SubscriberID{Id: "99999", Type: lte_protos.SubscriberID_IMSI},
+			Lte:        &lte_protos.LTESubscription{State: lte_protos.LTESubscription_ACTIVE, AuthKey: []byte{}},
+			Non_3Gpp:   &lte_protos.Non3GPPUserProfile{ApnConfig: []*lte_protos.APNConfiguration{}},
+			NetworkId:  &protos.NetworkID{Id: "n1"},
+			SubProfile: "default",
+		},
+	}
+
+	req := &lte_protos.ListSubscribersRequest{
```
hcgatewood:
```suggestion
	}
	req := &lte_protos.ListSubscribersRequest{
```

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer_test.go
```diff
@@ -295,8 +296,116 @@ func TestSubscriberdbCloudServicer(t *testing.T) {
...
+		{
+			Sid:        &lte_protos.SubscriberID{Id: "11111", Type: lte_protos.SubscriberID_IMSI},
+			Lte:        &lte_protos.LTESubscription{State: lte_protos.LTESubscription_ACTIVE, AuthKey: []byte{}},
+			Non_3Gpp:   &lte_protos.Non3GPPUserProfile{ApnConfig: []*lte_protos.APNConfiguration{}},
+			NetworkId:  &protos.NetworkID{Id: "n1"},
+			SubProfile: "default",
+		},
+	}, expectedProtos...)
+
+	req = &lte_protos.ListSubscribersRequest{
```
hcgatewood:
```suggestion
	}, expectedProtos...)
	req = &lte_protos.ListSubscribersRequest{
```

lte/cloud/go/services/subscriberdb/subscriberdb/main.go
```diff
@@ -54,11 +54,15 @@ func main() {
 		glog.Fatalf("Error initializing IP lookup storage: %v", err)
 	}
 
+	// Load service configs
```
hcgatewood: Nit: I know Orc8r codebase does a bad job of this, but "topic sentence" comments are a bad code smell https://magma.github.io/magma/docs/next/contributing/contribute_conventions#comment-with-why-not-what

lte/protos/subscriberdb.proto
```diff
@@ -208,16 +208,30 @@ service SubscriberDBCloud {
   rpc ListSubscribers (ListSubscribersRequest) returns (ListSubscribersResponse) {}
 }
 
+message SubscribersDigest {
+  // Base64 encoded MD5 digest of the full list of subscriber data
+  string md5_base64_digest = 1;
```
hcgatewood: https://magma.github.io/magma/docs/next/contributing/contribute_conventions#follow-style-conventions

```suggestion
  // md5_base64_digest is a base64-encoded MD5 digest of a set of subscriber protos.
  string md5_base64_digest = 1;
```

hcgatewood (approved):
Looks great! Just a few cosmetic requests, then let's go ahead and merge

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
wangyyt1013: Will be using the `curToken = ""; nextToken = "start"; for nextToken != ""` strategy for this one @hcgatewood

themarwhal (approved):

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -95,51 +95,51 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
...
-			index := subProto.Sid.Id
-			subProtosById[index] = subProto
-		}
-		if nextToken == "" {
-			break
+			subProtosById[subProto.Sid.Id] = subProto
 		}
-		curToken = nextToken
+		curToken = token
+		nextToken = token
```
hcgatewood: I don't know about the "nextToken" pattern, using "init" is a little unclear

Try something like

```
foundEmptyToken := false
for !foundEmptyToken {
	// ...
	foundEmptyToken = nextToken == "" 
}
```

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
electronjoe: Late to the party but also wanted to point out - I think the GetMessage() calls return a "default constructed" object if nil? Is that right? And I personally would far prefer nil and the knowledge that a field wasn't set - to a default-constructed object that is perhaps misleading or erroneous and shouldn't be operated on.

PR #7314, opened 2021-06-03
[NMS]Made Gateways register only unregistered enodeB's

ato-dani:
Signed-off-by: Dani Menewuyelet <dmenewuy@fb.com>

## Summary

In the GatewayDetailConfigEdit.js file, I made the ran edit component to filter out an already registered eNodeB's under a different gateway. This way, when trying to register eNodeBs, it now shows only unregistered eNodeBs. Therefore it makes sure that an eNodeB can not be registered to more than one Gateway.

## Test Plan

Tested it by creating eNodeBs and access gateways and registering the eNodeBs to some access gateway and seeing what happens when trying to register the same eNodeB to another gateway. Video is provided. Also, I ran the previous tests that were written and made sure it is compatible.

## Additional Information

unregistered_eNodeB_test_part_1
unregistered_eNodeB_test_part_2

### DIFF
--- a/nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
+++ b/nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
@@ -72,7 +72,7 @@ import {
 } from '../../components/GatewayUtils';
 import {colors, typography} from '../../theme/default';
 import {makeStyles} from '@material-ui/styles';
-import {useContext, useEffect, useState} from 'react';
+import {useCallback, useContext, useEffect, useState} from 'react';
 import {useEnqueueSnackbar} from '@fbcnms/ui/hooks/useSnackbar';
 import {useRouter} from '@fbcnms/ui/hooks';
 
@@ -782,7 +782,7 @@ export function RanEdit(props: Props) {
   const handleDnsChange = (key: string, val) => {
     setDnsConfig({...dnsConfig, [key]: val});
   };
-  const isEnodebUnregistered = React.useCallback(
+  const isEnodebUnregistered = useCallback(
     (enb: enodeb, currentGateway: lte_gateway) => {
       const gatewaysList = Object.keys(ctx.state);
       for (const gatewayId of gatewaysList) {
--- a/nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
+++ b/nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
@@ -72,7 +72,7 @@ import {
 } from '../../components/GatewayUtils';
 import {colors, typography} from '../../theme/default';
 import {makeStyles} from '@material-ui/styles';
-import {useCallback, useContext, useEffect, useState} from 'react';
+import {useContext, useEffect, useState} from 'react';
 import {useEnqueueSnackbar} from '@fbcnms/ui/hooks/useSnackbar';
 import {useRouter} from '@fbcnms/ui/hooks';
 
@@ -782,8 +782,8 @@ export function RanEdit(props: Props) {
   const handleDnsChange = (key: string, val) => {
     setDnsConfig({...dnsConfig, [key]: val});
   };
-  const isEnodebUnregistered = useCallback(
-    (enb: enodeb, currentGateway: lte_gateway) => {
+  useEffect(() => {
+    const isEnodebUnregistered = (enb: enodeb, currentGateway: lte_gateway) => {
       const gatewaysList = Object.keys(ctx.state);
       for (const gatewayId of gatewaysList) {
         if (
@@ -794,10 +794,7 @@ export function RanEdit(props: Props) {
         }
       }
       return true;
-    },
-    [ctx?.state],
-  );
-  useEffect(() => {
+    };
     const newUnregisteredEnodebs = {};
     if (enbsCtx?.state?.enbInfo) {
       Object.keys(enbsCtx.state.enbInfo).map(enbSerial => {
@@ -808,7 +805,7 @@ export function RanEdit(props: Props) {
       });
     }
     setUnregisteredEnodebs(newUnregisteredEnodebs);
-  }, [enbsCtx?.state?.enbInfo, isEnodebUnregistered, props?.gateway]);
+  }, [ctx?.state, enbsCtx?.state?.enbInfo, props?.gateway]);
   const onSave = async () => {
     try {
       const gateway = {
--- a/nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
+++ b/nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
@@ -775,7 +775,7 @@ export function RanEdit(props: Props) {
   const [connectedEnodebs, setConnectedEnodebs] = useState<enodeb_serials>(
     props.gateway.connected_enodeb_serials,
   );
-  const [unregisteredEnodebs, setUnregisteredEnodebs] = useState({});
+  const [unregisteredEnbsInfo, setUnregisteredEnbsInfo] = useState([]);
   const handleRanChange = (key: string, val) => {
     setRanConfig({...ranConfig, [key]: val});
   };
@@ -795,16 +795,22 @@ export function RanEdit(props: Props) {
       }
       return true;
     };
-    const newUnregisteredEnodebs = {};
+    const newUnregisteredEnbsInfo = [];
     if (enbsCtx?.state?.enbInfo) {
       Object.keys(enbsCtx.state.enbInfo).map(enbSerial => {
-        newUnregisteredEnodebs[enbSerial] = isEnodebUnregistered(
-          enbsCtx.state.enbInfo[enbSerial].enb,
-          props.gateway,
-        );
+        if (
+          isEnodebUnregistered(
+            enbsCtx.state.enbInfo[enbSerial].enb,
+            props.gateway,
+          )
+        ) {
+          newUnregisteredEnbsInfo.push({
+            [enbSerial]: enbsCtx.state.enbInfo[enbSerial],
+          });
+        }
       });
     }
-    setUnregisteredEnodebs(newUnregisteredEnodebs);
+    setUnregisteredEnbsInfo(newUnregisteredEnbsInfo);
   }, [ctx?.state, enbsCtx?.state?.enbInfo, props?.gateway]);
   const onSave = async () => {
     try {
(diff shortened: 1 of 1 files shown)


nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
```diff
@@ -808,7 +809,21 @@ export function RanEdit(props: Props) {
       setError(e.response?.data?.message ?? e.message);
     }
   };
-
+  const isEnodebUnregistered = React.useCallback(
```
karthiksubraveti: @MightyDani Basic question why is this a callback and not a simple function ?

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
```diff
@@ -808,7 +809,21 @@ export function RanEdit(props: Props) {
...
+    (enb: enodeb, currentGateway: lte_gateway) => {
+      const gatewaysList = Object.keys(ctx.state);
+      for (const gatewayId of gatewaysList) {
```
karthiksubraveti: why iterate over gateways when we already have the lte_gateway
```
export type lte_gateway = {
    apn_resources ? : apn_resources,
    cellular: gateway_cellular_configs,
    connected_enodeb_serials: enodeb_serials,
    description: gateway_description,
    device: gateway_device,
    id: gateway_id,
    magmad: magmad_gateway_configs,
    name: gateway_name,
    status ? : gateway_status,
    tier: tier_id,
};
```

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
```diff
@@ -855,15 +870,28 @@ export function RanEdit(props: Props) {
...
-                    <Checkbox checked={connectedEnodebs.includes(enbSerial)} />
-                    <ListItemText
-                      primary={enbsCtx.state.enbInfo[enbSerial].enb.name}
-                      secondary={enbSerial}
-                    />
-                  </MenuItem>
-                ))}
+                Object.keys(enbsCtx.state.enbInfo).map(enbSerial => {
+                  if (
+                    isEnodebUnregistered(
```
karthiksubraveti: I think we can eliminate this function and simply have a straightforward check here
`props.gateway.connected_enodeb_serials?.includes(enb.serial)`. Am I missing anything ?

karthiksubraveti (commented):
Thanks for the changes @MightyDani . Left a few comments.

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
ato-dani: Andrei asked me to call this function only when the state changed instead of calling it every time. You can see his comments on the original peer request I made.

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
ato-dani: Your logic is used when checking if the current gateway includes the enodeb and put a checkmark on it if so. However, the function(the one I wrote) is used to make sure that when displaying list of enodebs to be registered to a gateway, we only list enodebs that are not registered by another gateway. In short this function makes sure that an enodeb can not be registered by more than one access gateway. So, if the other gateways do not include the enodeb serial, then that means it is free to be registered since it is not registered by other access gateway. But if we are just checking if the current gateway includes the enodeb serial and list those, then that means we are only listing enodebs that are already registered under the current gateway(and not listing the free/unregistered ones)

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
karthiksubraveti: Thanks for clarification. I missed that point. If i understand useCallback correctly, it only prevents a new function from getting created, however it is going to return the function instance for the same ctx.state, so we will end up calling the function eitherways no ? 
My suggestion is to add a useEffect(with dependencies on gateway and enodeb context) and pre populate unregistered eNodeB state. We can then simply use that in that loop. The above approach works as well, but we might be running  (number of enodeb times * number of gateways) on every render.

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
karthiksubraveti: Sounds good. Thanks for the clarification

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
ato-dani: Sorry, I didn't explain it the first time well. The useCallback helps us do memoization. So, the first time it is called, it does all the checking and renders it. However, after the first call, when we call it second time or any number of times, since it already memoized it, it knows and returns the values as long as the the state didn't change. Does that makes sense?

MightyDani: Just to give you some option, I did the useEffect(just as what you suggested) at the start and gave it a dependency of gateway id and the connected enodeb serials because if any of the 2 changed we need to redo it again. I tested it and the functionality worked. If you don't want to go with the memoization technique just reply to me and I will push the new changes I made.  
Screen Shot 2021-06-04 at 11 54 47 AM
Screen Shot 2021-06-04 at 11 55 26 AM
[git_diff.txt](https://github.com/magma/magma/files/6599742/git_diff.txt)

MightyDani: I also did the suggestions you made and wrote a useEffect function which has a dependency list containing the gateway id and the connected enodeb serials for that gateway id since the useEffect needs to run again whenever one of the two things changes. If you would want to go with the useEffect method instead of the memoization technique, just call me and I will push the changes.
Screen Shot 2021-06-04 at 11 54 47 AM
Screen Shot 2021-06-04 at 11 55 26 AM
[git_diff.txt](https://github.com/magma/magma/files/6599969/git_diff.txt)

ato-dani: I have pushed a commit which uses useeffect and precompute the values. It works well and run that expensive function only once.

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
```diff
@@ -774,13 +775,40 @@ export function RanEdit(props: Props) {
...
   );
+  const [unregisteredEnodebs, setUnregisteredEnodebs] = useState({});
   const handleRanChange = (key: string, val) => {
     setRanConfig({...ranConfig, [key]: val});
   };
   const handleDnsChange = (key: string, val) => {
     setDnsConfig({...dnsConfig, [key]: val});
   };
-
+  const isEnodebUnregistered = React.useCallback(
```
andreilee: At the top of the file import this with the other hooks:
```
import {useCallback, useContext, useEffect, useState} from 'react';
```

andreilee (approved):

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
ato-dani: Just did that. Sorry for the late response, didn't see this comment.

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
```diff
@@ -774,13 +775,40 @@ export function RanEdit(props: Props) {
...
+    if (enbsCtx?.state?.enbInfo) {
+      Object.keys(enbsCtx.state.enbInfo).map(enbSerial => {
+        newUnregisteredEnodebs[enbSerial] = isEnodebUnregistered(
+          enbsCtx.state.enbInfo[enbSerial].enb,
+          props.gateway,
+        );
+      });
+    }
+    setUnregisteredEnodebs(newUnregisteredEnodebs);
+  }, [enbsCtx?.state?.enbInfo, isEnodebUnregistered, props?.gateway]);
```
karthiksubraveti: you can move isEnodebUnregistered function inside useEffect and eliminate useCallback

karthiksubraveti (approved):

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
ato-dani: I just did that.

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
```diff
@@ -774,13 +775,37 @@ export function RanEdit(props: Props) {
   const [connectedEnodebs, setConnectedEnodebs] = useState<enodeb_serials>(
     props.gateway.connected_enodeb_serials,
   );
+  const [unregisteredEnodebs, setUnregisteredEnodebs] = useState({});
```
karthiksubraveti: minor : This can be a newUnregisteredEnbInfo and can be a list

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
```diff
@@ -774,13 +775,37 @@ export function RanEdit(props: Props) {
...
+        ) {
+          return false;
+        }
+      }
+      return true;
+    };
+    const newUnregisteredEnodebs = {};
+    if (enbsCtx?.state?.enbInfo) {
+      Object.keys(enbsCtx.state.enbInfo).map(enbSerial => {
+        newUnregisteredEnodebs[enbSerial] = isEnodebUnregistered(
```
karthiksubraveti: You can make the map method return the list of unregisteredEnbInfo and directly use it as 
`Object.keys(unregisteredEnbInfo).map(enbSerial =`

karthiksubraveti (approved):

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
ato-dani: I just did that. I mapped through unregisteredEnbsInfo and show only unregistered enodebs and removed the if else logic.

nms/app/packages/magmalte/app/views/equipment/GatewayDetailConfigEdit.js
ato-dani: Done.

PR #7317, opened 2021-06-03
[NMS] Made the subscriber add button to now call the new bulk add endpoint instead of adding subscribers singly

ato-dani:
## Summary
I changed the subscribers add logic to call the new API endpoint which allows us to add more than one subscriber at a time. So, when a user uploads a csv file to add a list of subscribers, it now adds all of the subscribers calling the API endpoint only once instead of calling the API one by one for each to be added subscriber. To do this, I also changed the subscriber's value property to now also accept the mutable_subscribers type and also regenerated the magma api bindings file to use the new API end point. 

## Test Plan
I edited the existing subscribers add test which used to call the previous API endpoint that used to add only one subscriber at a time, to the now call the new endpoint. This worked without errors. There is also a gif and some photos which shows that the add button works properly.

## Additional Information

Screen Shot 2021-06-03 at 7 03 20 PM
bulkadd

https://user-images.githubusercontent.com/34602743/120723663-2d5fda00-c4a0-11eb-9098-c5d72c038449.mov

### DIFF
--- a/nms/app/packages/magmalte/app/components/context/SubscriberContext.js
+++ b/nms/app/packages/magmalte/app/components/context/SubscriberContext.js
@@ -16,6 +16,7 @@
 import type {
   gateway_id,
   mutable_subscriber,
+  mutable_subscribers,
   subscriber,
   subscriber_id,
   subscriber_state,
@@ -42,7 +43,7 @@ export type SubscriberContextType = {
   gwSubscriberMap: {[gateway_id]: Array<subscriber_id>},
   setState?: (
     key: string,
-    val?: mutable_subscriber,
+    val?: mutable_subscriber | mutable_subscribers,
     newState?: {[string]: subscriber},
     newSessionState?: {[string]: subscriber_state},
   ) => Promise<void>,
--- a/nms/app/packages/magmalte/app/state/lte/SubscriberState.js
+++ b/nms/app/packages/magmalte/app/state/lte/SubscriberState.js
@@ -20,6 +20,7 @@ import type {SubscriberContextType} from '../../components/context/SubscriberCon
 import type {SubscriberRowType} from '../../views/subscriber/SubscriberOverview';
 import type {
   mutable_subscriber,
+  mutable_subscribers,
   network_id,
   subscriber,
   subscriber_state,
@@ -187,7 +188,7 @@ type SubscriberStateProps = {
   setSubscriberMap: ({[string]: subscriber}) => void,
   setSessionState: ({[string]: subscriber_state}) => void,
   key: string,
-  value?: mutable_subscriber,
+  value?: mutable_subscriber | mutable_subscribers,
   newState?: {[string]: subscriber},
   newSessionState?: {[string]: subscriber_state},
 };
@@ -210,6 +211,18 @@ export async function setSubscriberState(props: SubscriberStateProps) {
   if (newSessionState) {
     // $FlowIgnore
     setSessionState(newSessionState.sessionState);
+    return;
+  }
+  if (Array.isArray(value)) {
+    await MagmaV1API.postLteByNetworkIdSubscribersV2({
+      networkId,
+      subscribers: value,
+    });
+    const newSubscriberMap = {};
+    value.map(newSubscriber => {
+      newSubscriberMap[newSubscriber.id] = newSubscriber;
+    });
+    setSubscriberMap({...subscriberMap, newSubscriberMap});
     return;
   }
   if (value != null) {
--- a/nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
+++ b/nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
@@ -13,7 +13,7 @@
  * @flow strict-local
  * @format
  */
-import type {subscriber} from '@fbcnms/magma-api';
+import type {mutable_subscribers, subscriber} from '@fbcnms/magma-api';
 
 import ActionTable from '../../components/ActionTable';
 import ApnContext from '../../components/context/ApnContext';
@@ -42,7 +42,11 @@ import Tabs from '@material-ui/core/Tabs';
 import TypedSelect from '@fbcnms/ui/components/TypedSelect';
 import nullthrows from '@fbcnms/util/nullthrows';
 
-import {AltFormField, PasswordInput} from '../../components/FormField';
+import {
+  AltFormField,
+  LinearProgressWithLabel,
+  PasswordInput,
+} from '../../components/FormField';
 import {SelectEditComponent} from '../../components/ActionTable';
 import {base64ToHex, hexToBase64, isValidHex} from '@fbcnms/util/strings';
 import {colors, typography} from '../../theme/default';
@@ -112,6 +116,7 @@ const MAX_UPLOAD_FILE_SZ_BYTES = 10 * 1024 * 1024;
 const SUBSCRIBER_TITLE = 'Subscriber';
 const TRAFFIC_TITLE = 'Traffic Policy';
 const STATIC_IPS_TITLE = 'APNs Static IPs';
+const SUBSCRIBERS_CHUNK_SIZE = 1000;
 
 type SubscriberInfo = {
   name: string,
@@ -490,7 +495,6 @@ export function SubscriberEditDialog(props: DialogProps) {
     </Dialog>
   );
 }
-
 function AddSubscriberDetails(props: DialogProps) {
   const classes = useStyles();
   const ctx = useContext(SubscriberContext);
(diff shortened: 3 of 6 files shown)


nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -547,8 +553,13 @@ function AddSubscriberDetails(props: DialogProps) {
...
             sub_profile: subscriber.dataPlan,
           },
-        });
+        };
         successCountRef.current = successCountRef.current + 1;
+        addedSubscribers.push(newSubscriber);
+        //bulk add subscribers at the end
+        if (subscriberIdx == subscribers.length - 1) {
+          await ctx.setState?.('', addedSubscribers);
+        }
```
andreilee: This should sit outside of the for loop. The resulting error message if this fails should also be difference, since it's a bulk save. Make sure to revert the for loop as well.

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -520,7 +520,13 @@ function AddSubscriberDetails(props: DialogProps) {
   );
   const saveSubscribers = async () => {
```
andreilee, line 540: What happens if a csv is uploaded with some new subscribers, and some edited subscribers?

andreilee (dismissed):

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
ato-dani: It will reject the already existing subscribers since this is an add dialog, and it will set error for those specific edited subscriber because it first loops and validates the subscriber info at line 531. Also, the endpoint API that we are calling here would also return error, if we try to edit an existing subscriber. So, I made sure to just set the error for an edited subscriber and not add it to the list of subscribers that will be sent to the endpoint to make sure that the whole operation doesn't fail.

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -520,7 +520,13 @@ function AddSubscriberDetails(props: DialogProps) {
...
     successCountRef.current = 0;
-    for (const subscriber of subscribers) {
+    const addedSubscribers = [];
+    for (
+      let subscriberIdx = 0;
+      subscriberIdx < subscribers.length;
+      ++subscriberIdx
+    ) {
```
andreilee: You should be able to revert this to `for (const subscriber of subscribers) {`

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
ato-dani: Done.

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
ato-dani: Done.

andreilee:
@karthiksubraveti Could you review this as well?

nms/app/packages/magmalte/app/components/context/SubscriberContext.js
```diff
@@ -42,7 +43,7 @@ export type SubscriberContextType = {
   gwSubscriberMap: {[gateway_id]: Array<subscriber_id>},
   setState?: (
     key: string,
-    val?: mutable_subscriber,
+    val?: mutable_subscriber | mutable_subscribers,
```
karthiksubraveti, line 46: Lets just use mutable_subscribers and pass in array of mutable subscriber in other locations to keep this consistent.

nms/app/packages/magmalte/app/state/lte/SubscriberState.js
```diff
@@ -212,6 +213,18 @@ export async function setSubscriberState(props: SubscriberStateProps) {
     setSessionState(newSessionState.sessionState);
     return;
   }
+  if (Array.isArray(value)) {
```
karthiksubraveti, line 216: Based on above comment, if value is always an array we can eliminate this check and unify the logic. If a user passes in empty array, then we can consider that as a delete option. We might also in implement bulk delete in orc8r.

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -547,14 +548,25 @@ function AddSubscriberDetails(props: DialogProps) {
             state: subscriber.state,
             sub_profile: subscriber.dataPlan,
           },
-        });
+        };
         successCountRef.current = successCountRef.current + 1;
```
karthiksubraveti: You can get rid of successCountRef

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -547,14 +548,25 @@ function AddSubscriberDetails(props: DialogProps) {
...
+    if (addedSubscribers.length > 0) {
+      try {
+        await ctx.setState?.('', addedSubscribers);
+      } catch (e) {
+        const errMsg = e.response?.data?.message ?? e.message ?? e;
+        setError('error saving subscribers to the api : ' + errMsg);
+        return;
+      }
+    }
     enqueueSnackbar(
```
karthiksubraveti, line 595: You can get rid of LinearProgressWithLabel since we don't know about the progress here.

karthiksubraveti (changes requested):
Left a few comments here. Thanks for making this change.

nms/app/packages/magmalte/app/components/context/SubscriberContext.js
ato-dani: I think it would not work because we only have bulk add api on the swagger api, apart from that the other ones like edit work on one subscriber at a time. Also, doing this change would lead to complications like if a user passes an array of some edited and some new added because the bulk add endpoint returns error for those cases. So, I think it is better to only allow an array to be passed when we are actually using it to call the bulk add endpoint.

ato-dani: Resolved after speaking with Karthik.

nms/app/packages/magmalte/app/state/lte/SubscriberState.js
ato-dani: Resolved after speaking with Karthik.

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
ato-dani: I think It shouldn't be removed because successCountRef is used in the Dialog content in calculating the value for LinearProgressWithLabel. Because, before we call the bulk add endpoint, we first verify each subscribers info and In case an error happens like for example a subscriber already exists then the successCountRef would help us calculate the percentage of how much we are done and how much we are left with. So, a user would correct the errors and finally when there is no error and the user presses the save button, the bulk api endpoint would be finally called. So, these change ensures that api endpoint is called lazily (only when everything is right) and saves us from wasting time by calling the api when we already know there is an error like a subscriber already exists. Does that sound right?

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
ato-dani: Instead of completely removing the error message, changed it to show the error on the Snackbar incase the communication with the API fails, so that the user knows the operation failed.

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
ato-dani: I have removed it

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
ato-dani: Completely removed Linear Progress and replaced by snackbar as per our chat

karthiksubraveti (approved):

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -547,11 +566,35 @@ function AddSubscriberDetails(props: DialogProps) {
             state: subscriber.state,
             sub_profile: subscriber.dataPlan,
           },
-        });
-        successCountRef.current = successCountRef.current + 1;
+        };
+        addedSubscribers.push(newSubscriber);
+        // bulk add chunked subscribers
+        if (addedSubscribers.length == SUBSCRIBERS_CHUNK_SIZE) {
```
karthiksubraveti: Suggestion: Change above for of loop to include a counter var
`for (const [i, subscriber] of subscribers.entries()) `

Include the below condition within this conditional by adding 
(addedSubscribers.length == SUBSCRIBERS_CHUNK_SIZE) || (i === subscribers.length - 1) {
 // bulk add
    }

ato-dani: Done.

PR #7330, opened 2021-06-04
[LTE][AGW] EMM data context _ms_network_capability attribute copy: wrong sizeof()

lionelgo:

## Summary

I changed   
memset( &ctxt->_ms_network_capability, 0, sizeof(ctxt->_ue_network_capability));
to
memset( &ctxt->_ms_network_capability, 0, sizeof(ctxt->_ms_network_capability));

so that attribute declared right after _ms_network_capability won't be overwritten when clearing _ms_network_capability value through emm_ctx_clear_ms_nw_cap().

## Test Plan

This bug was certainly modifying emm_context_t/_drx_parameter, should not now.


ulaskozat (approved):
LGTM

PR #7339, opened 2021-06-04
[mme] Remove AssertFatal NULL checks after itti_alloc_new_message calls

andreilee:

## Summary

A number of times after calling `itti_alloc_new_message(..)`, the return value is checked via `AssertFatal(..)` if it is non-NULL. This is redundant since `itti_alloc_new_message already checks if its result of `malloc` is non-NULL.

This PR is part of a series to rework `AssertFatal(..)` usage in Magma's MME service

## Test Plan

Only built MME service:

```
.
.
.
[1971/1971] Linking CXX executable oai_mme/mme
```

## Additional Information


ardzoht (approved):
lgtm

PR #7340, opened 2021-06-04
Update AGW to support the flat digest pattern, including storing and communicating digests

wangyyt1013:

## Summary

1. Add function in `subscriberdb/store/sqlite.py` to store and retrieve the current digest for the gateway
2. Add flat digest logic in `subscriberdb/client.py` to fetch, receive, and update digests, as well as skip subscriber data resyncing when `no_updates=true`
3. Add unit tests to cover added functionalities

## Test Plan

1. Run all existing agw python tests. All should pass
2. Add new unit tests in `subscriberdb/tests/store/store_tests.py` and `subscriberdb/tests/client_tests.py`
3. Locally verified flow with `subscriberdb` cloud in docker

## Additional Information

- This pr contains changes introduced in #7302

### DIFF
--- a/lte/gateway/python/magma/subscriberdb/client.py
+++ b/lte/gateway/python/magma/subscriberdb/client.py
@@ -23,7 +23,6 @@ from lte.protos.subscriberdb_pb2 import (
     ListSubscribersRequest,
     LTESubscription,
     SubscriberData,
-    SyncSubscribersRequest,
 )
 from magma.common.grpc_client_manager import GRPCClientManager
 from magma.common.rpc_utils import grpc_async_wrapper
@@ -84,15 +83,9 @@ class SubscriberDBCloudClient(SDWatchdogTask):
         if in_sync:
             return
 
-        resync = await self._sync_subscribers()
-        if not resync:
+        subscribers, flat_digest = await self._get_all_subscribers()
+        if subscribers is None:
             return
-
-        ret = await self._get_all_subscribers()
-        if ret is None:
-            return
-
-        subscribers, flat_digest = ret
         self._update_flat_digest(flat_digest)
         self._process_subscribers(subscribers)
 
@@ -126,41 +119,7 @@ class SubscriberDBCloudClient(SDWatchdogTask):
             return False
         return res.in_sync
 
-    async def _sync_subscribers(self) -> bool:
-        """
-        Query server-side set of per-subscriber digests and fetches the
-        subscriber data changeset
-
-        If the change is within a service-configurable size limit, updates the
-        local data with the returned values; if not, returns the signal for an
-        overall resync
-
-        Returns:
-            boolean value for whether a resync is required
-        """
-        subscriberdb_cloud_client = self._grpc_client_manager.get_client()
-        req = SyncSubscribersRequest()
-        try:
-            res = await grpc_async_wrapper(
-                subscriberdb_cloud_client.SyncSubscribers.future(
-                    req,
-                    self.SUBSCRIBERDB_REQUEST_TIMEOUT,
-                ),
-                self._loop,
-            )
-        except grpc.RpcError as err:
-            logging.error(
-                "Sync subscribers request error! [%s] %s", err.code(),
-                err.details(),
-            )
-            return True
-
-        if res.resync:
-            return True
-        # TODO (wangyyt1013): update susbcriber data according to changeset
-        return False
-
-    async def _get_all_subscribers(self) -> [SubscriberData, str]:
+    async def _get_all_subscribers(self) -> (SubscriberData, Digest):
         subscriberdb_cloud_client = self._grpc_client_manager.get_client()
         subscribers = []
         flat_digest = None
@@ -197,7 +156,7 @@ class SubscriberDBCloudClient(SDWatchdogTask):
                     time_elapsed.total_seconds() * 1000,
                 )
                 SUBSCRIBER_SYNC_FAILURE_TOTAL.inc()
-                return None
+                return None, None
         logging.info(
             "Successfully fetched all subscriber "
             "pages from the cloud!",
@@ -208,7 +167,7 @@ class SubscriberDBCloudClient(SDWatchdogTask):
             time_elapsed.total_seconds() * 1000,
         )
 
-        return [subscribers, flat_digest]
+        return subscribers, flat_digest
 
     def _update_flat_digest(self, flat_digest: Digest) -> None:
         if Digest is None:
--- a/lte/gateway/python/magma/subscriberdb/client.py
+++ b/lte/gateway/python/magma/subscriberdb/client.py
@@ -18,9 +18,12 @@ import grpc
 from lte.protos.s6a_service_pb2 import DeleteSubscriberRequest
 from lte.protos.s6a_service_pb2_grpc import S6aServiceStub
 from lte.protos.subscriberdb_pb2 import (
+    CheckSubscribersInSyncRequest,
+    Digest,
     ListSubscribersRequest,
     LTESubscription,
     SubscriberData,
+    SyncSubscribersRequest,
 )
 from magma.common.grpc_client_manager import GRPCClientManager
 from magma.common.rpc_utils import grpc_async_wrapper
@@ -77,32 +80,113 @@ class SubscriberDBCloudClient(SDWatchdogTask):
         self._grpc_client_manager = grpc_client_manager
 
     async def _run(self) -> None:
-        subscribers = await self._get_subscribers()
-        if subscribers is not None:
-            self._process_subscribers(subscribers)
+        in_sync = await self._check_subscribers_in_sync()
+        if in_sync:
+            return
 
-    async def _get_subscribers(self) -> [SubscriberData]:
+        resync = await self._sync_subscribers()
+        if not resync:
+            return
+
+        ret = await self._get_all_subscribers()
+        if ret is None:
+            return
+
+        subscribers, flat_digest = ret
+        self._update_flat_digest(flat_digest)
+        self._process_subscribers(subscribers)
+
+    async def _check_subscribers_in_sync(self) -> bool:
+        """
+        Check if the local subscriber data is up-to-date with the cloud by
+        comparing flat digests
+
+        Returns:
+            boolean value for whether the local data is in sync
+        """
+        subscriberdb_cloud_client = self._grpc_client_manager.get_client()
+        req = CheckSubscribersInSyncRequest(
+            flat_digest=Digest(
+                md5_base64_digest=self._store.get_current_digest(),
+            ),
+        )
+        try:
+            res = await grpc_async_wrapper(
+                subscriberdb_cloud_client.CheckSubscribersInSync.future(
+                    req,
+                    self.SUBSCRIBERDB_REQUEST_TIMEOUT,
+                ),
+                self._loop,
+            )
+        except grpc.RpcError as err:
+            logging.error(
+                "Check subscribers in sync request error! [%s] %s", err.code(),
+                err.details(),
+            )
+            return False
+        return res.in_sync
+
+    async def _sync_subscribers(self) -> bool:
+        """
+        Query server-side set of per-subscriber digests and fetches the
+        subscriber data changeset
+
+        If the change is within a service-configurable size limit, updates the
+        local data with the returned values; if not, returns the signal for an
+        overall resync
+
+        Returns:
+            boolean value for whether a resync is required
+        """
+        subscriberdb_cloud_client = self._grpc_client_manager.get_client()
+        req = SyncSubscribersRequest()
+        try:
+            res = await grpc_async_wrapper(
+                subscriberdb_cloud_client.SyncSubscribers.future(
+                    req,
+                    self.SUBSCRIBERDB_REQUEST_TIMEOUT,
+                ),
+                self._loop,
+            )
+        except grpc.RpcError as err:
+            logging.error(
+                "Sync subscribers request error! [%s] %s", err.code(),
+                err.details(),
+            )
+            return True
+
+        if res.resync:
+            return True
+        # TODO (wangyyt1013): update susbcriber data according to changeset
+        return False
+
+    async def _get_all_subscribers(self) -> [SubscriberData, str]:
         subscriberdb_cloud_client = self._grpc_client_manager.get_client()
         subscribers = []
+        flat_digest = None
         req_page_token = ""  # noqa: S105
-        res_page_token = "start_token"  # noqa: S105
+        found_empty_token = False
         sync_start = datetime.datetime.now()
-        while res_page_token != "":  # noqa: S105
+
+        # Next page token empty means read all updates
+        while not found_empty_token:  # noqa: S105
             try:
                 req = ListSubscribersRequest(
                     page_size=self._subscriber_page_size,
                     page_token=req_page_token,
                 )
-                response = await grpc_async_wrapper(
+                res = await grpc_async_wrapper(
                     subscriberdb_cloud_client.ListSubscribers.future(
                         req,
                         self.SUBSCRIBERDB_REQUEST_TIMEOUT,
                     ),
                     self._loop,
                 )
-                subscribers.extend(response.subscribers)
-                res_page_token = response.next_page_token
-                req_page_token = response.next_page_token
+                subscribers.extend(res.subscribers)
+                flat_digest = res.flat_digest
+                req_page_token = res.next_page_token
+                found_empty_token = req_page_token == ""
+
             except grpc.RpcError as err:
                 logging.error(
                     "Fetch subscribers error! [%s] %s", err.code(),
@@ -123,7 +207,13 @@ class SubscriberDBCloudClient(SDWatchdogTask):
         SUBSCRIBER_SYNC_LATENCY.observe(
             time_elapsed.total_seconds() * 1000,
         )
-        return subscribers
+
+        return [subscribers, flat_digest]
+
+    def _update_flat_digest(self, flat_digest: Digest) -> None:
+        if Digest is None:
+            return
+        self._store.update_digest(flat_digest.md5_base64_digest)
 
     def _process_subscribers(self, subscribers: SubscriberData) -> None:
         active_subscriber_ids = []
--- a/lte/gateway/python/magma/subscriberdb/store/sqlite.py
+++ b/lte/gateway/python/magma/subscriberdb/store/sqlite.py
@@ -15,6 +15,7 @@ import logging
 import sqlite3
 from collections import defaultdict
 from contextlib import contextmanager
+from datetime import datetime
 
 from lte.protos.subscriberdb_pb2 import SubscriberData
 from magma.subscriberdb.sid import SIDUtils
@@ -34,13 +35,13 @@ class SqliteStore(BaseStore):
     def __init__(self, db_location, loop=None, sid_digits=2):
         self._sid_digits = sid_digits  # last digits to be included from subscriber id
         self._n_shards = 10**sid_digits
-        self._db_locations = self._create_db_locations(
-            db_location, self._n_shards,
-        )
+        self._db_locations, self._digest_db_location = \
+            self._create_db_locations(db_location, self._n_shards)
         self._create_store()
         self._on_ready = OnDataReady(loop=loop)
 
-    def _create_db_locations(self, db_location, n_shards):
+    def _create_db_locations(self, db_location: str, n_shards: int) \
+            -> [list, str]:
         # in memory if db_location is not specified
         if not db_location:
             db_location = "/var/opt/magma/"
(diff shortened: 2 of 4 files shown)


lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +78,47 @@ def __init__(
         self._grpc_client_manager = grpc_client_manager
 
     async def _run(self) -> None:
-        subscribers = await self._get_subscribers()
-        if subscribers is not None:
+        subscribers, no_updates = await self._get_subscribers()
+        # No need to update local db if it's already up-to-date.
+        if (subscribers is not None) and (not no_updates):
             self._process_subscribers(subscribers)
```
hcgatewood: - Always try to make boolean logic as simple as possible. One tool in your toolset: https://en.wikipedia.org/wiki/De_Morgan%27s_laws
- Early returns are nice, especially when they allow for simpler boolean logic

```suggestion
        if subscribers is None or no_updates:
            return
        self._process_subscribers(subscribers)
```

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +78,47 @@ def __init__(
...
-                res_page_token = response.next_page_token
-                req_page_token = response.next_page_token
+                subscribers.extend(res.subscribers)
+                res_page_token = res.next_page_token
+                req_page_token = res.next_page_token
+                no_updates = res.no_updates
+
+                if cur_digest != res.digest:
+                    self._store.update_digest(res.digest.md5_base64_digest)
+                    cur_digest = res.digest
```
hcgatewood: Just do this once after completely reading all pages

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -59,11 +59,17 @@ def _create_db_locations(self, db_location, n_shards):
                 + ".db?cache=shared",
             )
             logging.info("db location: %s", db_location_list[shard])
-        return db_location_list
+
+        digest_db_location = 'file:' + db_location + \
```
hcgatewood, line 64: ?
```suggestion
        digest_db_location = 'file:' + digest_db_location + \
```

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -266,6 +283,40 @@ def resync(self, subscribers):
                 conn.close()
         self._on_ready.resync(subscribers)
 
+    def get_current_digest(self):
+        """
+        Method that returns the current subscriber digest stored in the db.
```
hcgatewood: "Method that" is redundant

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -266,6 +283,40 @@ def resync(self, subscribers):
...
+        """
+        conn = sqlite3.connect(self._digest_db_location, uri=True)
+        try:
+            with conn:
+                res = conn.execute(
+                    "SELECT * FROM subscriber_digest",
```
hcgatewood: As a general rule: don't use the blanket select (`SELECT *`) in application code except with specific reason -- (a) for readability and (b) won't break [if the columns ever got reordered](https://stackoverflow.com/questions/6805426/how-to-move-columns-in-a-mysql-table)

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -266,6 +283,40 @@ def resync(self, subscribers):
                 conn.close()
         self._on_ready.resync(subscribers)
 
+    def get_current_digest(self):
```
hcgatewood: Here and throughout your code: include Python type signatures

Our codebase doesn't do a good job of this yet, but that's because it was written before Python had type signatures 🙂 Python type signatures makes things significantly easier on future devs

lte/gateway/python/magma/subscriberdb/tests/client_tests.py
```diff
@@ -272,3 +280,38 @@ def test_detach_deleted_subscribers(self, s6a_service_mock_stub):
         mock.DeleteSubscriber.future.assert_called_once_with(
             DeleteSubscriberRequest(imsi_list=["101", "303"]),
         )
+
+    @ unittest.mock.patch(
+        'magma.common.service_registry.ServiceRegistry.get_rpc_channel',
+    )
```
hcgatewood, line 300: What does this do

lte/gateway/python/magma/subscriberdb/tests/client_tests.py
```diff
@@ -272,3 +280,38 @@ def test_detach_deleted_subscribers(self, s6a_service_mock_stub):
...
+            # subsequent requests.
+            cur_digest = self.subscriberdb_cloud_client._store.\
+                get_current_digest()
+            self.assertEqual("11111", cur_digest)
+
+            subscribers, no_updates = (
+                await self.subscriberdb_cloud_client._get_subscribers()
+            )
+            self.assertEqual(True, no_updates)
+            self.assertEqual([], subscribers)
```
hcgatewood: Also check that the digest hasn't been changed

lte/gateway/python/magma/subscriberdb/tests/client_tests.py
```diff
@@ -272,3 +280,38 @@ def test_detach_deleted_subscribers(self, s6a_service_mock_stub):
         mock.DeleteSubscriber.future.assert_called_once_with(
             DeleteSubscriberRequest(imsi_list=["101", "303"]),
         )
+
+    @ unittest.mock.patch(
+        'magma.common.service_registry.ServiceRegistry.get_rpc_channel',
+    )
+    def test_get_subscribers_no_updates(self, get_grpc_mock):
```
hcgatewood: Does this also test how things work with multiple pages?

hcgatewood (changes requested):
Looks good, couple fixes and couple questions

lte/gateway/python/magma/subscriberdb/store/sqlite.py
wangyyt1013: the `db_location` here refers to the directory where all db files are stored (defaulted to `/var/opt/magma/`). What is the reason why we shouldn't store the `subscriber-digest.db` file in the same directory? @hcgatewood

lte/gateway/python/magma/subscriberdb/store/sqlite.py
wangyyt1013: This entire doc uses this prefix for the method docstrings; should I go ahead and change all of the docstrings?

lte/gateway/python/magma/subscriberdb/tests/client_tests.py
wangyyt1013: I think this creates a local mock for the cloud service registry (hence all test cases that make rpc calls to the cloud use this)

lte/gateway/python/magma/subscriberdb/tests/client_tests.py
wangyyt1013: This test suite has a mock `ListSubscribers` for the test cases to work with, and the mock implements the same logic as on the cloud, as in it does return multiple pages of subscribers on request, but applies the flat digest logic on the first page (as in if the digests match for the request for the first page, it returns `no_updates=True`).

This helps test the full logic for the agw, where if it receives a `no_updates=True` for any page, it stops loading the succeeding pages.

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +78,47 @@ def __init__(
         self._grpc_client_manager = grpc_client_manager
 
     async def _run(self) -> None:
-        subscribers = await self._get_subscribers()
-        if subscribers is not None:
-            self._process_subscribers(subscribers)
+        subscribers, no_updates = await self._get_subscribers()
+        # No need to update local db if it's already up-to-date.
```
hcgatewood: Nit
```suggestion
        # No need to update local db if it's already up-to-date
```

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +78,47 @@ def __init__(
...
+        self._process_subscribers(subscribers)
 
     async def _get_subscribers(self) -> [SubscriberData]:
         subscriberdb_cloud_client = self._grpc_client_manager.get_client()
         subscribers = []
         req_page_token = ""  # noqa: S105
         res_page_token = "start_token"  # noqa: S105
+        cur_digest = SubscribersDigest(
+            md5_base64_digest=self._store.get_current_digest(),
+        )
```
hcgatewood: Side note: I hate how Python convention is to aggressively wrap at 80 chars. It makes the code so much harder to quickly parse / read

Nothing to change in your PR, just evangelizing my position 🙂 

Support for my position from e.g. Linus: http://lkml.iu.edu/hypermail/linux/kernel/2005.3/08168.html

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +78,47 @@ def __init__(
...
+        no_updates = False
         sync_start = datetime.datetime.now()
-        while res_page_token != "":  # noqa: S105
+
+        # If response.no_updates = True, it means local data is up-to-date &
+        # therefore no need for updates from the cloud.
+        while res_page_token != "" and not no_updates:  # noqa: S105
```
hcgatewood: Update the cloud such that it sends an empty page token when there are no updates

The canonical way of reading all subs should be, simply, read till returned page token is empty. no_updates tells you what to do *after* you finish reading all subs, not whether you're done reading

```suggestion
        while res_page_token != "":  # noqa: S105
```

hcgatewood: P.S. I prefer the "while not found_empty_token" construct we used in the cloud servicer tests https://github.com/magma/magma/pull/7302/files#diff-3ac163790ecd8ba19e710823f9528dec68bd82b18f0de71cad82c1066fc3d0c2R131-R133

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +78,47 @@ def __init__(
...
+        if subscribers is None or no_updates:
+            return
+        self._process_subscribers(subscribers)
 
     async def _get_subscribers(self) -> [SubscriberData]:
         subscriberdb_cloud_client = self._grpc_client_manager.get_client()
         subscribers = []
         req_page_token = ""  # noqa: S105
         res_page_token = "start_token"  # noqa: S105
+        cur_digest = SubscribersDigest(
```
hcgatewood:
```suggestion
        digest = SubscribersDigest(
```

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -123,7 +139,9 @@ async def _get_subscribers(self) -> [SubscriberData]:
         SUBSCRIBER_SYNC_LATENCY.observe(
             time_elapsed.total_seconds() * 1000,
         )
-        return subscribers
+
+        self._store.update_digest(cur_digest.md5_base64_digest)
+        return subscribers, no_updates
```
hcgatewood: Make sure we test what happens when cloud sends down empty digest

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -34,13 +35,12 @@ class SqliteStore(BaseStore):
...
-        self._db_locations = self._create_db_locations(
-            db_location, self._n_shards,
-        )
+        self._db_locations, self._digest_db_location = \
+            self._create_db_locations(db_location, self._n_shards)
         self._create_store()
         self._on_ready = OnDataReady(loop=loop)
 
-    def _create_db_locations(self, db_location, n_shards):
+    def _create_db_locations(self, db_location, n_shards) -> [list, str]:
```
hcgatewood: Here and elsewhere that you change a fn signature: go ahead and annotate the parameter types as well. Basically, in Python, any time you update an untyped fn signature, go ahead and type the entire fn signature

lte/gateway/python/magma/subscriberdb/store/sqlite.py
hcgatewood: Ah I see. Yes then this is correct

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -76,9 +82,19 @@ def _create_store(self):
...
+                    "CREATE TABLE IF NOT EXISTS subscriber_digest"
+                    "(digest string PRIMARY KEY, updated_at timestamp)",
+                )
+        finally:
+            conn.close()
+
     def add_subscriber(self, subscriber_data):
         """
-        Method that adds the subscriber.
+        Add the subscriber.
```
hcgatewood:
```suggestion
        Add the subscriber to store.
```

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -266,6 +283,40 @@ def resync(self, subscribers):
...
+            with conn:
+                conn.execute("DELETE FROM subscriber_digest")
+
+                conn.execute(
+                    "INSERT INTO subscriber_digest(digest, updated_at) "
+                    "VALUES (?, ?)", (new_digest, datetime.now()),
+                )
+        finally:
+            conn.close()
+
```
hcgatewood: Add some debug-level logs to both these

lte/gateway/python/magma/subscriberdb/tests/client_tests.py
```diff
@@ -69,11 +70,14 @@ def ListSubscribers(self, request, context):  # noqa: N802
             ListSubscribersResponse
         """
         # Add in logic to allow error handling testing
+        digest, no_updates = request.digest, False
         if request.page_size == 1:
             raise grpc.RpcError("Test Exception")
         if request.page_token == "":
             next_page_token = "aaa"  # noqa: S105
-            subscribers = [
+            digest = SubscribersDigest(md5_base64_digest="11111")
```
hcgatewood: General note: use alphabetized metasyntactic variables with common names, e.g. "digest_apple", "digest_banana"

Ref: https://magma.github.io/magma/docs/next/contributing/contribute_conventions#follow-style-conventions

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +78,47 @@ def __init__(
...
-                subscribers.extend(response.subscribers)
-                res_page_token = response.next_page_token
-                req_page_token = response.next_page_token
+                subscribers.extend(res.subscribers)
+                res_page_token = res.next_page_token
+                req_page_token = res.next_page_token
+                no_updates = res.no_updates
+
+                if cur_digest != res.digest:
+                    cur_digest = res.digest
```
hcgatewood: This if-guard doesn't do anything, can just set it
```suggestion
                cur_digest = res.digest
```

hcgatewood (changes requested):
Looking good, couple cleanup items

lte/gateway/python/magma/subscriberdb/client.py
wangyyt1013: This email piece should definitely be circulated more!

lte/gateway/python/magma/subscriberdb/client.py
wangyyt1013: Currently the logic works as such: when cloud sends down an empty digest, the gateway will store the empty gateway and send it in its next request; however, if `ListSubscribersRequest.Digest = ""`, then cloud will not enable the `no_updates` logic, but will send down the full subscriber list (as well as a new digest if it has one) as usual. Wondering if you think this is a reasonable logic? @hcgatewood 

P.S. the "cloud not enabling `no_updates` logic when `digest = ""`" is included in #7472

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +78,44 @@ def __init__(
...
+            md5_base64_digest=self._store.get_current_digest(),
+        )
+        no_updates = False
         req_page_token = ""  # noqa: S105
-        res_page_token = "start_token"  # noqa: S105
+        found_empty_token = False
         sync_start = datetime.datetime.now()
-        while res_page_token != "":  # noqa: S105
+
+        # If response.no_updates = True, response.next_page_token = ""
```
hcgatewood:
```suggestion
        # Next page token empty means either no updates found or read all updates
```

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -123,7 +136,9 @@ async def _get_subscribers(self) -> [SubscriberData]:
         SUBSCRIBER_SYNC_LATENCY.observe(
             time_elapsed.total_seconds() * 1000,
         )
-        return subscribers
+
+        self._store.update_digest(digest.md5_base64_digest)
```
hcgatewood: To me, "get subscribers" is a read-only action. So it should return the updated digest, and then either "process subscribers" or something like "process digest" should be the one to update/handle the digest

hcgatewood (approved):
Lgtm

1. Fix one small nit
2. Wait to merge until we get both PRs ready and manually verified

lte/gateway/python/magma/subscriberdb/client.py
hcgatewood: Yes this is how it should work, makes sense. My note was just to verify that this is being tested

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +80,113 @@ def __init__(
...
-        if subscribers is not None:
-            self._process_subscribers(subscribers)
+        in_sync = await self._check_subscribers_in_sync()
+        if in_sync:
+            return
+
+        resync = await self._sync_subscribers()
+        if not resync:
+            return
+
```
hcgatewood: Let's remove all the sync_subs logic from the gw at this point, just cause it's not doing anything positive, at the cost of an extra round-trip per update

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +80,113 @@ def __init__(
...
+        ret = await self._get_all_subscribers()
```
hcgatewood: Nit: I don't like using `ret` for variable names -- use something more descriptive

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +80,113 @@ def __init__(
...
+                self._loop,
+            )
+        except grpc.RpcError as err:
+            logging.error(
+                "Sync subscribers request error! [%s] %s", err.code(),
+                err.details(),
+            )
+            return True
+
+        if res.resync:
```
hcgatewood: "return res.resync" 🙂

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +80,113 @@ def __init__(
...
+            )
+        except grpc.RpcError as err:
+            logging.error(
+                "Check subscribers in sync request error! [%s] %s", err.code(),
+                err.details(),
+            )
+            return False
+        return res.in_sync
+
+    async def _sync_subscribers(self) -> bool:
```
hcgatewood: Yeah let's just remove this fn for now (you can save it in a txt file to use during milestone 3)

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -123,7 +207,13 @@ async def _get_subscribers(self) -> [SubscriberData]:
         SUBSCRIBER_SYNC_LATENCY.observe(
             time_elapsed.total_seconds() * 1000,
         )
-        return subscribers
+
+        return [subscribers, flat_digest]
```
hcgatewood: Prefer immutability where possible. May also have to update the fn signature
```suggestion
        return (subscribers, flat_digest)
```

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -77,32 +80,113 @@ def __init__(
...
+        in_sync = await self._check_subscribers_in_sync()
+        if in_sync:
+            return
+
+        resync = await self._sync_subscribers()
+        if not resync:
+            return
+
+        ret = await self._get_all_subscribers()
+        if ret is None:
```
hcgatewood: Nit: emulate golang-style multi-returns -- so set `subscribers, flat_digest = await self._get_all_subs`, then `if subscribers is None: return`

Imo it's easier to parse/follow when you don't mix return types

hcgatewood (approved):
Couple small changes then let's merge. Also need an agw review, msg @themarwhal

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -113,7 +156,7 @@ async def _get_subscribers(self) -> [SubscriberData]:
                     time_elapsed.total_seconds() * 1000,
                 )
                 SUBSCRIBER_SYNC_FAILURE_TOTAL.inc()
-                return None
+                return None, None
         logging.info(
```
ardzoht, line 160: For tuple structs like this I would recommend using a NamedTuple and explicitly name the parameters instead of returning a tuple here, it doesn't have to be in this PR you can add it as a TODO and add this later.

ardzoht (approved):
Lgtm, left one comment but it can be added in a separate PR

lte/gateway/python/magma/subscriberdb/client.py
wangyyt1013: sounds good, will do!

PR #7344, opened 2021-06-05
[feg_relay] fix wrong teid

uri200:

## Summary

Feg relay was using the wrong teid to find the AGW. This should fix the issue

## Test Plan

teravm once landed

## Additional Information

### DIFF
--- a/feg/cloud/go/services/feg_relay/servicers/create_bearer_pgw.go
+++ b/feg/cloud/go/services/feg_relay/servicers/create_bearer_pgw.go
@@ -36,11 +36,11 @@ func (srv *FegToGwRelayServer) CreateBearer(
 	if err := validateFegContext(ctx); err != nil {
 		return nil, err
 	}
-	if req == nil || req.BearerContext == nil || req.BearerContext.UserPlaneFteid == nil {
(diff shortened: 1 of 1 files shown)


feg/cloud/go/services/feg_relay/servicers/create_bearer_pgw.go
```diff
@@ -36,11 +36,11 @@ func (srv *FegToGwRelayServer) CreateBearer(
 	if err := validateFegContext(ctx); err != nil {
 		return nil, err
 	}
-	if req == nil || req.BearerContext == nil || req.BearerContext.UserPlaneFteid == nil {
+	if req == nil {
 		glog.Errorf("unable to send CreateBearerPGWUnverified, Teid not found  %v", req)
```
emakeev: you know, req == nil here, what's a point in Errorf & %v?
Just do glog.Error("unable to send CreateBearerPGWUnverified, nil request")

emakeev (approved):

feg/cloud/go/services/feg_relay/servicers/create_bearer_pgw.go
uri200: Before it made sense when the conditions were more. Now it does not.

PR #7354, opened 2021-06-07
[LTE][AGW] S6A/Diameter Enable extended bitrates

lionelgo:
## Summary

Enable Diameter extended bitrates
Handle ULA Supported-Features AVP for NSA
Send in ULR Dual-Registration-5G-Indicator
Test Plan

## Test Plan

Tested with dsTester NSA scenario:
Provision subscribers in OAI HSS with bitrate(s) greater than 4294967295 bits/s

## Additional Information
    This change is not backward breaking.
    This changes only S6a/Diameter interface, to enable NSA compatibility between Magma and OAI HSS network function.

### DIFF
--- a/lte/gateway/c/core/oai/tasks/s6a/s6a_supported_features.c
+++ b/lte/gateway/c/core/oai/tasks/s6a/s6a_supported_features.c
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the OpenAirInterface (OAI) Software Alliance under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The OpenAirInterface Software Alliance licenses this file to You under
+ * the terms found in the LICENSE file in the root of this source tree.
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


lte/gateway/c/core/oai/tasks/s6a/s6a_supported_features.c
```diff
@@ -0,0 +1,97 @@
...
+ * For more information about the OpenAirInterface (OAI) Software Alliance:
+ *      contact@openairinterface.org
+ */
+
+/*! \file s6a_supported_features.c
+  \brief
+  \author Lionel Gauthier
+  \company Eurecom
+  \email: lionel.gauthier@eurecom.fr
+*/
```
ulaskozat: we should use the standard Magma header for new files.

ulaskozat (approved):
LGTM with a nit comment about using standard Magma copyright statement.

lte/gateway/c/core/oai/tasks/s6a/s6a_supported_features.c
lionelgo: I changed it to the Magma header

PR #7357, opened 2021-06-07
Fix conntrackd zone verification

koolzz:
## Summary
Instead of checking all conntrackd events only check those which match the zone set by ovs

## Test Plan

Checked logs when running conntrack pipelined test, works as expected

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/conntrack.py
+++ b/lte/gateway/python/magma/pipelined/app/conntrack.py
@@ -14,7 +14,6 @@ from magma.pipelined.app.base import ControllerType, MagmaController
 from magma.pipelined.openflow import flows
 from magma.pipelined.openflow.magma_match import MagmaMatch
 from ryu.lib.packet import ether_types
-from ryu.ofproto.nicira_ext import ofs_nbits
 
 
 class ConntrackController(MagmaController):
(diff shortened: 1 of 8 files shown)


lte/gateway/python/magma/pipelined/app/conntrack.py
```diff
@@ -98,13 +99,13 @@ def _install_default_flows(self, datapath):
...
         match = MagmaMatch(eth_type=ether_types.ETH_TYPE_IP,
-                           ct_zone=ofs_nbits(14, 15),
+                           ct_zone=self.zone,
                            ct_state=(self.CT_NEW | self.CT_TRK,
                                      self.CT_NEW | self.CT_TRK))
         actions = [parser.NXActionCT(
             flags=0x1,
             zone_src=None,
-            zone_ofs_nbits=ofs_nbits(14, 15),
+            zone_ofs_nbits=self.zone,
```
pshelar, line 107: I am not sure about this.
the zone_src is reg and zone_ofs_nbits bits to copy from that reg.
https://github.com/magma/magma/blob/b8c11fca3d6cf5d8cb453eea9776c49c45798096/lte/gateway/deploy/roles/magma/files/nx_actions_3.5.py#L2695

am I missing something?

koolzz: The flows seem to be installed correctly I've ran the test_conntrack.py test

I think if zone_src is None its set by the zone_ofs_nbits as said here
```

zone_ofs_nbits | Start and End for the OXM/NXM field. Setting method refer to the&nbsp;nicira_ext.ofs_nbits. If you need set the Immediate value for zone, zone_src must be set to None or empty character string.
```

pshelar (approved):

PR #7358, opened 2021-06-07
Fix Enforcement Poll datapath select

koolzz:

## Summary

It doesn't make sense to iterate over all different datapath values just use the one that we save in controller init.

## Test Plan

Tests pass

## Additional Information


pshelar (approved):
LGTM

PR #7368, opened 2021-06-07
[docker] links session_manager config files

waqaraqeel:

## Summary

When `make test_session_manager` is run in the docker environment, tests that expect config files in `/etc/magma` fail because the docker env doesn't link the required files in the right place during setup. This PR adds a step in Dockerfile.

## Test Plan

Executed `Rebuild Container` in VSCode.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -295,3 +295,6 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git 
     make install && \
     cd / && \
     rm -rf freediameter
+
+#### Link configuration files needed for session_manager
+RUN ln -s lte/gateway/configs/ /etc/magma
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -295,6 +295,3 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git 
     make install && \
     cd / && \
     rm -rf freediameter
-
-#### Link configuration files needed for session_manager
-RUN ln -s lte/gateway/configs/ /etc/magma
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -53,6 +53,7 @@
 	"mounts": [
 		"source=/var/run/docker.sock,target=/var/run/docker-host.sock,type=bind"
 	],
+	"postCreateCommand": "sudo ln -s /workspaces/magma/lte/gateway/configs /etc/magma",
 	"runArgs": [
 		"--init"
 	],
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -53,7 +53,7 @@
 	"mounts": [
 		"source=/var/run/docker.sock,target=/var/run/docker-host.sock,type=bind"
 	],
-	"postCreateCommand": "sudo ln -s /workspaces/magma/lte/gateway/configs /etc/magma",
+	"postCreateCommand": "sudo ln -s ${containerWorkspaceFolder}/lte/gateway/configs /etc/magma",
 	"runArgs": [
 		"--init"
 	],


electronjoe:
Nice find @waqaraqeel!

We will have to think about how we address this.  Ultimately that session_manager tests require these configuration files is tech debt, I believe.  But I think we should probably not stray into fixup for you - we want to get you progress on this project.

The problem with the solution here is that these Dockerfiles are intended only to set up the environment up to but not including the source git repo checkout.  I'm not actually clear how the change here builds successfully - actually maybe you did not test a docker build (understandable, I need to give you some instructions).  The magma repo isn't checked out in this dockerfile.

First some context for a fix here.

1. We have two primary dockerfiles that I am aware of that want to run these tests
    1. `.devcontainer/Dockerfile`
    1. `lte/gateway/dockerfile/mme/Dockerfile.ubuntu20.04`
1. We solve this issue in CI (which uses the `...mme/Dockerfile.ubuntu20.04 image` by [mounting the configs by docker command line](https://github.com/magma/magma/blob/master/.circleci/config.yml#L936)

Here are some brainstormed solutions, by my estimate of preference.

## Option 1 ##

You might do this by adding a [mounts clause here](https://github.com/magma/magma/blob/master/.devcontainer/devcontainer.json#L53-L55).  See the description of [devcontainer.json configuration for mount](https://code.visualstudio.com/docs/remote/devcontainerjson-reference).  Note that the example given here *does appear to bind from our checked out code location of ${localWorkspaceFolder}* so this may be exactly what we want. I am not sure what the warning that `Codespaces ignores "bind" mounts with the exception of the Docker socket.` means.  I would try this option first.

```
	"mounts": [
		"source=/var/run/docker.sock,target=/var/run/docker-host.sock,type=bind",
		"source=${localWorkspaceFolder}/lte/gateway/configs/,target=/etc/magma,type=bind",
	],
```

`${localWorkspaceFolder}` is an env parameter that vs code devcontainer will populate with the path to our checked out repo code inside the devcontainer (checked out by VS code remote-container extension).

## Option 2 ##

Looking at the [devcontainer.json options](https://code.visualstudio.com/docs/remote/devcontainerjson-reference), it appears we can specify `postCreateCommand` or `postStartCommand`. Maybe these can be our symbolic link (unclear at what stage of create vs start VS code remote containers mounts our checked out repo to $(localWorkspaceFolder}). You could put your symbolic link command here.

# On testing

You can test changes to your local `.devcontainer` from within VS code. If already operating in remote container "mode" - hit `F1` and then type `Remote-Containers: Rebuild and Reopen in Container` - this will trigger a re-build of the Dockerfile and a fresh re-load in VS Code remote container mode (locally.. remote container is a bit of a misnomer - you can run remote but we aren't).

There are other useful devonctainer commands if you hit `F1` and begin typing `Remote-Containers:`

waqaraqeel:
Thanks for the detailed feedback @electronjoe! I have removed the `ln` command from `.devcontainer/Dockerfile`. It wasn't building as you figured.

##### Option 1
This would bind from the host (in my case Windows) and introduce, in my opinion, unnecessary coupling between the two environments. I don't even have `magma` cloned in Windows anymore and am staying within the Docker volume. So I went for option 2.

##### Option 2
This seems to work well. I used `postCreateCommand` since we don't want to run the `ln` command on every start. `${localWorkspaceFolder}` does not seem to have been initialized at that point, so I just used `/workspaces/magma`. Is that safe? What do you think?

Tested using VSCode `Rebuild Container`.

electronjoe:
> `${localWorkspaceFolder}` does not seem to have been initialized at that point, so I just used `/workspaces/magma`. Is that safe? What do you think?

OK, lets go with hardcoded for now. Not ideal but not awful.  I have noted that some env variables aren't available at all phases.

waqaraqeel:
> > `${localWorkspaceFolder}` does not seem to have been initialized at that point, so I just used `/workspaces/magma`. Is that safe? What do you think?
> 
> OK, lets go with hardcoded for now. Not ideal but not awful. I have noted that some env variables aren't available at all phases.

`${localWorkspaceFolder}` is the right env var, it's just not supported when using "Clone repository in container volume". Replaced with `${containerWorkspaceFolder}` which is supported. No hardcoding now.

themarwhal (commented):
this looks great! I'll let @electronjoe approve / merge

themarwhal:
I think we should better advertise the dev container setup here: https://docs.magmacore.org/docs/contributing/contribute_onboarding#getting-started

We have docs for Intellij setup but none for visual studio.

electronjoe:
> I think we should better advertise the dev container setup here: https://docs.magmacore.org/docs/contributing/contribute_onboarding#getting-started
> 
> We have docs for Intellij setup but none for visual studio. 

Good point Marie. I'll do this tomorrow morning.

electronjoe (approved):
Looks great

PR #7373, opened 2021-06-07
[mme] Remove additional unnecessary AssertFatal calls

andreilee:
## Summary

Removes two unnecessary `AssertFatal` calls in MME service.

- `itti_alloc_new_message` cannot return NULL, as it asserts in its function body for non-NULL return values.
- `bdestroy_wrapper` will always clean up the input, so the `AssertFatal` check after is unnecessary.

## Test Plan

Only built MME:

```
.
.
.
[1973/1973] Linking CXX executable oai_mme/mme
[16/16] Completed 'MagmaCore'
```

## Additional Information


ardzoht (approved):
lgtm

PR #7377, opened 2021-06-07
[mme][refactor] Remove AssertFatal from mme_api_get_emm_config

andreilee:

## Summary

This PR is among a series of changes to rework how the `AssertFatal` macro is used in MME. As part of this, API methods should not use `AssertFatal`, instead allowing callers higher up to decide whether or not to stop MME.

In this PR, function `mme_api_get_emm_config(..)` is modified to remove all calls to the `AssertFatal` macro, instead returning `RETURNerror`. The only caller of this function is `emm_main_initialize(..)`, which has been modified to call `Fatal`.

## Test Plan

Only built MME service:

```
.
.
.
[1971/1971] Linking CXX executable oai_mme/mme
```

## Additional Information


ulaskozat (changes requested):
We lose granularity on what config was the culprit for initialization failure.

andreilee:
@ulaskozat MME should still throw, no? And it looks like there's only one config that gets processed by that function as well.

lte/gateway/c/core/oai/tasks/nas/emm/emm_main.c
```diff
@@ -62,8 +63,7 @@ void emm_main_initialize(const mme_config_t* mme_config_p) {
    */
   memset(&_emm_data.conf, 0, sizeof(_emm_data.conf));
   if (mme_api_get_emm_config(&_emm_data.conf, mme_config_p) != RETURNok) {
-    OAILOG_ERROR(
-        LOG_NAS_EMM, "EMM-MAIN  - Failed to get MME configuration data");
+    Fatal("EMM-MAIN  - Failed to get MME configuration data");
```
ulaskozat: Maybe add more context to message, e.g., "EMM-MAIN - Failed to get all required MME config data, check the error logs for missing configs"

ulaskozat (approved):
Okay I guess we are printing the Error logs so one can figure out the offending configs. I added a nit comment so that message provides hint to look for the missing config.

PR #7382, opened 2021-06-08
[5G SA][LTE][AGW][SMF]: Handling SetUPFSessionsConfig in SMF

GANESH-WAVELABS:
## Summary

[5G SA][LTE][AGW][SMF]: Handling SetUPFSessionsConfig in SMF

1)  Periodic messages about UPF session config
2) Read the SessionMap from global session store.
3) Search with session search criteria of IMSI and session_id and find  respective sesion to operate
4) If No session found in SessionMap for IMSI with teid continue with other IMSI sessions.
5) If version mismatch then send the snapshot of session to UPF

## Test Plan
1) Added new test case "test_upf_session_config" in the file test_upf_node_state.cpp to test the functionality of "SetUPFSessionConfig" and "m5g_send_session_request_to_upf"

## Additional Information

Signed-off-by: GANESH-WAVELABS <ganesh.irrinki@wavelabs.ai>


lte/gateway/c/session_manager/test/test_upf_node_state.cpp
```diff
@@ -90,6 +95,48 @@ TEST_F(SetUPFNodeState, test_upf_node_state) {
...
+      uint32_t version     = ses_config.upf_session_state(i).session_version();
+      uint32_t teid        = ses_config.upf_session_state(i).local_f_teid();
+      auto session_map     = session_store->read_sessions({imsi});
+      /* Search with session search criteria of IMSI and session_id and
+       * find  respective sesion to operate
+       */
+      SessionSearchCriteria criteria(imsi, IMSI_AND_TEID, teid);
+      auto session_it = session_store->find_session(session_map, criteria);
+      if (!session_it) {
+        EXPECT_FALSE(session_it);
```
themarwhal: What does this line do? If the condition is `!session_it` isn't this always going to be false?

lte/gateway/c/session_manager/test/test_upf_node_state.cpp
```diff
@@ -90,6 +95,48 @@ TEST_F(SetUPFNodeState, test_upf_node_state) {
...
+      /* Validating UPF verions of session imsi of teid received version
+       * with SMF latest version
+       */
+      EXPECT_LE(version, cur_version);
+
+      // Sending the session request to UPF
+      session_enforcer->m5g_send_session_request_to_upf(imsi, session);
+
+      // Validating count increment
+      EXPECT_TRUE(count++);
```
themarwhal: What does `EXPECT_TRUE` do in this context?

lte/gateway/c/oai/tasks/mme_app/mme_app_context.c
```diff
@@ -2224,7 +2224,6 @@ static bool mme_app_recover_timers_for_ue(
   OAILOG_FUNC_RETURN(LOG_MME_APP, false);
 }
 
-
```
themarwhal: might want to remove the no-op oai changes to reduce the number of necessary reviewers :)

lte/gateway/c/session_manager/SessionState.h
```diff
@@ -164,6 +164,8 @@ class SessionState {
   void insert_far(SetGroupFAR* rule);
 
   void remove_all_rules();
+  /* method to increment retransmit count every time we resend*/
```
themarwhal:
```suggestion
  
  /**
   * Increment retransmit counter and return the updated value
   */
```

lte/gateway/c/session_manager/SessionState.h
```diff
@@ -719,6 +721,8 @@ class SessionState {
   uint64_t pdp_end_time_;
   /*5G related message to handle session state context */
   uint32_t current_version_;  // To compare with incoming session version
+  // SMF-UPF version missmatch, max re-sending throttle count
```
themarwhal:
```suggestion
  // Counter to keep track of number of retries in case of SMF-UPF version mismatch
```

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
```diff
@@ -604,6 +617,12 @@ bool SessionStateEnforcer::set_upf_node(
   return true;
 }
 
+bool SessionStateEnforcer::inc_rtx_counter(
```
themarwhal: No need to abbreviate here I think. 
`is_incremented_rtx_counter_within_max` maybe is more explicit

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
```diff
@@ -328,6 +330,17 @@ void SessionStateEnforcer::m5g_remove_rules_and_update_upf(
   return;
 }
 
+void SessionStateEnforcer::m5g_send_session_request_to_upf(
+    const std::string& imsi, const std::unique_ptr<SessionState>& session) {
```
themarwhal: imsi doesn't seem to be used in this function. should we omit it?

lte/gateway/c/session_manager/SessionStateEnforcer.h
```diff
@@ -87,6 +92,8 @@ class SessionStateEnforcer {
    */
   bool set_upf_node(const std::string& node_id, const std::string& n3_addr);
 
+  bool inc_rtx_counter(const std::unique_ptr<SessionState>& session);
```
themarwhal: This looks like a function that belongs better in `SessionState` rather than `SessionStateEnforcer`

themarwhal: ooo nvm we need the max number.
Then let's use the doxygen style doc here to be more clear about what this function mutates / returns.
Should look something like this
```
  /**
   * TODO: Fill in
   *
   * @param session
   * @return TODO: fill in
   */
```

lte/gateway/c/session_manager/UpfMsgManageHandler.cpp
```diff
@@ -70,4 +71,56 @@ void UpfMsgManageHandler::SetUPFNodeState(
...
+    ServerContext* context, const UPFSessionConfigState* sess_config,
+    std::function<void(Status, SmContextVoid)> response_callback) {
+  auto& ses_config = *sess_config;
+  int32_t count    = 0;
+  conv_enforcer_->get_event_base().runInEventBaseThread([this, &count,
+                                                         ses_config]() {
+    for (int i = 0; i < ses_config.upf_session_state_size(); i++) {
+      // Deleting the IMSI prefix from imsi
+      std::string imsi_upf = ses_config.upf_session_state(i).subscriber_id();
+      std::string imsi     = imsi_upf.substr(4, imsi_upf.length() - 4);
```
themarwhal, line 88: why do we remove the prefix here?

lte/gateway/c/session_manager/UpfMsgManageHandler.cpp
```diff
@@ -70,4 +71,56 @@ void UpfMsgManageHandler::SetUPFNodeState(
...
+ *
+ */
+void UpfMsgManageHandler::SetUPFSessionsConfig(
+    ServerContext* context, const UPFSessionConfigState* sess_config,
+    std::function<void(Status, SmContextVoid)> response_callback) {
+  auto& ses_config = *sess_config;
+  int32_t count    = 0;
+  conv_enforcer_->get_event_base().runInEventBaseThread([this, &count,
+                                                         ses_config]() {
+    for (int i = 0; i < ses_config.upf_session_state_size(); i++) {
```
themarwhal: Might be nicer to iterate like
`for (auto upf_session : ses_config.upf_session_state)` unless you want the index.

themarwhal:
Thanks for keeping the PR relatively short :) Left some comments. (also please address the merge conflicts)

lte/gateway/c/session_manager/UpfMsgManageHandler.cpp
GANESH-WAVELABS: While calling SessionSearchCriteria() with IMSI_AND_TEID it expects without prefix sub string 'IMSI'. I think in UPF they expect like so.

lte/gateway/c/session_manager/test/test_upf_node_state.cpp
GANESH-WAVELABS: session_it is not always false, sometimes If we don't find a session with SessionSearchCriteria() continue with next imsi + SearchCriteria. So, I wanted to validate the false case here.

lte/gateway/c/session_manager/test/test_upf_node_state.cpp
GANESH-WAVELABS: I wanted to validate the count increment, as it was not looking good previously I have modified it.

lte/gateway/c/session_manager/test/test_upf_node_state.cpp
GANESH-WAVELABS: session_it is not always false. Sometimes if we don't find session with SessionSearchCriteria(), we will continue with next imsi+SearchCriteria , So I wanted to validate the false case here.

lte/gateway/c/oai/tasks/mme_app/mme_app_context.c
GANESH-WAVELABS: not committed this oai file this time.

lte/gateway/c/session_manager/SessionState.h
GANESH-WAVELABS: Modified as you suggested.

lte/gateway/c/session_manager/SessionState.h
GANESH-WAVELABS: Modified as you suggested.

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
GANESH-WAVELABS: removed imsi as you suggested.

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
GANESH-WAVELABS: renamed the function as you suggested

lte/gateway/c/session_manager/SessionStateEnforcer.h
GANESH-WAVELABS: added the doxygen style doc as you suggested

lte/gateway/c/session_manager/UpfMsgManageHandler.cpp
GANESH-WAVELABS: removed the index from for loop, and updated the for loop as you suggested.

lte/gateway/c/session_manager/UpfMsgManageHandler.cpp
themarwhal: I'm concerned that this may cause some issues in the future. All other types of sessions have IMSI stored with the prefix set. We should converge to one. This is out of scope for this PR, do you mind filing an issue to track this?

themarwhal: @GANESH-WAVELABS

themarwhal (approved):

lte/gateway/c/session_manager/UpfMsgManageHandler.cpp
GANESH-WAVELABS: Sure Marie I will file an issue to track this.

PR #7384, opened 2021-06-08
[MME][ZMQ] protect against using a destroyed zmq context

talkhasib:

## Summary

When a zmq context is destroyed during shutdown of MME, the context might be used by a pending GRPC service. This causes a benign assert that might mask other issues.

The assert that we have been observing due to this issue was
"Sending to task without push socket. id: %s to %s!"
And is only observed during an MME shutdown

This change adds a "ready" flag that gets cleared when the context is destroyed.

## Test Plan

Built and tested on local focal VM with s1aptester

### DIFF
--- a/lte/gateway/c/core/oai/lib/itti/intertask_interface.c
+++ b/lte/gateway/c/core/oai/lib/itti/intertask_interface.c
@@ -80,6 +80,9 @@ const int itti_debug = ITTI_DEBUG_ISSUES | ITTI_DEBUG_MP_STATISTICS;
 #define MESSAGE_SIZE(mESSAGEiD)                                                \
   (sizeof(MessageHeader) + itti_desc.messages_info[mESSAGEiD].size)
 
+#define likely(x)      __builtin_expect(!!(x), 1)
+#define unlikely(x)    __builtin_expect(!!(x), 0)
+
 typedef volatile enum task_state_s {
   TASK_STATE_NOT_CONFIGURED,
   TASK_STATE_STARTING,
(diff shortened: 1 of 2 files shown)


lte/gateway/c/core/oai/lib/itti/intertask_interface.c
```diff
@@ -118,23 +118,25 @@ static itti_desc_t itti_desc;
...
-      message, sizeof(MessageHeader) + message->ittiMsgHeader.ittiMsgSize);
-  assert(frame);
-
-  // Protect against multiple threads using this context
-  pthread_mutex_lock(&task_zmq_ctx_p->send_mutex);
-  int rc =
-      zframe_send(&frame, task_zmq_ctx_p->push_socks[destination_task_id], 0);
-  assert(rc == 0);
-  pthread_mutex_unlock(&task_zmq_ctx_p->send_mutex);
+  if (task_zmq_ctx_p->ready != false) {
```
ulaskozat: simplify to `if (task_zmq_ctx_p->ready) {`

lte/gateway/c/core/oai/lib/itti/intertask_interface.c
```diff
@@ -118,23 +118,25 @@ static itti_desc_t itti_desc;
...
+        message, sizeof(MessageHeader) + message->ittiMsgHeader.ittiMsgSize);
+    assert(frame);
+
+    // Protect against multiple threads using this context
+    pthread_mutex_lock(&task_zmq_ctx_p->send_mutex);
+    int rc =
+        zframe_send(&frame, task_zmq_ctx_p->push_socks[destination_task_id], 0);
+    assert(rc == 0);
+    pthread_mutex_unlock(&task_zmq_ctx_p->send_mutex);
+  }
```
ulaskozat, line 146: want to add a log with else clause?

ulaskozat (approved):
left some nit comments, otherwise looks good.

talkhasib:
addressed nits
Also added a __builtin_expect to help compiler make an educated guess. Not sure how much it would help

PR #7385, opened 2021-06-08
[orc8r][tracing] Context propagation in all orc8r client libraries except configurator

xjtian:
## Summary

- Make client_api.go files context-aware and propagate contexts to GRPC calls
- No configurator yet because that one will require its own PR
- Mostly automated refactoring. bootstrapper servicer and identity decorator are the only interesting files in this diff

## Test Plan

- CI, `make test` locally

## Additional Information

Signed-off-by: Jacky Tian <jacky@xjtian.com>


themarwhal (approved):

fbinternal/cloud/go/services/testcontroller/statemachines/enodebd_test.go
```diff
@@ -110,7 +110,7 @@ func Test_EnodebdE2ETestStateMachine_HappyPath(t *testing.T) {
 	// Check upgrade status, gateway hasn't upgraded yet
 	// ---
 	gatewayRecord := &models2.GatewayDevice{HardwareID: "hw1", Key: &models2.ChallengeKey{KeyType: "ECHO"}}
-	err = device.RegisterDevice("n1", orc8r.AccessGatewayRecordType, "hw1", gatewayRecord, serdes.Device)
+	err = device.RegisterDevice(context.Background(), "n1", orc8r.AccessGatewayRecordType, "hw1", gatewayRecord, serdes.Device)
```
hcgatewood, line 113: For all test files: in my mind, the first-best solution for tests is to have one context.Background shared across each test. That way, hypothetically, we could hook up jaeger with any particular misbehaving test to peer in better understand the issues

That said, if it's non-trivial effort to switch to this pattern, fk it and we can keep things simple

orc8r/cloud/go/services/certifier/client_api_test.go
```diff
@@ -14,13 +14,15 @@ limitations under the License.
...
 import (
+	"context"
 	"crypto/x509"
 	"fmt"
 	"testing"
 	"time"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/stretchr/testify/assert"
+	context2 "golang.org/x/net/context"
```
hcgatewood, line 25: For a followup PR: can you grep-replace all usages of this pkg, in favor of the stdlib `context`

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
```diff
@@ -1006,14 +1006,10 @@ func TestCreateGateway(t *testing.T) {
...
-		&models.GatewayDevice{
-			HardwareID: "hw2",
-			Key:        &models.ChallengeKey{KeyType: "ECHO"},
-		},
-		serdes.Device,
-	)
+	err = device.RegisterDevice(context.Background(), "n1", orc8r.AccessGatewayRecordType, "hw2", &models.GatewayDevice{
+		HardwareID: "hw2",
+		Key:        &models.ChallengeKey{KeyType: "ECHO"},
+	}, serdes.Device)
```
hcgatewood: Here and elsewhere: ideally let's preserve/make uniform the formatting of these

Or, basically the min of what I'm looking for is newline after `(` and newline before `)`, other than that I care less

orc8r/cloud/go/services/accessd/accessd_test.go
```diff
@@ -27,6 +28,7 @@ import (
 
 func TestAccessManager(t *testing.T) {
 	accessd_test_service.StartTestService(t)
+	ctx := context.Background()
```
hcgatewood, line 31: From earlier comment: yeah this was the test-based pattern I had in mind

orc8r/cloud/go/services/bootstrapper/servicers/bootstrapper.go
```diff
@@ -282,6 +284,16 @@ func getChallengeKey(hwID string) (protos.ChallengeKey_KeyType, []byte, error) {
 	return protos.ChallengeKey_KeyType(keyType), key, nil
 }
 
+// Bootstrapper needs to ignore incoming SN/CN headers in ctx when it fans
+// out calls to upstream services, or else the identity middleware will trip.
+func strippedIncomingCtx(ctx context.Context) context.Context {
```
hcgatewood, line 289: 👍

orc8r/cloud/go/services/dispatcher/servicers/sync_rpc_service.go
```diff
@@ -169,7 +169,7 @@ func (srv *SyncRPCService) receiveFromStream(
 			return
 		} else {
 			glog.V(2).Infof("HWID %v: processing response", coordinator.GwID)
-			err := srv.processSyncRPCResp(syncRPCResp, coordinator.GwID)
+			err := srv.processSyncRPCResp(coordinator.Ctx, syncRPCResp, coordinator.GwID)
```
hcgatewood, line 172: Idk if this is the correct ctx but assuming you figured it out

orc8r/cloud/go/services/orchestrator/obsidian/handlers/gateway_handler_factory.go
```diff
@@ -155,13 +155,14 @@ func GetReadGatewayDeviceHandler(path string, serdes serde.Registry) obsidian.Ha
 				return nerr
 			}
 
+			reqCtx := c.Request().Context()
```
hcgatewood, line 158: We're a little non-uniform with how we're naming/handling the `echo.Context` contexts. Not end of the world, but my preference would be: at the start of every echo handler func that needs to propagate its context, go ahead and grab `ctx := c.Request().Context()` at line 1 of the function, then pass that ctx in everywhere it's required

hcgatewood (changes requested):
Thanks for the fix! Couple small nits. I'm down to stamp as-is, but preference would be to add the small couple cleanups here and going forward, for uniformity purposes. Lmk what you think

fbinternal/cloud/go/services/testcontroller/statemachines/enodebd_test.go
xjtian: It's a lot of files being changed, so where you see context.Background() it's from an automated refactor to update the function signature (except in some edge cases). I think it'll be better to update these as needed the future once you have a concrete idea of how you'll want to configure tracing in unit tests.

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
xjtian: yeah this is the refactorer's doing, I can do a once-over and check

orc8r/cloud/go/services/orchestrator/obsidian/handlers/gateway_handler_factory.go
xjtian: I save the `reqCtx` variable when there's more than 1 place in the callsite that will need to propagate the context, otherwise I just use this inline at the client call

edit: I declare the variable as late as possible so as not to clutter the head of the functions with something that won't be used if the req fails validation for e.g.

xjtian:
Updated PR to include all orc8r module services

orc8r/cloud/go/services/orchestrator/obsidian/handlers/gateway_handler_factory.go
hcgatewood: In my mind, the context would always be the first defn -- to such a degree that ideally it wouldn't even need to be declared / would have language-level support for implicitly passing ctxs around

But in practice, I'm fine with this approach as well

feg/cloud/go/services/basic_acct/tests/setup.go
```diff
@@ -146,11 +147,7 @@ func SetupNetworks(t *testing.T) {
 		serdes.Entity,
 	)
 	assert.NoError(t, err)
-	err = device.RegisterDevice(
-		FederatedLteNetworkID, orc8r.AccessGatewayRecordType, AgwHwId,
-		&models.GatewayDevice{HardwareID: AgwHwId, Key: &models.ChallengeKey{KeyType: "ECHO"}},
-		serdes.Device,
-	)
+	err = device.RegisterDevice(context.Background(), FederatedLteNetworkID, orc8r.AccessGatewayRecordType, AgwHwId, &models.GatewayDevice{HardwareID: AgwHwId, Key: &models.ChallengeKey{KeyType: "ECHO"}}, serdes.Device)
```
hcgatewood, line 150: Here and elsewhere, for now or later: line too long, prefer to break up these types of lines in a canonical way

lte/cloud/go/services/lte/analytics/calculations/state.go
```diff
@@ -129,6 +133,8 @@ func (x *SiteMetricsCalculation) Calculate(prometheusClient query_api.Prometheus
 	gatewayVersionCfg, gatewayVersionCfgOk := x.AnalyticsConfig.Metrics[metrics.GatewayMagmaVersionMetric]
 	enbConnectedCfg, enbConnectedOk := x.AnalyticsConfig.Metrics[metrics.EnodebConnectedMetric]
 
+	// TODO: you'll want to set some tracing params on this ctx
+	outgoingContext := context.TODO()
```
hcgatewood, line 137: For my understanding: why TODO here

orc8r/cloud/go/tools/accessc/handlers/add_existing.go
```diff
@@ -90,7 +93,7 @@ func addExisting(cmd *commands.Command, args []string) int {
 				log.Fatalf(
 					"Cannot parse certificate from '%s': %s\n", certFName, err)
 			}
-			err = certifier.AddCertificate(operator, certDERBlock.Bytes)
+			err = certifier.AddCertificate(context.Background(), operator, certDERBlock.Bytes)
```
hcgatewood: For all the accessc handlers: I'd imagined that each cmd should use a single ctx for the entirety of the cmd

hcgatewood (approved):
Lgtm, thanks for adding Jacky 🙂 

Main comment at this point is the per-CLI-handler contexts. But generally lgtm. Either way, msg me when you're ready and I'll merge for you

lte/cloud/go/services/lte/analytics/calculations/state.go
xjtian: This is the point of origination for the network call, so you'd want to start your trace here and use that context. But that'll depend on your tracing client library

xjtian:
@hcgatewood good to merge whenever

Issue #7136, opened 2021-05-25
[MME] Address uncovered MmeNasStateConverter test failures


electronjoe:
## Summary

In order to commit a randomized MmeNasStateConverter property test of #6903, which validates that a random `proto` -> `MME Nas State Struct` -> `proto_prime` always satisfies `proto == proto_prime`, some of the uncovered failures need to be addressed.  These failures do not necessarily represent bugs that hit production today - but represent a suite of `{Bugs, Better Engineering, Stability Vulnerabilities, Require Unit Test Work-Arounds}`.

The list of uncovered issues will be provided here.

- [ ] Failure by [stoul](https://en.cppreference.com/w/cpp/string/basic_string/stoul) to convert `mme_ue_contexts.guti_ue_id_htbl[].key` gives uncaught exception [0]
- [ ] The `magma::lte::oai::MmeNasState` sub-field `statistic_timer_id` is not preserved [1]
- [ ] The `MmeNasStateConverter::proto_to_guti_table(...)` function fails to clean up the dynamic guti_t* when exception is thrown by `mme_app_convert_string_to_guti` [2]
- [ ] The `mme_app_ue_s1ap_id_generator` must be set in our proto or it is initialized to 1 (is this safe?) [3]

```
[0]
IN:mme_ue_contexts {
  guti_ue_id_htbl {
    key: "\016"
    value: 0
  }
}
nb_enb_connected: 294912
nb_ue_attached: 1147136
nb_ue_connected: 36864
nb_default_eps_bearers: 402653188
nb_ue_managed: 3489661440
nb_ue_idle: 133136
nb_bearers_managed: 24584
nb_ue_since_last_stat: 1
nb_bearers_since_last_stat: 40960
mme_app_ue_s1ap_id_generator: 524544
unknown file: Failure
C++ exception with description "stoul" thrown in the test body.
[  FAILED  ] MmeAppStateConverter.DoubleConversionPropertyTest (30 ms)
```

```
[1]
Beginning Mutation #199
IN:nb_enb_connected: 8266039
nb_ue_attached: 4279336960
nb_ue_connected: 277389888
nb_default_eps_bearers: 1082156052
nb_s1u_bearers: 1981390848
statistic_timer_id: 300193089560839192
nb_ue_managed: 3503949888
nb_ue_idle: 1090727024
nb_bearers_managed: 68448264
nb_ue_since_last_stat: 4093806593
nb_bearers_since_last_stat: 346170193
[...]

Issue #6903, opened 2021-05-13
[MME] Implement state converter property test for mme_app


electronjoe:
## Summary

Per #5042, explore authoring a property test for one of three today-untested state converters. Here, the mme_app state converters.

Per comments in #4996, explore use of `libprotobuf-mutator` in the property test.

PR #7387, opened 2021-06-08
[MME] Remove unused MmeNasState proto field

electronjoe:
## Summary

This proto field is unused in state backup, and to prevent its use in the randomized state converter property tests (which would cause test failure) here we move it to a reserved tag ID. This fixup is required for #6903 and is enumerated in #7136.

## Test Plan

Run unit tests and lte `integ_test`.


electronjoe:
Looks like I'm failing https://github.com/magma/magma/blob/master/.circleci/config.yml#L566-L568 and need to regenerate some proto outputs that we have committed to the repo.

ekowtaylor:
@electronjoe this is in draft but are you still planning to move forward with this?

electronjoe:
Can re open later

PR #7388, opened 2021-06-08
[MME] Migrate to unique_ptr to release allocation on exception

electronjoe:
## Summary

This fixup is required for #6903 and is enumerated in #7136.  The new randomized state converter property test sometimes throws an exception when converting some string fields to integers.  This results in ASAN failures by the unit test, as memory is leaked by some allocations when exceptions are thrown.

This PR wraps the allocation that is leaking - with a unique ptr. Ensuring cleanup even when exceptions are thrown.

## Test Plan

Unit tests, property tests, and `lte integ_test`.


ekowtaylor:
@electronjoe, this seems like a small change. Do you want to move forward with it still and when do you expect to have this closed if that is the case?

PR #7393, opened 2021-06-08
fix bug where policy versions are overwritten when processing policy

themarwhal:
stats

## Summary
Currently the `policy_version_and_stats` field in `SessionStateUpdateCriteria` is an all or nothing field. If there are no updates, it is empty. If there is any update, it contains the entire map. 

We introduced a regression where we were only partially updating the map when we received some rule stats. This led to some policy versions being wiped away and lost. I noticed this when I was running some s1ap tests and kept seeing this log: `RuleID ims-voice doesn't have a version registered for IMSI001010000000001-798160, this is unexpected`

I added a unit test that fails on master but passes with the new test

## Test Plan
add a new unit test to prevent this regression in the future
CI

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/SessionState.cpp
+++ b/lte/gateway/c/session_manager/SessionState.cpp
@@ -185,9 +185,6 @@ SessionState::SessionState(
   for (auto& rule : marshaled.gy_dynamic_rules) {
     gy_dynamic_rules_.insert_rule(rule);
   }
-  for (auto& it : marshaled.policy_version_and_stats) {
-    policy_version_and_stats_[it.first] = it.second;
-  }
 }
 
 SessionState::SessionState(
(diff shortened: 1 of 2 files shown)


lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -185,9 +185,6 @@ SessionState::SessionState(
   for (auto& rule : marshaled.gy_dynamic_rules) {
     gy_dynamic_rules_.insert_rule(rule);
   }
-  for (auto& it : marshaled.policy_version_and_stats) {
```
themarwhal, line 188: This is repetitive logic. see line 147 in the same file

uri200 (approved):

PR #7431, opened 2021-06-09
Fix metrics collection chunks

ardzoht:
## Summary

- The current collection of metrics samples is based on splitting chunks by an even number of buckets, however this can lead to a `message larger than gRPC max allowed size` since the splitting does not take into account the size of each chunk.
- This change updates the chunking of samples to do the splitting based on size of each bucket, this way we ensure there is not a sent message that is larger than the allowed maximum.

## Test Plan

- Used local script to generate metrics on the AGW so they can be scraped by prometheus, provided by @karthiksubraveti 
- Before (one chunk with size 417 bytes and the second with 8093664 bytes):
```
Jun 08 23:35:12 magma-dev-focal magmad[282473]: INFO:root:Preparing chunk 0
Jun 08 23:35:12 magma-dev-focal magmad[282473]: INFO:root:Metric container size: 417
Jun 08 23:35:12 magma-dev-focal magmad[282473]: INFO:root:Preparing chunk 1
Jun 08 23:35:12 magma-dev-focal magmad[282473]: INFO:root:Metric container size: 8093664
Jun 08 23:35:13 magma-dev-focal magmad[282473]: ERROR:root:Prometheus Target Metrics upload error! [StatusCode.RESOURCE_EXHAUSTED] Sent message larger than max (8093664 vs. 4194304)
```
- After:
```
Jun 08 23:24:26 magma-dev-focal magmad[279947]: INFO:root:Preparing chunk 0
Jun 08 23:24:26 magma-dev-focal magmad[279947]: INFO:root:Metric container size: 2698432
Jun 08 23:24:26 magma-dev-focal magmad[279947]: INFO:root:Submitting chunk num 1 with size 2697801 bytes
Jun 08 23:24:26 magma-dev-focal magmad[279947]: INFO:root:Preparing chunk 1
Jun 08 23:24:26 magma-dev-focal magmad[279947]: INFO:root:Metric container size: 2697844
Jun 08 23:24:27 magma-dev-focal magmad[279947]: INFO:root:Submitting chunk num 2 with size 2697800 bytes
Jun 08 23:24:27 magma-dev-focal magmad[279947]: INFO:root:Preparing chunk 2
Jun 08 23:24:27 magma-dev-focal magmad[279947]: INFO:root:Metric container size: 2697843
Jun 08 23:24:27 magma-dev-focal magmad[279947]: INFO:root:Submitting final chunk  with size 2697800 bytes
```

## Additional Information

### DIFF
--- a/orc8r/gateway/python/magma/magmad/metrics_collector.py
+++ b/orc8r/gateway/python/magma/magmad/metrics_collector.py
@@ -13,9 +13,8 @@ limitations under the License.
 import asyncio
 import calendar
 import logging
-import math
 import time
-from typing import Callable, Dict, List, NamedTuple, Optional
+from typing import Callable, Dict, List, NamedTuple, Optional, Union
 
 import metrics_pb2
 import prometheus_client.core
@@ -50,7 +49,7 @@ class MetricsCollector(object):
         collect_interval: int,
         sync_interval: int,
         grpc_timeout: int,
-        grpc_max_msg_size_mb: int,
+        grpc_max_msg_size_mb: Union[int, float],
         loop: Optional[asyncio.AbstractEventLoop] = None,
         post_processing_fn: Optional[Callable] = None,
         scrape_targets: [ScrapeTarget] = None,
--- a/orc8r/gateway/python/magma/magmad/metrics_collector.py
+++ b/orc8r/gateway/python/magma/magmad/metrics_collector.py
@@ -14,7 +14,6 @@ import asyncio
 import calendar
 import logging
 import math
-import sys
 import time
 from typing import Callable, Dict, List, NamedTuple, Optional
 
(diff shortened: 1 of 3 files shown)


karthiksubraveti (changes requested):
Can you also include the chunking logic for metric scrape targets ?

ardzoht:
> Can you also include the chunking logic for metric scrape targets ?

Yes I missed this on the new branch, will update, thanks!

orc8r/gateway/python/magma/magmad/metrics_collector.py
```diff
@@ -219,15 +218,26 @@ def _make_sync_done_func(self, service_name, chunk):
...
-        sample_length = len(samples)
-        chunk_size = sample_length // buckets
-
-        for i in range(0, sample_length, chunk_size):
-            yield samples[i:i + chunk_size]
+        chunk_size_bytes = sample_size_bytes // buckets
+
+        chunked_samples = []
+        chunked_samples_size = 0
+        for s in samples:
```
karthiksubraveti, line 224: If we use this logic, then we can simply iterate till `chunked_samples_size + s.ByteSize() <= chunked_samples_size + s.ByteSize() <=`, we can get rid of the buckets computation.

ardzoht: We do need the buckets computation to check what is the maximum bytes size for each chunk:

```
        buckets = math.ceil(
            sample_size_bytes / self.grpc_max_msg_size_bytes,
        )
        chunk_size_bytes = sample_size_bytes // buckets
```

karthiksubraveti: Why do we need to stop at chunk_size_bytes, we can probably iterate till total number of chunks is less than the grpc msg size no?

ardzoht: Ok sorry didn't get your point at first, yes basically we chunk at `self.grpc_max_msg_size_bytes`

karthiksubraveti (approved):
Given that this method `chunk_samples` is well contained, would it be possible to add a unit test to verify this method alone perhaps. Thanks @ardzoht for making this change.

PR #7435, opened 2021-06-09
count rule records with 0 usage in sessions_with_reporting_flows

themarwhal:

## Summary
we currently only count records with non-zero usage as `sessions_with_reporting_flows`. This leads to weird logs when printing 
```
   MLOG(MINFO) << "Received stats for " << sessions_with_reporting_flows.size()
                << " active sessions and " << dead_sessions_to_cleanup.size()
                << " stale sessions";
```

This variable is used to determine any terminating sessions that no longer has stats in pipelined. We should not treat a record with 0 values as an indication of pipelined clearing state for the rule. Pipelined will stop sending any record once it cleans up state. 

## Test Plan
Had to modify a test since it was making an assumption that we don't remove the drop rule on termination. -> we do.

terravm test 2+5+5capacity
Screen Shot 2021-06-14 at 10 14 18 AM

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -248,8 +248,8 @@ void LocalEnforcer::aggregate_records(
 
     auto& session                 = **session_it;
     const std::string& session_id = session->get_session_id();
+    sessions_with_reporting_flows.insert(ImsiAndSessionID(imsi, session_id));
     if (record.bytes_tx() > 0 || record.bytes_rx() > 0) {
-      sessions_with_reporting_flows.insert(ImsiAndSessionID(imsi, session_id));
       MLOG(MDEBUG) << session_id << " used " << record.bytes_tx()
                    << " tx bytes and " << record.bytes_rx()
                    << " rx bytes for rule " << record.rule_id();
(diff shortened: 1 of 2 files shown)


uri200:
The idea is to wait until pipelined removes the rule, then we can assure the rule is gone. 

If for any reason pipelined keeps reporting 0, that means the rule hasn't been removed for some reason. Why is the rush to terminate if the rule hasn't really been removed?

I think that part of the code was tested with teravm gxgy_05 test and gxgy_capacity test

themarwhal:
> The idea is to wait until pipelined removes the rule, then we can assure the rule is gone. 
> 
> 
> 
> If for any reason pipelined keeps reporting 0, that means the rule hasn't been removed for some reason. Why is the rush to terminate if the rule hasn't really been removed?
> 
> 
> 
> I think that part of the code was tested with teravm gxgy_05 test and gxgy_capacity test
> 
> 
> 
> 
> 
> 
> 
> 
@uri200 
There is no rush. I'm trying to fix ending early here. In the current logic, we only count the rules with non 0 values towards the set of active sessions we have. I'm trying to move that to include 0 value records as well.

Current logic:
If session is in termination AND we receive either no rule record or records with 0 values, we complete termination.

New logic:
If session is in termination and we receive no rule record, we complete termination.

uri200:
Ohh then I got the logic reverse
this is right 
```
New logic:
If session is in termination and we receive no rule record, we complete termination.
```

uri200 (approved):

PR #7436, opened 2021-06-09
Fix Ingress Flow Direction logic for CWF

koolzz:

## Summary

In CWF we saw that the traffic wasn't egressed properly thats due to the passthrough logic for lte. Changing the default flow to be CWF specific

## Additional Information


pshelar (approved):
LGTM

PR #7447, opened 2021-06-09
Fix default rule drop flow metrics

### LINKED ISSUE #7143: sessiond: handle drop all rule stats properly for dropped metrics
### Your Environment

- **Version:** master 
- **Affected Subcomponent:** SessionD

### Describe the Issue
We currently ignore reported stats for `internal_default_drop_flow_rule`, which we need for the dropped tx/rx metrics. 
This feature broke when we migrated to flat rate reporting, tracking the breakage here.

koolzz:

## Summary
Resolves #7143

## Test Plan

make test_session_manager   runs fine                        
## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/SessionState.cpp
+++ b/lte/gateway/c/session_manager/SessionState.cpp
@@ -46,8 +46,8 @@ const char* LABEL_SESSION_ID        = "session_id";
 const char* DROP_ALL_RULE = "internal_default_drop_flow_rule";
 }  // namespace
 
-using magma::service303::increment_counter;
 using magma::service303::remove_counter;
+using magma::service303::set_gauge;
 
 namespace magma {
 
@@ -473,7 +473,6 @@ optional<RuleStats> SessionState::get_rule_delta(
     const std::string& rule_id, uint64_t rule_version, uint64_t used_tx,
     uint64_t used_rx, uint64_t dropped_tx, uint64_t dropped_rx,
     SessionStateUpdateCriteria* session_uc) {
-  // TODO(@koolzz): Handle drop all stats properly GH7143
   if (policy_version_and_stats_.find(rule_id) ==
       policy_version_and_stats_.end()) {
     if (rule_id.compare(DROP_ALL_RULE)) {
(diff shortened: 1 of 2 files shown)


karthiksubraveti (changes requested):
Can you edit these queries, can you replace the method increase -> sum_over_time. Thanks.
https://github.com/magma/magma/blob/faba77bb55a87400dbc89733f6a2caabbafa0b3b/nms/app/packages/magmalte/app/views/subscriber/SubscriberChart.js#L77
```
  const queries = [
    {
      q: `sum(increase(ue_reported_usage{IMSI="${subscriberId}",direction="down"}[${step}]))`,
      color: colors.secondary.dodgerBlue,
      label: 'download',
    },
    {
      q: `sum(increase(ue_reported_usage{IMSI="${subscriberId}",direction="up"}[${step}]))`,
      color: colors.data.flamePea,
      label: 'upload',
    },
  ];
```

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -550,6 +549,10 @@ void SessionState::add_rule_usage(
       rule_id, rule_version, used_tx, used_rx, dropped_tx, dropped_rx,
       session_uc);
   if (!rule_delta) {
+    if (strcmp(rule_id.c_str(), DROP_ALL_RULE) == 0) {
```
themarwhal: Should we compare first, and then skip `get_rule_delta` if it's a drop rule?

karthiksubraveti (commented):
FYI This PR https://github.com/magma/magma/pull/7495#pullrequestreview-682150669 might conflict with your changes.

lte/gateway/c/session_manager/SessionState.cpp
karthiksubraveti: Also this being c++, you can simply add DROP_ALL_RULE as a c++ string and compare `rule_id == DROP_ALL_RULE`, this is much safer than using strcmp.

karthiksubraveti (approved):

uri200 (approved):

themarwhal:
@koolzz + @karthiksubraveti 
do we only want the drop flow value to be gauge not counter?
I think this change may have modified `ue_reported_usage` metric to be gauge not counter as well. Just want to make sure that was intentional. Otherwise, I can revert the change for `ue_reported_usage`.

themarwhal:
Or rather, if we want both to be gauges, is this correct?
https://sourcegraph.com/github.com/magma/magma/-/blob/lte/gateway/c/session_manager/SessionState.cpp?L591
`update_data_metrics(UE_USED_COUNTER_NAME, delta.tx, delta.rx);`


Resolving it here: https://github.com/magma/magma/pull/7495

PR #7452, opened 2021-06-10
[Agw][MME] Remove caller null checks on return of itti_alloc_new_message

andreilee:

## Summary

Function `itti_alloc_new_message(..)` is documented to return NULL values when out of memory. This is untrue as there's an `AssertFatal` call inside which will cause MME to exit when out of memory. Modified the documentation block for this function, and also all callers of the function to no longer do any NULL checks on the return value

## Test Plan

Only build MME

```
.
.
.
[7/7] Completed 'MagmaCore'
```

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/nas/emm/sap/emm_fsm.c
+++ b/lte/gateway/c/core/oai/tasks/nas/emm/sap/emm_fsm.c
@@ -172,6 +172,20 @@ int emm_fsm_set_state(
 
 /****************************************************************************
  **                                                                        **
+ ** Name:    sgs_la_update_requested_handler() **
+ **                                                                        **
+ ** Description: Handles the behaviour of the UE in MME while the          **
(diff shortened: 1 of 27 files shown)


ulaskozat (commented):
Can you put a comment before each function call that it will assert inside the function?

lte/gateway/c/core/oai/tasks/nas/emm/sap/emm_fsm.c
```diff
@@ -170,6 +170,20 @@ int emm_fsm_set_state(
   OAILOG_FUNC_RETURN(LOG_NAS_EMM, RETURNerror);
 }
 
+/****************************************************************************
+ **                                                                        **
+ ** Name:    sgs_la_update_requested_handler() **
+ **                                                                        **
+ ** Description: Handles the behaviour of the UE in MME while the          **
```
ardzoht, line 177: legit warning on the typo

PR #7461, opened 2021-06-10
[orc8r][helm] Fix dependency conflict for orc8r metric chart

quentinDERORY:
## Summary

A new version of the metric helm chart raised an issue in the chart generation

## Test Plan

run the dependency update against the new set of versions and it is working

## Additional Information

Maybe it would be good to run the dependency check when filing a PR

### DIFF
--- a/orc8r/cloud/helm/orc8r/Chart.lock
+++ b/orc8r/cloud/helm/orc8r/Chart.lock
@@ -4,7 +4,7 @@ dependencies:
   version: 0.1.10
 - name: metrics
   repository: ""
-  version: 1.4.22
+  version: 1.4.23
 - name: nms
   repository: ""
   version: 0.1.11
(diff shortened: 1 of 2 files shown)


tmdzk (approved):

karthiksubraveti:
> 
> 
> ## Summary
> 
> 
> 
> A new version of the metric helm chart raised an issue in the chart generation
> 
> 
> 
> ## Test Plan
> 
> 
> 
> run the dependency update against the new set of versions and it is working
> 
> 
> 
> ## Additional Information
> 
> 
> 
> Maybe it would be good to run the dependency check when filing a PR
> 
> 

Thanks for fixing this. Do we need to bump up the version for this change ?

quentinDERORY:
> > ## Summary
> > A new version of the metric helm chart raised an issue in the chart generation
> > ## Test Plan
> > run the dependency update against the new set of versions and it is working
> > ## Additional Information
> > Maybe it would be good to run the dependency check when filing a PR
> 
> Thanks for fixing this. Do we need to bump up the version for this change ?

Good call, just bumped orc8r chart to 1.5.23

karthiksubraveti (approved):

PR #7474, opened 2021-06-10
[feg] Add NH support to FeG 2 GW relay (issue #7473)

### LINKED ISSUE #7473: [feg_relay] find Served AGW Id when NH is configured
When we receive a request from core, FEG sends the request to Orc8r. Then Orc8r finds out the AGW that UE is being served.

When NH is enabled, this functionallity stops working because 
```
servedIds, err := getFegServedIds(gw.GetNetworkId()) 
```
does not return the AGW newtorks served by that whole NH netowrk. 

We need that function to take the `served_nh_ids` and check what agw networks it is serving to find the AGW
```
    "served_nh_ids": [
      "terravm1_feg_network"
    ],
```

emakeev:

## Summary

Currently Neutral Host functionality is only supported on GW->FeG relay (see issue 7473). 
For FeG to GW calls only 'local' served networks are considered. This PR adds support for served NH networks as well.

## Test Plan
Unit tests; test w staging/test orc8r

## Additional Information

### DIFF
--- a/feg/cloud/go/services/feg_relay/servicers/feg_to_gw_relay.go
+++ b/feg/cloud/go/services/feg_relay/servicers/feg_to_gw_relay.go
@@ -128,6 +128,8 @@ func getAllGWSGSServiceConnCtx(ctx context.Context) ([]*grpc.ClientConn, []conte
 	return connList, ctxList, nil
 }
 
+// getFegServedIds returns ServedNetworkIds of the given FeG networkId and appends to the list all ServedNetworkIds of
+// the network's Neutral Host Network if any
 func getFegServedIds(networkId string) ([]string, error) {
 	if len(networkId) == 0 {
 		return []string{}, fmt.Errorf("Empty networkID provided.")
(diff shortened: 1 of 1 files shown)


feg/cloud/go/services/feg_relay/servicers/feg_to_gw_relay.go
```diff
@@ -140,7 +140,29 @@ func getFegServedIds(networkId string) ([]string, error) {
 	if !ok || networkFegConfigs == nil {
 		return []string{}, fmt.Errorf("invalid federation network config found for network: %s", networkId)
 	}
-	return networkFegConfigs.ServedNetworkIds, nil
+	// If this is a NH FeG Network, add served networks from ServedNhIds FeG networks, if not - return ServedNhIds
```
uri200: isnt it possible to have a nh network and still have some gateways direclty connected?

Wouldnt it be better append the results? Meaning if we have nh return the configured local networks and the nh networks

emakeev: yea, that's what it needs to do - it takes takes local Served Network IDs (nids := networkFegConfigs.ServedNetworkIds) & appends Served Network IDs from NH FeGs (nids = append(nids, nhNetworkFegConfigs.ServedNetworkIds...).
The comment is misleading, will fix it

uri200 (approved):

PR #7478, opened 2021-06-10
orc8r: Populate subscribers with MSISDN values

### LINKED ISSUE #3722: Add orc8r MSISDN to subscriber model
[Due: phase-2, so April 2021. But a nice-to-have for debugging purposes.]

The orc8r's MSISDN value should be included in
- the subscriber model (and populated for the subscriber endpoints)
- the subscriber proto (using existing field, and streamed down to gw)

The orc8r MSISDN value is currently a config-only value used to alias an IMSI, so the MSISDN->IMSI lookup only supports that direction. The `SubscriberLookup` service needs to add support for looking up the reverse map (IMSI->MSISDN).

hcgatewood:
## Summary

Resolves #3722.

## Additional Information


hcgatewood:
Closing due to deprio

PR #7479, opened 2021-06-10
[mme] add _get_sgw_s5s8_if_name

uri200:

## Summary
Refactor to separate the logic of getting if name from the _get_context function

## Test Plan
make test

## Additional Information

### DIFF
--- a/lte/gateway/python/scripts/generate_oai_config.py
+++ b/lte/gateway/python/scripts/generate_oai_config.py
@@ -196,13 +196,7 @@ def _get_restricted_imeis(service_mconfig):
         return service_mconfig.restricted_imeis
     return {}
 
-
-def _get_context():
-    """
-    Create the context which has the interface IP and the OAI log level to use.
-    """
-    mme_service_config = load_service_mconfig('mme', MME())
-    nat = _get_enable_nat(mme_service_config)
+def _get_sgw_s5s8_if_name(nat):
     if nat:
         iface_name = get_service_config_value(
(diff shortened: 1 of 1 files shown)


lte/gateway/python/scripts/generate_oai_config.py
```diff
@@ -211,11 +205,26 @@ def _get_context():
         iface_name = get_service_config_value(
             'spgw', 'sgw_s5s8_up_iface_name_non_nat', '',
         )
+    return iface_name
+
+def _get_sgw_s5s8_ip_from_if_cidr(s5s8_if_name):
+    if not s5s8_if_name:
+        return ""
+    return get_ip_from_if_cidr(s5s8_if_name),
```
ardzoht, line 213: extra comma at the end?

lte/gateway/python/scripts/generate_oai_config.py
```diff
@@ -211,11 +205,26 @@ def _get_context():
...
+
+def _get_context():
+    """
+    Create the context which has the interface IP and the OAI log level to use.
+    """
+    mme_service_config = load_service_mconfig('mme', MME())
+    nat = _get_enable_nat(mme_service_config)
```
ardzoht, line 220: not related to your changes, but let's rename `nat` to `nat_enabled` for more meaningful naming

ardzoht (approved):
lgtm, I would suggest before landing your changes, test this with provisioning your VM again just to make sure the generation of OAI config is fine and add this to your test plan, thanks!

Issue #4086, opened 2020-12-16
Remove default in orc8r database storage


mpgermano:
Currently we default to using an in memory db for the orc8r `storage.Database`. This can cause issues for an unaware user that is attempting to persist via a production db.

- Add a new `MustGetEnv` function in orc8r/lib/go/definitions/env.go
- Set SQL driver and DB source in orc8r/cloud/go/storage/storage.go with `MustGetEnv` instead of the "with default" getter

hcgatewood:
Closed by #7484

PR #7484, opened 2021-06-11
Remove default in orc8r database storage

mstre123:
Remove default in orc8r database storage

## Summary

Add a new MustGetEnv function in orc8r/lib/go/definitions/env.go
Set SQL driver and DB source in orc8r/cloud/go/storage/storage.go with MustGetEnv instead of the "with default" getter
Fix tests
## Test Plan
./build.py -t

## Additional Information

#4086

### DIFF
--- a/orc8r/cloud/go/services/state/state/main.go
+++ b/orc8r/cloud/go/services/state/state/main.go
@@ -59,7 +59,7 @@ func main() {
 	if err != nil {
 		glog.Fatalf("Error creating state service %v", err)
 	}
-	db, err := sqorc.Open(storage.SQLDriver, storage.DatabaseSource)
+	db, err := sqorc.Open(storage.GetDatabaseSource(), storage.GetDatabaseSource())
 	if err != nil {
 		glog.Fatalf("Error connecting to database: %v", err)
 	}
@@ -105,7 +105,7 @@ func newIndexerManagerServicer(cfg *config.ConfigMap, db *sql.DB, store blobstor
 	reindexer := reindex.NewReindexer(queue, reindex.NewStore(store))
 	servicer := servicers.NewIndexerManagerServicer(reindexer, autoReindex)
 
-	if autoReindex && storage.SQLDriver != sqorc.PostgresDriver {
+	if autoReindex && storage.GetDatabaseSource() != sqorc.PostgresDriver {
 		glog.Warning(nonPostgresDriverMessage)
 	}
 
--- a/orc8r/cloud/go/services/state/state/main.go
+++ b/orc8r/cloud/go/services/state/state/main.go
@@ -59,7 +59,7 @@ func main() {
 	if err != nil {
 		glog.Fatalf("Error creating state service %v", err)
 	}
-	db, err := sqorc.Open(storage.GetDatabaseSource(), storage.GetDatabaseSource())
+	db, err := sqorc.Open(storage.GetSQLDriver(), storage.GetDatabaseSource())
 	if err != nil {
 		glog.Fatalf("Error connecting to database: %v", err)
 	}
@@ -105,7 +105,7 @@ func newIndexerManagerServicer(cfg *config.ConfigMap, db *sql.DB, store blobstor
 	reindexer := reindex.NewReindexer(queue, reindex.NewStore(store))
 	servicer := servicers.NewIndexerManagerServicer(reindexer, autoReindex)
 
-	if autoReindex && storage.GetDatabaseSource() != sqorc.PostgresDriver {
+	if autoReindex && storage.GetSQLDriver() != sqorc.PostgresDriver {
 		glog.Warning(nonPostgresDriverMessage)
 	}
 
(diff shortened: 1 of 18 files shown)


orc8r/lib/go/definitions/env.go
```diff
@@ -24,3 +27,13 @@ func GetEnvWithDefault(variable string, defaultValue string) string {
...
+
+// MustGetEnv returns the string value of the environment variable,
+// returns error if it doesn't exist
+func MustGetEnv(variable string) (string, error) {
+	value := os.Getenv(variable)
+	if len(value) == 0 {
+		return "", fmt.Errorf("%s env not found", variable)
+	}
+	return value, nil
+}
```
hcgatewood: "Must" prefix indicating we want to panic if it fails. Will panic pretty immediately if the env var isn't set, in all services, so better to be noisy and panic than swallow the err

orc8r/cloud/go/test_utils/test_db.go
```diff
@@ -36,7 +35,7 @@ var (
 // GetSharedMemoryDB returns a singleton in-memory database connection.
 func GetSharedMemoryDB() (*sql.DB, error) {
 	var err error
-	once.Do(func() { instance, err = sqorc.Open(storage2.SQLDriver, ":memory:") })
+	once.Do(func() { instance, err = sqorc.Open("sqlite3", ":memory:") })
```
hcgatewood: I think there's a const for this somewhere in the codebase. If not, let's add one -- I think there's one for postgres you could look for to emulate

hcgatewood (changes requested):
Thanks for the change! Couple small details then ready to go

orc8r/cloud/go/test_utils/test_db.go
mstre123: couldn't find the postgres const so I moved it to the definitions const file

orc8r/lib/go/definitions/const.go
```diff
@@ -23,4 +23,6 @@ const (
 	MetricsdServiceName     = "metricsd"
 	StateServiceName        = "state"
 	StreamerServiceName     = "streamer"
+
+	SQLite3 = "sqlite3"
```
hcgatewood: https://github.com/magma/magma/blob/6aaa8e0f705fc7957dd89ccba3dc2b2c08dce294/orc8r/cloud/go/sqorc/open.go#L30

orc8r/lib/go/definitions/env.go
```diff
@@ -24,3 +27,13 @@ func GetEnvWithDefault(variable string, defaultValue string) string {
 	}
 	return value
 }
+
+// MustGetEnv returns the string value of the environment variable,
+// panics it doesn't exist
+func MustGetEnv(variable string) string {
+	value := os.Getenv(variable)
+	if len(value) == 0 {
+		panic(fmt.Errorf("%s env not found", variable))
```
hcgatewood: Use the github errors pkg

orc8r/cloud/go/services/state/state/main.go
```diff
@@ -105,7 +105,7 @@ func newIndexerManagerServicer(cfg *config.ConfigMap, db *sql.DB, store blobstor
 	reindexer := reindex.NewReindexer(queue, reindex.NewStore(store))
 	servicer := servicers.NewIndexerManagerServicer(reindexer, autoReindex)
 
-	if autoReindex && storage.SQLDriver != sqorc.PostgresDriver {
+	if autoReindex && storage.GetDatabaseSource() != sqorc.PostgresDriver {
```
hcgatewood: Here and elsewhere: some of these are mixing `SQLDriver` and `DatabaseSource`

hcgatewood (changes requested):
Looks good, couple notes

orc8r/cloud/go/services/state/state/main.go
mstre123: good catch went a little overboard with find/replace

orc8r/lib/go/definitions/const.go
```diff
@@ -23,4 +23,6 @@ const (
 	MetricsdServiceName     = "metricsd"
 	StateServiceName        = "state"
 	StreamerServiceName     = "streamer"
+
+	SQLiteDriver = "sqlite3"
```
hcgatewood: Just use https://github.com/magma/magma/blob/6aaa8e0f705fc7957dd89ccba3dc2b2c08dce294/orc8r/cloud/go/sqorc/open.go#L30 instead of creating a new var

hcgatewood (changes requested):
One last aesthetic fix then am ready to merge

orc8r/lib/go/definitions/const.go
mstre123: oh nice didn't see that

orc8r/lib/go/definitions/env.go
```diff
@@ -24,3 +28,13 @@ func GetEnvWithDefault(variable string, defaultValue string) string {
 	}
 	return value
 }
+
+// MustGetEnv returns the string value of the environment variable,
+// panics it doesn't exist
```
hcgatewood, line 33: Nit
```suggestion
// panics it doesn't exist.
```

hcgatewood (approved):
Lgtm, thanks for the fix!

You'll need to re-push (e.g. force-push a noop commit or amend the head commit) to re-kick off the failed flaky cirlceci job

themarwhal (approved):
feg changes looks good!

PR #7490, opened 2021-06-11
AGW: add IRQ CPU affinity configuration

pshelar:
Add IRQ CPU affinity configuration for AGW.
This patch sets default configuration that restricts
datapath processing on two core. This would reduce max
packet per second performance of AGW but it improves
control plane responsiveness

## Test Plan
Validate on lab AGW and vagrant VM.

## Additional Information

### DIFF
--- a/lte/gateway/deploy/roles/magma/files/set_irq_affinity
+++ b/lte/gateway/deploy/roles/magma/files/set_irq_affinity
@@ -1,0 +1,265 @@
+#!/bin/bash
+#
+# Copyright (c) 2015, Intel Corporation
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     * Redistributions of source code must retain the above copyright notice,
+#       this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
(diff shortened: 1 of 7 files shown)


lte/gateway/python/magma/pipelined/datapath_setup.py
```diff
@@ -0,0 +1,90 @@
+"""
```
koolzz, line 1: Should this file be in python/scripts/?

lte/gateway/deploy/roles/magma/files/set_irq_affinity
```diff
@@ -0,0 +1,265 @@
...
+		core=$i
+		set_affinity
+		((n++))
+	done
+}
+
+# now the actual useful bits of code
+
+# these next 2 lines would allow script to auto-determine interfaces
+#[ -z "$IFACES" ] && IFACES=$(ls /sys/class/net)
```
koolzz, line 203: rm ?

koolzz (approved):

lte/gateway/deploy/roles/magma/files/set_irq_affinity
pshelar: I basically copied this script from gist, so I would like to keep it as it is for easier diff in future.

pshelar:
the script is from: https://gist.github.com/SaveTheRbtz/8875474

PR #7513, opened 2021-06-11
orc8r: Improve AGW subscriber sync scalability

hcgatewood:
## Summary

Currently, 20k subscribers and 50 gateways are *just barely* manageable. This change resolves some immediate issues experienced, in preparation for testing 100/200 gateways.

Additional items we will tweak to try to reach 200 gateways as-is:

- 10m sync interval for all gateways
- Better DB: beefier instance, upgrade to latest Postgres version

If we can hit 200 gateways as-is, I have confidence the subscriber digest work will land us near 500 gateways

## Additional Information

### DIFF
--- a/lte/gateway/configs/subscriberdb.yml
+++ b/lte/gateway/configs/subscriberdb.yml
@@ -17,10 +17,10 @@
 print_grpc_payload: False
 
 # Size of subscriber pages returned from cloud
-subscriber_page_size: 3000
+subscriber_page_size: 5000
 
 # Interval in seconds between gateway and cloud sync
-subscriberdb_sync_interval: 180
+subscriberdb_sync_interval: 300  # 5m
 
 # Host address of the Diameter/S6A MME server
 host_address: 0.0.0.0 # Bind to all interfaces
--- a/lte/gateway/python/magma/subscriberdb/main.py
+++ b/lte/gateway/python/magma/subscriberdb/main.py
@@ -11,8 +11,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 import asyncio
+import hashlib
 import logging
 
+import snowflake
 from lte.protos.mconfig import mconfigs_pb2
 from lte.protos.subscriberdb_pb2_grpc import SubscriberDBCloudStub
 from magma.common.grpc_client_manager import GRPCClientManager
@@ -64,7 +66,9 @@ def main():
             service_stub=SubscriberDBCloudStub,
             max_client_reuse=60,
         )
-        sync_interval = service.config.get('subscriberdb_sync_interval')
+        sync_interval = _randomize_sync_interval(
+            service.config.get('subscriberdb_sync_interval'),
+        )
         subscriber_page_size = service.config.get('subscriber_page_size')
         subscriberdb_cloud_client = SubscriberDBCloudClient(
             service.loop,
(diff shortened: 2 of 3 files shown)


lte/gateway/python/magma/subscriberdb/main.py
```diff
@@ -136,5 +140,15 @@ def _get_s6a_manager(service, processor):
     )
 
 
+def _randomize_sync_interval(interval: int) -> int:
+    """_randomize_sync_interval increases sync interval to ameliorate
+    thundering herd at Orc8r."""
+    h = hashlib.md5()
+    h.update(bytes(snowflake.snowflake(), 'utf8'))  # digest of hwid
+    multiplier = (hash(h.hexdigest()) % 100) / 100  # to interval [0, 1]
+    delta = multiplier * (interval / 5)  # up to 1/5 of target interval
```
ulaskozat, line 152: looks a bit convoluted, why not use random?

lte/gateway/configs/subscriberdb.yml
```diff
@@ -17,10 +17,10 @@
 print_grpc_payload: False
 
 # Size of subscriber pages returned from cloud
-subscriber_page_size: 3000
+subscriber_page_size: 5000
 
 # Interval in seconds between gateway and cloud sync
-subscriberdb_sync_interval: 180
+subscriberdb_sync_interval: 300  # 5m
```
ulaskozat, line 23: What is the plan to move to delta updates?

karthiksubraveti: Due to our level triggered semantics, we don't plan to move to delta updates.

lte/gateway/python/magma/subscriberdb/main.py
karthiksubraveti: I agree, using random library might be more straightforward.
```
random.seed(snowflake.snowflake())
random.uniform(interval, interval + (interval/5)) 
```

lte/gateway/configs/subscriberdb.yml
ulaskozat: Remaining within the level triggered semantics, we can still compress how the levels are communicated and when the data of a particular level  is fetched. E.g., AGW can make the RPC call stating what level (e.g., version number) it is at for different buckets of subscribers, then Orc8r can stream down the buckets that have a higher version number. If we want to protect against data corruption, rather than version number, we can also use hash. I thought some level of bucketization (if not Merkle tree based approach) was on the table.

lte/gateway/python/magma/subscriberdb/main.py
```diff
@@ -136,5 +140,15 @@ def _get_s6a_manager(service, processor):
     )
 
 
+def _randomize_sync_interval(interval: int) -> int:
+    """_randomize_sync_interval increases sync interval to ameliorate
+    thundering herd at Orc8r."""
```
ardzoht: Please use google style docstring for the function, that will also fix the python linting comments

karthiksubraveti (approved):
Changes LGTM

lte/gateway/python/magma/subscriberdb/main.py
hcgatewood: To make the delta deterministic. Discussed offline with Karthik, will add a comment to make the intent clearer

lte/gateway/configs/subscriberdb.yml
hcgatewood: We are planning to support a mild form of delta updates, where the delta is computed as the set-difference between the gateway's view of the subs vs. the Orc8r's view of the subs. This will effectively be "level-triggered deltas"

However, this will not be in v1.6, as it's part of Elaine's intern project and won't be finished in time (as planned)

Note: this set-difference is a replacement for the merkle tree construction I'd previously described, since the set-difference is simpler to implement

ardzoht (approved):
lgtm

PR #7517, opened 2021-06-12
Create a super build for LiAgent

amarpad:
## Summary

Change creates a super build for LiAgent
Also, fixes the build script to capture the new updated path for connection tracker

## Test Plan

Build all, verified sym links are correct

## Additional Information


amarpad:
@ymasmoudi looks like build-magma.sh is not copying liagentd?

lte/gateway/c/li_agent/src/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+# we can find memory leaks in production.
+set(CMAKE_C_FLAGS_RELWITHDEBINFO
+    "${CMAKE_C_FLAGS_RELWITHDEBINFO} -fsanitize=leak -fno-omit-frame-pointer")
+set(CMAKE_EXE_LINKER_FLAGS_RELWITHDEBINFO
+    "${CMAKE_EXE_LINKER_FLAGS_RELWITHDEBINFO} -fsanitize=leak")
+
+add_definitions(-DLOG_WITH_GLOG)
+
+# compile the needed macros
+create_proto_dir("lte" LTE_CPP_OUT_DIR)
```
themarwhal, line 47: this is unrelated to your change but i think we don't need `spgw_service` for this service? or do we? or policydb :o

themarwhal:
Does this fix https://github.com/magma/magma/issues/7518 by any chance? (maybe it is unrelated as it happens during session manager build)

lte/gateway/c/li_agent/src/CMakeLists.txt
amarpad: Let me take a look. It seemed like session_manager added a dependency into policydb.

amarpad:
> Does this fix #7518 by any chance? (maybe it is unrelated as it happens during session manager build)

I suspect not.

lte/gateway/c/li_agent/src/CMakeLists.txt
amarpad: ok so mobilityd client brings in subscriberdb which brings in APN. Thats the minimum set of protos we need. Fixed and updated, thanks for the review.

themarwhal: thanks! that was bothering me for a bit :)

koolzz (approved):
cc @ymasmoudi

ardzoht (approved):

themarwhal:
@amarpad locally I don't see this path "${C_BUILD}/connection_tracker/src/src" but I see "${C_BUILD}/connection_tracker/src"

I think this is breaking the LTE deploy job: https://github.com/magma/magma/issues/7601

PR #7526, opened 2021-06-14
fix(pipelined): Use correct ip-addr value to get hash entry

pshelar:

## Test Plan
`make test`

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/rule_mappers.py
+++ b/lte/gateway/python/magma/pipelined/rule_mappers.py
@@ -106,8 +106,9 @@ class SessionRuleToVersionMapper:
             ip_addr_str = ip_addr.address.decode('utf-8').strip()
         with self._lock:
             for k, v in self._version_by_imsi_and_rule.items():
-                _, imsi, ip_addr_str, _ = SubscriberRuleKey(*json.loads(k))
-                if imsi == encoded_imsi and ip_addr_str == ip_addr_str:
+                _, cur_imsi, cur_ip_addr_str, _ = SubscriberRuleKey(*json.loads(k))
+                if cur_imsi == encoded_imsi and (ip_addr_str == "" or
+                                                 ip_addr_str == cur_ip_addr_str):
                     self._version_by_imsi_and_rule[k] = v + 1
 
     def save_version(self, imsi: str, ip_addr: IPAddress,


lte/gateway/python/magma/pipelined/rule_mappers.py
```diff
@@ -101,13 +101,14 @@ def update_all_ue_versions(self, imsi: str, ip_addr: IPAddress):
         """
         encoded_imsi = encode_imsi(imsi)
         if ip_addr is None or ip_addr.address is None:
-            ip_addr_str = ""
+            ip_addr_str = None
```
koolzz: Can you keep it "", when saving I used the empty str (helps with cwf compat)
https://github.com/magma/magma/blob/3f76e5fae724102b30debc5e52e72ca09f40d11d/lte/gateway/python/magma/pipelined/rule_mappers.py#L122

pshelar: sure

lte/gateway/python/magma/pipelined/rule_mappers.py
```diff
@@ -106,8 +106,9 @@ def update_all_ue_versions(self, imsi: str, ip_addr: IPAddress):
             ip_addr_str = ip_addr.address.decode('utf-8').strip()
         with self._lock:
             for k, v in self._version_by_imsi_and_rule.items():
-                _, imsi, ip_addr_str, _ = SubscriberRuleKey(*json.loads(k))
-                if imsi == encoded_imsi and ip_addr_str == ip_addr_str:
+                _, cur_imsi, cur_ip_addr_str, _ = SubscriberRuleKey(*json.loads(k))
+                if cur_imsi == encoded_imsi and (ip_addr_str == "" or
```
koolzz, line 110: I think its better to do `cur_imsi == encoded_imsi and ip_addr_str== cur_ip_addr_str` (we don't want this check to pass when ip_addr_str is "" but cur_ip_addr_str is some ip addr)

koolzz (approved):

lte/gateway/python/magma/pipelined/rule_mappers.py
pshelar: this function needs to update version for all rules for given IMSI when IP address is not specified.
if the `ip_addr_str == ""` match is dropped it would not match any rule with valid IP address. Am I missing something?

koolzz: it updates all version when rule_id is not given/ip address is assumed to be available

PR #7539, opened 2021-06-14
perf(pipeilined): Remove rule map entry in rule Deactivate call

pshelar:
pipelined updates version of the rule on Deactivate call. This could
result in lower performance is case of higher number of UEs.
Deleting the entry improves performance and has same result.

## Test Plan
`make test`

## Additional Information


koolzz (approved):
lgtm, I think with the new save/version from sessiond this is a safe change. @themarwhal just tagging you here for reference

themarwhal:
> lgtm, I think with the new save/version from sessiond this is a safe change. @themarwhal just tagging you here for reference

This seems safe I think. When we delete the rules, we probably don't care about the version anymore in PipelineD.

PR #7547, opened 2021-06-14
[agw][cloudstrapper] Export AGW AMI to qcow2

quentinDERORY:
## Summary

Added the convertion to qcow2 to the agw ami pipeline.
The script will export it in qcow2 format in a pre-configured bucket: qcow2images

## Test Plan

Tested the pipeline in circle ci https://app.circleci.com/pipelines/github/magma/magma/26537/workflows/b623c96b-96cb-492a-bf6a-b0261626e4e0/jobs/283388

## Additional Information

Disabling default EBS encryption if enabled to permit export of ami.
Using the default bucket to store the intermediate vmdk file.
Got rid of failing cleaning commands.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1477,7 +1477,7 @@ jobs:
           name: Propagate names for AWS essential components
           command: |
             sed -i -e '/^secgroupDefault: /s/:.*$/: publish-amis-to-marketplace-secgroup/' $VARS_DIR/defaults.yaml
-            sed -i -e '/^bucketDefault: /s/:.*$/: publish-amis-to-marketplace-bucket/' $VARS_DIR/defaults.yaml
+            sed -i -e '/^bucketDefault: /s/:.*$/: publish-amis-to-marketplace-bucket2/' $VARS_DIR/defaults.yaml
             sed -i -e '/^stackEssentialsDefault: /s/:.*$/: publish-amis-to-marketplace-stack/' $VARS_DIR/defaults.yaml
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -1477,7 +1477,7 @@ jobs:
           name: Propagate names for AWS essential components
           command: |
             sed -i -e '/^secgroupDefault: /s/:.*$/: publish-amis-to-marketplace-secgroup/' $VARS_DIR/defaults.yaml
-            sed -i -e '/^bucketDefault: /s/:.*$/: publish-amis-to-marketplace-bucket/' $VARS_DIR/defaults.yaml
+            sed -i -e '/^bucketDefault: /s/:.*$/: publish-amis-to-marketplace-bucket2/' $VARS_DIR/defaults.yaml
```
arunuke, line 1480: Since bucket names are globally unique, is there a way to add a check that validates if bucket already exists and has write permissions for the circle-ci user?

arunuke (changes requested):

.circleci/config.yml
quentinDERORY: Just added  existence and write permission check for the export bucket but not for the default bucket.
The default bucket is created by the playbooks and playbooks will fail if it already exists.

arunuke: LGTM.

arunuke (approved):
LGTM on the check for buckets.

PR #7550, opened 2021-06-15
[component:agw] Fix for NAS Decode failure on GUTI

panyogesh:
Fixes:
  1. Fix for NAS decode failure on getting GUTI request
     from TerraVM simulator.
  2. Fix for LTE test failures.

Testing:
  1. Tested basic flow with simulator.
  2. Run unit test of ngap and amf.

## Additional Information


lte/gateway/c/oai/tasks/nas5g/src/M5GRegistrationRequest.cpp
```diff
@@ -119,10 +119,16 @@ int RegistrationRequestMsg::DecodeRegistrationRequestMsg(
...
-        DECODE_U8(buffer + decoded + type_len, decoded_result, decoded);
+        if (type == REGISTRATION_REQUEST_NAS_MESSAGE_CONTAINER_TYPE) {
+          type_len   = sizeof(uint8_t);
+          length_len = 2 * sizeof(uint8_t);
+          DECODE_U16(buffer + decoded + type_len, decoded_result, decoded);
+        } else {
+          type_len   = sizeof(uint8_t);
+          length_len = sizeof(uint8_t);
+          DECODE_U8(buffer + decoded + type_len, decoded_result, decoded);
+        }
```
ulaskozat, line 131: rather than adding another if/else, you should fix the switch statement so that this case is handled properly.

PR #7578, opened 2021-06-15
modify stop command to also stop sctpd service

themarwhal:

## Summary
`make stop` currently only stops `magma@*` but not the `sctpd` service. Adding sctpd to the make command.

## Test Plan
ran `make stop`

## Additional Information


ulaskozat:
@themarwhal we may want to keep this behavior.

themarwhal:
> @themarwhal we may want to keep this behavior.

if we want this we can keep it around. I just assumed `make stop` should stop everything. :)

PR #7579, opened 2021-06-15
Don't build protos in common

amarpad:
## Summary

Don't build protos in common

This is not needed as the specific super builds will build protos
Also, its probably not common if we are building specific service protos

## Test Plan

make run

## Additional Information


amarpad:
Sorry re-creating this pull request as the previous one was on an upstream branch.

uri200:
I had to use debian vm because ubuntu is taking forever.

With debian I can still run the test and use the debugger using the ide

themarwhal (approved):

PR #7583, opened 2021-06-15
[lte][agw] Update timer id for expired NAS procedure timers

ulaskozat:

## Summary

- After moving the image to Ubuntu, `test_attach_detach_duplicate_nas_resp_messages.py` started failing. The PR updates the timer IDs in the handler functions for expired timers of NAS procedures. This prevents unnecessary stoppage of already expired one-off timers. 
- We still need to root cause why (presumed to be) idempotent "stop_timer" call is causing an issue for already expired timers.

## Test Plan

Run the failing integ tests and the full test suite.

## Additional Information


ardzoht (approved):
lgtm

PR #7584, opened 2021-06-16
[orc8r][lte]: Add sentry.io configuration models and API endpoints

waqaraqeel:

## Summary
Working towards #7613.

This PR adds the models and API endpoints for getting and putting sentry.io configuration params as the first step in #7613. The config params are not actually used yet.

## Test Plan

Ran `./build.py --test`. Ran swagger UI locally. Tested by doing PUT and GET on `/networks/{network_id}/sentry`. Verified that records were created/updated in `postgresql`.


lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -165,6 +165,7 @@ func (s *builderServicer) Build(ctx context.Context, request *builder_protos.Bui
 			ServiceAreaMaps:          getServiceAreaMaps(nwEpc.ServiceAreaMaps),
 			FederatedModeMap:         getFederatedModeMap(federatedNetworkConfigs),
 			CongestionControlEnabled: swag.BoolValue(gwEpc.CongestionControlEnabled),
+			SentryConfig:             &lte_mconfig.SentryConfig{},
```
themarwhal: I'm guessing you'll have a separate PR for actually filling these out right? We can just leave the field unset for any empty values. (proto serialization will omit them anyways)

lte/protos/mconfig/mconfigs.proto
```diff
@@ -18,6 +18,16 @@ import "orc8r/protos/common.proto";
 package magma.mconfig;
 option go_package = "magma/lte/cloud/go/protos/mconfig";
 
+// --------------------------------------------------------------------------
+// SentryConfig stores the network-wide sentry.io configuration
+// --------------------------------------------------------------------------
+message SentryConfig {
+    string url_python = 1;
```
themarwhal, line 27: Let's add the same documentation strings from the config file

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
```diff
@@ -5229,6 +5229,36 @@ paths:
...
+      responses:
+        "200":
+          description: Network-wide Sentry.io configuration
+          schema:
+            $ref: '#/definitions/network_sentry_config'
+        default:
+          $ref: '#/responses/UnexpectedError'
+      summary: Get the Sentry.io configuration for network
+      tags:
+      - Networks
```
themarwhal, line 5245: is the linter warning just wrong? or does yaml not care about indententation?

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
```diff
@@ -10295,6 +10325,31 @@ definitions:
     required:
     - bandwidth_mhz
     type: object
+  network_sentry_config:
+    description: Sentry.io configuration
+    properties:
+      sample_rate:
+        example: 0.5
+        format: float
+        maximum: 1
```
themarwhal, line 10334: 👍

themarwhal (approved):
looks good to me! :)

themarwhal:
Mind the `insync-checkin` job failure. You'll just have to run `./build.py -g` and commit everything again.

lte/cloud/go/services/lte/servicers/builder_servicer.go
waqaraqeel: Yes I'll have it in the next PR.

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
waqaraqeel: That's interesting... This is a generated file. In the manually edited `orc8r/cloud/go/services/orchestrator/obsidian/models/swagger.v1.yml`, list items are indented correctly. But this generated file has unindented lists throughout. It's a `combine_swagger` issue I think.

orc8r/cloud/go/services/orchestrator/obsidian/models/conversion.go
```diff
@@ -75,6 +75,14 @@ func (m *NetworkFeatures) ToUpdateCriteria(network configurator.Network) (config
 	return GetNetworkConfigUpdateCriteria(network.ID, orc8r.NetworkFeaturesConfig, m), nil
 }
 
+func (m *NetworkSentryConfig) GetFromNetwork(network configurator.Network) interface{} {
```
electronjoe, line 78: Nit and possibly not worth addressing - but it appears to me that codecov.io says this function, and the ToUpdateCriteria - lack unit test coverage. Is it hard to generate coverage?

electronjoe: Also FWIW there is a plugin for this. If you install Sourcegraph plugin for Chrome. There is a sub-feature for sourcegraph plugin in chrome that enables codecov.io coverage highlighting for PRs. See [info here](https://docs.codecov.com/docs/browser-extension).

orc8r/cloud/go/services/orchestrator/obsidian/models/validate.go
```diff
@@ -36,6 +36,10 @@ func (m *NetworkFeatures) ValidateModel() error {
 	return m.Validate(strfmt.Default)
 }
 
+func (m *NetworkSentryConfig) ValidateModel() error {
+	return m.Validate(strfmt.Default)
```
electronjoe, line 40: This line appears uncovered as well - in unit tests? Per codecov plugin via sourcegraph extension for chrome. Though as I look at this I'm wondering what it's doing right now. Is it aplaceholder function?

waqaraqeel: It is calling generated Go code to validate fields based on the YAML model declaration, e.g., `sample_rate` must be between 0 and 1.

The `PartialNetworkModel` interface in `network_handler_factory.go` requires these three methods, `GetFromNetwork`, `ToUpdateCriteria` and `ValidateModel`.

orc8r/cloud/go/services/orchestrator/obsidian/models/conversion.go
waqaraqeel: Thanks for pointing this out and suggesting the extension. It's quite helpful. I'll make sure to fully test-cover future PRs. 

It turns out that integrating this new field (SentryConfig) with existing tests of the Network model is non-trivial. I'm now working on another PR that does that along with actually filling out the mconfig objects. Thanks!

lte/cloud/go/services/lte/servicers/builder_servicer_test.go
```diff
@@ -1156,6 +1168,7 @@ func TestBuilder_Build_MMEPool(t *testing.T) {
 			CloudSubscriberdbEnabled: false,
 			AttachedEnodebTacs:       nil,
 			NatEnabled:               true,
+			SentryConfig:             &lte_mconfig.SentryConfig{},
```
hcgatewood: Here and throughout: just have this be implicitly nil

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
hcgatewood: This file is generated. The output is correct, yaml is weird

lte/protos/mconfig/mconfigs.proto
```diff
@@ -18,6 +18,21 @@ import "orc8r/protos/common.proto";
...
 option go_package = "magma/lte/cloud/go/protos/mconfig";
 
+// --------------------------------------------------------------------------
+// SentryConfig stores the network-wide sentry.io configuration
+// --------------------------------------------------------------------------
+message SentryConfig {
+    // If set, the Sentry Python SDK will be initialized for all python services
+    string url_python = 1;
+    // If set, the Sentry Native SDK will be initialized for MME and SessionD
+    string url_native = 2;
```
hcgatewood: Everywhere: use golang-style doc comments
```suggestion
    // url_native causes the Sentry Native SDK to be initialized for MME and SessionD,
    // when non-empty
    string url_native = 2;
```

lte/cloud/go/services/lte/servicers/builder_servicer.go
hcgatewood: Okay so, big question for @themarwhal and @waqaraqeel: will the SentryConfig ever be different *per-service*? Because if *no*, then we shouldn't be sending down copies of the same config to each service that needs it (since, you know, each service will need it)

And we *definitely* shouldn't be sending it down to just 1 arbitrary+non-core service then having that value apply across multiple services, especially cross-language -- so like, don't send it to `mme` then apply it to `magmad` and `bootstrapper`, that violates the module dependency tree https://magma.github.io/magma/docs/next/orc8r/dev_dependencies

So if it is the case that we really just want 1 normalized SentryConfig sent down, perhaps it's time to support per-module pseudo services. In this case, either use an `orc8r` pseudo service to hold the SentryConfig, or, if you want to maintain backwards compat, use the `magmad` service

But I think it's a messy, messy idea to mix and match like (it seems) we're doing now. Lmk if I'm misunderstanding though 🙂

hcgatewood (changes requested):
One main request for clarification

lte/cloud/go/services/lte/servicers/builder_servicer.go
themarwhal: Hey @hcgatewood! To your first point, the plan for now was to only stream the config to MME and SessionD services only for now. They are the services that have been benefitting most from Sentry, so I don't want to enable it for the Python services just yet. (Which brings me to think that we should probably remove `url_python` from the swagger API to not expose it until we start using it.)

To your second point, we are not aiming to send it to one service and apply it to multiple.

My thought is, if we keep going with this approach of sending the config down per-service, we should split up the mconfig object so that we only send what we need for each service.
`SessionD` -> `url_native` and `sample_rate`
`MME` -> `url_native`, `sample_rate`, and `upload_mme_log`

I also thought about sending the sentry config under a global?non-service specific key, but since we don't use the config everywhere yet, it may not be worth the trouble just yet.

lte/cloud/go/services/lte/servicers/builder_servicer_test.go
waqaraqeel: Took these placeholders out.

lte/protos/mconfig/mconfigs.proto
waqaraqeel: Improved comments.

lte/cloud/go/services/lte/servicers/builder_servicer.go
waqaraqeel: I agree @themarwhal. It would make sense to move to `magmad` or another cross-service config solution when we enable Sentry for all/more services. We could eliminate the unified `SentryConfig` proto structure and send individual fields to services, but I think the unified structure is actually cleaner unless we care about every byte of config we send.

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
```diff
@@ -10295,6 +10325,31 @@ definitions:
...
+      upload_mme_log:
+        example: false
+        type: boolean
+      url_native:
+        example: https://examplePublicKey@o0.ingest.sentry.io/0
+        format: uri
+        minLength: 0
+        type: string
+        x-nullable: false
+      url_python:
```
themarwhal, line 10346: Let's remove this for now? So that we don't have a field that doens't do anything.

waqaraqeel: I think it's okay to keep it... Once we have a cross-service version of SentryConfig, we'll need to bring this field back in. The API endpoints and tests will be the same too. I think it's a nice, encapsulated config struct that needs minimal change (only on the `mconfig` side) to eliminate per-service copies. Let me know if you think otherwise.

hcgatewood (approved):
For now, replicating the sentry config to each service is fine

I still think though, we should've a priori decided how we would like this to look. If we just send one sentry config per service, that's fine. But if we're trying to coalesce into a "global" config at any point, it's going to be harder to do (from backwards compat perspective) if we don't think through it ahead of time cc @themarwhal

PR #7591, opened 2021-06-16
fix(pipelined): improve check to handle missing  ethtool

pshelar:
This fixes the pipelined crash[1] on debian, but does not
enable the tuning. we can do it once its verified it on debian.

[1] https://gist.github.com/themarwhal/dd2d9a90f72f2997db842f6359f8b1e2

## Test Plan
varied it on debian and ubuntu AGW.

## Additional Information


ulaskozat (approved):
LGTM

PR #7592, opened 2021-06-16
[Orc8r] Adding APN swagger changes and skeleton code for servicer and api handlers

karthiksubraveti:

## Summary
Added swagger changes for gateway_apn_configs
Defined following new REST API endpoints

Screen Shot 2021-06-15 at 9 43 12 PM

## Test Plan
Tested the APIs with skeleton handlers. 

## Additional Information


lte/protos/apn.proto
```diff
@@ -80,17 +82,15 @@ message APNConfiguration {
 // Network wide APN configuration
 message APNConfig {
   uint32 id = 1;
-  // Contains either the APN Name or wildcard "*"
-  string service_selection = 2;
```
ulaskozat, line 84: why did you remove the service_selection? This is used to name the APN (e.g., "magma.ipv4", "ims", "internet", etc.)

lte/protos/apn.proto
```diff
@@ -80,17 +82,15 @@ message APNConfiguration {
...
+  QoSProfile qos_profile = 2;
+  AggregatedMaximumBitrate ambr = 3;
   enum PDNType {
       IPV4 = 0;
       IPV6 = 1;
       IPV4V6 = 2;
       IPV4_OR_IPV6 = 3;
   }
-  PDNType pdn = 5;
+  PDNType pdn = 4;
```
ulaskozat, line 93: we do not renumber the legacy fields for backwards compatibility. use deprecated for fields that are no longer used.

lte/protos/apn.proto
karthiksubraveti: Earlier APNConfig was housed within a subscriber and it made sense to name the APN. This is the new config model for APN which is independently streamed down, the service_selection attribute needn't be a part of the APN config. In subscriber config, I added a new attribute
```
message SubscriberAPNConfig {
  uint32 apn_id = 1;
  string assigned_static_ip = 2;
}
```
subscriber_apn_config . This does the apn selection and also provides any additional APN specific subscriber configuration.

lte/protos/apn.proto
karthiksubraveti: This is not yet being used. This is part of the new proto struct I'm adding as a part of the feature. Hence did the renumbering.

lte/protos/apn.proto
```diff
@@ -123,12 +123,12 @@ message GatewayAPNConfig {
     APNFilter apn_filter = 2;
```
ulaskozat, line 123: who generates apn_id? how is it used?

What is the purpose of APNFilter?

karthiksubraveti: apn_id(Its incorrectly marked as u32, will change this to string) identifies the network wide APN entity. APNFilter is just a future feature, where we can customize the APN specific behavior per gateway on whether we choose to permit or block. I can remove it if you feel its a distraction.

lte/protos/apn.proto
ulaskozat: let's not change the field names such as  service_selection or context_id otherwise we will lose parity on s6a interface in federated vs. non-federated cases.

ulaskozat: currently subscriberdb (in non-federated case) generates the context_id (uint32), but actually it makes more sense for the orc8r to assign it as well.

themarwhal:
@karthiksubraveti Is this PR still up to date / open? Seeing merge conflicts + check failures.

hcgatewood:
Closing this PR till @karthiksubraveti returns from paternity leave

PR #7597, opened 2021-06-16
Sending PLMN of mme within serving network in Create Session Request

rsarwad:
[lte][agw] : Sending PLMN of mme within serving network in Create Session Request

## Summary
Sending PLMN of mme within serving network in Create Session Request
## Test Plan


ulaskozat (approved):
LGTM

PR #7614, opened 2021-06-17
fix(agw): Change the location where metrics scraping callback is added

karthiksubraveti:

## Summary

In the latest set of changes introduced by PR https://github.com/magma/magma/pull/7431, we added the ability to chunk grpc messages for the metrics collected from scrape targets. This however schedules same target to be scraped in the future done callback for each of the chunks. This is unnecessary and aggressive. This change moves the scrape target callback to be set in the original routine instead of scrape_done callback. 

## Additional Information

### DIFF
--- a/orc8r/gateway/python/magma/magmad/metrics_collector.py
+++ b/orc8r/gateway/python/magma/magmad/metrics_collector.py
@@ -282,6 +282,11 @@ class MetricsCollector(object):
                     self.scrape_done, future, target,
                 ),
             )
+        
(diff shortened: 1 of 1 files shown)


ardzoht (approved):

themarwhal (approved):
can you run the format command? otherwise, looks good to me.

hcgatewood (approved):

hcgatewood:
Force-merging over semantic commit failure because Karthik is ooo

PR #7502, opened 2021-06-11
[NMS]Federation network dashboard

ato-dani:
## Summary
The Events - Alerts graph, Alerts KPI, and Events table all were reused from the LTE dashboard without any change since they don't depend on the type of the network. For the Events KPI, currently it only has servicing access gateways kpi which shows how many gateways are serviced by this network. Since the serviced access gateways has the potential to be a very large number, a useEffect hook was used to get the information that we only need instead of sharing it using context providers. However a context provider would be used in the future when sharing information about the federation gateways(This is not currently done but will be pushed next week) . 

## Test Plan
As said earlier, the Events - Alerts graph, Alerts KPI and Events Table didn't need any new test since they were reused from the Lte network dashboard and have been tested. For the new component which is servicing gateway kpi, a test was added to fully test it, in the KPI-test file and it works properly. As usual, previous tests were run and it works there too.

## Additional Information

https://user-images.githubusercontent.com/34602743/121740454-19812d00-cacb-11eb-93df-8b461146610a.mov

PR #7641, opened 2021-06-17
[NMS]Federation Network: Network page - Displaying Network information

ato-dani:
## Summary
This is a draft PR because it is a stacked pr upon an exisiting open PR #7502. Once, that PR is approved, this PR would get rebased and would be changed from draft to a normal PR. A network page was created for the federation network. The network page was hooked up to an existing JSON Edit component for editing the network. A new Network info component was created which simply displays the id, description, name and federation of the network. Also, a context provider was created for sharing the network information between different pages instead of fetching the data.

## Test Plan
A test was added under NetworkTest.js file to test that the new network info component displays correct information about the network. 

## Additional Information
Screen Shot 2021-06-17 at 4 00 16 PM
Screen Shot 2021-06-17 at 4 00 38 PM

### DIFF
--- a/nms/app/packages/magmalte/app/components/feg/FEGContext.js
+++ b/nms/app/packages/magmalte/app/components/feg/FEGContext.js
@@ -1,0 +1,94 @@
+/**
+ * Copyright 2020 The Magma Authors.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @flow strict-local
+ * @format
+ */
+
+import * as React from 'react';
+import FEGNetworkContext from '../context/FEGNetworkContext';
+import LoadingFiller from '@fbcnms/ui/components/LoadingFiller';
+import MagmaV1API from '@fbcnms/magma-api/client/WebClient';
+import useMagmaAPI from '@fbcnms/ui/magma/useMagmaAPI';
+
+import type {feg_network, network_id, network_type} from '@fbcnms/magma-api';
+
+import {UpdateNetworkState as UpdateFegNetworkState} from '../../state/feg/NetworkState';
+import {useCallback, useState} from 'react';
+import {useEnqueueSnackbar} from '@fbcnms/ui/hooks/useSnackbar';
+
+type Props = {
+  networkId: network_id,
+  networkType: network_type,
+  children: React.Node,
+};
+
+/**
+ * Fetches and returns information about the federation network inside
+ * a context provider.
+ * @param {object} props: contains the network id and its type
+ */
+export function FEGNetworkContextProvider(props: Props) {
+  const {networkId} = props;
+  const [fegNetwork, setFegNetwork] = useState<$Shape<feg_network>>({});
+  const enqueueSnackbar = useEnqueueSnackbar();
+  const {error, isLoading} = useMagmaAPI(
+    MagmaV1API.getFegByNetworkId,
+    {networkId: networkId},
+    useCallback(response => setFegNetwork(response), []),
+  );
+
+  if (error) {
+    enqueueSnackbar?.('failed fetching network information', {
+      variant: 'error',
+    });
+  }
+
+  if (isLoading) {
+    return <LoadingFiller />;
+  }
+
+  return (
+    <FEGNetworkContext.Provider
+      value={{
+        state: fegNetwork,
+        updateNetworks: props => {
+          let refreshState = true;
+          if (networkId !== props.networkId) {
+            refreshState = false;
+          }
+          return UpdateFegNetworkState({
+            networkId,
+            setFegNetwork,
+            refreshState,
+            ...props,
+          });
+        },
+      }}>
+      {props.children}
+    </FEGNetworkContext.Provider>
+  );
+}
+/**
+ * A context provider for federation networks. It is used in sharing
+ * information like the network information or the gateways information.
+ * @param {object} props contains the network id and its type
+ */
+export function FEGContextProvider(props: Props) {
+  const {networkId, networkType} = props;
+
+  return (
+    <FEGNetworkContextProvider {...{networkId, networkType}}>
+      {props.children}
+    </FEGNetworkContextProvider>
+  );
+}
--- a/nms/app/packages/magmalte/app/components/feg/FEGSections.js
+++ b/nms/app/packages/magmalte/app/components/feg/FEGSections.js
@@ -23,6 +23,8 @@ import FEGConfigure from './FEGConfigure';
 import FEGDashboard from '../../views/dashboard/feg/FEGDashboard';
 import FEGGateways from './FEGGateways';
 import FEGMetrics from './FEGMetrics';
+import FEGNetworkDashboard from '../../views/network/FEGNetworkDashboard';
+import NetworkCheckIcon from '@material-ui/icons/NetworkCheck';
 import React from 'react';
 import SettingsCellIcon from '@material-ui/icons/SettingsCell';
 import ShowChartIcon from '@material-ui/icons/ShowChart';
@@ -34,6 +36,12 @@ export function getFEGSections(dashboardV2Enabled: boolean): SectionsConfigs {
       label: 'Gateways',
       icon: <CellWifiIcon />,
       component: FEGGateways,
+    },
+    {
+      path: 'network',
+      label: 'Network',
+      icon: <NetworkCheckIcon />,
+      component: FEGNetworkDashboard,
     },
     {
       path: 'configure',
--- a/nms/app/packages/magmalte/app/components/main/Index.js
+++ b/nms/app/packages/magmalte/app/components/main/Index.js
@@ -16,10 +16,13 @@
 
 import MagmaV1API from '@fbcnms/magma-api/client/WebClient';
 
+import {FEG} from '@fbcnms/types/network';
+import {FEGContextProvider} from '../feg/FEGContext';
 import {LteContextProvider} from '../lte/LteContext';
 import {coalesceNetworkType} from '@fbcnms/types/network';
 import type {NetworkType} from '@fbcnms/types/network';
 import type {Theme} from '@material-ui/core';
+import type {network_id, network_type} from '@fbcnms/magma-api';
 
 import * as React from 'react';
 import AppContent from '../layout/AppContent';
@@ -54,6 +57,12 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
+type Props = {
+  networkId: network_id,
+  networkType: network_type,
+  children: React.Node,
+};
+
 export default function Index() {
   const classes = useStyles();
   const {match} = useRouter();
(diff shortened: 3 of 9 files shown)


ato-dani:
@andreilee This is the network page part 1 PR

nms/app/packages/magmalte/app/views/network/FEGNetworkInfo.js
```diff
@@ -0,0 +1,60 @@
...
+    [
+      {
+        category: 'Description',
+        value: props.fegNetwork.description || '-',
+      },
+    ],
+    [
+      {
+        category: 'Federation',
+        value: props.fegNetwork.id,
```
uri200: why is `props.fegNetwork.id,` two times, here and in line 37?

nms/app/packages/magmalte/app/views/network/FEGNetworkInfo.js
```diff
@@ -0,0 +1,60 @@
...
+      },
+    ],
+  ];
+  return <DataGrid data={kpiData} testID="feg_info" />;
+}
```
uri200, line 70: another item that it would be interesting to show is inside `federation` key

```
    "served_network_ids": [
      "terravm2_inbound_agw_network"
    ],
    "served_nh_ids": [
      "terravm2_feg_network"
    ],
```

`served_network_ids` indicates the LTE federated networks this federation serves.

`served_nh_ids` indicates the Neutral Host (or Virtual) Fedration Networks it serves

Both are array of strings.

nms/app/packages/magmalte/app/views/network/FEGNetworkInfo.js
ato-dani: Because the network ID is the same as the federation network id, since it is already a federation network. And on the design doc, they still wanted it to be included even though they are the same thing. This would have been different if it was a feg_lte network because the network id is different from the federation network id(feg_network_id) for that case.

nms/app/packages/magmalte/app/views/network/FEGNetworkInfo.js
ato-dani: Added them.
Screen Shot 2021-06-25 at 9 46 41 AM

nms/app/packages/magmalte/app/components/main/Index.js
```diff
@@ -83,21 +84,23 @@ export default function Index() {
...
-              isSuperUser: user.isSuperUser,
-              ssoEnabled,
-            })}
-            user={user}
-          />
-          <AppContent>
-            <SectionRoutes />
-          </AppContent>
-        </div>
+        <FEGContextProvider networkId={networkId} networkType={networkType}>
```
andreilee, line 126: Semantically, it seems a little strange to me for the `FEGContextProvider` to be nested inside the `LteContextProvider`. I think it makes sense to provide a `NetworkContextProvider` that returns either one, though that might be too much extra code.

nms/app/packages/magmalte/app/views/network/FEGNetworkInfo.js
andreilee: If the design mock-up doesn't make sense, we should either change it, or perhaps get some clarification on what was intended.

andreilee (changes requested):
Update with clarification from design so we don't have the federation network ID presented twice in the same form

nms/app/packages/magmalte/app/components/main/Index.js
ato-dani: I did that, but yes there is some extra code. However, it is semantically correct now.

ato-dani: Also removed federation category.

nms/app/packages/magmalte/app/components/feg/FEGContext.js
```diff
@@ -0,0 +1,103 @@
...
+
+import * as React from 'react';
+import FEGNetworkContext from '../context/FEGNetworkContext';
+import LoadingFiller from '@fbcnms/ui/components/LoadingFiller';
+import MagmaV1API from '@fbcnms/magma-api/client/WebClient';
+
+import type {feg_network, network_id, network_type} from '@fbcnms/magma-api';
+
+import {UpdateNetworkState as UpdateFegNetworkState} from '../../state/feg/NetworkState';
+
```
karthiksubraveti: minor nit: remove newline here

nms/app/packages/magmalte/app/components/feg/FEGContext.js
```diff
@@ -0,0 +1,103 @@
...
+ * @flow strict-local
+ * @format
+ */
+
+import * as React from 'react';
+import FEGNetworkContext from '../context/FEGNetworkContext';
+import LoadingFiller from '@fbcnms/ui/components/LoadingFiller';
+import MagmaV1API from '@fbcnms/magma-api/client/WebClient';
+
+import type {feg_network, network_id, network_type} from '@fbcnms/magma-api';
```
karthiksubraveti, line 23: move "import type". to the top

nms/app/packages/magmalte/app/components/feg/FEGContext.js
```diff
@@ -0,0 +1,103 @@
...
+
+import {useEffect, useState} from 'react';
+import {useEnqueueSnackbar} from '@fbcnms/ui/hooks/useSnackbar';
+
+type Props = {
+  networkId: network_id,
+  networkType: network_type,
+  children: React.Node,
+};
+import {FEG} from '@fbcnms/types/network';
```
karthiksubraveti: move this to be with rest of the imports

nms/app/packages/magmalte/app/components/feg/FEGContext.js
```diff
@@ -0,0 +1,103 @@
...
+ * @param {object} props: contains the network id and its type
+ */
+export function FEGNetworkContextProvider(props: Props) {
+  const {networkId} = props;
+  const [fegNetwork, setFegNetwork] = useState<$Shape<feg_network>>({});
+  const [isLoading, setIsLoading] = useState(true);
+  const enqueueSnackbar = useEnqueueSnackbar();
+
+  useEffect(() => {
+    const fetchState = async () => {
```
karthiksubraveti: If we are just setting Feg Network we can use useMagmaAPI instead. https://github.com/facebookincubator/fbc-js-core/blob/886a66f2e4934f03c6b5e16ec7d0e5fcd036a536/fbcnms-packages/fbcnms-ui/magma/useMagmaAPI.js

nms/app/packages/magmalte/app/components/feg/FEGContext.js
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * A context provider for federation networks. It is used in sharing
+ * information like the network information or the gateways information.
+ * @param {object} props contains the network id and its type
+ */
+export function FEGContextProvider(props: Props) {
+  const {networkId, networkType} = props;
+  const fegNetwork = networkType === FEG;
+  if (!fegNetwork) {
+    return props.children;
```
karthiksubraveti: Is this a valid condition?

nms/app/packages/magmalte/app/components/feg/FEGSections.js
```diff
@@ -63,6 +65,12 @@ export function getFEGSections(dashboardV2Enabled: boolean): SectionsConfigs {
       icon: <DashboardIcon />,
       component: FEGDashboard,
     });
+    sections.splice(1, 0, {
```
karthiksubraveti: Let's add this under default itself.

nms/app/packages/magmalte/app/views/network/FEGNetworkDashboard.js
```diff
@@ -0,0 +1,155 @@
...
+ * Returns information about the federation network and a table of the servicing
+ * access gateways alongside the serviced networks they are under.
+ */
+export function NetworkDashboardInternal() {
+  const classes = useStyles();
+  const ctx = useContext(FEGNetworkContext);
+
+  return (
+    <div className={classes.dashboardRoot}>
+      <Grid container spacing={4}>
```
karthiksubraveti, line 145: Do we need these Grid containers ?

nms/app/packages/magmalte/app/views/network/FEGNetworkInfo.js
karthiksubraveti: @uri200 Can you suggest right hints to be added with each of these items. It will be extremely confusing for an average user to understand what served federated LTE network ID is and served virtual federation network ID

karthiksubraveti (commented):
Left a few comments here. Thanks.

nms/app/packages/magmalte/app/components/feg/FEGContext.js
ato-dani: I tried to change it so that we use useMagmaAPI, but it turns out that you can't use that if you are setting the response you get from that function into a react state variable because after you get the response and set that response to a react state variable it re-renders the page and this process goes on and on making the page to rerender infinite times. If I was using this in a simple page(like in Subscribers.js file), and I dont need to save the response into a state variable this was fine but since I need to set the response to a state variable because this is a context provider, unfortunately, I can't use that function.

nms/app/packages/magmalte/app/views/network/FEGNetworkDashboard.js
MightyDani: Oh, I forgot to remove that once I created a network context provider. Since the network context provider now take care of that, I have removed it.

nms/app/packages/magmalte/app/components/feg/FEGContext.js
MightyDani: Oh, I forgot to remove that once I created a network context provider. Since the network context provider now take care of that, I have removed it.

ato-dani: Oh, I forgot to remove that once a network context provider was added. It is not valid anymore since the network context provider takes care of that. I have removed it. Thanks!

nms/app/packages/magmalte/app/components/feg/FEGContext.js
karthiksubraveti: > I tried to change it so that we use useMagmaAPI, but it turns out that you can't use that if you are setting the response you get from that function into a react state variable because after you get the response and set that response to a react state variable it re-renders the page and this process goes on and on making the page to rerender infinite times. If I was using this in a simple page(like in Subscribers.js file), and I dont need to save the response into a state variable this was fine but since I need to set the response to a state variable because this is a context provider, unfortunately, I can't use that function.

You don't need to set the response to the state variable. Instead you use a callback pattern as follows. Look at the example here -> https://github.com/magma/magma/blame/d691319dd40e0a7d2822a993f819beca7490e65d/nms/app/packages/magmalte/app/components/admin/FEGNetworkDialog.js#L49

andreilee: Is it completely unavoidable without simple changes? I'm referencing [this](https://daveceddia.com/useeffect-hook-examples/#prevent-useeffect-from-running-every-render)

nms/app/packages/magmalte/app/views/network/FEGNetworkInfo.js
uri200: Hey @karthiksubraveti, I don't think the user should configure gateways based on the hints. The hints are useful to find the field and show the format they need to input but they should not used as a definition. The users will have to check the guide first to see if those fields are even necessary. 

For example `served_nh_ids` is used on Inbound Roaming. It indicates other fegs this feg will be forwarding request. Once we have final field on NMS I will add a note to the instructions to use that (right now this is done by the API)

`served_network_ids`  is the ide of the federated LTE networks this FEG serves


Note both fields are an array, so it would be good to display the format the user needs to use to add more than one

nms/app/packages/magmalte/app/views/network/FEGNetworkDashboard.js
ato-dani: Yes, those grid containers are needed for styling purpose. They will make more sense on the next PR I made

nms/app/packages/magmalte/app/components/feg/FEGContext.js
ato-dani: Oh, I have changed it now using the useCallback function.

nms/app/packages/magmalte/app/views/network/FEGNetworkInfo.js
karthiksubraveti: > Hey @karthiksubraveti, I don't think the user should configure gateways based on the hints. The hints are useful to find the field and show the format they need to input but they should not used as a definition. The users will have to check the guide first to see if those fields are even necessary.
> 
> For example `served_nh_ids` is used on Inbound Roaming. It indicates other fegs this feg will be forwarding request. Once we have final field on NMS I will add a note to the instructions to use that (right now this is done by the API)
> 
> `served_network_ids` is the ide of the federated LTE networks this FEG serves
> 
> Note both fields are an array, so it would be good to display the format the user needs to use to add more than one

Hey @uri200 , Yes that's right. I understand that user shouldn't configure gateways based on hints. This component is not relevant to the user who is configuring alone, it is also relevant to other read only users. Hints will help them understand the page. We can link the guide if necessary in the config component. cc @jingtien

karthiksubraveti (approved):
Changes mostly LGTM. Please make any further changes based on @uri200 's responses here.

nms/app/packages/magmalte/app/views/network/FEGNetworkInfo.js
uri200: Hey @karthiksubraveti I really don't understand what is the problem: when configuring a feg the person doing so is expected to know that a feg serves a set of lte networks.  That is the field where they can configure those lte networks

In fact, if you want to make it easier, just create a combo box listing all the existing LTE federated networks on that orc8r and let the user add as many as they want. Then there will be no confusion

uri200 (approved):

andreilee (approved):

PR #7642, opened 2021-06-17
[NMS] Full Network page: Including Servicing Access Gateway Table

ato-dani:
## Summary
This is a draft PR because it is a stacked pr upon an exisiting open PR #7641. Once, that PR is approved, this PR would get rebased and would be changed from draft to a normal PR. Servicing access gateway table was added to the network page. Since servicing access gateways had a potential to be a large number, the gateway names and the network they are under were fetched and displayed in a table.

## Test Plan
A test file was created which verifies that all of the serviced access gateways are found under the table.  

## Additional Information
Screen Shot 2021-06-24 at 3 22 23 PM

### DIFF
--- a/nms/app/packages/magmalte/app/components/GatewayUtils.js
+++ b/nms/app/packages/magmalte/app/components/GatewayUtils.js
@@ -177,6 +177,10 @@ type GatewayStatusPayload = {
 
 const GATEWAY_KEEPALIVE_TIMEOUT_MS = 1000 * 5 * 60;
 
+export const HEALTHY_GATEWAY = 'Good';
+
+export const UNHEALTHY_GATEWAY = 'Bad';
+
 export default function isGatewayHealthy({status}: lte_gateway) {
   if (status != null) {
     const checkin = status.checkin_time;
--- a/nms/app/packages/magmalte/app/views/network/FEGServicingAccessGatewayTable.js
+++ b/nms/app/packages/magmalte/app/views/network/FEGServicingAccessGatewayTable.js
@@ -1,0 +1,169 @@
+/*
+ * Copyright 2020 The Magma Authors.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @flow strict-local
+ * @format
+ */
+
+import type {EnqueueSnackbarOptions} from 'notistack';
+import type {
+  feg_lte_network,
+  gateway_id,
+  gateway_name,
+  lte_gateway,
+  network_id,
+  network_name,
+} from '@fbcnms/magma-api';
+
+import ActionTable from '../../components/ActionTable';
+import Link from '@material-ui/core/Link';
+import LoadingFiller from '@fbcnms/ui/components/LoadingFiller';
+import React, {useEffect, useState} from 'react';
+import isGatewayHealthy from '../../components/GatewayUtils';
+import nullthrows from '@fbcnms/util/nullthrows';
+
+import {FetchGateways} from '../../state/lte/EquipmentState';
+import {
+  HEALTHY_GATEWAY,
+  UNHEALTHY_GATEWAY,
+} from '../../components/GatewayUtils';
+import {getServicedAccessNetworks} from '../../components/FEGServicingAccessGatewayKPIs';
+import {useEnqueueSnackbar} from '@fbcnms/ui/hooks/useSnackbar';
+import {useRouter} from '@fbcnms/ui/hooks';
+
+type ServicingAccessGatewayRowType = {
+  networkId: network_id,
+  networkName: network_name,
+  gatewayId: gateway_id,
+  gatewayName: gateway_name,
+  gatewayHealth: string,
+};
+
+/**
+ * Returns an array which holds information about each serviced
+ * access gateways servicied by this federation network.
+ * The information about the serviced gateways includes the gateway
+ * id, name, network name & network id under which the gateway exists.
+ *
+ * @param {Array<feg_lte_network>} servicedAccessNetworks List of federated LTE networks serviced by this federation network.
+ * @param {(msg, cfg) => ?(string | number),} enqueueSnackbar A snackbar to display errors.
+ * @returns An Array of the serviced access gateways with information about each.
+ */
+async function getServicedAccessGatewaysInfo(
+  servicedAccessNetworks: Array<feg_lte_network>,
+  enqueueSnackbar: (
+    msg: string,
+    cfg: EnqueueSnackbarOptions,
+  ) => ?(string | number),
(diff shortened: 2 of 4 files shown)


nms/app/packages/magmalte/app/views/network/FEGServicingAccessGatewayTable.js
```diff
@@ -0,0 +1,135 @@
...
+    ): Promise<Array<ServicingAccessGatewayRowType>> => {
+      const newServicedAccessGatewaysInfo = [];
+      for (const servicedAccessNetwork of servicedAccessNetworks) {
+        const servicedAccessGateways = await FetchGateways({
+          networkId: servicedAccessNetwork.id,
+          undefined,
+          enqueueSnackbar,
+        });
+        //Add the gateways of the serviced network
+        for (const servicedAccessGatewayId in servicedAccessGateways) {
```
karthiksubraveti: using servicedAccessGateways.map might be a more common pattern here.

nms/app/packages/magmalte/app/views/network/FEGServicingAccessGatewayTable.js
```diff
@@ -0,0 +1,135 @@
...
+          {title: 'Access Network', field: 'networkName'},
+          {
+            title: 'Access Gateway',
+            field: 'gatewayName',
+            render: currRow => (
+              <Link
+                variant="body2"
+                component="button"
+                onClick={() => {
+                  window.open(
```
karthiksubraveti, line 152: We have generally used history push pattern instead of window open pattern. For e.g.
https://github.com/magma/magma/blob/d691319dd40e0a7d2822a993f819beca7490e65d/nms/app/packages/magmalte/app/views/equipment/GatewayDetailEnodebs.js#L95

nms/app/packages/magmalte/app/views/network/FEGServicingAccessGatewayTable.js
```diff
@@ -0,0 +1,135 @@
...
+          undefined,
+          enqueueSnackbar,
+        });
+        //Add the gateways of the serviced network
+        for (const servicedAccessGatewayId in servicedAccessGateways) {
+          const newServicedAccessGatewayInfo: ServicingAccessGatewayRowType = {
+            networkId: servicedAccessNetwork.id,
+            networkName: servicedAccessNetwork.name,
+            gatewayId: servicedAccessGatewayId,
+            gatewayName: servicedAccessGateways[servicedAccessGatewayId]?.name,
```
karthiksubraveti: Can you add access gateway health as well in this table ? @uri200 is there any other fields which will be interesting to look at from federation perspective ?

nms/app/packages/magmalte/app/views/network/FEGServicingAccessGatewayTable.js
```diff
@@ -0,0 +1,135 @@
...
+  }
+  return (
+    <div>
+      <ActionTable
+        data={servicedAccessGatewaysInfo}
+        columns={[
+          {title: 'Access Network', field: 'networkName'},
+          {
+            title: 'Access Gateway',
+            field: 'gatewayName',
```
karthiksubraveti, line 146: Let's add gatewayID as well to be displayed here.

uri200:
On the columns at the right I see AGW (gateway itself) is accessible through a link. Would it be possible to make the same with AGW network?

nms/app/packages/magmalte/app/views/network/FEGServicingAccessGatewayTable.js
ato-dani: Yeah, I tried that but history doesn't work for this case because we are going from one network to another network, the context providers need to run again since they have different data. If we are **moving** from one page to another page **in the same network**, it would work since the already loaded data would be used and don't need to be changed. So, I discussed this with Andrei, last time and he suggested to open a new tab for now because that way we would force to load the data of the new network that we are traveling into.

ato-dani:
> On the columns at the right I see AGW (gateway itself) is accessible through a link. Would it be possible to make the same with AGW network?

Yeah, they are accessible through a link. It is possible to edit the federated lte network itself so that the access gateways there are linked to the federation network they are served under. I can go and edit the federated lte network to do that case if I have time after finishing my projects on the federation network.

nms/app/packages/magmalte/app/views/network/FEGServicingAccessGatewayTable.js
ato-dani: Did that & also updated test file to test that.

nms/app/packages/magmalte/app/views/network/FEGServicingAccessGatewayTable.js
```diff
@@ -0,0 +1,148 @@
...
+type ServicingAccessGatewayRowType = {
+  networkId: network_id,
+  networkName: network_name,
+  gatewayId: gateway_id,
+  gatewayName: gateway_name,
+  gatewayHealth: string,
+};
+
+/**
+ * Returns a table consisiting of the serviced access gateways alongside
```
andreilee: typo

nms/app/packages/magmalte/app/views/network/FEGServicingAccessGatewayTable.js
```diff
@@ -0,0 +1,148 @@
...
+            networkId: servicedAccessNetwork.id,
+            networkName: servicedAccessNetwork.name,
+            gatewayId: servicedAccessGatewayId,
+            gatewayName:
+              servicedAccessGateways[servicedAccessGatewayId]?.name || '',
+            gatewayHealth: isGatewayHealthy(
+              servicedAccessGateways[servicedAccessGatewayId] || {},
+            )
+              ? 'Good'
+              : 'Bad',
```
andreilee: Move these into constants. The notion of gateway health is used in multiple places in the NMS, so these strings should be shared.

nms/app/packages/magmalte/app/views/network/FEGServicingAccessGatewayTable.js
```diff
@@ -0,0 +1,148 @@
...
+  const networkId: string = nullthrows(match.params.networkId);
+  const [servicedAccessGatewaysInfo, setServicedAccessGatewaysInfo] = useState<
+    Array<ServicingAccessGatewayRowType>,
+  >([]);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    const getServicedAccessGatewaysInfo = async (
+      servicedAccessNetworks: Array<feg_lte_network>,
+    ): Promise<Array<ServicingAccessGatewayRowType>> => {
```
andreilee: Move this function definition out of the component

andreilee (changes requested):

nms/app/packages/magmalte/app/components/GatewayUtils.js
```diff
@@ -177,6 +177,10 @@ type GatewayStatusPayload = {
 
 const GATEWAY_KEEPALIVE_TIMEOUT_MS = 1000 * 5 * 60;
 
+export const GOOD_GATEWAY = 'Good';
+
+export const BAD_GATEWAY = 'Bad';
```
andreilee: Lol rename these constants to indicate it's for labeling a healthy or unhealthy gateway

ato-dani: Haha, my bad. I just did that.

andreilee (approved):

PR #7663, opened 2021-06-18
perf(datapath): Increase NIC queue size of S1 interface.

pshelar:
This queus size shows better stability on UE sessions after the
attachment.

## Test Plan
validated on baremetal AGW

## Additional Information

### DIFF
--- a/lte/gateway/configs/pipelined.yml_prod
+++ b/lte/gateway/configs/pipelined.yml_prod
@@ -171,6 +171,6 @@ ovs_internal_conntrack_fwd_tbl_number: 202
 dp_irq:
  enable: True
  S1_cpu: '2-3'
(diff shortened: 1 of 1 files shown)


koolzz (approved):

koolzz:
@pshelar should this be backported to 1.6?

ulaskozat (approved):
LGTM. Can we quantify the gains and mention cons of increasing the queue size if any?

pshelar:
> LGTM. Can we quantify the gains and mention cons of increasing the queue size if any?


I see much less idle UEs (20-30%) with this config in higher UE count, which results in stable throughput in test result.

downside is increased memory locked for NIC queue, but its not  alot (1024*1500) bytes.

Issue #7643, opened 2021-06-17
Make `subscriberdb` sync interval a network-wide orc8r-pushed config


waqaraqeel:
### Describe the Feature

The sync interval for the `subscriberdb` lte service is currently pulled from [lte/gateway/configs/subscriberdb.yml](https://github.com/magma/magma/blob/32bbd1edd739d62a0ec297ea4159b9066fd485b3/lte/gateway/configs/subscriberdb.yml#L22-L23). We want to make this configuration parameter configurable network-wide and pushed through the orc8r. It should still be overridable per AGW.

**Describe the solution you'd like**

These steps will be required:
- [x] Change Swagger model (`lte_network`) and create API endpoints (`/lte/{network_id}`)
- [x] Integrate with `lte_network` tests
- [x] Send config model as part of service config for `subscriberdb`
- [x] Use orc8r config and override with AGW specific config to initialize `subscriberdb`
- [ ] ~~Add docs with instructions on how to configure the sync interval and override for AGW~~

PR #7664, opened 2021-06-18
feat: make `subscriberdb_sync_interval` a network-wide orc8r-pushed config item

waqaraqeel:
## Summary

Working towards #7643. This PR makes Swagger model changes for both nw-wide and gw-specific configs, proto message changes, and implements the override mechanism.

The config is not actually used at the AGW yet. That will be a separate PR.

It may need a migration for existing networks and gateways.

## Test Plan
Changed existing tests to get and update `subscriberdb_sync_interval` for networks and gateways. Added new tests for override mechanism. Ran `./build.py --test` and manually verified through Swagger UI.


karthiksubraveti (commented):
I briefly looked at the changes here and feel the NetworkEPCConfigs and GatewayEPCConfigs is becoming like a kitchen sink for arbitrary gateway configs. Might be cleaner to add some structure to these configs.

waqaraqeel:
> I briefly looked at the changes here and feel the NetworkEPCConfigs and GatewayEPCConfigs is becoming like a kitchen sink for arbitrary gateway configs. Might be cleaner to add some structure to these configs.

Hey @karthiksubraveti! I agree on that point. It made sense to put this particular config (`subscriberdb_sync_interval`) beside the existing `cloud_subscriberdb_enabled`. Should we file an issue for `{nw,gw}EpcConfig` reorganization?

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -1749,6 +1749,12 @@ definitions:
       gx_gy_relay_enabled:
         type: boolean
         example: false
+      subscriberdb_sync_interval:
```
themarwhal, line 1752: If a description is allowed, I would mention that the unit is seconds here. (Is it?)
Or just postpend the field with `_seconds`/`_secs`, even.

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -602,3 +603,14 @@ func getNetworkProbeConfig(networkID string) ([]*lte_mconfig.NProbeTask, *lte_mc
 	}
 	return npTasks, liUes
 }
+
+// getSyncInterval takes network-wide subscriber_db sync interval and overrides it if also set for gateway
+func getSyncInterval(nwEpc *lte_models.NetworkEpcConfigs, gwEpc *lte_models.GatewayEpcConfigs) uint32 {
+	sync_interval := nwEpc.SubscriberdbSyncInterval
+	fmt.Printf("nw: %d\n", nwEpc.SubscriberdbSyncInterval)
```
themarwhal: If these are just for testing, let's remove them. Otherwise, we use `glog` for logging.

waqaraqeel: oh my.. Thanks for catching these. Unintentional.

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
waqaraqeel: Good point! Added description.

lte/protos/mconfig/mconfigs.proto
```diff
@@ -372,6 +372,8 @@ message SubscriberDB {
 
     // Enable relaying S6a messages via FeG RPC
     bool hss_relay_enabled = 6;
+
```
ardzoht, line 400: Add a description of the field, including what it represents (secs?)

waqaraqeel: Added. Thanks!

lte/cloud/go/services/lte/obsidian/models/defaults.go
```diff
@@ -38,6 +38,7 @@ func NewDefaultTDDNetworkConfig() *NetworkCellularConfigs {
 			GxGyRelayEnabled:         swag.Bool(false),
 			CloudSubscriberdbEnabled: false,
 			DefaultRuleID:            "",
+			SubscriberdbSyncInterval: 180,
```
hcgatewood: Here and below: default is 300 (5m)

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -1749,6 +1749,13 @@ definitions:
...
+        description: Configuration for subscriberdb service. Interval in seconds between gateway and cloud sync.
```
hcgatewood:
```suggestion
        description: Interval in seconds between gateway requests to sync its subscriberdb with the cloud.
```

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -1749,6 +1749,13 @@ definitions:
...
+        type: integer
+        format: uint32
+        minimum: 0
+        x-nullable: false
+        example: 180
```
hcgatewood: Again from above: default is 5m, so use that for the example

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -1749,6 +1749,13 @@ definitions:
       gx_gy_relay_enabled:
         type: boolean
         example: false
+      subscriberdb_sync_interval:
+        description: Configuration for subscriberdb service. Interval in seconds between gateway and cloud sync.
+        type: integer
+        format: uint32
+        minimum: 0
+        x-nullable: false
```
hcgatewood: If it's not required (which it's not), it needs to be nullable

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -1749,6 +1749,13 @@ definitions:
       gx_gy_relay_enabled:
         type: boolean
         example: false
+      subscriberdb_sync_interval:
+        description: Configuration for subscriberdb service. Interval in seconds between gateway and cloud sync.
+        type: integer
+        format: uint32
+        minimum: 0
```
hcgatewood: These two seem redundant (?)

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -2018,6 +2025,12 @@ definitions:
         type: boolean
         example: true
         default: true
+      subscriberdb_sync_interval:
```
hcgatewood, line 2028: Why two `subscriberdb_sync_interval` ? Create a separate object definition and then reference it in each location

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -602,3 +603,11 @@ func getNetworkProbeConfig(networkID string) ([]*lte_mconfig.NProbeTask, *lte_mc
 	}
 	return npTasks, liUes
 }
+
+// getSyncInterval takes network-wide subscriber_db sync interval and overrides it if also set for gateway
+func getSyncInterval(nwEpc *lte_models.NetworkEpcConfigs, gwEpc *lte_models.GatewayEpcConfigs) uint32 {
+	if gwEpc.SubscriberdbSyncInterval != 0 {
+		return gwEpc.SubscriberdbSyncInterval
+	}
+	return nwEpc.SubscriberdbSyncInterval
```
hcgatewood: What if both aren't set? We need some true default. Pull the true default from a cloud service config, and set that default value to 5m

lte/cloud/go/services/lte/servicers/builder_servicer_test.go
```diff
@@ -655,6 +658,65 @@ func TestBuilder_Build_BaseCase(t *testing.T) {
 	assert.Equal(t, expected, actual)
 }
 
+func TestBuilder_Build_ConfigOverride(t *testing.T) {
+	lte_test_init.StartTestService(t)
+
+	nwConfig := lte_models.NewDefaultTDDNetworkConfig()
+	// change sync interval from the default 180
```
hcgatewood: Nit: sentence-case your comments

hcgatewood (changes requested):

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -2147,6 +2151,13 @@ definitions:
           - 'HEX2BIN'
         example: 'BASE64'
 
+  subscriberdb_sync_interval:
+    description: Interval in seconds between gateway requests to sync its subscriberdb with the cloud.
+    type: integer
+    format: uint32
+    x-nullable: true
```
hcgatewood: This is the default I believe, so can omit it

lte/cloud/go/services/lte/servicers/builder_servicer.go
hcgatewood: Still need to handle this need for a "true default" sync interval -- there should be some default, and operators shouldn't have to set it for it to take effect

hcgatewood (changes requested):

lte/cloud/go/services/lte/servicers/builder_servicer.go
waqaraqeel: I've been thinking about this. A default can also be set on the gateway (in `gateway.mconfig`) and orc8r may not send anything. If we want to definitely send something from the orc8r, I haven't figured out where I should set that default value. I could put it right here in this function, but wonder if there's a more suitable place.

waqaraqeel: Could also have it in `lte/cloud/go/services/lte/obsidian/models/defaults.go`?

waqaraqeel: Placed default config in `lte/cloud/configs/subscriberdb.yml`

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
waqaraqeel: I looked into this and it seems `nullable` is an interesting thing in go-swagger: https://goswagger.io/use/models/schemas.html

It appears that `x-nullable: false` is the default. `x-nullable: true` turns basic data types into pointers with `nil` values when the field is unset, whereas by default, the type would hold empty values such as `0` and `""`. Keeping `x-nullable: false` by omitting it means that the value would be `0` when not set. It won't be required. I think we are okay with that in this case and can treat `0` as unset.

lte/cloud/configs/subscriberdb.yml
```diff
@@ -12,3 +12,7 @@
 
 # flatDigestEnabled is a feature flag for the flat digest functionality.
 flatDigestEnabled: true
+
+# syncInterval is the interval in seconds between gateway requests to sync its
+# subscriberdb with the cloud.
+syncInterval: 300
```
hcgatewood: This is actually the "default" sync interval, so make that clear in the var name and docstring, as well as where the value gets pulled in (service config)

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -627,3 +629,16 @@ func getNetworkSentryConfig(network *configurator.Network) *lte_mconfig.SentryCo
 		UrlPython:    string(sentryConfig.URLPython),
 	}
 }
+
+// getSyncInterval takes network-wide subscriber_db sync interval and overrides it if also set for gateway.
```
hcgatewood:
```suggestion
// getSyncInterval takes network-wide subscriberdb sync interval and overrides it if also set for gateway.
```

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -627,3 +629,16 @@ func getNetworkSentryConfig(network *configurator.Network) *lte_mconfig.SentryCo
...
+// If sync interval is unset for both network and gateway, a default is read from lte/cloud/configs/subscriberdb.yml
+func getSyncInterval(nwEpc *lte_models.NetworkEpcConfigs, gwEpc *lte_models.GatewayEpcConfigs) uint32 {
+	syncInterval := gwEpc.SubscriberdbSyncInterval
+	if syncInterval == 0 {
+		syncInterval = nwEpc.SubscriberdbSyncInterval
+	}
+	if syncInterval == 0 {
+		return subscriberdb.MustGetServiceConfig().SyncInterval
```
hcgatewood: Only ever read a `Must*` function during service startup. Instead, pass the read config in to this servicer, from the relevant main fn

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -627,3 +629,16 @@ func getNetworkSentryConfig(network *configurator.Network) *lte_mconfig.SentryCo
...
+	}
+	return uint32(syncInterval)
```
hcgatewood: Refrain from reassigning a variable where not necessary. So either use early returns, or a switch statement, or just use what you have but use new var names that are descriptive of their contents

lte/cloud/go/services/lte/servicers/builder_servicer_test.go
```diff
@@ -676,6 +679,75 @@ func TestBuilder_Build_BaseCase(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Equal(t, expected["subscriberdb"], actual["subscriberdb"])
+
+	nwConfig.Epc.SubscriberdbSyncInterval = 0
+	gatewayConfig.Epc.SubscriberdbSyncInterval = 0
+
+	// nw-wide and gw-specific not set. Service-level default expected
+	expected["subscriberdb"].(*lte_mconfig.SubscriberDB).SyncInterval = 300
+
+	actual, err = build_non_federated(&nw, &graph, "gw1")
```
hcgatewood: If you want to make my day, open a followup PR that fixes the usages of snake_case in build_non_federated and its surroundings

lte/cloud/go/services/subscriberdb/config.go
```diff
@@ -22,7 +22,8 @@ import (
 
 type Config struct {
 	// FlatDigestEnabled is a feature flag for the flat digest functionality.
-	FlatDigestEnabled bool `yaml:"flatDigestEnabled"`
+	FlatDigestEnabled bool   `yaml:"flatDigestEnabled"`
+	SyncInterval      uint32 `yaml:"syncInterval"`
```
hcgatewood: Yeah add a docstring for this, so that readers don't have to refer to the .yml file

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
```diff
@@ -11481,6 +11485,13 @@ definitions:
       icmp:
         $ref: '#/definitions/icmp_status'
     type: object
+  subscriberdb_sync_interval:
+    description: Interval in seconds between gateway requests to sync its subscriberdb
+      with the cloud.
+    example: 300
+    format: uint32
+    minimum: 60
```
hcgatewood, line 11511: So I think it is good to include this validation here. However, I think we also need to include the validation when creating the proto -- because operators could set the "default" value in lte.yml to be less than 1m, which would cause weird shit to happen

So keep this northbound validation, but also add something in your "get sync interval" method that returns `max{1m, calculatedSyncInterval}`

hcgatewood (changes requested):

lte/cloud/go/services/lte/servicers/builder_servicer_test.go
waqaraqeel: https://github.com/magma/magma/pull/7911

lte/cloud/go/services/lte/servicers/builder_servicer.go
waqaraqeel: I didn't find a way to directly pass config from `subscriberdb/main.go` to `builder_servicers`. So I instead exposed a method in `subscriberdb/client_api.go`. I'm not sure if this is the right way.

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
waqaraqeel: I added the extra check in `getSyncInterval()`.

lte/cloud/go/services/lte/servicers/builder_servicer.go
waqaraqeel: Moved the `defaultSyncInterval` configuration to `lte.yml`

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -627,3 +633,24 @@ func getNetworkSentryConfig(network *configurator.Network) *lte_mconfig.SentryCo
 		UrlPython:    string(sentryConfig.URLPython),
 	}
 }
+
+// getSyncInterval takes network-wide subscriberdb sync interval and overrides it if also set for gateway.
+// If sync interval is unset for both network and gateway, a default is read from lte/cloud/configs/subscriberdb.yml
+func getSyncInterval(nwEpc *lte_models.NetworkEpcConfigs, gwEpc *lte_models.GatewayEpcConfigs, defaultSyncInterval uint32) uint32 {
+	// minSyncInterval enforces a minimum sync interval to prevent too many
+	// sync requests if operators set the default lte.yml lower than 60
```
hcgatewood:
```suggestion
	// sync requests if operators sets the default in lte.yml to lower than 60
```

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -627,3 +633,24 @@ func getNetworkSentryConfig(network *configurator.Network) *lte_mconfig.SentryCo
 		UrlPython:    string(sentryConfig.URLPython),
 	}
 }
+
+// getSyncInterval takes network-wide subscriberdb sync interval and overrides it if also set for gateway.
+// If sync interval is unset for both network and gateway, a default is read from lte/cloud/configs/subscriberdb.yml
+func getSyncInterval(nwEpc *lte_models.NetworkEpcConfigs, gwEpc *lte_models.GatewayEpcConfigs, defaultSyncInterval uint32) uint32 {
```
hcgatewood: Instead of passing in defaultSyncInterval, just make this a method on the obj and use the defaultSyncInterval attribute

lte/cloud/go/services/lte/servicers/builder_servicer_test.go
```diff
@@ -676,6 +679,75 @@ func TestBuilder_Build_BaseCase(t *testing.T) {
...
+	expected["subscriberdb"].(*lte_mconfig.SubscriberDB).SyncInterval = 90
+
+	actual, err = buildNonFederated(&nw, &graph, "gw1")
+	assert.NoError(t, err)
+	assert.Equal(t, expected["subscriberdb"], actual["subscriberdb"])
+
+	nwConfig.Epc.SubscriberdbSyncInterval = 0
+	gatewayConfig.Epc.SubscriberdbSyncInterval = 0
+
+	// nw-wide and gw-specific not set. Service-level default expected
```
hcgatewood, line 743: Also add a test for when neither are set and the service-level is less than 60

orc8r/lib/go/service/config/service_config.go
```diff
@@ -86,6 +86,15 @@ func GetStructuredServiceConfig(moduleName string, serviceName string, out inter
 	return GetStructuredServiceConfigExt(moduleName, serviceName, main, legacy, overwrite, out)
 }
 
+// MustGetStructuredServiceConfig is the same as GetStructuredServiceConfig,
+// but it fails on errors.
+func MustGetStructuredServiceConfig(moduleName string, serviceName string, out interface{}) {
+	_, _, err := GetStructuredServiceConfig(moduleName, serviceName, &out)
+	if err != nil {
+		glog.Fatal(err)
```
hcgatewood: Wrap the error before you fatal, for more info

hcgatewood (approved):
Lgtm. Please fix nits then msg me to merge

lte/cloud/go/services/lte/servicers/builder_servicer_test.go
waqaraqeel: I skipped testing this since it will take significant plumbing. `lte_test_init.StartTestService()` will have to be modified to take in an `lte_service.Config` object since it creates the `BuilderServicer` object on the fly and does not expose it for later updates. The `StartTestService()` method is used in 22 places...

hcgatewood: You can either add a new `StartTestServiceWithConfig` method, or use e.g. IntelliJ's "refactor method" functionality to make the change in O(1) time

waqaraqeel: Should have thought of the `StartTestServiceWithConfig` way. Done now.

ardzoht (approved):
lgtm, looks like you need to fix some merge conflicts

waqaraqeel:
> lgtm, looks like you need to fix some merge conflicts

Resolved. Thanks!

PR #7682, opened 2021-06-21
make the test results related operations warn only

themarwhal:

## Summary
sometimes these operations fail because the VM is not found, https://app.circleci.com/pipelines/github/magma/magma/26608/workflows/7e9227ab-0871-4757-b00a-4ef8aca20d13/jobs/284942
I suspect they are getting shutdown somehow. 

Anyhow, if the integration tests pass, I don't want the job to fail for test summary related reasons, so putting a warn only flag.

## Additional Information


tmdzk (approved):
let's use that for now.

PR #7684, opened 2021-06-21
feat: Add support for Semantic Pull Requests

pshelar:
This documents commit message and subject line format.

## Additional Information

### DIFF
--- a/.github/workflows/pr_bot.yml
+++ b/.github/workflows/pr_bot.yml
@@ -27,6 +27,7 @@ jobs:
 
             Please take a moment to read through the Magma project's
             - [Contributing Conventions](https://docs.magmacore.org/docs/next/contributing/contribute_conventions) for norms around contributed code
+            - [Commit msg style](https://docs.magmacore.org/docs/next/contributing/contribute_commit_msg) recommended style for writing commit msg
 
             If this is your first Magma PR, also consider reading
             - [Developer Onboarding](https://docs.magmacore.org/docs/next/contributing/contribute_onboarding) for onboarding as a new Magma developer
--- a/docs/readmes/contributing/contribute_commit_msg.md
+++ b/docs/readmes/contributing/contribute_commit_msg.md
@@ -1,0 +1,96 @@
+---
+id: contribute_commit_msg
+title: Commit message style
+hide_title: true
+---
+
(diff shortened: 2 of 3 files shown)


docs/readmes/contributing/contribute_commit_msg.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Advantage of semantic commits.
+1. This state intention of the commit msg in concise manner in PR title.
+2. It can be used for automatically compiling change log for magma release
+   documentation.
+3. Analytics related to number of fixes vs feature commits in each release.
+4. This would force developer to rethink when they mix different type of
+   changes in same PR.
+
+## type
```
themarwhal: For type and scope it might be better to link to the config file so that the doc stays up to date.

themarwhal (approved):
We should probably add a link to this in the hello bot message. https://github.com/magma/magma/blob/master/.github/workflows/pr_bot.yml

docs/readmes/contributing/contribute_commit_msg.md
pshelar: I am not sure how to do it, can you point me to example?

themarwhal: I didn't mean anything fancy. I just meant adding a line like: `for an up-to-date list of types we support see <INSERT LINK TO https://github.com/magma/magma/blob/master/.github/semantic.yml>`

Just so that people know where to look. (I know you mention the file in the doc somewhere else, but people might not be careful readers :) )

.github/workflows/pr_bot.yml
```diff
@@ -27,6 +27,7 @@ jobs:
 
             Please take a moment to read through the Magma project's
             - [Contributing Conventions](https://docs.magmacore.org/docs/next/contributing/contribute_conventions) for norms around contributed code
+            - [Commit msg style](https://docs.magmacore.org/docs/next/contributing/contribute_commit_msg) recommended style for writting commit msg
```
themarwhal: legit spelling annotation
writting -> writing

themarwhal:
Oh actually I think this is a good spot to add docs on this: https://docs.magmacore.org/docs/next/contributing/contribute_workflow#guidelines

PR #7686, opened 2021-06-21
docs: Fix header enrichment doc.

pshelar:
Update URL requirement for the HE APIs

## Test Plan
Validated in lab.

## Additional Information

### DIFF
--- a/docs/readmes/howtos/header_enrichment.md
+++ b/docs/readmes/howtos/header_enrichment.md
@@ -91,7 +91,9 @@ Following example show required parameters for Header enrichment rule.
 2. Mandatory field: `Direction` needs to be `uplink`.
 3. Mandatory field: `tcp_dst` port needs to be `80 (http)`.
 4. Mandatory field: `ip_dst` is IP address range of the HTTP servers.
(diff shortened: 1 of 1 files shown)


docs/readmes/howtos/header_enrichment.md
```diff
@@ -91,7 +91,8 @@ Following example show required parameters for Header enrichment rule.
 2. Mandatory field: `Direction` needs to be `uplink`.
 3. Mandatory field: `tcp_dst` port needs to be `80 (http)`.
 4. Mandatory field: `ip_dst` is IP address range of the HTTP servers.
-5. Mandatory field: `header_enrichment_targets` which is list of URLs.
+5. Mandatory field: `header_enrichment_targets` which is list of URLs. Do not
+   specify http protocol in URL name, for example `http://abc.com` is not supported.
```
sudhikan: Does it make sense to add a valid example as well? For instance:

` ... for example http://abc.com is not supported, but abc.com is a valid target`.

pshelar: It is shown in the example just above these points. But lets add to be clear.

themarwhal (approved):

PR #7697, opened 2021-06-22
[component:agw] Mobilityd IP Release fix

panyogesh:
Issue:
  1. Once the session is released IP addresse
     needs to be freed in mobilityD

Fix:
  1. Enhanced the M5G mobility Service client to support
     IP address release message to MobolityD
  2. Done some changes in IP Address allocation part
     to store the information in SMF context.

Test:
  1. Tested on UERANSIM and verified logs of mobilityd

Log : Released IP 192.168.128.223 for sid IMSI901700000000001

## Additional Information

### DIFF
--- a/lte/gateway/c/oai/include/amf_app_messages_types.h
+++ b/lte/gateway/c/oai/include/amf_app_messages_types.h
@@ -126,7 +126,7 @@ typedef struct itti_amf_ip_allocation_response_s {
   uint8_t gnb_gtp_teid_ip_addr[16];
 
   /* UE IP Address Allocated by MobilityD */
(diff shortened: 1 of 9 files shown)


panyogesh:
[Uploading Test-Logs-7697.txt…]()

ulaskozat:
There are some intricacies about how the async grpc callbacks should be handled as the callback runs on a separate thread than the AMF task. If you check the 4G path callbacks shoot ITTI message back the the calling thread (in this case spgw task) which then decides on what to do with the context it owns. I am landing this as is but let's keep an eye on that.

PR #7709, opened 2021-06-22
[lte][agw] Add detach timer to the clean up

ulaskozat:

## Summary

- Detach timer was not being cleaned up when UE context was removed. This is a potential fix for #7566. 

## Test Plan

Integ tests.

## Additional Information


ardzoht (approved):
nice catch, lgtm

PR #7710, opened 2021-06-22
Migrate stats collection to the mme_app thread

### LINKED ISSUE #7421: magma::lte::MmeNasStateManager::get_state
Hitting assert fatal: AssertFatal(state_cache_p, "mme_nas_state is NULL");

Sentry Issue: [LAB-AGWS-NATIVE-N](https://sentry.io/organizations/lf-9c/issues/2444965165/?referrer=github_integration)

```
SIGILL / ILL_ILLOPN: Fatal Error: SIGILL / ILL_ILLOPN
  File "mme_app_state_manager.cpp", line 90, in magma::lte::MmeNasStateManager::get_state
  File "mme_app_state.cpp", line 41, in get_mme_nas_state
  File "service303_mme_stats.c", line 26, in service303_mme_statistics_read
  File "service303_mme_stats.c", line 34, in service303_statistics_read
  File "service303_task.c", line 76, in handle_timer
...
(4 additional frame(s) were not displayed)
```

amarpad:
## Summary
Currently service303 reaches across service boundaries to read
mme_app state. Fix this by pushing the stats over an ITTI message
Also remove the check for MME service as the timer is not started
in the SPGW context (as it is only run in the MME_APP)

Also may close #7421 where the nas datastructure is null as there is probably
races with clearing out the state_p.

Also, addresses race in initialization where the timer was started outside
mme_app_thread thus the zmq context might not be initialized

This reverts commit ab6d9cbe3dcd403d569df6b9142fe5c181e240ca.

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/lib/message_utils/service303_message_utils.h
+++ b/lte/gateway/c/core/oai/lib/message_utils/service303_message_utils.h
@@ -26,4 +26,8 @@
 int send_app_health_to_service303(
     task_zmq_ctx_t* task_zmq_ctx_p, task_id_t origin_id, bool healthy);
 
+int send_stats_to_service303(
+    task_zmq_ctx_t* task_zmq_ctx_p, task_id_t origin_id,
+    application_mme_stats_msg_t* stats_msg);
+
 #endif /* FILE_SERVICE303_MESSAGE_UTILS */
--- a/lte/gateway/c/core/oai/tasks/mme_app/mme_app_main.c
+++ b/lte/gateway/c/core/oai/tasks/mme_app/mme_app_main.c
@@ -55,6 +55,7 @@
 static void check_mme_healthy_and_notify_service(void);
 static bool is_mme_app_healthy(void);
 static void mme_app_exit(void);
+static void start_stats_timer(void);
 
 bool mme_hss_associated = false;
 bool mme_sctp_bounded   = false;
@@ -62,6 +63,7 @@ task_zmq_ctx_t mme_app_task_zmq_ctx;
 bool mme_congestion_control_enabled = true;
 long mme_app_last_msg_latency;
 long pre_mme_task_msg_latency;
+static long epc_stats_timer_id;
 
 mme_congestion_params_t mme_congestion_params;
 
@@ -520,6 +522,7 @@ static void* mme_app_thread(__attribute__((unused)) void* args) {
 
   // Service started, but not healthy yet
   send_app_health_to_service303(&mme_app_task_zmq_ctx, TASK_MME_APP, false);
+  start_stats_timer();
 
   zloop_start(mme_app_task_zmq_ctx.event_loop);
   mme_app_exit();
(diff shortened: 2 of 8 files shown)


lte/gateway/c/core/oai/tasks/mme_app/mme_app_main.c
```diff
@@ -520,6 +522,7 @@ static void* mme_app_thread(__attribute__((unused)) void* args) {
 
   // Service started, but not healthy yet
   send_app_health_to_service303(&mme_app_task_zmq_ctx, TASK_MME_APP, false);
+  start_stats_timer();
```
amarpad, line 525: I had to move the start to this worker thread instead of the main thread as otherwise there is a race with zmq starting.

lte/gateway/c/core/oai/lib/message_utils/service303_message_utils.h
```diff
@@ -26,4 +26,9 @@
 int send_app_health_to_service303(
     task_zmq_ctx_t* task_zmq_ctx_p, task_id_t origin_id, bool healthy);
 
+int send_stats_to_service303(
+    task_zmq_ctx_t* task_zmq_ctx_p, task_id_t origin_id,
+    uint32_t nb_enb_connected, uint32_t nb_ue_attached,
+    uint32_t nb_ue_connected);
```
ulaskozat: can you actually use the structure that you defined? It will be easier to extend when we want to pass more stats/states in the future.

lte/gateway/c/core/oai/tasks/mme_app/mme_app_main.c
ulaskozat: It belongs to mme_app thread anyhow, so this is indeed the right location.

lte/gateway/c/core/oai/lib/message_utils/service303_message_utils.h
amarpad: Good point, I have a follow-up change that converts this to a C++ object. Can I put a pin on this and get back within a week?
Technically if it is a C object we could reuse application_mme_stats_msg, so if you would like to see a struct here, I can replace this interface with that as well.

amarpad: The underlying reason for this change apart from being right is to convert the dataobjects in mme_app to C++ and that would require service303 not reaching out into the mme_app as it doesn't compile as I am not migrating service303 thread.

ulaskozat: let's reuse application_mme_stats_msg to pack the parameters now, as I only see the benefits of doing so.

amarpad: Updated with application_mme_stats_msg_t for now. As noted I am still figuring out the right way to replace the union, will float out a proposal to support a polymorphic ITTI msg object in C++. Thanks for the review

ulaskozat (approved):
LGTM

ulaskozat:
we want to backport to 1.6, right?

amarpad:
right we do, I am still trying to figure out why this breaks the OAI MME. waiting for help from @rdefosse

rdefosse:
@amarpad 

It is better than yesterday. It passes the DsTester testsuite when deployed on `Ubuntu18`.

But on the `RHEL8` sanity check deployment, it still fails. I manually built your branch into an new image and ran manually the test.

I ran as the test and then I ran even more manually with a gdb in the MME executable. I just launched the MME after the `stcpd` executable and after 2 or 3 minutes I got an assertion:

```bash
$ gdb ./bin/oai_mme
(gdb) run -c ./etc/mme.conf 

oai_mme: src/zlistx.c:182: zlistx_add_end: Assertion `item' failed.

Thread 10 "TASK_SERVICE303" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffe69f7700 (LWP 338)]
0x00007ffff09d937f in raise () from /lib64/libc.so.6

```

See all logs at `https://jenkins-oai.eurecom.fr/job/MAGMA-MME-RHEL8-Deploy-Check/268/`

Versus a passing run at `https://jenkins-oai.eurecom.fr/job/MAGMA-MME-RHEL8-Deploy-Check/272/`

Raphael

rdefosse:
@amarpad 

I just tried also manually on the `Ubuntu18` image based on your branch.

I see the same behavior. I started `HSS`, `sctpd`, `MME` and waited for a while....

```bash
root@mme:/magma-mme# ./bin/oai_mme -c ./etc/mme.conf 
Initializing OAI Logging
Initializing OAI Logging to syslog
Initializing shared logging
Initializing shared logging Done
Initializing TIMER module
Initializing TIMER module: DONE
mme_config_parse_opt_line mme_config.config_file ./etc/mme.conf
filename /var/run/oai_mme.pid failed to be written
[/magma/orc8r/gateway/c/common/config/ServiceConfigLoader.cpp:35] Override file not found for service control_proxy
[/magma/orc8r/gateway/c/common/config/ServiceConfigLoader.cpp:35] Override file not found for service service_registry
oai_mme: src/zlistx.c:172: zlistx_add_end: Assertion `item' failed.
Aborted (core dumped)
```

After 2/3 minutes, MME stopped. Our DsTester tests were fast enough to pass. Lucky one.

Raphael

amarpad:
As discussed async @rdefosse will validate it longer in the OAI jenkins and then post back, his time. Thanks for the help @rdefosse

rdefosse (approved):
@amarpad 

I approve from a CI point of view. 

I deployed an `RHEL8` image based on your last commit and MME ran for 20 minutes without any problem.

Thanks a lot for making the effort to pass the OAI pipeline.

Regards,
Raphael

Issue #7613, opened 2021-06-17
Provide orc8r-pushed, per AGW overridable configuration for sentry.io


waqaraqeel:
### Describe the Feature

[Sentry.io](https://sentry.io/welcome/) has been integrated for lte with `sessiond` and `mme` services. Sentry configuration is currently read from `lte/gateway/configs/control_proxy.yml`. We want these parameters to be configurable network-wide, and pushed through the orc8r. They should still be overridable per AGW.

**Describe the solution you'd like**

These steps will be required:
- [x] Create Swagger config model and API endpoints
- [x] Integrate with Network model and tests
- [x] Send config model as part of service config for `sessiond` and `mme`
- [x] Use orc8r config and override with AGW specific config to initialize `sentry`
- [x] Update [integrate_sentry](https://docs.magmacore.org/docs/next/lte/integrate_sentry) documentation to reflect the changes

Refer to more detailed instructions [here](https://fb.quip.com/ehjOAx1IVNC2).

PR #7718, opened 2021-06-23
feat: use sessiond mconfig to initialize sentry in sessiond

waqaraqeel:

## Summary
Working towards #7613.

This PR makes changes to sessiond sentry initialization so that relevant configuration sent from orc8r are used. Configurations in `control_proxy.yml` still take precedence over those sent from orc8r.

## Test Plan

Ran `make test_session_manager` in `lte/gateway` with `SENTRY_ENABLED` flag set.


themarwhal (commented):
Can we modify `lte/gateway/configs/control_proxy.yml` so that the default sample rate is 1?

lte/gateway/c/session_manager/SentryWrappers.cpp
```diff
@@ -37,33 +38,56 @@ using std::experimental::optional;
...
 // TODO(@themarwhal) pull common sentry functions into lib common
 
-optional<std::string> get_sentry_url(YAML::Node control_proxy_config) {
-  std::string sentry_url;
+optional<std::string> get_sentry_url(
+    const magma::mconfig::SentryConfig& sentry_config,
+    YAML::Node control_proxy_config) {
   if (control_proxy_config[SENTRY_NATIVE_URL].IsDefined()) {
-    const std::string sentry_dns =
+    const std::string override =
```
themarwhal: `dns_override` just to be extra clear

lte/gateway/c/session_manager/SentryWrappers.cpp
```diff
@@ -37,33 +38,56 @@ using std::experimental::optional;
...
+    return sentry_url;
+  }
   return {};
 }
 
+float get_sentry_sample_rate(
+    const magma::mconfig::SentryConfig& sentry_config,
+    YAML::Node control_proxy_config) {
+  if (control_proxy_config[SENTRY_SAMPLE_RATE].IsDefined()) {
+    const float override = control_proxy_config[SENTRY_SAMPLE_RATE].as<float>();
```
themarwhal: `sample_rate_override`

themarwhal (approved):
the changes look good thanks!

My only concern is that the sample rate from the config is 0.5 by default at the moment, we might want that value to be consistent with the default value we use here.

waqaraqeel:
> the changes look good thanks!
> 
> My only concern is that the sample rate from the config is 0.5 by default at the moment, we might want that value to be consistent with the default value we use here.

I have changed the default `sample_rate` in `control_proxy.yml` to 1.0. Also changed the local variable names to reflect your suggestions. Thanks!

PR #7719, opened 2021-06-23
refactor(agw): move sync interval randomization from agw to orc8r

waqaraqeel:

## Summary
Working towards #7643.

This PR moves the subscriberdb sync interval randomization (added in #7513) from the gateway to the orc8r. The intent is to keep the gateway dumb. The gateway will use the randomized subscriberdb sync interval as received from the orc8r (change made in #7664).

## Test Plan

Modified tests in `builder_servicer_tests` to accept randomized values from the `builder_servicer`. Ran those tests and others from `./build.py -t`. Ran `make test_python` from `lte/gateway` in the magma VM.

### DIFF
--- a/lte/cloud/go/services/lte/servicers/builder_servicer.go
+++ b/lte/cloud/go/services/lte/servicers/builder_servicer.go
@@ -16,6 +16,7 @@ package servicers
 import (
 	"context"
 	"fmt"
+	"hash/fnv"
 	"sort"
 
 	feg "magma/feg/cloud/go/feg"
@@ -196,7 +197,7 @@ func (s *builderServicer) Build(ctx context.Context, request *builder_protos.Bui
 			LteAuthAmf:      nwEpc.LteAuthAmf,
 			SubProfiles:     getSubProfiles(nwEpc),
 			HssRelayEnabled: swag.BoolValue(nwEpc.HssRelayEnabled),
-			SyncInterval:    s.getSyncInterval(nwEpc, gwEpc),
+			SyncInterval:    s.getRandomizedSyncInterval(cellGW.Key, nwEpc, gwEpc),
 		},
 		"policydb": &lte_mconfig.PolicyDB{
 			LogLevel: protos.LogLevel_INFO,
@@ -634,7 +635,7 @@ func getNetworkSentryConfig(network *configurator.Network) *lte_mconfig.SentryCo
 	}
 }
 
-// getSyncInterval takes network-wide subscriberdb sync interval and overrides it if also set for gateway.
+// getSyncInterval takes network-wide subscriberdb sync interval in seconds and overrides it if also set for gateway.
 // If sync interval is unset for both network and gateway, a default is read from lte/cloud/configs/lte.yml
 func (s *builderServicer) getSyncInterval(nwEpc *lte_models.NetworkEpcConfigs, gwEpc *lte_models.GatewayEpcConfigs) uint32 {
 	// minSyncInterval enforces a minimum sync interval to prevent too many
@@ -653,4 +654,21 @@ func (s *builderServicer) getSyncInterval(nwEpc *lte_models.NetworkEpcConfigs, g
 		return s.defaultSubscriberdbSyncInterval
 	}
 	return minSyncInterval
+}
+
+// getRandomizedSyncInterval returns the interval received from getSyncInterval as seconds and increases it by a random
+// delta in the range [0, getSyncInterval() / 5.0]. Increased sync interval ameliorates the thundering herd effect at
+// the orc8r. Delta is deterministic based on gwKey
+func (s *builderServicer) getRandomizedSyncInterval(gwKey string, nwEpc *lte_models.NetworkEpcConfigs, gwEpc *lte_models.GatewayEpcConfigs) uint32 {
+	syncInterval := s.getSyncInterval(nwEpc, gwEpc)
+
+	// FNV-1 is a non-cryptographic hash function that is fast and very simple to implement
+	h := fnv.New32a()
+	_, err := h.Write([]byte(gwKey))
+	if err != nil {
+		return syncInterval
+	}
+	multiplier := float32(h.Sum32()%100) / 100.0
+	delta := multiplier * (float32(syncInterval) / 5.0)
+	return syncInterval + uint32(delta)
 }
--- a/lte/cloud/go/services/lte/servicers/builder_servicer_test.go
+++ b/lte/cloud/go/services/lte/servicers/builder_servicer_test.go
@@ -39,6 +39,16 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const (
+	// randomizedInterval300 and similar constants are calculated by adding delta * (300/5) to 300 where delta is a
+	// random fraction between 0 and 1 based on the FNV-1 hash of the string "gw1". These constants emulate return
+	// values of builder_servicer.getRandomizedSyncInterval("gw1", 300) etc.
+	randomizedInterval300 = 344
+	randomizedInterval120 = 137
+	randomizedInterval90  = 103
+	randomizedInterval60  = 68
+)
+
 func TestBuilder_Build(t *testing.T) {
 	lte_test_init.StartTestService(t)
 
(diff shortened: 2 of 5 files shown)


lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -103,6 +104,8 @@ func (s *builderServicer) Build(ctx context.Context, request *builder_protos.Bui
 	gwRan := cellularGwConfig.Ran
 	gwEpc := cellularGwConfig.Epc
 	gwNonEpsService := cellularGwConfig.NonEpsService
+	// gwKey is a unique string for each gateway since hardware ID is not accessible here
+	gwKey := network.ID + cellGW.Type + cellGW.Key
```
waqaraqeel: @hcgatewood Is this okay to use as a unique key for a gateway since the hardware ID of the gateway is not available? Randomization of the sync interval will be based on this key.

lte/gateway/python/magma/subscriberdb/main.py
```diff
@@ -66,9 +64,7 @@ def main():
             service_stub=SubscriberDBCloudStub,
             max_client_reuse=60,
         )
-        sync_interval = _randomize_sync_interval(
-            service.config.get('subscriberdb_sync_interval'),
-        )
+        sync_interval = service.mconfig.sync_interval
```
waqaraqeel, line 67: @hcgatewood Is this completely safe? I think it is because a `sync_interval` has been added to the default `gateway.mconfig` and we have made sure that the orc8r sends one too.

lte/cloud/go/services/lte/servicers/builder_servicer.go
mstre123: couldn't we randomize the sync interval by just generating a random delay using unix time as the seed? my understanding is this was to prevent a storm of traffic, so an occasional collision probably isn't the end of the world

waqaraqeel: We want to make the randomization deterministic for each gateway. The original PR (#7513) does it this way as well probably to avoid config updates on gateways.

mstre123: got it, hadn't seen the first pr but looks like this was already hashed out (pun intended :))

waqaraqeel: Added #7513 in summary as context.

lte/gateway/python/magma/subscriberdb/main.py
hcgatewood: Seems safe to me. Will defer to @ardzoht. Also, you can probably test this yourself with local AGW setup if you want to, but may not be necessary

lte/cloud/go/services/lte/servicers/builder_servicer_test.go
```diff
@@ -1612,3 +1613,11 @@ func setEpcNetworkServiceAreaMap(nw *configurator.Network, serviceAreaMaps map[s
 	cellularNwConfig.Epc.ServiceAreaMaps = serviceAreaMaps
 	nw.Configs[lte.CellularNetworkConfigType] = cellularNwConfig
 }
+
+func getRandomizedInterval(nwID string, gwKey string, interval uint32) uint32 {
```
hcgatewood: Eh just make this public instead of copying, though idc either way

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -16,6 +16,7 @@ package servicers
 import (
 	"context"
 	"fmt"
+	"hash/fnv"
```
hcgatewood, line 19: Why fnv sum? Add note somewhere on choice

lte/cloud/go/services/lte/servicers/builder_servicer.go
hcgatewood: You could conceivably pull the hwid from configurator, but it's not worth the call. Also, we don't actually need this to be globally unique (or really that unique at all, just "unique enough") -- so can just use cellGW.Key as the key

hcgatewood (changes requested):
Shape looks great, couple small changes

lte/gateway/python/magma/subscriberdb/main.py
waqaraqeel: I tested it with an AGW connected to the orc8r. Will test with a headless one too.

lte/cloud/go/services/lte/servicers/builder_servicer_test.go
waqaraqeel: It is weird to have the function being tested directly provide the expected value. Problems with the function wouldn't be caught. I could hardcode the expected randomized values if we don't want to copy the function...

lte/cloud/go/services/lte/servicers/builder_servicer.go
waqaraqeel: Using `cellGW.Key`

lte/cloud/go/services/lte/servicers/builder_servicer_test.go
hcgatewood: Agreed. Let's go with hard-coding then, with brief comments indicating how you calculated the hardcoded values

lte/gateway/python/magma/subscriberdb/main.py
ardzoht: Adding this to the default gateway.mconfig and validating this field is always generated by orc8r should be enough,  you can always do `if service.mconfig.sync_interval` to validate existence but it should be good as it is.

ardzoht (approved):
AGW changes look good to me

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -654,3 +657,16 @@ func (s *builderServicer) getSyncInterval(nwEpc *lte_models.NetworkEpcConfigs, g
 	}
 	return minSyncInterval
 }
+
+// getRandomizedSyncInterval returns the interval received from getSyncInterval increased by a random
+// delta in the range [0, getSyncInterval() / 5.0]. Increased sync interval ameliorates the thundering
+// herd effect at the orc8r. Delta is deterministic based on gwKey
```
ardzoht: Missed this, but let's make sure to mention on the docs or on the variable itself that this is measured in secs

ardzoht (changes requested):

lte/cloud/go/services/lte/servicers/builder_servicer_test.go
waqaraqeel: Done.

lte/cloud/go/services/lte/servicers/builder_servicer.go
waqaraqeel: Included unit in doc string.

lte/cloud/go/services/lte/servicers/builder_servicer.go
waqaraqeel: Added note on choice. Basically "use the simplest tool that suffices."

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -654,3 +655,20 @@ func (s *builderServicer) getSyncInterval(nwEpc *lte_models.NetworkEpcConfigs, g
...
 	return minSyncInterval
 }
+
+// getRandomizedSyncInterval returns the interval received from getSyncInterval as seconds and increases it by a random
+// delta in the range [0, getSyncInterval() / 5.0]. Increased sync interval ameliorates the thundering herd effect at
+// the orc8r. Delta is deterministic based on gwKey
+func (s *builderServicer) getRandomizedSyncInterval(gwKey string, nwEpc *lte_models.NetworkEpcConfigs, gwEpc *lte_models.GatewayEpcConfigs) uint32 {
+	syncInterval := s.getSyncInterval(nwEpc, gwEpc)
+
+	// FNV-1 is a non-cryptographic hash function that is fast and very simple to implement
```
hcgatewood, line 665: non-cryptographic + fast is sufficient for me, idc if it's simple to implement

hcgatewood (approved):
Lgtm thanks @waqaraqeel!

ardzoht (approved):
lgtm

Issue #7036, opened 2021-05-20
Scale tests leave dirty mme_nas_state


ulaskozat:
In scaling tests we end up in duplicate attach requests and ignoring legitimate users. Inspecting redis state shows dirty guti map although there are no sessions in the system:

```
magma@1vagw2:~$ state_cli.py parse "mme_nas_state" | grep guti_ue_id_htbl | wc -l
1429
```

rsarwad:
Hi @ulaskozat  can we get mme.log and wireshark capture

ulaskozat:
@rsarwad I do not have the pcaps but the mme log and the state info.

[log.zip](https://github.com/magma/magma/files/6524434/log.zip)

rsarwad:
Hi @ulaskozat  I am not able to view the mme.log attached within log.zip.  It's appearing as below 
![image](https://user-images.githubusercontent.com/30685618/119281247-9e0e2780-bc52-11eb-8d0f-9a5fe9f14857.png)

ulaskozat:
@rsarwad most likely I copied mme binary rather than the mme log message. I will rerun the test and capture pcaps as well.

ulaskozat:
@rsarwad attached the new logs and pcaps. I also included leftover states. There are some on MME and some on SPGW. There are a good number of session rejects due to IP address allocation failures I think.

[log_300UE_10UEpsec.zip](https://github.com/magma/magma/files/6529808/log_300UE_10UEpsec.zip)

[sim_300UE_rate10.pcapng.zip](https://github.com/magma/magma/files/6529812/sim_300UE_rate10.pcapng.zip)

ardzoht:
To follow up on this, I ran a test with 500 UEs / 5 UE per sec attach for 30 iterations (which took around 10 hours), I monitored over time the changes on state, SPGW and MME UE state maintained the same number of entries as the `get_subscriber_table` command. Also the `nb_ue_attached` on `mme_nas_state` followed the same pattern.

These are the results at the end of the runs:

```
magma@phy-u6:~$ state_cli.py keys "*:SPGW" | wc -l
0
magma@phy-u6:~$ state_cli.py keys "*:MME" | wc -l
0
magma@phy-u6:~$ state_cli.py keys "*:S1AP" | wc -l
0
magma@phy-u6:~$ state_cli.py parse s1ap_state | grep mmeid2associd | wc -l
0
[...]

PR #7721, opened 2021-06-23
Rsarwad fix guti stale entries

rsarwad:
Addresses the issue, #7036
GUTI stale entries were seen in various scenarios as listed below:

When mme sends Service reject, clear_emm_context() was getting called due to which it was resulting guti stale entries

When MME sends TAU reject; clear_emm_context() was getting called due to which it was resulting guti stale entries

While handling duplicate attach requests; MME implicitly detaches the first UE and if second attach request is received in UL nas message, as part of handling new attach request mme re-initializes emm context, added fix such that mme shall remove the previous GUTI entry before mme re-initialize emm context.
This scenario was observed on issue, #7036. Attaching the pcap, mme_ue_s1ap_id_1679.zip

While handling duplicate attach requests; Consider the case that as part of first attach request, UE is registered and later the UE is moved to idle state. Then mme receives second attach request in initial UE message. So mme implicitly detach the first UE contexts. While handling implicit detach of first UE, mme_app was sending "ue context release command" but since UE is already in idle mode, s1ap was not able to get ue_reference_p for first UE's mme_ue_s1ap_id. So message was getting dropped, neither s1ap was sending ue_context_release_complete to mme_app;
So contexts were not getting deleted at mme_app;

When new context is created, mme_insert_ue_context() is called, in this fun, it checks for change in the guti. When UE is created m_tmsi is initialized with "-1", check fails and creates guti entry with value, "000000000000ffffffff". Fixed all these issues.

## Test Plan
## Additional Information
Once OAI team support lbi field in Delete Session Response from spgw; mme shall delete the pdn context corresponding to lbi.
'nb_active_pdn_contexts' needs to be decremented as part of deleting the pdn context.
Also merge the functions, mme_app_free_pdn_context  and pdn_connectivity_delete


lte/gateway/c/core/oai/tasks/nas/emm/Attach.c
```diff
@@ -2594,82 +2594,114 @@ static int emm_attach_update(
...
-          ue_context_p->sgs_context->ts9_timer.id ==
-          MME_APP_TIMER_INACTIVE_ID) {
-        /* Notify S1AP to send UE Context Release Command to eNB or free
-         * s1 context locally.
-         */
+   * So release previous Attach Request's contexts
+   */
+  if (attach_info.is_mm_ctx_new) {
+    if (ue_context_p->ecm_state == ECM_IDLE) {
+      mme_remove_ue_context(mme_ue_context_p, ue_context_p);
```
ardzoht, line 2622: I would suggest adding a debug (maybe info? level) log here to specify UE context is removed due to being on idle mode

lte/gateway/c/core/oai/tasks/nas/esm/PdnConnectivity.c
```diff
@@ -550,9 +550,6 @@ proc_tid_t pdn_connectivity_delete(
         "\n",
         pdn_cid, ue_mm_context->mme_ue_s1ap_id);
   }
-  if (ue_mm_context->nb_active_pdn_contexts > 0) {
-    ue_mm_context->nb_active_pdn_contexts -= 1;
-  }
```
ulaskozat, line 555: rather than removing should not we simply say:
```
if ((ue_context_p->nb_active_pdn_contexts > 0) &&
      (ue_context_p->emm_context.esm_ctx.is_pdn_disconnect)) {
    ue_context_p->nb_active_pdn_contexts -= 1;
  }
```
as this is a counterpart to https://github.com/magma/magma/blob/c1f15ee8cf1ff8eb60cadcf7bfcc504e3806564d/lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c#L902 ?

ulaskozat: I missed that you actually removed the check `(!ue_context_p->emm_context.esm_ctx.is_pdn_disconnect)`

ulaskozat (approved):
LGTM

PR #7724, opened 2021-06-23
fix(pipelined):Cleanup flows on timed out installation

koolzz:

## Summary

- Change default rpc timeout to 6 same as in sessiond
- Attempt to delete flows that were possibly installed

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/rpc_servicer.py
+++ b/lte/gateway/python/magma/pipelined/rpc_servicer.py
@@ -43,6 +43,7 @@ from lte.protos.pipelined_pb2 import (
     UESessionSet,
     UPFSessionContextState,
     VersionedPolicy,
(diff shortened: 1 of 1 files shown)


lte/gateway/python/magma/pipelined/rpc_servicer.py
```diff
@@ -71,7 +72,7 @@
 )
 
 grpc_msg_queue = queue.Queue()
-DEFAULT_CALL_TIMEOUT = 15
+DEFAULT_CALL_TIMEOUT = 6
```
pshelar: can we reduce it to 5 sec to have lower probability of inconsistent timeout between pipelineD and sessionD?

pshelar (approved):
LGTM

PR #7732, opened 2021-06-23
build(mme): fix perm error in AGW docker build appliance

markjen:
## Summary

`find .` fails in some environments (e.g. linux 5.12.10-300.fc34.x86_64 w/
docker 20.10.7) because some paths in /proc/... have special permissions
(e.g. /proc/tty/driver is 500 and owned by nobody:nogroup). Also, there is no
reason to `find .` from `/`. This change instructs libfluid_build to output
to `/tmp` and looks for the resulting package there for install.

## Test Plan

docker build succeeds and `make oai_mme` in /lte/gateway within the resulting docker image succeeds.

### DIFF
--- a/lte/gateway/docker/mme/Dockerfile.ubuntu20.04
+++ b/lte/gateway/docker/mme/Dockerfile.ubuntu20.04
@@ -260,8 +260,8 @@ RUN git clone https://github.com/include-what-you-use/include-what-you-use && \
 
 #### libfluid is requird to build MME with OVS support
 COPY third_party /tmp/third_party/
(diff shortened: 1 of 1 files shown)


themarwhal (approved):

electronjoe (approved):
Thanks for the improvement Mark!

pshelar (approved):
PR looks good.
Can you change the commit subject line to 'build(mme):...', this would make semantic pull request check happy. lte is not included in the scope list.

markjen:
#ready2merge

PR #7741, opened 2021-06-24
fix(pipelined): add recovery for ovs restarts 

koolzz:

## Summary

 - Make pipelined part of ovs to avoid crazy errors when ovs goes down
 - Increment rule versions when ovs gets restarted to avoid issues with rules being out of sync if ovs goes down.

## Additional Information


pshelar:
yes, Let keep it for master.
For 1.6 release, we need to document what operator need to do if OVS is restarted.
It would be nice if there is script to detect that OVS was restarted, so we dont need to check it manually.

koolzz:
> yes, Let keep it for master.
> For 1.6 release, we need to document what operator need to do if OVS is restarted.
> It would be nice if there is script to detect that OVS was restarted, so we dont need to check it manually.

I meant for 1.6 we can apply this PR. it'll auto restart pipelined if ovs restarts.

Its not perfect because I think we lose a few seconds of traffic accounting because on startup flow diff logic(because we bump the rule versions in sessiond with this PR) we don't track stats from the deleted flows. But thats a minor difference which is not worth tackling for 1.6

lte/gateway/c/session_manager/LocalSessionManagerHandler.cpp
```diff
@@ -206,8 +207,16 @@ void LocalSessionManagerHandlerImpl::call_setup_pipelined(
   }
   pipelined_state_ = PipelineDState::SETTING_UP;
 
+  auto session_map    = session_store_.read_all_sessions();
+  auto session_update = SessionStore::get_default_session_update(session_map);
+  enforcer_->increment_all_policy_versions(session_update);
```
themarwhal: Should we just pass in `session_map` here?

themarwhal:
Yeah as I said before, my only concern here is the rule versions might be incremented too many times if we get stuck in a setup loop.

If we do some extra logic in the setup handling we may be able to increment the versions only once per setup exchange. (as opposed to incrementing every time we call setup)

setup-states

Since this is for 1.6, I think we can land this for now, and if we need to we can add some extra logic on the sessiond side.

lte/gateway/c/session_manager/LocalSessionManagerHandler.cpp
```diff
@@ -206,8 +207,16 @@ void LocalSessionManagerHandlerImpl::call_setup_pipelined(
...
+  bool update_success = session_store_.update_sessions(session_update);
+  if (!update_success) {
+    MLOG(MERROR) << "Failed to update rule versions after ovs restart";
+  }
+  MLOG(MERROR) << "Successfully updated rule versions after ovs restart";
```
themarwhal: `MLOG(MERROR)` -> `MLOG(MINFO)`

lte/gateway/c/session_manager/LocalSessionManagerHandler.cpp
```diff
@@ -77,8 +77,12 @@ void LocalSessionManagerHandlerImpl::ReportRuleStats(
     MLOG(MINFO) << "Pipelined has been restarted, attempting to sync flows,"
                 << " old epoch = " << current_epoch_
                 << ", new epoch = " << reported_epoch_;
+
     enforcer_->get_event_base().runInEventBaseThread(
-        [this, epoch = reported_epoch_]() { call_setup_pipelined(epoch); });
+        [this, epoch = reported_epoch_,
+         update_rule_versions = request_cpy.update_rule_versions()]() {
+          call_setup_pipelined(epoch, update_rule_versions);
```
pshelar, line 84: why not call deactivate flows for all the rules?

koolzz: 2 things:
 - when we do this we lose some flow state, basically any traffic since last report will not get reported(because we bump the version and remove old flow)
 - extra processing to remove all / add all rules back (which happen swhen you increment version)

pshelar (approved):

themarwhal (approved):

PR #7743, opened 2021-06-24
ci: Turn off codecov comment

hcgatewood:
## Summary

Comment is useless noise due to MME codecov pipeline. Turning off to remove the noisy comments.

## Test Plan

test_in_prod

## Additional Information


electronjoe (approved):
nit: YAML lint that doc should start with `---` could be addressed now (or not)

PR #7758, opened 2021-06-24
Move stats collection of eNB stats into s1ap thread

amarpad:
## Summary
The current access pattern violates the ITTI framework so
1 - Create a stats timer for S1ap to send stats to service303
2 - Remove fields from mme_app structure relating to eNB (s1).
3 - Move the increment of enb stats to post s1setup request vs when the
association gets enabled. This makes more sense and maps to where the
increment used to happen for the cross thread stats
Also was missing a break for handling APPLICATION_MME_APP_STATS_MSG which
triggered the premature termination of the service303 thread, regression
introduced in by me in 118e113. Fixed that as well.

## Test Plan

1 - Added instrumentation logs to validate that the timer was getting triggered multiple times

2 - make integ_test (on debian)

## Additional Information


lte/gateway/c/core/oai/tasks/service303/service303_task.c
```diff
@@ -81,10 +81,14 @@ static int handle_service_message(zloop_t* loop, zsock_t* reader, void* arg) {
...
       service303_set_application_health(APP_UNHEALTHY);
     } break;
-    case APPLICATION_STATS_MSG: {
-      service303_mme_statistics_read(
-          &received_message_p->ittiMsg.application_mme_stats_msg);
-    }
+    case APPLICATION_MME_APP_STATS_MSG: {
+      service303_mme_app_statistics_read(
+          &received_message_p->ittiMsg.application_mme_app_stats_msg);
+    } break;
```
amarpad, line 87: @ulaskozat , @themarwhal : Oversight on missing the break here, so the first sample gets recorded correctly but subsequent samples don't as the thread exists. Apologies.

ulaskozat:
Can you move stats printing in debug mode from mme app to 303 as it has now all the stats?

amarpad:
> Can you move stats printing in debug mode from mme app to 303 as it has now all the stats?

I think that is a separate cleanup. This is a correctness fix.

ulaskozat (approved):
LGTM

ardzoht (approved):
lgtm

PR #7762, opened 2021-06-24
refactor: [ci] Put all helm dependency checks into one check

quentinDERORY:

## Summary
Helm checks were too noisy in the PR checks tab.
So we're trimming it down to one github action job.
Still have same functionality

## Test Plan

Tested on PR itself

## Additional Information

### DIFF
--- a/.github/workflows/helm-chart-dependency-check.yml
+++ b/.github/workflows/helm-chart-dependency-check.yml
@@ -7,28 +7,59 @@ jobs:
     env:
       MAGMA_ROOT: "${{ github.workspace }}"
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        charts:
-          - ["orc8r", "$MAGMA_ROOT/orc8r/cloud/helm/orc8r/"]
-          - ["cwf-orc8r", "$MAGMA_ROOT/cwf/cloud/helm/cwf-orc8r/"]
-          - ["lte-orc8r", "$MAGMA_ROOT/lte/cloud/helm/lte-orc8r/"]
-          - ["feg-orc8r", "$MAGMA_ROOT/feg/cloud/helm/feg-orc8r/"]
-          - ["fbinternal-orc8r", "$MAGMA_ROOT/fbinternal/cloud/helm/fbinternal-orc8r/"]
-          - ["wifi-orc8r", "$MAGMA_ROOT/wifi/cloud/helm/wifi-orc8r/"]
     name: Check dependency of helm chart ${{ matrix.charts[0] }}
     steps:
       - uses: actions/checkout@v2
-      - name: Get chart.lock digest
+      - name: Check Orc8r
         run: |
-          echo "DIGEST=$(cat ${{ matrix.charts[1] }}Chart.lock | grep digest | cut -d ":" -f 2-3 | xargs)" >> $GITHUB_ENV
-      - name: Run dependency check against ${{ matrix.charts[0] }}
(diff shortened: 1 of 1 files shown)


.github/workflows/helm-chart-dependency-check.yml
```diff
@@ -7,28 +7,59 @@ jobs:
...
-          echo "DIGEST=$(cat ${{ matrix.charts[1] }}Chart.lock | grep digest | cut -d ":" -f 2-3 | xargs)" >> $GITHUB_ENV
-      - name: Run dependency check against ${{ matrix.charts[0] }}
+          echo "DIGEST=$(cat $MAGMA_ROOT/orc8r/cloud/helm/orc8r/Chart.lock | grep digest | cut -d ":" -f 2-3 | xargs)" >> $GITHUB_ENV
+          helm dependency update "$MAGMA_ROOT/orc8r/cloud/helm/orc8r/"
+          echo "NEW_DIGEST=$(cat $MAGMA_ROOT/orc8r/cloud/helm/orc8r/Chart.lock | grep digest | cut -d ":" -f 2-3 | xargs)" >> $GITHUB_ENV
+          if [ "$NEW_DIGEST" != "$NEW_DIGEST" ]; then
+            exit 1
+          fi
+      - name: Check cwf-orc8r
+        if: always()
```
hcgatewood, line 22: `if: always()` seems like we can just remove that line (?)

.github/workflows/helm-chart-dependency-check.yml
```diff
@@ -7,28 +7,59 @@ jobs:
...
+            exit 1
+          fi
+      - name: Check lte-orc8r
+        if: always()
         run: |
-          echo "NEW_DIGEST=$(cat ${{ matrix.charts[1] }}Chart.lock | grep digest | cut -d ":" -f 2-3 | xargs)" >> $GITHUB_ENV
-      - name: Verify digest is the same
-        if: ${{ env.DIGEST != env.NEW_DIGEST }}
-        run: exit 1
+          echo "DIGEST=$(cat $MAGMA_ROOT/lte/cloud/helm/lte-orc8r/Chart.lock | grep digest | cut -d ":" -f 2-3 | xargs)" >> $GITHUB_ENV
```
hcgatewood, line 33: Is there any way to store this run cmd as a "function", and call it as a one-liner for each job?

hcgatewood, line 33: Is there any way to store this run cmd as a "function", and call it as a one-liner for each job?

hcgatewood (approved):
Thanks for making this fix 🙂 Couple small comments

.github/workflows/helm-chart-dependency-check.yml
quentinDERORY: No we can't because we want all the checks to run even if one fails before.

.github/workflows/helm-chart-dependency-check.yml
quentinDERORY: That's where github actions is lacking. 
It does not support creation of function or yaml anchoring.
So there are three solutions possible: create my own github action module(in another file), use an external script or the matrix trick(^^) 
The first two are kind of overkill and would just be too much files. And we don't want matrix so I think we're best just repeating this code.

.github/workflows/helm-chart-dependency-check.yml
hcgatewood: Ah I see. Good to know, I didn't realize that was the default, thanks for explaining 🙂

.github/workflows/helm-chart-dependency-check.yml
hcgatewood: Yeah makes sense, agreed

PR #7776, opened 2021-06-25
initial experimentations on bazel build

themarwhal:

## Important
use this new dockerfile
```
cd lte/gateway/docker/mme
docker build -t bazel_base -f Dockerfile.ubuntu20.04.bazel ../../../../
docker run -v $MAGMA_ROOT:/magma -v $MAGMA_ROOT/lte/gateway/configs:/etc/magma -i -t bazel_base:latest /bin/bash
```

### Running Bazel commands
```
cd /magma
bazel build $PATH_TO_PACKAGE:$LIBRARY_NAME
bazel build ...
bazel test ...
# example: bazel build orc8r/gateway/c/common/config:config
```

### Notable Changes
1. upgraded cpp-redis to be latest release since that version included Bazel support. This meant I've had to slightly fix up some types but nothing too big.
2. session manager on master has two sub directories called datastore and policdb. It was annoying to deal with the dependency chain across directories, so I just flattened it into the main session manager directory. (The two sub directories in total had 5 files, so I figured it was no big deal)

## Additional Information


lte/gateway/c/session_manager/magma_logging.h
```diff
@@ -0,0 +1,41 @@
...
+#include <iostream>
+
+// for non glog use cases, just log to std cout
+struct _MLOG_NEWLINE {
+  ~_MLOG_NEWLINE() { std::cout << std::endl; }
+};
+#define MLOG(VERBOSITY)                                                        \
+  (_MLOG_NEWLINE(), std::cout << "[" << __FILE__ << ":" << __LINE__ << "] ")
+
+#endif
```
uri200, line 41: isnt this a common library?

or are you placing this just for test?

themarwhal: This I just placed until I figure out how to import it from common

Taking some shortcuts to get some sort of build first :)

themarwhal: Well importing isn't the hard part. Making it work for bazel while not breaking the cmake build is harder😩

uri200: well copying temporary there to allow a migration is fine. We can remove it later

Problem is if this happens with many more.

Is it possible to have hybrid builds, meaning, sessiond all using Bazel, but the rest still using cmake?

themarwhal: should be yes! they are separate builds already, so that should not be a problem

PR #7782, opened 2021-06-25
Break after handing stats message

amarpad:
## Summary
Missing break during case handling

## Test Plan

make integ_test

## Additional Information


ulaskozat (approved):
LGTM

PR #7788, opened 2021-06-25
[MME][LOG] Use unsigned decimal for all UE ID logging

talkhasib:

## Summary

We currently use a mix of decimal and hex logging for UE IDs. This is a bit confusing when parsing MME logs.
Unify logging to use unsigned decimal

## Test Plan

Built on local vagrant box and tested using s1ap integ test


ardzoht (approved):
lgtm

PR #7794, opened 2021-06-26
[lte][agw] Change is_standalone to address multiple concurrent sessions

ulaskozat:

## Summary

- Spirent tests revealed that concurrent standalone PDN requests were not handled properly as any of the responses to standalone activation request was reseting the boolean flag. Instead the flag is changed to int and pending standalone requests are counted.

## Test Plan

Integ tests.
Plan to add new integ test.

## Additional Information


ardzoht (approved):
lgtm, great catch

PR #7801, opened 2021-06-28
perf(pipelined): Use async callback for QoS to reduce attach time

pshelar:
Creating QoS object is expensive in terms of time. This PR moves
bulk of the work in async callback to reduce latency of session
create.

## Test Plan
validated with integ-tests on local setup.

## Additional Information


koolzz (approved):
maybe rename to cleanup_rule_lambda or cleanup_rule_func

PR #7746, opened 2021-06-24
feat(nms): Equipment Dashboard Federation Gateway KPIs

ato-dani:
## Summary
This is a draft PR because it is a stacked pr upon an existing open PR #7700 . Once, that PR is approved, this PR would get rebased and would be changed from draft to a normal PR. Context provider was added to share information about the federation gateways, their health and the active gateway id instead of fetching it in each page. Since federation gateways can be a maximum of 2, sharing this information doesn't consume memory. Using the Gateway information, KPIs were added to it to better understand the gateway status. 

## Test Plan
A test was added in a file FEGEquipmentGatewayTest.js. It tests the KPIs information and how the context providers gets the data that is used in the calculation to get the KPI. Also, previous tests were ran and all worked without error.

## Additional Information
Screen Shot 2021-06-24 at 10 08 38 AM

PR #7804, opened 2021-06-28
feat(nms): Equipment dashboard Federation Gateway Table

### LINKED ISSUE #2842: Add Federation Equipment dashboard and Gateway overview page

ato-dani:
## Summary
This is a draft PR because it is a stacked pr upon an existing open PR #7746 . Once, that PR is approved, this PR would be changed from draft to a normal PR. A federation gateway table was added to the equipment dashboard. It would display the basic configurations of the federation gateways. 

## Test Plan
A test was added under FEGEquipmentGatewayTest.js file to test how the federation gateway table is rendered. 

## Additional Information
Screen Shot 2021-06-28 at 1 51 09 PM

### DIFF
--- a/nms/app/packages/magmalte/app/components/context/RefreshContext.js
+++ b/nms/app/packages/magmalte/app/components/context/RefreshContext.js
@@ -15,16 +15,28 @@
  */
 'use strict';
 import {FetchEnodebs, FetchGateways} from '../../state/lte/EquipmentState';
+import {
+  FetchFegGateways,
+  getActiveFegGatewayId,
+  getFegGatewaysHealthStatus,
+} from '../../state/feg/EquipmentState';
 import {FetchSubscriberState} from '../../state/lte/SubscriberState';
 import {useContext, useEffect, useRef, useState} from 'react';
 import type {EnqueueSnackbarOptions} from 'notistack';
 
 export const REFRESH_INTERVAL = 30000;
 
+export const RefreshTypeEnum = Object.freeze({
+  SUBSCRIBER: 'subscriber',
+  GATEWAY: 'gateway',
+  FEG_GATEWAY: 'feg_gateway',
+  ENODEB: 'enodeb',
+});
+
 type Props = {
   context: typeof React.Context,
   networkId: string,
-  type: refreshType,
+  type: RefreshType,
   interval?: number,
   id?: string,
   enqueueSnackbar?: (
@@ -35,13 +47,23 @@ type Props = {
   lastRefreshTime?: string,
 };
 
-type refreshType = 'subscriber' | 'gateway' | 'enodeb';
+type RefreshType = $Values<typeof RefreshTypeEnum>;
 
 export function useRefreshingContext(props: Props) {
   const ctx = useContext(props.context);
-  const [state, setState] = useState(
-    props.type === 'subscriber' ? {sessionState: ctx.sessionState} : ctx.state,
-  );
+  const initState = () => {
+    if (props.type === RefreshTypeEnum.FEG_GATEWAY) {
+      return {
+        fegGateways: ctx?.state,
+        health: ctx?.health,
+        activeFegGatewayId: ctx?.activeFegGatewayId,
+      };
+    } else if (props.type === 'subscriber') {
+      return {sessionState: ctx.sessionState};
+    }
+    return ctx.state;
+  };
+  const [state, setState] = useState(initState());
 
   const [autoRefreshTime, setAutoRefreshTime] = useState(props.lastRefreshTime);
   async function fetchState(props: FetchProps) {
@@ -69,18 +91,23 @@ export function useRefreshingContext(props: Props) {
     if (id !== null && id !== undefined) {
       if (props.type === 'subscriber') {
         newState = {
-          // $FlowIgnore
+          // $FlowIgnore because state may not contain sessionState for other refresh type like enodeb
           sessionState: Object.keys(state.sessionState || {}).length
             ? {
                 ...ctx.sessionState,
-                // $FlowIgnore
+                // $FlowIgnore because state may not contain sessionState for other refresh type like enodeb
                 [id]: state.sessionState?.[id],
               }
             : {},
         };
       } else if (props.type === 'enodeb') {
-        // $FlowIgnore
+        // $FlowIgnore because state may not contain enbInfo for other refresh type like feg_gateway
         newState = {...ctx.state, [id]: state.enbInfo?.[id]};
+      } else if (props.type === RefreshTypeEnum.FEG_GATEWAY) {
+        newState = {
+          // $FlowIgnore because state may not contain fegGateways for other refresh type like subscriber
+          fegGateways: {...ctx.fegGateways, [id]: state?.fegGateways?.[id]},
+        };
       } else {
         newState = {...ctx.state, [id]: state?.[id]};
       }
@@ -88,6 +115,8 @@ export function useRefreshingContext(props: Props) {
     if (props.type === 'subscriber') {
       // update subscriber session state
       return ctx.setState(null, null, null, newState);
+    } else if (props.type === RefreshTypeEnum.FEG_GATEWAY) {
+      return ctx.setState(null, null, newState?.fegGateways || {});
     }
     return ctx.setState(null, null, newState);
   }
@@ -134,7 +163,7 @@ export function useRefreshingContext(props: Props) {
 }
 
 type FetchProps = {
-  type: refreshType,
+  type: RefreshType,
   networkId: string,
   id?: string,
   enqueueSnackbar?: (
@@ -164,6 +193,17 @@ async function fetchRefreshState(props: FetchProps) {
     });
 
     return gateways;
+  } else if (type === RefreshTypeEnum.FEG_GATEWAY) {
+    const fegGateways = await FetchFegGateways({
+      id: id,
+      networkId,
+      enqueueSnackbar,
+    });
+    const [health, activeFegGatewayId] = await Promise.all([
+      getFegGatewaysHealthStatus(networkId, fegGateways, enqueueSnackbar),
+      getActiveFegGatewayId(networkId, fegGateways, enqueueSnackbar),
+    ]);
+    return {fegGateways, health, activeFegGatewayId};
   } else {
     const enodebs = await FetchEnodebs({
       id: id,
--- a/nms/app/packages/magmalte/app/state/feg/EquipmentState.js
+++ b/nms/app/packages/magmalte/app/state/feg/EquipmentState.js
@@ -19,6 +19,7 @@ import type {FederationGatewayHealthStatus} from '../../components/GatewayUtils'
 import type {
   federation_gateway,
   gateway_id,
+  mutable_federation_gateway,
   network_id,
 } from '@fbcnms/magma-api';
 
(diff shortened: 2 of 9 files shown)


uri200:
I think that table would make more sense from top to down, this way we can add more columns per each service.

Does each service reports it's health to the cloud? If so we could show the health for gx, gy, s6a . And the whole gatway.

ato-dani:
> I think that table would make more sense from top to down, this way we can add more columns per each service.
> 
> Does each service reports it's health to the cloud? If so we could show the health for gx, gy, s6a . And the whole gatway.

@uri200 Those information would be displayed in the Gateway Detail Page. That would be my next milestone project. When you click the gateways, you should go to the details page, and there detailed information would be present so that we won't overload all info about the network just in one page.

uri200:
> > I think that table would make more sense from top to down, this way we can add more columns per each service.
> > Does each service reports it's health to the cloud? If so we could show the health for gx, gy, s6a . And the whole gatway.
> 
> @uri200 Those information would be displayed in the Gateway Detail Page. That would be my next milestone project. When you click the gateways, you should go to the details page, and there detailed information would be present so that we won't overload all info about the network just in one page.

oh I see. I was missing the header in that picture, so I couldn't tell. Thanks for clarifying

themarwhal (approved):
hmm not sure why I'm requested as an owner, will look into this. But approving to unblock. Since karthik is off, I'll add andrei and hanna for review

nms/app/packages/magmalte/app/components/context/RefreshContext.js
```diff
@@ -35,13 +42,24 @@ type Props = {
...
+type refreshType = 'subscriber' | 'gateway' | 'feg_gateway' | 'enodeb';
 
 export function useRefreshingContext(props: Props) {
   const ctx = useContext(props.context);
-  const [state, setState] = useState(
-    props.type === 'subscriber' ? {sessionState: ctx.sessionState} : ctx.state,
-  );
+  const initState = () => {
+    if (props.type === 'feg_gateway') {
+      // return federation gateways, their health status and the active federation gateway id
```
andreilee: this comment is redundant - the same info is obvious in the return statement right below.

nms/app/packages/magmalte/app/components/context/RefreshContext.js
```diff
@@ -35,13 +42,24 @@ type Props = {
   lastRefreshTime?: string,
 };
 
-type refreshType = 'subscriber' | 'gateway' | 'enodeb';
+type refreshType = 'subscriber' | 'gateway' | 'feg_gateway' | 'enodeb';
```
andreilee: String union types are a little strange to use. If somebody needs to update one of the values they would need to find and replace all occurrences.

```suggestion
const RefreshTypeEnum = Object.freeze({SUBSCRIBER: 'subscriber', GATEWAY: 'gateway', ...})
type RefreshType = $Values<typeof RefreshTypeEnum>;
```

https://stackoverflow.com/questions/287903/how-can-i-guarantee-that-my-enums-definition-doesnt-change-in-javascript
https://flow.org/en/docs/types/utilities/

andreilee: Where there's comparisons to `'feg_gateway'`, they can then be updated as well
```
if (props.type === RefreshTypeEnum.FEG_GATWAY) {
```

nms/app/packages/magmalte/app/components/context/RefreshContext.js
```diff
@@ -81,13 +99,20 @@ export function useRefreshingContext(props: Props) {
       } else if (props.type === 'enodeb') {
         // $FlowIgnore
         newState = {...ctx.state, [id]: state.enbInfo?.[id]};
+      } else if (props.type === 'feg_gateway') {
+        newState = {
+          // $FlowIgnore
```
andreilee: Why's this `$FlowIgnore` here? Add details to the comment at least

andreilee: Also add to the other `$FlowIgnore` annotations in the function

andreilee (changes requested):

nms/app/packages/magmalte/app/state/feg/EquipmentState.js
```diff
@@ -136,3 +233,50 @@ async function getActiveFegGatewayId(
...
+ * the specific gateway if the id is provided.
+ *
+ * @param {network_id} networkId: Id of the federation network.
+ * @param {gateway_id} id id of the federation gateway.
+ * @param {(msg, cfg,) => ?(string | number),} enqueueSnackbar Snackbar to display error
+ * @returns returns a list of the federation gateways or a federation gateway with the given id.
+ */
+export async function FetchFegGateways(props: FetchProps) {
+  const {networkId, id, enqueueSnackbar} = props;
+  if (id !== undefined && id !== null) {
```
andreilee: You should be able to do a direct truthiness check on `id` here. A zero-length `gateway_id` is nonsense anyways.

https://www.sitepoint.com/javascript-truthy-falsy/

nms/app/packages/magmalte/app/state/feg/EquipmentState.js
```diff
@@ -136,3 +233,50 @@ async function getActiveFegGatewayId(
...
+        variant: 'error',
+      });
+    }
+  } else {
+    try {
+      return await MagmaV1API.getFegByNetworkIdGateways({
+        networkId: networkId,
+      });
+    } catch (e) {
+      enqueueSnackbar?.('failed fetching gateway information', {
```
andreilee: Capitalize the error messages. These are read by NMS users and would be better as understandable sentences.

nms/app/packages/magmalte/app/state/feg/EquipmentState.js
```diff
@@ -136,3 +233,50 @@ async function getActiveFegGatewayId(
...
+};
+
+/**
+ * Fetches and returns the list of gateways under the federation network or
+ * the specific gateway if the id is provided.
+ *
+ * @param {network_id} networkId: Id of the federation network.
+ * @param {gateway_id} id id of the federation gateway.
+ * @param {(msg, cfg,) => ?(string | number),} enqueueSnackbar Snackbar to display error
+ * @returns returns a list of the federation gateways or a federation gateway with the given id.
```
andreilee: Update this to specify the type - not a list

andreilee: Also either this specify that it throws an error when it fails to fetch gateways, or returns an empty set

nms/app/packages/magmalte/app/components/context/RefreshContext.js
```diff
@@ -164,6 +193,17 @@ async function fetchRefreshState(props: FetchProps) {
...
 
     return gateways;
+  } else if (type === RefreshTypeEnum.FEG_GATEWAY) {
+    const fegGateways = await FetchFegGateways({
+      id: id,
+      networkId,
+      enqueueSnackbar,
+    });
+    const [health, activeFegGatewayId] = await Promise.all([
+      getFegGatewaysHealthStatus(networkId, fegGateways || {}, enqueueSnackbar),
```
andreilee: `fegGateways || {}` shouldn't be necessary here. Getting the health is arbitrary if there aren't any `fegGateways`.

nms/app/packages/magmalte/app/state/feg/EquipmentState.js
```diff
@@ -74,6 +75,102 @@ export async function InitGatewayState(props: InitGatewayStateProps) {
...
+ * @param {network_id} networkId Id of the federation network
+ * @param {{[gateway_id]: federation_gateway}} fegGateways Federation gateways of the network.
+ * @param {{[gateway_id]: FederationGatewayHealthStatus}} fegGatewaysHealthStatus Health status of the federation gateways.
+ * @param {({[string]: federation_gateway}) => void} setFegGateways Sets federation gateways.
+ * @param {({[string]: FederationGatewayHealthStatus}) => void} setFegGatewaysHealthStatus Sets federation gateways health status.
+ * @param {(gatewayId:gateway_id) => void} setActiveFegGatewayId Sets the active gateway id.
+ * @param {gateway_id} key Id of the gateway to be added, deleted or edited.
+ * @param {mutable_federation_gateway} value New Value for the gateway with the id: key.
+ * @param {[gateway_id]: federation_gateway} newState New State of the Federation Gateway.
+ * @param {(msg, cfg,) => ?(string | number),} enqueueSnackbar Snackbar to display error
```
andreilee: Specify the only `@param` as type `GatewayStateProps`, and add these helpful comments to the definition of `GatewayStateProps` above.

nms/app/packages/magmalte/app/state/feg/EquipmentState.js
```diff
@@ -136,3 +233,50 @@ async function getActiveFegGatewayId(
...
+        networkId: networkId,
+        gatewayId: id,
+      });
+      if (gateway) {
+        return {[id]: gateway};
+      }
+    } catch (e) {
+      enqueueSnackbar?.('failed fetching gateway information', {
+        variant: 'error',
+      });
```
andreilee: Either the whole function body should be wrapped in the same `try..catch`, or these two error messages should be different and provide some additional context. Right now it's just redundant.

nms/app/packages/magmalte/app/views/equipment/FEGGatewayTable.js
```diff
@@ -0,0 +1,231 @@
...
+          : '-',
+        gx: gateway.federation?.gx?.server?.address || '-',
+        gy: gateway.federation?.gy?.server?.address || '-',
+        swx: gateway.federation?.swx?.server?.address || '-',
+        s6a: gateway.federation?.s6a?.server?.address || '-',
+        csfb: gateway.federation?.csfb?.client?.server_address || '-',
+      });
+    });
+
+  return (
```
andreilee, line 138: We should probably display something different if there's no federation gateways. Say, if they just haven't been configured yet, the NMS would just appear broken. Defer to @jingtien if necessary.

nms/app/packages/magmalte/app/views/equipment/FEGGatewayTable.js
```diff
@@ -0,0 +1,231 @@
...
+    .filter((g: federation_gateway) => g.federation && g.id)
+    .map((gateway: federation_gateway) => {
+      fegGatewayRows.push({
+        name: gateway.name,
+        id: gateway.id,
+        is_primary: activeFegGatewayId === gateway.id,
+        health: health[gateway.id].status
+          ? health[gateway.id]?.status === HEALTHY_STATUS
+            ? HEALTHY_GATEWAY
+            : UNHEALTHY_GATEWAY
```
andreilee: Actually, these health status constants can be put into an enum as well.

nms/app/packages/magmalte/app/views/equipment/FEGGatewayTable.js
```diff
@@ -0,0 +1,231 @@
...
+          : '-',
```
andreilee: Include these into the health status enum as an unknown value type

andreilee (changes requested):

nms/app/packages/magmalte/app/views/equipment/FEGGatewayTable.js
ato-dani: I don't understand why it would appear broken. I am saying this because if there are no federation gateways in the network then all the headers of the table are there(name, id, gx ,...), but there is no data in there. Anyone seeing that would notice that there is no federation gateway to display under those headers. If you created a new test lte network go to the equipment page without adding any gateway, a blank table will be seen there. 
Screen Shot 2021-07-21 at 2 33 13 PM

I think anyone seeing this would notice that there are no gateways in the network(the same applies here too). Besides the count of 0 on table title will tell you that there are no federation gateways. Does that make sense?

andreilee: Sounds good. Thanks for the detailed response.

andreilee (approved):

PR #7810, opened 2021-06-28
Create subscriberdb_cache logic to register per subscriber digests

wangyyt1013:

## Summary

1. Add `perSubDigestLookup` store under the `DigestLookup` interface that stores digests of individual subscriber data objects, indexed by network and subscriber ID
2. Add logic in `subscriberdb` package for generating per subscriber digests of a network
3. Add logic in `subscriberdb_cache` that constantly updates the per subscriber digests at a configurable interval

## Test Plan

1. Ran all existing tests. All should pass
2. Added unit tests corresponding to the added functionalities

## Additional Information


lte/cloud/go/lte/const.go
```diff
@@ -95,5 +95,6 @@ const (
 	SubscriberStateType = "subscriber_state"
 
 	// MSISDNBlobstoreType etc. denote blob types stored in blobstore tables.
-	MSISDNBlobstoreType = "msisdn"
+	MSISDNBlobstoreType       = "msisdn"
+	PerSubDigestBlobstoreType = "per_sub_digest"
```
hcgatewood: This type is namespaced to the particular blobstore table, so looks like can store it in subscriberdb/storage

lte/protos/subscriberdb.proto
```diff
@@ -267,6 +267,13 @@ message ListSubscribersResponse {
   map<string, Digest> per_sub_digests = 4;
 }
 
+message SubscriberDigestByID {
```
hcgatewood: This is a weird name, since "by" usually means it's a map. Something like "DigestWithID" makes more sense to me -- thoughts?

lte/cloud/go/services/subscriberdb/protos/subscriberdb.proto
```diff
@@ -126,3 +127,10 @@ message ApnResourceInternal {
   // apn_resource is the original apn resource proto message.
   APNConfiguration.APNResource apn_resource = 3;
 }
+
+// SubscriberDigestByIDs is used internally in per-subscriber store to serialize and deserialize
+// a list of per-sub digests.
+message SubscriberDigestByIDs {
```
hcgatewood: Per the other comment in subscribers.proto, update this name analogously

orc8r/cloud/go/mproto/mproto.go
```diff
@@ -78,6 +78,12 @@ func MarshalDeterministic(pb proto.Message) ([]byte, error) {
 	return buf.Bytes(), err
 }
 
+// UnmarshalDeterministic decodes the proto message serialized by MarshalDeterministic.
+func UnmarshalDeterministic(buf []byte, pb proto.Message) error {
```
hcgatewood: Unmarshalling is always deterministic -- i.e the process of `obj -> bytes -> obj` is `1 -> many -> 1`, since you necessarily need the same object on either side of the serialization (or, at least, it's as "deterministic" as the language will allow [(ref)](https://github.com/golang/go/issues/33702))

So just call this "Unmarshal" -- that way you still have a paired unmarshal fn to use with MarshalDeterministic, but it isn't confusing/misleading in the name

The other way you could do this is create a "deterministic binary encoder" object that implements the  `BinaryMarshaler` and `BinaryUnmarshaler` interfaces https://golang.org/pkg/encoding/#BinaryMarshaler. But I'm fine with the way you're doing it since it's clear how to use them without adding the separate object

lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -52,9 +53,92 @@ func GetDigest(networkID string) (string, error) {
 	return digest, nil
 }
 
+// GetPerSubDigests generates a set of individual subscriber digests ordered by their IDs.
+func GetPerSubDigests(network string) ([]*lte_protos.SubscriberDigestByID, error) {
```
uri200: To make it consistent, can we use `Subscriber` instead of `Sub`

lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -52,9 +53,92 @@ func GetDigest(networkID string) (string, error) {
...
+		for _, subProto := range subProtos {
+			digest, err := mproto.HashDeterministic(subProto)
+			if err != nil {
+				// Swallow errors related to a single subscriber digest to avoid affecting the rest
+				glog.Errorf("Failed to generate digest for subscriber %+v of network %+v: %+v", subProto.Sid.Id, network, err)
+				digest = ""
+			}
+			fullDigest := digest + apnDigest
+			perSubDigest := &lte_protos.SubscriberDigestByID{
+				Digest: &lte_protos.Digest{Md5Base64Digest: fullDigest},
```
uri200, line 88: `fullDigest := digest + apnDigest` is the string concatenation of two Md5Base64 digests.

Is `fullDigest` stil a MD5base64?

orc8r/cloud/go/mproto/mproto_test.go
```diff
@@ -139,3 +139,79 @@ func getTestDataCompound() []*test.TestDataCompound {
 		},
 	}
 }
+
+func TestUnmarshalDeterministic(t *testing.T) {
```
hcgatewood: This can be a much simpler test -- just a sanity check. Because it's clear from the fns' impl that this should pass

lte/cloud/go/services/subscriberdb_cache/subscriberdb_cache/main.go
```diff
@@ -36,13 +38,19 @@ func main() {
...
+	if err := fact.InitializeFactory(); err != nil {
+		glog.Fatalf("Error initializing per-sub digest storage: %+v", err)
+	}
+	perSubDigestStore := subscriberdb_storage.NewPerSubDigestLookup(fact)
+
 	serviceConfig := subscriberdb_cache.MustGetServiceConfig()
 	glog.Infof("Subscriberdb_cache service config %+v", serviceConfig)
 
-	go subscriberdb_cache.MonitorDigests(digestStore, serviceConfig)
+	go subscriberdb_cache.MonitorDigests(digestStore, perSubDigestStore, serviceConfig)
```
hcgatewood: Nit: throughout the changes, anywhere you pass these 3 arguments, pass the service config first

lte/cloud/go/services/subscriberdb_cache/test_init/test_service_init.go
```diff
@@ -36,10 +38,13 @@ func StartTestService(t *testing.T) {
 	assert.NoError(t, err)
 	digestStore := storage.NewDigestLookup(db, sqorc.GetSqlBuilder())
 	assert.NoError(t, digestStore.Initialize())
+	fact := blobstore.NewEntStorage(subscriberdb.PerSubDigestTableBlobstore, db, sqorc.GetSqlBuilder())
```
hcgatewood: Here and throughout: don't use ent in tests https://github.com/magma/magma/blob/60df24fd4470e09dd02ab03c5429946010ad61eb/orc8r/cloud/go/blobstore/ent.go#L33-L39

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -47,48 +47,68 @@ func MonitorDigests(store storage.DigestLookup, config Config) {
...
+		err = perSubDigestStore.SetDigest(network, perSubDigests)
 		if err != nil {
-			multierror.Append(errors.Wrapf(err, "set digest"))
+			multierror.Append(errors.Wrapf(err, "set per sub digest"))
 			continue
 		}
-		digestsByNetwork[network] = digest
 	}
-	return digestsByNetwork, errs.ErrorOrNil()
+	return flatDigestsByNetwork, errs.ErrorOrNil()
```
hcgatewood: Return the persubdigests as well, for consistency/testability

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -47,48 +47,68 @@ func MonitorDigests(store storage.DigestLookup, config Config) {
...
 }
 
 // getNetworksToUpdate returns networks to renew or delete in the store.
-func getNetworksToUpdate(store storage.DigestLookup, updateIntervalSecs int) ([]string, []string, error) {
+func getNetworksToUpdate(digestStore storage.DigestLookup, updateIntervalSecs int) ([]string, []string, error) {
```
hcgatewood: Just keep this as "store", it's clear it's a digest store from the type and context

lte/cloud/go/services/subscriberdb_cache/worker_test.go
```diff
@@ -49,20 +55,26 @@ func TestSubscriberdbCacheWorker(t *testing.T) {
 	allNetworks, err := storage.GetAllNetworks(digestStore)
 	assert.NoError(t, err)
 	assert.Equal(t, []string{}, allNetworks)
-	digest, _, err := storage.GetDigest(digestStore, "n1")
+	digest, err := storage.GetDigest(digestStore, "n1")
 	assert.NoError(t, err)
 	assert.Equal(t, "", digest)
```
hcgatewood: Replace this type of check with assert.Empty

lte/cloud/go/services/subscriberdb_cache/worker_test.go
```diff
@@ -49,20 +55,26 @@ func TestSubscriberdbCacheWorker(t *testing.T) {
...
+	perSubDigests, err := perSubDigestStore.GetDigest("n1")
+	assert.NoError(t, err)
+	assert.Equal(t, 0, len(perSubDigests))
```
hcgatewood: Replace this type of check with assert.Empty

lte/cloud/go/services/subscriberdb_cache/worker_test.go
```diff
@@ -76,16 +88,30 @@ func TestSubscriberdbCacheWorker(t *testing.T) {
...
+			},
 		},
 		serdes.Entity,
 	)
 	assert.NoError(t, err)
 
 	clock.SetAndFreezeClock(t, clock.Now().Add(10*time.Minute))
-	subscriberdb_cache.RenewDigests(digestStore, serviceConfig)
-	digest, _, err = storage.GetDigest(digestStore, "n1")
+	subscriberdb_cache.RenewDigests(digestStore, perSubDigestStore, serviceConfig)
```
hcgatewood: Everywhere you call RenewDigests: grab the returned err and assert whether you expect it to be nil or not

lte/cloud/go/services/subscriberdb_cache/worker_test.go
```diff
@@ -76,16 +88,30 @@ func TestSubscriberdbCacheWorker(t *testing.T) {
...
+	assert.NotEqual(t, digestExpected, digest)
+
+	perSubDigests, err = perSubDigestStore.GetDigest("n1")
 	assert.NoError(t, err)
-	assert.NotEqual(t, digestCanon, digest)
+	// The individual subscriber digests are ordered by subscriber ID
+	assert.Equal(t, "11111", perSubDigests[0].Sid.Id)
+	assert.NotEqual(t, "", perSubDigests[0].Digest.GetMd5Base64Digest())
+	assert.Equal(t, "99999", perSubDigests[1].Sid.Id)
+	assert.NotEqual(t, "", perSubDigests[1].Digest.GetMd5Base64Digest())
```
hcgatewood: assert.NotEmpty

lte/cloud/go/services/subscriberdb_cache/worker_test.go
```diff
@@ -76,16 +88,30 @@ func TestSubscriberdbCacheWorker(t *testing.T) {
...
+	digest, err = storage.GetDigest(digestStore, "n1")
+	assert.NoError(t, err)
+	assert.NotEqual(t, digestExpected, digest)
+
+	perSubDigests, err = perSubDigestStore.GetDigest("n1")
 	assert.NoError(t, err)
-	assert.NotEqual(t, digestCanon, digest)
+	// The individual subscriber digests are ordered by subscriber ID
+	assert.Equal(t, "11111", perSubDigests[0].Sid.Id)
+	assert.NotEqual(t, "", perSubDigests[0].Digest.GetMd5Base64Digest())
```
hcgatewood: Add at least one (minimal, sanity-check) check in this test that ensures the *contents* of the per-sub digests are as-expected

lte/cloud/go/services/subscriberdb_cache/worker_test.go
```diff
@@ -76,16 +88,30 @@ func TestSubscriberdbCacheWorker(t *testing.T) {
...
+	subscriberdb_cache.RenewDigests(digestStore, perSubDigestStore, serviceConfig)
+	digest, err = storage.GetDigest(digestStore, "n1")
+	assert.NoError(t, err)
+	assert.NotEqual(t, digestExpected, digest)
+
+	perSubDigests, err = perSubDigestStore.GetDigest("n1")
 	assert.NoError(t, err)
-	assert.NotEqual(t, digestCanon, digest)
+	// The individual subscriber digests are ordered by subscriber ID
+	assert.Equal(t, "11111", perSubDigests[0].Sid.Id)
```
hcgatewood, line 126: Before indexing into perSubDigests, do an assert.Len

lte/cloud/go/services/subscriberdb/const.go
```diff
@@ -19,4 +19,6 @@ const (
 	EntityType = "subscriber"
 
 	LookupTableBlobstore = "subscriber_lookup_blobstore"
+
```
hcgatewood: Remove this newline since the vars are related

lte/cloud/go/services/subscriberdb/digest.go
hcgatewood: No, but doesn't matter since it's just a "small blob of arbitrary data" to the user/gw's perspective -- the only affordance we care about is determinism in creation and equality comparison

lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -52,9 +53,92 @@ func GetDigest(networkID string) (string, error) {
...
+
+	apnsByName, err := LoadApnsByName(network)
+	if err != nil {
+		return nil, err
+	}
+	perSubDigests := []*lte_protos.SubscriberDigestByID{}
+	token := ""
+	foundEmptyToken := false
+	for !foundEmptyToken {
+		// the resultant subProtos lists are already ordered by subscriber id due to the nature
```
hcgatewood: Consisency: sentence-case your comments

lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -52,9 +53,92 @@ func GetDigest(networkID string) (string, error) {
...
+		subProtos, nextToken, err := LoadSubProtosPage(0, token, network, apnsByName, lte_models.ApnResources{})
+		if err != nil {
+			return nil, err
+		}
+		for _, subProto := range subProtos {
+			digest, err := mproto.HashDeterministic(subProto)
+			if err != nil {
+				// Swallow errors related to a single subscriber digest to avoid affecting the rest
+				glog.Errorf("Failed to generate digest for subscriber %+v of network %+v: %+v", subProto.Sid.Id, network, err)
+				digest = ""
```
hcgatewood: Hm in this case, the sub's digest is still getting set. So I think, if we fail in any particular sub here, just fail for the entire network

Then, I think it makes sense to add some logic when we're retrieving digests in subscriberdb -- if the "last updated" for either type of digest is outdated by like 3x or more the regen interval, return "out of sync" for safety. Thoughts?

lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -52,9 +53,92 @@ func GetDigest(networkID string) (string, error) {
...
+	}
+
+	return perSubDigests, nil
+}
+
+// GetPerSubDigestsDiff computes the changeset between two lists of per-subscriber digests,
+// ordered by their subscriber IDs (unique within a network). It returns
+// 1. A set of subscribers that have been added/modified, with the new digests.
+// 2. A list of subscribers that have been removed.
+func GetPerSubDigestsDiff(
```
hcgatewood: General: don't wrap fn definitions unless they're like 120-140 lines in length

lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -52,9 +53,92 @@ func GetDigest(networkID string) (string, error) {
...
+		foundEmptyToken = nextToken == ""
+		token = nextToken
+	}
+
+	return perSubDigests, nil
+}
+
+// GetPerSubDigestsDiff computes the changeset between two lists of per-subscriber digests,
+// ordered by their subscriber IDs (unique within a network). It returns
+// 1. A set of subscribers that have been added/modified, with the new digests.
```
hcgatewood, line 102: 1 and 2 are both "ordered lists"

lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -52,9 +53,92 @@ func GetDigest(networkID string) (string, error) {
...
+func GetPerSubDigestsDiff(
+	all []*lte_protos.SubscriberDigestByID,
+	tracked []*lte_protos.SubscriberDigestByID,
+) (map[string]string, []string) {
+	n, m, i, j := len(all), len(tracked), 0, 0
+	toRenew := map[string]string{}
+	deleted := []string{}
+
+	for i < n && j < m {
+		if all[i].Sid.Id == tracked[j].Sid.Id {
```
hcgatewood: Define `iSid, jSid` at the top of this block for conciseness

lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -52,9 +53,92 @@ func GetDigest(networkID string) (string, error) {
...
+			perSubDigests = append(perSubDigests, perSubDigest)
+		}
+		foundEmptyToken = nextToken == ""
+		token = nextToken
+	}
+
+	return perSubDigests, nil
+}
+
+// GetPerSubDigestsDiff computes the changeset between two lists of per-subscriber digests,
```
hcgatewood: Great job on this fn, I was worried its impl would look messy but you made it nice and clear

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -251,3 +252,193 @@ func TestGetDigestApnResourceAssocs(t *testing.T) {
...
+
+func TestGetPerSubDigests(t *testing.T) {
+	lte_test_init.StartTestService(t)
+	configurator_test_init.StartTestService(t)
+
+	err := configurator.CreateNetwork(configurator.Network{ID: "n1"}, serdes.Network)
+	assert.NoError(t, err)
+	_, err = configurator.CreateEntity("n1", configurator.NetworkEntity{Type: orc8r.MagmadGatewayType, Key: "g1", PhysicalID: "hw1"}, serdes.Entity)
+	assert.NoError(t, err)
+	gw, err := configurator.CreateEntity("n1", configurator.NetworkEntity{Type: lte.CellularGatewayEntityType, Key: "g1"}, serdes.Entity)
```
hcgatewood, line 264: Nit: create these 2 in 1 CreateEntities call

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -251,3 +252,193 @@ func TestGetDigestApnResourceAssocs(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	// Initially there should only be the placeholder apn digest in the generated digest map
+	perSubDigests, err := subscriberdb.GetPerSubDigests("n1")
+	assert.NoError(t, err)
+	assert.Equal(t, []*lte_protos.SubscriberDigestByID{}, perSubDigests)
+
+	// Generate individual digests for each newly detected subscriber in the network
+	_, err = configurator.CreateEntities("n1", configurator.NetworkEntities{
```
hcgatewood: Nit here and elsewhere: move each arg to "their own" line -- so newline after opening `(` and before closing `)`, etc

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -251,3 +252,193 @@ func TestGetDigestApnResourceAssocs(t *testing.T) {
...
+		configurator.NetworkEntity{
+			Type: lte.APNEntityType, Key: "apn",
+			Config: &lte_models.ApnConfiguration{},
+		},
+	}, serdes.Entity)
+	assert.NoError(t, err)
+
+	perSubDigests, err = subscriberdb.GetPerSubDigests("n1")
+	assert.NoError(t, err)
+	assert.Equal(t, "0001", perSubDigests[0].Sid.Id)
```
hcgatewood, line 301: Here and elsewhere: assert.Len before indexing into an obj

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -251,3 +252,193 @@ func TestGetDigestApnResourceAssocs(t *testing.T) {
...
+		configurator.NetworkEntity{
+			Type: lte.SubscriberEntityType, Key: "IMSI0001",
+			Config: &models.SubscriberConfig{
+				Lte: &models.LteSubscription{State: "INACTIVE"},
+			},
+		}, serdes.Entity,
+	)
+	assert.NoError(t, err)
+
+	perSubDigests, err = subscriberdb.GetPerSubDigests("n1")
```
hcgatewood: Add a comment summarizing what you're testing in this block

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -251,3 +252,193 @@ func TestGetDigestApnResourceAssocs(t *testing.T) {
...
+		toRenew, deleted := subscriberdb.GetPerSubDigestsDiff(all, tracked)
+
+		assert.Equal(t, map[string]string{
+			"00000": "apple",
+			"00001": "banana",
+			"00002": "cherry",
+		}, toRenew)
+		assert.Equal(t, []string{}, deleted)
+	})
+	t.Run("both not empty, case 1", func(t *testing.T) {
```
hcgatewood: "case 1" and "case 2" aren't descriptive enough -- perhaps "basic" and "involved" or something

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -251,3 +252,193 @@ func TestGetDigestApnResourceAssocs(t *testing.T) {
...
+		all := []*lte_protos.SubscriberDigestByID{
```
hcgatewood: Here and elsewhere: "all" and "tracked" don't connote the difference between the two. Which is which?

Maybe something like "prev" and "next" ?

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -251,3 +252,193 @@ func TestGetDigestApnResourceAssocs(t *testing.T) {
...
+			{Sid: &lte_protos.SubscriberID{Id: "00001"}, Digest: &lte_protos.Digest{Md5Base64Digest: "apple"}},
+			{Sid: &lte_protos.SubscriberID{Id: "00002"}, Digest: &lte_protos.Digest{Md5Base64Digest: "banana"}},
```
hcgatewood, line 416: Your use of metasyntactic variables is still a little confusing to me, here and below. Here's what I'd do

- Always keep the mapping between sid->metasyntactic_variable the same, e.g. "00001"->"apple"
- If you need to "update" the digest, go from "apple" to e.g. "apple2"
- Only 1 name per letter, i.e. don't use both "cherry" and "citrus" -- use "cherry" and "date"
- Since we're alphabetizing, don't skip around in the alphabet -- use e.g. "date" "egg" "fig" instead of "sand" "dirt" "sky" 

This is kind of a nit but it makes reading the tests much easier. Lmk if you want to chat about it

lte/cloud/go/services/subscriberdb/storage/digest.go
```diff
@@ -32,27 +33,27 @@ type DigestLookup interface {
...
 
-	digestLookupNidCol             = "network_id"
-	digestLookupDigestCol          = "digest"
-	digestLookupLastUpdatedTimeCol = "last_updated_at"
+	digestNidCol             = "network_id"
+	digestDigestCol          = "digest"
+	digestLastUpdatedTimeCol = "last_updated_at"
 )
 
 func NewDigestLookup(db *sql.DB, builder sqorc.StatementBuilder) DigestLookup {
```
hcgatewood: "NewDigestStore" perhaps, and change the obj name etc

lte/cloud/go/services/subscriberdb/storage/digest.go
```diff
@@ -191,14 +190,15 @@ func GetOutdatedNetworks(l DigestLookup, lastUpdatedBefore int64) ([]string, err
...
 	if err != nil {
 		return nil, err
 	}
 	networks := digestInfos.Networks()
-	return networks, nil
+	networksUniq := funk.UniqString(networks)
+	return networksUniq, nil
 }
 
 type digestInfo struct {
```
hcgatewood: digestInfo and its fields should all be public -- since we're making DigestInfos public

lte/cloud/go/services/subscriberdb/storage/digest.go
```diff
@@ -166,18 +167,16 @@ func (l *digestLookup) DeleteDigests(networks []string) error {
...
-func GetDigest(l DigestLookup, network string) (string, int64, error) {
+func GetDigest(l DigestLookup, network string) (string, error) {
 	digestInfos, err := l.GetDigests([]string{network}, clock.Now().Unix())
 	if err != nil {
-		return "", 0, err
+		return "", err
 	}
-	// There should be at most 1 digest for each network
-	// if digest not found, return default value
+	// Each network has at most 1 digest; if digest not set yet, return empty digest
```
hcgatewood: I like "...; if digest not found, ..." since it more accurately describes the actual experience

lte/cloud/go/services/subscriberdb/storage/per_sub_digest.go
```diff
@@ -0,0 +1,118 @@
...
+	"magma/orc8r/cloud/go/mproto"
+	"magma/orc8r/cloud/go/storage"
+	merrors "magma/orc8r/lib/go/errors"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+)
+
+const (
+	perSubDigestBlobstoreNetworkKey = "per_sub_digest_network"
```
hcgatewood: Add note that this is a placeholder network, since we're gonna use the network as the key

Also suffix with "_internal"

lte/cloud/go/services/subscriberdb/storage/per_sub_digest.go
```diff
@@ -0,0 +1,118 @@
...
+func (l *PerSubDigestLookup) GetDigest(network string) ([]*lte_protos.SubscriberDigestByID, error) {
+	store, err := l.fact.StartTransaction(&storage.TxOptions{ReadOnly: true})
+	if err != nil {
+		return nil, errors.Wrapf(err, "error starting transaction")
+	}
+	defer store.Rollback()
+
+	blob, err := store.Get(perSubDigestBlobstoreNetworkKey, storage.TypeAndKey{Type: lte.PerSubDigestBlobstoreType, Key: network})
+	if err == merrors.ErrNotFound {
+		// if per-sub digests for this network is not set yet, return empty list
```
hcgatewood: Sentence-case comments

lte/cloud/go/services/subscriberdb/storage/per_sub_digest.go
```diff
@@ -0,0 +1,118 @@
...
+// SetDigest creates or updates the per-subscriber digests of a network.
+func (l *PerSubDigestLookup) SetDigest(network string, digests []*lte_protos.SubscriberDigestByID) error {
+	store, err := l.fact.StartTransaction(&storage.TxOptions{ReadOnly: true})
+	if err != nil {
+		return errors.Wrapf(err, "error starting transaction")
+	}
+	defer store.Rollback()
+
+	// The sorted list of per-sub digests are serialized to be stored as a single blob.
+	// This is to preserve the sorted order of the digests AND decrease the frequency
```
hcgatewood: AND -> and

hcgatewood (changes requested):
Shape looks perfect. Big handful of comments then we can move forward

lte/cloud/go/services/subscriberdb_cache/worker_test.go
wangyyt1013: Could you elaborate on this? Currently the contents of the per-sub digests are generated by the worker code and can't exactly be predicted by the test code, unless we introduce `mproto` library here and try to manually generate a per-subscriber digest ourselves @hcgatewood

lte/cloud/go/services/subscriberdb/digest.go
wangyyt1013: Currently the per-sub digest store doesn't store `last_updated` information; this info is stored exclusively in the flat digest store as the only source of truth when it comes to last-updated time. Therefore, if per-sub digest generation failed for a network, should the overall digest generation (including flat digest) just fail for this network, so that the `last_updated` time for this network is not updated to `now()`? @hcgatewood

lte/cloud/go/services/subscriberdb/digest.go
wangyyt1013: I think here 1 is a map of sid -> digest (unordered) @hcgatewood

lte/cloud/go/services/subscriberdb_cache/worker_test.go
hcgatewood: Yeah I was imagining we use mproto to generate the digest in the test then compare

lte/cloud/go/services/subscriberdb/digest.go
hcgatewood: Yeah I think let's go with that -- on a per-network basis, only update the "last_updated" info if you fully+successfully update all digests for that network

Then on the read side of things, we can take into account the "last_updated" and, like I mentioned, if it's too long ago we can say "must resync"

Thoughts?

lte/cloud/go/services/subscriberdb/digest.go
hcgatewood: Ah you're right. Why are we using a map vs. slice for 1 vs. 2? Preference for them to be the same, but lmk if there's a concrete reason why we want to keep them as-is

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -251,3 +254,202 @@ func TestGetDigestApnResourceAssocs(t *testing.T) {
...
+
+	err := configurator.CreateNetwork(configurator.Network{ID: "n1"}, serdes.Network)
+	assert.NoError(t, err)
+	gw, err := configurator.CreateEntity("n1", configurator.NetworkEntity{Type: lte.CellularGatewayEntityType, Key: "g1"}, serdes.Entity)
+	assert.NoError(t, err)
+
+	// Initially the per-sub digest list should be empty (but not nil)
+	perSubDigests, err := subscriberdb.GetPerSubscriberDigests("n1")
+	assert.NoError(t, err)
+	assert.Empty(t, []*lte_protos.SubscriberDigestWithID{}, perSubDigests)
```
hcgatewood: Double check all the places you use assert.Empty -- the third arg is "message" or something, so you actually only want to pass 2 args

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -251,3 +254,202 @@ func TestGetDigestApnResourceAssocs(t *testing.T) {
...
+
+		assert.Equal(t, map[string]string{
+			"00000": "apple",
+			"00001": "banana",
+			"00002": "cherry",
+		}, toRenew)
+		assert.Empty(t, deleted)
+	})
+	t.Run("both not empty, basic", func(t *testing.T) {
+		next := []*lte_protos.SubscriberDigestWithID{
```
hcgatewood: When doing "prev" and "next", move the "prev" first, above "next", since it reads better in that order

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -251,3 +254,202 @@ func TestGetDigestApnResourceAssocs(t *testing.T) {
...
+		}
+		prev := []*lte_protos.SubscriberDigestWithID{
+			{Sid: &lte_protos.SubscriberID{Id: "00001"}, Digest: &lte_protos.Digest{Md5Base64Digest: "orange"}},
+			{Sid: &lte_protos.SubscriberID{Id: "00002"}, Digest: &lte_protos.Digest{Md5Base64Digest: "banana"}},
+			{Sid: &lte_protos.SubscriberID{Id: "00004"}, Digest: &lte_protos.Digest{Md5Base64Digest: "dragonfruit"}},
+			{Sid: &lte_protos.SubscriberID{Id: "00006"}, Digest: &lte_protos.Digest{Md5Base64Digest: "fig"}},
+			{Sid: &lte_protos.SubscriberID{Id: "00007"}, Digest: &lte_protos.Digest{Md5Base64Digest: "grape"}},
+		}
+		toRenew, deleted := subscriberdb.GetPerSubscriberDigestsDiff(next, prev)
+		assert.Equal(t, map[string]string{
```
hcgatewood: Nit for here and elsewhere: like single-line returns, I prefer single-line asserts. So when the "expected" is too long to inline into the assert, move the expected to a defn on its own line, then assert.Equal(t, toRenewExpected, toRenew)

lte/cloud/go/services/subscriberdb/protos/subscriberdb.proto
```diff
@@ -126,3 +127,10 @@ message ApnResourceInternal {
   // apn_resource is the original apn resource proto message.
   APNConfiguration.APNResource apn_resource = 3;
 }
+
+// SubscriberDigestWithIDs is used internally in per-subscriber store to serialize and deserialize
```
hcgatewood: "used internally" is implicit based on this proto's location in the filesystem

lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -52,9 +54,88 @@ func GetDigest(networkID string) (string, error) {
...
+	}
+
+	return perSubDigests, nil
+}
+
+// GetPerSubscriberDigestsDiff computes the changeset between two lists of per-subscriber digests,
+// ordered by their subscriber IDs (unique within a network). It returns
+// 1. A set of subscribers that have been added/modified, with the new digests.
+// 2. An ordered list of subscribers that have been removed.
+func GetPerSubscriberDigestsDiff(all []*lte_protos.SubscriberDigestWithID, tracked []*lte_protos.SubscriberDigestWithID) (map[string]string, []string) {
```
hcgatewood: From previous comment: "all" and "tracked" are confusing to me on first read, consider "prev" and "next" or "tracked" and "requested" or something like that

lte/cloud/go/services/subscriberdb/storage/per_sub_digest.go
```diff
@@ -0,0 +1,121 @@
...
+	"magma/orc8r/cloud/go/mproto"
+	"magma/orc8r/cloud/go/storage"
+	merrors "magma/orc8r/lib/go/errors"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+)
+
+const (
+	PerSubDigestBlobstoreType = "per_sub_digest"
```
hcgatewood: For both of these: add golang-style docstrings. Also no need to add the "NOTE:" since it's just the docstring

lte/cloud/go/services/subscriberdb/storage/per_sub_digest.go
```diff
@@ -0,0 +1,121 @@
...
+			Type: PerSubDigestBlobstoreType,
+			Key:  network,
+		}})
+		if err != nil {
+			multierror.Append(errs, err)
+		}
+	}
+
+	if errs.ErrorOrNil() != nil {
+		return errors.Wrapf(errs, "Delete per-sub digests from blobstore")
```
hcgatewood: Errors aren't sentence-cased unless you're directly logging the err

lte/cloud/go/services/subscriberdb/storage/per_sub_digest_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"magma/lte/cloud/go/services/subscriberdb"
+	"magma/lte/cloud/go/services/subscriberdb/storage"
+	"magma/orc8r/cloud/go/blobstore"
+	"magma/orc8r/cloud/go/sqorc"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPerSubDigestStore(t *testing.T) {
+	db, err := sqorc.Open("sqlite3", ":memory:")
```
hcgatewood: Replace these lines with https://github.com/magma/magma/blob/51843e3245e0b785a738d991f62657c2cac328b6/orc8r/cloud/go/test_utils/test_db.go#L50-L52

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -27,14 +28,15 @@ import (
...
-		generatedDigestsByNetworks, err := RenewDigests(store, config)
+		flatDigestsByNetworks, perSubDigestsByNetworks, err := RenewDigests(config, digestStore, perSubDigestStore)
 		if err != nil {
 			glog.Errorf("Error monitoring digests: %+v", err)
 		}
-		if len(generatedDigestsByNetworks) > 0 {
-			glog.Infof("Generated digests per network: %+v", generatedDigestsByNetworks)
+		if len(flatDigestsByNetworks) > 0 {
+			glog.Infof("Generated digests per network: %+v", flatDigestsByNetworks)
+			glog.Infof("Generated per-sub digests per network: %+v", perSubDigestsByNetworks)
```
hcgatewood: Make this one a V2 log, since it may be O(20k+)

lte/cloud/go/services/subscriberdb/storage/per_sub_digest.go
```diff
@@ -0,0 +1,121 @@
...
+	store, err := l.fact.StartTransaction(&storage.TxOptions{ReadOnly: true})
+	if err != nil {
+		return errors.Wrapf(err, "error starting transaction")
+	}
+	defer store.Rollback()
+
+	// The sorted list of per-sub digests are serialized to be stored as a single blob.
+	// This is to preserve the sorted order of the digests, and to decrease the frequency
+	// of writes to the store.
+	blobValue := &protos.SubscriberDigestWithIDs{Digests: digests}
```
hcgatewood: Inline this defn into the next line, if possible

lte/cloud/go/services/subscriberdb/storage/per_sub_digest.go
```diff
@@ -0,0 +1,121 @@
...
+	blobValueSerialized, err := mproto.MarshalDeterministic(blobValue)
```
hcgatewood: No need to use deterministic serialization here. Instead, use https://github.com/magma/magma/blob/509e5b4bcdb68b0021044fb1dc8cf47554446b8f/orc8r/lib/go/protos/marshaler.go#L27

Then you can remove the mproto.Unmarshal fn you added, along with the accompanying tests

hcgatewood (changes requested):
Couple more small things, looking good

lte/cloud/go/services/subscriberdb/digest.go
wangyyt1013: The reason for the current implementation is simply that if a subscriber is to be deleted, there's no use for sending down that subscriber data to the agw as well, and so the subscriber data portion is omitted for `deleted` @hcgatewood

lte/cloud/go/services/subscriberdb/digest.go
wangyyt1013: I think that makes sense! In this case, if per-sub digest generation fails, the `last_updated_at` for this network would not update (and will stay at an "outdated" status), so in the next loop a resync will be enforced. @hcgatewood

hcgatewood: Yes. We just need to be careful in the "SyncSubscribers" rpc to ensure we check the "last updated time" -- if it's too old, SyncSubscribers should *also* return "resync"

lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -101,33 +101,33 @@ func GetPerSubscriberDigests(network string) ([]*lte_protos.SubscriberDigestWith
 // ordered by their subscriber IDs (unique within a network). It returns
 // 1. A set of subscribers that have been added/modified, with the new digests.
 // 2. An ordered list of subscribers that have been removed.
-func GetPerSubscriberDigestsDiff(all []*lte_protos.SubscriberDigestWithID, tracked []*lte_protos.SubscriberDigestWithID) (map[string]string, []string) {
-	n, m, i, j := len(all), len(tracked), 0, 0
+func GetPerSubscriberDigestsDiff(next []*lte_protos.SubscriberDigestWithID, prev []*lte_protos.SubscriberDigestWithID) (map[string]string, []string) {
```
hcgatewood: In the args and throughout the fn -- be consistent about "prev" first, then "next". It reads better that way

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -267,7 +267,7 @@ func TestGetPerSubscriberDigests(t *testing.T) {
 	// Initially the per-sub digest list should be empty (but not nil)
 	perSubDigests, err := subscriberdb.GetPerSubscriberDigests("n1")
 	assert.NoError(t, err)
-	assert.Empty(t, []*lte_protos.SubscriberDigestWithID{}, perSubDigests)
+	assert.Equal(t, []*lte_protos.SubscriberDigestWithID{}, perSubDigests)
```
hcgatewood, line 271: This is how I meant for all of these
```suggestion
	assert.Empty(t, perSubDigests)
```

hcgatewood (approved):
Lgtm! Fix two small nits then msg me to merge

lte/cloud/go/services/subscriberdb/digest_test.go
wangyyt1013: That makes sense, although in this specific instance I wanted to make sure that the `perSubDigests` is an empty slice instead of a `nil` slice, whereas `assert.Empty` will return `true` for either empty or nil slices. Therefore, I chose to use `assert.Equal` to make it clear that the slice needs to be empty, not `nil`. Wondering what's your thoughts on this? @hcgatewood

hcgatewood: I see. Okay so generally, the difference between a nil slice and an empty slice should be as nonexistent as possible [0] -- why do you want to ensure they're different here?

[0] Main reason, in my experience, where they differ is related to serialization. Is that what you're checking here, or ?

wangyyt1013: The reason is that the return value of `GetPerSubscriberDigests`, i.e. the cloud-side per-subscriber digests, will be sent back to the agw in a `SyncSubscribersResponse`, and currently the agw is set to iterate over the list to update the agw per-sub digests (if the rpc call was successful). In python, iterating over `None` will result in an error, and that's why I wanted to enforce that the cloud side should always send back a non-nil array; should I stick to this pattern or try to add extra parsing steps on the agw side? @hcgatewood

hcgatewood: No that's a good reason. Just add a comment in the code at a relevant point to make it clear that's what you're doing

lte/cloud/go/services/subscriberdb/digest.go
uri200: Sounds good. But, If possible, I would use different name instead of `Md5Base64Digest`

hcgatewood: Ah yes true I understand the concern. The idea is that, once we complete the APN refactor, this can indeed be "actually" base64-encoded bytes, so I don't think it's worth changing in the interim 🙂

themarwhal (approved):

PR #7815, opened 2021-06-28
refactor(sessiond): move events and update report callback into LocalEnforcer from LocalSessionManagerHandler

veshkemburu:

## Summary

After receiving stats from pipelined, sessiond performs actions on them such as reporting and record aggregation

## Test Plan

Transfer internal function callback to separate function. 
Move each LocalSessionManagerHandler function to LocalEnforcer
Add calls from LocalSessionManagerHandler to enforcer using local enforcer object

## Additional Information


themarwhal (changes requested):
Please add a summary to indicate why this change is necessary. Thanks :)

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -147,6 +148,123 @@ void LocalEnforcer::setup(
...
+        }
+        // Success!
+        update_session_credits_and_rules(
+            *session_map_ptr, response, session_uc);
+        report_session_update_event(*session_map_ptr, updates_by_session);
+        session_store_.update_sessions(session_uc);
+      });
+}
+
+void LocalEnforcer::HandlePipelinedResponse(
```
themarwhal: `HandlePipelinedResponse` -> `handle_pipelined_response`

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -577,6 +577,8 @@ def get_stats(self, cookie: int = 0, cookie_mask: int = 0):
         Use Ryu API to send a stats request containing cookie and cookie mask, retrieve a response and 
         convert to a Rule Record Table
         """
+        if not self._datapath:
```
themarwhal: Should this be a separate PR?

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -147,6 +148,123 @@ void LocalEnforcer::setup(
...
+    MLOG(MDEBUG) << "Aggregating " << resp.records_size() << " records";
+    aggregate_records(session_map, resp, update);
+
+    check_usage_for_reporting(std::move(session_map), update);
+  }
+}
+
+void LocalEnforcer::poll_stats_enforcer() {
+  pipelined_client_->poll_stats(
+      0, 0, std::bind(&LocalEnforcer::HandlePipelinedResponse, this, _1, _2));
```
themarwhal: Can you add a comment on why we need to bind the arguments here?

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -147,6 +148,123 @@ void LocalEnforcer::setup(
...
+  session_store_.set_and_save_reporting_flag(true, request, session_uc);
+
+  // Before reporting and returning control to the event loop, increment the
+  // request numbers stored for the sessions in SessionStore
+  session_store_.sync_request_numbers(session_uc);
+
+  // report to cloud
+  reporter_->report_updates(
+      request,
+      [this, request, session_uc,
```
themarwhal: If this callback function is shared between here and LocalsessionManagerHandler, pull the callback function into a single function defined in `SessionReporter`. 
Always try to minimize the amount of code you have to copy :)

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
veshkemburu: It's a minor adjustment to handle an edge case, I can make it a separate PR if desired but I thought it would be easier to just make the fix and put it in with this PR.

themarwhal: Let's separate it out :) 
Each PR should handle independent changes
Also, I think if we separate it out, we can get it into master faster as it only requires @koolzz 's approval.

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -147,13 +148,127 @@ void LocalEnforcer::setup(
   }
 }
 
-void LocalEnforcer::poll_stats_enforcer() {
-  pipelined_client_->poll_stats(0, 0, [](Status status, RuleRecordTable resp) {
-    if (!status.ok()) {
-      MLOG(MERROR) << "Could not successfully poll stats: "
-                   << status.error_message();
+void LocalEnforcer::report_session_update_event(
```
themarwhal, line 151: Should the corresponding function in LocalSessionManagerHandler removed in this PR?

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -147,13 +148,127 @@ void LocalEnforcer::setup(
...
+                     << " because it couldn't be found";
+      continue;
     }
-  });
+    events_reporter_->session_updated(
+        session_id, (**session_it)->get_config(), it.second);
+  }
+}
+
+void LocalEnforcer::report_session_update_event_failure(
```
themarwhal: Same deal here

themarwhal: Also, while you're at it can you rename this function to `report_session_update_failure_event`?

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -147,13 +148,127 @@ void LocalEnforcer::setup(
...
+    MLOG(MERROR) << "Could not successfully poll stats: "
+                 << status.error_message();
+  } else {
+    auto session_map = session_store_.read_all_sessions();
+    SessionUpdate update =
+        SessionStore::get_default_session_update(session_map);
+    MLOG(MDEBUG) << "Aggregating " << resp.records_size() << " records";
+    aggregate_records(session_map, resp, update);
+
+    check_usage_for_reporting(std::move(session_map), update);
```
themarwhal: Instead of `std::move` ing the entire map, you can pass by reference here.

```
void LocalEnforcer::check_usage_for_reporting(
     SessionMap& session_map, SessionUpdate& updates)
```

lte/gateway/c/session_manager/LocalEnforcer.h
```diff
@@ -294,6 +294,27 @@ class LocalEnforcer {
...
+
+  /*
+   * Report flow stats from pipelined and track the usage per rule
+   */
+  void check_usage_for_reporting(
+      SessionMap session_map, SessionUpdate& session_uc);
+
+  void handle_pipelined_response(Status status, RuleRecordTable resp);
+
+  void report_updates_callback(
```
themarwhal: `report_updates_callback` -> `handle_session_update_response`

lte/gateway/c/session_manager/LocalEnforcer.h
```diff
@@ -294,6 +294,27 @@ class LocalEnforcer {
...
+      UpdateSessionRequest& request,
```
themarwhal: Should this be const as the linter suggests?

themarwhal (changes requested):

themarwhal:
please add a test plan + fix PR title (`refactor(sessiond): move events and update report callback into LocalEnforcer`) .

lte/gateway/c/session_manager/LocalEnforcer.cpp
veshkemburu: Tried this before but it errors out.

themarwhal (approved):

PR #7816, opened 2021-06-28
Fixed Subscriber add dialog to show saved errors

ato-dani:
Signed-off-by: Dani Menewuyelet <dmenewuy@fb.com>

## Summary
The add dialogue bulkily adds 1000 subscribers at a time. If it gets any error in the middle of reading the 1000 subscribers, it saves them and finally shows them when it reaches the 1000th subscriber or if it reaches the end of the subscribers list(if we were adding less than 1000 subscriber). However, the previous implementation missed the condition where the last subscriber to be added resulted in an error. For that case, it would append the error but since there are no more subscribers to loop at, it returns without showing the saved errors. So, the changes in this PR makes sure that if the last subscriber resulted in an error, all the saved errors should be shown since we wouldn't get another chance to report those errors. 

## Additional Information


andreilee (approved):

karthiksubraveti:
Closes #7817

PR #7835, opened 2021-06-29
Add edge case to Get Stats Function in EnforcementStats

veshkemburu:

## Summary

Sometimes the GetStats API is called before the datapath is populated. In this case we just send an empty RuleRecordsTable. 

## Additional Information


lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -577,6 +577,8 @@ def get_stats(self, cookie: int = 0, cookie_mask: int = 0):
         Use Ryu API to send a stats request containing cookie and cookie mask, retrieve a response and 
         convert to a Rule Record Table
         """
+        if not self._datapath:
```
koolzz, line 580: maybe add error log, saying` datapath is not initialized for stats retrieval `

koolzz (approved):

PR #7841, opened 2021-06-29
feat(agw): Add commands and logs in show-tech playbook

wallyrb:

## Summary

- "command" in ansible doesn't take pipes. Changing to "dpkg -l magma*" in order to get sctpd and other magma package related versions
- Add "sudo" in the documentation
- Add ovs commands to be collected
- Check service status from ovs and scptd
- Collect all /var/log folder from AGW
- Since we are collecting syslogs, remove the journalctl -u magma@* command

## Test Plan

Tested in 1.5 Ubuntu AGW

### DIFF
--- a/show-tech/constants.yml
+++ b/show-tech/constants.yml
@@ -12,10 +12,15 @@ general_commands:
   - "top -b -n 1"
   - "df -kh"
 debian_commands:
-  - "dpkg -l | grep magma"
+  - "dpkg -l magma*"
+  - "dpkg -l *openvswitch*"
   - "ovs-vsctl show"
+  - "ovs-ofctl show gtp_br0"
+  - "ovs-ofctl dump-flows gtp_br0"
   - "apt show magma"
   - "service magma@* status"
+  - "service sctpd status"
+  - "service openvswitch-switch status"
   - "show_gateway_info.py"
   - "checkin_cli.py"
   - "mobility_cli.py get_subscriber_table"
@@ -23,7 +28,6 @@ debian_commands:
   - "enodebd_cli.py get_all_status"
   - "ip addr"
   - "ping google.com -I eth0 -c 5"
-  - "journalctl -u magma@*"
   - "timeout 60 sudo tcpdump -i any sctp -w {{report_dir_output.stdout}}/sctp.pcap"
   - "timeout 60 sudo tcpdump -i any port 48080 -w {{report_dir_output.stdout}}/any-48080.pcap"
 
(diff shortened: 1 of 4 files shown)


show-tech/constants.yml
```diff
@@ -34,10 +37,8 @@ paths_to_collect:
   - "/etc/magma/configs/*"
   - "/etc/systemd/system/*"
   - "/usr/local/bin/mme"
-  - "/var/log/syslog"
   - "/var/core/*"
-  - "/var/log/MME.*"
-  - "/var/log/enode.*"
+  - "/var/log/*"
```
pshelar, line 42: can you also include '/var/log/openvswitch/'

karthiksubraveti (changes requested):
We should be able to set privilege escalation to individual tasks instead of running ansible-playbook in sudo mode. Can you add sudo to tasks which require this alone. https://docs.ansible.com/ansible/2.3/become.html

show-tech/constants.yml
```diff
@@ -12,18 +12,21 @@ general_commands:
   - "top -b -n 1"
   - "df -kh"
 debian_commands:
-  - "dpkg -l | grep magma"
+  - "dpkg -l magma*"
```
pshelar, line 15: lets also get ovs version: `dpkg -l *openvswitch*`

wallyrb:
> We should be able to set privilege escalation to individual tasks instead of running ansible-playbook in sudo mode. Can you add sudo to tasks which require this alone. https://docs.ansible.com/ansible/2.3/become.html

@karthiksubraveti I tried using become in the tasks that required it. However, for the module "copy" and any other module with "src", become will apply only for writing the file not for reading. Therefore, if the local user doesn't have permissions to read "src"  it will fail with "an error occurred while trying to read the file Permission denied"

https://superuser.com/questions/1522927/ansible-copy-module-become-permission-denied

We could either use "sudo' when run the playbook or try with another module that is not copy. (maybe command)

themarwhal (approved):
doc changes lgtm

electronjoe:
I was waiting for clear-up of the questions posted by @pshelar and @karthiksubraveti  - before giving docs LGTM from owner-side. I'll give stamp now but not clear to me if all parties are yet satisfied above?

electronjoe (approved):
Thank you for updating docs Wally!

pshelar (approved):

karthiksubraveti:
> become

Sounds good. Let's go with this and change it when we have a better approach

karthiksubraveti (approved):

wallyrb:
hey @electronjoe ! could you help me check if something is pending to merge this? Will appreciate any suggestions to complete this. Thanks!

PR #7848, opened 2021-06-30
Reconcile protobuf state with actual state as maintained by MME

amarpad:
## Summary
Currently num_enbs counter may by out of sync with the actual associations
Reconcile this by clobbering the state as maintained by the protobuf

## Test Plan

make integ_test
restarted sctpd connection verified without the change the num_enbs is out of sync, after change it is in_sync

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/s1ap/s1ap_state_converter.cpp
+++ b/lte/gateway/c/core/oai/tasks/s1ap/s1ap_state_converter.cpp
@@ -60,6 +60,14 @@ void S1apStateConverter::state_to_proto(s1ap_state_t* state, S1apState* proto) {
     FREE_HASHTABLE_KEY_ARRAY(keys);
   }
 
+  keys                        = hashtable_ts_get_keys(&state->enbs);
+  uint32_t expected_enb_count = keys ? keys->num_keys : 0;
+  if (expected_enb_count != state->num_enbs) {
+    OAILOG_ERROR(
+        LOG_S1AP, "Updating num_eNBs from maintained to actual count %u->%u",
+        state->num_enbs, expected_enb_count);
+    state->num_enbs = expected_enb_count;
+  }
   proto->set_num_enbs(state->num_enbs);
 }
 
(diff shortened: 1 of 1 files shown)


lte/gateway/c/core/oai/tasks/s1ap/s1ap_state_converter.cpp
```diff
@@ -60,6 +60,14 @@ void S1apStateConverter::state_to_proto(s1ap_state_t* state, S1apState* proto) {
     FREE_HASHTABLE_KEY_ARRAY(keys);
   }
 
+  keys                        = hashtable_ts_get_keys(&state->enbs);
+  uint32_t expected_enb_count = keys ? keys->num_keys : 0;
+  if (expected_enb_count != state->num_enbs) {
+    OAILOG_ERROR(
+        LOG_S1AP, "Updating num_eNBs from maintained to actual count %u->%u",
+        state->num_enbs, expected_enb_count);
+    state->num_enbs = expected_enb_count;
```
amarpad, line 69: We could also remove this state from the persistence layer and then just reconstruct it, but this helps us make sure that our stats counting is sane as well.

lte/gateway/c/core/oai/tasks/s1ap/s1ap_state_converter.cpp
```diff
@@ -60,6 +60,14 @@ void S1apStateConverter::state_to_proto(s1ap_state_t* state, S1apState* proto) {
     FREE_HASHTABLE_KEY_ARRAY(keys);
   }
 
+  keys                        = hashtable_ts_get_keys(&state->enbs);
+  uint32_t expected_enb_count = keys ? keys->num_keys : 0;
+  if (expected_enb_count != state->num_enbs) {
+    OAILOG_ERROR(
```
ardzoht, line 66: Shouldn't this be WARNING level as well? Or do we want to explicitly call out this case as an error?

ardzoht (approved):
lgtm, left one minor comment regarding logging level

lte/gateway/c/core/oai/tasks/s1ap/s1ap_state_converter.cpp
amarpad: Its an error if we are trying to write state that is inconsistent i.e. it shouldn't happen, it's a warning if the state we read is inconsistent as it is reconciling external state, is how I have been interpreting it, I am ok switching to warning if you prefer.

ardzoht: Makes sense to me, let's keep them as separate, thanks

PR #7863, opened 2021-06-30
remove unused apn.proto import from session_manager.proto

themarwhal:

## Summary
apn.proto is not used, and I keep seeing warnings about it during compilation

## Test Plan
CI

## Additional Information


amarpad:
Would you mind eyeballing the cmake files to make sure we are not building apn because we ended up building session_manager? Thanks!

amarpad (approved):

ardzoht (approved):

pshelar (approved):

themarwhal:
> Would you mind eyeballing the cmake files to make sure we are not building apn because we ended up building session_manager? Thanks!

Yeah that is a good thing to double check! Thanks :)

karthiksubraveti:
> Would you mind eyeballing the cmake files to make sure we are not building apn because we ended up building session_manager? Thanks!

I moved some structs from subscriberdb proto to apn proto. We do need to build apn proto

karthiksubraveti:
> > Would you mind eyeballing the cmake files to make sure we are not building apn because we ended up building session_manager? Thanks!
> 
> Yeah that is a good thing to double check! Thanks :)

I had originally added it because of a few errors which cropped up during lte tests. I forgot the original error, it might be worth to look out for that error

themarwhal:
> > > Would you mind eyeballing the cmake files to make sure we are not building apn because we ended up building session_manager? Thanks!
> > 
> > 
> > Yeah that is a good thing to double check! Thanks :)
> 
> I had originally added it because of a few errors which cropped up during lte tests. I forgot the original error, it might be worth to look out for that error

Everything builds and tests fine without the apn.proto so I think we're OK. In proto you only need to explicitly include another proto if it directly relies on it.

karthiksubraveti:
> > > > Would you mind eyeballing the cmake files to make sure we are not building apn because we ended up building session_manager? Thanks!
> > > 
> > > 
> > > Yeah that is a good thing to double check! Thanks :)
> > 
> > 
> > I had originally added it because of a few errors which cropped up during lte tests. I forgot the original error, it might be worth to look out for that error
> 
> Everything builds and tests fine without the apn.proto so I think we're OK. In proto you only need to explicitly include another proto if it directly relies on it.

Yes. I am aware of that and I made changes in subscriberdb proto to include apn proto. From what I remember, there were also structs in apn proto which was being used by these proto files, so added this import. If it works fine, then that's great.

PR #7885, opened 2021-07-01
feat(agw): enable GTP-U echo response.

pshelar:
AGW can make use of OVS echo response if available.

## Test Plan
config changes for the gtp port validated on vagrant.

## Additional Information


pshelar:
This depends on #7888

uri200 (approved):

PR #7890, opened 2021-07-01
[AGW]Make sure we don't install 1.6.0 on debian

tmdzk:
[AGW]Make sure we don't install 1.6.0 on debian

## Summary

[AGW]Make sure we don't install 1.6.0 on debian

### DIFF
--- a/lte/gateway/release/upgrade_magma.sh
+++ b/lte/gateway/release/upgrade_magma.sh
@@ -22,6 +22,10 @@ done
 
 if grep -q 'Debian' /etc/issue; then
   OS_VERSION="stretch"
+  # We don't support 1.6.0 for debian so bumping down to 1.5.2
+  if [ "$MAGMA_VERSION" == "1.6.0" ]; then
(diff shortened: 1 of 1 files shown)


ardzoht (approved):

lte/gateway/release/upgrade_magma.sh
```diff
@@ -22,6 +22,10 @@ done
 
 if grep -q 'Debian' /etc/issue; then
   OS_VERSION="stretch"
+  # We don't support 1.6.0 for debian so bumping down to 1.5.2
+  if [ "$MAGMA_VERSION" == "1.6.0" ]; then
+    MAGMA_VERSION="1.5.2"
+  fi
```
pshelar, line 28: it would better to compare all newer release: 'if [[ "$MAGMA_VERSION" > "1.6.0" || "$MAGMA_VERSION" == "1.6.0"  ]]; then'

it should handle most of versions.

PR #7894, opened 2021-07-01
Variable Poll Stats Interval

veshkemburu:

## Summary

Add config value for polling stats every variable amount of seconds.

## Test Plan

Added config flag to multiple yaml files.
Ran sessiond manager unit tests on Magma VM
Removed flag and altered interval values and used journalctf to check polling times

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/sessiond_main.cpp
+++ b/lte/gateway/c/session_manager/sessiond_main.cpp
@@ -338,13 +338,12 @@ int main(int argc, char* argv[]) {
   std::thread periodic_stats_requester_thread([&]() {
     // random value assigned for interval period, the value will be loaded
     // from a config field later
+    uint32_t interval = DEFAULT_POLL_INTERVAL_TIME;
     if (config["poll_stats_interval"].IsDefined()) {
-      periodic_stats_requester->start_loop(
-          local_enforcer, config["poll_stats_interval"].as<uint32_t>());
-    } else {
-      periodic_stats_requester->start_loop(
-          local_enforcer, DEFAULT_POLL_INTERVAL_TIME);
+      interval = config["poll_stats_interval"].as<uint32_t>();
     }
+    periodic_stats_requester->start_loop(
+          local_enforcer, config["poll_stats_interval"].as<uint32_t>());
     periodic_stats_requester->stop();
   });
 
--- a/lte/gateway/c/session_manager/sessiond_main.cpp
+++ b/lte/gateway/c/session_manager/sessiond_main.cpp
@@ -335,9 +335,10 @@ int main(int argc, char* argv[]) {
   // every fixed interval of time
   auto periodic_stats_requester = std::make_shared<magma::StatsPoller>();
   std::thread periodic_stats_requester_thread([&]() {
-    //random value assigned for interval period, the value will be loaded
-    //from a config field later
-    periodic_stats_requester->start_loop(local_enforcer, 5);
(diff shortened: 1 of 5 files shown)


koolzz (approved):
Lgtm but maybe have a default val

lte/gateway/configs/sessiond.yml
```diff
@@ -72,3 +72,7 @@ enable_ipfix: false
 # Session manager max resending throttle count when the UPF session info version is
 # not matching with UPF received version no.
 session_rtx_count: 3
+
+# set to a certain interval measured in seconds for which sessiond should poll pipelined
+# for relevant stats
+poll_stats_interval: 5
```
uri200, line 78: look for all `sessiond.yml` in our codebase and add this parameter. Otherwise `config["poll_stats_interval"].as<uint32_t>()` will crash

Normally cmd+shift+f or cmd+shif+o will open the file search in your vscode (I use intellij)

veshkemburu: Makes sense. I believe there are only 2 sessiond.yml files, and I've added the field to both.

veshkemburu: Ah never mind I see more.

lte/gateway/c/session_manager/sessiond_main.cpp
```diff
@@ -335,9 +336,15 @@ int main(int argc, char* argv[]) {
...
-    periodic_stats_requester->start_loop(local_enforcer, 5);
+    // random value assigned for interval period, the value will be loaded
+    // from a config field later
+    if (config["poll_stats_interval"].IsDefined()) {
+      periodic_stats_requester->start_loop(
+          local_enforcer, config["poll_stats_interval"].as<uint32_t>());
+    } else {
+      periodic_stats_requester->start_loop(
+          local_enforcer, DEFAULT_POLL_INTERVAL_TIME);
+    }
```
themarwhal: something like this would be cleaner to read
```suggestion
	uint32_t interval = DEFAULT_POLL_INTERVAL_TIME;
    if (config["poll_stats_interval"].IsDefined()) {
      interval = config["poll_stats_interval"].as<uint32_t>();
    }
    periodic_stats_requester->start_loop(local_enforcer, interval);
```

themarwhal (changes requested):
Can you fill in the test plan section?
Left a style related comment, but change itself looks good!

themarwhal:
lmk if you need help resolving the semantic pull request check

lte/gateway/c/session_manager/sessiond_main.cpp
```diff
@@ -335,9 +336,14 @@ int main(int argc, char* argv[]) {
...
-    //from a config field later
-    periodic_stats_requester->start_loop(local_enforcer, 5);
+    // random value assigned for interval period, the value will be loaded
+    // from a config field later
+    uint32_t interval = DEFAULT_POLL_INTERVAL_TIME;
+    if (config["poll_stats_interval"].IsDefined()) {
+      interval = config["poll_stats_interval"].as<uint32_t>();
+    }
+    periodic_stats_requester->start_loop(
+          local_enforcer, config["poll_stats_interval"].as<uint32_t>());
```
themarwhal: do you mean
```suggestion
    periodic_stats_requester->start_loop(local_enforcer, interval);
```
?

themarwhal (approved):

PR #7905, opened 2021-07-02
refactor: [ci] dco pr comment job

quentinDERORY:
## Summary

Current DCO python library does not support pull_request_target GHA event (granting write permission on PR)
So I had to create a job listening on the DCO check to comment the PR
Second issue is GHA does not propagate PR number so I had to save it as artifact

## Test Plan

Tested in my repo https://github.com/quentinDERORY/magma/pull/13

## Additional Information

### DIFF
--- a/.github/workflows/dco-check.yml
+++ b/.github/workflows/dco-check.yml
@@ -1,3 +1,4 @@
+---
 name: DCO check
 on:  # yamllint disable-line rule:truthy
   pull_request:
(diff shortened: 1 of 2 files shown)


.github/workflows/dco-check.yml
```diff
@@ -8,32 +9,23 @@ jobs:
...
-          - [Howto: sign-off most-recent commit](https://stackoverflow.com/questions/13043357/git-sign-off-previous-commits)
-          - [Howto: sign-off multiple past commits](https://gist.github.com/kwk/d70f20d17b18c4f3296d)`
-
-          github.issues.createComment({
-            issue_number: context.issue.number,
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            body: msg,
-          })
+      - name: Get PR Commits
```
hcgatewood: Nit: can you retain the previous indentation to avoid superfluous diff lines

.github/workflows/dco-check.yml
```diff
@@ -8,32 +9,23 @@ jobs:
...
+        id: 'get-pr-commits'
+        uses: tim-actions/get-pr-commits@master
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: DCO Check
+        uses: tim-actions/dco@master
+        with:
+          commits: ${{ steps.get-pr-commits.outputs.commits }}
+      # Need to save PR number as github action does not propagate it with workflow_run event
```
hcgatewood:
```suggestion
      # Need to save PR number as GitHub action does not propagate it with workflow_run event
```

hcgatewood (approved):
Does this resolve the issue of the dco-comment-pr being able to make comments on the PR?

quentinDERORY:
> Does this resolve the issue of the dco-comment-pr being able to make comments on the PR?

Yes It's solving the issue :)

themarwhal (approved):
This is awesome! Instead of having a dedicated comment job for the DCO check, I think we want a job that listens on any required check failure just so that we don't have a different comment job for ~20 checks. (We can do this once we merge this and validate it works, ofc)

The comment should link to a comprehensive doc on required checks -> POC + resolution steps. (something like https://docs.magmacore.org/docs/next/contributing/contribute_ci_checks but it needs improvement). 
We also need to make it so that a job can't be made required without editing this doc, etc. 

Anyhow, this would be great to have <3

quentinDERORY:
> This is awesome! Instead of having a dedicated comment job for the DCO check, I think we want a job that listens on any required check failure just so that we don't have a different comment job for ~20 checks. (We can do this once we merge this and validate it works, ofc)
> 
> The comment should link to a comprehensive doc on required checks -> POC + resolution steps. (something like https://docs.magmacore.org/docs/next/contributing/contribute_ci_checks but it needs improvement).
> We also need to make it so that a job can't be made required without editing this doc, etc.
> 
> Anyhow, this would be great to have <3

Yes I think going forward, we will just modify this job, add additional trigger and switch the message content in function of the trigger.
The job should run in parallel and comment for each failing workflow, we could always add this doc to the bottom of each comment.

hcgatewood:
> > This is awesome! Instead of having a dedicated comment job for the DCO check, I think we want a job that listens on any required check failure just so that we don't have a different comment job for ~20 checks. (We can do this once we merge this and validate it works, ofc)
> > The comment should link to a comprehensive doc on required checks -> POC + resolution steps. (something like [docs.magmacore.org/docs/next/contributing/contribute_ci_checks](https://docs.magmacore.org/docs/next/contributing/contribute_ci_checks) but it needs improvement).
> > We also need to make it so that a job can't be made required without editing this doc, etc.
> > Anyhow, this would be great to have <3
> 
> Yes I think going forward, we will just modify this job, add additional trigger and switch the message content in function of the trigger.
> The job should run in parallel and comment for each failing workflow, we could always add this doc to the bottom of each comment.

Fantastic

Issue #7869, opened 2021-06-30
[lte][agw] enodebd fails to configure device due to Invalid Parameters


ardzoht:
Prior to submitting an issue, check to see if one has already been created. If there is currently an open issue, add a thumbs-up emoji to identify that it is also affecting you.

### Your Environment

- **Version:** 1.6.0-1625058082-58fc8527
- **Affected Component:** Access Gateway
- **Affected Subcomponent:** enodebd
- **Deployment Environment:** e.g. bare-metal (AGW)

### Describe the Issue

On AGW we are connecting a managed eNodeb device, with following configuration:

```
            "enbConfigsBySerial": {
                "120200018118BVJ0013": {
                    "earfcndl": 9430,
                    "subframeAssignment": 2,
                    "specialSubframePattern": 7,
                    "pci": 60,
                    "transmitEnabled": true,
                    "deviceClass": "Baicells Nova-233 G2 OD FDD",
                    "bandwidthMhz": 10,
                    "tac": 1,
                    "cellId": 171160
                }
            }
```

On enodebd logs we can see that on the response to SetParameters message, there is a Fault response by the Tr069 server, which impedes the configuration of the device:

```
2021-06-30 19:56:02,462 DEBUG 10.0.2.243 - "POST / HTTP/1.1" 200 0
2021-06-30 19:56:03,211 WARNING Parameter <PTP status> not defined in model
2021-06-30 19:56:03,214 WARNING Parameter <PTP status> not defined in model
2021-06-30 19:56:06,631 WARNING Parameter <PTP status> not defined in model
"/var/log/enodebd.log" 5631L, 1085578C                                                                                                                                                                      1,1           Top
2021-06-30 20:19:14,905 DEBUG State transition to <wait_set_params>
[...]

PR #7906, opened 2021-07-04
fix(enodebd): Add workaround logic for stale msg name in dictionary

talkhasib:
## Summary
https://github.com/magma/magma/issues/7869

"SetParameterValues" is going out as "GetParameterValues" message type. Although, the message body is correctly sent.
This causes the eNB to reject the message and send back a fault.

Issue traced down to the following
We set message name "tx.descriptor.out_message.Attributes.sub_name" to the correct message name. But, when the parent "serialize" function is called, it pulls a stale value from the dictionary "_attrcache"

Workaround is to clear the element in the dictionary to force "_attrcache" to repopulate using the provided descriptor.

## Test Plan

Tested on local home setup with BaiCells eNB

### DIFF
--- a/lte/gateway/python/magma/enodebd/tr069/spyne_mods.py
+++ b/lte/gateway/python/magma/enodebd/tr069/spyne_mods.py
@@ -140,6 +140,15 @@ class Tr069Soap11(Soap11):
         return super(Tr069Soap11, self).get_call_handles(ctx)
 
     def serialize(self, ctx, message):
+        # Workaround for issue https://github.com/magma/magma/issues/7869
+        # Updates to ctx.descriptor.out_message.Attributes.sub_name are taking
+        # effect on the descriptor. But when puled from _attrcache dictionary,
+        # it still has a stale value.
+        # Force repopulation of dictionary by deleting entry
+        # TODO Remove this code once we have a better fix
+        if (ctx.descriptor.out_message in self._attrcache):
+            del self._attrcache[ctx.descriptor.out_message]  # noqa: WPS529
+
         super(Tr069Soap11, self).serialize(ctx, message)
 
         # Keep XSD namespace
--- a/lte/gateway/python/magma/enodebd/data_models/transform_for_magma.py
+++ b/lte/gateway/python/magma/enodebd/data_models/transform_for_magma.py
@@ -54,7 +54,7 @@ def gps_tr181(value: str) -> str:
     """
     try:
         return str(float(value) / 1e6)
-    except ValueError:
+    except Exception:  # pylint: disable=broad-except
         return value
 
 


lte/gateway/python/magma/enodebd/tr069/spyne_mods.py
```diff
@@ -140,6 +140,15 @@ def get_call_handles(self, ctx):
         return super(Tr069Soap11, self).get_call_handles(ctx)
 
     def serialize(self, ctx, message):
+        # Workaround for issue https://github.com/magma/magma/issues/7869
+        # Updates to ctx.descriptor.out_message.Attributes.sub_name are taking
+        # effect on the descriptor. But when puled from _attrcache dictionary,
+        # it still has a stale value.
+        # Force repopulation of dictionary by deleting entry
+        # TODO Remove this code once we have a better fix
+        if (self._attrcache.__contains__(ctx.descriptor.out_message)):
```
amarpad: `# noqa: WPS731?`

lte/gateway/python/magma/enodebd/tr069/spyne_mods.py
```diff
@@ -140,6 +140,15 @@ def get_call_handles(self, ctx):
...
+            self._attrcache.__delitem__(ctx.descriptor.out_message)
+
```
amarpad, line 151: There is a Baicells test where we simulate inputs if you want to add a test there, you can get rid of this. But optional.

amarpad (approved):

amarpad:
Great catch btw.

lte/gateway/python/magma/enodebd/tr069/spyne_mods.py
ardzoht:
```suggestion
        if (ctx.descriptor.out_message in self._attrcache):
```

lte/gateway/python/magma/enodebd/tr069/spyne_mods.py
```diff
@@ -140,6 +140,15 @@ def get_call_handles(self, ctx):
...
 
     def serialize(self, ctx, message):
+        # Workaround for issue https://github.com/magma/magma/issues/7869
+        # Updates to ctx.descriptor.out_message.Attributes.sub_name are taking
+        # effect on the descriptor. But when puled from _attrcache dictionary,
+        # it still has a stale value.
+        # Force repopulation of dictionary by deleting entry
+        # TODO Remove this code once we have a better fix
+        if (self._attrcache.__contains__(ctx.descriptor.out_message)):
+            self._attrcache.__delitem__(ctx.descriptor.out_message)
```
ardzoht:
```suggestion
            del self._attrcache[ctx.descriptor.out_message]
```

ardzoht (changes requested):
couple of changes requested due to magic methods usage

lte/gateway/python/magma/enodebd/tr069/spyne_mods.py
talkhasib: State machine tests do not cover TR69 Soap11 functions. And TR69 Soap11 tests are simple single message tests.
This requires a new test that combines a bit of both

talkhasib:
Added one extra little change that might not be related but was  blocker for my testing
Code in gps_tr181(value: str)was not catching TypeError when the passed value was None

``` 
    try:
        return str(float(value) / 1e6)
    except ValueError:
        return value
```

themarwhal:
For Python Format check, please run
```
# have docker running
./$MAGMA_ROOT/lte/gateway/python/precommit.py --build
./$MAGMA_ROOT/lte/gateway/python/precommit.py --format --path lte/gateway/python/magma/enodebd/data_models/transform_for_magma.py
```
And for the Semantic Pull Request check, you'll have to follow the instructions here: https://sourcegraph.com/github.com/magma/magma/-/blob/docs/readmes/contributing/contribute_commit_msg.md

So eomthing like `fix(enodebd): Add workaround logic for ...`

ardzoht (approved):
looks like there's some formatting left to do (on Marie's comment), apart from that lgtm

PR #7911, opened 2021-07-06
refactor: remove snake case naming from test methods

waqaraqeel:

## Summary

This PR just renames internal test functions to use camelCase. snake_case should not be used in Golang.

## Test Plan

Ran `./build.py -t`. All passed.


themarwhal (approved):
good catch! wonder how that sneaked in :D

PR #7915, opened 2021-07-07
[orc8r] add swagger per service health status

uri200:

## Summary

This PR adds per service health status to swagger health API. This will allow to see what specific services are not healthy.

Note that if health service is disabled in any service, its entry will NOT show at all.

Being `additionalProp1` the name of each service with health enabled

```
    "Name_of_service": {
      "health_status": "HEALTHY",
      "service_state": "AVAILABLE"
    },
```
## Test Plan
image

## Additional Information


emakeev (approved):
LGTM

Issue #7348, opened 2021-06-06
Magma not deleting UE context after UE release command leading to asynchronous UE state between eNodeB and AGW


shehzadalee:
Your Environment
**Version:**1.3.3-1614917907-371c6462
Affected Component: Access Gateway
Affected Subcomponent: Mobilityd
Deployment Environment: bare-metal (AGW)

### Describe the Issue

Due to Radio Coverage Flap , UE Release Procedure takes place i.e. UE Release Request received from eNodeB with Cause 21 based on which UE Release command is sent by Magma and received UE Release complete. Yet the MME never removed the context of this UE instead added a paging flow. Later the same UE sent another GUTI based Attach Request (another UE ID)  which was accepted by Magma, but this new UE Id didnt update in MME. Later, Magma released old UE ID of this UE due to Expired Implicit Detach Timer, (since the UE is no more using this ue id, therefore no signaling happens with this UE and therefore timer expires) , Since the UE ID at Magma was never updated, Magma releases old UE ID due to Implicit Detach and sends UE Release Command for old UE ID to the eNodeB, while the eNodeB doesnt have this context any more since it already deleted during previous release procedure and maintains another (new) UE ID against this UE. This results in asynchronous state of UE between eNodeB and AGW.

**To Reproduce**
1. Enable DHCP IP assignment for UEs
2. Attach the UE
3. Simulate UE Release request with Cause 21.
4. Try another GUTI Based Attach.
5. check the logs

**Expected behavior**
Magma MME should have released the context during first UE Release procedure. and should have updated new UE ID for the new GUTI based Attached for this UE.

**Screenshots**

@15:46:35 UE Release Request was received from the eNodeB with Cause 21 against UE ID 6, IMSI 314030000015580 which followed the normal procedure until UE Release complete received. However magma added a paging flow for IP 10.1.39.174, the IP which was assigned by the DHCP Server.
![image](https://user-images.githubusercontent.com/77636702/120908376-48148900-c62f-11eb-87dd-3ae3130e2465.png)

@15:46:49 GUTI based Attached request received for same UE with UE ID 17, Magma successfully Attaches the UE.
![image](https://user-images.githubusercontent.com/77636702/120908428-c4a76780-c62f-11eb-8a28-a795a4daa303.png)

Re-allocation of IP 10.1.39.174 for sid IMSI314030000015580.internet without MME releasing can be clearly seen
[...]

PR #7916, opened 2021-07-07
fix(mme): Fix for handling unknown GUTI attach

pruthvihebbani:
[bug fix][lte][agw] Fix for handling unknown GUTI attach

## Summary

Issue : UE does GUTI attach with 0 M-TMSI, MME is not able to fetch the old context and perform implicit detach. Due to this the mobile reachability timer which was started during the previous attach does not get stopped. MME proceeds with the new attach without clearing the old context which leads to expiry of mobile reachability timer. 
This PR fixes the issue https://github.com/magma/magma/issues/7348 by performing identification procedure in case MME is not able to fetch context using GUTI. After receiving IMSI in identification response message, if context already exists for the user, MME triggers implicit detach and then proceeds with the new attach

## Test Plan

- Added a new s1ap tester TC to send GUTI attach with 0 M-TMSI and verified that MME deletes the old context after identification procedure and then proceeds with the new attach.
- Verified s1ap tester sanity


lte/gateway/c/core/oai/tasks/nas/emm/Attach.c
```diff
@@ -616,6 +622,10 @@ status_code_e emm_proc_attach_request(
           VOICE_DOMAIN_PREF_UE_USAGE_SETTING;
     }
   }
+  if (ue_mm_context->is_unknown_guti) {
+    ue_mm_context->is_unknown_guti = false;
```
amarpad: Why are we doing this? if we set it as an unknown_guti is false will we trigger an implicit detach on the old context?

lte/gateway/c/core/oai/tasks/nas/emm/Identification.c
```diff
@@ -257,7 +262,33 @@ status_code_e emm_proc_identification_complete(
...
         imsi64_t imsi64 = imsi_to_imsi64(imsi);
-        int emm_cause   = check_plmn_restriction(*imsi);
+        // If context already exists for this IMSI ,perform implicit detach
+        mme_app_desc_t* mme_app_desc_p      = get_mme_nas_state(false);
+        ue_mm_context_t* old_imsi_ue_mm_ctx = mme_ue_context_exists_imsi(
+            &mme_app_desc_p->mme_ue_contexts, imsi64);
+        if (emm_ctx->is_unknown_guti && old_imsi_ue_mm_ctx) {
+          OAILOG_INFO_UE(
+              LOG_NAS_EMM, imsi64,
+              "EMMAS-SAP - Old ue context already exists for for ue_id "
```
ardzoht:
```suggestion
              "EMMAS-SAP - UE context already exists for for ue_id "
```

lte/gateway/c/core/oai/tasks/nas/emm/Identification.c
```diff
@@ -53,6 +54,9 @@ extern mme_congestion_params_t mme_congestion_params;
 
 extern int check_plmn_restriction(imsi_t imsi);
 extern int validate_imei(imeisv_t* imeisv);
+extern void create_new_attach_info(
```
ardzoht: Would it make sense to move these function definitions to a common header file? Not sure what would be the best place to put these maybe `common_utility_funcs.h` or create a header file under `tasks/nas/emm/` to put these just to avoid having definitions on an implementation file

ardzoht (changes requested):

lte/gateway/c/core/oai/tasks/nas/emm/Attach.c
pruthvihebbani: I removed ue_mm_context->is_unknown_guti and added a local flag. This flag is initially set when attach request is received with unknown GUTI because at this point emm_context is not yet created. Once the new emm context is created, if  is_unknown_guti is set I am setting is_unknown_guti flag is emm_context because to emm_attach_run_procedure() and emm_proc_identification() functions only emm_context gets passed. After receiving identification complete, emm_ctx->is_unknown_guti flag is used to decide whether implicit detach should be triggered. After implicit detach, this flag is used to skip emm_proc_create_procedure_attach_request() and emm_attach_run_procedure() and directly initiate authentication procedure

lte/gateway/c/core/oai/tasks/nas/emm/Identification.c
pruthvihebbani: I moved this function to nas/emm/EmmCommon.c where other emm common functions are defined

lte/gateway/c/core/oai/tasks/nas/emm/Attach.c
amarpad: Hi Pruthvi I would model this as an enum instead of a bool as it is not describing the state of the current emm_context is unknown_guti -> new_emm_context_created. The emm_context that gets passed is the new emm_context.

lte/gateway/c/core/oai/tasks/nas/emm/emm_data.h
```diff
@@ -119,6 +119,11 @@ typedef struct emm_security_context_s {
 struct emm_common_data_s;
 
 typedef enum { SUCCESS, FAILURE } sgs_loc_updt_status_t;
+typedef enum {
+  NEW_EMM_CONTEXT_NOT_CREATED,
+  NEW_EMM_CONTEXT_CREATED,
+  UNKOWN_GUTI
```
ardzoht:
```suggestion
  UNKNOWN_GUTI
```

ardzoht (approved):
lgtm, please resolve the typo on the enum

PR #7920, opened 2021-07-07
Pull Stats Model Config

veshkemburu:

## Summary

Easily enable/disable the pull stats model using a flag. This flag connects directly to sessiond_main which creates the threads that poll pipelined. Disabling the flag would revert to the default push model.

## Test Plan

Added config flag to sessiond.yml files
Debugged output for both models(pull and push)

## Additional Information

### DIFF
--- a/cwf/gateway/configs/sessiond.yml
+++ b/cwf/gateway/configs/sessiond.yml
@@ -61,3 +61,6 @@ enable_ipfix: true
 # set to a certain interval measured in seconds for which sessiond should poll pipelined
 # for relevant stats
 poll_stats_interval: 5
+
+#set to true to enable pull model for stats(polling pipelined from sessiond)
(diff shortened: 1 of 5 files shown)


koolzz (approved):

cwf/gateway/configs/sessiond.yml
```diff
@@ -61,3 +61,6 @@ enable_ipfix: true
 # set to a certain interval measured in seconds for which sessiond should poll pipelined
 # for relevant stats
 poll_stats_interval: 5
+
+#set to true to enable pull model for stats(polling pipelined from sessiond)
```
themarwhal: `#set` -> `# set` for all configs, sorry I missed this earlier

themarwhal (approved):

PR #7923, opened 2021-07-07
feat(agw): Adapt subscriberdb agw client to make SyncSubscribers calls

wangyyt1013:
## Summary

1. Add functions in `SqliteStore` that stores and manages gateway-local per-subscriber digests
2. Add functions to ensure thread-safety of servicer-response-triggered `per-sub digests` store and `subscribers` store operations
3. Adapt the `subscriberdb` agw client to incorporate `SyncSubscribers` calls to cloud as part of its callpath

## Test Plan

2. Added unit tests for the new `SqliteStore` functions, the modified`subscriberdb` client callpath, as well as the `OnDataReady` and `OnDigestsReady` mixin interfaces 

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/subscriberdb/client.py
+++ b/lte/gateway/python/magma/subscriberdb/client.py
@@ -13,7 +13,7 @@ limitations under the License.
 import asyncio
 import datetime
 import logging
-from collections import namedtuple
+from typing import List, NamedTuple, Optional
 
 import grpc
 from lte.protos.s6a_service_pb2 import DeleteSubscriberRequest
@@ -24,6 +24,7 @@ from lte.protos.subscriberdb_pb2 import (
     ListSubscribersRequest,
     LTESubscription,
     SubscriberData,
+    SubscriberDigestWithID,
     SyncSubscribersRequest,
 )
 from magma.common.grpc_client_manager import GRPCClientManager
@@ -37,10 +38,11 @@ from magma.subscriberdb.metrics import (
 )
 from magma.subscriberdb.store.sqlite import SqliteStore
 
-CloudSubscribersInfo = namedtuple(
-    'CloudSubscribersInfo',
-    ['subscribers', 'flat_digest', 'per_sub_digests'],
-)
+
+class CloudSubscribersInfo(NamedTuple):
+    subscribers: List[SubscriberData]
+    flat_digest: Optional[Digest]
+    per_sub_digests: Optional[List[SubscriberDigestWithID]]
 
 
 class SubscriberDBCloudClient(SDWatchdogTask):
@@ -174,7 +176,7 @@ class SubscriberDBCloudClient(SDWatchdogTask):
 
         return res.resync
 
-    async def _get_all_subscribers(self) -> CloudSubscribersInfo:
+    async def _get_all_subscribers(self) -> Optional[CloudSubscribersInfo]:
         subscriberdb_cloud_client = self._grpc_client_manager.get_client()
         subscribers = []
         flat_digest = None
@@ -234,17 +236,20 @@ class SubscriberDBCloudClient(SDWatchdogTask):
         )
         return subscribers_info
 
-    def _update_flat_digest(self, flat_digest: Digest) -> None:
+    def _update_flat_digest(self, flat_digest: Optional[Digest]) -> None:
         if Digest is None:
             return
         self._store.update_digest(flat_digest.md5_base64_digest)
 
-    def _update_per_sub_digests(self, per_sub_digests: list) -> None:
+    def _update_per_sub_digests(
+            self,
+            per_sub_digests: Optional[List[SubscriberDigestWithID]],
+    ) -> None:
         if per_sub_digests is None:
             return
         self._store.update_per_sub_digests(per_sub_digests)
 
-    def _process_subscribers(self, subscribers: SubscriberData) -> None:
+    def _process_subscribers(self, subscribers: List[SubscriberData]) -> None:
         active_subscriber_ids = []
         sids = []
         for subscriber in subscribers:
--- a/lte/gateway/python/magma/subscriberdb/client.py
+++ b/lte/gateway/python/magma/subscriberdb/client.py
@@ -13,6 +13,7 @@ limitations under the License.
 import asyncio
 import datetime
 import logging
+from collections import namedtuple
 
 import grpc
 from lte.protos.s6a_service_pb2 import DeleteSubscriberRequest
@@ -23,6 +24,7 @@ from lte.protos.subscriberdb_pb2 import (
     ListSubscribersRequest,
     LTESubscription,
     SubscriberData,
+    SyncSubscribersRequest,
 )
 from magma.common.grpc_client_manager import GRPCClientManager
 from magma.common.rpc_utils import grpc_async_wrapper
@@ -34,6 +36,11 @@ from magma.subscriberdb.metrics import (
     SUBSCRIBER_SYNC_SUCCESS_TOTAL,
 )
 from magma.subscriberdb.store.sqlite import SqliteStore
+
+CloudSubscribersInfo = namedtuple(
+    'CloudSubscribersInfo',
+    ['subscribers', 'flat_digest', 'per_sub_digests'],
+)
 
 
 class SubscriberDBCloudClient(SDWatchdogTask):
@@ -83,13 +90,18 @@ class SubscriberDBCloudClient(SDWatchdogTask):
         if in_sync:
             return
 
-        subscribers, flat_digest = await self._get_all_subscribers()
-        if subscribers is None:
+        resync = await self._sync_subscribers()
+        if not resync:
+            return
+
+        subscribers_info = await self._get_all_subscribers()
+        if subscribers_info is None:
             return
         # Process subscriber data before digest data, in case there's a gateway
         # failure between the calls
-        self._process_subscribers(subscribers)
-        self._update_flat_digest(flat_digest)
+        self._process_subscribers(subscribers_info.subscribers)
+        self._update_flat_digest(subscribers_info.flat_digest)
+        self._update_per_sub_digests(subscribers_info.per_sub_digests)
 
     async def _check_subscribers_in_sync(self) -> bool:
         """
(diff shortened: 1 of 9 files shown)


lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -119,10 +131,50 @@ async def _check_subscribers_in_sync(self) -> bool:
...
+                "Sync subscribers request error! [%s] %s", err.code(),
+                err.details(),
+            )
+            return True
+
+        if not res.resync:
+            self._update_flat_digest(res.flat_digest)
+            self._update_per_sub_digests(res.per_sub_digests)
+
+            for sid in res.to_renew:
```
hcgatewood: Fix lint (use `.items()`)

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -119,10 +131,50 @@ async def _check_subscribers_in_sync(self) -> bool:
...
+                subscriber_data = res.to_renew[sid]
+                self._store.upsert_subscriber(subscriber_data)
+            for sid in res.deleted:
+                self._store.delete_subscriber(sid)
```
hcgatewood, line 177: Edit: I see that this is how we were doing it before. Let's keep it as-is then, and we can defer the bulk-edit functionality to a future todo. Can you add a `TODO(hcgatewood)`?

This could be up to 500 subscribers -- instead of doing one at a time, can we bulk-upsert and/or bulk-delete?

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -170,13 +223,22 @@ async def _get_all_subscribers(self) -> (SubscriberData, Digest):
             time_elapsed.total_seconds() * 1000,
         )
 
-        return subscribers, flat_digest
+        return CloudSubscribersInfo(
```
hcgatewood: Nit: I prefer single-line returns, even in Python code. So move the var decl to the previous line

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -292,7 +348,8 @@ def get_current_digest(self) -> str:
         try:
             with conn:
                 res = conn.execute(
-                    "SELECT digest, updated_at FROM subscriber_digest",
+                    "SELECT digest, updated_at FROM subscriber_digest "
+                    "ORDER BY updated_at DESC",
```
hcgatewood, line 367: Why order by?

hcgatewood (commented):
Here or in DM: can you explain the concurrency/event loop changes you made? That's the main part that I'm not following

wangyyt1013 (commented):
Test comment for toast

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -119,10 +134,51 @@ async def _check_subscribers_in_sync(self) -> bool:
...
+            # TODO(hcgatewood): switch to bulk editing subscriber data rows
+            for item in res.to_renew.items():
+                subscriber_data = item[1]
+                self._store.upsert_subscriber(subscriber_data)
+            for sid in res.deleted:
+                self._store.delete_subscriber(sid)
+
+        return res.resync
+
+    async def _get_all_subscribers(self) -> CloudSubscribersInfo:
```
ardzoht:
```suggestion
    async def _get_all_subscribers(self) -> Optional[CloudSubscribersInfo]:
```

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -35,13 +39,14 @@ class SqliteStore(BaseStore):
...
+        self._db_locations = self._create_db_locations(db_location, self._n_shards)
+        self._digest_db_location, self._per_sub_digest_db_location = \
+            self._create_digest_db_locations(db_location)
         self._create_store()
         self._on_ready = OnDataReady(loop=loop)
+        self._on_digests_ready = OnDigestsReady(loop=loop)
 
-    def _create_db_locations(self, db_location: str, n_shards: int) \
-            -> [list, str]:
+    def _create_db_locations(self, db_location: str, n_shards: int) -> list:
```
ardzoht:
```suggestion
    def _create_db_locations(self, db_location: str, n_shards: int) -> List:
```

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -61,11 +66,21 @@ def _create_db_locations(self, db_location: str, n_shards: int) \
...
 
-        return db_location_list, digest_db_location
+        per_sub_digest_db_location = 'file:' + db_location + \
+                                     'per-subscriber-digest.db?cache=shared'
+        logging.info(
+            "per-sub digest db location: %s",
+            per_sub_digest_db_location,
+        )
+
+        return digest_db_location, per_sub_digest_db_location
```
ardzoht: Let's create a NamedTuple (maybe name it DigestDBInfo?) obj for this

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -146,7 +171,37 @@ def edit_subscriber(self, subscriber_id):
...
-    def delete_subscriber(self, subscriber_id):
+    def upsert_subscriber(self, subscriber_data: SubscriberData) -> None:
+        """
+        Check if the given subscriber exists in store. If so, update subscriber
+        data; otherwise, add subscriber.
+        """
+        sid = SIDUtils.to_str(subscriber_data.sid)
+        data_str = subscriber_data.SerializeToString()
+        db_location = self._db_locations[self._sid2bucket(sid)]
+        conn = sqlite3.connect(db_location, uri=True)
```
ardzoht, line 197: what is the scope for the number of calls `upsert_subscriber`? also this should happen every certain amount of time right?

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -321,10 +378,52 @@ def update_digest(self, new_digest: str) -> None:
             conn.close()
 
         logging.info("update digest stored in gateway: %s", new_digest)
+        self._on_digests_ready.update_digest(new_digest)
+
+    def get_current_per_sub_digests(self) -> list:
```
ardzoht:
```suggestion
    def get_current_per_sub_digests(self) -> List:
```

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -321,10 +378,52 @@ def update_digest(self, new_digest: str) -> None:
...
+                        sid=SIDUtils.to_pb(row[0]),
+                        digest=Digest(md5_base64_digest=row[1]),
+                    )
+                    digests.append(digest)
+        finally:
+            conn.close()
+
+        return digests
+
+    def update_per_sub_digests(self, new_digests: list) -> None:
```
ardzoht:
```suggestion
    def update_per_sub_digests(self, new_digests: List) -> None:
```

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -35,6 +37,11 @@
 )
 from magma.subscriberdb.store.sqlite import SqliteStore
 
+CloudSubscribersInfo = namedtuple(
```
ardzoht: Use NamedTuple as it supports type hinting as well

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -61,11 +66,21 @@ def _create_db_locations(self, db_location: str, n_shards: int) \
...
 
+        return db_location_list
+
+    def _create_digest_db_locations(self, db_location: str) -> [str, str]:
         digest_db_location = 'file:' + db_location + \
                              'subscriber-digest.db?cache=shared'
         logging.info("digest db location: %s", digest_db_location)
 
-        return db_location_list, digest_db_location
+        per_sub_digest_db_location = 'file:' + db_location + \
```
ardzoht, line 87: Maybe I'm missing something here, but what is the difference between `per_sub_digest_db` and `digest_db`? afaik `digest_db` is used to store the flat digests right? Isn't that per subscriber as well?

ardzoht (changes requested):
Some minor nits on functions type hinting and left questions regarding the purpose of the DBs

lte/gateway/python/magma/subscriberdb/store/sqlite.py
wangyyt1013: `digest_db` is used to store exclusively *flat digests*, which are digests for an entire network (including all subscribers in it);
`per_sub_digest_db` stores exclusively *per-subscriber digests*, which are digests for every individual subscriber for every network. Hope this helps! @ardzoht

lte/gateway/python/magma/subscriberdb/store/sqlite.py
wangyyt1013: `upsert_subscriber` is called only within the `SyncSubscribers` handler (i.e. `_sync_subscribers()` in `client.py`) right now. Each changeset sent back from the cloud should be no larger than a configurable size (currently 500), so `upsert_subscriber` won't be called for more than 500 times for every sync.

The `upsert_subscriber` in `_sync_subscribers` is called only when 
1. the cloud store has changed, and the AGW data is out-of-date AND
2. the changeset between cloud and AGW is no larger than 500

As a result, the AGW checks whether it's in sync with the cloud every 5 minutes (configurable), but `_sync_subscribers` is not necessarily triggered, or wouldn't call `upsert_subscriber`, every time there's a sync. @ardzoht

hcgatewood (changes requested):
Looks like you may need to detach_deleted_subscribers for all the deleted subscribes during the sync update (?)

Other than that, I'll be ready to stamp and can defer AGW reviews to @themarwhal and/or @ardzoht

ardzoht:
Before landing, I would suggest running `make integ_test` on the VM to validate these changes as part of sanity testing and add this validation to your test plan.

ardzoht (approved):
synced offline with Elaine, `make integ_test` ran successfully, lgtm

hcgatewood (approved):

PR #7924, opened 2021-07-07
fix(ci): Try fix running build.py from different dir

hcgatewood:
## Summary

Try fix https://magmacore.slack.com/archives/C0191LLL03S/p1625653920366600?thread_ts=1624980159.325600&cid=C0191LLL03S

## Additional Information


themarwhal (approved):

hcgatewood:
Closing since it seems using PWD is the only way to access a dockerfile outside the indicated context, and I don't think it's worth copying the dockerfiles into the context just for this purpose

Ref: https://github.com/docker/compose/issues/4926#issuecomment-370500372

PR #7446, opened 2021-06-09
feat(mme): NSA-FLAGS addition in Create session procedure and Update location procedure. 

Vyas20:
[Converged-MME]  

NSA-FLAGS for create session request and Update location procedure (pending from PR #5749 ).

This PR is tested with S1AP-Test suite, ds-Tester.

PR #7930, opened 2021-07-08
[Converged-MME]AMBR Extended-2 field addition in EPS QoS IE and in APN-AMBR. 

Vyas20:
[Converged-MME]AMBR Extended-2 field addition in NAS layer for EPS QoS IE and in APN-AMBR IE . 

This is used to accommodate AMBR values greater than 256 Mbps in both uplink and downlink direction.

This PR is based on PR #7446 and is tested using S1AP tester test suite and dsTester.


Vyas20:
Changes from this PR will be landed in a different PR because this was cascaded to another PR (#7446 ).

PR #7946, opened 2021-07-08
perf(agw): set proportional memory limit to the total

pshelar:
In cases AGW has higher RAM than 4GB. it could make
use of higher memory limit to avoid using swap.
Following patch changes the memory limit of service which impacts
control-plane and dataplane performance: mme, sessionD and pipelineD.
mme and pipelineD changed to 13%, which translates to 532 MB
on 4GB RAM and sessionD to 8% which translate to 327 MB.

## Additional Information


uri200:
By the way, what happen if agw has less than 8gb. Isn't it going to hit performance?
Maybe is somewhere in the doc, but are we advertising somewhere AGW has to run with min of 8gb?

themarwhal:
> By the way, what happen if agw has less than 8gb. Isn't it going to hit performance?
> Maybe is somewhere in the doc, but are we advertising somewhere AGW has to run with min of 8gb?

We seem to say 4GB here? https://docs.magmacore.org/docs/next/basics/prerequisites#access-gateways

pshelar:
> > By the way, what happen if agw has less than 8gb. Isn't it going to hit performance?
> > Maybe is somewhere in the doc, but are we advertising somewhere AGW has to run with min of 8gb?
> 
> We seem to say 4GB here? https://docs.magmacore.org/docs/next/basics/prerequisites#access-gateways

I see, our lab AGW and vagrant has 8GB. so I thought its the minimum required memory. I can double the percentage limit.

uri200:
> > > By the way, what happen if agw has less than 8gb. Isn't it going to hit performance?
> > > Maybe is somewhere in the doc, but are we advertising somewhere AGW has to run with min of 8gb?
> > 
> > We seem to say 4GB here? https://docs.magmacore.org/docs/next/basics/prerequisites#access-gateways
> 
> I see, our lab AGW and vagrant has 8GB. so I thought its the minimum required memory. I can double the percentage limit.

Yes we needed to increase Vagrant from 4 to 6 and then to 8gb some time ago to be able to build, but not to run magma.

Also not sure if this is related to an issue reported today on slack, but if so, that was probably caused because they built magma using `fab dev`, instead of `fab test` (I gave them the wrong command initially).

pshelar:
> > > > By the way, what happen if agw has less than 8gb. Isn't it going to hit performance?
> > > > Maybe is somewhere in the doc, but are we advertising somewhere AGW has to run with min of 8gb?
> > > 
> > > 
> > > We seem to say 4GB here? https://docs.magmacore.org/docs/next/basics/prerequisites#access-gateways
> > 
> > 
> > I see, our lab AGW and vagrant has 8GB. so I thought its the minimum required memory. I can double the percentage limit.
> 
> Yes we needed to increase Vagrant from 4 to 6 and then to 8gb some time ago to be able to build, but not to run magma.
> 
> Also not sure if this is related to an issue reported today on slack, but if so, that was probably caused because they built magma using `fab dev`, instead of `fab test` (I gave them the wrong command initially).

Agree, the dev build has debugging flags set which needs more memory.
But there is another issue reported where AGW is using swap space even when there is huge amount of RAM available, due to the memory limit. the swap usage can cause performance issue on slower disc swap device.

themarwhal (approved):

uri200 (approved):

ulaskozat:
I suggest to revisit these numbers as although current recommended minimum value is 4GB, there are deployments with less. We should not drop below the original hard limits in those cases.

pshelar:
> I suggest to revisit these numbers as although current recommended minimum value is 4GB, there are deployments with less. We should not drop below the original hard limits in those cases.

Even in case of 2GB RAM we need to make sure it is divided proportionately. otherwise mme can have disproportionate memory. what exactly are you worried in lower memory AGW?

PR #7949, opened 2021-07-08
docs(ci): Document semantic commit requirement

hcgatewood:
## Additional Information


themarwhal (approved):
doc looks good. Why the semantic config change?

hcgatewood:
> doc looks good. Why the semantic config change?

Should be a nil-effect. Just cleaning up for conciseness

PR #7950, opened 2021-07-08
fix(ci): point mme_test job determinator to check c/core not c/oai

themarwhal:

## Summary
We moved mme code from lte/gateway/c/oai to lte/gateway/c/core ~ a month ago, but we didn't change the build determinator for the mme unit test job.

This means the precommit check for mme unit test was a no-op for close to a month. (big yikes)

## Test Plan
this should work 

## Additional Information


amarpad (approved):
oops.

quentinDERORY (approved):

PR #7967, opened 2021-07-09
fix(ci): change the python format check so it comments on failure

themarwhal:

## Summary
changing the trigger event to pull_request_target may allow us to have a comment on check failures

## Test Plan
test on master :/ 

## Additional Information

### DIFF
--- a/.github/workflows/python-formatting-check.yml
+++ b/.github/workflows/python-formatting-check.yml
@@ -1,10 +1,9 @@
 ---
 name: Python Format Check
 on:  # yamllint disable-line rule:truthy
-  pull_request:
+  pull_request_target:
     types:
       - opened
-      - reopened
       - synchronize
     paths:  # order matters here, always put includes then excludes
       - 'lte/gateway/python/**.py'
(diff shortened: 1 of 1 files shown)


quentinDERORY (changes requested):
There are security issues  involved when using pull_request_target with checkout.
Malicious code could be run from the forked PR which will gain permission on secrets.
There is a restriction for forked PR workflow run but it's only for first time contributors.
It's described in more details here: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
To avoid this issue you can create another workflow like I did here: https://github.com/magma/magma/pull/7905/files

.github/workflows/python-formatting-check.yml
```diff
@@ -51,3 +50,18 @@ jobs:
         run: |
```
quentinDERORY, line 50: You can delete this step now.

themarwhal:
i see! didn't realize you already had a solution for this. I'll close this for now and do another change once your workflow is working

PR #7980, opened 2021-07-09
fix(mme): add config flag to control GTP-U echo

pshelar:
This is turned off by default to fix perf regression.

## Test Plan
validated it on vagrant test setup.

## Additional Information

### DIFF
--- a/lte/gateway/configs/templates/spgw.conf.template
+++ b/lte/gateway/configs/templates/spgw.conf.template
@@ -46,6 +46,7 @@ S-GW :
       UPLINK_PORT_NUM                      = {{ ovs_uplink_port_number }};
       UPLINK_MAC                           = "{{ ovs_uplink_mac }}";
       MULTI_TUNNEL                         = "{{ ovs_multi_tunnel }}";
+      GTP_ECHO                             = "{{ ovs_gtpu_echo_resp }}";
       PIPELINED_CONFIG_ENABLED             = "{{ pipelined_managed_tbl0 }}";
     };
 };
--- a/lte/gateway/c/core/oai/include/sgw_config.h
+++ b/lte/gateway/c/core/oai/include/sgw_config.h
@@ -59,6 +59,7 @@
 #define SGW_CONFIG_STRING_OVS_UPLINK_PORT_NUM "UPLINK_PORT_NUM"
 #define SGW_CONFIG_STRING_OVS_UPLINK_MAC "UPLINK_MAC"
 #define SGW_CONFIG_STRING_OVS_MULTI_TUNNEL "MULTI_TUNNEL"
+#define SGW_CONFIG_STRING_OVS_GTP_ECHO "GTP_ECHO"
 #define SGW_CONFIG_STRING_OVS_PIPELINED_CONFIG_ENABLED                         \
   "PIPELINED_CONFIG_ENABLED"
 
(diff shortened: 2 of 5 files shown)


themarwhal (approved):

lte/gateway/configs/templates/spgw.conf.template
```diff
@@ -46,6 +46,7 @@ S-GW :
       UPLINK_PORT_NUM                      = {{ ovs_uplink_port_number }};
       UPLINK_MAC                           = "{{ ovs_uplink_mac }}";
       MULTI_TUNNEL                         = "{{ ovs_multi_tunnel }}";
+      GTP_ECHO                             = "{{ ovs_gtpu_echo_resp }}";
```
ardzoht, line 49: where is this generated from? there should be a `ovs_gtpu_echo_resp` config param on `spgw.yml` that this value updates from.

ardzoht (changes requested):
There is a missing config value on yml file I think. Also for these type of changes I'd suggest including on your test plan an output that validates how this value gets updated (on mme.log for example)

lte/gateway/configs/templates/spgw.conf.template
pshelar: in case of missing value, the feature is off. thats why I did not add the variable to yml file.

pshelar:
> There is a missing config value on yml file I think. Also for these type of changes I'd suggest including on your test plan an output that validates how this value gets updated (on mme.log for example)

I ran test-attach-detach with this change. and it works. do you see any potential issue due to missing config in this case?

lte/gateway/configs/templates/spgw.conf.template
ardzoht: Ok got your point, MME config is checking on the existence of the value. 

I'd strongly advise against this, at least on MME and OAI configuration generation there is an explicit declaration for all configuration both on mme.yml and spgw.yml files which `generate_oai_config` parses the values and puts them on the given template.

My preference would be to declare the value on the yml file and use that to control the enablement/disablement of the feature, this way we also have expectations between what different versions of what service configs files should contain.

ardzoht:
> > There is a missing config value on yml file I think. Also for these type of changes I'd suggest including on your test plan an output that validates how this value gets updated (on mme.log for example)
> 
> I ran test-attach-detach with this change. and it works. do you see any potential issue due to missing config in this case?

I just want to confirm validation that `generate_oai_config` is working as expected for this value, as I mentioned on my other comment `generate_oai_config.py` should be the manager for validating existence of config and not the MME service itself.

lte/gateway/configs/templates/spgw.conf.template
pshelar: ok. updated the PR.

pshelar:
> > > There is a missing config value on yml file I think. Also for these type of changes I'd suggest including on your test plan an output that validates how this value gets updated (on mme.log for example)
> > 
> > 
> > I ran test-attach-detach with this change. and it works. do you see any potential issue due to missing config in this case?
> 
> I just want to confirm validation that `generate_oai_config` is working as expected for this value, as I mentioned on my other comment `generate_oai_config.py` should be the manager for validating existence of config and not the MME service itself.

MME restart and performs as per the config for the attach-detach-test. it does validate generate_oai_config, right?

lte/gateway/configs/templates/spgw.conf.template
ardzoht: cool, and thanks for the validation!

ardzoht (approved):
lgtm

PR #7983, opened 2021-07-10
fix(agw): Process subscriber data before digest data

wangyyt1013:
## Summary

1. Switched ordering of processing subscriber data (now first) and digest data (now second) in case there's a gw failure between the two calls

## Test Plan

## Additional Information

- This fix is meant to be back-ported to v1.6


hcgatewood (changes requested):
As discussed offline: open this PR against master and add the `apply-v1.6` label

hcgatewood (commented):
As discussed offline: need to fix the base branch

hcgatewood (approved):

ardzoht (approved):
lgtm

PR #7985, opened 2021-07-11
fix(mme): [S1 handover] fix bugs in security context for HandoverRequest Msg

talkhasib:

## Summary

Security context next hop info were not properly computed or populated for HandoverRequest Msg. This PR resolves the following issues

1- A copy of emm_context was used to update next_hop and nhcc. These changes did not persist
2- We only copied 1 byte of the next_hop vector into the security context ie
3- We needed to bump the NHCC and next_hop before we copied into security context

The last point above needs investigation as noted with a comment on the code. We are either missing a bump to NHCC and next_hop somewhere else. Or this is the correct behaviour

Another place that needs investigation is the path switch msg for X2 handover as it populates a similar security context.

## Test Plan

Tested on local setup with 2 baicells eNBs and a pixel3 phone
pcap is attached
[S1HO_UE_Baicells.pcap.zip](https://github.com/magma/magma/files/6797598/S1HO_UE_Baicells.pcap.zip)

### DIFF
--- a/lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c
+++ b/lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c
@@ -292,7 +292,7 @@ void mme_app_handle_conn_est_cnf(
     nas_establish_rsp_t* const nas_conn_est_cnf_p) {
   OAILOG_FUNC_IN(LOG_MME_APP);
   struct ue_mm_context_s* ue_context_p                             = NULL;
-  emm_context_t emm_context                                        = {0};
+  emm_context_t* emm_context_p                                     = NULL;
   MessageDef* message_p                                            = NULL;
   itti_mme_app_connection_establishment_cnf_t* establishment_cnf_p = NULL;
   int rc                                                           = RETURNok;
@@ -313,7 +313,7 @@ void mme_app_handle_conn_est_cnf(
     bdestroy_wrapper(&nas_conn_est_cnf_p->nas_msg);
     OAILOG_FUNC_OUT(LOG_MME_APP);
   }
-  emm_context = ue_context_p->emm_context;
+  emm_context_p = &ue_context_p->emm_context;
   /* Check that if Service Request is recieved in response to SGS Paging for MT
    * SMS */
   if (ue_context_p->sgs_context) {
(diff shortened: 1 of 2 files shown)


lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c
```diff
@@ -3336,6 +3334,13 @@ void mme_app_handle_handover_required(
         ue_context_p->emm_context._security.vector_index,
         ue_context_p->mme_ue_s1ap_id);
   }
+
+  /* TODO: investigate
+     1- Should we bump next_hop and use it in this message (as is)
+     2- or, use the previously calculated next_hop and bump up for next
+        iteration. In this case, we are missing a bump to next_hop someewhere
+        else, per testing
+  */
```
ulaskozat: TL;DR We should bump up next_hop and ncc, store these new values and send them to target eNB in S1 Handover Request.

Longer explanation:

https://www.etsi.org/deliver/etsi_ts/133400_133499/133401/15.07.00_60/ts_133401v150700p.pdf section 7.2.8.4.3 indicates that (source and target MMEs in this case are the same):
- "Upon reception of the HANDOVER REQUIRED message the source MME shall increase its locally kept NCC value
by one and compute a fresh NH from its stored data" 
- "The source MME shall store that fresh pair and send it to the target MME in the S10 FORWARD RELOCATION REQUEST message." (here S10 can be ignored as source and target MMEs are the same.
- "The target MME shall store locally the {NH, NCC} pair received from the source MME.The target MME shall then send the received {NH, NCC} pair to the target eNB within the S1 HANDOVER REQUEST."

ulaskozat (approved):
I think bump up behavior is correct as noted in my comment.

ulaskozat:
@talkhasib Why is the context release command is going to enb with ip ending 245 but not to enab with 249 (the requester)?

ulaskozat (changes requested):
Context release is ending up at the wrong eNB?

talkhasib:
> Context release is ending up at the wrong eNB?

Good catch
There is another issue, MME needs to send the release command to the source eNB once it gets the handover notify. We currently rely on the source eNB sending a release request.

I think these issues need to be addressed in a separate PR

shaddi:
Thank you for finally finding the fix for this. :-)

ulaskozat:
> > Context release is ending up at the wrong eNB?
> 
> Good catch
> There is another issue, MME needs to send the release command to the source eNB once it gets the handover notify. We currently rely on the source eNB sending a release request.
> 
> I think these issues need to be addressed in a separate PR

Okay, then let me stamp it again and you can merge it. I assume you did run the integ tests on this locally?

ulaskozat (approved):
LGTM. The remaining issues will be handled in a separate PR.

PR #7997, opened 2021-07-12
fix(pipelined): avoid cleaning up uplink-br0 in stateless mode.

pshelar:
Cleaning up uplink-br0 can disrupt traffic during service restart.

## Test Plan
Test it in lab setup.
 

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/uplink_bridge.py
+++ b/lte/gateway/python/magma/pipelined/app/uplink_bridge.py
@@ -54,6 +54,7 @@ class UplinkBridgeController(MagmaController):
 
         self.config = self._get_config(kwargs['config'])
         self.logger.info("uplink bridge app config: %s", self.config)
+        self._clean_restart = kwargs['config']['clean_restart']
 
     def _get_config(self, config_dict) -> namedtuple:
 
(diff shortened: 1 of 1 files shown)


amarpad (changes requested):
Don't you still have to check the flow table is inline with what it needs to be, what happens if the config changes?

amarpad (commented):
Also can you explain def initialize_on_connect(self, datapath) and if it wouldn't clear the flows as well? It has a delete_all_flows

pshelar:
> Don't you still have to check the flow table is inline with what it needs to be, what happens if the config changes?

initialize_on_connect() delete flows takes care of cleanup.

pshelar:
> Also can you explain def initialize_on_connect(self, datapath) and if it wouldn't clear the flows as well? It has a delete_all_flows

yes, but the uplink-controller adding flows right after that, so OVS would be missing flows for few milliseconds.
we can improve it by generation flows based on flow diff.

pshelar:
I validated it on bare metal AGW, it works fine.
even without this patch I dont see loosing flows in uplink_br0, so we dont needs to backport this PR.

amarpad:
But on 1.5 I saw the flows getting nuked because at a later point pipelined crashed. I still think we should be conservative and diff against forwarding state and not nuke flows as the downside of bricking a box is really big

amarpad (approved):
This is a different issue than the one we saw, approving this.

PR #7998, opened 2021-07-12
fix(cloud): Upgrading Cloudstrapper for v1.6

arunuke:

## Summary

Upgrading terraform to 0.15 or above (default for Ubuntu 20.04) since orc8r now supports higher versions of terraform

## Test Plan

Deploy Cloudstrapper and check terraform version. It should be 0.15 or above. 

## Additional Information

Older versions of Orc8r (1.5 or earlier) would not work with terraform 0.15


quentinDERORY (approved):
LGTM

PR #7999, opened 2021-07-12
fix(agw): Add 10.0.2.3 in resolv conf for Virtualbox VMs

tmdzk:

fix(agw) Add 10.0.2.3 in resolv conf for Virtualbox VMs

## Summary

fix(agw) Add 10.0.2.3 in resolv conf for Virtualbox VMs

## Test Plan
```
cd lte/gateway && vagrant destroy -f && vagrant up && vagrant ssh 
ping google.com
```

## Additional Information

None

### DIFF
--- a/lte/gateway/deploy/roles/dev_common/tasks/main.yml
+++ b/lte/gateway/deploy/roles/dev_common/tasks/main.yml
@@ -432,4 +432,4 @@
   copy:
     src: magma_resolv.conf
     dest: /etc/systemd/resolved.conf
-  when: preburn and ansible_distribution == "Ubuntu"
+  when: ansible_distribution == "Ubuntu"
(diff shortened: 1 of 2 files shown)


lte/gateway/deploy/roles/dev_common/tasks/main.yml
```diff
@@ -432,4 +432,4 @@
   copy:
     src: magma_resolv.conf
     dest: /etc/systemd/resolved.conf
-  when: preburn and ansible_distribution == "Ubuntu"
+  when: ansible_distribution == "Ubuntu"
```
pshelar, line 435: Now we could also remove check for distribution.

pshelar (approved):

lte/gateway/deploy/roles/dev_common/tasks/main.yml
tmdzk: True but this will be back to preburn soon as I'll build new images with the rnew esolved.conf in it, this is just tmp but yea I'll do some cleanup on this role

PR #7859, opened 2021-06-30
WIP [CI] Bump up ubuntu version for lte test

tmdzk:
Signed-off-by: Timothée Dzik <timdzik@fb.com>

WIP [CI] Bump up ubuntu version for lte test

## Summary

WIP [CI] Bump up ubuntu version for lte test

## Test Plan

Let CI go

PR #8000, opened 2021-07-12
refactor(agw): remove leftover dependency lists and doc refs for Debian

waqaraqeel:

## Summary

We have stopped supporting Debian for Magma release 1.6 onwards. There are some leftovers from when we supported both Debian and Ubuntu. This PR removes Debian-specific dependency lists from `{lte,orc8r}/gateway/python/setup.py`, removes `magma.lockfile.debian` and makes appropriate changes in the packaging README.

## Test Plan

Did a `vagrant destroy magma` and built packages according to instructions [here](https://docs.magmacore.org/docs/lte/build_install_magma_pkg_in_agw). The `magma` package was installed and checkin with orc8r was successful.

## Additional Information

Removes support for Debian on AGW. For Magma v1.6 and onwards, we do not support Debian. We only support Ubuntu.


docs/readmes/howtos/README_package.md
```diff
@@ -8,9 +8,7 @@ TL;DR
 -----
 1. Running `fab test package:git` on the host creates the package inside the
 gateway VM
```
themarwhal:
```suggestion
1. Run `fab test package:git` on the host to create packages inside the
gateway VM
```
reads nicer I think

themarwhal (approved):
awesome! one small nit in the doc :)

themarwhal:
cc @pshelar for the final check :)

docs/readmes/howtos/README_package.md
```diff
@@ -51,7 +48,7 @@ you had released your package and run `apt-get install magma`.
 
 If you want to test upgrading versions, skip the part where you spin up a fresh
 prod VM, and instead set up a VM with whatever version you're trying to test
-upgrade from. For example, if you want to test upgrading from the currrently
```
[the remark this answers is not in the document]
waqaraqeel: I considered changing it to `currrrently`

pshelar:
there are few checks for distribution in https://github.com/magma/magma/blob/master/lte/gateway/release/build-magma.sh. can you also take care of those.

orc8r/gateway/python/setup.py
```diff
@@ -34,92 +34,6 @@ def os_release():
 # version this represents. 'None' defaults to 0.0.0.
 VERSION = os.environ.get('PKG_VERSION', None)
 
-release_info = os_release()
```
themarwhal, line 37: should we remove the os_release function as well?

waqaraqeel: I left it because it gives you the version as well. But if we want version-specific stuff later, we can bring it back I guess.

waqaraqeel: Removed.

docs/readmes/howtos/README_package.md
waqaraqeel: Done.

waqaraqeel:
> there are few checks for distribution in [https://github.com/magma/magma/blob/master/lte/gateway/release/build-magma.sh](https://github.com/magma/magma/blob/master/lte/gateway/release/build-magma.sh?rgh-link-date=2021-07-12T20%3A56%3A14Z). can you also take care of those.

Pushed a commit for it. The `lte-test` is currently failing as it runs on Ubuntu 16.04. It should pass once #7859 is merged.

pshelar (approved):
Thanks for working on this.

themarwhal:
One last note: can you specify why this is backwards incompatible and add notes on what actions need to be taken?

The assumption post 1.6 is that we only support Debian, but it's good to be explicit. :)

docs/readmes/howtos/README_package.md
```diff
@@ -6,11 +6,9 @@ hide_title: true
 # Packaging
 TL;DR
 -----
-1. Running `fab test package:git` on the host creates the package inside the
-gateway VM
-2. Commit changes to build-magma.sh and magma.lockfile.debian.
-3. To build packages for Ubuntu, bring up magma_focal VM. Run
-`fab dev package:vcs=git,vm=magma_focal,os=ubuntu` to build packages.
+1. Run `fab test package:git` on the host to create packages inside the
```
hcgatewood: I think the `:git` is now unnecessary cc @tmdzk (same below etc)

hcgatewood (approved):
Lgtm

waqaraqeel:
> One last note: can you specify why this is backwards incompatible and add notes on what actions need to be taken?
> 
> The assumption post 1.6 is that we only support Debian, but it's good to be explicit. :)

Done. I'm not aware of what actions should be taken though... Do we mean just switch to Ubuntu? Is there a version assumption as well? Ubuntu 20.04 onwards only?

waqaraqeel:
> Pushed a commit for it. The `lte-test` is currently failing as it runs on Ubuntu 16.04. It should pass once #7859 is merged.

cc: @tmdzk

waqaraqeel:
@rdefosse This PR also upgrades the circleci `lte-test` image from Ubuntu 16.04 to Ubuntu 20.04. The upgrade was decided and attempted in #7859, but that one wasn't passing tests because of a missing dependency. It required an updated `magma.lockfile.ubuntu` that this PR includes. The two PRs are mutually dependent :)

rdefosse (approved):

themarwhal (approved):
Thanks for fixing the failing job as well😍🤩🥳

PR #8003, opened 2021-07-13
feat(sessiond): Add new shard representation class for UEs in Sessiond

veshkemburu:
## Summary

Polling all stats at once is inefficient and leads to issues on the pipelined side
We can’t do variable polling rates for different UEs as there is currently no grouping that can identify a bunch of UEs. 

An approach is to create a shard class that stores the global state of all shards, basically the frequency of UEs per shard ID. We need to be able to dynamically add, find, and remove these UEs. Load balancing will be incorporate later on.

This is the link to the PR for the design doc related to this feature: [https://github.com/magma/magma/pull/8087](https://github.com/magma/magma/pull/8087)

Note: Some of the changes are from another PR that needs to be merged.

## Test Plan

Ran make run in vagrant VM on magma/lte/gateway
Added UEShard files to CMakeFiles.txt
There is no need for a unit test for this PR since the skeleton is fairly simple and just keeps track of global shard state. Once further API calls are added interacting with this shard class, unit tests will need to be written verifying the proper flow.

## Additional Information


lte/gateway/c/session_manager/UEShard.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+UEShard::UEShard() {
+  number_of_shards = 0;
+}
+
+// returns shard_id associated with UE added
+int UEShard::add_ue(std::string imsi) {
+  int shard_id;
+  if (shards.size() == 0) {
+    shard_id = 1;
```
uri200: Unless necessary, I recommend to start with 0. It will make your life easier

uri200: Also, maybe explore the option of using an array/vector and initialize each shard, instead of use a dictionary. You can save the need for a key.

lte/gateway/c/session_manager/UEShard.cpp
```diff
@@ -0,0 +1,94 @@
...
+        shards[shard_id].push_back(imsi);
+        return shard_id;
+      }
+    }
+    shard_id = ++number_of_shards;
+    shards[shard_id].push_back(imsi);
+  }
+  return shard_id;
+}
+
```
uri200: Shards are normally used to spread the load among them. Per what I understand, this algorithm will fill a shard completely before moving to the next shard. So that means, if you have 100 UEs, all UEs will fall in the same shard and all the load will be on that shard. Is this intentional?

If not, maybe a good way to distribute the load is to use a uniform hash like MD5 of the IMSI (or any other uniform distributed algorithm that doesn't take a hit on performance)

lte/gateway/c/session_manager/UEShard.cpp
```diff
@@ -0,0 +1,94 @@
...
+  if (shard_id <= 0 || shard_id > number_of_shards) {
+    return 0;
+  }
+  if (total_ues_for_shard(shard_id) == MAX_SHARD_SIZE) {
+    return 0;
+  }
+  remove_ue(imsi);
+  shards[shard_id].push_back(imsi);
+  return shard_id;
+}
```
uri200: moving elements from shards is normally not recommended since require lots of effort and sometimes moving doesn't improve the performance at all if not done using statistics of other shards. What will be the use case of this?

uri200 (commented):
Probably you are working on this, but I would add some tests.

I also would create a test generating some UEs and polling them randomly. Then get what the stats per shard to see if any shard is queried more often than others
(Note i meant u it test)

lte/gateway/c/session_manager/UEShard.h
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+#include <string>
+#include <utility>
+#include <vector>
+#include <map>
+
+namespace magma {
+
+class UEShard {
```
amarpad: I think this change needs a C++ face lift also I will not start the class with private.

lte/gateway/c/session_manager/UEShard.h
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+#include <string>
+#include <utility>
+#include <vector>
+#include <map>
+
+namespace magma {
+
```
amarpad: Class needs comments

lte/gateway/c/session_manager/UEShard.h
```diff
@@ -0,0 +1,42 @@
...
+  std::map<int, std::vector<std::string>> shards;
+  int number_of_shards;
+  int max_shard_size;
+
+  UEShard();
+
+  // returns shard_id associated with UE added
+  int add_ue(std::string imsi);
+
+  std::pair<int, int> find_ue_shard(std::string imsi);
```
amarpad: All the methods need comments, also pass with ref and const in this case as imsi is immutable.

lte/gateway/c/session_manager/UEShard.cpp
amarpad: +1, if you really want to loadbalance then do a ring or something so you can move a hash bucket vs the entire set of UEs 1 at a time.

amarpad (changes requested):
The sub commit messages I think can just be squashed.

lte/gateway/c/session_manager/UEShard.cpp
veshkemburu: Yeah the move isn't necessary yet(I think I'll make load balancing a separate PR later down the line). The main idea was to just have some of the skeleton for basic functionality.

lte/gateway/c/session_manager/UEShard.cpp
veshkemburu: It's intentional. The idea according to Nick was to have a fixed size for each shard(like 100), fill up that shard for insertion, and once it's full to go to the next shard. I think right now it's just meant to be simple and I'll incorporate rebalancing down the line

lte/gateway/c/session_manager/UEShard.h
veshkemburu: What is a C++ face lift?

lte/gateway/c/session_manager/UEShard.cpp
veshkemburu: Array/vector definitely makes sense, the hashmap isn't necessary if the ids for shards are as such. Good point.

lte/gateway/c/session_manager/UEShard.cpp
uri200: @koolzz probably you already shared this it the past but can you provide more insgith on the need of this sharding please?
- Will sessiond send a list of all the UEs in the shard every time we request update for that shard? Or will pipelined know what UEs are included on that shard?
- What is the trigger to request an update for shard A and not B?
- Why can't we just poll those subcribers independently based on current state of the quota (if over 80% we hyper poll them). 
- What is the final objective? Poll ovs  by smaller groups of IMSI? 

Probably I am missing some context, but I am concerned because it looks like that the intention is to poll smaller groups of UEs. This works only for very specific cases. But on big sets you may end up in situations where you need to poll at least one UE in each shard all the time. Which means you will be polling all the shards. That would be as expensive as polling all the UEs as we are doing so far

lte/gateway/c/session_manager/UEShard.h
amarpad: Just more C++ best practices friendly

lte/gateway/c/session_manager/UEShard.cpp
veshkemburu: Answers from Pravin: 
1. sessionD sends shard ID to pipelineD to retrieve stats for all UEs in the shard
2. the trigger is a request for updated stats for any UE in the shard(to differentiate among multiple shards)
3. querying each individual UE generates too much overhead
The reason to introduce sharding as solution is to scale UP the system, with multiple shards you can process each shard independently. sessionD does not need multiple threads, but pipelineD can make use of multiple requests for using async IO more effectively

uri200: Thanks. 

I think a change like this that increases complexity of the system requires a proper reason and  design document (no need to be big document, just show the call flow and what is the intention of this algorithm on the long term). It would be great to see a diagram of how this system is expected to evolve when running. 

before starting any coding we should discuss a proper load balance strategy that can help to get better performance. If there is no clear strategy, sharding will become as bad as querying all UEs.

lte/gateway/c/session_manager/UEShard.cpp
```diff
@@ -0,0 +1,78 @@
...
+      if (total_ues_for_shard(i) < max_shard_size) {
+        shards[i].push_back(imsi);
+        return i;
+      }
+    }
+    shard_id = number_of_shards++;
+    shards.push_back({imsi});
+  }
+  return shard_id;
+}
```
pshelar: rather than keeping imsi list per shard, it would be better to keep the shard ID in session-state for each imsi. since the shard-id needs to be persistent. it would save another persistent map in sessionD.

pshelar: once we move the shard-id to session-state, UEShard needs to keep track of shard level info, e.g. count of UEs in the shard.

lte/gateway/c/session_manager/UEShard.cpp
pshelar: I think this feature is documented in google doc, @veshkemburu can you post a PR for design doc for this feature in markdown format.

uri200: Please, add some diagrams of the call flow with differnt UEs and different polling frequencies to understand what will be the benefit of this.

I think it would be good to put in a diagram how you expect the system behave (with load balancing included, even you don't have the algo)

veshkemburu: I'll have the design doc pushed in a separate PR.

lte/gateway/c/session_manager/UEShard.cpp
```diff
@@ -0,0 +1,52 @@
...
+int UEShard::add_ue() {
+  int shard_id;
+  if (shards.size() == 0) {
+    shard_id = 0;
+    shards.push_back(1);
+    number_of_shards++;
+  } else {
+    for (auto i = 0; i < number_of_shards; i++) {
+      if (shards[i] < MAX_SHARD_SIZE) {
+        shards[i]++;
```
koolzz: we might need to actually store imsi here for easier subscriber reallocation but for now this seems good to merge

koolzz (approved):

lte/gateway/c/session_manager/CMakeLists.txt
```diff
@@ -152,6 +152,8 @@ add_library(SESSION_MANAGER
     CreditKey.h
     StatsPoller.cpp
     StatsPoller.h
+    UEShard.cpp
```
themarwhal: `ShardTracker` maybe?

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }
+    shard_id = number_of_shards++;
+    shards.push_back(1);
+  }
+  return shard_id;
+}
+
+void ShardTracker::remove_ue(int shard_id) {
+  shards[shard_id]--;
```
themarwhal: I might be missing something but how are we guaranteeing that the shard_id index is available?

lte/gateway/c/session_manager/UEShard.cpp
themarwhal: Let's add a TODO here so that the intent is clear here :) Extra bonus points if you file an issue and link it in the TODO

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,38 @@
...
+
+namespace magma {
+
+// Shards represent groups of UEs placed into buckets of
+// a certain size, to make polling more manageable
+class ShardTracker {
+  ShardTracker();
+
+  // add UE to shards based on availability
+  int add_ue();
```
themarwhal, line 30: Add doxygen style docs for all these functions
```
  /**
   * FILL IN WITH DESCRIPTION
   * @param paramA description
   * @return description
   */
```

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,38 @@
...
+
+  // remove UE from shard
+  void remove_ue(int shard_id);
+
+ private:
+  std::vector<int> shards;
```
themarwhal: Can we add comments on what this contains? What is the int in this case? Similarly we could probably use a better variable name than shards here. 

The PR description says `An approach is to create a shard class that stores a map of shard IDs to a vector of UEs with unique IMSIs`, but I can't tell if that's outdated or not.

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,38 @@
...
+  int max_shard_size;
```
themarwhal: Add a doc on how this is initialized

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,38 @@
...
+  int number_of_shards;
```
themarwhal: maybe use uint? unless the value can be negative :)

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "ShardTracker.h"
+#define MAX_SHARD_SIZE 100
+
+namespace magma {
+
+ShardTracker::ShardTracker() {
+  number_of_shards = 0;
+}
+
+int ShardTracker::add_ue() {
```
themarwhal, line 21: This function seems to be doing a lot. Can we add some comments to make it easier to read?

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <map>
+
+#include "ShardTracker.h"
+#define MAX_SHARD_SIZE 100
+
+namespace magma {
+
+ShardTracker::ShardTracker() {
+  number_of_shards = 0;
+}
```
themarwhal:
```suggestion
ShardTracker::ShardTracker():number_of_shards(0) {}
```

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,38 @@
...
+class ShardTracker {
+  ShardTracker();
+
+  // add UE to shards based on availability
+  int add_ue();
+
+  // remove UE from shard
+  void remove_ue(int shard_id);
+
+ private:
```
themarwhal, line 39: All class variables should have an underscore at the end. https://google.github.io/styleguide/cppguide.html#General_Naming_Rules
See under Class Data Members

themarwhal (changes requested):
Left some comments! If there is a proposal PR/doc, can we link it in the descriptions if it's not already?

lte/gateway/c/session_manager/ShardTracker.cpp
veshkemburu: So the shard_id index is available based on the number of UEs it has. I'll add a comment for this, but every time the UE is added a check is made for the number of UEs at a particular index before addition.

lte/gateway/c/session_manager/ShardTracker.h
veshkemburu: Thanks for the PR description catch, that should definitely be updated.

lte/gateway/c/session_manager/UEShard.cpp
veshkemburu: Will do, I'll file an issue as well.

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "ShardTracker.h"
+#define MAX_SHARD_SIZE 100
+
+namespace magma {
+
+ShardTracker::ShardTracker() : number_of_shards_(0) {}
+
+int ShardTracker::add_ue() {
+  int shard_id;
+  //If we have no shards, we add an entry for the shard representing
```
themarwhal: Should be able to get rid of all these by running `make format_session_manager`

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,52 @@
...
+  // TODO: Store IMSI as well for easier subscriber reallocation
+  int add_ue();
+
+  /**
+   * remove UE from shard
+   * @param shard_id location of UE to be removed
+   */
+  void remove_ue(int shard_id);
+
+ /* shards: a vector of quantities, where the indices represent
```
themarwhal: let's put the comment right above the variable declaration

so something like
```c++
private:
  // a vector of quantities, where the indices represent ...
  std::vector<uint16_t> shards_;
  // largest number of UEs that can...
  uint16_t max_shard_size_;
  // number of shards there exists in the system...
  uint16_t number_of_shards_;
```

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,52 @@
...
+/*Shards represent groups of UEs placed into buckets of
+a certain size, to make polling more manageable*/
+class ShardTracker {
+  ShardTracker();
+
+  /**
+   * add UE to shards based on availability
+   * @return index(shard id) where UE was placed
+   */
+  // TODO: Store IMSI as well for easier subscriber reallocation
```
themarwhal: Let's address this TODO lint

themarwhal: If you have a GH issue we can do
```c++
// TODO(veshkemburu): Store IMSI as well for easier subscriber reallocation (GHXXXX)
```

veshkemburu: Added issue: #8167

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace magma {
+
+ShardTracker::ShardTracker() : number_of_shards_(0) {}
+
+int ShardTracker::add_ue() {
+  int shard_id;
+  // If we have no shards, we add an entry for the shard representing
+  // of 1 representing a single UE at the new shard*/
+  if (shards_.size() == 0) {
```
themarwhal: Let's return early to clean up the logic
```suggestion
  if (shards_.size() == 0) {
    shards_.push_back(1);
    number_of_shards_++;
    return 0;
  }
```

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "ShardTracker.h"
+#define MAX_SHARD_SIZE 100
+
+namespace magma {
+
+ShardTracker::ShardTracker() : number_of_shards_(0) {}
+
+int ShardTracker::add_ue() {
+  int shard_id;
+  // If we have no shards, we add an entry for the shard representing
```
themarwhal: omit we to make the comment shorter
```suggestion
  // If there no shards, add an entry for the shard representing
```

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,57 @@
...
+  // of 1 representing a single UE at the new shard*/
```
themarwhal: remove trailing */

themarwhal: Also not sure what this comment/sentence means, can you clarify?

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,57 @@
...
+  if (shards_.size() == 0) {
+    shard_id = 0;
+    shards_.push_back(1);
+    number_of_shards_++;
+  } else {
+    // We iterate through all existing shards, if all are full
+    // we create a new shard with quantity 1, otherwise we increment
+    // the quantity of UEs in the latest shard
```
themarwhal: Omit we
```suggestion
    // Iterate through all existing shards. If all are full,
    // create a new shard with quantity 1. Otherwise, we increment
    // the quantity of UEs in the latest shard
```

themarwhal (changes requested):
thanks for addressing earlier comments! Left a few more to ask for clarifications on some docs

themarwhal:
cc @amarpad for re-review

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+#include <string>
+#include <utility>
+#include <vector>
+#include <algorithm>
+#include <map>
+
+#include "ShardTracker.h"
+#define MAX_SHARD_SIZE 100
```
amarpad: nit: use consts in C++ also this can be a part of the class as that encapsulates the entire behavior of the class.

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,51 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <string>
+#include <utility>
+#include <vector>
+#include <map>
```
amarpad: Please ensure headers are all used.

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,51 @@
...
+
+namespace magma {
+
+/*Shards represent groups of UEs placed into buckets of
```
amarpad: Please ensure spacing is accurate for comments.

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,51 @@
...
+a certain size, to make polling more manageable*/
+class ShardTracker {
+  ShardTracker();
+
+  /**
+   * add UE to shards based on availability
```
amarpad: s/add/Add, same for remove below.

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,51 @@
...
+  int add_ue();
+
+  /**
+   * remove UE from shard
+   * @param shard_id location of UE to be removed
+   */
+  void remove_ue(int shard_id);
+
+ private:
+  // a vector of quantities, where the indices represent
```
amarpad: consistently use block comments or multiline comments, I think we shouldn't mix and match them.

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,51 @@
...
+  // the shard id and the values represent the number of
+  // UEs held in each shard
+  std::vector<uint16_t> shards_;
+  // largest number of UEs that can fill a shard
+  uint16_t max_shard_size_;
```
amarpad: unused?

lte/gateway/c/session_manager/ShardTracker.cpp
amarpad: +1 also please fix the grammar.

lte/gateway/c/session_manager/ShardTracker.cpp
amarpad: size() = 0 -> empty(), standard guarantees constant time for empty, so its stylistically preferred. +1 on return early pattern

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,57 @@
...
+  // of 1 representing a single UE at the new shard*/
+  if (shards_.size() == 0) {
+    shard_id = 0;
+    shards_.push_back(1);
+    number_of_shards_++;
+  } else {
+    // We iterate through all existing shards, if all are full
+    // we create a new shard with quantity 1, otherwise we increment
+    // the quantity of UEs in the latest shard
+    for (auto i = 0; i < number_of_shards_; i++) {
```
amarpad: stylistic -> for (auto& shard: shards_)

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,57 @@
...
+      if (shards_[i] < MAX_SHARD_SIZE) {
+        shards_[i]++;
```
amarpad: This seems like a suboptimal way for sharding right? You will always fill up the buckets with lower indices first. This seems like bin packing to me. If so we should use a different name for the class.
If your objective is to minimize the number of calls to flows in pipelined then I would definitely call it bin packing. If the objective is to truly shard then a vector is the wrong datastructure.

amarpad (changes requested):

amarpad:
@veshkemburu
1 - I have two other requests, please squash the git history a lot of the subcommits are not relevant.
2 - Please self review before posting for public review, some of the stylistic issues come across as a lack of craftsmanship/attention to detail and are easily fixable before posting.

koolzz:
> @veshkemburu
> 1 - I have two other requests, please squash the git history a lot of the subcommits are not relevant.
> 2 - Please self review before posting for public review, some of the stylistic issues come across as a lack of craftsmanship/attention to detail and are easily fixable before posting.

Agree with style comments but, what's the benefits of squashing subcommits? We squash in the end anyway when we merge

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,51 @@
...
+  // the number of shards we currently have
+  uint16_t number_of_shards_;
```
themarwhal: Is `number_of_shards_` always equal to `shards_.size()`? Or do we want this number to omit  some of 0 values in the shards_ vector? 

If we just want the number of shards that exist, maybe we can just use the size of shards_.
If not, we should clarify it to be like `number_of_shards_with_ues_`?

lte/gateway/c/session_manager/ShardTracker.cpp
koolzz: Yes, this is just initial implementation. We want to get some semblance of it merged and working and then we'll move onto better allocation and moving UEs around. Iterative approach

amarpad:
> > @veshkemburu
> > 1 - I have two other requests, please squash the git history a lot of the subcommits are not relevant.
> > 2 - Please self review before posting for public review, some of the stylistic issues come across as a lack of craftsmanship/attention to detail and are easily fixable before posting.
> 
> Agree with style comments but, what's the benefits of squashing subcommits? We squash in the end anyway when we merge

The commit message will be cluttered with a lot of one line inputs otherwise. It's just also cleaner practice no?

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,51 @@
...
+   * remove UE from shard
+   * @param shard_id location of UE to be removed
+   */
+  void remove_ue(int shard_id);
+
+ private:
+  // a vector of quantities, where the indices represent
+  // the shard id and the values represent the number of
+  // UEs held in each shard
+  std::vector<uint16_t> shards_;
```
themarwhal: `ue_count_per_shard_` is clearer I think. The X_per_Y / X_by_Y naming scheme is often pretty useful for maps. (This var is technically a vector but it serves like a map)

lte/gateway/c/session_manager/ShardTracker.h
themarwhal: I think `max_shard_size_` sounds like the max size for the `shards_` vector. We could rename it to something more explicit like `max_ue_count_per_shard_`. This one I feel less strongly about than the comment above regarding `shards_` vector though. If we rename the vector, this variable name isn't as confusing.

koolzz:
> > > @veshkemburu
> > > 1 - I have two other requests, please squash the git history a lot of the subcommits are not relevant.
> > > 2 - Please self review before posting for public review, some of the stylistic issues come across as a lack of craftsmanship/attention to detail and are easily fixable before posting.
> > 
> > 
> > Agree with style comments but, what's the benefits of squashing subcommits? We squash in the end anyway when we merge
> 
> The commit message will be cluttered with a lot of one line inputs otherwise. It's just also cleaner practice no?

We can always edit the commit message during merge, I think we should have a general guideline if we want to enforce some specific style

amarpad:
> > > > @veshkemburu
> > > > 1 - I have two other requests, please squash the git history a lot of the subcommits are not relevant.
> > > > 2 - Please self review before posting for public review, some of the stylistic issues come across as a lack of craftsmanship/attention to detail and are easily fixable before posting.
> > > 
> > > 
> > > Agree with style comments but, what's the benefits of squashing subcommits? We squash in the end anyway when we merge
> > 
> > 
> > The commit message will be cluttered with a lot of one line inputs otherwise. It's just also cleaner practice no?
> 
> We can always edit the commit message during merge, I think we should have a general guideline if we want to enforce some specific style

I just want to make sure you are not being too religious here by throwing in the rule book :D. This is not worth us arguing. 
My read of why this is happening is that we are not familiar with the git flow of amending after rebases vs introducing new commits and longer term there is no substitute for git familiarity. This seems like an initial diff from @veshkemburu so good time to ramp and getting the toolsets right. If you think that read is wrong, then happy to let it go :D. 
Here is how I squash changes, probably takes less than a min.
git rebase -i HEAD~xx where xx is the number of commits in the patch that can be squashed
this will open an interactive rebase, replace all the change numbers by s that will pop up a squashed commit message that we close the file.

lte/gateway/c/session_manager/ShardTracker.cpp
veshkemburu: Are for range loops discouraged(didn't see this mentioned in the style guide). The index is actually needed to return the relevant shard id(mentioned in ShardTracker.h), so wouldn't changing up the for loop form just make for more code?

lte/gateway/c/session_manager/ShardTracker.h
veshkemburu: So for this, I'm using docstring style comments for function descriptions and class descriptions, and normal comments elsewhere. If that's not very clear, I can change it to a different style to be more consistent.

lte/gateway/c/session_manager/ShardTracker.h
veshkemburu: It was just meant to be the shards size, but it isn't necessary, you're right.

lte/gateway/c/session_manager/ShardTracker.cpp
amarpad: ah right you need the offset index, missed that.

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,50 @@
...
+namespace magma {
+
+int ShardTracker::add_ue() {
+  int shard_id;
+  // If there are no shards, add an entry for the shard
+  // representing a single UE at the new shard
+  if (ue_count_per_shard_.size() == 0) {
+    ue_count_per_shard_.push_back(1);
+    return 0;
+  } else {
```
themarwhal: Don't need the else clause because we're returning early above.

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,50 @@
...
+    // Iterate through all existing shards, if all are full
+    // create a new shard with quantity 1, otherwise increment
+    // the quantity of UEs in the latest shard
+    for (auto i = 0; i < ue_count_per_shard_.size(); i++) {
```
themarwhal: Since the index refers to ShardID in this implementation, let's use shard_id instead of i.
```suggestion
    for (auto shard_id = 0; shard_id < ue_count_per_shard_.size(); shard_id++) {
```

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,50 @@
...
+      if (ue_count_per_shard_[i] < max_shard_size_) {
```
themarwhal: ^ Can you resolve this build warning?
```
comparison of integer expressions of different signedness: 'int' and 'std::vector<short unsigned int>::size_type' {aka 'long unsigned int'} [-Wsign-compare]
```

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,50 @@
...
+        ue_count_per_shard_[i]++;
+        return i;
+      }
+    }
+    shard_id = ue_count_per_shard_.size();
+    ue_count_per_shard_.push_back(1);
```
themarwhal: If we assign a UE by adding a new shard, can't we just return the new shard id here? Do we need to keep going with the loop?

themarwhal: If we have 
* max_shard_size_ = 5
* shard ID 0 with 5UEs
* shard ID 1 with 1UE

This would add a new shard ID 2 with 1UE when `add_ue` is called, is that right?

veshkemburu: It's not because it would add another UE to shard ID 1(since it only has 1 UE). But this handles the case where we need to add a new shard id. Meaning all the shard ids elsewhere are already filled. So like if shard ID 0 has 5 UEs and shard ID one has 5 UEs, then we create new shard id 2, and add 1 UE.

lte/gateway/c/session_manager/ShardTracker.cpp
veshkemburu: Should have been resolved when I changed auto to size_t. Wonder why it's still here(didn't commit with new change).

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,47 @@
...
+    }
+  }
+  ue_count_per_shard_.push_back(1);
+  return ue_count_per_shard_.size() - 1;
+}
+
+void ShardTracker::remove_ue(int shard_id) {
+  // Since we only keep global state of all UEs, we just
+  // need to decrement the number of UEs at a particular id
+  ue_count_per_shard_[shard_id]--;
```
themarwhal, line 54: Could `ue_count_per_shard_[shard_id]` be 0? If not, can you leave a comment as to why this is the case?

Should we do some sanity check here that 0 <= shard_id < ue_count_per_shard_.size() ? May want to log some debug statements in case we have some weird behaviors.

themarwhal (changes requested):
Final comment! I think the rest looks good.

lte/gateway/c/session_manager/ShardTracker.cpp
veshkemburu: So it shouldn't reach 0 if there are shards. Once the last shard is removed only then should it return 0.

themarwhal (approved):
lgtm!

veshkemburu:
> > > > > @veshkemburu
> > > > > 1 - I have two other requests, please squash the git history a lot of the subcommits are not relevant.
> > > > > 2 - Please self review before posting for public review, some of the stylistic issues come across as a lack of craftsmanship/attention to detail and are easily fixable before posting.
> > > > 
> > > > 
> > > > Agree with style comments but, what's the benefits of squashing subcommits? We squash in the end anyway when we merge
> > > 
> > > 
> > > The commit message will be cluttered with a lot of one line inputs otherwise. It's just also cleaner practice no?
> > 
> > 
> > We can always edit the commit message during merge, I think we should have a general guideline if we want to enforce some specific style
> 
> I just want to make sure you are not being too religious here by throwing in the rule book :D. This is not worth us arguing.
> My read of why this is happening is that we are not familiar with the git flow of amending after rebases vs introducing new commits and longer term there is no substitute for git familiarity. This seems like an initial diff from @veshkemburu so good time to ramp and getting the toolsets right. If you think that read is wrong, then happy to let it go :D.
> Here is how I squash changes, probably takes less than a min.
> git rebase -i HEAD~xx where xx is the number of commits in the patch that can be squashed
> this will open an interactive rebase, replace all the change numbers by s that will pop up a squashed commit message that we close the file.

Your absolutely right about the style. I rushed a few of the commits yesterday, and made some careless errors. I've also looked at the C++ style guide.

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <vector>
+
+#include "ShardTracker.h"
+
+namespace magma {
+
+int ShardTracker::add_ue() {
+  // If there are no shards, add an entry for the shard
+  // representing a single UE at the new shard
+  if (ue_count_per_shard_.size() == 0) {
```
amarpad, line 26: Nit: Always use empty for containers, as the standard guarantees constant time lookup.

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <string>
+#include <vector>
+
+namespace magma {
+
+/* Shards represent groups of UEs placed into buckets of
```
amarpad, line 19: Nit fix the style
/*
 *
 */

lte/gateway/c/session_manager/ShardTracker.h
amarpad: https://google.github.io/styleguide/cppguide.html#Comment_Style

I have generally seen only one style in a file.

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -0,0 +1,48 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
```
amarpad: #pragma once?

amarpad (approved):
Couple of nits, looks good otherwise, but do address.

PR #8009, opened 2021-07-13
docs(feg): [proposal] N7 Interface and Network initiated session modi…

pmishra-wavelabs:
## Summary
This document provides an approach for the following 
    • Implementing N7 interface towards PCF through Federation Gateway
        ◦ Implement 5G Services Based Interface (SBI) for Federation Gateway
        ◦ Feg Relay for SM Policy control create, update, delete and update-notify
        ◦ Feg Session Proxy for SM Policy control create, update, delete and update-notify 
    • Implementing Network initiated session modification
        ◦ Network Initiated Session Policy association update
        ◦ Network initiated session Policy association termination
Corresponding Github Issue : https://app.zenhub.com/workspaces/magma--5g-sa-60871e99b9326f001696b1ee/issues/magma/magma/7970

## Test Plan
    • N7 interface for create and terminate session should pass with DSTester and TeraVM (PCF).
    • Network requested session modification (Policy Update) should be tested with DSTester and TeraVM (PCF).

## Additional Information

Signed-off-by: Prabhat Mishra <prabhat.mishra@wavelabs.ai>


docs/readmes/proposals/p020_n7_Interface_n_network_initiated_session_modification.md
```diff
@@ -0,0 +1,165 @@
...
+*Author: @pmishra-wavelabs*\
+*Last Updated: 07/12/2021*
+
+## Background and Objective 
+This document provides an approach for the following 
+- Implementing N7 interface towards PCF through Federation Gateway
+  -- Implement 5G Services Based Interface (SBI) for Federation Gateway
+  -- Feg Relay for SM Policy control create, update, delete and update-notify
+  -- Feg Session Proxy for SM Policy control create, update, delete and 
+     update-notify
```
uri200, line 17: Session_proxy is tight to diameter so at a first sight I don't think this should be implemented on session_proxy. Probably this will require a brand new service at the feg.

How much PCEF and PCF share on its messaging?

docs/readmes/proposals/p020_n7_Interface_n_network_initiated_session_modification.md
```diff
@@ -0,0 +1,165 @@
...
+## Background and Objective 
+This document provides an approach for the following 
+- Implementing N7 interface towards PCF through Federation Gateway
+  -- Implement 5G Services Based Interface (SBI) for Federation Gateway
+  -- Feg Relay for SM Policy control create, update, delete and update-notify
+  -- Feg Session Proxy for SM Policy control create, update, delete and 
+     update-notify 
+- Implementing Network initiated session modification
+  -- Network Initiated Session Policy association update
+  -- Network initiated session Policy association termination
```
uri200, line 20: Note FEG is just a translator from GRPC into 3gpp world. session_proxy this translator for the Gx/Gy interface for LTE so it is directly tight to diameter. I think N7 will require a brand new service

Does pcrf and PCF share any kind of messaging protocol/structure?

Are the messages sent from sessiond to FEG exactly the same for LTE and 5g?

docs/readmes/proposals/p020_n7_Interface_n_network_initiated_session_modification.md
```diff
@@ -0,0 +1,165 @@
...
+## Use Cases
```
uri200, line 21: is network_initiated_session_modification_sequence_diagram the only call flow this feature will support?
If not, then we need `create, update, delete and update-notify` call flows too.  Service definition should be planned based on all callflows to implement

docs/readmes/proposals/p020_n7_Interface_n_network_initiated_session_modification.md
```diff
@@ -0,0 +1,165 @@
...
+SBI is a HTTP/2 protocol based interface. SBI is composed of both HTTP/2 client 
+and server. The client is used to send requests and the server is used to 
+receive callbacks and notifications. This SBI is part of Feg Session Proxy
+for the N7 interface.
+
+#### Session Proxy
+Session Proxy is reused for interfacing with the PCF through N7 interface. 
+The CentralSessionController GRPC service implementation in the Session Proxy 
+invokes the N7 SBI requests when M5GSMSessionContext is received in the 
+Create/Update/Terminate session requests.
```
uri200, line 58: grpc server is an interface in go, so no need to modify session_proxy. A new service can implement the same methods as session_proxy and receive the same messages.

emakeev: Magma is microservice architecture & session_proxy is already pretty complicated & hardly "micro". If a significant new logic/functionality is required - I'd suggest implementing it in a new service.

uri200 (commented):
per previous experiences, adding a new service to feg we need to:
- Understand the protocol (grpc and rest API)
- Define callflows clearly (each of them)
- Define needed grpc messages and its content
- Define needed grpc servers

Also, if this moves forward I would recommend to start a google docs to facilitate collaboration (same we did for S8 interface on inbound roaming https://github.com/magma/magma/blob/master/docs/readmes/proposals/p008_inbound_roaming_with_subscriberdb.md). I can create an empty document on our gdocs repo once this proposal is more defined

hcgatewood (commented):
Clearing review queue. Deferring to @uri200. P.S. agreed that a design doc in e.g. Google Docs would be a better way to quickly iterate on this doc

ulaskozat:
@hcgatewood @karthiksubraveti @uri200 @emakeev since N7 interface is a restful interface, shall we rather expose these as NB APIs from the orc8r?

uri200:
> @hcgatewood @karthiksubraveti @uri200 @emakeev since N7 interface is a restful interface, shall we rather expose these as NB APIs from the orc8r?

Hey @ulaskozat  the reason of having FEG gateway is that it provide us with the ability to have our 3gpp  services next to the core services without worrying about network configuration (nats, firewalls, proxies).

If the or8rc is not in prem, then operator will have to expose ports to access to those interfaces.

Orc8r can talk to the FEG even if it is behind a firewall.

To me being diameter or restful Api doesn't make any difference. We still have to convert the grpc messages into the 3gpp format (either diameter or rest). 

This also kept me thinking those days: why do we need to maintain grpc anymore if 5g will work with rest?

ulaskozat:
> > @hcgatewood @karthiksubraveti @uri200 @emakeev since N7 interface is a restful interface, shall we rather expose these as NB APIs from the orc8r?
> 
> Hey @ulaskozat the reason of having FEG gateway is that it provide us with the ability to have our 3gpp services next to the core services without worrying about network configuration (nats, firewalls, proxies).
> 
> If the or8rc is not in prem, then operator will have to expose ports to access to those interfaces.
> 
> Orc8r can talk to the FEG even if it is behind a firewall.
> 
> To me being diameter or restful Api doesn't make any difference. We still have to convert the grpc messages into the 3gpp format (either diameter or rest).
> 
> This also kept me thinking those days: why do we need to maintain grpc anymore if 5g will work with rest?

@uri200 I guess it makes sense to move "non-generic" technology specific Restful APIs over to FEG? As there was a legitimate push back by the Orc8r team. Maybe what we need is just a matter of figuring out where this FEG micro-service be sitting?

ulaskozat:
@pmishra-wavelabs @sureshgorijavolu (cc @uri200, @andreilee, @themarwhal), I think we are jumping too much ahead about integrating with a third party PCF. Initially, I suggest to focus on fetching session policies from policydb. As a prerequisite, I would like to see why and how you plan to augment the existing policy API at Orc8r as well as why and how you want to change the interfacing/interaction between sessiond and policydb.

uri200:
Note: I typed this before seeing your last reply @ulaskozat 

Hey @ulaskozat I don't know much about this interface but if this is just another interface, the service should be sitting with the other FEG services. Same way we added s8_proxy we could add n7_proxy.

But note this is a new 5g service that uses a total brand new interface we have never implemented. It would be good to do a proper design so we can set the standard for future 5g interfaces.

This also brings to me an existential question of the reason to use GRPC for 5g when 5g already have a flat protocol (for 4g it made sense because we had lots of different protocols).

electronjoe:
Where are we at with regards to this proposal? @pmishra-wavelabs 

@others the proposal process as documented involves committing a design doc (like this) with only review being for typos. Then discussion on the GH Issue.

electronjoe (approved):

electronjoe:
@pmishra-wavelabs if you want to merge this design doc for discussion, add the #ready2merge label and I'll do so.

Edit: there is in fact one typo to fix per linter findings:

`"propogate" is a misspelling of "propagate"`

Please address for merge.

@aswin-alexander FYI

hcgatewood:
@pmishra-wavelabs @electronjoe @ulaskozat this proposal is 22 days stale. If it's accepted please merge, otherwise where are we at with it

ulaskozat:
@pmishra-wavelabs can you update the proposal and diagrams such that a new micro-service on FEG side is used rather than session_proxy? Note that protos and rpc calls can be still reused between 5G thread on sessiond and this new N7 Proxy service on FEG side. cc @uri200

ulaskozat:
@hcgatewood this proposal is actually discussed and needs to go through changes. cc @uri200

hcgatewood:
> @hcgatewood this proposal is actually discussed and needs to go through changes. cc @uri200

@ulaskozat sounds good, just making sure it didn't fall through the cracks 🙂

hcgatewood:
> @hcgatewood this proposal is actually discussed and needs to go through changes. cc @uri200

@ulaskozat sounds good, just making sure it didn't fall through the cracks 🙂

uri200:
@pmishra-wavelabs It would be good to think how to make the FEG part extendeable. Meaning, we should create a server that takes swagger API and becomes N7. That server should be reusable with other set of swagger definition to create other interfaces.

The N7 feg service should be just translating GRPC into swagger specs and api calls.

GANESH-WAVELABS:
The updated proposal is maintained in the following path as a Google Doc - https://docs.google.com/document/d/18umXHz4OBNh2XhNqEfa6G3ObBJ1nC2M1aMWtZS-jkGE/edit#heading=h.czv0vpn4mw66.

@ulaskozat the above document has the updates your have requested for.

@uri200 the "Service Based Interface" heading in the above document mentions about the extendable approach.

ekowtaylor:
I'll go ahead and close this proposal as the doc will be the live document moving forward

PR #8016, opened 2021-07-13
feat(nms): Gateway Detail Page Displaying Gateway Status

ato-dani:
## Summary
This is stacked PR upon the open PR #8015, and therefore is a Draft PR until the other PR gets merged(Only last commit represents changes).  Gateway Detail Status component was added to display different information about the status of the gateway like its health, last checkin time and the like. 

## Test Plan
A test file named GatewayDetailStatusTest was added to test that the gateway status is displayed correctly. Also previous tests were run and all run without errors. 

## Additional Information

- [ ] This change is backwards-breaking

Screen Shot 2021-07-13 at 11 29 18 AM

PR #8018, opened 2021-07-13
feat(nms): Gateway Detail Page Displaying Managed Subscribers

### LINKED ISSUE #2842: Add Federation Equipment dashboard and Gateway overview page

### LINKED ISSUE #2844: Federation gateway detail page
https://www.figma.com/proto/aRJzvSa3J9nhD2n3yERd1s/Magma---Phase-2-(Updated-Styles)?node-id=1035%3A1754&scaling=min-zoom

ato-dani:
## Summary
This is stacked PR upon the open PR #8016. The managed subscriber table was added to the gateway detail page which shows the subscribers managed by the federation network. The subscribers come from the subscribers of the federated_lte networks serviced by the federation network.

## Test Plan
A test file named FEGGatewayDetailSubscribers.js to test the subscribers table. Also, all previous tests run without error. 

## Additional Information

Screen Shot 2021-07-13 at 1 33 14 PM

Screen Shot 2021-07-28 at 3 09 13 PM (2)

### DIFF
--- a/nms/app/packages/magmalte/app/components/FEGServicingAccessGatewayKPIs.js
+++ b/nms/app/packages/magmalte/app/components/FEGServicingAccessGatewayKPIs.js
@@ -37,7 +37,7 @@ import {useRouter} from '@fbcnms/ui/hooks';
  */
 export async function getServicedAccessNetworks(
   federationNetworkId: network_id,
-  enqueueSnackbar: (
+  enqueueSnackbar?: (
     msg: string,
     cfg: EnqueueSnackbarOptions,
   ) => ?(string | number),
--- a/nms/app/packages/magmalte/app/components/context/FEGSubscriberContext.js
+++ b/nms/app/packages/magmalte/app/components/context/FEGSubscriberContext.js
@@ -1,0 +1,34 @@
+/**
+ * Copyright 2020 The Magma Authors.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @flow strict-local
+ * @format
+ */
+
+import type {
+  network_id,
+  subscriber_id,
+  subscriber_state,
+} from '@fbcnms/magma-api';
+
+import React from 'react';
+
+export type FEGSubscriberContextType = {
+  sessionState: {
+    [networkId: network_id]: {[subscriberId: subscriber_id]: subscriber_state},
+  },
+  setSessionState: (newSessionState: {
+    [string]: {[string]: subscriber_state},
+  }) => void,
+};
+
+export default React.createContext<FEGSubscriberContextType>({});
--- a/nms/app/packages/magmalte/app/components/context/RefreshContext.js
+++ b/nms/app/packages/magmalte/app/components/context/RefreshContext.js
@@ -20,6 +20,7 @@ import {
   getActiveFegGatewayId,
   getFegGatewaysHealthStatus,
 } from '../../state/feg/EquipmentState';
+import {FetchFegSubscriberState} from '../../state/feg/SubscriberState';
 import {FetchSubscriberState} from '../../state/lte/SubscriberState';
 import {useContext, useEffect, useRef, useState} from 'react';
 import type {EnqueueSnackbarOptions} from 'notistack';
@@ -30,6 +31,7 @@ export const RefreshTypeEnum = Object.freeze({
   SUBSCRIBER: 'subscriber',
   GATEWAY: 'gateway',
   FEG_GATEWAY: 'feg_gateway',
+  FEG_SUBSCRIBER: 'feg_subscriber',
   ENODEB: 'enodeb',
 });
 
@@ -58,7 +60,10 @@ export function useRefreshingContext(props: Props) {
         health: ctx?.health,
         activeFegGatewayId: ctx?.activeFegGatewayId,
       };
-    } else if (props.type === 'subscriber') {
+    } else if (
+      props.type === RefreshTypeEnum.SUBSCRIBER ||
+      props.type === RefreshTypeEnum.FEG_SUBSCRIBER
+    ) {
       return {sessionState: ctx.sessionState};
     }
     return ctx.state;
@@ -108,6 +113,8 @@ export function useRefreshingContext(props: Props) {
           // $FlowIgnore because state may not contain fegGateways for other refresh type like subscriber
           fegGateways: {...ctx.fegGateways, [id]: state?.fegGateways?.[id]},
         };
+      } else if (props.type === RefreshTypeEnum.FEG_SUBSCRIBER) {
+        newState = {...ctx.sessionState};
       } else {
         newState = {...ctx.state, [id]: state?.[id]};
       }
@@ -117,6 +124,8 @@ export function useRefreshingContext(props: Props) {
       return ctx.setState(null, null, null, newState);
     } else if (props.type === RefreshTypeEnum.FEG_GATEWAY) {
       return ctx.setState(null, null, newState?.fegGateways || {});
+    } else if (props.type === RefreshTypeEnum.FEG_SUBSCRIBER) {
+      return ctx.setSessionState(newState);
     }
     return ctx.setState(null, null, newState);
   }
@@ -204,6 +213,12 @@ async function fetchRefreshState(props: FetchProps) {
       getActiveFegGatewayId(networkId, fegGateways, enqueueSnackbar),
     ]);
     return {fegGateways, health, activeFegGatewayId};
+  } else if (type === RefreshTypeEnum.FEG_SUBSCRIBER) {
+    const sessions = await FetchFegSubscriberState({
+      networkId,
+      enqueueSnackbar,
+    });
+    return {sessionState: sessions};
   } else {
     const enodebs = await FetchEnodebs({
       id: id,
--- a/nms/app/packages/magmalte/app/state/feg/SubscriberState.js
+++ b/nms/app/packages/magmalte/app/state/feg/SubscriberState.js
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2020 The Magma Authors.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @flow strict-local
+ * @format
+ */
+
+import type {EnqueueSnackbarOptions} from 'notistack';
+import type {network_id} from '@fbcnms/magma-api';
+
+import {FetchSubscriberState} from '../lte/SubscriberState';
+import {getServicedAccessNetworks} from '../../components/FEGServicingAccessGatewayKPIs';
+
+/**
+ * Props passed when fetching subscriber state.
+ *
+ * @param {network_id} networkId Id of the federation network.
+ * @param {(msg, cfg,) => ?(string | number),} enqueueSnackbar Snackbar to display error.
+ */
+type FetchProps = {
+  networkId: network_id,
+  enqueueSnackbar?: (
+    msg: string,
+    cfg: EnqueueSnackbarOptions,
+  ) => ?(string | number),
+};
+
+/**
+ * Fetches and returns the subscriber session state of all the serviced
+ * federated lte networks under by this federation network.
+ *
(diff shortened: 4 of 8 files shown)


nms/app/packages/magmalte/app/components/FEGServicingAccessGatewayKPIs.js
```diff
@@ -37,7 +37,7 @@ import {useRouter} from '@fbcnms/ui/hooks';
  */
 export async function getServicedAccessNetworks(
   federationNetworkId: network_id,
-  enqueueSnackbar: (
+  enqueueSnackbar?: (
```
ato-dani, line 40: Made this optional because we need to call this function from the useRefreshingContext and in that function enqueueSnackbar is optional which forces us to adhere to that. I think the enqueueSnackbar in useRefreshingContext was made optional so as to don't make errors noisy sometimes.

nms/app/packages/magmalte/app/state/feg/SubscriberState.js
```diff
@@ -0,0 +1,94 @@
...
+};
+
+/**
+ * Fetches and returns the subscriber session state of all the serviced
+ * federated lte networks under by this federation network and whose
+ * subscriber session is not managed by the HSS.
+ *
+ * @param {FetchProps} props an object containing the network id and snackbar to display error.
+ */
+export async function FetchFegSubscriberState(props: FetchProps) {
```
andreilee: Annotate with return type

nms/app/packages/magmalte/app/state/feg/SubscriberState.js
```diff
@@ -0,0 +1,94 @@
...
+  const {networkId, enqueueSnackbar} = props;
+  const servicedAccessNetworks = await getServicedAccessNetworks(
+    networkId,
+    enqueueSnackbar,
+  );
+  const sessionState = {};
+  for (const servicedAccessNetwork of servicedAccessNetworks) {
+    // only save subscriber state of networks with hss disabled
+    if (!servicedAccessNetwork?.cellular?.epc?.hss_relay_enabled) {
```
andreilee: If the `cellular` property is missing, then this will still fetch subscribers. Is that intended here?

nms/app/packages/magmalte/app/views/equipment/FEGGatewayDetailSubscribers.js
```diff
@@ -0,0 +1,138 @@
...
+  service: string,
+};
+
+/**
+ * Returns a table of subscribers serviced by the federation gateway.
+ *
+ * @param {federation_gateway} gwInfo The Federation Gateway
+ * @param {boolean} refresh Boolean Value telling to refresh or not.
+ */
+export default function GatewayDetailSubscribers(props: FEGGatewayDetailType) {
```
andreilee: the annotations don't match the parameters

nms/app/packages/magmalte/app/views/equipment/FEGGatewayDetailSubscribers.js
```diff
@@ -0,0 +1,138 @@
...
+    networkId: networkId,
+    type: RefreshTypeEnum.FEG_SUBSCRIBER,
+    interval: REFRESH_INTERVAL,
+    refresh: props.refresh,
+  });
+  const sessionState = ctx?.sessionState || {};
+  const subscriberToNetworkIdMap = {};
+
+  Object.keys(sessionState).map(servicedNetworkId => {
+    // $FlowIgnore because refresh context returns other things too like state, enbInfo and the like
```
andreilee: There's a standard way to comment for `FlowIgnore`: https://flow.org/en/docs/errors/

nms/app/packages/magmalte/app/views/equipment/FEGGatewayDetailSubscribers.js
```diff
@@ -0,0 +1,138 @@
...
+ * @param {boolean} refresh Boolean Value telling to refresh or not.
+ */
+export default function GatewayDetailSubscribers(props: FEGGatewayDetailType) {
+  const {history, match} = useRouter();
+  const networkId: string = nullthrows(match.params.networkId);
+  const [subscriberRows, setSubscriberRows] = useState<
+    Array<SubscriberRowType>,
+  >([]);
+  const [isLoading, setIsLoading] = useState(true);
+  // Auto refresh  every 30 seconds
```
andreilee: Don't include the reference to 30 seconds. If the refresh interval changes, then this will be wrong.

nms/app/packages/magmalte/app/components/context/FEGSubscriberContext.js
```diff
@@ -0,0 +1,28 @@
...
+ * @flow strict-local
+ * @format
+ */
+
+import type {network_id, subscriber_state} from '@fbcnms/magma-api';
+
+import React from 'react';
+
+export type FEGSubscriberContextType = {
+  sessionState: {[network_id]: {[string]: subscriber_state}},
```
andreilee: Rename this please - it's confusing in other files that this is keyed on network id

andreilee (changes requested):
The screenshots need to show subscribers in the table for when the managed network isn't using HSS, and also when it is.

Also, mind the typos.

nms/app/packages/magmalte/app/state/feg/SubscriberState.js
MightyDani: Yes  that is intended. But we also can guarantee that all of those 3 property won't be missing because they are required properties. So, I have removed the question marks in order to remove confusions.

ato-dani: Nvm, I have removed that line because we have to display when the network is using HSS too.

hcgatewood:
Is this still draft? If yes convert it to draft PR. If no, let's work on getting it merged cc @andreilee

nms/app/packages/magmalte/app/components/context/FEGSubscriberContext.js
```diff
@@ -0,0 +1,28 @@
...
+ * @flow strict-local
+ * @format
+ */
+
+import type {subscriber_state} from '@fbcnms/magma-api';
+
+import React from 'react';
+
+export type FEGSubscriberContextType = {
+  sessionState: {[string]: {[string]: subscriber_state}},
```
andreilee: Key type should still be `network_id`. Could you rename `sessionState` to something more meaningful? What does `session` refer to, in this case?

nms/app/packages/magmalte/app/components/context/RefreshContext.js
```diff
@@ -117,6 +124,8 @@ export function useRefreshingContext(props: Props) {
       return ctx.setState(null, null, null, newState);
     } else if (props.type === RefreshTypeEnum.FEG_GATEWAY) {
       return ctx.setState(null, null, newState?.fegGateways || {});
+    } else if (props.type === RefreshTypeEnum.FEG_SUBSCRIBER) {
+      return ctx?.setSessionState(newState);
```
andreilee: Why is optional chaining used here?

nms/app/packages/magmalte/app/state/feg/SubscriberState.js
```diff
@@ -0,0 +1,93 @@
...
+ * @param {InitSubscriberStateProps} props an object containing the necessary values to initialize the subscriber state
+ */
+export async function InitSubscriberState(props: InitSubscriberStateProps) {
+  const {networkId, setSessionState, enqueueSnackbar} = props;
+  const sessionState = await FetchFegSubscriberState({
+    networkId,
+    enqueueSnackbar,
+  });
+  setSessionState(sessionState);
+}
```
andreilee: Maybe you're planning on using this function more, but I don't see it being very useful here. Just inline the function in `FEGContext.js`, the function call takes 5 lines, and the body of this function is 6 lines, so this isn't really making anything easier to read.

nms/app/packages/magmalte/app/views/equipment/FEGGatewayDetailMain.js
```diff
@@ -149,6 +159,17 @@ function FEGGatewayOverview() {
               />
               <FEGGatewayDetailStatus refresh={refresh} />
             </Grid>
+            <Grid item>
+              <CardTitleRow
+                icon={PeopleIcon}
+                label="Managed Subscribers"
+                filter={() => filter(refreshSubscribers, setRefreshSubscribers)}
+              />
```
andreilee: Could you add a tooltip here, actually? Since we've had to clarify the definition of "Managed Subscribers", users will likely need the clarification too.

nms/app/packages/magmalte/app/views/equipment/FEGGatewayDetailMain.js
```diff
@@ -44,6 +48,11 @@ const useStyles = makeStyles(theme => ({
   },
 }));
 
+export type FEGGatewayDetailType = {
+  gwInfo: federation_gateway,
+  refresh: boolean,
+};
+
```
andreilee: Add comments on this type and the fields - what's the intention here?

andreilee: Also you should probably move this to `FEGGatewayDetailSubscribers.js`.

nms/app/packages/magmalte/app/views/equipment/FEGGatewayDetailSubscribers.js
```diff
@@ -0,0 +1,138 @@
...
+  service: string,
+};
+
+/**
+ * Returns a table of subscribers serviced by the federation gateway.
+ *
+ * @param {FEGGatewayDetailType} props An object containing the federation gateway and a boolean
+ *  telling to refresh or not
+ */
+export default function GatewayDetailSubscribers(props: FEGGatewayDetailType) {
```
andreilee: Don't just add an annotation describing what fields the argument object contains - people can easily see that by checking the type definition. It's better to just leave the annotation bare-bones if there's nothing helpful to put.

nms/app/packages/magmalte/app/views/equipment/FEGGatewayDetailSubscribers.js
```diff
@@ -0,0 +1,138 @@
...
+            networkId: subscriberToNetworkIdMap[subscriberImsi],
+            id: subscriberImsi,
+          });
+          newSubscriberRows.push({
+            name: subscriberInfo?.name || subscriberImsi,
+            id: subscriberImsi,
+            service: subscriberInfo?.lte?.state || '-',
+          });
+        }),
+      );
```
andreilee, line 105: Could you add a TODO referencing my name `andreilee` saying to update this to use a bulk fetch API?

nms/app/packages/magmalte/app/views/equipment/FEGGatewayDetailSubscribers.js
```diff
@@ -0,0 +1,138 @@
...
+      setSubscriberRows(newSubscriberRows);
+      setIsLoading(false);
+    };
+    fetchSubscribersInfo();
+    // rerun only when a new subscriber session has been added
+  }, [subscribersImsi]); // eslint-disable-line react-hooks/exhaustive-deps
```
andreilee, line 111: Something to note, but for me to chase down later, is that this won't work very well for a large network with thousands of subscribers.

andreilee: We'll need a paginated API for this on Orc8r side down the road.

andreilee (changes requested):

nms/app/packages/magmalte/app/components/context/FEGSubscriberContext.js
ato-dani: I think sessionState is the appropriate name here. I am saying this because It would be good to do the same thing as what is done in the LTE network side: https://github.com/magma/magma/blob/master/nms/app/packages/magmalte/app/components/context/SubscriberContext.js#L41. This way we wouldn't create confusion for new developers if we use the same names for same thing. Also, here we are saving the subscriber state(not the subscriber him/herself) which equates to a subscriber's session. That is why session state was used in both the access network side and here. So, don't you think that keeping it as sessionState is appropriate?

nms/app/packages/magmalte/app/components/context/RefreshContext.js
ato-dani: Oh, I was being extra protective so as to not get flow error because ctx doesn't contain the setSessionState function for other types like enodebs or gateways but it turns out flow doesn't complain about that. So, I have removed it in order to make it similar to the other code.

nms/app/packages/magmalte/app/state/feg/SubscriberState.js
ato-dani: Yeah, this function has a potential to include many other things with fetching subscriber info being one of them or the subscriber metrics and the like. So, I created a function so as to help make it easier for anyone who adds those informations have an easy time(I will not use it for the rest of my project). However, I will just put it in `FEGContext.js` file itself, just as you suggested.

nms/app/packages/magmalte/app/views/equipment/FEGGatewayDetailSubscribers.js
ato-dani: Yeah, I can't agree more. I have tried to make sure that this expensive function to run most of the time only once but still, a pagination support is needed to truly make it scalable. This is because we still load all the data at the beginning/ pre-populate instead of lazily loading the data.

andreilee (approved):

PR #7847, opened 2021-06-30
Add debug logging to the service303 thread

amarpad:
## Summary

Move logging to the service303 thread

Signed-off-by: Amar Padmanabhan <amar@freedomfi.com>

## Test Plan

make integ_test
validated output after running test
002005 Wed Jun 30 00:59:22 2021 7FECB9095700 DEBUG SERVIC tasks/service303/service303_mme_:0045    ======================================= STATISTICS ============================================
002006 Wed Jun 30 00:59:22 2021 7FECB9095700 DEBUG SERVIC tasks/service303/service303_mme_:0046                   |   Current Status|
002007 Wed Jun 30 00:59:22 2021 7FECB9095700 DEBUG SERVIC tasks/service303/service303_mme_:0049    Attached UEs   |          1      |
002008 Wed Jun 30 00:59:22 2021 7FECB9095700 DEBUG SERVIC tasks/service303/service303_mme_:0052    Connected UEs  |          1      |
002009 Wed Jun 30 00:59:22 2021 7FECB9095700 DEBUG SERVIC tasks/service303/service303_mme_:0055    Connected eNBs |          1      |
002010 Wed Jun 30 00:59:22 2021 7FECB9095700 DEBUG SERVIC tasks/service303/service303_mme_:0058    Default Bearers|          1      |
002011 Wed Jun 30 00:59:22 2021 7FECB9095700 DEBUG SERVIC tasks/service303/service303_mme_:0061    S1-U Bearers   |          1      |
002012 Wed Jun 30 00:59:22 2021 7FECB9095700 DEBUG SERVIC tasks/service303/service303_mme_:0066    ======================================= STATISTICS ============================================

## Additional Information

- [ ] This change is backwards-breaking

PR #8019, opened 2021-07-13
fix(mme): reduce EPC_STATS_TIMER_MSEC back to 20s

uri200:

## Summary

Stats on MME are used to debug issues. Reducing its frequency to 1 min defeats the purpose of this log.

Setting this back to 20s until we have a config parameter to change that value

## Test Plan

make test at agw

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/include/service303.h
+++ b/lte/gateway/c/core/oai/include/service303.h
@@ -27,8 +27,8 @@
 
 #define NO_BOUNDARIES 0
 #define NO_LABELS 0
-#define EPC_STATS_TIMER_MSEC 60000          // In milliseconds
-#define EPC_STATS_DISPLAY_TIMER_MSEC 60000  // In milliseconds
+#define EPC_STATS_TIMER_MSEC 20000          // In milliseconds
+#define EPC_STATS_DISPLAY_TIMER_MSEC 20000  // In milliseconds
 
 void service303_mme_app_statistics_read(
     application_mme_app_stats_msg_t* stats_msg_p);


amarpad (approved):
cc @ulaskozat 
Just want to make sure everyone is ok with this. This is going to be an ITT_MSG every 20 sec.

ulaskozat:
@uri200 The idea is to not use this for debugging purposes only. Have you tested this on Spirent against GA targets? Rather than setting it to a lower value as default, could you add this to mme configuration?

uri200:
> @uri200 The idea is to not use this for debugging purposes only. Have you tested this on Spirent against GA targets? Rather than setting it to a lower value as default, could you add this to mme configuration?

Hey @ulaskozat, note those questions should have been answered here https://github.com/magma/magma/pull/7847 when this value was increased from 20 to 60 seconds: Is there any prof the increase of this value from 20 to 60  really helped? Was this tested against GA  targets before and after the change?

If this has a real impact on performance, of course let's set it to 60. But if it does not have any, we should not compromise other features of the system.

amarpad:
Hi @uri200 the frequency of ITTI messages were always 60s. The service303 and S1 threads were reaching into MME_APP which was a correctness issue not a logging issue. So I moved the message to ITTI. Since the MME_APP thread no longer had all the state we decided to move logging to the service303 thread and there is no point having a display frequency of 20s if the state is updated every 60s

uri200:
Hey @amarpad so that means that to increase the frequency of that message we would need to increase the frequency of the whole IITI queue?

amarpad:
> Hey @amarpad so that means that to increase the frequency of that message we would need to increase the frequency of the whole IITI queue?

EPC_STATS_TIMER_MSEC determines the frequency of sending ITTI messages which you are updating so this works, just that it increases the load on ITTI messages.

uri200:
yep, but my question is: with the changes done on the previous PR the logging function can only report changes as fast as the ITTI queue. Is that right?

ardzoht:
I think we should make this a configurable field, it is clear that there's a need for adjusting the frequency for different purposes (TeraVM as Oriol explains) with the trade-offs of performance, we can add a local service config to control this, wdyt? cc @ulaskozat @amarpad

amarpad:
> yep, but my question is: with the changes done on the previous PR the logging function can only report changes as fast as the ITTI queue. Is that right?

yep right. All tasks report their state to the service303 thread and threads don't access each others state. MME datastructures are not thread safe so what was happening could trigger issues like https://github.com/magma/magma/issues/7421

ulaskozat:
> > @uri200 The idea is to not use this for debugging purposes only. Have you tested this on Spirent against GA targets? Rather than setting it to a lower value as default, could you add this to mme configuration?
> 
> Hey @ulaskozat, note those questions should have been answered here #7847 when this value was increased from 20 to 60 seconds: Is there any prof the increase of this value from 20 to 60 really helped? Was this tested against GA targets before and after the change?
> 
> If this has a real impact on performance, of course let's set it to 60. But if it does not have any, we should not compromise other features of the system.

@uri200 while writing a long message on the context, I see that @amarpad has already provided it. Let's not beat ourselves on this and make this a configurable parameter with default as 60 seconds and @uri200 you can use any value you want for observability during tests.

ulaskozat:
@uri200 @amarpad @ardzoht I will raise a PR to make this configurable...

PR #8020, opened 2021-07-13
refactor(sessiond): Add default output for polling with empty traffic

veshkemburu:
## Summary

Stats requests with empty traffic should still show empty rule record table outputs in sessiond, instead of no output.
Cleaning up pipelined log output.

## Test Plan

Run traffic tests in ng40 
Run sudo journalctl -fu magma@sessiond in access gateway VM

## Additional Information


lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -409,8 +409,8 @@ def _report_usage(self, usage):
         record_table = RuleRecordTable(records=usage.values(),
                                        epoch=global_epoch)
         if self._print_grpc_payload:
-            record_msg = 'Sending RPC payload: {0}{{\n{1}}}'.format(
-                record_table.DESCRIPTOR.name, str(record_table))
+            record_msg = 'Sending RPC payload: \n{0}{{{1}}}'.format(
+                record_table.DESCRIPTOR.name, "\n"+str(record_table))
```
uri200, line 413: not sure if we are trying to ident the whole message or just to separate headers here. But note this will only add \n to the first line. The rest of the lines will not have any indentation

veshkemburu: It was just meant to separate headers. Ill need to look at indentation maybe try pulling that grpcutils function in

koolzz: @veshkemburu can you show the difference between before/after this change?

hcgatewood:
@themarwhal @uri200 PR is looking stale, can we get it merged or closed

Issue #6210, opened 2021-04-19
AGW: s1ap: MME: LeakSanitizer: detected memory leaks


pshelar:
### Your Environment

- **Version:** master
- **Affected Component:** Access Gateway
- **Affected Subcomponent:** MME
- **Deployment Environment:** Ubuntu Vagrant (AGW)

### Describe the Issue

There are multiple memory leak detected by LeakSanitizer.

**To Reproduce**
Run S1AP test. check syslogd

```
Apr 19 05:48:55 magma-dev-focal mme[41014]: ==41014==ERROR: LeakSanitizer: detected memory leaks
Apr 19 05:48:55 magma-dev-focal mme[41014]: Direct leak of 48 byte(s) in 1 object(s) allocated from:
Apr 19 05:48:55 magma-dev-focal mme[41014]:     #0 0x7fde49b0bdc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
Apr 19 05:48:55 magma-dev-focal mme[41014]:     #1 0x7fde48c7dc0f in zlist_new (/lib/x86_64-linux-gnu/libczmq.so.4+0x24c0f)
Apr 19 05:48:55 magma-dev-focal mme[41014]: Direct leak of 16 byte(s) in 1 object(s) allocated from:
Apr 19 05:48:55 magma-dev-focal mme[41014]:     #0 0x7fde49b0bdc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
Apr 19 05:48:55 magma-dev-focal mme[41014]:     #1 0x5596c4e42633 in create_new_attach_info /home/vagrant/magma/lte/gateway/c/oai/tasks/nas/emm/Attach.c:2550
Apr 19 05:48:55 magma-dev-focal mme[41014]:     #2 0x5596c4e1728e in emm_proc_attach_request /home/vagrant/magma/lte/gateway/c/oai/tasks/nas/emm/Attach.c:399
Apr 19 05:48:55 magma-dev-focal mme[41014]:     #3 0x5596c4ede88e in emm_recv_attach_request /home/vagrant/magma/lte/gateway/c/oai/tasks/nas/emm/sap/emm_recv.c:422
Apr 19 05:48:55 magma-dev-focal mme[41014]:     #4 0x5596c4f6743b in emm_as_recv /home/vagrant/magma/lte/gateway/c/oai/tasks/nas/emm/sap/emm_as.c:377
[...]

PR #8029, opened 2021-07-14
fix(agw): Fixed memory leaks during execution of s1ap sanity test suite

rsarwad:
fix(agw): Fixes the issue, Fixes the issue, https://github.com/magma/magma/issues/6210

## Summary
All the memory leaks reported in issue, https://github.com/magma/magma/issues/6210 are not observed on latest master codebase. And provided fixes for which memory leaks observed on latest master codebase.

## Test Plan

### DIFF
--- a/lte/gateway/c/core/oai/tasks/nas/emm/TrackingAreaUpdate.c
+++ b/lte/gateway/c/core/oai/tasks/nas/emm/TrackingAreaUpdate.c
@@ -682,11 +682,8 @@ static int emm_tracking_area_update_accept(nas_emm_tau_proc_t* const tau_proc) {
       emm_sap.u.emm_as.u.establish.emergency_number_list = NULL;
 
       emm_sap.u.emm_as.u.establish.eps_network_feature_support =
-          calloc(1, sizeof(eps_network_feature_support_t));
-      emm_sap.u.emm_as.u.establish.eps_network_feature_support->b1 =
-          _emm_data.conf.eps_network_feature_support[0];
-      emm_sap.u.emm_as.u.establish.eps_network_feature_support->b2 =
-          _emm_data.conf.eps_network_feature_support[1];
+          (eps_network_feature_support_t*) &_emm_data.conf
+              .eps_network_feature_support;
       emm_sap.u.emm_as.u.establish.additional_update_result = NULL;
       emm_sap.u.emm_as.u.establish.t3412_extended           = NULL;
       emm_sap.u.emm_as.u.establish.nas_msg =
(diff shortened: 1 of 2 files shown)


lte/gateway/c/core/oai/tasks/nas/emm/TrackingAreaUpdate.c
```diff
@@ -682,11 +682,8 @@ static int emm_tracking_area_update_accept(nas_emm_tau_proc_t* const tau_proc) {
       emm_sap.u.emm_as.u.establish.emergency_number_list = NULL;
 
       emm_sap.u.emm_as.u.establish.eps_network_feature_support =
-          calloc(1, sizeof(eps_network_feature_support_t));
-      emm_sap.u.emm_as.u.establish.eps_network_feature_support->b1 =
-          _emm_data.conf.eps_network_feature_support[0];
-      emm_sap.u.emm_as.u.establish.eps_network_feature_support->b2 =
-          _emm_data.conf.eps_network_feature_support[1];
+          (eps_network_feature_support_t*) &_emm_data.conf
```
ardzoht, line 685: could this lead to eps_network_feature_support ptr being freed later and then another place trying to access it? I assume we do a nullptr check if that's the case, just want to make sure this doesn't lead to crash

ardzoht (approved):
lgtm, thanks! just left one question regarding the `eps_network_feature_support` ptr

lte/gateway/c/core/oai/tasks/nas/emm/TrackingAreaUpdate.c
rsarwad: Here we are not actually allocating memory, just sending address of parameter, eps_network_feature_support which is part of configuration structure.
While accessing we have null check.

ulaskozat (approved):
LGTM

themarwhal:
Hey @ulaskozat & @rsarwad, there are too many conflicts to backport into 1.5. I'll be removing the apply-v1.5 label. Please raise a PR against the 1.5 branch. Thanks! :)

themarwhal:
Actually, looks like there's a merge conflict for 1.6 as well. Would be great if they can be backported as a PR too. Thanks!

ulaskozat:
cc @rsarwad for backporting PRs

rsarwad:
> cc @rsarwad for backporting PRs

Two memory leaks were introduced by snippet of code that is neither in v1.5 nor in v1.6. So I shall check the memory in individual branches and update.

ulaskozat:
but TAU one is there as far as I see.

PR #8042, opened 2021-07-14
chore(agw): remove CLion/CMake gitignore

markjen:
## Summary

Remove gitignore entries for cmake-build-debug folders. CLion/CMake pollutes checkouts with cmake-build-* folders, but they can be ignored by developers via their own global gitignore settings or entries in $MAGMA_ROOT/.gitignore/info/excludes

## Test Plan

Worked in my dev environment

## Additional Information

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,8 +65,8 @@ aws_keys.tfvars
 # Helm
 orc8r/cloud/helm/**/.secrets/
 
-# CMake debug files
-lte/gateway/c/session_manager/cmake-build-debug/
+# Certs for accessing the cloud
+.certs
 
 # Log files
 *.log


electronjoe:
@markjen Is the script for proto generation to aid in CLion / IntelliJ use? Just to clarify use case.

electronjoe:
Also, any context on what the generation of `cmake-build-debug` is about? Was there no gitignore for this prior to the /core/ migration? I'm not familiar with the source of `cmake-build-debug`.

markjen:
> @markjen Is the script for proto generation to aid in CLion / IntelliJ use? Just to clarify use case.

@uri200 had this script in a private paste/snippet and it's useful for seeing generated proto files that are being used. I had been using it to generate protos, but I figured out how to make things work without this script. I'll update our wiki with info.

> Also, any context on what the generation of `cmake-build-debug` is about? Was there no gitignore for this prior to the /core/ migration? I'm not familiar with the source of `cmake-build-debug`.

This dir is the default name of a local cache of cmake build files that are copied into a dev checkout by CLion upon remote execution. The same ignore exists for lte/gateway/c/session_manager; I'll make a more general .gitignore for all dirs under lte/gateway/c (any CMakeLists.txt project opened by CLion will create this directory)

markjen:
note: an alternative approach is to remove this ignore entirely and have the developer specify it in their own .git/info/excludes or global gitignore if they're using CLion. We could take this approach for the IntelliJ and VSCode ignores in the .gitignore file as well.

.gitignore
```diff
@@ -76,8 +76,8 @@ orc8r/cloud/helm/**/.secrets/
 # Certs for accessing the cloud
 .certs
 
-# CMake debug files
-lte/gateway/c/session_manager/cmake-build-debug/
+# CMake debug files; copied into source tree by CLion when using remote CMake
+lte/gateway/c/**/cmake-build-debug
```
uri200: note this directory depends on the build directory defined on clion so I never added it.

See below how many different flavors I currently have on my mac

```
#      untracked:  [4] c/core/oai/cmake-build-debug/
#      untracked:  [5] c/oai/cmake-build-debug/
#      untracked:  [6] c/session_manager/cmake-build-debug-remote-host-2222/
#      untracked:  [7] c/session_manager/cmake-build-debug-remote-host_default/
```

Would something like this work?
```
lte/gateway/c/**/cmake-build-**
```

Or maybe what @electronjoe mentioned with `exclude` file


image

uri200 (commented):
> > @markjen Is the script for proto generation to aid in CLion / IntelliJ use? Just to clarify use case.
> 
> @uri200 had this script in a private paste/snippet and it's useful for seeing generated proto files that are being used. I had been using it to generate protos, but I figured out how to make things work without this script. I'll update our wiki with info.
> 
> > Also, any context on what the generation of `cmake-build-debug` is about? Was there no gitignore for this prior to the /core/ migration? I'm not familiar with the source of `cmake-build-debug`.
> 
> This dir is the default name of a local cache of cmake build files that are copied into a dev checkout by CLion upon remote execution. The same ignore exists for lte/gateway/c/session_manager; I'll make a more general .gitignore for all dirs under lte/gateway/c (any CMakeLists.txt project opened by CLion will create this directory)

Thanks @markjen  for automating the script. I had this script on my wiki annd since I am the only one using the debugger on sessiond, and it was some kind of manual process, I never thought to add it there. 


Btw, can somehow add this to sessiond CMAKE using `add_custom_command`. Using a flag we can trigger the creation from clion every time we run the debugger If we set it behind a flag (same we do with `BUILD_TESTS`).ay we can trigger the creation of those protos every time we run test on debugger on CLION

Note if we ever change or add locations of the protos this will break. I think remember that happened to me shortly after I created the script.

uri200:
> note: an alternative approach is to remove this ignore entirely and have the developer specify it in their own .git/info/excludes or global gitignore if they're using CLion. We could take this approach for the IntelliJ and VSCode ignores in the .gitignore file as well.

Hey @electronjoe , I think that solution is better not to pollute gitignore. 
I just added `cmake-build-*` to that file and now git is not taking any directory from clion anymore

```
cat ~/magma/.git/info/exclude
cmake-build-*
``

markjen:
> > note: an alternative approach is to remove this ignore entirely and have the developer specify it in their own .git/info/excludes or global gitignore if they're using CLion. We could take this approach for the IntelliJ and VSCode ignores in the .gitignore file as well.
> 
> Hey @electronjoe , I think that solution is better not to pollute gitignore.
> I just added `cmake-build-*` to that file and now git is not taking any directory from clion anymore
> 
> ```
> cat ~/magma/.git/info/exclude
> cmake-build-*
> ``
> ```

I updated this PR to just remove the entries in .gitignore

electronjoe:
> I updated this PR to just remove the entries in .gitignore

SGTM! Our gitignore is a bit of a cluster already =( But at least your change here was very precise (full path) - I am totally against wildcards that don't give full paths (which we have...).

electronjoe (approved):

uri200:
I see the part to generate the protos is gone.  Is that intentional?

markjen:
> I see the part to generate the protos is gone. Is that intentional?

Yeah, I removed that from this PR. If you want, I can add it in a separate PR (it has less to do with this one anyways)

electronjoe:
There is a conflict on this PR so you'll need to rebase and re-push, I enabled auto-merge once that's satisfied. @markjen

markjen:
> There is a conflict on this PR so you'll need to rebase and re-push, I enabled auto-merge once that's satisfied. @markjen

Re-pushed, but I think automerge is blocked because the `Semantic Pull Request` check is stuck?

PR #8043, opened 2021-07-14
feat(ci): Test to run hourly teravm Job in CI

uri200:

## Summary

This is a test job to try to run TeraVM regression test on CI.

The idea is to run an hourly test on master. Similar to the xwf hourly test we already run.
The Job will run the fab script on the lab remotely using a third party node.

Note the job will fail because we have one test that fails. But this is fine

## Test Plan

Tested on a PR branch (but this will run on master once landed)
https://app.circleci.com/pipelines/github/magma/magma/28866/workflows/3f17cc0a-e75e-418b-ad7c-840f7e57c15d/jobs/323370

image

## Additional Information

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1349,6 +1349,28 @@ jobs:
           version_path: versions/nms_version
       - magma_slack_notify
 
+
+  ### Regression
+
+  teravm-regression:
+    machine:
+      image: ubuntu-1604:201903-01
+      docker_layer_caching: true
+    resource_class: large
+    environment:
+      MAGMA_ROOT: /home/circleci/project
+    steps:
+      - checkout
+      - run:
+          name: run terravm test
+          no_output_timeout: 60m
+          command: |
+            cd ${MAGMA_ROOT}
+            cd ci-scripts/teravm/go/teravm_ci/
+            ./teravm_ci -key $TERAVM_REMOTE_ACCESS_KEY
+
+
+
   ### XWF
 
   xwfm-test:
@@ -1661,6 +1683,14 @@ workflows:
           images: 'gateway_python|gateway_pipelined|gateway_go'
           tag-latest: true
 
+
+  hourly_teravm_regression:
+    triggers:
+      - schedule:
+          cron: "0 * * * *"
+          <<: *only_master
+    jobs:
+      - teravm-regression
 
   cwf_operator:
     jobs:
--- a/ci-scripts/teravm/fabfile.py
+++ b/ci-scripts/teravm/fabfile.py
@@ -18,6 +18,7 @@ import os
 import re
 import sys
 import time
+from datetime import datetime
 
 from fabric.api import cd, env, hide, local, run, settings
 from fabric.operations import put, sudo
@@ -88,8 +89,8 @@ def upgrade_and_run_3gpp_tests(
     if err:
         sys.exit(1)
 
-    fastprint("\nSleeping for 30 seconds to make sure system is read\n\n")
-    time.sleep(30)
+    fastprint("\nSleeping for 60 seconds to make sure system is read\n\n")
+    time.sleep(60)
 
     verdicts = run_3gpp_tests(setup, key_filename, custom_test_file)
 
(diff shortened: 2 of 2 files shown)


lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -86,8 +86,10 @@ async def _run(self) -> None:
         subscribers, flat_digest = await self._get_all_subscribers()
         if subscribers is None:
             return
-        self._update_flat_digest(flat_digest)
+        # Process subscriber data before digest data, in case there's a gateway
+        # failure between the calls
         self._process_subscribers(subscribers)
+        self._update_flat_digest(flat_digest)
```
uri200: renoving this change

rdefosse:
@uri200 @quentinDERORY 

I thought we would avoid adding new CI jobs to `circle-ci` env.

Regards
Raphael

quentinDERORY:
> @uri200 @quentinDERORY
> 
> I thought we would avoid adding new CI jobs to `circle-ci` env.
> 
> Regards
> Raphael

It's true it would be best to add new jobs to github actions instead of circle ci.
But for here specifically, this job will need to get migrated to the new jenkins as we don't want to use remote internal lab node.
It's fine to add this one still.

ci-scripts/teravm/fabfile.py
```diff
@@ -386,9 +395,11 @@ def _get_gateway_image(hash):
 
 def _get_latest_agw_tag(setup, key_filename):
     _setup_env("magma", VM_IP_MAP[setup]["gateway"], key_filename)
+    """
```
quentinDERORY: Can we just delete commented code?

rdefosse:
> 
> 
> > @uri200 @quentinDERORY
> > I thought we would avoid adding new CI jobs to `circle-ci` env.
> > Regards
> > Raphael
> 
> It's true it would be best to add new jobs to github actions instead of circle ci.
> But for here specifically, this job will need to get migrated to the new jenkins as we don't want to use remote internal lab node.
> It's fine to add this one still.

Hi @quentinDERORY 

Yes I meant migrating to LF Jenkins (not necessarily to GHA).

I just wanted to avoid double work (another migration to do).

Regards
Raphael

quentinDERORY (approved):
LGTM except linting/python style issues

uri200:
Hey @rdefosse , @quentinDERORY , this is a regression job we have been waiting for the last 2 years. We have now been able to workaround the issues and integrate it.

For now the idea is to run it as experimental and see how it responds and what we should improve on our regression test bed. It will run as a cron job and will not block master.

Note this has to run on an internal lab since teravn is a full environment (not virtual).

ci-scripts/teravm/fabfile.py
uri200: this should not have been here. Thanks

.circleci/config.yml
```diff
@@ -1349,6 +1349,28 @@ jobs:
...
+      - run:
+          name: run terravm test
+          no_output_timeout: 60m
+          command: |
+            cd ${MAGMA_ROOT}
+            cd ci-scripts/teravm/go/teravm_ci/
+            ./teravm_ci -key $TERAVM_REMOTE_ACCESS_KEY
+
+
+
```
uri200, line 1373: note this space is the same amount we are leaving in other parts of this file

rdefosse:
> 
> 
> Hey @rdefosse , @quentinDERORY , this is a regression job we have been waiting for the last 2 years. We have now been able to workaround the issues and integrate it.
> 
> For now the idea is to run it as experimental and see how it responds and what we should improve on our regression test bed. It will run as a cron job and will not block master.
> 
> Note this has to run on an internal lab since teravn is a full environment (not virtual).

@uri200 

I understand the pain! I just wanted to flag that we were supposed to migrate to LF Jenkins.

Regards
Raphael

ekowtaylor:
@uri200 it looks like this change has gone stale, what is the plan to move this forward

uri200:
> @uri200 it looks like this change has gone stale, what is the plan to move this forward

We have to find some time to test this more. But the idea is to use it in the near future while we don't have any other solution for teravm + CI

ekowtaylor:
@uri200, this PR has now gone stale. To maintain a healthy codebase, it will automatically closed unless there are plans to rebase and merge by close of week

uri200:
> @uri200, this PR has now gone stale. To maintain a healthy codebase, it will automatically closed unless there are plans to rebase and merge by close of week

Hey @ekowtaylor this is still on hold due to other priorities. Needs to be tested more to see if it works well. It is on my to do list.

PR #8044, opened 2021-07-14
feat(orc8r): Implement logic for caching subscriber protos

wangyyt1013:

## Summary

1. Add `SubProtoStore` that manages cached subscriber protos
2. Add logic in `subscriberdb_cache` that constantly updates the `SubProtoStore`
3. Add logic in `subscriberdb` that loads subscriber protos from the cahce, instead of from `configurator` (behind a feature flag)

## Test Plan

2. Added new test cases for added code

## Additional Information


lte/protos/subscriberdb.proto
```diff
@@ -242,7 +242,7 @@ message SyncSubscribersResponse {
   repeated SubscriberDigestWithID per_sub_digests = 3;
   // to_renew is the set of all subscriber data modified/added on the server side,
   // indexed by their IDs (IMSI).
-  map<string, SubscriberData> to_renew = 4;
+  repeated SubscriberData to_renew = 4;
```
hcgatewood, line 245: 1. Why the change? (And update the comment if we keep the change)
2. Will probably need to make an accompanying change on the gateway

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -28,9 +29,9 @@ import (
 	"github.com/thoas/go-funk"
 )
 
-func MonitorDigests(config Config, digestStore storage.DigestStore, perSubDigestStore *storage.PerSubDigestStore) {
+func MonitorDigests(config Config, digestStore storage.DigestStore, perSubDigestStore *storage.PerSubDigestStore, subProtoStore *storage.SubProtoStore) {
```
hcgatewood: Just call it "subStore"

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -66,43 +69,108 @@ func RenewDigests(
...
+	err = subProtoStore.DeleteSubProtos(networksToRemove)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "remove sub protos of invalid networks")
+	}
 
 	errs := &multierror.Error{}
-	flatDigestsByNetwork := map[string]string{}
 	perSubDigestsByNetwork := map[string][]*lte_protos.SubscriberDigestWithID{}
+	flatDigestsByNetwork := map[string]string{}
 	for _, network := range networksToRenew {
```
hcgatewood, line 80: The logic in this for loop is wonky (and don't use goto except with dire need). Pull each iteration of the loop into a fn, that should help

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -66,43 +69,108 @@ func RenewDigests(
...
+		return nil, errors.Wrapf(err, "get per sub dgests to update")
+	}
+	err = perSubDigestStore.SetDigest(network, perSubDigests)
+	if err != nil {
+		return nil, errors.Wrapf(err, "set per sub digest")
+	}
+	return perSubDigests, nil
+}
+
+func updateDigestByNetwork(network string, digestStore storage.DigestStore) (string, string, error) {
```
hcgatewood: "updateDigest"

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -66,43 +69,108 @@ func RenewDigests(
...
+			multierror.Append(errs, errors.Wrapf(err, "update subscriber protos for network %+v", network))
 		}
-		flatDigestsByNetwork[network] = digest
-		perSubDigestsByNetwork[network] = perSubDigests
+		// TODO(wangyyt1013): add logs for updated sub protos
 	}
 	return flatDigestsByNetwork, perSubDigestsByNetwork, errs.ErrorOrNil()
 }
 
+func updatePerSubDigestsByNetwork(network string, perSubDigestStore *storage.PerSubDigestStore) ([]*lte_protos.SubscriberDigestWithID, error) {
```
hcgatewood: "updatePerSubDigests"

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -66,43 +69,108 @@ func RenewDigests(
...
+		return "", "", errors.Wrapf(err, "get previous flat digest")
+	}
+	err = digestStore.SetDigest(network, digest)
+	if err != nil {
+		return "", "", errors.Wrapf(err, "set flat digest")
+	}
+	return digest, prevDigest, nil
+}
+
+func updateSubProtosByNetwork(network string, subProtoStore *storage.SubProtoStore) error {
```
hcgatewood: "updateSubscribers"

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -66,43 +69,108 @@ func RenewDigests(
...
+
+	token := ""
+	foundEmptyToken := false
+	for !foundEmptyToken {
+		subProtos, nextToken, err := subscriberdb.LoadSubProtosPage(0, token, network, apnsByName, lte_models.ApnResources{})
+		if err != nil {
+			return err
+		}
+
+		err = subProtoStore.InsertManyByNetwork(network, subProtos)
```
hcgatewood: Okay just a general pattern throughout: no need to say "by network" because (a) you're only taking the action for a single network and (b) given that you pass `network` as a param, it's clear that it's scoped to the particular networ

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+func NewSubProtoStore(db *sql.DB, builder sqorc.StatementBuilder) *SubProtoStore {
+	return &SubProtoStore{db: db, builder: builder}
+}
+
+// getSqlDialect looks up the sql dialect of the current environment once and stores it
+// in memory cache for use.
+func getSqlDialect() string {
+	once.Do(func() {
+		dialect, envFound := os.LookupEnv("SQL_DIALECT")
+		// Default to postgresql
```
hcgatewood: Do we do this in other parts of the codebase? iirc we actually require sql_dialect to be set and error if it's not

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+			Column(subProtoSidCol).Type(sqorc.ColumnTypeText).NotNull().EndColumn().
+			Column(subProtoProtoCol).Type(sqorc.ColumnTypeBytes).NotNull().EndColumn().
+			PrimaryKey(subProtoNidCol, subProtoSidCol).
+			RunWith(tx).
+			Exec()
+		if err != nil {
+			return nil, errors.Wrap(err, "initialize sub proto store table")
+		}
+		_, err = l.builder.CreateTable(subProtoTmpTableName).
+			IfNotExists().
```
hcgatewood: Can you just do "create table if not exists like subtable" ?

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+			sqlDialect = sqorc.PostgresDialect
+		} else {
+			sqlDialect = dialect
+		}
+	})
+	return sqlDialect
+}
+func (l *SubProtoStore) Initialize() error {
+	txFn := func(tx *sql.Tx) (interface{}, error) {
+		_, err := l.builder.CreateTable(subProtoTableName).
```
hcgatewood: Here and throughout: "sub proto" isn't necessary, just call it "subscriber" e.g. `subscriberTableName`

And actually, for conciseness just call it `subscriberTable`

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// CommitUpdateByNetwork cleans up and re-populates the subscriber proto store table with data from the temporary
+// table for a particular network, and then truncates the temporary table.
+func (l *SubProtoStore) CommitUpdateByNetwork(network string) error {
+	txFn := func(tx *sql.Tx) (interface{}, error) {
+		// HACK: hard coding sql query because there currently doesn't exist good support
+		// for this following query with squirrel
+		deleteRowsOnlyInRealTableQuery := fmt.Sprintf(
+			"DELETE FROM %s WHERE (%s = \"%s\") AND ((%s, %s) NOT IN (SELECT %s, %s FROM %s))",
```
hcgatewood: You can still use squirrel for part of this, just inside the `.Where(...)` you'll have to hardcode that bit

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+		_, err := insertQuery.RunWith(tx).Exec()
+		return nil, errors.Wrapf(err, "insert sub protos into store for network %+v", network)
+	}
+	_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
+	return err
+}
+
+// CommitUpdateByNetwork cleans up and re-populates the subscriber proto store table with data from the temporary
+// table for a particular network, and then truncates the temporary table.
+func (l *SubProtoStore) CommitUpdateByNetwork(network string) error {
```
hcgatewood: Call this "ApplyUpdate" since "commit" has a slightly different meaning in sql -- i.e., "commit" makes me think it's committing a *transaction*, whereas really we're committing "to the accumulated changes"

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+	return nextToken, nil
+}
+
+// decodePageToken returns the deserialized page token proto.
+func decodePageToken(token string) (*configurator_storage.EntityPageToken, error) {
+	marshaledToken, err := base64.StdEncoding.DecodeString(token)
+	if err != nil {
+		return nil, errors.Wrapf(err, "decode page token")
+	}
+	buf := &configurator_storage.EntityPageToken{}
```
hcgatewood: Add a note somewhere that we're reusing the configurator entity page token type for simplicity, but that this is technically an unrelated page token

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	err := rows.Err()
+	if err != nil {
+		return nil, errors.Wrap(err, "get sub protos from store, SQL rows error")
+	}
+	return subProtos, nil
+}
+
+// getNextPageToken returns the next page token based on the lastIncludedEntity in the current page.
+func getNextPageToken(subProtos []*lte_protos.SubscriberData) (string, error) {
```
hcgatewood: These two page token fns probably do or should already exist in configurator -- can we just make those public (or move these to configurator and make them public) and then reuse them in both places

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+	subProtos := ret.([]*lte_protos.SubscriberData)
+	nextPageToken, err := getNextPageToken(subProtos)
+	if err != nil {
+		return nil, "", errors.Wrapf(err, "get next page token")
+	}
+	return subProtos, nextPageToken, nil
+}
+
+// GetByIDs returns an ordered list of subscriber protos with matching IDs.
+func (l *SubProtoStore) GetByIDs(network string, sids []string) ([]*lte_protos.SubscriberData, error) {
```
hcgatewood: "GetSubscribers"

hcgatewood: Consider adding a guard that "if len(sids) > max page size then return err" -- and document that in the docstring

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+			Exec()
+		return nil, errors.Wrapf(err, "delete sub protos for networks %+v", networks)
+	}
+	_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
+	return err
+}
+
+// GetPage gets a page of subscriber protos based on the page token and size, and also returns
+// the token for the next page.
+func (l *SubProtoStore) GetPage(network string, token string, pageSize uint64) ([]*lte_protos.SubscriberData, string, error) {
```
hcgatewood: "GetSubscribersPage"

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		lastIncludedSid = decoded.LastIncludedEntity
+	}
+	txFn := func(tx *sql.Tx) (interface{}, error) {
+		rows, err := l.builder.
+			Select(subProtoProtoCol).
+			From(subProtoTableName).
+			Where(squirrel.And{
+				squirrel.Eq{subProtoNidCol: network},
+				squirrel.Gt{subProtoSidCol: lastIncludedSid},
```
hcgatewood: Interesting that empty string is less than non-empty string, but makes logic work

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+			RunWith(tx).
+			Exec()
+		return nil, errors.Wrapf(err, "clear sub protos tmp table")
+	}
+	_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
+	return err
+}
+
+// DeleteSubProtos deletes the cached protos for a list of networks.
+func (l *SubProtoStore) DeleteSubProtos(networks []string) error {
```
hcgatewood: "PurgeNetworks" maybe, or "DeleteSubscribersForNetworks"

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "clean up tmp sub proto store table")
+		}
+		return nil, nil
+	}
+	_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
+	return err
+}
+
+func (l *SubProtoStore) ClearTmpTable() error {
```
hcgatewood: "InitiateUpdate" -- the fact that we're clearing a tmp table should be an implementation detail. Instead, it can be "InitiateUpdate", do some work, then "ApplyUpdate"

And put the fns in order -- "InitateUpdate" should be higher in the file than "ApplyUpdate"

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	txFn := func(tx *sql.Tx) (interface{}, error) {
+		_, err := insertQuery.RunWith(tx).Exec()
+		return nil, errors.Wrapf(err, "insert sub protos into store for network %+v", network)
+	}
+	_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
+	return err
+}
+
+// CommitUpdateByNetwork cleans up and re-populates the subscriber proto store table with data from the temporary
```
hcgatewood: From other comment: docstrings should tell "what" the fn is doing (the abstract conceptual affordances the caller cares about) instead of "how" the fn is doing it (these are implementation details, and belong either inside the fn or in a comment on the object defn

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+// table for a particular network, and then truncates the temporary table.
+func (l *SubProtoStore) CommitUpdateByNetwork(network string) error {
+	txFn := func(tx *sql.Tx) (interface{}, error) {
+		// HACK: hard coding sql query because there currently doesn't exist good support
+		// for this following query with squirrel
+		deleteRowsOnlyInRealTableQuery := fmt.Sprintf(
```
hcgatewood: var name too long

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+		)
+		_, err := tx.Exec(deleteRowsOnlyInRealTableQuery)
+		if err != nil {
+			return nil, errors.Wrapf(err, "clean up previous sub proto store table")
+		}
+
+		// the upsertColumnPrefix is used to refer to the the table whose column value is
+		// used in the "on conflict update" operation; it depends on the sql dialect
+		// e.g. for mariaDB, "f1=t2.f1"; for postgres, "f1=excluded.f1"
+		var upsertColumnPrefix string
```
hcgatewood: Move this whole thing (get sql dialect, choose which upsert format string, and actually formatting the string) to sqorc helper fn

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+		switch getSqlDialect() {
```
hcgatewood: i.e. "getsqldialect" shouldn't be exposed in this part of the codebase

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+		case sqorc.PostgresDialect:
+			upsertColumnPrefix = postgresUpsertColumnPrefix
+		case sqorc.MariaDialect:
+			upsertColumnPrefix = subProtoTmpTableName
+		}
+		_, err = l.builder.
```
hcgatewood: This query is a bit convoluted (necessarily so), so add a comment above the var defn with "query should look something like this: ..." for readability

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+	return sqlDialect
+}
+func (l *SubProtoStore) Initialize() error {
+	txFn := func(tx *sql.Tx) (interface{}, error) {
+		_, err := l.builder.CreateTable(subProtoTableName).
+			IfNotExists().
+			Column(subProtoNidCol).Type(sqorc.ColumnTypeText).NotNull().EndColumn().
+			Column(subProtoSidCol).Type(sqorc.ColumnTypeText).NotNull().EndColumn().
+			Column(subProtoProtoCol).Type(sqorc.ColumnTypeBytes).NotNull().EndColumn().
+			PrimaryKey(subProtoNidCol, subProtoSidCol).
```
hcgatewood: :+1:

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+			return nil, errors.Wrap(err, "initialize sub proto store tmp table")
+		}
+		return nil, nil
+	}
+	_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
+	return err
+}
+
+// InsertManyByNetwork serializes and inserts a batch of subscriber protos into the temporary table.
+func (l *SubProtoStore) InsertManyByNetwork(network string, subProtos []*lte_protos.SubscriberData) error {
```
hcgatewood: Same here: document and add a check for len(subprotos)

lte/cloud/go/services/subscriberdb/test_init/test_service_init.go
```diff
@@ -34,6 +34,11 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const (
+	// testMaxProtosLoadSize is the maxProtosLoadSize used in a test subscriberdb service.
```
hcgatewood: Generally: wrap docstrings to 79 chars

lte/cloud/go/services/subscriberdb_cache/worker_test.go
```diff
@@ -141,30 +155,129 @@ func TestSubscriberdbCacheWorker(t *testing.T) {
...
+	assert.Empty(t, subProtos)
+	assert.Empty(t, nextToken)
 
 	allNetworks, err = storage.GetAllNetworks(digestStore)
 	assert.NoError(t, err)
 	assert.Equal(t, []string{"n2"}, allNetworks)
 	clock.UnfreezeClock(t)
 }
+
+func TestUpdateSubProtosByNetworkNoChange(t *testing.T) {
```
hcgatewood, line 204: Add a short comment summarizing what this fn is testing

lte/cloud/go/services/subscriberdb/config.go
```diff
@@ -19,4 +19,10 @@ type Config struct {
 	// ChangesetSizeTheshold specifies the max size of the cloud-agw changeset
 	// past which a resync signal will be sent back to the agw.
 	ChangesetSizeTheshold int `yaml:"changesetSizeTheshold"`
+	// UseSubProtoStore is a feature flag for loading subscriber data from the
+	// cache store, instead of directly from configurator.
+	UseSubProtoStore bool `yaml:"useSubProtoStore"`
```
hcgatewood: Eh no need for a separate config -- just use "flat digest enabled" to gate the entirety of this functionality

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -36,20 +36,27 @@ import (
 type subscriberdbServicer struct {
```
hcgatewood, line 36: For this file: seems like we could potentially coalesce a decent bit of the functionality between this file and the worker.go. Lmk if you think so

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
```diff
@@ -0,0 +1,335 @@
...
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree.
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package storage
```
hcgatewood: Just call this file "subscribers.go" and "subscribers_test.go"

hcgatewood (changes requested):
Shape looks great!

lte/protos/subscriberdb.proto
wangyyt1013: 1. This is mostly for consistently. `SyncSubscribers` and `ListSubscribers` both read their results from the subscriber proto cache, which returns an ordered list, and I thought it was unnecessary to convert the list in `SyncSubscribers` to a map, especially since on the AGW side a list works the same way as a map
2. Yep! I was waiting to make that change either here or in the agw `SyncSubscribers` PR, depending on which one gets merged first @hcgatewood

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
wangyyt1013: This (setting postgres as default) is what's done in `sqorc.GetSqlBuilder` @hcgatewood

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
wangyyt1013: I don't think it's supported by squirrel @hcgatewood

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
wangyyt1013: The storage functions don't have access to the max page size config (for the `subscriberdb` servicer), but maybe this guard can be added in the `SyncSubscribers` handler of the servicer? @hcgatewood

lte/protos/subscriberdb.proto
hcgatewood: Cool. Make the AGW change in this PR as well then, so nothing breaks in-between commits

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
hcgatewood: 👍

lte/cloud/go/services/subscriberdb/storage/sub_proto.go
hcgatewood: That's fine, just make a note of it here and then can enforce it in SyncSubsc

lte/protos/subscriberdb.proto
wangyyt1013: Yep, will have to wait until the code for `SyncSubscribers` in AGW come in from the other PR though @hcgatewood

lte/cloud/configs/subscriberdb.yml
```diff
@@ -12,6 +12,9 @@
 
 # flatDigestEnabled is a feature flag for the flat digest functionality.
 flatDigestEnabled: true
-# changesetSizeTheshold specifies the max size of the cloud-agw changeset
+# changesetSizeThreshold specifies the max size of the cloud-agw changeset
```
hcgatewood: "cloud-AGW"

lte/cloud/configs/subscriberdb.yml
```diff
@@ -12,6 +12,9 @@
 
 # flatDigestEnabled is a feature flag for the flat digest functionality.
 flatDigestEnabled: true
```
hcgatewood: While we're here: let's change this to "digestsEnabled" since this guards more than just the flat digests

orc8r/cloud/go/sqorc/builder.go
```diff
@@ -318,6 +318,26 @@ func ClearStatementCacheLogOnError(cache *squirrel.StmtCache, callsite string) {
 	}
 }
 
+// FmtConflictUpdateTarget generates the string used to refer to target value
```
hcgatewood: "used to refer to target column"

orc8r/cloud/go/sqorc/builder.go
```diff
@@ -318,6 +318,26 @@ func ClearStatementCacheLogOnError(cache *squirrel.StmtCache, callsite string) {
...
+// used in an on-conflict upsert operation; the specific syntax depends on the
+// sql dialect.
```
hcgatewood: "SQL"

orc8r/cloud/go/sqorc/builder.go
```diff
@@ -318,6 +318,26 @@ func ClearStatementCacheLogOnError(cache *squirrel.StmtCache, callsite string) {
...
+// e.g. for mariaDB, "f1=t2.f1"; for postgres, "f1=excluded.f1"
```
hcgatewood: Turn this into a list, making it clear that these are the two options we support. Also, use "t" for table and "c" for column

orc8r/cloud/go/sqorc/builder.go
```diff
@@ -318,6 +318,26 @@ func ClearStatementCacheLogOnError(cache *squirrel.StmtCache, callsite string) {
...
+func FmtConflictUpdateTarget(tableName string, colName string) string {
+	dialect, envFound := os.LookupEnv("SQL_DIALECT")
```
hcgatewood, line 327: "dialect, ok" is standard and concise enough for me

orc8r/cloud/go/sqorc/builder.go
```diff
@@ -318,6 +318,26 @@ func ClearStatementCacheLogOnError(cache *squirrel.StmtCache, callsite string) {
...
+		dialect = PostgresDialect
+	}
+
+	var upsertColumnPrefix string
+	switch strings.ToLower(dialect) {
+	case PostgresDialect:
+		upsertColumnPrefix = "excluded"
+	case MariaDialect:
+		upsertColumnPrefix = tableName
+	}
```
hcgatewood, line 341: Always include a default in switch statements. In this case it's not strictly necessary, but it protects against future code changes

Since default here always means there's a problem, you can panic. Or if that feels icky, you can just return the postgres default

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -66,41 +69,120 @@ func RenewDigests(
...
+	subStore *storage.SubStore,
+) (string, []*lte_protos.SubscriberDigestWithID, error) {
+	errs := &multierror.Error{}
+	digest, prevDigest, err := updateDigest(network, digestStore)
+	if err != nil {
+		multierror.Append(errs, err)
+	}
+
+	var perSubDigests []*lte_protos.SubscriberDigestWithID
+	if errs.ErrorOrNil() == nil {
```
hcgatewood: `errs.ErrorOrNil() == nil` is an idiosyncratic construction. Just use named single errs (e.g. `updateDigestErr`, then e.g. before your final return you can aggregate them into a multi-error

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -66,41 +69,120 @@ func RenewDigests(
...
+		return nil, errors.Wrapf(err, "get per sub dgests to update")
+	}
+	err = perSubDigestStore.SetDigest(network, perSubDigests)
+	if err != nil {
+		return nil, errors.Wrapf(err, "set per sub digest")
+	}
+	return perSubDigests, nil
+}
+
+func updateDigest(network string, digestStore storage.DigestStore) (string, string, error) {
```
hcgatewood: Add docstring indicating the ordering of the next and prev digest return values

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -66,41 +69,120 @@ func RenewDigests(
...
+	if err != nil {
+		return "", "", errors.Wrapf(err, "generate flat digest")
+	}
+	prevDigest, err := storage.GetDigest(digestStore, network)
+	if err != nil {
+		return "", "", errors.Wrapf(err, "get previous flat digest")
+	}
+	err = digestStore.SetDigest(network, digest)
+	if err != nil {
+		return "", "", errors.Wrapf(err, "set flat digest")
```
hcgatewood: Here and throughout: use `errors.Wrap` unless you have additional info to format (the `f` stands for "format", like in `Printf`)

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -66,41 +69,120 @@ func RenewDigests(
...
+	return digest, prevDigest, nil
+}
+
+func updateSubscribers(network string, subStore *storage.SubStore) error {
+	apnsByName, err := subscriberdb.LoadApnsByName(network)
+	if err != nil {
+		return err
+	}
+	// Ensure the temporary table is empty for use
+	err = subStore.InitiateUpdate()
```
hcgatewood: I think I actually like "InitializeUpdate" better

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -66,41 +69,120 @@ func RenewDigests(
...
+		return "", "", errors.Wrapf(err, "get previous flat digest")
+	}
+	err = digestStore.SetDigest(network, digest)
+	if err != nil {
+		return "", "", errors.Wrapf(err, "set flat digest")
+	}
+	return digest, prevDigest, nil
+}
+
+func updateSubscribers(network string, subStore *storage.SubStore) error {
```
hcgatewood: subStore -> store since it's the only store in this fn

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -66,41 +69,120 @@ func RenewDigests(
...
+	apnsByName, err := subscriberdb.LoadApnsByName(network)
+	if err != nil {
+		return err
+	}
+	// Ensure the temporary table is empty for use
```
hcgatewood: No need for this comment -- the store's inner workings are irrelevant to this fn

lte/cloud/go/services/subscriberdb_cache/worker_test.go
```diff
@@ -141,30 +154,130 @@ func TestSubscriberdbCacheWorker(t *testing.T) {
...
 
 	allNetworks, err = storage.GetAllNetworks(digestStore)
 	assert.NoError(t, err)
 	assert.Equal(t, []string{"n2"}, allNetworks)
 	clock.UnfreezeClock(t)
 }
+
+// TestUpdateSubProtosByNetworkNoChange checks that, given there's no error in
+// digest generation for the network, subscribers cache is only updated when
+// the newly generated flat digest is different from the previous digest
```
hcgatewood: Nit: add trailing period

lte/cloud/go/services/subscriberdb/storage/subscribers.go
```diff
@@ -0,0 +1,302 @@
...
+			Exec()
+		return nil, errors.Wrapf(err, "clear sub protos tmp table")
+	}
+	_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
+	return err
+}
+
+// InsertMany inserts a batch of subscriber data into the temporary table.
+func (l *SubStore) InsertMany(network string, subProtos []*lte_protos.SubscriberData) error {
+	// NOTE: Caller of this function should guard against the case where
```
hcgatewood: Here and below: this is a weird place for this comment. Instead, add a note in the docstring that the caller should enforce some reasonable max size of subProtos, in accordance with its max page size

lte/cloud/go/services/subscriberdb/storage/subscribers.go
```diff
@@ -0,0 +1,302 @@
...
+	_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
+	return err
+}
+
+// ApplyUpdate applies all subscriber data updates onto the db table and completes the batch update.
+func (l *SubStore) ApplyUpdate(network string) error {
+	txFn := func(tx *sql.Tx) (interface{}, error) {
+		// HACK: hard coding part of this sql query because there currently doesn't exist good support
+		// for "WHERE (row NOT IN other_table)" with squirrel
+		_, err := l.builder.
```
hcgatewood, line 132: Similar to a previous comment: add a comment that gives a summary of the SQL that this builder actually generates, for easy skimming

lte/cloud/go/services/subscriberdb/storage/subscribers.go
```diff
@@ -0,0 +1,302 @@
...
+				)),
+			}).
+			RunWith(tx).
+			Exec()
+		if err != nil {
+			return nil, errors.Wrapf(err, "clean up previous sub proto store table")
+		}
+
+		// The upsert query should look something like
+		// "INSERT INTO subscribers
```
hcgatewood: No need to add quotes here and below

lte/cloud/go/services/subscriberdb/storage/subscribers.go
```diff
@@ -0,0 +1,302 @@
...
+		subProtos = append(subProtos, subProto)
+	}
+	err := rows.Err()
+	if err != nil {
+		return nil, errors.Wrap(err, "get sub protos from store, SQL rows error")
+	}
+	return subProtos, nil
+}
+
+// getNextPageToken returns the next page token based on the lastIncludedEntity in the current page.
```
hcgatewood: Throughout: wrap docstrings to 79 chars

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -207,3 +235,26 @@ func loadAPNs(gateway *protos.Identity_Gateway) (map[string]*lte_models.ApnConfi
 
 	return apnsByName, apnResources, nil
 }
+
+// appendApnResourcesToSubProtos adds the gateway-specific apn resources data to subscriber protos before
+// returning to AGWs.
+func appendApnResourcesToSubProtos(subProtos []*lte_protos.SubscriberData, apnResources lte_models.ApnResources) []*lte_protos.SubscriberData {
```
hcgatewood: Maybe "injectAPNResources"

hcgatewood: Rather than taking the apnResources as an arg, can the fn load the apnResources itself?

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -76,6 +85,10 @@ func (s *subscriberdbServicer) SyncSubscribers(
 	ctx context.Context,
 	req *lte_protos.SyncSubscribersRequest,
 ) (*lte_protos.SyncSubscribersResponse, error) {
```
hcgatewood, line 87: This servicer method is getting just a bit long. Can you move the "get sub changeset" logic to its own method?

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -100,19 +113,23 @@ func (s *subscriberdbServicer) SyncSubscribers(
...
 	}
+	// Since the cached sub protos don't contain gateway-specific apn resources data, need to
+	// append that to each sub proto in this handler
+	subProtosToRenew = appendApnResourcesToSubProtos(subProtosToRenew, apnResourcesByAPN)
 
 	res := &lte_protos.SyncSubscribersResponse{
 		FlatDigest:    &lte_protos.Digest{Md5Base64Digest: flatDigest},
 		PerSubDigests: cloudPerSubDigests,
-		ToRenew:       subProtosById,
+		ToRenew:       subProtosToRenew,
```
hcgatewood: "subProtosToRenew" -> "renewed", because that var actually does contain the renewed values, not just a list of which subs should be renewed

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -139,15 +156,31 @@ func (s *subscriberdbServicer) ListSubscribers(ctx context.Context, req *lte_pro
 	if err != nil {
 		return nil, err
 	}
-	subProtos, nextToken, err := subscriberdb.LoadSubProtosPage(req.PageSize, req.PageToken, networkID, apnsByName, apnResourcesByAPN)
-	if err != nil {
-		return nil, err
+
+	var subProtos []*lte_protos.SubscriberData
```
hcgatewood, line 155: Encapsulate this new block you added into its own method

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -207,3 +235,26 @@ func loadAPNs(gateway *protos.Identity_Gateway) (map[string]*lte_models.ApnConfi
 
 	return apnsByName, apnResources, nil
 }
+
+// appendApnResourcesToSubProtos adds the gateway-specific apn resources data to subscriber protos before
+// returning to AGWs.
+func appendApnResourcesToSubProtos(subProtos []*lte_protos.SubscriberData, apnResources lte_models.ApnResources) []*lte_protos.SubscriberData {
+	for i := range subProtos {
+		subProto := subProtos[i]
+		if subProto.GetNon_3Gpp() == nil || subProto.Non_3Gpp.GetApnConfig() == nil {
```
hcgatewood: This is where the guarded getters are nice -- just do `subProto.GetNon_3Gpp().GetApnConfig()`

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -207,3 +235,26 @@ func loadAPNs(gateway *protos.Identity_Gateway) (map[string]*lte_models.ApnConfi
...
+			continue
+		}
+
+		for j := range subProto.Non_3Gpp.ApnConfig {
+			apnKey := subProto.Non_3Gpp.ApnConfig[j].ServiceSelection
+			var apnResource *lte_protos.APNConfiguration_APNResource
```
hcgatewood: This block can be simplified I think -- no need to create the nil `var` declaraction, just assign directly in the if block

hcgatewood (changes requested):
Couple more small things then I think we'll be good to go

orc8r/cloud/go/sqorc/builder.go
wangyyt1013: "dialect, envFound" is used in the `GetSqlBuilder` function in the same file; do you suggest I prioritize consistency or conciseness? @hcgatewood

hcgatewood: Ah I see 🙂 In that case either is fine. Just make them both the same

lte/cloud/go/services/subscriberdb/load.go
```diff
@@ -80,21 +80,19 @@ func LoadSubProtosByID(
 		lc, serdes.Entity,
 	)
 	if err != nil {
-		return nil, errors.Wrapf(err, "Load added/modified subscriber entities")
+		return nil, errors.Wrap(err, "Load added/modified subscriber entities")
```
hcgatewood: Lowercase errors
```suggestion
		return nil, errors.Wrap(err, "load added/modified subscriber entities")
```

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -207,3 +261,27 @@ func loadAPNs(gateway *protos.Identity_Gateway) (map[string]*lte_models.ApnConfi
...
+		if subProto.GetNon_3Gpp().GetApnConfig() == nil {
+			continue
+		}
+		for j := range subProto.Non_3Gpp.ApnConfig {
+			apnKey := subProto.Non_3Gpp.ApnConfig[j].ServiceSelection
+			if apnResourceModel, ok := apnResources[apnKey]; ok {
+				subProto.Non_3Gpp.ApnConfig[j].Resource = apnResourceModel.ToProto()
+			}
+		}
+		subProtos[i] = subProto
```
hcgatewood: Reading and setting subProto like this is idiosyncratic. Try "for i, subProto := range subProtos", then just edit the subProto in-place (it's a ptr, so no need to add it back to the array)

hcgatewood (approved):
Structure's correct, code looks beautiful. Fix two nits then msg me to merge

lte/protos/subscriberdb.proto
```diff
@@ -242,7 +242,7 @@ message SyncSubscribersResponse {
   repeated SubscriberDigestWithID per_sub_digests = 3;
   // to_renew is the set of all subscriber data modified/added on the server side,
   // indexed by their IDs (IMSI).
```
ardzoht: update the description to reflect the change as this is not indexed by IMSI anymore

ardzoht (approved):
minor nit for a comment, AGW changes LGTM

lte/protos/subscriberdb.proto
wangyyt1013: Thanks for the catch!

PR #8120, opened 2021-07-15
test(sessiond): Use shard class and write unit test

veshkemburu:
## Summary

Change Shard Structure to store IMSI information as well.
Initialize shard object in sessiond_main and pass it through local enforcer.
SessionState only really initialized in 2 places(so initialize shard id there and add). 

## Test Plan

Investigated where to place UEShard object, and how to assign UE Sessions the appropriate shard ID.
Using templates from test_session_store, wrote unit test that shards a large number of UEs and ensures that all sessions within a particular UE receive the same shard ID. 

## Additional Information

NOTE: Implementation does not contain state recovery(when pipelined resets), which will be added in a later PR


lte/gateway/c/session_manager/SessionStore.cpp
```diff
@@ -148,9 +150,16 @@ SessionMap SessionStore::read_sessions_for_deletion(const SessionRead& req) {
...
 bool SessionStore::create_sessions(
     const std::string& subscriber_id, SessionVector sessions) {
-  auto session_map           = SessionMap{};
+  auto session_map = SessionMap{};
+  int shard_id     = ue_shards_->add_ue();
+  for (size_t i = 0; i < sessions.size(); i++) {
+    sessions[i]->set_shard_id(shard_id);
+  }
+  MLOG(MINFO) << "UE with IMSI: " << subscriber_id
+              << " and shard_id: " << shard_id;
```
uri200: Note this breaks abstraction of Session Store. Session store should be dealing with storage of session, not defining fields. 

One place it could be added is on [update_tunnel_ids](http://github.com/magma/magma/blob/7441d3f1467d142e0f01f101c0e5c0a7a1228592/lte/gateway/c/session_manager/LocalEnforcer.cpp#L1023-L1023). That functions does many changes on sessions. 

This should be added before `handle_session_activate_rule_updates`, that is where the rules are sent to pipelined to be installed (and pipelined will need to know the shard id on activation)

veshkemburu: I'll be moving the shard object out of SessionStore and into LocalEnforcer. It makes sense to just stick the relevant operations in LocalEnforcer and avoid SessionStore entirely.

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -72,6 +72,7 @@ LocalEnforcer::LocalEnforcer(
       spgw_client_(spgw_client),
       aaa_client_(aaa_client),
       session_store_(session_store),
+      shards_(std::make_shared<ShardTracker>()),
```
koolzz: Should we be initializing this in the sessiond_main and passing it?

How are we accessing this struct from the pollerthread class if its defined here

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1051,6 +1052,15 @@ bool LocalEnforcer::update_tunnel_ids(
     session->receive_monitor(monitor, nullptr);
   }
 
+  // iterate through sessions in session map and update shard ids
+  SessionMap::iterator iter;
+  for (iter = session_map.begin(); iter != session_map.end(); iter++) {
+    int shard_id = shards_->add_ue();
```
koolzz: This seems like it will just reassign every ue a new shard, but a UE should already have a shard associated with it. Can we reconstruct state here?

lte/gateway/c/session_manager/LocalEnforcer.cpp
veshkemburu: So sessiond_main does have a local enforcer object, and sessiond_main can access fields in there like the shards_ object if needed.

lte/gateway/c/session_manager/LocalEnforcer.cpp
veshkemburu: It'll assign every UE _session_ a new shard. The reason I do it here is because there is no object definition of a UE and I want to avoid giving each session it's own shard. It would be ideal for a UE to already have a shard associated with it, but I didn't see the place to make the addition(Oriol suggested the update_tunnel_ids function, which is why I've put it in here)

koolzz: Don't we save it in session state
marshaled["shard_id"]          = stored.shard_id;

koolzz: We need to recover what we previously assigned its a hard requirement, or pipelined state will get recked

veshkemburu: Yeah we do, but it's never assigned. Assignment happens when we add a UE. I thought if I assigned it there that it would add a "new UE" per session instead of per UE, which is why I didn't do marshaled["shard_id"] = add_ue();

uri200: This looks wrong. `session_map` contains all existing sessions in the system.  Why do we have to change shard id 

Plus if the idea is to maintain this in both sessiond and pipelined, that means that if we have 1000 sessions, we will have to update 1000 ues on sessiond and then send an update to pipelined with to inform of the new shard id. 

Update Tunnel applies to one single session (defined at `auto& session = **session_it;`). But this functions seems to iterate over the whole session map.

veshkemburu: You are right. The idea was to set shard id after session creation because of the issue of where to put the shard object and avoiding creating duplicate entries for multiple sessions in a single UE. I thought this would be a good place to set the shard id. I'm working on a new approach.

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
```diff
@@ -96,6 +96,7 @@ bool SessionStateEnforcer::m5g_init_session_credit(
     const std::string& session_id, const SessionConfig& cfg) {
   /* creating SessionState object with state CREATING
    * This calls constructor and allocates memory*/
+  // TODO(veshkemburu): Figure out how to handle sharding(object)
   auto session_state =
       std::make_unique<SessionState>(imsi, session_id, cfg, *rule_store_);
```
koolzz, line 101: are you missing a 0 in the end there?

koolzz: (placeholder for 0 shard)

lte/gateway/c/session_manager/LocalEnforcer.cpp
veshkemburu: I'll be handling state recovery in a separate PR, but it should be easy to add in with this approach.

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
veshkemburu: No need for a 0 here because I can just initialize shard id to be 0 within the class itself. It's a different constructor

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,58 @@
...
-  * representing a single UE at the new shard
-  */
-  if (ue_count_per_shard_.size() == 0) {
-    ue_count_per_shard_.push_back(1);
-    return 0;
+int ShardTracker::add_ue(std::string imsi) {
+  /**
+   * If there are no shards, add a single entry set
+   * to shards list containing the IMSI
+   * */
```
uri200: Comments should be used when the code is not clear enough. But in this case code is self explanatory

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,58 @@
...
+  if (shards_.empty()) {
+    std::set<std::string> new_shard = {imsi};
+    shards_.push_back(new_shard);
```
uri200: shouldn't this be part of the constructor?

uri200: I still don't understand well this object but looks like it may use singleton pattern
https://www.tutorialspoint.com/Explain-Cplusplus-Singleton-design-pattern
https://stackoverflow.com/questions/1008019/c-singleton-design-pattern

Is ShardTracker a set of buckets where to place IMSIS? Note a design document with pictures  would help a lot to visual

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,58 @@
...
-      ue_count_per_shard_[shard_id]++;
+
+  for (size_t shard_id = 0; shard_id < shards_.size(); shard_id++) {
+    /**
+     * If the UE is already in the shard, return the shard id. This check
+     * is meant to avoid multiple sessions for a UE being assigned duplicate
+     * entries.
+     * */
+    if (shards_[shard_id].find(imsi) != shards_[shard_id].end()) {
+      return shard_id;
```
uri200: Aren't we storing shard id on session_map. If the UE already exists on session_map, we can get the shard id from there

In case we do not store anymore. Does it means every time we have to add an shard we have to go through all the shards in the systen?

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,58 @@
...
+    }
+    /**
```
uri200: this is my main concern on this design: it looks like the whole objective of this feature is to do groups of 100. 

But the question I still have is, what will we do with those groups? Are they going to be polled in parallel? How are those parallel workers going to functio: does it uses multithread, multiprocess? Has there been any test done to see that parallel request to OVS do not clog ovs somewhere else?

What happen if you have 20 groups of 100? Will that mean you will poll one group each 10 seconds, which is equivalent to set polling every 200 seconds on current code as is?

uri200 (commented):
@pshelar  @koolzz  I **strongly** recommend to have a design stating things like
- What is the issue
- Where is the issue (is it in OVS, is it in pipelined): look for the functions that take longer on pipelined to understand where is this clogged.
- Find or build a test case were you can reproduce this issue easily and quantify it. This should be used to compare once this is completed)
- Purpose some solutions (multithread? multiprocessing?) and try to test them ONLY with pipelined (before polluting other systems)
- If you are using shariding, do think possible sharding algorithms that may work, try to visualize how will they work (with pictures on design doc) before starting to build a shard structure.

lte/gateway/c/session_manager/ShardTracker.cpp
veshkemburu: So the object is a vector of sets, where each set holds the IMSIs associated with the shard id. I can make a diagram for this. It's just that the design wound up changing here.

lte/gateway/c/session_manager/ShardTracker.cpp
koolzz: It's easier for ovs/ryu to handle a bunch of small requests for subset of UEs than serve 1 huge one. We're going to try processing them in parallel(20 groups of 100 won't mean that we wait for one group reply to send a request for another one, we'll just attach a callback for once we recieve the stats to process the updates on sessiond)

lte/gateway/c/session_manager/ShardTracker.cpp
veshkemburu: That's a good point I can simplify this by just accessing it directly. For the second question, the answer is yes. I mentioned in the PR description that I still need to implement state recovery, and to do that I'd need to go through all the shards.

lte/gateway/c/session_manager/ShardTracker.cpp
koolzz: > @pshelar @koolzz I **strongly** recommend to have a design stating things like
> 
> * What is the issue
> * Where is the issue (is it in OVS, is it in pipelined): look for the functions that take longer on pipelined to understand where is this clogged.
> * Find or build a test case were you can reproduce this issue easily and quantify it. This should be used to compare once this is completed)
> * Purpose some solutions (multithread? multiprocessing?) and try to test them ONLY with pipelined (before polluting other systems)
> * If you are using shariding, do think possible sharding algorithms that may work, try to visualize how will they work (with pictures on design doc) before starting to build a shard structure.

those are valid points, ty for review:
I think we're confusing a few things I'm working on a broader design doc I'll share shortly but I think as this is just the skeleton for the sessiond->pipelined pull model which as **we all agreed on multiple times** we should be fine merging this.

 - The issue is to scale up stats to support large number of UEs (returning all UEs * rules at once is becoming complicated)
 - Smart polling and moving to pull stats structure are 2 different efforts(even though connected smart polling is way more involved and design is not fully fleshed out yet). 
 - Sharding allows us to process stats for smaller groups of UEs, which is very convenient for datapath. 
 - We're not polluting sessiond, sessiond is the brains of the system, pipelined can't make any smart deicions because only sessiond know how much quota the UEs has. Down the line if we want to do any smart polling and group UEs with high traffic usage or break up GX/GY polling we need to do it from sessiond because we can't change the rules on pipelined side as they're installed by sessiond.

pshelar: right. the key point is 'scaling up'. Its not about the efficiency or bottleneck of existing stats reporting scheme. we can continue to improve it along with shard based stats reports.

lte/gateway/c/session_manager/ShardTracker.cpp
veshkemburu: I will also add a small diagram to the design doc to make this clear.

lte/gateway/c/session_manager/ShardTracker.cpp
veshkemburu: I ran into some errors that I could not figure out for some reason when trying to access the shard id directly. Since this approaches work and isn't too much more inefficient(O(number of shards), and we'll probably use very few shards(as discussed in our design doc), I am going to keep this check in for now, but it'll be addressed in a future PR(probably the state recovery PR).

lte/gateway/c/session_manager/sessiond_main.cpp
```diff
@@ -315,12 +315,14 @@ int main(int argc, char* argv[]) {
 
   // Some setup work for the SessionCredit class
   set_consts(config);
+  // initialize shard tracker object
```
koolzz: rm comment

lte/gateway/c/session_manager/ShardTracker.cpp
uri200: Have you guys tried to run parallel queries in pipelined to see what happens? Does it improve performance? Can you run 2 queries in parallel without performance impact, 20, 200?  What is the impact of a shard size of 100, what about 200 or 50? Answering those questions is basic to come up with a proper sharding

Why don't you start modifying pipelined to take parallel queries and do some ground testing to see if parallel queries are feasible?

koolzz: @veshkemburu will gather that on spirient. Although I'd prefer we land this PR to avoid the unnecessary confusion

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1014,9 +1015,11 @@ void LocalEnforcer::init_session(
     const std::string& session_id, const SessionConfig& cfg,
     const CreateSessionResponse& response) {
   const auto time_since_epoch = magma::get_time_in_sec_since_epoch();
-  auto session                = std::make_unique<SessionState>(
+  // avoid duplicate sessions having multiple shard entries
```
uri200: this comment here is out of context. Maybe you want to move it inside add_ue

The reader, if interested will see what happen when IMSI already exists

lte/gateway/c/session_manager/ShardTracker.cpp
uri200: @koolzz just curious: is pipelined code to run queries in parallel already landed?  I mean, do we have the capability to run simultaneous queries? Because that would help a lot

lte/gateway/c/session_manager/SessionState.h
```diff
@@ -267,6 +267,8 @@ class SessionState {
 
   std::string get_imsi() const { return config_.common_context.sid().id(); }
 
+  int get_shard_id() { return shard_id_; }
+
```
uri200, line 273: you will need to run sm_precommit on your vagrant VM for the format to be proper

lte/gateway/c/session_manager/test/SessionStateTester.h
```diff
@@ -37,7 +37,7 @@ class SessionStateTest : public ::testing::Test {
     rule_store    = std::make_shared<StaticRuleStore>();
     session_state = std::make_shared<SessionState>(
         IMSI1, SESSION_ID_1, cfg, *rule_store, tgpp_ctx, pdp_start_time,
-        CreateSessionResponse{});
+        CreateSessionResponse{}, 0);
```
uri200: why initialized to 0?  What is the point to initialize if we don't even have a value yet? You can just default it in the constructuor

lte/gateway/c/session_manager/SessionState.h
```diff
@@ -732,6 +734,7 @@ class SessionState {
  private:
   std::string imsi_;
   std::string session_id_;
+  int shard_id_;
```
uri200: when is this initialized with the actual value of the shard (not 0)

uri200:
please take a look at this implementation 

NOTE code it IS NOT proper and there are many things left to make it work. But it can be used as start point

https://github.com/magma/magma/pull/8236

lte/gateway/c/session_manager/ShardTracker.cpp
koolzz: yeah I've landed it, we're sending sending a ovs req for specific shard (whenever we get a grpc req from pipelined) and whenever we get the result back we reply)

uri200: What about duplicating the query you receive n times and run them in parallel to see the performance?

koolzz: no test results yet unfortunately, only been back for a few days 😛

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -64,7 +64,7 @@ LocalEnforcer::LocalEnforcer(
     std::shared_ptr<aaa::AAAClient> aaa_client,
```
themarwhal, line 64: This is just my personal preference, but I think we should consolidate the object parameters. 
`std::shared_ptr<ShardTracker> shards` can go below `aaa_client`

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -13,39 +13,37 @@
...
 /* Shards represent groups of UEs placed into buckets of
 a certain size, to make polling more manageable*/
 class ShardTracker {
-  ShardTracker();
-
+ public:
   /**
    * Add UE to shards based on availability
    * @return index(shard id) where UE was placed
-   * TODO(veshkemburu): Store IMSI as well for easier subscriber reallocation
```
themarwhal, line 27: Add documentation on what would happen if the UE already has a shardID. Does it return the existing shardID?

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -192,7 +195,7 @@ SessionState::SessionState(
     const std::string& imsi, const std::string& session_id,
     const SessionConfig& cfg, StaticRuleStore& rule_store,
     const magma::lte::TgppContext& tgpp_context, uint64_t pdp_start_time,
-    const CreateSessionResponse& csr)
+    const CreateSessionResponse& csr, const int shard_id)
```
themarwhal: Please address all `build_session_manager` warnings here and everywhere else

lte/gateway/c/session_manager/SessionState.h
themarwhal: Should this be unsigned?

lte/gateway/c/session_manager/SessionStore.h
```diff
@@ -110,15 +110,15 @@ class SessionStore {
    * storage interface.
    * @param req
    * @return Last written values for requested sessions. Returns an empty vector
-   *         for subscribers that do not have active sessions.
```
themarwhal, line 113: 💯

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,53 @@
...
-  /*
-  * Iterate through all existing shards, if all are full
-  * create a new shard with quantity 1, otherwise increment
-  * the quantity of UEs in the latest shard
-  */
-  for (size_t shard_id = 0; shard_id < ue_count_per_shard_.size(); shard_id++) {
-    if (ue_count_per_shard_[shard_id] < max_shard_size_) {
-      ue_count_per_shard_[shard_id]++;
+  for (size_t shard_id = 0; shard_id < shards_.size(); shard_id++) {
+    /**
```
themarwhal: Did we decide on using `//` for multi line comments in the last PR? Or `/**`? (Just making sure, I can't remember which :) )

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,53 @@
...
 
-int ShardTracker::add_ue() {
-  /*
-  * If there are no shards, add an entry for the shard
-  * representing a single UE at the new shard
-  */
-  if (ue_count_per_shard_.size() == 0) {
-    ue_count_per_shard_.push_back(1);
-    return 0;
+int ShardTracker::add_ue(std::string imsi) {
```
themarwhal: Should sharID be unsigned?

themarwhal (changes requested):

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -13,39 +13,37 @@
...
-   * (GH8167)
    */
-  int add_ue();
+  int add_ue(std::string imsi);
```
themarwhal: uint here for the return value?

Also, can this function ever fail to add a UE? if there is a case, would be good to document and update the return type.

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -13,39 +13,37 @@
...
-  */
-  std::vector<uint16_t> ue_count_per_shard_;
   /*
-  * largest number of UEs that can fill a shard
-  */
+   * a vector of quantities, where the indices represent
+   * the shard id and the values are the UEs(IMSIs) assigned
+   * to that shard id
+   */
+  std::vector<std::set<std::string>> shards_;
```
themarwhal: imsis_per_shard_

lte/gateway/c/session_manager/test/SessionStateTester.h
themarwhal: ShardID allocation should happen outside of SessionState I think. For testing, let's use a non-0 value though.

lte/gateway/c/session_manager/test/test_local_enforcer.cpp
```diff
@@ -56,9 +56,10 @@ class LocalEnforcerTest : public ::testing::Test {
     aaa_client           = std::make_shared<MockAAAClient>();
     events_reporter      = std::make_shared<MockEventsReporter>();
     auto default_mconfig = get_default_mconfig();
+    auto shards          = std::make_shared<ShardTracker>();
```
themarwhal:
```suggestion
    auto shard_tracker        = std::make_shared<ShardTracker>();
```

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1014,9 +1015,11 @@ void LocalEnforcer::init_session(
     const std::string& session_id, const SessionConfig& cfg,
     const CreateSessionResponse& response) {
   const auto time_since_epoch = magma::get_time_in_sec_since_epoch();
-  auto session                = std::make_unique<SessionState>(
+  // avoid duplicate sessions having multiple shard entries
+  int shard_id = shards_->add_ue(imsi);
```
themarwhal: Could shard allocation ever fail?

lte/gateway/c/session_manager/SessionState.h
veshkemburu: I can make it unsigned, wasn't aware of the specifications for declaring variables, but from what I've seen it makes sense to make variables whose values are solely going to positive or zero unisgned.

lte/gateway/c/session_manager/LocalEnforcer.cpp
veshkemburu: I stuck the parameters in the end to make debugging easier especially when local enforcer objects are initialized in test files. This also makes sense though.

lte/gateway/c/session_manager/ShardTracker.cpp
veshkemburu: Changed to // comments for multilines(except for function docs).

lte/gateway/c/session_manager/test/SessionStateTester.h
veshkemburu: So the reason was we wanted to add the ue before intitializing the session. If you look at local_enforcer->init_session I add the UE then set the shard id for the session. I may change this approach a bit, based on the template Oriol provided.

lte/gateway/c/session_manager/SessionState.h
veshkemburu: It's initialized when calling init session in local enforcer. I pass the shard id in through there.

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1014,9 +1016,10 @@ void LocalEnforcer::init_session(
     const std::string& session_id, const SessionConfig& cfg,
     const CreateSessionResponse& response) {
   const auto time_since_epoch = magma::get_time_in_sec_since_epoch();
+  int shard_id                = shard_tracker_->add_ue(imsi);
```
themarwhal: shouldn't this be uint16_t?

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,47 @@
...
-  * create a new shard with quantity 1, otherwise increment
-  * the quantity of UEs in the latest shard
-  */
-  for (size_t shard_id = 0; shard_id < ue_count_per_shard_.size(); shard_id++) {
-    if (ue_count_per_shard_[shard_id] < max_shard_size_) {
-      ue_count_per_shard_[shard_id]++;
+  for (size_t shard_id = 0; shard_id < imsis_per_shard_.size(); shard_id++) {
+    // If the UE is already in the shard, return the shard id. This check
+    // is meant to avoid multiple sessions for a UE being assigned duplicate
+    // entries.
```
themarwhal, line 31: Assigning a reference to the set a variable might make this cleaner to read.
Something like
```c++
std::set<std::string>& imsis = imsis_per_shard_[shard_id];
```

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,47 @@
...
-  * If there are no shards, add an entry for the shard
-  * representing a single UE at the new shard
-  */
-  if (ue_count_per_shard_.size() == 0) {
-    ue_count_per_shard_.push_back(1);
-    return 0;
+uint16_t ShardTracker::add_ue(std::string imsi) {
+  if (imsis_per_shard_.empty()) {
+    std::set<std::string> new_shard = {imsi};
+    imsis_per_shard_.push_back(new_shard);
```
themarwhal:
```suggestion
    imsis_per_shard_.push_back({imsi});
```

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,47 @@
...
+      imsis_per_shard_[shard_id].insert(imsi);
       return shard_id;
     }
   }
-  ue_count_per_shard_.push_back(1);
-  return ue_count_per_shard_.size() - 1;
+  // If all shards are filled, add a new shard and insert the UE,
+  // return it's index
+  std::set<std::string> new_shard = {imsi};
+  imsis_per_shard_.push_back(new_shard);
```
themarwhal:
```suggestion
  imsis_per_shard_.push_back({imsi});
```

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,47 @@
...
 }
 
-bool ShardTracker::remove_ue(int shard_id) {
-  /*
-  * Since we only keep global state of all UEs, we just
-  * need to decrement the number of UEs at a particular id
-  * if there are no UEs at the shard, removal should fail
-  */
-  if (ue_count_per_shard_.empty()) {
+bool ShardTracker::remove_ue(std::string imsi, int shard_id) {
```
themarwhal: should be unsigned

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,47 @@
...
+  // Check if there are any shards, any UEs at a particular shard,
+  // and whether the UE is actually part of the shard, before removal
+  if (imsis_per_shard_.empty() || imsis_per_shard_[shard_id].empty() ||
```
themarwhal: Isn't the second clause covered in the third clause?

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -13,39 +13,39 @@
...
-  int add_ue();
+  uint16_t add_ue(std::string imsi);
 
   /**
    * Remove UE from shard
    * @param shard_id location of UE to be removed
    * @return true for successful removal, false for failed removal
    */
-  bool remove_ue(int shard_id);
+  bool remove_ue(std::string imsi, int shard_id);
```
themarwhal: shard_id should be unsigned right?

lte/gateway/c/session_manager/test/test_local_enforcer.cpp
```diff
@@ -3561,6 +3561,53 @@ TEST_F(LocalEnforcerTest, test_receiving_stats_for_subset_of_rules) {
...
+    EXPECT_EQ(session_map[imsi_id].size(), session_count);
+    session_store->create_sessions(imsi_id, std::move(sessions));
+    sessions.clear();
+  }
+  // check shards for all 501 UEs to make share there are 6, all sessions should
+  // have the same shard id per IMSI
+  for (int i = 1; i <= 501; i++) {
+    // check the sessions to make sure they have the right shard id and all the
+    std::stringstream imsiStream;
+    imsiStream << "IMSI" << std::string(16 - std::to_string(i).length(), '0')
```
themarwhal: `imsiStream` -> `imsi_stream`, but neat! didn't know std::string could do that.

themarwhal (changes requested):
another round of reviews. added some new comments

lte/gateway/c/session_manager/ShardTracker.h
veshkemburu: Yep and both should be constant. Thanks for the catch.

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1014,9 +1016,10 @@ void LocalEnforcer::init_session(
     const std::string& session_id, const SessionConfig& cfg,
     const CreateSessionResponse& response) {
   const auto time_since_epoch = magma::get_time_in_sec_since_epoch();
+  unsigned shard_id           = shard_tracker_->add_ue(imsi);
```
uri200: I think the placement on the template of add_ue is fine (inside the update_tunnelId). Right before installing the flows. There is the very first time you need this information.


Plus if you put it before you will need to clear it from the shard_tracker in case of failure

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1014,9 +1016,10 @@ void LocalEnforcer::init_session(
...
   auto session                = std::make_unique<SessionState>(
       imsi, session_id, cfg, *rule_store_, response.tgpp_ctx(),
-      time_since_epoch, response);
+      time_since_epoch, response, shard_id);
```
uri200: lets use set/get instead of modifying the signature of the constructor.  

At this point the flows are not even installed. No need to trigger a procedure that may not be needed. Plus if we ever use rebalancing, we will be triggering a rebalance to early with a UE that does not have  any flows installed.

If you want you can initialize shard id to -1 in the constructor in case we ever need to know if it was already initialized

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -133,6 +135,7 @@ SessionState::SessionState(
     const StoredSessionState& marshaled, StaticRuleStore& rule_store)
     : imsi_(marshaled.imsi),
       session_id_(marshaled.session_id),
+      shard_id_(marshaled.shard_id),
```
uri200, line 138: per the `-Wreorder` you may need to initialize this after other variables.
In this case i would set it to `-1`


https://stackoverflow.com/questions/1828037/whats-the-point-of-g-wreorder

lte/gateway/c/session_manager/SessionState.cpp
```diff
@@ -225,7 +229,8 @@ SessionState::SessionState(
       config_(cfg),
       current_version_(0),
       rtx_counter_(0),
-      static_rules_(rule_store) {}
+      static_rules_(rule_store),
+      shard_id_(0) {}
```
uri200: if 0 is a valid shard, i would not initialize it to that value.

It is better to have one single place were the values are set. Even if you are setting all the UEs to the same shard for now, do it at add_shard level

lte/gateway/c/session_manager/SessionState.h
```diff
@@ -267,6 +267,8 @@ class SessionState {
 
   std::string get_imsi() const { return config_.common_context.sid().id(); }
 
+  uint16_t get_shard_id() { return shard_id_; }
```
uri200: add set_shard_id as int  the template and use that function to modify the value, no constructors

The reason to use set is the same I mentioned beforeL: it is way better to have just one single entry point to see where the property is modified. Also if we ever do rebalancing we will need to use set_shard_id anyway

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,43 @@
...
 
-int ShardTracker::add_ue() {
-  /*
-  * If there are no shards, add an entry for the shard
-  * representing a single UE at the new shard
-  */
-  if (ue_count_per_shard_.size() == 0) {
-    ue_count_per_shard_.push_back(1);
-    return 0;
+uint16_t ShardTracker::add_ue(const std::string imsi) {
```
uri200, line 27: Sorry to say it again burt I don't like this design: now we have 2 places where to maintain imsis.

What we really need either now or when we do rebalance is to know what shard a UE is at and how many UEs are per shard. 

Per the design document we are not ready yet for rebalancing, so let's not do optimizations on something we don't even want to discuss.

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,43 @@
...
+  if (imsis_per_shard_.empty()) {
+    imsis_per_shard_.push_back({imsi});
```
uri200: the very first shard should be initiated at the constructor. This line will be hit in all the requests and it is only used once.

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,43 @@
...
-  */
-  for (size_t shard_id = 0; shard_id < ue_count_per_shard_.size(); shard_id++) {
-    if (ue_count_per_shard_[shard_id] < max_shard_size_) {
-      ue_count_per_shard_[shard_id]++;
+  for (size_t shard_id = 0; shard_id < imsis_per_shard_.size(); shard_id++) {
+    // If the UE is already in the shard, return the shard id. This check
+    // is meant to avoid multiple sessions for a UE being assigned duplicate
+    // entries.
+    std::set<std::string>& imsis_at_id = imsis_per_shard_[shard_id];
+    if (imsis_at_id.find(imsi) != imsis_at_id.end()) {
```
uri200: As shown in the template we already have a way to find in O(1) if the imsi has a shard id without having to maintain a parallel list of imsis

This is inefficient, not only because you have to go through all the shards but because you will have to  go through all of them just to find out it is not there.

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -13,45 +13,43 @@
...
 }
 
-bool ShardTracker::remove_ue(int shard_id) {
-  /*
-  * Since we only keep global state of all UEs, we just
-  * need to decrement the number of UEs at a particular id
-  * if there are no UEs at the shard, removal should fail
-  */
-  if (ue_count_per_shard_.empty()) {
+bool ShardTracker::remove_ue(const std::string imsi, const uint16_t shard_id) {
```
uri200, line 45: why do you need to return a value here

uri200:
Please, use the same logic in the template to add and remove UEs to a shard. Reasons:
- Reuses information we already have and maintain in the system. No need to maintain another list of IMSIS.
- Reuses search functions we already have on the system that are efficient.
- Logic is simple and clear.


Please, build an object that hosts those two functions and that stores a vector of ints.  For now lets initialize that vector on the constructor with  a vector of 1000 ints. That will give us a total of 100x1000 = 100k UEs. We can change the way we init vector later and make that more efficient.

Create a test for that object.  Also modify one of the current tests at LocalEnforcer to check what shard value is given. 

Once you have that you will have to modify all Activate Flows calls from sessiond to include shard id.

That will allow us to do test once @koolzz is done adding parallelization un pipelined

lte/gateway/c/session_manager/LocalEnforcer.cpp
veshkemburu: On the template your add UE is within the handle_command_level_result_code_for_monitors method, not update_tunnel_id. Are you talking about the one from before?

lte/gateway/c/session_manager/ShardTracker.cpp
veshkemburu: It's just in case remove fails somehow(when invoked). it'll be useful for debugging any errors and the check can be removed later

veshkemburu:
> Please, use the same logic in the template to add and remove UEs to a shard. Reasons:
> 
> * Reuses information we already have and maintain in the system. No need to maintain another list of IMSIS.
> * Reuses search functions we already have on the system that are efficient.
> * Logic is simple and clear.
> 
> Please, build an object that hosts those two functions and that stores a vector of ints. For now lets initialize that vector on the constructor with a vector of 1000 ints. That will give us a total of 100x1000 = 100k UEs. We can change the way we init vector later and make that more efficient.
> 
> Create a test for that object. Also modify one of the current tests at LocalEnforcer to check what shard value is given.
> 
> Once you have that you will have to modify all Activate Flows calls from sessiond to include shard id.
> 
> That will allow us to do test once @koolzz is done adding parallelization un pipelined

I think I will keep the IMSIs since there seems to be consensus elsewhere that storing them is a good thing. But I will use a lot of your template since the logic does flow better(rather than initializing variables at unnecessary places).

lte/gateway/c/session_manager/LocalEnforcer.h
```diff
@@ -713,6 +713,29 @@ class LocalEnforcer {
...
    */
   void cleanup_dead_sessions(const RuleRecordSet dead_sessions_to_cleanup);
+
+  /**
+   * Given a UE id and session, add the session to the session map,
+   * and invoke API that adds shard to shard tracker if the UE doesn't already
+   * exist in the session map. If the UE already exists, set the shard id of
+   * the session to the same as that of another session of the UE.
+   * @param imsi id of UE
+   * @param session_map map of UE to vector of session states
```
themarwhal, line 723: Is the SessionState passed into this function included in the session map? Or does it not matter for this function?

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1016,13 +1016,44 @@ void LocalEnforcer::init_session(
...
+      time_since_epoch, response);
   session_map[imsi].push_back(std::move(session));
 }
 
+void LocalEnforcer::add_ue_to_shard(
+    const std::string imsi, SessionMap& session_map, SessionState& session) {
+  // Check if UE has other sessions and reuse shard_id if so
+  auto sm_it     = session_map.find(imsi);
+  auto& sessions = sm_it->second;
+  for (auto it = sessions.begin(); it != sessions.end(); ++it) {
```
themarwhal:
```suggestion
  for (auto session_it = sessions.begin(); session_it != sessions.end(); ++session_it) {
```

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1016,13 +1016,44 @@ void LocalEnforcer::init_session(
...
       imsi, session_id, cfg, *rule_store_, response.tgpp_ctx(),
-      time_since_epoch, response, shard_id);
+      time_since_epoch, response);
   session_map[imsi].push_back(std::move(session));
 }
 
+void LocalEnforcer::add_ue_to_shard(
+    const std::string imsi, SessionMap& session_map, SessionState& session) {
+  // Check if UE has other sessions and reuse shard_id if so
+  auto sm_it     = session_map.find(imsi);
```
themarwhal, line 1027: I think we might have to check that this is a valid iterator right?

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1016,13 +1016,44 @@ void LocalEnforcer::init_session(
...
+  auto& sessions = sm_it->second;
+  for (auto it = sessions.begin(); it != sessions.end(); ++it) {
+    if ((*it)->get_session_id() != session.get_session_id()) {
+      session.set_shard_id((*it)->get_shard_id());
+      break;
```
themarwhal: we could just return here right?

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1016,13 +1016,44 @@ void LocalEnforcer::init_session(
...
+    }
+  }
+  // if the UE doesn't have any sessions, add to shard tracker and set id
+  if (sm_it == session_map.end()) {
```
themarwhal: Let's handle this case before we iterate through the sessions + return early. I think that'll be easier to read.

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1016,13 +1016,44 @@ void LocalEnforcer::init_session(
...
+    int shard_id = shard_tracker_->add_ue(imsi);
+    session.set_shard_id(shard_id);
+  }
+}
+
+bool LocalEnforcer::remove_ue_from_shard(
+    const std::string imsi, SessionMap& session_map, SessionState& session) {
+  auto session_map_it = session_map.find(imsi);
```
themarwhal, line 1044: Let's return early if there are no sessions for the IMSI

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: I can't remember what find returns if the item is not found but let's handle cases in this order:
1. IMSI not in session map -> Is this illegal?
2. Found a different session in session map for the IMSI -> assign shard to new session
3. No other sessions found for the IMSI -> allocate new shard

lte/gateway/c/session_manager/SessionState.h
```diff
@@ -267,6 +267,10 @@ class SessionState {
...
```
themarwhal: This should be a const function ^ see get_imsi above

lte/gateway/c/session_manager/LocalEnforcer.h
veshkemburu: It is included in session map, this is just adding it to the shard. It has to be included in order to set the shard id.

lte/gateway/c/session_manager/LocalEnforcer.cpp
veshkemburu: So I can add a check, but there should be no need to check because adding the UE to shard should happen after the session has already been added to the map. To answer those questions:
1. It's illegal.
2. For loop addresses it
3. The last if condition performs the allocation.
I can add further documentation to make all this more clear.

veshkemburu: My logic for this part is actually incorrect. 
1. It's illegal so shouldn't even be checked.
2. Addressed in for loop
3. Should be addressed at end, but I incorrectly checked if IMSI exists in the map(it always does).

lte/gateway/c/session_manager/LocalEnforcer.cpp
veshkemburu: Shouldn't be possible(could add the case for debugging purposes, not sure what the behavior would be).

lte/gateway/c/session_manager/LocalEnforcer.cpp
veshkemburu: So it shouldn't ever get called if there are no sessions for the IMSI(since it's dependent on the session existing in the map during invocation in update tunnel ids).

themarwhal (approved):

PR #8123, opened 2021-07-16
[component: agw] Fix for for getenvoy-deb support

panyogesh:
Fix:
  Updated the Vagrant file version

Test:
  Basic Vagrant bring up and make run

## Additional Information


tmdzk (changes requested):
Hello, 
Thanks for your PR, however that won't work as the version you're pointing to is an Ubuntu VM, which needs a different Vagrant configuration/provisioning from the Debian VM.

panyogesh:
Closing the PR.

PR #8124, opened 2021-07-16
[nms,agw,orc8r] Adding MME_IP to eNB configs

hakeen0618:
Signed-off-by: HakeenHan hanhufeng@baicells.com
## Summary

- Adding MME_IP value to eNB  configs
- MME_IP parameter is used to set enb  in the nms UI .

## Test Plan

- make precommit
- running locally with eNB configured manually.

## Additional Information

This adds mme_ip field for managed configs. however we don't have any of these registered yet on any of the environments so we should be good.


hakeen0618:
NMS: 
# #8124 
Screenshot from 2021-07-16 01-06-53

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -2286,6 +2286,10 @@ definitions:
           - 'Baicells ID TDD/FDD'
           - 'NuRAN Cavium OC-LTE'
         x-nullable: false
+      mme_ip:
+        type: string
+        format: ipv4
+        example: '10.0.2.1'
```
uri200, line 2292: maybe you can set a default value to empty string
```
default: ""
```

amarpad:
@ardzoht why can't this just be derived from the AGW?

lte/gateway/python/magma/enodebd/device_config/configuration_init.py
```diff
@@ -199,6 +199,8 @@ def _get_enb_config(
             tac = enb_config.tac
             bandwidth_mhz = enb_config.bandwidth_mhz
             cell_id = enb_config.cell_id
+            mme_address = enb_config.mme_ip
+            mme_port = 36412
```
ardzoht: What is this value from? Are you adding configuration value for this as well?

nms/app/packages/magmalte/app/components/lte/AddEditEnodebDialog.js
```diff
@@ -83,6 +83,9 @@ export default function AddEditEnodebDialog(props: Props) {
   );
   const [serialId, setSerialId] = useState<string>(editingEnodeb?.serial || '');
   const [tac, setTac] = useState(String(editingEnodeb?.config.tac || ''));
+  const [mme_ip, setMmeIP] = useState(
+    String(editingEnodeb?.config.mme_ip || '10.0.2.1'),
```
ardzoht: Defaulting to @andreilee for this, but it seems iffy to me to define a hard coded string here, do we have a constants file for this to be defined?

ardzoht (changes requested):

ardzoht:
Requested some changes, may I ask what is the long term plan to add this config value? With this change we will add NMS support to stream this value and display it, but it doesn’t change any of the actual device config, is this planned for configuring managed radios?

if you could add some context on the Summary section that would be good to follow up from there, thanks!

amarpad (changes requested):
MME IP should just be derived from the AGW the radio is getting attached to.

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
hakeen0618: This example is only use in swagger api response doc, and I set it reference  the ip_addr configuration as following picture in the history.
ip

nms/app/packages/magmalte/app/components/lte/AddEditEnodebDialog.js
hakeen0618: ok , i will make changes and git push.
mme_nms_default

lte/gateway/python/magma/enodebd/device_config/configuration_init.py
hakeen0618: yes, I  use DEFAULT_S1_PORT and replace it.
mme_hardcode

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
hakeen0618: > Requested some changes, may I ask what is the long term plan to add this config value? With this change we will add NMS support to stream this value and display it, but it doesn’t change any of the actual device config, is this planned for configuring managed radios?
> 
> if you could add some context on the Summary section that would be good to follow up from there, thanks!

OK, I will add the Summary section.

sourabh-nanoti:
> Requested some changes, may I ask what is the long term plan to add this config value? With this change we will add NMS support to stream this value and display it, but it doesn’t change any of the actual device config, is this planned for configuring managed radios?
> 
> if you could add some context on the Summary section that would be good to follow up from there, thanks!
@ardzoht I think the enodebd has been changed so that TR069 sends this parameter to the ENB 
@hakeen0618  Can you please add some clarification here as requested by @ardzoht ? Thanks.

sourabh-nanoti:
> MME IP should just be derived from the AGW the radio is getting attached to.

cc  @talkhasib

talkhasib:
I agree that MME IP should be derived from the AGW managing the eNB. Not through NMS.
But, there is value in this if we ever end up moving enodebd to cloud. Thus, MME IP is different for enodebd host.

talkhasib:
But, in the current context. This is not necessary and adds confusion.
There are no plans to move enodebd out of the AGW. Thus, I support closing this PR.
This was a good exercise for the authors to get a parameter added e2e. Lets discuss what other parameters make better sense.

talkhasib (changes requested):
Lets close this PR

PR #8130, opened 2021-07-16
feat(sessiond): Refactor Pipelined poll stats API to poll for shards

veshkemburu:
## Summary

The current pull model polls based on cookie and cookie mask. We want to poll based on shards. Let's change the API
to poll by shards on sessiond side. Pipelined will automatically map each shard id to cookie mask. 

Once PRs #8003 and #8120 are approved/merged, this PR will be updated and ready for review.

## Test Plan

Modified API definitions in sessiond side. 
Had API take in vector of shard ids and send a request for each id.
Modified sessiond tests and reran both sessiond and pipelined tests.
Might need to write an additional unit test


lte/gateway/python/magma/pipelined/app/enforcement.py
```diff
@@ -173,7 +173,7 @@ def _get_rule_match_flow_msgs(self, imsi, msisdn: bytes, uplink_tunnel: int, ip_
                     imsi, msisdn, uplink_tunnel, ip_addr, apn_ambr, flow, rule_num, priority,
                     rule.qos, rule.hard_timeout, rule.id, rule.app_name,
                     rule.app_service_type, self.next_main_table,
-                    version, self._qos_mgr, self._enforcement_stats_tbl, rule.he.urls))
+                    version, self._qos_mgr, self._enforcement_stats_tbl, shard_id, rule.he.urls))
```
koolzz, line 176: put shard_id in the end

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -581,7 +576,11 @@ def get_stats(self, cookie: int = 0, cookie_mask: int = 0):
             self.logger.error("Could not initialize datapath for stats retrieval")
             return RuleRecordTable()
         parser = self._datapath.ofproto_parser
-        message = parser.OFPFlowStatsRequest(datapath=self._datapath, cookie = cookie, cookie_mask = cookie_mask)
+        if (cookie == -1):
```
koolzz: hm what is this for? If you don't want to match on cookie you can just give 0 for cookie_mask

lte/gateway/python/magma/pipelined/rpc_servicer.py
```diff
@@ -354,7 +354,8 @@ def _install_flows_gy(
         self._update_version(request, ip_address)
         # Install rules in enforcement stats
         enforcement_stats_res = self._activate_rules_in_enforcement_stats(
-            request.sid.id, request.msisdn, request.uplink_tunnel, ip_address, request.apn_ambr,
+            request.sid.id, request.shard_id,
```
koolzz: shard id seems to be in wrong place

lte/gateway/python/scripts/pipelined_cli.py
```diff
@@ -156,6 +156,7 @@ def activate_flows(client, args):
             ),
         ],
         request_origin=RequestOriginType(type=RequestOriginType.GX),
+        shard_id=args.shard_id,
```
koolzz, line 159: is shard_id defined as on of the args?

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
veshkemburu: According to Pravin, getting all flows would be not passing anything at all. I assumed since cookie mapped directly to shard that I needed to not pass anything to poll all shards.

veshkemburu: So when I asked Pravin about polling all shards he mentioned that it meant not passing in any cookie value at all. Since a shard id of 0 would ideally be directly linked to a cookie value of 0, I assumed that meant I should just pass in nothing.

themarwhal:
Does this PR need to be rebased onto master?

veshkemburu:
> Does this PR need to be rebased onto master?

Yep, haven't done it yet :) But also, the main reason for the high amount of file changes is because of the pipelined test adjustments

lte/gateway/python/magma/pipelined/app/enforcement.py
veshkemburu: rule.he.urls is a default argument here and non-default arguments have to be placed before the default ones.

PR #8132, opened 2021-07-16
refactor: [ci] Refactor semantic PR check

### LINKED ISSUE #7951: Refactor: semantic PR check
we’re using https://github.com/zeke/semantic-pull-requests for enforcing semantic commits right now, but we’ve received some feedback that it’s a bit confusing
Similarly to how you improved the DCO check with a helpful message on failure, we’re wanting to use https://github.com/amannn/action-semantic-pull-request to replace the probot-based semantic commit check, and provide a helpful error message on failure
For context: https://www.conventionalcommits.org/en/v1.0.0/#summary

quentinDERORY:
Now it will comment on the PR why the check is failing

## Summary

Put semantic PR check in a workflow to enable PR comment on failure.
Clean PR comment on check failure job

## Test Plan

will test it in my repo

## Additional Information


.github/workflows/comment-pr-on-check-failure.yml
```diff
@@ -35,34 +36,32 @@ jobs:
...
+          - [About the \`signoff\` feature](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for)
+          - [Howto: sign-off most-recent commit](https://stackoverflow.com/questions/13043357/git-sign-off-previous-commits)
+          - [Howto: sign-off multiple past commits](https://gist.github.com/kwk/d70f20d17b18c4f3296d)" >> $GITHUB_ENV
+      - name: Python format comment message
         if: ${{ github.event.workflow.name == 'Python Format Check' }}
+        run: |
+          echo "MSG=Oops! Looks like you failed the \`Python Format Check\`.
+          Instructions on running the formatter locally are provided [here](https://github.com/magma/magma/tree/master/lte/gateway/docker/python-precommit)." >> $GITHUB_ENV
+      - name: Semantic PR comment message
+        if: ${{ github.event.workflow.name == 'Semantic PR' }}
```
hcgatewood, line 59: So for my understanding: does this only work if the check is a GH Actions check, or could we leave the check as the existing probot check and still test its outcome?

.github/workflows/comment-pr-on-check-failure.yml
```diff
@@ -35,34 +36,32 @@ jobs:
...
+        run: |
+          echo "MSG=Oops! Looks like you failed the \`Semantic PR check\`.
+          Instructions on formatting your PR title [here](https://www.conventionalcommits.org/en/v1.0.0/#summary)." >> $GITHUB_ENV
```
hcgatewood: 1. Use descriptive hyperlink text https://magma.github.io/magma/docs/next/contributing/contribute_conventions#all-documentation
2. Link to this instead https://magma.github.io/magma/docs/next/contributing/contribute_workflow#guidelines

.github/workflows/comment-pr-on-check-failure.yml
```diff
@@ -35,34 +36,32 @@ jobs:
...
+      - name: Comment on PR
         uses: actions/github-script@v3
         with:
           script: |
             var fs = require('fs');
             var issue_number = Number(fs.readFileSync('./NR'));
-            var msg = `Oops! Looks like you failed the \`Python Format Check\`.
-                        Instructions on running the formatter locally are provided [here](https://github.com/magma/magma/tree/master/lte/gateway/docker/python-precommit).`
+            const msg = process.env.MSG
```
hcgatewood: 1. I love this architecture of having a single workflow that handles all the "failure commenting"
2. Any thoughts on how to handle when multiple jobs fail -- should we create 1 PR comment with the n messages, or just comment with the "highest priority" failure?

(2) can be a separate PR if you decide on the former

.github/workflows/pr_bot.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
             A couple initial guidelines
             - [All commits must be signed off](https://docs.magmacore.org/docs/contributing/contribute_workflow#guidelines). This is enforced by \`DCO check\`.
-            - [All PR titles must follow the semantic commits format](https://docs.magmacore.org/docs/contributing/contribute_workflow#guidelines). This is enforced by \`Semantic Pull Request\`.
+            - [All PR titles must follow the semantic commits format](https://docs.magmacore.org/docs/contributing/contribute_workflow#guidelines). This is enforced by \`Semantic PR\`.
```
hcgatewood, line 22: 👍

.github/workflows/semantic-pr.yml
```diff
@@ -0,0 +1,103 @@
...
+    steps:
+      # Please look up the latest version from
+      # https://github.com/amannn/action-semantic-pull-request/releases
+      - uses: amannn/action-semantic-pull-request@v3.4.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # Configure which types are allowed.
+          # Default: https://github.com/commitizen/conventional-commit-types
+          types: |
```
hcgatewood, line 23: Can we remove `types` to cause it to just use the default list?

.github/workflows/semantic-pr.yml
```diff
@@ -0,0 +1,103 @@
...
+          # this option and therefore this action allows you to opt-in to using the
+          # special "[WIP]" prefix to indicate this state. This will avoid the
+          # validation of the PR title and the pull request checks remain pending.
+          # Note that a second check will be reported if this is enabled.
+          wip: true
+          # When using "Squash and merge" on a PR with only one commit, GitHub
+          # will suggest using that commit message instead of the PR title for the
+          # merge commit, and it's easy to commit this by mistake. Enable this option
+          # to also validate the commit message for one commit PRs.
+          validateSingleCommit: true
```
hcgatewood, line 92: Oh this is nice, maybe this is a reason to stay with this gh actions workflow for the check (?)

.github/workflows/semantic-pr.yml
```diff
@@ -0,0 +1,103 @@
+---
+name: "Semantic PR"
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
```
hcgatewood, line 9: What does `synchronize` indicate / can you point me to a doc with the different types? I had trouble finding one last time I looked

hcgatewood (changes requested):
Excited for this, thanks @quentinDERORY

.github/workflows/semantic-pr.yml
quentinDERORY: Here is the extensive list. https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request
Here is a shortlist with meaning https://frontside.com/blog/2020-05-26-github-actions-pull_request/
Synchronize: 	commit(s) pushed to the pull request

.github/workflows/comment-pr-on-check-failure.yml
quentinDERORY: yes we could let the bot lives if we want. I would have to revert the deletion of the config file tough

.github/workflows/comment-pr-on-check-failure.yml
quentinDERORY: 2. Glad you're mentioning this, I think I should get the reference of the last comment and update it. (Might do it in another PR)
For now it's generating one comment per check failure.

.github/workflows/semantic-pr.yml
quentinDERORY: I personally like the availability to see at a glance what is expected. This workflow file should not get longer that that anyway...

.github/workflows/comment-pr-on-check-failure.yml
quentinDERORY: tracking 2 here: https://github.com/magma/magma/issues/8165
Another issue came to mind: we currently comment each time the check failed it would be best to update the comment instead of creating a new one: https://github.com/magma/magma/issues/8164

rdefosse (approved):

.github/workflows/comment-pr-on-check-failure.yml
hcgatewood: Gotcha. The one nice thing about the existing bot is it updates ~immediately when you change the PR title, which makes it easy for me to update PR titles for contributors and have the check immediately pass

See my below comment about the `validateSingleCommit` option -- can you open a PR against the existing bot and check whether it handles this correctly? I think that will be the deciding factor on whether we should stay with probot or move to this GH Action

.github/workflows/comment-pr-on-check-failure.yml
hcgatewood: Perfect thanks @quentinDERORY

hcgatewood (changes requested):
One last request before merging (feel free to ignore if desired) -- https://github.com/magma/magma/pull/8132#discussion_r673731739

I'd prefer to stay with the probot check if possible, but the test^ I mention in that comment should be the go/no-go determination

.github/workflows/comment-pr-on-check-failure.yml
quentinDERORY: The edit trigger will do the same with this github action workflow, it will recheck on a title change just like the bot.
And for the one commit PR, the pr bot passes as it only checks the title but when merging, github does take the commit message as merge commit title.
For reference: the PR https://github.com/quentinDERORY/magma/pull/16 and what it looks like in master (my repo lol) 
image

PR #8133, opened 2021-07-16
fix(feg): multiple imports

alrs:
`feg/gateway/registry` was importing the same package, twice.


uri200 (approved):
Thanks for finding this!

PR #8138, opened 2021-07-17
chore(orc8r): Move generated API stubs to own go.mod

hcgatewood:
## Summary

Reduce build time for main Orc8r, and also makes golang version updates easier by drastically reducing the amount of generated code in the main module

These stubs are *not* meant to be used directly from the main Orc8r application code, so moving them to their own go.mod is an easy win

## Additional Information


mstre123 (approved):

electronjoe:
I think I'm just stamping this? 70k lines changed thousands of files =D

Edit: Stamping is FINE by me and an acceptable path, so long as this is automated in it's generation and it's passing tests. If it fails on merge we need more tests.

electronjoe (approved):

themarwhal (approved):
stamping, the CODEOWNER change should have a / prefix to indicate full path (probably doesn't matter in this case but I changed the rest this morning)

PR #8141, opened 2021-07-18
fix(agw): stop/start swap during provisioning to prevent error

uri200:

## Summary

When doing `vagrant provision magma` on an already provisioned magma VM, we were getting the error below.

It looks like activating the swap on the first provisioning works, but once it is activated we can not modify swap file anymore, so running provision again will fail

This PR adds swapoff and sappwon wrapping the command that enables swap. 

```
╰─ vagrant provision magma
==> magma: You assigned a static IP ending in ".1" to this machine.
==> magma: This is very often used by the router and can cause the
==> magma: network to not work properly. If the network doesn't work
==> magma: properly, try changing this IP.
==> magma: Running provisioner: shell...
    magma: Running: inline script
    magma: fallocate: fallocate failed: Text file busy
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

## Test Plan

Checked if swap is still on after provisioning 
```
root@magma-dev-focal:/home/vagrant# swapon -s
Filename				Type		Size	Used	Priority
/swapfile                              	file    	4194300	0	-2
```
## Additional Information

### DIFF
--- a/lte/gateway/Vagrantfile
+++ b/lte/gateway/Vagrantfile
@@ -28,7 +28,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.vm.box = "magmacore/magma_dev"
     config.vm.box_version = "1.1.20210618"
      # Enable Dynamic Swap Space to prevent Out of Memory crashes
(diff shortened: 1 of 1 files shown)


themarwhal (approved):
thanks for fixing this!

ulaskozat (approved):
LGTM

Issue #4908, opened 2021-02-11
[FFWA] UeContext.msisdn proto with bad format


uri200:
When running test on FFWA case we see this message  during athentication mme syslog

`'magma.lte.oai.UeContext.msisdn' contains invalid UTF-8 data when serializing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes.`

It looks like there is a bad format defined on the protos that causes the serialization to fail

Complete log:
```
sudo journalctl -fu magma@mme
Feb 11 08:52:31 agw mme[6255]: [libprotobuf ERROR google/protobuf/wire_format_lite.cc:534] String field 'magma.lte.oai.UeContext.msisdn' contains invalid UTF-8 data when serializing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes.
Feb 11 08:52:32 agw mme[6255]: [libprotobuf ERROR google/protobuf/wire_format_lite.cc:534] String field 'magma.lte.oai.UeContext.msisdn' contains invalid UTF-8 data when serializing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes.
Feb 11 08:52:32 agw mme[6255]: Received SGW IP Address 0x62800000bf50
Feb 11 08:52:32 agw mme[6255]: [DEBUG] Success logging event: attach_success
Feb 11 08:52:32 agw mme[6255]: [libprotobuf ERROR google/protobuf/wire_format_lite.cc:534] String field 'magma.lte.oai.UeContext.msisdn' contains invalid UTF-8 data when serializing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes.
```

PR #8144, opened 2021-07-18
fix(agw): Added code to correct the grpc encoding format of msisdn within ue_context

rsarwad:
Fix(agw): Added code to correct the grpc encoding format of msisdn within ue_context
This fixes the issue, https://github.com/magma/magma/issues/4908

## Summary
Issue description: While writing msisdn to Redis DB, mme was throwing encoding format error as below:
'magma.lte.oai.UeContext.msisdn' contains invalid UTF-8 data when serializing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes.

Resolution: Corrected the gRPC encoding format of msisdn while writing to DB.
## Test Plan
Executed s1ap sanity test suite and verified the changes in Tera vm by executing inbound_roaming_01.ntl test case


ardzoht (approved):
lgtm, thanks for this!

crbertoldo:
@rsarwad @rmeleromira @uri200, please do you happen to know what triggers the [issue](https://github.com/magma/magma/issues/4908)n, and what are the real impacts of the serialization failure (crash, attach failure, etc.)?

rsarwad:
> @rsarwad @rmeleromira @uri200, please do you happen to know what triggers the [issue](https://github.com/magma/magma/issues/4908), and what are the real impacts of the serialization failure (crash, attach failure, etc.)?

Hi @crbertoldo, Let me share my understanding and analysis
Having defined the data type as string in proto file, the value needs to be in UTF-8 encoded format. But with data type as byte doesn't need to have specific encoding format.
Msisdn value is received from HSS and stored in UE context, the same value is stored in Redis DB and also used while sending create session request from both spgw and sgw_s8.
Since we were seeing the error:
'magma.lte.oai.UeContext.msisdn' contains invalid UTF-8 data when serializing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes.
In spgw_state.proto file, msisdn is of type byte, https://github.com/magma/magma/blob/master/lte/protos/oai/spgw_state.proto#L71
So we changed msisdn to bytes in mme_nas_state.proto file. 

@uri200, please correct me if I am wrong. 
@crbertoldo, could you please let us know any issue with this change?

crbertoldo:
Thanks a lot, @rsarwad and @uri200!

PR #8146, opened 2021-07-19
fix(agw): restore SGi bridge flows to keep connectivity.

pshelar:
In case AGW service crash loop, the bridged mode AGW can
completely loose connectivity due to OVS reset done by
service monitoring the crash loop.
Following PR restores flows to keep the connectivity
to AGW via SGi interface.

## Test Plan
validated flow restoration on vagrant vm.

## Additional Information


ulaskozat (commented):
can we add some unit tests for this or at least include some test results against flakiness?

pshelar:
> can we add some unit tests for this or at least include some test results against flakiness?

there is not framework for UT for this type of change. let me explore it.

pshelar:
> > can we add some unit tests for this or at least include some test results against flakiness?
> 
> there is not framework for UT for this type of change. let me explore it.

I have added unit test for script using pipelineD test framework.

ulaskozat (approved):
Thanks for adding the test!

koolzz (approved):

ardzoht (approved):
lgtm

PR #8160, opened 2021-07-19
fix: Fix cwf integ test

quentinDERORY:
Use virtual test docker repo

## Summary

Fix cwf integ test.
Use virtual docker test repo instead of cwf test docker repository

## Test Plan

In master :/

## Additional Information

### DIFF
--- a/cwf/gateway/docker/.env
+++ b/cwf/gateway/docker/.env
@@ -12,7 +12,7 @@
 COMPOSE_PROJECT_NAME=cwf
 DOCKER_USERNAME=
 DOCKER_PASSWORD=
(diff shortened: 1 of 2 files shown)


themarwhal (approved):

themarwhal:
There's a cwag integ test job on PRs that get triggered with the label component:agw. Might as well merge it as it's broken but for future references.

quentinDERORY:
> There's a cwag integ test job on PRs that get triggered with the label component:agw. Might as well merge it as it's broken but for future references.

thanks for the info

uri200:
is this correct? the .env var should point to the local repository, then overwritten by install scripts. But here the address is hardcoded. Isn't this going to prevent us to run cwf from our local repository?

PR #8163, opened 2021-07-19
fix: [ci] Reverting changes to make integ test passes

quentinDERORY:

## Summary

Reverting jfrog reference as integ test use it.

## Test Plan

On PR

## Additional Information


uri200:
I think this is causing issues on master. Per what I understand after building we need to get the version from the local docker repository.

uri200 (approved):

quentinDERORY:
> I think this is causing issues on master. Per what I understand after building we need to get the version from the local docker repository.

yes I wanted to reference the new jfrog artifactory everywhere but we're using this file as is in our tests :( .

PR #8205, opened 2021-07-21
feat(pipelined): Use future for poll stats retrieval

koolzz:

## Summary
The current library that we use is wonky and errors out, replaced with standard future implementation to remove the unknown xid log.

## Test Plan
Unit tests pass fine
## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/enforcement_stats.py
+++ b/lte/gateway/python/magma/pipelined/app/enforcement_stats.py
@@ -10,8 +10,10 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+import concurrent.futures
 import os
 from collections import defaultdict
+from concurrent.futures import Future
 from datetime import datetime, timedelta
 
 from lte.protos.pipelined_pb2 import RuleModResult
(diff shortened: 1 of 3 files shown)


lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -574,29 +583,31 @@ def _get_rule_id(self, flow):
...
-        parser = self._datapath.ofproto_parser
-        message = parser.OFPFlowStatsRequest(datapath=self._datapath, cookie = cookie, cookie_mask = cookie_mask)
         try:
-            response = ofctl_api.send_msg(self, message, reply_cls=parser.OFPFlowStatsReply,
-                    reply_multi=False)
-            if response == None:
+            xid = flows.send_stats_request(self._datapath, self.tbl_num, cookie,
+                                           cookie_mask)
+            self._poll_futures[xid] = Future()
+            res = self._poll_futures[xid].result(timeout=self._stats_wait_timeout)
```
pshelar, line 597: Can you delete the map entry once you are done with it to avoid holding ref to the object already processed.

koolzz: right, I had that in mind but forgot to do it. Thanks

koolzz: added

pshelar (approved):
LGTM

PR #8228, opened 2021-07-22
fix: Add docker retry in cwag-deploy

jwagantall:
Add a manual retry in case docker compose step
fails in cwag-deploy

## Additional Information

### DIFF
--- a/.github/workflows/cwag-workflow.yml
+++ b/.github/workflows/cwag-workflow.yml
@@ -150,12 +150,32 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Run docker compose
+        id: cwag-docker-compose
+        continue-on-error: true
         # yamllint disable rule:line-length
         env:
           DOCKER_REGISTRY: cwf_
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/cwag-workflow.yml
```diff
@@ -156,6 +156,13 @@ jobs:
         run: |
           cd ${MAGMA_ROOT}/cwf/gateway/docker
           docker-compose --file docker-compose.yml --file docker-compose.override.yml build --parallel
+      - name: Retry docker compose on failure
+        if: failure()
```
quentinDERORY: Use this instead: https://github.community/t/how-to-retry-a-failed-step-in-github-actions-workflow/125880/2
If you just use failure(), it will not be specific to the docker compose step

quentinDERORY (changes requested):
See comment

.github/workflows/cwag-workflow.yml
jwagantall: thank you!

quentinDERORY (approved):
LGTM

PR #8229, opened 2021-07-22
fix(mobilityd): Fixing mconfig reload of IP block allocation

### LINKED ISSUE #7650: [1.6]AGW mobilityd does not forgot old IP pool once updated from NMS

### Your Environment

- **Version:** 1.6
- **Affected Component:** AGW
- **Affected Subcomponent:** mobilityd
- **Deployment Environment:** AWS (orc8r), bare-metal (AGW), 

### Describe the Issue

Modifying the IP pool from the NMS does not overwrite what's in the gateway, it keeps on adding to the list. 

* Take a new NAT gateway (which has a default IP block of `192.168.128.0/24`) and modify the gateway config that changes the IP pool. Output of `mobility_cli.py list_ipv4_blocks` shows the old block and the new block. Updating it again will create a third entry ... 
image

**Expected behavior**
If IP block is modified on the NMS, old block should be removed from the gateway. 

**Screenshots**
See attached

ardzoht:

## Summary

- On mobilityd initialization process, IP block is allocated from mconfig ip block value, however, due to the reload and restart logic, the old IP block is not cleaned up, which can cause `OverlappedIPBlock` exception, and it doesn't update the new block.
- This PR updates the mconfig value load, by cleaning up already allocated block, and allocates the new one
- Applied some logging level updates and removed some duplicated logs
- Closes #7650 

## Test Plan

- make integ_test and make precommit
- tested with local orc8r to ensure update of new IP block is applied to mobilityd 

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/mobilityd/main.py
+++ b/lte/gateway/python/magma/mobilityd/main.py
@@ -160,19 +160,17 @@ def main():
     # Load IPAddressManager
     ip_address_man = IPAddressManager(ipv4_allocator, ipv6_allocator, store)
 
-    # Add all servicers to the server
-    mobility_service_servicer = MobilityServiceRpcServicer(
-        ip_address_man, config.get('print_grpc_payload', False),
-    )
-    mobility_service_servicer.add_to_server(service.rpc_server)
-
     # Load IPv4 and IPv6 blocks from the configurable mconfig file
     # No dynamic reloading support for now, assume restart after updates
     ipv4_block = _get_ip_block(mconfig.ip_block, "IPv4")
     if ipv4_block is not None:
         logging.info('Adding IPv4 block')
         try:
-            mobility_service_servicer.add_ip_block(ipv4_block)
+            allocated_ip_blocks = ip_address_man.list_added_ip_blocks()
+            if ipv4_block not in allocated_ip_blocks:
+                # Cleanup previously allocated IP blocks
+                ip_address_man.remove_ip_blocks(*allocated_ip_blocks, force=True)
+                ip_address_man.add_ip_block(ipv4_block)
         except OverlappedIPBlocksError:
             logging.warning("Overlapped IPv4 block: %s", ipv4_block)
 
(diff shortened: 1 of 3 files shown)


lte/gateway/python/magma/mobilityd/main.py
```diff
@@ -160,31 +160,32 @@ def main():
...
     # No dynamic reloading support for now, assume restart after updates
     ipv4_block = _get_ip_block(mconfig.ip_block, "IPv4")
     if ipv4_block is not None:
         logging.info('Adding IPv4 block')
         try:
-            mobility_service_servicer.add_ip_block(ipv4_block)
+            # Cleanup previously allocated IP blocks
+            allocated_ip_blocks = ip_address_man.list_added_ip_blocks()
+            ip_address_man.remove_ip_blocks(*allocated_ip_blocks, force=True)
+            ip_address_man.add_ip_block(ipv4_block)
```
pshelar: After this change, how do we maintain UE session IP address after service restart? this is expected in stateless mode.

ardzoht: Added pre-check to only cleanup for different IP blocks

lte/gateway/python/magma/mobilityd/main.py
```diff
@@ -160,31 +160,33 @@ def main():
...
+                ip_address_man.add_ip_block(ipv4_block)
         except OverlappedIPBlocksError:
             logging.warning("Overlapped IPv4 block: %s", ipv4_block)
 
     ipv6_block = _get_ip_block(mconfig.ipv6_block, "IPv6")
     if ipv6_block is not None:
         logging.info('Adding IPv6 block')
         try:
-            mobility_service_servicer.add_ip_block(ipv6_block)
+            ip_address_man.add_ip_block(ipv6_block)
```
pshelar: can we do same for IPv6 blocks?

pshelar (approved):
LGTM

PR #8236, opened 2021-07-22
feat(sessiond): example of simple shard id (THIS IS A JUST A TEMPLATE, not to be merged)

uri200:

## Summary

This is just an example of adding and removing shards ids from UEs. 

The main idea is shards_ will be just a vector of integers. In other words, just a counter of how many UEs we have in each shard.

Each session will store the value of its shard

Every session for one specific UE will have the same shard (that can be easily change removing some conditions)

When UE is added we will either get the shard from a previous session for that UE or get a new one and ++ the counter for that shard

When UE leaves, we just check if the session is the last one for that UE. If so, we -- on that specific shard

Note code IT IS NOT PROPER 
- Location of shards_ is wrong
- shards_ is not initialized 
- access to shards_ is not good
- functions are not defined on headers

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -39,7 +39,11 @@ bool LocalEnforcer::SEND_IPFIX             = true;
 
 using google::protobuf::RepeatedPtrField;
 
+
+
 using namespace std::placeholders;
+
+
 
 static void handle_command_level_result_code_for_monitors(
     const std::string& imsi, const std::string& session_id,
@@ -1051,6 +1055,8 @@ bool LocalEnforcer::update_tunnel_ids(
     session->receive_monitor(monitor, nullptr);
   }
 
+  add_ue_to_shard(imsi, session_map, session);
+
   handle_session_activate_rule_updates(
       *session, csr, charging_credits_received);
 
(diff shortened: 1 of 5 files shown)


lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -20,18 +20,18 @@ namespace magma {
 
 int ShardTracker::add_ue() {
   /*
-  * If there are no shards, add an entry for the shard
-  * representing a single UE at the new shard
-  */
+   * If there are no shards, add an entry for the shard
```
uri200, line 23: i didn't change anything after this point. IDE fixed this for me

lte/gateway/c/session_manager/LocalEnforcer.h
```diff
@@ -39,6 +39,8 @@
 namespace magma {
 using std::experimental::optional;
 
+using shards_ = std::vector<int>;
```
uri200, line 42: this should be proper initialized

Maybe you will also need functions to grow and shrink the vector.

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1211,6 +1257,8 @@ void LocalEnforcer::complete_termination(
   // clear all metrics associated with this session
   session->clear_session_metrics();
 
+  remove_ue_to_shard(imsi, session_map, session);
+
```
veshkemburu, line 1261: What's the reason to remove shard ID here?

PR #8239, opened 2021-07-23
chore(ci): Store SwaggerHub account info to sops

hcgatewood:
## Summary

PGP key and passphrase is stored in lastpass. After injecting PGP key locally, can use sops as follows

```bash
sops secrets/ci.yaml  # edit

sops -d secrets/ci.yaml  # view
```

Stored in encrypted yaml, but unencrypted looks like this

```yaml
magma-bot:
  about: Bot account for automating Magma actions
  email: bots@magmacore.org
  accounts:
    gmail.com:
      password: XXX
    github.com:
      username: magma-bot
      password: XXX
      tokens:
        magma-docusaurus-bot: XXX
    swaggerhub.com:
      password: XXX
      api_key: XXX
```

## Additional Information

### DIFF
--- a/secrets/ci.yaml
+++ b/secrets/ci.yaml
@@ -1,19 +1,24 @@
 magma-bot:
-    gmail:
-        username: ENC[AES256_GCM,data:vVsbjRVfR422Y0ZrSsjcVMGf,iv:GwSUUAR1bcS7MQxJmrAjdZ+ZRO3EpCdBC3hi1950cSw=,tag:4cflzSkz3A2Ylw8PILSm2w==,type:str]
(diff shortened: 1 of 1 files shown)


quentinDERORY:
I did not know sops, it's cooler than ansible vault!
Few questions:
This file is just for reference and will not be used by any CI?
If no and this is just for reference, do we want to store all ci-bots credentials in there?

quentinDERORY (approved):

hcgatewood:
> I did not know sops, it's cooler than ansible vault!
> Few questions:
> This file is just for reference and will not be used by any CI?
> If no and this is just for reference, do we want to store all ci-bots credentials in there?

My thought was eventually we could store all CI secrets directly in sops -- then our CI env would just need the 1 pgp key, and each CI job would have a "hydrate secrets" step that adds the necessary secrets from sops into the env

But yeah for now I'm just using it to store account information in a unified location -- it's not actually being used by CI directly

PR #8242, opened 2021-07-23
fix(mme): sentry_config was not set in OAI dockerfiles

rdefosse:

## Summary

I was not using `jinja` template for CI-PR runs `dockerfiles`
But it is used in the full-from-scratch `dockerfiles` and this morning `master` branch checks failed because of this.

See [failed run today](https://jenkins-oai.eurecom.fr/job/MAGMA-MME-production/5117/).

What I've added:

- a check in template for existence of `sentry_config`
- jinja calls in CI-PR runs `dockerfiles` to avoid the same kind of error.

## Test Plan

I've tested manually with and without `sentry_config` the `jinja` command.
Normally the `OAI` CI pipeline will finalize the fix.

## Additional Information

### DIFF
--- a/ci-scripts/docker/Dockerfile.mme.ci.rhel8
+++ b/ci-scripts/docker/Dockerfile.mme.ci.rhel8
@@ -21,6 +21,35 @@ RUN export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/local/lib/pkgconfig/ && \
     make build_oai && \
     make build_sctpd
 
+# Prepare config file
+RUN yum install -y python3-pip && \
+    pip3 install jinja2-cli && \
+    cd $MAGMA_ROOT/lte/gateway/docker/mme/configs/ && \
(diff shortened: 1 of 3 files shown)


themarwhal (approved):
Not too familiar with the change so I'll let @waqaraqeel take a look.
Is it OK that the oai test is failing on this PR as well?

rdefosse:
> 
> 
> Not too familiar with the change so I'll let @waqaraqeel take a look.
> Is it OK that the oai test is failing on this PR as well?

Not it is not OK. I am working on it.

I've got other CI problems.

I will push again

waqaraqeel (approved):

waqaraqeel:
Changes lgtm.

Question: In the CI job, do we not want to use the `generate_oai_config.py` script and provide it with a JSON service config file?

Thanks for the fix!

quentinDERORY (approved):
LGTM

PR #8246, opened 2021-07-23
feat(sessiond): Change Poll Stats API to Poll for Shards

veshkemburu:

## Summary

From sessiond's side, we need to modify the poll stats API to poll for shards. We just need to invoke the API to be able to invoke a vector of shard ids. Also handled activate flows request processing from sessiond's side, to pass shard id into it for pipelined to process, and have reverted changes in Nick's PR from: [https://github.com/magma/magma/pull/8205](https://github.com/magma/magma/pull/8205), due to issues with futures.

## Test Plan

Ran sessiond unit tests. Also changed polling pipelined test to poll using a cookie mask that matches all shards just for a particular cookie, to illustrate the needed behavior. 

## Additional Information


lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -262,14 +263,22 @@ void LocalEnforcer::handle_pipelined_response(
...
-void LocalEnforcer::poll_stats_enforcer(int cookie, int cookie_mask) {
+void LocalEnforcer::poll_stats_enforcer(std::vector<int> shard_ids) {
   // we need to pass in a function pointer. Binding is required because
   // the function is part of the LocalEnforcer class and has arguments
   // so we bind to the object and the two arguments the function needs
   // which are the status and RuleRecordTable response
-  pipelined_client_->poll_stats(
-      cookie, cookie_mask,
-      std::bind(&LocalEnforcer::handle_pipelined_response, this, _1, _2));
+  if (shard_ids.empty()) {
```
koolzz: I think we chatted about this already, lets move this logic into StatsPoller, and just call poll_stats_enforcer with different cookie/cookie_mask args

veshkemburu: Right, I'll move it back another level.

lte/gateway/c/session_manager/StatsPoller.cpp
```diff
@@ -13,17 +13,27 @@
...
-#define COOKIE 0
-#define COOKIE_MASK 0
+#define OVS_COOKIE_MATCH_ALL 0xffffffff
 
 namespace magma {
 
 void StatsPoller::start_loop(
     std::shared_ptr<magma::LocalEnforcer> local_enforcer,
     uint32_t loop_interval_seconds) {
+  // for now poll all shards by passing in an empty vector
```
koolzz: This comment is a bit confusing maybe

// When no shard_ids are available poll all UEs

lte/gateway/c/session_manager/StatsPoller.cpp
```diff
@@ -13,17 +13,27 @@
...
+  // to poll by cookie pass in a vector of shard ids, where
+  // each element is a shard id that maps directly to a cookie
   while (true) {
-    local_enforcer->poll_stats_enforcer(COOKIE, COOKIE_MASK);
+    std::vector<int> shard_ids{};
```
koolzz: can you add the logic for using your shardtracker thing, you already have everything so lets just do it in this pr

lte/gateway/c/session_manager/sessiond_main.cpp
```diff
@@ -358,7 +364,7 @@ int main(int argc, char* argv[]) {
       if (config["poll_stats_interval"].IsDefined()) {
         interval = config["poll_stats_interval"].as<uint32_t>();
       }
-      periodic_stats_requester->start_loop(local_enforcer, interval);
+      periodic_stats_requester->start_loop(local_enforcer, interval, shard_ids);
```
koolzz: if you do it this way then shard_ids will always be empty. You have to iterate over the shard tracker class you have

koolzz (changes requested):

lte/gateway/c/session_manager/StatsPoller.cpp
```diff
@@ -13,17 +13,32 @@
...
 
 namespace magma {
 
 void StatsPoller::start_loop(
     std::shared_ptr<magma::LocalEnforcer> local_enforcer,
     uint32_t loop_interval_seconds) {
+  // when no shard ids are available poll all UEs
+  // to poll by cookie pass in a vector of shard ids, where
+  // each element is a shard id that maps directly to a cookie
+  auto shard_tracker = local_enforcer->get_shard_tracker();
```
koolzz: Hmm I think it might be better to just pass shard_tracker in constructor to StatsPoller? adding a get method for it in LocalEnforcer seems convoluted

lte/gateway/c/session_manager/StatsPoller.cpp
```diff
@@ -13,17 +13,32 @@
...
   while (true) {
-    local_enforcer->poll_stats_enforcer(COOKIE, COOKIE_MASK);
+    // TODO(veshkemburu): add smart polling function to determine which
+    // shard ids to poll
+    if (shard_tracker.get_size_of_tracker() == 0) {
+      local_enforcer->poll_stats_enforcer(0, 0);
+    } else {
+      // iterate through all shard ids, but only poll shard IDs that have UEs
+      for (size_t i = 0; i < shard_tracker.get_size_of_tracker(); i++) {
+        if (!shard_tracker.get_imsis_per_shard().at(i).empty()) {
```
koolzz: This feels, complicated. Lets just add a function to shard_tracker that returns a vector of active shards?

veshkemburu: I was thinking about doing that, and frankly I like that idea much better.

lte/gateway/c/session_manager/StatsPoller.cpp
```diff
@@ -13,17 +13,32 @@
...
+  // to poll by cookie pass in a vector of shard ids, where
+  // each element is a shard id that maps directly to a cookie
   while (true) {
-    local_enforcer->poll_stats_enforcer(COOKIE, COOKIE_MASK);
+    // TODO(veshkemburu): add smart polling function to determine which
+    // shard ids to poll
+    if (shard_tracker->get_size_of_tracker() == 0) {
+      local_enforcer->poll_stats_enforcer(0, 0);
+    } else {
+      // retrieve a list of all active shard ids and poll them
```
koolzz: I don't think this comment brings any value, rm?

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -53,4 +53,18 @@ bool ShardTracker::remove_ue(const std::string imsi, const uint16_t shard_id) {
   return true;
 }
 
+uint16_t ShardTracker::get_size_of_tracker() {
```
koolzz: get_size_of_tracker maybe -> get_shard_list_size ?

lte/gateway/c/session_manager/StatsPoller.cpp
```diff
@@ -13,17 +13,31 @@
...
+    uint32_t loop_interval_seconds,
+    std::shared_ptr<magma::ShardTracker> shard_tracker) {
+  // when no shard ids are available poll all UEs
+  // to poll by cookie pass in a vector of shard ids, where
+  // each element is a shard id that maps directly to a cookie
   while (true) {
-    local_enforcer->poll_stats_enforcer(COOKIE, COOKIE_MASK);
+    // TODO(veshkemburu): add smart polling function to determine which
+    // shard ids to poll
+    if (shard_tracker->get_shard_list_size() == 0) {
```
themarwhal: Is this ever going to be true? I saw this in the constructor
```

ShardTracker::ShardTracker() {
  // initialize with at least one empty vector to avoid checking for empty
  imsis_per_shard_.push_back({});
}
```

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -37,6 +37,18 @@ class ShardTracker {
...
+   * Retrieve number of shards currently in shard tracker
+   * @return int representing number of shards that have been generated
+   */
+  uint16_t get_shard_list_size();
+
+  /**
+   * Retrieve a list of active shard ids
+   * @return vector<int> representing shard IDs of shards that have UEs
+   */
+  std::vector<int> get_active_shards();
```
themarwhal: `std::vector< uint16_t >`?

lte/gateway/c/session_manager/StatsPoller.cpp
```diff
@@ -13,17 +13,31 @@
...
-#define COOKIE_MASK 0
+#define OVS_COOKIE_MATCH_ALL 0xffffffff
 
 namespace magma {
 
 void StatsPoller::start_loop(
     std::shared_ptr<magma::LocalEnforcer> local_enforcer,
-    uint32_t loop_interval_seconds) {
+    uint32_t loop_interval_seconds,
+    std::shared_ptr<magma::ShardTracker> shard_tracker) {
```
themarwhal, line 24: nit: I think the parameter order should be local_enforcer, shard_tracker, loop_interval_seconds

lte/gateway/c/session_manager/test/test_polling_pipelined.cpp
```diff
@@ -18,6 +18,7 @@
 #include "ProtobufCreators.h"
 #include "SessiondMocks.h"
 #include "SessionStore.h"
+#define OVS_COOKIE_MATCH_ALL 0xffffffff
```
themarwhal, line 21: maybe define this in Consts.h

themarwhal (changes requested):
a few comments :) but looks good overall

lte/gateway/c/session_manager/ShardTracker.h
veshkemburu: The issue is that cookies can be up to size uint32 on pipelined side, and these values will get mapped directly to cookies and type mismatch errors happen.

themarwhal: oh interesting. Then why did we pick uint16_t on SessionD's side?

veshkemburu: I picked uint16 because I didn't think we wanted a lot of shards in concept. We want for shards to be relatively large in size capabale of holding a lot of UEs, and they'll help well in multiprocessing and smart polling.

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -845,12 +848,13 @@ void LocalEnforcer::schedule_dynamic_rule_activation(
...
 
-        auto config          = session->get_config();
-        const auto ip_addr   = config.common_context.ue_ipv4();
-        const auto ipv6_addr = config.common_context.ue_ipv6();
-        const Teids teids    = config.common_context.teids();
-        const auto ambr      = config.get_apn_ambr();
-        const auto msisdn    = config.common_context.msisdn();
+        auto config             = session->get_config();
+        const auto ip_addr      = config.common_context.ue_ipv4();
+        const auto ipv6_addr    = config.common_context.ue_ipv6();
```
veshkemburu: Just style changes(make format_session_manager)

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1046,18 +1051,8 @@ void LocalEnforcer::init_session(
 
 void LocalEnforcer::add_ue_to_shard(
     const std::string imsi, SessionMap& session_map, SessionState& session) {
-  auto sm_it     = session_map.find(imsi);
-  auto& sessions = sm_it->second;
-  // Check if UE has other sessions and reuse shard_id if so
-  for (auto session_it = sessions.begin(); session_it != sessions.end();
-       ++session_it) {
```
veshkemburu, line 1053: This approach caused bugs in the polling, so I reverted to just checking if the UE already exists within the add method itself.

lte/gateway/c/session_manager/ShardTracker.cpp
```diff
@@ -53,4 +56,20 @@ bool ShardTracker::remove_ue(const std::string imsi, const uint16_t shard_id) {
   return true;
 }
 
+uint16_t ShardTracker::get_shard_list_size() {
+  return imsis_per_shard_.size();
+}
+
+std::vector<int> ShardTracker::get_active_shards() {
```
veshkemburu: int data type because cookie values passed in must be ints(but the number of shards itself will probably be small enough to be uint16).

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -47,7 +59,7 @@ class ShardTracker {
   /*
    * largest number of UEs that can fill a shard
    */
-  const uint16_t max_shard_size_ = 100;
+  const uint16_t max_shard_size_ = 300;
```
veshkemburu, line 63: Default value I have set from running the tests(on UEs at multiples of 100, 600-800 UEs). Three shards is a good number to have(though ten has performed slightly better).

lte/gateway/c/session_manager/LocalEnforcer.h
```diff
@@ -458,7 +458,8 @@ class LocalEnforcer {
    */
   void propagate_rule_updates_to_pipelined(
       const SessionConfig& config, const RulesToProcess& pending_activation,
-      const RulesToProcess& pending_deactivation, bool always_send_activate);
+      const RulesToProcess& pending_deactivation, bool always_send_activate,
+      uint16_t shard_id);
```
themarwhal: nit: the order looks off, the parameters that identify the session/UE should come first and then information about what/how to propagate
```
void propagate_rule_updates_to_pipelined(
        const SessionConfig& config, uint16_t shard_id, 
        const RulesToProcess& pending_activation,
        const RulesToProcess& pending_deactivation, bool always_send_activate);
```

lte/gateway/c/session_manager/PipelinedClient.h
```diff
@@ -128,7 +128,7 @@ class PipelinedClient {
       const std::string& imsi, const std::string& ip_addr,
       const std::string& ipv6_addr, const Teids default_teids,
       const std::string& msisdn, const optional<AggregatedMaximumBitrate>& ambr,
-      const RulesToProcess to_process,
+      const RulesToProcess to_process, uint16_t shard_id,
```
themarwhal: Same note as above here. ShardID should come before `const RulesToProcess to_process`.

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: What kind of bugs?

lte/gateway/c/session_manager/ShardTracker.h
themarwhal: Let's put this comment inside the code so that we don't lose context :)

lte/gateway/c/session_manager/StatsPoller.cpp
```diff
@@ -13,18 +13,42 @@
...
+      local_enforcer->poll_stats_enforcer(0, 0);
+      std::this_thread::sleep_for(std::chrono::seconds(interval));
+    } else {
+      for (size_t shard_index = 0; shard_index < active_shard_ids.size();
+           shard_index++) {
+        local_enforcer->poll_stats_enforcer(
+            active_shard_ids[shard_index], OVS_COOKIE_MATCH_ALL);
+        std::this_thread::sleep_for(std::chrono::seconds(interval));
+      }
+    }
```
themarwhal: Could we clean up this logic to look like 
```
if (active_shard_ids.size() == 0) {
// poll everything
// sleep for 10 seconds
// continue
}

// Do the else clause
```
?

themarwhal: Also, if there are no active shards `active_shard_ids.size() == 0`, why do we need to poll at all?

themarwhal: Also, let's use `active_shard_ids.empty()`

lte/gateway/c/session_manager/ShardTracker.h
```diff
@@ -28,14 +28,26 @@ class ShardTracker {
    * an addition
    * @return index(shard id) where UE was placed
    */
-  uint16_t add_ue(const std::string imsi);
+  unsigned int add_ue(const std::string imsi);
```
themarwhal, line 31: 👍

themarwhal (changes requested):

lte/gateway/c/session_manager/StatsPoller.cpp
veshkemburu: So this is a change to see if it optimizes polling. Polling everything and sleep for 10 seconds was how I did polling before. The new idea was to reduce load by polling a shard every few seconds rather than polling at once. I guess technically we don't need to poll when there are no active shard ids(this was more just in case)

lte/gateway/c/session_manager/LocalEnforcer.cpp
veshkemburu: It wasn't properly finding the session ids or setting them. Just seemed simpler to handle the case inside, without much performance compromise.

ekowtaylor:
@veshkemburu , @koolzz  this PR has now gone stale. To maintain a healthy codebase, it will automatically closed unless there are plans to rebase and merge by close of week

themarwhal:
Closing it out to clear out notification.

Issue #8028, opened 2021-07-14
Export control path network metrics


waqaraqeel:
# Export Control Path Network Metrics

## Problem

Services running on the access gateway communicate with the orc8r and generate
control traffic. We do not have visibility into the amount, patterns, and
service-level breakdown of this traffic. As we try to minimize control plane
bandwidth consumption for our deployment partners, especially those with
satellite backhaul, visibility into control path bandwidth consumption is crucial.

## Solution

- Use eBPF on the access gateway to monitor network traffic going out.
- Maintain per-service, per-destination (`ip:port`) byte and packet counters in the eBPF program.
- Create a new Python service to load the eBPF program into the kernel, read its counters and expose those counters as Prometheus metrics.
- Let `magmad` collect those Prometheus metrics and upload them to the orc8r so they are available in the NMS.

## Non-goals

- Change collection/export methods for existing data-path metrics.

**[Design doc](https://magma.github.io/magma/docs/next/proposals/p018_control_network_metrics)**

amarpad:
Nice!

electronjoe:
@waqaraqeel do you mind posting a PR for the design doc to Magma's repo, tagging this GH Issue in the description? I'll stamp it for merge (discussion will occur here) but that versions your design doc and allows conversation here on well defined document versions.

waqaraqeel:
> @waqaraqeel do you mind posting a PR for the design doc to Magma's repo, tagging this GH Issue in the description? I'll stamp it for merge (discussion will occur here) but that versions your design doc and allows conversation here on well defined document versions.

Added in #8040.

electronjoe:
Thanks Waqar!  Dropping feedback on the proposal revision here.

## Proposal

> Every magma service running on the orchestrator will get a separate Prometheus Gauge by the name `{service}_bytes_sent` and `{service}_packets_sent`

For this style of metric, the advice is to use a [Counter](https://prometheus.io/docs/concepts/metric_types/#counter) type. Any metric that purely increments (omitting restart) should prefer `Counter` type.  What you'll find is that counters decimate much more pleasantly and with fewer surprises (and decimation is a common need for both querying, and time series database compression in the background). By decimation I mean e.g. taking your reporting interval of every 5 seconds and reducing it to every 20 seconds.  Accumulator style metrics simply discard. Gauges require a policy for decimation and it can get dicey.  And in the event of a telemetry outage (e.g. network disconnection or some such) you are able to capture behavior during the period of outage.

[...]

Issue #8173, opened 2021-07-20
[AGW] Add telemetry monitoring for open FD count per process


electronjoe:
The first question that comes to mind for the long running SEV discussed in https://github.com/magma/magma/issues/4833 is whether our Magma binaries are leaking resources - in this case in the form of open sockets created by gRPC.

We should make detection of all forms of resource leakage easy to detect via telemetry, so if missing (it seems to be so far) we should add this telemetry stream.

We might leverage the eBPF work of @waqaraqeel - once he completes landing the packet and byte counter telemetry streams it should not be complex to add eBPF counters to monitor open FD count per Magma process.

electronjoe:
@waqaraqeel's network telemetry design proposal is [p0008](https://github.com/magma/magma/blob/master/docs/readmes/proposals/p018_control_network_metrics.md)

electronjoe:
I'll note that @waqaraqeel pointed out our Python Prometheus Client automatically exports FD resource count.  The C++ client does not appear to, however.

PR #8251, opened 2021-07-23
feat: add kernsnoopd service implementation

waqaraqeel:

## Summary

Working towards #8028.

This PR provides the implementation for the _kernsnoopd_ service discussed in #8028 and in the design document referenced there. There are some deviations from the design doc that were made deliberately during implementation. The design doc will be updated. Please see comments on code.

Before merging, it must be ensured that `ebpf_programs` list in `kernsnoopd.yml` is empty. If not, _kernsnoopd_ will be enabled by default. 

## Test Plan

Added unit tests for the front-end program in `handler.py`. Ran _kernsnoopd_ in gateway VM and observed log messages to ensure correct operation.

Measured the performance impact of instrumenting the `tcp_sendmsg` syscall with the Spirent lab setup. 4 tests were run in total, each with 600 UEs, 5 MB DL per UE, and an attach rate of 10. The tests ran for 30 minutes each.

For 2 tests, the `tcp_sendmsg` instrumentation was enabled. These tests generated the following throughput plots:

For the other 2 tests, the `tcp_sendmsg` instrumentation was disabled. These tests generated the following throughput plots:

Hence no distinguishable downgrade in throughput was observed when `tcp_sendmsg` instrumentation was enabled. The mean DL throughput for each of the four cases hovered around 530 Mbps.


lte/gateway/configs/kernsnoopd.yml
```diff
@@ -0,0 +1,21 @@
...
+
+log_level: INFO
+
+# collect_interval is the interval in seconds at which kernsnoopd collects
+# counters from the kernel
+collect_interval: 10
+
+# ebpf_programs are the ebpf source filenames that kernsnoopd will load.
+ebpf_programs:
+  - packet_count
```
waqaraqeel: List of eBPF programs to load. `packet_count` is the current one. I will also provide one which listens to TCP state change events so performance impact can be lowered. #8173 suggests another program to keep count of open file descriptors.

lte/gateway/python/magma/kernsnoopd/handlers.py
```diff
@@ -0,0 +1,170 @@
...
+            d_port = ntohs(key.dport)
+            service_name = None
+
+            try:
+                # TODO: destination service name inference does not work when
+                # control_proxy is enabled
+                service_name = self._get_source_service(key)
+                # get destination service from host and port
+                dest_service = self._registry.get_service_name(d_host,
+                                                               d_port)
```
waqaraqeel: If two Magma services are communicating, we want to increment MAGMA counters with source and destination service names as labels. This functionality doesn't work yet when traffic is proxied through `control_proxy`.

lte/gateway/python/magma/kernsnoopd/snooper.py
```diff
@@ -0,0 +1,107 @@
...
+import os
+
+from bcc import BPF
+from jinja2 import Template
+from magma.common.job import Job
+
+from magma.kernsnoopd.handlers import ebpf_handlers
+
+# TODO: what's the right way to look for these files?
+EBPF_SRC_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'ebpf')
```
waqaraqeel: Will non-Python files be here when Magma is packaged? Beside the Python source file?

lte/gateway/python/magma/kernsnoopd/snooper.py
```diff
@@ -0,0 +1,107 @@
...
+            filename = os.path.join(EBPF_SRC_DIR, f'{basename}.bpf.c')
+            try:
+                sources.append(self._get_ebpf_source(filename, self._context))
+                handler = ebpf_handlers[basename](self._service_registry)
+                self._handlers.append(handler)
+            except FileNotFoundError:
+                logging.error('Could not open eBPF source file %s' % filename)
+            except KeyError:
+                # TODO: How to terminate so magmad doesn't attempt restart
+                logging.error('Fatal: did not find handler for %s' % basename)
```
waqaraqeel, line 104: There is nothing to load for the service. No need to keep running. How to stop without Magmad attempting restart?

lte/gateway/release/build-magma.sh
```diff
@@ -137,6 +137,7 @@ MAGMA_DEPS=(
     "nlohmann-json3-dev"
     "sentry-native"   # sessiond
     "td-agent-bit >= 1.7.8"
+    "bpfcc-tools" # required for kernsnoopd
```
waqaraqeel, line 140: Ships an integrated clang. 63.0 MB in total. Where else do I put this dep?

lte/gateway/python/magma/kernsnoopd/snooper.py
```diff
@@ -0,0 +1,107 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import logging
+import os
+
+from bcc import BPF
```
waqaraqeel, line 16: Should this be guarded? Log and terminate service if bcc not installed.

orc8r/gateway/python/magma/common/service_registry.py
```diff
@@ -78,6 +78,23 @@ def list_services():
         """
         return ServiceRegistry.get_registry()["services"]
 
+    @staticmethod
```
themarwhal: Looks like the linter wants @classmethod instead.. https://wemake-python-stylegui.de/en/latest/pages/usage/violations/oop.html#wemake_python_styleguide.violations.oop.StaticMethodViolation

I don't use Python often enough to have an opinion on this, will research best practices :)

lte/gateway/release/build-magma.sh
themarwhal: @tmdzk / @quentinDERORY on adding new dependencies

lte/gateway/python/magma/kernsnoopd/tests/packet_counter_tests.py
```diff
@@ -0,0 +1,125 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import unittest
+from socket import htons
+from unittest.mock import MagicMock
+
+from magma.kernsnoopd.handlers import PacketCounter
+
+
```
themarwhal: Just commenting to say thank you for adding such well written unit tests

orc8r/gateway/python/magma/common/service_registry.py
waqaraqeel: The ServiceRegistry class is full of @staticmethod... The linter is complaining about a lot of other things. Still working on pleasing it.

themarwhal: Yeah dw about existing lint warnings. Would be good to not introduce new warnings OR identify any linter warnings that should be disabled.

lte/gateway/python/magma/kernsnoopd/snooper.py
themarwhal: What is the dependency to have BPF here? Is it `bpfcc-tools` mentioned in build-magma.sh?

lte/gateway/python/magma/kernsnoopd/snooper.py
themarwhal: cc @tmdzk / @quentinDERORY 

I suspect you'll have to handle them in build-magma.sh / post install but will let the CI folks give their expertise

lte/gateway/release/build-magma.sh
quentinDERORY: Here https://github.com/magma/magma/blob/87603aff84a8dda0e75d362455f75b369972f849/lte/gateway/deploy/roles/magma/tasks/main.yml#L240

lte/gateway/python/magma/kernsnoopd/snooper.py
waqaraqeel: Yes. _Snooper_ needs bcc to compile and load eBPF programs to kernel.

lte/gateway/python/magma/kernsnoopd/snooper.py
themarwhal: I see what you mean. I don't know of a way off the top of my head. Does stopping the loop work?

waqaraqeel: I found a way to prevent the job from starting rather than stopping them afterwards.

lte/gateway/release/build-magma.sh
waqaraqeel: Done.

orc8r/gateway/python/magma/common/service_registry.py
waqaraqeel: Fixed all reported issues. I don't know why Python Format check is still failing. It's not telling me either...

lte/gateway/python/magma/kernsnoopd/snooper.py
waqaraqeel: It looks like I should use `glob_files()` in `build-magma.sh` to place my static files in `/var/opt/magma/ebpf/`, and directly read from absolute path in the service. Does that sound right? @tmdzk @quentinDERORY

lte/gateway/python/magma/kernsnoopd/ebpf/packet_count.bpf.c
```diff
@@ -0,0 +1,53 @@
...
+ * counts the number of bytes and packets sent by each linux binary.
+ * Expect considerable performance impact.
+ * https://patchwork.ozlabs.org/patch/309437/
+ */
+
+// Create a hash map with `key_t` as key type and `counters_t` as value type
+BPF_HASH(dest_counters, struct key_t, struct counters_t, 1000);
+
+// Attach hook for the `net_dev_start_xmit` kernel trace event
+TRACEPOINT_PROBE(net, net_dev_start_xmit) {
```
pshelar: since this is hooking up to all L2 traffic,
can you post perf numbers from the lab spirent setup, for scale test like, 600 UE @ 5 mbps DL traffic for each UE.

lte/gateway/python/magma/kernsnoopd/ebpf/packet_count.bpf.c
```diff
@@ -0,0 +1,53 @@
...
+  struct key_t key = {};
+  bpf_probe_read(&key.daddr, sizeof(sk->sk_daddr), &sk->sk_daddr);
+  // ignore packets destined for localhost
+  // IPv4 127.0.0.1 == 16777343, 0.0.0.0 == 0
+  if ((key.daddr == 16777343 || key.daddr == 0)) {
+    return 0;
+  }
+
+  // read binary name, pid, destination IP address, port
+  bpf_get_current_comm(key.comm, TASK_COMM_LEN);
```
pshelar: This might not return the app name that triggered the traffic since the packet could be processed asynchronously in the kernel.

pshelar (changes requested):

lte/gateway/python/magma/kernsnoopd/ebpf/packet_count.bpf.c
waqaraqeel: True. I am seeing traffic from the swapper process with pid 0 which I think is from asynchronous processing. How do I find the app or pid that actually triggered traffic? It looks like there is no sure fire way and we have to make do with best effort [0].

[0] https://www.brendangregg.com/blog/2016-11-30/linux-bcc-tcplife.html. See "4. Showing task context."

orc8r/gateway/python/magma/common/service_registry.py
themarwhal: It looks there are some leftover formatting changes for `lte/gateway/python/magma/kernsnoopd/handlers.py`
```
Running formatting commands for lte/gateway/python/magma/kernsnoopd/__init__.py...
Running formatting commands for lte/gateway/python/magma/kernsnoopd/handlers.py...
Rewriting lte/gateway/python/magma/kernsnoopd/handlers.py
```

What are you running locally?

lte/gateway/python/magma/kernsnoopd/ebpf/packet_count.bpf.c
waqaraqeel: The objective with this PR is to get the _kernsnoopd_ service code, and eBPF compiling and loading infrastructure checked in.

We expect high performance impact which is why this program is disabled by default. It can be enabled by adding `packet_count` to the `ebpf_programs` list in [kernsnoopd.yml](https://github.com/magma/magma/pull/8251/files#diff-588537ec9ea515e6ab0698e3a434e1a62a82c114896b7736e55ae0abd81f2255R20). This program will likely stay disabled in most deployments. A future PR will have a different program that listens to TCP state change events rather than individual packets for a much lower performance impact. Integration tests are also planned.

orc8r/gateway/python/magma/common/service_registry.py
waqaraqeel: I run `autopep8` directly and it's not making any changes to the existing file. I couldn't use the precommit.py script as it complained about not finding `magma/py-lint:latest` and not being able to download the image because of docker login issues.

themarwhal: Did you follow the instructions here? https://docs.magmacore.org/docs/next/lte/dev_unit_testing#format-agw

You have to use the build flag first to build the image

lte/gateway/python/magma/kernsnoopd/ebpf/packet_count.bpf.c
pshelar: it still would be nice to quantify the perf impact of this feature.

lte/gateway/python/magma/kernsnoopd/ebpf/packet_count.bpf.c
pshelar: you could export the 5-tuple to the map. In userspace we could map it to pid using `ss` type of utility.
We might miss short lived sockets/process in this way, but I dont think it will matter in most of cases.

orc8r/gateway/python/magma/common/service_registry.py
waqaraqeel: I hadn't followed this one. I was looking at the [contributing conventions](https://magma.github.io/magma/docs/next/contributing/contribute_conventions#python). Have made the required changes now. Thanks!

themarwhal: i should update that file then :)

themarwhal (approved):
Approving to clean out my review queue - deferring the last approval to @pshelar

lte/gateway/python/magma/kernsnoopd/ebpf/packet_count.bpf.c
pshelar: I am curious about the choice of xmit (L2) trace point.
why not use L4 tracepoint similar to https://github.com/iovisor/bcc/blob/master/tools/tcptop.py?
This would have much lower overhead.

waqaraqeel: We wanted to include L3 traffic from ping, traceroute etc. as well.

pshelar: do we have any guess on how much bandwidth is spent on pure L3 traffic?
is it work the overhead of tracking L4 level stats at L2 ?

waqaraqeel: I don't have any estimates/guesses but I don't think it would be much. _Magmad_ sends some ping and traceroute traffic for monitoring periodically, but bandwidth-wise it should be miniscule. If we switch to `tcp_send`, we will also lose the packet counts and would be left with byte counts only.

pshelar: you are not getting real packet count even in this function. Most of devices has TCP optimizations that offload TCP segments to the NIC. so you are really counting TCP segments. 
Looking at the tcp_info. All required stats are already there. so the eBPF code is pretty much redundant stats collector.
We could just make use of `ss -nti`

ref: https://www.brendangregg.com/blog/2016-10-15/linux-bcc-tcptop.html

waqaraqeel: The primary motivation for using eBPF over a ready-made approach is flexibility. Eventually, we want more than just byte counters keyed by 5-tuples.

For example, we want separate counters for traffic between pairs of AGW and orc8r services. At the network level, all this traffic may be going over a single socket (when _control_proxy_ is enabled) or over multiple sockets with the same destination host and port (when _control_proxy_ is disabled). eBPF will allow us to look into the authority HTTP header and figure out the destination service. We may also want to look at the HTTP/2 stream numbers in the future.

Another example could be open fd counters as discussed in #8173. While we currently have the `process_open_fds` metric that is collected through the `psutil` python package, that method will not work if we wanted to find out more about the kind of fds opened, how long they stayed opened etc. eBPF will allow us to instrument the `open()` and `close()` syscalls and keep track of that.

The plan is to make _kernsnoopd_ a home for all such metric collection. `packet_count` is the first such collector. The upside of this program is that it does not miss short-lived processes (assuming their traffic is not processed asynchronously,) and accounts for non-TCP traffic as well. The downside is the expected high performance impact which I'm working on measuring. I [will also write a separate program](https://github.com/magma/magma/pull/8251#discussion_r675767217) that listens to TCP events and correctly implements the AGW service to orc8r service mapping.

cc: @electronjoe

pshelar: > The primary motivation for using eBPF over a ready-made approach is flexibility. Eventually, we want more than just byte counters keyed by 5-tuples.
> 
> For example, we want separate counters for traffic between pairs of AGW and orc8r services. At the network level, all this traffic may be going over a single socket (when _control_proxy_ is enabled) or over multiple sockets with the same destination host and port (when _control_proxy_ is disabled). eBPF will allow us to look into the authority HTTP header and figure out the destination service. We may also want to look at the HTTP/2 stream numbers in the future.
> 
I see, the control proxy makes it bit complicated.
anyways, I think it make sense to use tcp level hooks to collect the packet data and expand it later.


> Another example could be open fd counters as discussed in #8173. While we currently have the `process_open_fds` metric that is collected through the `psutil` python package, that method will not work if we wanted to find out more about the kind of fds opened, how long they stayed opened etc. eBPF will allow us to instrument the `open()` and `close()` syscalls and keep track of that.
> 
> The plan is to make _kernsnoopd_ a home for all such metric collection. `packet_count` is the first such collector. The upside of this program is that it does not miss short-lived processes (assuming their traffic is not processed asynchronously,) and accounts for non-TCP traffic as well. The downside is the expected high performance impact which I'm working on measuring. I [will also write a separate program](https://github.com/magma/magma/pull/8251#discussion_r675767217) that listens to TCP events and correctly implements the AGW service to orc8r service mapping.
> 

I am  fine with the metric collection service, its good addition to AGW. I​ am bit worried about the direction this packet stats collector has taken.

waqaraqeel: > I see, the control proxy makes it bit complicated.

Indeed. But it's complicated even without _control_proxy_ because the orc8r is listening on the same port for all cloud services.

> I​ am bit worried about the direction this packet stats collector has taken.

I understand and agree. We were worried about performance too and discussed it in the [design doc](https://magma.github.io/magma/docs/next/proposals/p018_control_network_metrics#performance). I'll change the `packet_count` program to use the `tcp_send` hook and then measure the performance impact to complete it.

pshelar: > > I see, the control proxy makes it bit complicated.
> 
> Indeed. But it's complicated even without _control_proxy_ because the orc8r is listening on the same port for all cloud services.
> 

You could make use of 5 tuples in case of direct orc8r connection. AFAICS with control proxy there is no way to figure it out without parsing the HTTP headers or mapping current pid to task (as you have done).

> > I​ am bit worried about the direction this packet stats collector has taken.
> 
> I understand and agree. We were worried about performance too and discussed it in the [design doc](https://magma.github.io/magma/docs/next/proposals/p018_control_network_metrics#performance). I'll change the `packet_count` program to use the `tcp_send` hook and then measure the performance impact to complete it.

Thanks.

lte/gateway/python/magma/kernsnoopd/ebpf/packet_count.bpf.c
waqaraqeel: Done.

lte/gateway/python/magma/kernsnoopd/ebpf/packet_count.bpf.c
waqaraqeel: @pshelar I have now switched from instrumenting `start_xmit` to `tcp_sendmsg` and posted performance evaluation results in PR summary above.

lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c
```diff
@@ -0,0 +1,51 @@
...
+
+    // ignore packets destined for localhost
+    // IPv4 127.0.0.1 == 16777343, 0.0.0.0 == 0
+    if ((key.daddr == 16777343 || key.daddr == 0)) {
+      return 0;
+    }
+
+    // read binary name, pid, source and destination IP address and port
+    bpf_get_current_comm(key.comm, TASK_COMM_LEN);
+    key.pid = bpf_get_current_pid_tgid() >> 32;;
```
pshelar, line 41: I hope this pid is more reliable then reading it from the xmit() hook.

lte/gateway/python/magma/kernsnoopd/handlers.py
```diff
@@ -0,0 +1,159 @@
...
+        table = bpf['dest_counters']
+        for key, count in table.items():
+            d_host = inet_ntop(AF_INET, pack('I', key.daddr))
+            service_name = None
+
+            try:
+                service_name = self._get_source_service(key)
+                # TODO: destination service name inference does not work
+                # get destination service from host and port
+                logging.info(
```
pshelar: let log it at debug level.

lte/gateway/python/magma/kernsnoopd/ebpf/common.bpf.h
```diff
@@ -0,0 +1,35 @@
...
+#include <net/sock.h>
+
+
+struct key_t {
+  // binary name (task->comm in the kernel)
+  char comm[TASK_COMM_LEN];
+  u32 pid;
+  // source and destination addresses and ports
+  u32 saddr;
+  u32 daddr;
```
pshelar, line 31: daddr is not set, can you remove it?

lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c
```diff
@@ -0,0 +1,51 @@
...
+    key.saddr = sk->__sk_common.skc_rcv_saddr;
```
pshelar: why do you need saddr?

pshelar (commented):
Thanks for making the change and benchmarking it.

lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c
waqaraqeel: As far as I know, `tcp_sendmsg()` doesn't happen asynchronously. It has to be called from the right task, and I'm not seeing any 0 pids.

lte/gateway/python/magma/kernsnoopd/ebpf/common.bpf.h
waqaraqeel: daddr is set here: https://github.com/magma/magma/pull/8251/files/ea268589335a91178784cf10ff44830d0a63b392#diff-2c26eb9ca27c795eb35848591cbefa270853c569ee3df6253eecf352076d0813R31

lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c
waqaraqeel: We need the (saddr, lport, daddr, dport) to keep track of TCP sessions. We aren't doing it now, but we need it for destination service mapping, which is a TODO item in this PR.

lte/gateway/python/magma/kernsnoopd/handlers.py
waqaraqeel: Done.

lte/gateway/python/magma/kernsnoopd/ebpf/common.bpf.h
pshelar: ok, we can add the fields when we actually need it. you can remove it out for now. I am not fan of unused code.

lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c
pshelar: yes, I expected it. in most of cases tcp_sendmsg() is called in the userspace context.

lte/gateway/python/magma/kernsnoopd/ebpf/common.bpf.h
waqaraqeel: Removed.

lte/gateway/python/magma/kernsnoopd/ebpf/common.bpf.h
```diff
@@ -0,0 +1,34 @@
...
+#include <linux/sched.h>
+#include <uapi/linux/ptrace.h>
+#include <net/sock.h>
+
+
+struct key_t {
+  // binary name (task->comm in the kernel)
+  char comm[TASK_COMM_LEN];
+  u32 pid;
+  // source port and destination IP address, port
```
pshelar, line 30: can you update this comment.

lte/gateway/python/magma/kernsnoopd/handlers.py
```diff
@@ -0,0 +1,159 @@
...
+            Magma service name inferred from key
+
+        Raises:
+            ValueError: Could not infer service name from key
+        """
+
+        try:
+            # get python service name from command line args
+            # e.g. "python3 -m magma.state.main"
+            cmdline = psutil.Process(pid=key.pid).cmdline()
```
pshelar, line 115: it would be nice to cache it.

lte/gateway/python/magma/kernsnoopd/handlers.py
```diff
@@ -0,0 +1,159 @@
...
+        Handle() reads counters from the loaded byte_count program stored as
+        a dict in 'dest_counters' with key type key_t and value type counter_t
+        defined in ebpf/common.bpf.h
+
+        Args:
+            bpf: bcc.BPF object that was used to load eBPF program into kernel
+        """
+        table = bpf['dest_counters']
+        for key, count in table.items():
+            d_host = inet_ntop(AF_INET, pack('I', key.daddr))
```
pshelar, line 77: it would be nice to cache it.

pshelar (approved):
Thanks a lot for working on this.
I have couple of minor comment, you can address it in follow up PR.

PR #8260, opened 2021-07-24
ci: Remove unused semantic.yml

hcgatewood:
## Additional Information


hcgatewood:
Closing for now actually because I can't sort out which of the GH Actions checks to mark as required

PR #8272, opened 2021-07-26
feat(pipelined): Pipelined side for polling shards

veshkemburu:
## Summary

For our new sharding model, pipelined side needs to be updated to be able to process shard ids. We need to add these to ActivateFlowsRequest and update the snapshots for pipelined tests. Fix error causing some test failures. 

## Test Plan

Run pipelined unit tests.

## Additional Information


lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -194,7 +193,7 @@ def _handle_error(self, ev):
         self._msg_hub.handle_error(ev)
 
     # pylint: disable=protected-access,unused-argument
-    def _get_rule_match_flow_msgs(self, imsi, _, __, ip_addr, ambr, rule, version):
+    def _get_rule_match_flow_msgs(self, imsi, _, __, ip_addr, ambr, rule, version, shard_id):
```
themarwhal, line 196:
```suggestion
    def _get_rule_match_flow_msgs(self, imsi, _, _, ip_addr, ambr, rule, version, shard_id):
```

veshkemburu: Weird, that wasn't something I had changed(only added the shard id parameter).

koolzz (approved):

koolzz:
@veshkemburu do integ tests pass fine?

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
veshkemburu: It actually has to be double underscore since it's a duplicate parameter

pshelar (approved):
LGTM

lte/gateway/python/magma/pipelined/app/gy.py
```diff
@@ -283,7 +283,7 @@ def _get_rule_match_flow_msgs(self, imsi, msisdn: bytes, uplink_tunnel: int, ip_
                     imsi, msisdn, uplink_tunnel, ip_addr, apn_ambr, flow, rule_num, priority,
                     rule.qos, rule.hard_timeout, rule.id, rule.app_name,
                     rule.app_service_type, self.next_service_table,
-                    version, self._qos_mgr, self._enforcement_stats_tbl))
+                    version, self._qos_mgr, shard_id, self._enforcement_stats_tbl))
```
koolzz: ah I hate how long cwf integ runs... the issue is in this function call, shard_id should be the last argument

koolzz (changes requested):

koolzz (approved):

PR #8278, opened 2021-07-26
feat(orc8r): Generic SyncStore interface for propagating digest pattern

wangyyt1013:

## Summary

1. Create Orc8r module that implements the `SyncStore` interface, which stores digests and cached objects following the specifications of the digest pattern
2. (TODO in upcoming dependent PR) Adapt `subscriberdb` and `subscriberdb_cache` to migrate to this interface

## Test Plan

2. Added tests for the `SyncStore` implementation in `orc8r/cloud/go/syncstore`

## Additional Information


orc8r/cloud/go/syncstore/syncstore.go
```diff
@@ -0,0 +1,87 @@
...
+ limitations under the License.
+*/
+
+package syncstore
+
+import (
+	"magma/orc8r/lib/go/protos"
+)
+
+const (
```
hcgatewood: Move these consts to their respectively relevant files. This file should just contain the interfaces

orc8r/cloud/go/syncstore/syncstore.go
```diff
@@ -0,0 +1,87 @@
...
+type CacheWriter interface {
+	// InsertMany adds objects to the list staged for the batch update.
+	// NOTE: Caller of the function should enforce that the max size of the
+	// insertion aligns reasonably with the max page size of its corresponding
+	// load source.
+	InsertMany(objects map[string][]byte) error
+	// Apply completes the batch cache update.
+	Apply() error
+	// SetInvalid sets this writer to be invalid for future writes.
+	SetInvalid()
```
hcgatewood: As discussed, I don't think SetInvalid is necessary

orc8r/cloud/go/syncstore/writer.go
```diff
@@ -0,0 +1,240 @@
...
+	"magma/orc8r/lib/go/protos"
+
+	"github.com/Masterminds/squirrel"
+	"github.com/golang/protobuf/proto"
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+	"github.com/thoas/go-funk"
+)
+
+type syncStore struct {
```
hcgatewood: You can split the reader/writer implementations into separate files if you want, but you only need one backing object. Then you can have `NewSyncStore` and `NewSyncStoreReader` functions that will return the relevant interface types (achieving what we want), but those funcs can implementationally just return the same `syncStore` under the covers

Lmk if that makes sense, we can chat on Monday if desired

orc8r/cloud/go/syncstore/reader.go
```diff
@@ -0,0 +1,245 @@
...
+	db      *sql.DB
+	builder sqorc.StatementBuilder
+	fact    blobstore.BlobStorageFactory
+}
+
+func NewSyncStoreReader(db *sql.DB, builder sqorc.StatementBuilder, fact blobstore.BlobStorageFactory) SyncStoreReader {
+	return &syncStoreReader{db: db, builder: builder, fact: fact}
+}
+
+func (l *syncStoreReader) Initialize() error {
```
hcgatewood: I think the various table names should all be generated with a prefix that's passed in to the `New` func -- this will allow multiple different types of sync stores to be created (subscribers, apns, mconfigs, etc), without causing namespace issues on the table names

orc8r/cloud/go/syncstore/reader.go
```diff
@@ -0,0 +1,245 @@
...
+
+		digestTrees := map[string]*protos.DigestTree{}
+		for rows.Next() {
+			network, rootDigest, leafDigestsMarshaled, lastUpdatedTime := "", "", []byte{}, int64(0)
+			err = rows.Scan(&network, &rootDigest, &leafDigestsMarshaled, &lastUpdatedTime)
+			if err != nil {
+				return nil, errors.Wrapf(err, "get digests for network %+v, SQL row scan error", network)
+			}
+
+			digestTree := &protos.DigestTree{
```
hcgatewood: Nit: single-line this

orc8r/protos/digest.proto
```diff
@@ -0,0 +1,44 @@
...
+
+message LeafDigest {
+  // id is the network-wide unique identifier of the object.
+  string id = 1;
+  // digest is the deterministic digest of the object.
+  Digest digest = 2;
+}
+
+// LeafDigestsToSerialize is used to serialize a list of leaf digests en masse.
+message LeafDigestsToSerialize {
```
hcgatewood: Just call this "LeafDigests" and replace instances of "repeated LeafDigest" with this "LeafDigests", then add a short note that you did it this way so you can serialize a set of leaf digests

orc8r/cloud/go/syncstore/reader.go
```diff
@@ -0,0 +1,245 @@
...
+				RootDigest: &protos.Digest{Md5Base64Digest: rootDigest},
+			}
+			if loadLeaves {
+				leafDigests := &protos.LeafDigestsToSerialize{}
+				err = proto.Unmarshal(leafDigestsMarshaled, leafDigests)
+				if err != nil {
+					return nil, errors.Wrapf(err, "unmarshal leaf digests for network %+v", network)
+				}
+				digestTree.LeafDigests = leafDigests.GetDigests()
```
hcgatewood: Don't use guarded getter since you don't expect leafDigests to be nil

orc8r/cloud/go/syncstore/reader.go
```diff
@@ -0,0 +1,245 @@
...
+	ret := txRet.(map[string]*protos.DigestTree)
+	return ret, nil
+}
+
+func (l *syncStoreReader) GetCachedByID(network string, ids []string) ([][]byte, error) {
+	txFn := func(tx *sql.Tx) (interface{}, error) {
+		rows, err := l.builder.
+			Select(idCol, objCol).
+			From(cacheTableName).
+			Where(squirrel.And{
```
hcgatewood: General: have a newline after the opening paren and before the closing paren (move squirrel.And to its own line)

orc8r/cloud/go/syncstore/reader.go
```diff
@@ -0,0 +1,245 @@
...
+	objs, lastIncludedId, err := parseRows(rows)
+	if err != nil {
+		return nil, err
+	}
+
+	// Return empty token if we have definitely fetched all data in the db
+	if uint64(len(objs)) < pageSize {
+		return &pageInfo{token: "", objects: objs}, nil
+	}
+	nextToken, err := configurator_storage.SerializePageToken(&configurator_storage.EntityPageToken{
```
hcgatewood: Nit: create the token var on 1 line, then serialized on the next 1 line

orc8r/cloud/go/syncstore/writer.go
```diff
@@ -0,0 +1,240 @@
...
+	builder      sqorc.StatementBuilder
+	fact         blobstore.BlobStorageFactory
+}
+
+func NewSyncStore(db *sql.DB, builder sqorc.StatementBuilder, fact blobstore.BlobStorageFactory) SyncStore {
+	reader := NewSyncStoreReader(db, builder, fact)
+	return &syncStore{SyncStoreReader: reader, cacheWriters: map[string]CacheWriter{}, db: db, builder: builder, fact: fact}
+}
+
+func (l *syncStore) CollectGarbage(trackedNetworks []string) error {
```
hcgatewood: This impl is too long (too much info / not easy to scan). Split it up into private methods, one-per-table, and call those successively from CollectGarbage

orc8r/cloud/go/syncstore/writer.go
```diff
@@ -0,0 +1,240 @@
...
+				l.cacheWriters[blob.Key].SetInvalid()
+				delete(l.cacheWriters, blob.Key)
+			}
+		}
+	}
+	return store.Commit()
+}
+
+func (l *syncStore) SetDigest(network string, digests *protos.DigestTree) error {
+	rootDigest := digests.GetRootDigest().GetMd5Base64Digest()
```
hcgatewood: Same here/throughout: don't use guarded getters unless there's a realistic reason why the method's object would be nil

orc8r/cloud/go/syncstore/writer.go
```diff
@@ -0,0 +1,240 @@
...
+			return nil, errors.Wrapf(err, "insert digests for network %+v", network)
+		}
+		return nil, nil
+	}
+	_, err = sqorc.ExecInTx(l.db, nil, nil, txFn)
+	return err
+}
+
+func (l *syncStore) UpdateCache(network string) (CacheWriter, error) {
+	// the temporary table is namespaced with the unique id of the cacheWriter
```
hcgatewood: Sentence-case comments

orc8r/cloud/go/syncstore/writer.go
```diff
@@ -0,0 +1,240 @@
...
+	writerId := fmt.Sprintf("cache_writer_%s", strings.Replace((&storage.UUIDGenerator{}).New(), "-", "_", -1))
```
hcgatewood: Move this line to a separate fn, including info on why you're doing the replace

Also "ID" rather than "Id"

orc8r/cloud/go/syncstore/writer.go
```diff
@@ -0,0 +1,240 @@
...
+			RunWith(tx).
+			Exec()
+		return nil, errors.Wrap(err, "create cached objs tmp table")
+	}
+	_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
+	if err != nil {
+		return nil, err
+	}
+
+	// The creation time of cacheWriters is tracked by the store for garbage collection
```
hcgatewood: This comment actually belongs to the subsequent paragraph. But just make those subsequent paragraphs (recording the tx start time) their own private method

orc8r/cloud/go/syncstore/writer.go
```diff
@@ -0,0 +1,240 @@
...
+	}})
+	if err != nil {
+		return nil, errors.Wrapf(err, "set start time of network %+v, cachewriter %+v in blobstore", network, writerId)
+	}
+	err = store.Commit()
+	if err != nil {
+		return nil, err
+	}
+
+	l.cacheWriters[writerId] = NewCacheWriter(network, writerId, l.db, l.builder)
```
hcgatewood: No need to track the cache writers like this. If the cache writer lives longer than let's say 30m, it's clearly bad, so it's fine for the cache writer to just fail on the error "tmp table not found"

orc8r/cloud/go/syncstore/writer.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (l *syncStore) RecordResync(network string, gateway string, t uint64) error {
+	store, err := l.fact.StartTransaction(&storage.TxOptions{ReadOnly: true})
+	if err != nil {
+		return errors.Wrapf(err, "error starting transaction")
+	}
+	defer store.Rollback()
+
+	lastResyncBytes := make([]byte, 8)
```
hcgatewood: Add a little helper fn somewhere to make the "putuint64" a one-liner, which you can then inline into the CreateOrUpdate call (here and elsewhere)

Also, just use a generic byte slice of no specified len/cap, since 8 bytes doesn't matter for performance and that way we don't have to worry about any sort of byte len issues

orc8r/cloud/go/syncstore/cache_writer.go
```diff
@@ -0,0 +1,145 @@
...
+	txFn := func(tx *sql.Tx) (interface{}, error) {
+		_, err := insertQuery.RunWith(tx).Exec()
+		return nil, errors.Wrapf(err, "insert objs into store for network %+v", l.network)
+	}
+	_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
+	return err
+}
+
+func (l *cacheWriter) Apply() error {
+	if l.invalidWriter {
```
hcgatewood: So (a) just call this "applied"/"hasApplied"/"invalid", and set it immediately after this early return. And (b) there's no need to set it via the SetInvalid method, we can just let this silently fail if garbage collection gets to the tmp table before the final apply

orc8r/cloud/go/syncstore/syncstore.go
hcgatewood: Nit: also, let's rename the files just a bit
- syncstore{,_test}.go -> store{,_test}.go
- {reader,writer}.go -> store_{reader,writer}.go (if you decide to keep the separate reader/writer files
- cache_writer.go -> cache.go

hcgatewood (changes requested):
Wow looks awesome! Very pretty code, looks like the functionality is spot on as well. Couple requests, mostly nits, then let's get this merged on Monday

orc8r/cloud/go/syncstore/reader.go
wangyyt1013: My concern with replacing `repeated LeafDigest` in other messages with `LeafDigests` was that it will result in an extra layer of data abstraction, and we'd have to unpack an extra data object to get to the list of leaf digests. Given that, not sure if I should choose consistency (using `LeafDigests` everywhere) over simplicity (directly use `repeated LeafDigest` in `DigestTree`)? @hcgatewood

orc8r/protos/digest.proto
wangyyt1013: My concern with replacing `repeated LeafDigest` in other messages with `LeafDigests` was that it will result in an extra layer of data abstraction, and we'd have to unpack an extra data object to get to the list of leaf digests. Given that, not sure if I should choose consistency (using `LeafDigests` everywhere) over simplicity (directly use `repeated LeafDigest` in `DigestTree`)? @hcgatewood

orc8r/cloud/go/syncstore/writer.go
wangyyt1013: Looks like `PutUint64` requires a slice with a fixed length to put the value in, and declaring a generic slice through `[]byte{}` didn't work; wondering if there's anything that I'm missing? @hcgatewood

hcgatewood: Yeah make your own `PutUint64` method (either private and local to this file, or public and in the `math` pkg we created recently), which makes it a one-liner -- it will basically be a ~3 line wrapper around the stdlib putuint

orc8r/protos/digest.proto
hcgatewood: Hm. You're right that it does feel gross using a Message when we really want a slice. Okay I'm fine with leaving the others as slices (`repeated`), but still let's make this one `LeafDigests` and add a note that we're using a separate msg with the intention of it being for serialization

Rationale: I don't like naming things with intent -- the name should describe "what it is" not "how it's used". But adding a comment on "why" is totally fine

orc8r/cloud/go/syncstore/writer.go
wangyyt1013: Still kept the separate files for clarity, but adopted the method of having one single `syncStore` struct @hcgatewood

orc8r/cloud/go/syncstore/store_reader.go
```diff
@@ -0,0 +1,259 @@
...
+	leafDigestsCol     = "leaf_digests"
+	objCol             = "obj"
+	lastUpdatedTimeCol = "last_updated_at"
+
+	lastResyncBlobstoreType  = "gateway_last_resync_time"
+	cacheWriterBlobstoreType = "cache_writer_creation_time"
+)
+
+func NewSyncStoreReader(db *sql.DB, builder sqorc.StatementBuilder, fact blobstore.BlobStorageFactory, config Config) SyncStoreReader {
+	storeReader := &syncStore{
```
hcgatewood, line 51: Since you're passing in a config value, add a config.Validate method (and call it), to set e.g. that the validIntervalSecs etc is set (> 0) and prefix is set (!= "")

orc8r/cloud/go/syncstore/store_reader.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (l *syncStore) GetCachedByPage(network string, token string, pageSize uint64) ([][]byte, string, error) {
+	lastIncludedId := ""
+	if token != "" {
+		decoded, err := configurator_storage.DeserializePageToken(token)
+		if err != nil {
+			return nil, "", err
+		}
+		lastIncludedId = decoded.LastIncludedEntity
```
hcgatewood: Nit here+elsewhere: "ID" not "Id"

orc8r/cloud/go/syncstore/store_reader.go
```diff
@@ -0,0 +1,259 @@
...
+		return uint64(0), errors.Wrapf(err, "get last resync time of network %+v, gateway %+v from blobstore", network, gateway)
+	}
+
+	lastResync := binary.LittleEndian.Uint64(blob.Value)
+	return lastResync, store.Commit()
+}
+
+// parsePage parses the list of cached serialized objs, as well as returns the next
+// page token based on the lastIncludedEntity in the current page.
+// NOTE: The configurator_storage.EntityPageToken is used for simplicity & ease
```
hcgatewood, line 226: Add a (commented) newline before this note, to clarify that it's an addendum rather than the core docstring summary

orc8r/cloud/go/syncstore/store_writer.go
```diff
@@ -0,0 +1,393 @@
...
+	err = l.recordCacheWriterStartTime(network, writerID)
+	if err != nil {
+		return nil, errors.Wrapf(err, "record start time of cache writer %+v of network %+v", writerID, network)
+	}
+
+	return l.NewCacheWriter(network, writerID), nil
+}
+
+func (l *syncStore) RecordResync(network string, gateway string, t uint64) error {
+	store, err := l.fact.StartTransaction(&storage.TxOptions{ReadOnly: true})
```
hcgatewood: Not readonly! Cause we're doing a write. Double-check this elsewhere as well

orc8r/cloud/go/syncstore/store_writer.go
```diff
@@ -0,0 +1,393 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "set last resync time of network %+v, gateway %+v in blobstore", network, gateway)
+	}
+
+	return store.Commit()
+}
+
+// recordCacheWriterStartTime registers the creation time of a cache writer in store.
+func (l *syncStore) recordCacheWriterStartTime(network string, writerID string) error {
+	store, err := l.fact.StartTransaction(&storage.TxOptions{ReadOnly: true})
```
hcgatewood: Same here: not readonly, so can just omit the txoptions I think

orc8r/cloud/go/syncstore/store_writer.go
```diff
@@ -0,0 +1,393 @@
...
+		Value: encodeUint64ToBytes(uint64(clock.Now().Unix())),
+	}})
+	if err != nil {
+		return errors.Wrapf(err, "set start time of network %+v, cachewriter %+v in blobstore", network, writerID)
+	}
+	return store.Commit()
+}
+
+func (l *syncStore) CollectGarbage(trackedNetworks []string) {
+	err := l.collectGarbageSQL(trackedNetworks, []string{l.digestTableName, l.cacheTableName})
```
hcgatewood: Probably don't need to pass the table names cause they're already part of the obj

orc8r/cloud/go/syncstore/store_writer.go
```diff
@@ -0,0 +1,393 @@
...
+		}
+		_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
+		if err != nil {
+			multierror.Append(errs, errors.Wrapf(err, "collect garbage for table %+v", tableName))
+		}
+	}
+	return errs.ErrorOrNil()
+}
+
+func (l *syncStore) getInStoreNetworksSQL(tx *sql.Tx, tableName string) ([]string, error) {
```
hcgatewood: Nit throughout: I like "stored" rather than "inStore" -- so "storedNetworks"

orc8r/cloud/go/syncstore/store_writer.go
```diff
@@ -0,0 +1,393 @@
...
+	if err != nil {
+		multierror.Append(errs, err)
+	}
+	return errs.ErrorOrNil()
+}
+
+func (l *syncStore) collectGarbageCacheWriter(tracked []string) error {
+	errs := &multierror.Error{}
+
+	invalidByNetwork, err := getInvalidCacheWriter(l.fact, tracked, l.cacheWriterValidIntervalSecs)
```
hcgatewood: Yeah here and throughout: any time you find yourself passing in an attribute of the object, consider whether you can just make the func a method and using those attributes from inside the method

orc8r/cloud/go/syncstore/store_writer.go
```diff
@@ -0,0 +1,393 @@
...
+			multierror.Append(errs, errors.Wrapf(err, "list cache writers of deleted network %+v", network))
+			continue
+		}
+		invalidByNetwork[network] = keys
+	}
+
+	for _, network := range tracked {
+		filter := blobstore.CreateSearchFilter(&network, []string{cacheWriterBlobstoreType}, nil, nil)
+		criteria := blobstore.LoadCriteria{LoadValue: true}
+		networkBlobs, err := store.Search(filter, criteria)
```
hcgatewood: There's probably a static method you can use that wraps Search, to make this cleaner

orc8r/cloud/go/syncstore/store_writer.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+// generateCacheWriterUUID returns a universally unique ID for a cache writer.
+func generateCacheWriterUUID(tableNamePrefix string) string {
+	// Replace "-" symbols in the ID since they aren't supported by SQL variable names
+	id := strings.Replace((&storage.UUIDGenerator{}).New(), "-", "_", -1)
+	return fmt.Sprintf("%s_cache_writer_%s", tableNamePrefix, id)
+}
+
+func encodeUint64ToBytes(n uint64) []byte {
```
hcgatewood: Nit: can just call this encodeUint64, it's mostly implicit that you'd be encoding it to bytes

orc8r/cloud/go/syncstore/cache.go
```diff
@@ -0,0 +1,142 @@
...
+	network        string
+	id             string
+	cacheTableName string
+	db             *sql.DB
+	builder        sqorc.StatementBuilder
+	invalid        bool
+}
+
+func (l *syncStore) NewCacheWriter(network string, id string) CacheWriter {
+	return &cacheWriter{network: network, id: id, db: l.db, builder: l.builder, cacheTableName: l.cacheTableName, invalid: false}
```
hcgatewood: Create this object on multiple lines cause it's so long

hcgatewood (approved):
Looks great! Couple nits then msg me to merge

orc8r/cloud/go/syncstore/cache.go
```diff
@@ -0,0 +1,149 @@
...
+	id             string
+	cacheTableName string
+	db             *sql.DB
+	builder        sqorc.StatementBuilder
+	invalid        bool
+}
+
+func (l *syncStore) NewCacheWriter(network string, id string) CacheWriter {
+	writer := &cacheWriter{
+		network: network, id: id,
```
hcgatewood: This is weird, use separate lines

orc8r/cloud/go/syncstore/cache.go
```diff
@@ -0,0 +1,149 @@
...
+	"magma/orc8r/cloud/go/sqorc"
+
+	"github.com/Masterminds/squirrel"
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+)
+
+type cacheWriter struct {
+	network        string
+	id             string
```
hcgatewood: Throughout: I think "id" is actually the "type" of the cached objects? Let's use "type" if so

orc8r/cloud/go/syncstore/config.go
```diff
@@ -0,0 +1,43 @@
...
+	// NOTE: the caller should enforce that this value is smaller than the service
+	// worker loop interval, to prevent workers with "older" cache writers overwriting
+	// concurrent workers with "newer" ones.
+	CacheWriterValidIntervalSecs int64
+	// TableNamePrefix is used to namespace all syncstore tables to prevent
+	// naming collisions among different services using syncstore.
+	TableNamePrefix string
+}
+
+func (config *Config) Validate(writer bool) error {
```
hcgatewood, line 34: Why have this bool if it's always false?

orc8r/cloud/go/syncstore/store_reader.go
```diff
@@ -0,0 +1,264 @@
...
+	cacheWriterBlobstoreType = "cache_writer_creation_time"
+)
+
+func NewSyncStoreReader(db *sql.DB, builder sqorc.StatementBuilder, fact blobstore.BlobStorageFactory, config Config) (SyncStoreReader, error) {
+	err := config.Validate(false)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid configs for syncstore reader")
+	}
+	storeReader := &syncStore{
+		db: db, builder: builder, fact: fact,
```
hcgatewood: Here and elsewhere: eh just make these separate lines, usually I don't like mixing single vs multiple per-line

orc8r/cloud/go/syncstore/store_reader.go
```diff
@@ -0,0 +1,264 @@
...
+		if err != nil {
+			return nil, "", err
+		}
+		lastIncludedID = decoded.LastIncludedEntity
+	}
+	txFn := func(tx *sql.Tx) (interface{}, error) {
+		rows, err := l.builder.
+			Select(idCol, objCol).
+			From(l.cacheTableName).
+			Where(squirrel.And{
```
hcgatewood: Move the squirrel.And to its own line, here and elsewhere

orc8r/cloud/go/syncstore/store_test.go
```diff
@@ -0,0 +1,333 @@
...
+	"magma/orc8r/lib/go/protos"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSyncStore(t *testing.T) {
+	db, err := sqorc.Open("sqlite3", ":memory:")
+	assert.NoError(t, err)
+	fact := blobstore.NewSQLBlobStorageFactory("last_resync_time", db, sqorc.GetSqlBuilder())
```
hcgatewood: Here and elsewhere: use this instead https://github.com/magma/magma/blob/1b784496f3577c020f13236e815488cd8d9ee670/orc8r/cloud/go/test_utils/test_db.go#L49-L51

hcgatewood (approved):
Couple nits 🙂

orc8r/cloud/go/syncstore/config.go
wangyyt1013: it's true when used to initialize `SyncStore`, and false when used to initialize `SyncStoreReader` @hcgatewood

hcgatewood: Ah I see, GitHub was hiding the writer.go contents from me 🙂

wangyyt1013: the file was too large 🥲

orc8r/cloud/go/syncstore/cache.go
wangyyt1013: I thought the `id` here is the unique identifier for each cache writer (which is also conveniently used as the name for the temporary SQL table) @hcgatewood

PR #8279, opened 2021-07-27
fix(mme): Fix memory leak in TAU procedure

### LINKED ISSUE #8326: [agw] MME reaches high memory consumption / leaks memory

### Your Environment

- **Version:** 1.5.2
- **Affected Subcomponent:** MME
- **Deployment Environment:** bare-metal (AGW)

### Describe the Issue

- On production AGW deployment, multiple users have experienced and reported high memory consumption of MME service after an extended amount of time with around 30 UEs attached.
- This has also been reproducible on Spirent testing, using active/idle switching test case, after around 3 hours of iterations the service would hit memory limits (~511 MBs)

**Expected behavior**
- Memory should not keep increasing in a steadily manner, expected memory of MME service on a loaded system reaches between 250-300 UEs. The memory consumption with the tests have made the system reach 500MBs

**Screenshots**
image

ulaskozat:
## Summary

Backports only the TAU part of the memory leak fix in #8029.  The other part was introduced and fixed on the same release cycle, hence it does have no effect on v1.5.
 
## Test Plan

Integ tests. Spirent tests.

## Additional Information


ardzoht (approved):
lgtm

PR #8281, opened 2021-07-27
Idle Mode and Service Request Procedure

VinayKumar1591:
feat(agw)[5G][SA] Idle mode and service request procedure changes 

## Summary

1. Added changes to  support  Idle mode and service request procedure

## Test Plan

1. Tested on NGAP and DS Tester


VinayKumar1591:
New PR will be raised

PR #8285, opened 2021-07-27
fix: change nlohmann-json import path to be absolute

themarwhal:

## Summary
Fixing up the json include path to use the full path, not just json.hpp. (It works either way)

This change also includes clean up work to remove the unnecessary symlink from vagrant/docker.

## Test Plan
All unit tests
Built AGW and tested in phyu2

## Additional Information


rdefosse (approved):

lte/gateway/docker/mme/Dockerfile.rhel8
```diff
@@ -344,8 +344,7 @@ RUN cd json && \
     cd _build && \
     cmake3 -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release .. && \
     make -j`nproc` && \
-    make install && \
-    cp /usr/local/include/nlohmann/json.hpp /usr/local/include/
```
rdefosse, line 348: @themarwhal  Marie

This modification is not tested in the OAI CI pipeline. Only when your PR is `merged`. But it should be OK.

Ping me when it is merged. I will also have to recreate a base image with this modification.

Regards,
Raphael

electronjoe (approved):
Thanks for the reduction of work involving the phrase "hacks" =D

lte/gateway/docker/mme/Dockerfile.rhel8
themarwhal: Is there a way to ensure nothing breaks before merging?

rdefosse: Let me try manually

rdefosse: @themarwhal Marie

OK for me. You can merge!

ardzoht (approved):

PR #8308, opened 2021-07-27
chore: fix weird include patterns in common

themarwhal:

## Summary
Observed that we're including some prometheus related files in a weird way. Fixing them up so that it explicitly imports from prometheus.

## Test Plan
make build + tests locally

## Additional Information


ardzoht (approved):
lgtm

Issue #8299, opened 2021-07-27
bettereng: C++ builds - Use full path from root for includes rather than relative paths


themarwhal:

themarwhal:
@electronjoe completed this for all services except sessiond. closing for now.

PR #8319, opened 2021-07-28
feat: bazelify orc8r/gateway/c/common

themarwhal:
## Summary
This PR adds all BUILD files necessary to build all targets and unit tests in `orc8r/gateway/c/common`

Following the Bazel styleguide on naming conventions, etc. (https://docs.bazel.build/versions/main/skylark/build-style.html#build-style-guide)

## Issues and workarounds
* Bazel expects C++ headers to be full paths whereas our paths are relative to where the CMake sub projects are defined. (Instead of doing `#include "orc8r/gateway/c/common/service_registry/includes/ServiceRegistrySingleton.h"`, we do `#include "includes/ServiceRegistrySingleton.h"`)
  * To avoid doing a major refactor on this in all C++ files, I've added various `strip_include_prefix` in the libraries to make Bazel happy. (This makes it so that the files can be used without some specified prefix). If we decide to fully migrate to Bazel, this is definitely something that needs fixing up. (Issue created for this: https://github.com/magma/magma/issues/8299)
 * I've also had to add a custom BUILD file for cpp-redis as the version that we use does not natively support bazel. But good news here is that newer versions of cpp-redis does! (Added another issue to update cpp-redis eventually: https://github.com/magma/magma/issues/8321)
 * I've had to add a custom BUILD file for `prometheus-cpp` and its dependencies as the one in the repo depends on a prometheus client proto model that is not compatible with the metrics.proto file we use in repo.
* WORKSPACE file was getting fairly large so I moved some logic into `third_party_repositories.bzl` - still trying to figure out best practices on nicely managing repo-level dependencies in a readable way :) 

## Other notes
* All dependencies in bazel, regardless of where they are used, are declared at the top level. For now, I've split them into `WORKSPACE.bazel` and `third_party_repositories.bzl`. Bazel will only pull dependencies as required.
* I've filed an issue to eventually flatten out the orc8r/gateway/c/common to a single directory structure. (https://github.com/magma/magma/issues/8453) We can get small modular libraries with Bazel for free, so it'll make sense to get rid of this unnecessary structure. 
* I've filed an issue to upgrade the protobuf version we use. (https://github.com/magma/magma/issues/8457) We're currently using at least one method (`status.error_message()`) that is now deprecated in the latest prometheus version. Took a bit of effort to find the correct version to pull as we don't label any version information in the current build :'( 
* The test size in `cc_test` determines the appropriate timeout. Trying to see if there's a way to generate these values, but not sure yet. (doc: https://docs.bazel.build/versions/main/be/common-definitions.html#test.size)
* `.bazelversion` at the root specifies what bazel version to use
* `.bazelrc` at the root will be the place to configure any project wide settings (set c++ version as 14, set out directory to /tmp/bazel)

## Test Plan

Verified the existing build in magma VM.

Followed the instructions on building and execing into the bazel-base docker container: https://github.com/magma/magma/tree/master/experimental/bazel-base

Running all tests in the repo (for now it is the orc8r/gateway/c/common tests)
```bash
root@15df9af5e1e4:/magma# bazel test ...
INFO: Analyzed 118 targets (82 packages loaded, 6015 targets configured).
INFO: Found 113 targets and 5 test targets...
INFO: Elapsed time: 4.248s, Critical Path: 0.03s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
//orc8r/gateway/c/common/config/test:config_test                (cached) PASSED in 0.0s
//orc8r/gateway/c/common/service303/test:magma_service_test     (cached) PASSED in 0.2s
//orc8r/gateway/c/common/service303/test:metrics_test           (cached) PASSED in 0.2s
//orc8r/gateway/c/common/service303/test:service303_test        (cached) PASSED in 1.6s
//orc8r/gateway/c/common/service_registry/test:service_registry_test (cached) PASSED in 0.1s

Executed 0 out of 5 tests: 5 tests pass.
INFO: Build completed successfully, 1 total action
root@15df9af5e1e4:/magma#
```

To format all bazel changes, I ran
```bash
bazel run //:buildifier
```

## Additional Information


WORKSPACE
```diff
@@ -1,5 +1,6 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+load("//:third_party_repositories.bzl", "boost", "cpp_testing_deps", "grpc", "nlohmann_json", "prometheus_cpp_deps", "protobuf", "yamlcpp", "zlib")
```
electronjoe: Loading this way does reduce the number of lines for these external resource imports. I suppose it's nice to have them all in one file and clearly pinned.  The other form of complexity of this file is all of the sequencing dependencies on invocation of the resource configurations. Not sure if that can be bundled out of sight here more so - or if hiding that complexity is a bad idea... Wonder if @markjen has any experience in abstracting WORKSPACE setup complexity.

orc8r/gateway/c/common/async_grpc/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "async_grpc",
+    srcs = [
+        "GRPCReceiver.cpp",
+    ],
+    hdrs = [
+        "includes/GRPCReceiver.h",
```
electronjoe: Here and elsewhere in PR. I understand that we may not want to move these files around yet - but this `includes/...` pattern is tech debt for a monorepo, and even more so for Bazel. But it's likely all over the place and perhaps a separate discussion. Is this so common that we don't want a `TODO(GH8446): Remove includes/ project directories`?.

orc8r/gateway/c/common/async_grpc/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "async_grpc",
```
electronjoe: If your srcs is singular (which is a good goal often) - I'm used to the Bazel target name being 1:1 tied to the source file name.  So here, likely `grpc_receiver".  Is this badly named perhaps, and should be e.g. async_grpc?  If you have a group of files in a target, then it often makes sense to use a name that is more applicable to the collection (not the case here... yet).

orc8r/gateway/c/common/async_grpc/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
...
+    srcs = [
+        "GRPCReceiver.cpp",
+    ],
+    hdrs = [
+        "includes/GRPCReceiver.h",
+    ],
+    # TODO(@themarwhal): Migrate to using full path for includes - GH8299
```
electronjoe, line 19: Thanks for the breadcrumb!

electronjoe:
Did you apply the `buildifier` tool to auto-format these? I think you did. If so maybe mention that in your PR description so others can learn.

orc8r/gateway/c/common/config/BUILD.bazel
```diff
@@ -0,0 +1,24 @@
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "config",
```
electronjoe: When you bundle several compilation units like this - then it's OK to make up a reasonble name for the ensemble as you have done here.

orc8r/gateway/c/common/config/BUILD.bazel
```diff
@@ -0,0 +1,24 @@
...
+    srcs = [
+        "MConfigLoader.cpp",
```
electronjoe: Here and elsewhere in the PR. Are all three of these compilation units often necessary? If some dependencies may only need a subset, then we prefer to break up into a few cc_library declarations.  Though sometimes this is not possible due to e.g. cycles in dependencies between these compilation units (often a sign of tech debt itself).  If they are likely all necessary for some dep, then it's OK to group. Impact here is in compilation time mostly (see `To use fine-grained dependencies to allow parallelism and incrementality.` of [best practices](https://docs.bazel.build/versions/main/best-practices.html#best-practices)).

orc8r/gateway/c/common/config/test/BUILD.bazel
```diff
@@ -0,0 +1,13 @@
+load("@rules_cc//cc:defs.bzl", "cc_test")
+
+cc_test(
+    name = "config_test",
+    size = "small",
```
electronjoe, line 16: WOOHOO! Test sizes =D nice

electronjoe: You might call out what `size =` is doing in these tests (in top level PR description) as an educational call-out?

orc8r/gateway/c/common/eventd/BUILD.bazel
```diff
@@ -0,0 +1,24 @@
+package(default_visibility = ["//visibility:public"])
```
electronjoe, line 12: FYI: Public visibility is a fine default, but if there are areas of the code base that we don't want to "spread" any further, you can restrict visibility to existing dependencies! Probably uncommon though.

orc8r/gateway/c/common/service303/BUILD.bazel
```diff
@@ -0,0 +1,36 @@
...
+    hdrs = [
+        "ProcFileUtils.h",
+        "includes/MagmaService.h",
+        "includes/MetricsHelpers.h",
+        "includes/MetricsRegistry.h",
+        "includes/MetricsSingleton.h",
+    ],
+    # TODO(@themarwhal): Migrate to using full path for includes - GH8299
+    copts = [
+        "-DBAZEL=1",
```
electronjoe: Here ?and elsewhere?. Do you have a GH Issue for this DBazel work-around? To eventually remove it...

orc8r/gateway/c/common/service303/test/test_service303.cpp
```diff
@@ -48,7 +47,7 @@ namespace magma {
 
 class Service303Client {
  public:
-  Service303Client(const std::shared_ptr<Channel>& channel)
+  explicit Service303Client(const std::shared_ptr<Channel>& channel)
```
electronjoe, line 50: Was this a misc GCC warning fixup?

third_party/civetweb.BUILD
```diff
@@ -0,0 +1,25 @@
+cc_library(
+    name = "civetweb",
```
electronjoe, line 6: Just for my curiousity, what are we using the embedded http server for? Is this perhaps a dep of prometheus client or some such?

third_party/nlohmann_json.BUILD
```diff
@@ -0,0 +1,9 @@
+cc_library(
+    name = "json",
+    hdrs = glob([
+        "include/**/*.hpp",
+    ]),
+    includes = ["include"],
+    visibility = ["//visibility:public"],
+    alwayslink = 1,
```
electronjoe: Can you comment on the alwayslink = 1 here? And perhaps explain it's use in the top level PR description for visibility / education.   This is one I've not used before.

third_party/prometheus-cpp.BUILD
```diff
@@ -0,0 +1,33 @@
...
+        "lib/protobuf_delimited_serializer.h",
+        "lib/registry.cc",
+        "lib/serializer.h",
+        "lib/text_serializer.cc",
+        "lib/text_serializer.h",
+    ],
+    hdrs = glob([
+        "include/prometheus/*.h",
+    ]),
+    copts = ["-I."],
```
electronjoe: TODO(full paths) missing here?

third_party/prometheus-cpp.BUILD
```diff
@@ -0,0 +1,33 @@
...
+    linkstatic = 1,
```
electronjoe: Can you comment on the `linkstatic = 1` here? And perhaps explain it's use in the top level PR description for visibility / education. This is one I've not used before. Perhaps this third party library gets fed this linkstatic flag by Bazel to generate non-dynamic output? But since we don't use their build infra that's not clear to me how it matters (unless some sort of defines are used...)

third_party/civetweb.BUILD
electronjoe: Aah yes, below I see prometheus dependency

third_party/nlohmann_json.BUILD
```diff
@@ -0,0 +1,9 @@
+cc_library(
+    name = "json",
+    hdrs = glob([
+        "include/**/*.hpp",
+    ]),
+    includes = ["include"],
+    visibility = ["//visibility:public"],
```
electronjoe: Nit: here and elsewhere, I think package visibility is preferred unless you have heterogeneus package visibility needs. I get why here there may only ever be one target - but my brain scans for visibility package-wise first.  Also in other scenarios somebody will come along and add another package and likely need public visibility so you can save them some effort by putting it at package (far more regular than reducing visibility).

third_party/zlib.BUILD
```diff
@@ -0,0 +1,61 @@
...
+    "inftrees.h",
+    "trees.h",
+    "zconf.h",
+    "zlib.h",
+    "zutil.h",
+]
+
+_ZLIB_PREFIXED_HEADERS = ["zlib/include/" + hdr for hdr in _ZLIB_HEADERS]
+
+# In order to limit the damage from the `includes` propagation
```
electronjoe, line 23: Did you come up with this, or was it advised somewhere?

electronjoe (changes requested):
I think we should get a review by @markjen also - as the style you build up here for Bazel files will rapidly propagate everywhere =D

orc8r/gateway/c/common/async_grpc/BUILD.bazel
themarwhal: Yeah I think it was added to appease CMake, so if we do transition to Bazel we can act on that issue. (Will put it into the repo as a TODO so that I remember to do this.)

orc8r/gateway/c/common/async_grpc/BUILD.bazel
themarwhal: I see -  I was just taking the name from the directory. But assuming that eventually we'll flatten `orc8r/gateway/c/common` it seems reasonable to name it the same as the source files. (I'll add a TODO to rename the files to async_grpc_receiver, maybe.

orc8r/gateway/c/common/config/BUILD.bazel
themarwhal: Can look into breaking this up. Service config and mconfig could be separate, will see if there are weird dependency issues.

orc8r/gateway/c/common/config/test/BUILD.bazel
themarwhal: Good point - will do!

orc8r/gateway/c/common/service303/BUILD.bazel
themarwhal: Oh! I actually ended up not needing this I think :o Will double check and either remove or file an issue

orc8r/gateway/c/common/service303/test/test_service303.cpp
themarwhal: oh yes I think so - I can leave it out of this PR to keep it focused on bazel. (I think I saw it and couldn't resist haha)

third_party/civetweb.BUILD
themarwhal: A lot of these BUILD files for third_party I got from somewhere else. (I'm very sad I didn't include it in the build file initially) I'll find the links and include it.

third_party/zlib.BUILD
themarwhal: I think I got this from somewhere else - will search where

themarwhal:
Adding @markjen also to get maximum comments :D

WORKSPACE
themarwhal: Yes keeping WORKSPACE manageable is something I'm struggling with a bit. 

I was thinking about maybe having a `bazel` directory with separate .bzl files for different groups of dependencies. I've seen some projects that do this but not sure what is the widely accepted way.

third_party/zlib.BUILD
themarwhal: Oh good job past Marie! Looks like I added it at the top of the file: `# Copied from https://github.com/protocolbuffers/protobuf/blob/master/third_party/zlib.BUILD`

themarwhal: I also got some other from here I think: https://github.com/bazelbuild/bazel-federation

third_party/nlohmann_json.BUILD
themarwhal: By package do you mean project wide visibility setting?

third_party/nlohmann_json.BUILD
themarwhal: Looks like I got it from here: https://github.com/horance-liu/nlohmann_json_demo 
Will research and see what it does and either remove or comment

orc8r/gateway/c/common/service303/BUILD.bazel
themarwhal: removed

third_party/civetweb.BUILD
themarwhal: Added at the top of the file

third_party/prometheus-cpp.BUILD
themarwhal: removed

third_party/prometheus-cpp.BUILD
themarwhal: removed - unnecessary copy pasta it seems

third_party/nlohmann_json.BUILD
themarwhal: forgot what package meant :D fixed it in the new version

WORKSPACE
```diff
@@ -112,3 +100,21 @@ load("@com_github_grpc_grpc//bazel:grpc_deps.bzl", "grpc_deps")
 
 grpc_deps()
 ### PROTO / GRPC DEPENDENCIES ###
+
+### Dependencies for orc8r/gateway/c/common
```
markjen, line 114: does this mean everyone will have to pull down these deps even if they are only working on components that do not need them (e.g. orc8r / feg)?

perhaps specifying these dependencies explicitly for build targets that need them would be better?

markjen (approved):
1. do we need copyright notices at the top of our build files? I don't know the recommended practice and answer for magma repo.
2. for places where we borrowed BUILD files, do we need to check the license and see if we need to reproduce the license or any other info in our repo? If nothing else, perhaps it'd be helpful to also store the git SHA we copied at and what the license is at the moment.

WORKSPACE
themarwhal: @markjen  I think these archives only get pulled when they are needed. So it something like prometheus-cpp would not get pulled if you only want to build feg. 
Will double check on this though.

themarwhal: From bazel doc: https://docs.bazel.build/versions/main/external.html#fetching-dependencies: `By default, external dependencies are fetched as needed during bazel build. `

orc8r/gateway/c/common/async_grpc/BUILD.bazel
themarwhal: Added a TODO to rename files

hcgatewood (approved):
Excited!

electronjoe (approved):
Wooohoo!

ardzoht (approved):
lgtm

orc8r/gateway/c/common/service_registry/BUILD.bazel
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "service_registry",
```
amarpad, line 15: I was hoping we would create only one library or at least fewer than 6 or howmany ever we have.

themarwhal: @amarpad  I think it's OK to keep the libraries separate just so that we can do partial builds easier. But I do have a task tracking the work to flatten the orc8r/gateway/c/common subdirectories. (https://github.com/magma/magma/issues/8453)
I'm deferring this work to after we commit to Bazel migration. I'm trying to not affect the current CMake build at all in this iteration. :)

PR #8327, opened 2021-07-29
fix(agw): fix perf regression with GTP echo

pshelar:
Following patch improves GTP performance by fixing the
flow mask and setting right interface state for BFD.

## Test Plan
validated with OVS unit test and perf test on lab setup.

## Additional Information


themarwhal (approved):
stamping - I wonder who would be a good approvers-* owner for `third_party/gtp_ovs` ?

pshelar:
> stamping - I wonder who would be a good approvers-* owner for `third_party/gtp_ovs` ?

Thanks.
lets set owner as Nick and me.

PR #8352, opened 2021-07-30
fix(feg): single protos import

alrs:
Signed-off-by: Lars Lehtonen <lars.lehtonen@gmail.com>

## Summary

This fixes `feg/gateway/services/swx_proxy/servicers`, where the same package was being imported twice.

## Test Plan

`go test .` passed.

## Additional Information


themarwhal (approved):

uri200 (approved):
Thanks for fixing this 🙏

PR #8355, opened 2021-07-30
Revert "feat(pipelined): enable NAT traffic for eNB traffic in Non-NAT setup"

ulaskozat:
Reverts magma/magma#8134


pshelar (approved):

ulaskozat:
Original PR seems to break the CWAG integ tests

PR #8324, opened 2021-07-28
fix(mme): Fix memory leak in TAU procedure on 1.6 branch

ulaskozat:
Signed-off-by: Ulas Kozat <kozat@fb.com>

## Summary

Backports only the TAU part of the memory leak fix in #8029. The other part was introduced and fixed on the same release cycle, hence it does have no effect on v1.6.

## Test Plan

integ tests.

## Additional Information

- [ ] This change is backwards-breaking

PR #8414, opened 2021-07-31
fix(agw): Fixed memory leaks on branch v1.6

rsarwad:
fix(agw): Fixed memory leaks on branch v1.6
## Summary
Inconsistently memory leak was observed during execution of test_attach_detach_dedicated_deactivation_timer_expiry.py
Cause for leak is, on first timer timer expiry, mme shall start the timer with some arguments, for which memory would be allocated.  So part of start, if timer has valid timer id then first timer is stopped and started again. As part of stop, it returns the memory allocated for timer arguments. 
Since timer has expired, timer id will not be found in the timer queue, so timer argument returned from stop was corrupted/incorrect. When start the timer again; it was not starting. Since it was not started, it was resulting into memory leak.
Memory allocated to timer argument would be freed, either on final timer expiry or while stopping using api, esm_ebr_stop_timer()
## Test Plan
Executed test_attach_detach_dedicated_deactivation_timer_expiry.py test case and stopped all services to check memory leaks.  This is repeated multiple times. 


ulaskozat (commented):
Can you rebase it as this https://github.com/magma/magma/pull/8324 has landed the TAU parts.

ulaskozat (approved):
LGTM

PR #8420, opened 2021-08-02
refactor(nms): Changed subscriber detail KPIs to display data usage as MB/s and changed histogram to line chart

ato-dani:
## Summary
The data usage KPI of the subscribers were changed from showing the total bytes to avg data usage in Mb/s. Also, the histogram showing the data usage was changed to a line chart.

## Test Plan
A new testfile named, SubscriberChartTest.js, file was created to test that the correct KPIs are rendered.

Screen Shot 2021-08-02 at 12 39 35 PM

## Additional Information

### DIFF
--- a/nms/app/packages/magmalte/app/views/subscriber/SubscriberChart.js
+++ b/nms/app/packages/magmalte/app/views/subscriber/SubscriberChart.js
@@ -21,7 +21,6 @@ import type {network_id, subscriber_id} from '@fbcnms/magma-api';
 import Card from '@material-ui/core/Card';
 import CardHeader from '@material-ui/core/CardHeader';
 import CardTitleRow from '../../components/layout/CardTitleRow';
-import CustomHistogram from '../../components/CustomMetrics';
 import Divider from '@material-ui/core/Divider';
 
 import DataGrid from '../../components/DataGrid';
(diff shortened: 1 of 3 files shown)


nms/app/packages/magmalte/app/views/subscriber/SubscriberChart.js
```diff
@@ -148,26 +147,39 @@ function SubscriberDataKPI() {
     // fetch queries
     const fetchAllData = async () => {
       const stepCategoryMap = {
-        '1h': 'Last 1 Hour',
-        '24h': 'Last 1 Day',
-        '7d': 'Last 1 Week',
-        '30d': 'Last 1 Month',
+        '1h': {
+          category: 'Current Usage MB/s',
```
andreilee:
```suggestion
          category: 'Hourly Usage MB/s',
```

andreilee (commented):
Why is reported usage being changed from `sum` to `avg`?

ato-dani:
> Why is reported usage being changed from `sum` to `avg`?
@andreilee Because previously it was reporting the total data usage which means we needed to use `sum` but now as per the task, we want to report the avg data usage which means we need to use `avg`

andreilee (approved):

PR #8421, opened 2021-08-02
refactor: add caching to expensive functions in ByteCount front-end

waqaraqeel:
## Summary

Some function calls when processing byte counters from the kernel can be expensive. `psutil.Process(cmdline).cmdline()` is one such example. This PR adds simple LRU caching to two such functions as suggested in [comments](https://github.com/magma/magma/pull/8251#discussion_r679609960) on #8251.

## Test Plan

Ran unit tests in `byte_counter_tests`. All pass.


pshelar (approved):
LGTM

PR #8422, opened 2021-08-02
feat: add IPv6 support to byte_count.bpf.c

waqaraqeel:

## Summary

This PR adds IPv6 support for byte counting IPv6 traffic. Traffic from local AGW services to the control proxy is IPv6 with embedded IPv4. This PR also handles this traffic to count bytes that local services send to the cloud through control proxy.

## Test Plan

Updated `byte_counter_tests.py` to reflect changes. All tests pass.

### DIFF
--- a/lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c
+++ b/lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c
@@ -1,3 +1,4 @@
+
 /**
  * Copyright 2020 The Magma Authors.
  *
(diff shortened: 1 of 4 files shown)


lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c
```diff
@@ -23,28 +24,48 @@ BPF_HASH(dest_counters, struct key_t);
...
+      if (key.daddr == 0x100007F || key.daddr == 0) {
+        return 0;
+      }
+    }
+    // IPv6 localhost has embedded IPv4 in case of control_proxy
+    // uint128 is split into two u64 and high is FFFF:"127.0.0.1"
+    else if (family == AF_INET6 && key.dport != {{PROXY_PORT}}) {
+      uint64_t low = (uint64_t)key.daddr;
+      uint64_t high = (key.daddr >> 64);
+      if (low == 0 && high == 0x100007FFFFF0000) {
```
pshelar, line 55: it would be easier/faster to define constant unsigned char array and use memcmp() here.

lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c
```diff
@@ -23,28 +24,48 @@ BPF_HASH(dest_counters, struct key_t);
...
+        return 0;
+      }
     }
 
-    // read binary name, pid, source and destination IP address and port
+    // read binary name, pid, and source port
     bpf_get_current_comm(key.comm, TASK_COMM_LEN);
```
pshelar, line 61: I missed it in last round, can we avoid reading the command name, since we can map the pid to service name?

pshelar (changes requested):
Thanks for adding this support. It looks pretty close to finish.

lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c
waqaraqeel: This comm is only used as backup when the captured PID is no longer valid (process exited). We also want to record arbitrary Linux binaries generating traffic. PID to service/binary name mapping will happen every `collect_interval` seconds. The minimum `collect_interval` right now is 10 seconds and can be configured to higher values. If we don't capture the comm here, we will miss short-lived, perhaps frequent processes generating significant traffic.

lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c
waqaraqeel: Looked into it for a while. This would be the required byte string to compare with:
```
"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xFF\xFF\x7F\x00\x00\x01"
```

However, `memcmp` isn't allowed in eBPF. So if you use it, the program will compile but fail to load as the verifier will reject it. `__builtin_memcmp`, which should unroll at compile time, does not work because of a [clang bug](https://bugs.llvm.org/show_bug.cgi?id=26218). Now that bounded loops are supported in eBPF, one could write a custom memcmp(), but at that point, the two-part comparison above seems better to me.

pshelar (approved):

PR #8432, opened 2021-08-03
docs: Clarify underspecified path to proposal acceptance

hcgatewood:
## Summary

Process was unspecified for proposals that didn't require TSC approval, leading to multiple proposals sitting in purgatory.

With this change, proposals require TSC approval, which will help make the process more explicit -- TSC is the point of serialization, but can defer to `approvers-*` teams or the like. This should drastically improve the explicitness of the process by, in a public way, bringing the proposer+reviewer together to walk domain non-experts (the TSC) through the rationale for acceptance

## Additional Information

### DIFF
--- a/docs/readmes/contributing/contribute_proposals.md
+++ b/docs/readmes/contributing/contribute_proposals.md
@@ -10,12 +10,15 @@ This document describes the process by which interested parties may propose chan
 
 To start, consider skimming the existing [Magma project proposals](https://github.com/magma/magma/issues?q=is%3Aissue+label%3A%22type%3A+proposal%22+).
 
-## Tracking proposals
+## Overview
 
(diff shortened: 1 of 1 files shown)


themarwhal (approved):
lgtm! thanks

themarwhal:
Should we mention the slack channel? (maybe it's not needed but could be useful)

hcgatewood:
> Should we mention the slack channel? (maybe it's not needed but could be useful)

Looks like both #proposals and #governance-tsc-ama are mentioned, lmk if you had an additional channel in mind

themarwhal:
> > Should we mention the slack channel? (maybe it's not needed but could be useful)
> 
> Looks like both #proposals and #governance-tsc-ama are mentioned, lmk if you had an additional channel in mind

Missed those! Sounds good then :D

PR #8452, opened 2021-08-03
fix: handle int128 as couple of int64 and make hashable

waqaraqeel:

## Summary

IPv6 support in `ByteCounter` had a bug where a `c_ulong_Array_2` was being cast to tuple and then bytes in `_ip_addr_to_str`. That is incorrect as `inet_ntop` expects a `c_ulong_Array_2` object in case of `AF_INET6`. This PR fixes the bug by replace cast to bytes with cast to `c_ulong_Array_2`. The conversion to tuple is necessary for caching as it requires a hashable type.

Alternative would be to define `__hash__` and `__eq__` for the fixed-size array type. I think that is more clumsy.

## Test Plan

Unit tests did not catch the original bug as `key.daddr` was being mocked as `int` and `bytes` for IPv4 and IPv6 respectively. Modified the unit test to use the `c_ulong_Array_2` type. Tests pass.

### DIFF
--- a/lte/gateway/python/magma/kernsnoopd/handlers.py
+++ b/lte/gateway/python/magma/kernsnoopd/handlers.py
@@ -13,6 +13,7 @@ limitations under the License.
 
 import abc
 import logging
+from ctypes import c_ulong
 from functools import lru_cache
 from socket import AF_INET, AF_INET6, inet_ntop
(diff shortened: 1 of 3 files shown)


ardzoht (approved):
lgtm

PR #8456, opened 2021-08-03
clean(agw): use a newer message() method to get error messages

themarwhal:

## Additional Information


themarwhal:
doesn't work welp

PR #8471, opened 2021-08-04
feat(mme): amf client servicer layer

panyogesh:
Fixes:
  1. Removed the proto from AMF cmake as thats handled in oai/lib
  2. Introduced amf client servicer which will abstratct amf's client
     interaction like subscriberdb
  3. Introduced the Mock layer in amf unit test for the client testing
  4. Fixed the CMake for unit testing of amf

Testing:
  1. Basis unit testing
  2. Basic snaity with UERANSIM

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/amf/amf_app_main.cpp
+++ b/lte/gateway/c/core/oai/tasks/amf/amf_app_main.cpp
@@ -32,6 +32,7 @@ extern "C" {
 #include "amf_data.h"
 #include "amf_app_defs.h"
 #include "amf_authentication.h"
+#include "include/amf_client_servicer.h"
 #include "ngap_messages_types.h"
 #include "amf_app_state_manager.h"
 #include "amf_smfDefs.h"
@@ -182,6 +183,10 @@ extern "C" int amf_app_init(const amf_config_t* amf_config_p) {
   }
   /*Initialize UE state matrix */
   create_state_matrix();
+
+  // Initialize the client layer
+  amf_client_servicer_init();
+
   if (itti_create_task(TASK_AMF_APP, &amf_app_thread, NULL) < 0) {
     OAILOG_CRITICAL(LOG_AMF_APP, "Amf app create task failed\n");
     OAILOG_FUNC_RETURN(LOG_AMF_APP, RETURNerror);
--- a/lte/gateway/c/core/oai/tasks/amf/amf_app_main.cpp
+++ b/lte/gateway/c/core/oai/tasks/amf/amf_app_main.cpp
@@ -156,14 +156,19 @@ static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
  ***************************************************************************/
 void* amf_app_thread(void* args) {
   itti_mark_task_ready(TASK_AMF_APP);
-  const task_id_t tasks[]      = {TASK_NGAP, TASK_SERVICE303};
-  amf_metadata_t* amf_metadata = reinterpret_cast<amf_metadata_t*>(args);
+  const task_id_t tasks[] = {TASK_NGAP, TASK_SERVICE303};
+  std::shared_ptr<amf_metadata_t> amf_metadata =
+      std::make_shared<amf_metadata_t>();
+
+  // initialize the metadata and client service entries
+  amf_metadata_initialize(amf_metadata);
 
   init_task_context(
       TASK_AMF_APP, tasks, 2, handle_message, &amf_app_task_zmq_ctx);
   // Service started, but not healthy yet
   send_app_health_to_service303(&amf_app_task_zmq_ctx, TASK_AMF_APP, false);
   zloop_start(amf_app_task_zmq_ctx.event_loop);
+
   amf_app_exit();
   return NULL;
 }
@@ -186,10 +191,7 @@ extern "C" int amf_app_init(const amf_config_t* amf_config_p) {
   /*Initialize UE state matrix */
   create_state_matrix();
 
-  // initialize the metadata and client service entries
-  amf_metadata_intialize(&(amf_metadata));
-
-  if (itti_create_task(TASK_AMF_APP, &amf_app_thread, &(amf_metadata)) < 0) {
+  if (itti_create_task(TASK_AMF_APP, &amf_app_thread, NULL) < 0) {
     OAILOG_CRITICAL(LOG_AMF_APP, "Amf app create task failed\n");
     OAILOG_FUNC_RETURN(LOG_AMF_APP, RETURNerror);
   }
--- a/lte/gateway/c/core/oai/tasks/amf/amf_app_main.cpp
+++ b/lte/gateway/c/core/oai/tasks/amf/amf_app_main.cpp
@@ -25,6 +25,7 @@ extern "C" {
 #ifdef __cplusplus
 }
 #endif
+#include "include/amf_metadata_util.h"
 #include "amf_app_messages_types.h"
 #include "amf_config.h"
 #include "amf_fsm.h"
@@ -32,7 +33,6 @@ extern "C" {
 #include "amf_data.h"
 #include "amf_app_defs.h"
 #include "amf_authentication.h"
-#include "include/amf_client_servicer.h"
 #include "ngap_messages_types.h"
 #include "amf_app_state_manager.h"
 #include "amf_smfDefs.h"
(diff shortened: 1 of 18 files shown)


lte/gateway/c/core/oai/tasks/amf/include/amf_client_servicer.h
```diff
@@ -0,0 +1,57 @@
...
+  bool get_subscriber_authentication_info(
+      const std::string& imsi, uint8_t imsi_length, const char* snni,
+      amf_ue_ngap_id_t ue_id);
+
+  bool get_subscriber_authentication_info_resync(
+      const std::string& imsi, uint8_t imsi_length, const char* snni,
+      const void* resync_info, uint8_t resync_info_len, amf_ue_ngap_id_t ue_id);
+};
+
+// Initilaze the client servicer layer
```
ulaskozat: typo `Initilaze`

lte/gateway/c/core/oai/tasks/amf/amf_client_servicer.cpp
```diff
@@ -0,0 +1,57 @@
...
+AmfClientServicer& get_amf_client_server_ref() {
+  return amf_client_servicer_g;
+}
+
+// Initialize the client servicer layer
+void amf_client_servicer_init() {
+  auto authentication_client =
+      std::make_shared<AsyncM5GAuthenticationServiceClient>();
+
+  amf_client_servicer_g = AmfClientServicer(authentication_client);
```
ulaskozat: what is the significance of `_g`? is it to signify that this is grpc client? it does not help much to readability.

ulaskozat: ah this is to indicate that it is global?

lte/gateway/c/core/oai/tasks/amf/include/amf_client_servicer.h
panyogesh: Comments addressed.

lte/gateway/c/core/oai/tasks/amf/amf_client_servicer.cpp
panyogesh: Yes..intention to denote it as global...This layer will serve as interface to all clients. After the merge it will be extended t o serve mobilityd and sessiond clients.

lte/gateway/c/core/oai/tasks/amf/amf_client_servicer.cpp
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "include/amf_client_servicer.h"
+#include <memory>
+
+namespace magma5g {
+
+AmfClientServicer amf_client_servicer_g;
```
ulaskozat: The overall pattern looks more complex than it needs to be. You are wrapping a singleton class with another class and then instantiating a global instance. Instead of this, it makes more sense to convert singleton class into a non singleton one and define a global. Then you can write you mock class directly for that.

ulaskozat (changes requested):

ulaskozat:
@panyogesh reading your new comment and I understand that you want this class as an entry class for all the grpc clients. In this case, do we still need to define actual GRPC clients as singletons?

ulaskozat:
@panyogesh I think the better way is to avoid the global definition here completely. Let's create a struct that is a holder of each of the globals here (i.e., it is a meta class) and pass the pointer to that into `argp` parameter (currently passing as NULL) in `itti_create_task`. 

We can then mock each of the clients individually for mobilityd, sessiond, etc.

panyogesh:
> @panyogesh I think the better way is to avoid the global definition here completely. Let's create a struct that is a holder of each of the globals here (i.e., it is a meta class) and pass the pointer to that into `argp` parameter (currently passing as NULL) in `itti_create_task`.
> 
> We can then mock each of the clients individually for mobilityd, sessiond, etc.

1. Created the global struct : amf_metadata
2. Initialized during startup before launching the amf thread and it stores the information about clients (in this patch it stores authenticaiton client)
3. Its then passed on to the amf_thread.

lte/gateway/c/core/oai/tasks/amf/amf_app_main.cpp
```diff
@@ -155,7 +156,9 @@ static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
  ***************************************************************************/
 void* amf_app_thread(void* args) {
   itti_mark_task_ready(TASK_AMF_APP);
-  const task_id_t tasks[] = {TASK_NGAP, TASK_SERVICE303};
+  const task_id_t tasks[]      = {TASK_NGAP, TASK_SERVICE303};
+  amf_metadata_t* amf_metadata = (amf_metadata_t*) args;
```
ulaskozat: please use `reinterpret_cast`.

lte/gateway/c/core/oai/tasks/amf/amf_app_main.cpp
```diff
@@ -182,7 +185,11 @@ extern "C" int amf_app_init(const amf_config_t* amf_config_p) {
   }
   /*Initialize UE state matrix */
   create_state_matrix();
-  if (itti_create_task(TASK_AMF_APP, &amf_app_thread, NULL) < 0) {
+
+  // Initialze the metadata and client service entries
```
ulaskozat: Typo `Initialze`

lte/gateway/c/core/oai/tasks/amf/amf_app_main.cpp
panyogesh: Comments addressed.

lte/gateway/c/core/oai/tasks/amf/include/amf_metadata_util.h
```diff
@@ -0,0 +1,44 @@
...
+        std::make_shared<magma5g::AmfClientServicer>(authentication_client);
+  }
+
+  std::shared_ptr<magma5g::AmfClientServicer>& get_client_servicer_ref() {
+    return amf_client_servicer_;
+  }
+
+} amf_metadata_t;
+
+void amf_metadata_intialize(amf_metadata_t* metadata_p);
```
ardzoht:
```suggestion
void amf_metadata_initialize(amf_metadata_t* metadata_p);
```

lte/gateway/c/core/oai/tasks/amf/amf_app_main.cpp
panyogesh: Comments addressed.

lte/gateway/c/core/oai/tasks/amf/include/amf_metadata_util.h
```diff
@@ -0,0 +1,44 @@
...
+
+/*
+ * This strucuture is used in itti_create_task
+ * Container of amf structure/class at the task level.
+ */
+extern amf_metadata_t amf_metadata;
```
ardzoht: What's the constraint of this being exposed as global? I think from the AMF app's thread scope this should be dynamically allocated (on the thread initialization procedure) This way this global definition can be avoided. Also let's make sure if that's being allocated the memory is being freed when the task context is destroyed on the thread exit function. cc @ulaskozat

ardzoht (changes requested):

lte/gateway/c/core/oai/tasks/amf/include/amf_metadata_util.h
ulaskozat: +1

panyogesh: Comments addressed.

lte/gateway/c/core/oai/tasks/amf/include/amf_metadata_util.h
panyogesh: Comments addressed.

ardzoht (approved):
lgtm

lte/gateway/c/core/oai/tasks/amf/amf_metadata_util.cpp
```diff
@@ -0,0 +1,30 @@
...
+void amf_metadata_initialize(
+    const std::shared_ptr<amf_metadata_t>& metadata_p) {
+  metadata_info_ptr = std::move(metadata_p);
+
+  metadata_info_ptr->amf_client_servicer_init();
+}
+
+std::shared_ptr<magma5g::AmfClientServicer> amf_get_client_servicer_ref() {
+  return metadata_info_ptr->amf_client_servicer;
+}
```
ulaskozat, line 30: The original comment was about to eliminate globals. The pattern should be like the following:
1. Create meta data object during task initialization.
2. Have new function `init_task_context_with_args`  (in the same file as `init_task_context`) and pass the meta data  pointer as the arguments for handler function. Inside `init_task_context_with_args`, pass the argument to zloop_reader.
3. In handler function, pass the pointer for metadata to called functions if they need to access the meta data.

cc @ardzoht

ardzoht (changes requested):
Let's remove use of global shared ptr and update this per @ulaskozat' comment

lte/gateway/c/core/oai/tasks/amf/amf_metadata_util.cpp
panyogesh: Tried making the change. Metadata parameter will be used by function : start_authentication_information_procedure().  This function is coming at a stack depth of 16. So the changes are spanning across many files/functions.  Need to discuss on the same.

panyogesh:
Closing as the comments as its addressed by #9618

PR #8478, opened 2021-08-04
fix(agw): handle missing error code on service failure

pshelar:
Fixes following exception:
```
magma-dev-focal pipelined[2619]: Traceback (most recent call last):
magma-dev-focal pipelined[2619]:   File "/usr/local/bin/service_util.py", line 80, in <module>
magma-dev-focal pipelined[2619]:     update_stats(service_name)
magma-dev-focal pipelined[2619]:   File "/usr/local/bin/service_util.py", line 69, in update_stats
magma-dev-focal pipelined[2619]:     status = get_status()
magma-dev-focal pipelined[2619]:   File "/usr/local/bin/service_util.py", line 42, in get_status
magma-dev-focal pipelined[2619]:     ServiceExitStatus.ExitCode.Value(exit_code.upper())
magma-dev-focal pipelined[2619]: AttributeError: 'NoneType' object has no attribute 'upper'
```

## Additional Information

### DIFF
--- a/orc8r/gateway/python/scripts/service_util.py
+++ b/orc8r/gateway/python/scripts/service_util.py
@@ -28,7 +28,7 @@ def get_status() -> ServiceExitStatus:
     @returns a populated service exit status object
     """
     service_result = os.environ.get("SERVICE_RESULT")
-    exit_code = os.environ.get("EXIT_CODE")
+    exit_code = os.environ.get("EXIT_CODE", "EXITED")
     exit_status = os.environ.get("EXIT_STATUS")
 
     # Populate the service exit status string and exit code.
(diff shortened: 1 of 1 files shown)


andreilee (changes requested):
I think we need a root cause here of why `EXIT_CODE` env variable isn't populated - it seems like an underlying assumption of this code, rather than a small oversight. If we just use a default of `EXITED`, then we might end up missing service crashes.

pshelar:
> I think we need a root cause here of why `EXIT_CODE` env variable isn't populated - it seems like an underlying assumption of this code, rather than a small oversight. If we just use a default of `EXITED`, then we might end up missing service crashes.

'existed' is code I see in service status in such cases.

andreilee (approved):
Approving this on the basis that this will mitigate crashes, and that we are committing to following up with discovering and fixing the root cause of this issue.

PR #8455, opened 2021-08-03
fix(mme): assert on single thread exits

ulaskozat:
Signed-off-by: Ulas Kozat <kozat@fb.com>

## Summary

MME service is vulnerable to the cases where a single thread exits leaving the rest of the threads keep running. This PR asserts on such events forcing the process to exit.

## Test Plan

S1AP integ tests.

## Additional Information

- [ ] This change is backwards-breaking

PR #8480, opened 2021-08-04
fix(mme): assert on single thread exits (#8455)

ulaskozat:
* fix(mme): assert on single thread exits

## Summary

Backporting #8455 to eliminate individual thread exits.

## Test Plan

S1AP integ tests.

## Additional Information


ardzoht (approved):
lgtm

PR #8482, opened 2021-08-04
feat(agw): Inject state sync interval

### LINKED ISSUE #8384: app: state: Increase default state sync interval, make interval Orc8r-controlled

mstre123:
Signed-off-by: Eli Sutton <elisutton@fb.com>

## Summary
Make state service sync interval configurable from orchestrator
Issue #8482 
## Test Plan
Unit Tests
Local testing


lte/gateway/configs/gateway.mconfig
```diff
@@ -84,6 +84,11 @@
       "lteAuthAmf": "gAA=",
       "sync_interval": 300
     },
+    "state":{
+      "@type": "type.googleapis.com/magma.mconfig.State",
+      "logLevel": "INFO",
+      "sync_interval": 43
```
mstre123: temp using 43 to make it distinct and see what was being used downstream

orc8r/cloud/go/services/orchestrator/obsidian/handlers/network_handlers_test.go
```diff
@@ -133,12 +133,14 @@ func Test_GetNetworkHandlers(t *testing.T) {
...
 	tests.RunUnitTest(t, e, tc)
 
-	// add Sentry config
+	// add Sentry and state config
 	sentryConfig := models.NewDefaultSentryConfig()
+	stateConfig := models.NewDefaultStateConfig()
 	update1 = configurator.NetworkUpdateCriteria{
 		ID:                   "n1",
-		ConfigsToAddOrUpdate: map[string]interface{}{orc8r.NetworkSentryConfig: sentryConfig},
+		ConfigsToAddOrUpdate: map[string]interface{}{orc8r.NetworkSentryConfig: sentryConfig, orc8r.NetworkStateConfig: stateConfig},
```
hcgatewood: Nit: move these values to their own lines

orc8r/cloud/go/services/orchestrator/obsidian/handlers/network_handlers_test.go
```diff
@@ -511,6 +516,8 @@ func Test_PutNetworkHandlers(t *testing.T) {
 	network1.Features.Features["new-feature"] = "foobar"
 	network1.SentryConfig.SampleRate = swag.Float32(0.75)
 	network1.SentryConfig.UploadMmeLog = true
+	syncInterval := uint32(30)
+	network1.StateConfig.SyncInterval = &syncInterval
```
hcgatewood: Use the swag.Uint32 pattern (see above)

orc8r/cloud/go/services/orchestrator/obsidian/models/defaults.go
```diff
@@ -44,6 +44,13 @@ func NewDefaultSentryConfig() *NetworkSentryConfig {
 	}
 }
 
+func NewDefaultStateConfig() *NetworkStateConfig {
+	syncInterval := uint32(60)
```
hcgatewood: Part of the task includes actually increasing this default value

Also nit: here and elsewhere, use the swag.uint32 pattern I mention above

orc8r/cloud/go/services/orchestrator/obsidian/models/swagger.v1.yml
```diff
@@ -1410,6 +1412,17 @@ definitions:
...
 
+  network_state_config:
+    type: object
+    description: Network state configuration
+    properties:
+      sync_interval:
+        description: sync interval in seconds
+        type: number
+        format: uint32
+        minimum: 0
```
hcgatewood: Actual minimum should be 1m -- anything less than that could cause agw to get into weird states

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -191,3 +198,25 @@ func getVpnMconfig(gatewayConfig *models.MagmadGatewayConfigs) *mconfig_protos.O
 
 	return ret
 }
+
+func getStateMconfig(net configurator.Network, gwKey string) *mconfig_protos.State {
+	ret := &mconfig_protos.State{
```
hcgatewood: Nit: I don't like using "ret" as a variable, since it obfuscates the contents of the variable / its affordances. Just name it like you would any other variable

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -191,3 +198,25 @@ func getVpnMconfig(gatewayConfig *models.MagmadGatewayConfigs) *mconfig_protos.O
...
+	}
+	return ret
+}
+
+// getRandomizedSyncInterval returns the interval received from getSyncInterval
+// as seconds and increases it by a random jitter in the range of
+// [0, 0.2 * getSyncInterval()]. Increased sync interval ameliorates the thundering
+// herd effect at the Orc8r.
+func getRandomizedSyncInterval(gwKey string, syncInterval uint32) uint32 {
+	jitter := math.JitterUint32(syncInterval, gwKey, 0.2)
```
hcgatewood: How this fn works was recently changed on master. It now returns `n + jitter` instead of just `jitter`

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
```diff
@@ -76,6 +81,10 @@ func TestBaseOrchestratorMconfigBuilder_Build(t *testing.T) {
 				LogLevel:       protos.LogLevel_INFO,
 				EventVerbosity: -1,
 			},
+			"state": &mconfig_protos.State{
+				//expected value differs from map due to added jitter
+				SyncInterval: 68,
```
hcgatewood: Instead of adding a comment to a magic number, make the number a variable with descriptive name. If the comment is still necessary, add it as a doc comment on the variable

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
hcgatewood: Also let's make the value .25, such that default of n=3 minutes means the jitter is up to 45s

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
```diff
@@ -311,12 +340,162 @@ func TestBaseOrchestratorMconfigBuilder_Build(t *testing.T) {
...
+			},
+		}
+
+		actual, err := buildBaseOrchestrator(&nw, &graph, "gw1")
+		assert.NoError(t, err)
+		assert.Equal(t, expected, actual)
+	})
+	// No state config added to Network Configs
+	t.Run("no state config", func(t *testing.T) {
```
hcgatewood: No need to add a new test for this -- just don't include the state config in one of the earlier tests

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
```diff
@@ -311,12 +340,162 @@ func TestBaseOrchestratorMconfigBuilder_Build(t *testing.T) {
...
+				SyncInterval: 68,
+			},
 		}
 
 		actual, err := buildBaseOrchestrator(&nw, &graph, "gw1")
 		assert.NoError(t, err)
 		assert.Equal(t, expected, actual)
 	})
+
+	// test sync interval jitter
```
hcgatewood: Nit: sentence-case your comments

hcgatewood: How is this test different from the previous tests, with regard to testing sync interval jitter? The network id actually needs to be different (?)

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
```diff
@@ -350,6 +529,7 @@ func buildBaseOrchestrator(network *configurator.Network, graph *configurator.En
 		ret["td-agent-bit"] = configs["td-agent-bit"]
 		ret["eventd"] = configs["eventd"]
 	}
+	ret["state"] = configs["state"]
```
hcgatewood: Move this into the initial map defn -- line 522+. No need to have it separate down here

hcgatewood (changes requested):
Awesome. Fix nits, then we can move toward merging once you're able to verify e2e on dev machine 👍

orc8r/cloud/go/services/orchestrator/obsidian/models/defaults.go
mstre123: bumped it to 5 mins

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
mstre123: kk was following the convention in the other helpers in this class but can rename.

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
mstre123: the jitter is based on the gateway id not the network id

hcgatewood: Ah of course. In any case, ensure we have tests that cover different gateway IDs such that they cause different jittered values

mstre123: they do - to make it clear at the top of the test i have 2 diff expected sync intervals based on different gateway ids

hcgatewood: 👍💯

orc8r/cloud/go/services/orchestrator/obsidian/handlers/network_handlers_test.go
```diff
@@ -133,12 +133,16 @@ func Test_GetNetworkHandlers(t *testing.T) {
 	}
 	tests.RunUnitTest(t, e, tc)
 
-	// add Sentry config
+	// add Sentry and state config
```
hcgatewood: Nit: since you're changing this comment anyway, convert it to sentence-case

https://magma.github.io/magma/docs/next/contributing/contribute_conventions#leave-it-better-than-you-found-it

orc8r/cloud/go/services/orchestrator/obsidian/handlers/network_handlers_test.go
```diff
@@ -133,12 +133,16 @@ func Test_GetNetworkHandlers(t *testing.T) {
...
 	sentryConfig := models.NewDefaultSentryConfig()
+	stateConfig := models.NewDefaultStateConfig()
 	update1 = configurator.NetworkUpdateCriteria{
-		ID:                   "n1",
-		ConfigsToAddOrUpdate: map[string]interface{}{orc8r.NetworkSentryConfig: sentryConfig},
+		ID: "n1",
+		ConfigsToAddOrUpdate: map[string]interface{}{
+			orc8r.NetworkSentryConfig: sentryConfig,
+			orc8r.NetworkStateConfig:  stateConfig},
```
hcgatewood: If parens etc start on their own line, they should end on their own line

orc8r/cloud/go/services/orchestrator/obsidian/models/swagger.v1.yml
```diff
@@ -1426,6 +1428,17 @@ definitions:
         maximum: 1
         example: 0.5
 
+  network_state_config:
```
hcgatewood: For now let's just call this `state_config` -- cause it's likely this will be added to the agw-level as well at some point, and there's nothing specific to this that makes it "network-wide"

And fix description etc

orc8r/cloud/go/services/orchestrator/obsidian/models/swagger.v1.yml
```diff
@@ -1426,6 +1428,17 @@ definitions:
...
+    type: object
+    description: Network state configuration
+    properties:
+      sync_interval:
+        description: sync interval in seconds
```
hcgatewood: Sentence-case

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -191,3 +198,24 @@ func getVpnMconfig(gatewayConfig *models.MagmadGatewayConfigs) *mconfig_protos.O
 
 	return ret
 }
+
+func getStateMconfig(net configurator.Network, gwKey string) *mconfig_protos.State {
+	stateMconfigProto := &mconfig_protos.State{
```
hcgatewood: While we're here, let's go ahead and set the log level I guess. Set it to info for now. No need to consume it at the gateway though

hcgatewood: General nit: I prefer concise variable names, so e.g. the `state` in `stateMconfigProto` is superfluous because it's implicit given the context of the fn (`getStateMconfig`), so just name this one mconfigProto and netStateConfig->netConfig, e.g.

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -191,3 +198,24 @@ func getVpnMconfig(gatewayConfig *models.MagmadGatewayConfigs) *mconfig_protos.O
 
 	return ret
 }
+
+func getStateMconfig(net configurator.Network, gwKey string) *mconfig_protos.State {
```
hcgatewood, line 202: I would structure this fn like this:

1. Try get state mconfig from netconfig, else use default state mconfig
2. Augment the chosen mconfig value with jittered sync interval

This will (a) allow us to choose an "entire" mconfig as a pre-step, and will also remove the need for `getRandomizedSyncInterval` defn since can just call JitterUint directly

orc8r/cloud/go/services/orchestrator/obsidian/handlers/network_handlers_test.go
```diff
@@ -133,12 +133,16 @@ func Test_GetNetworkHandlers(t *testing.T) {
 	}
 	tests.RunUnitTest(t, e, tc)
```
hcgatewood, line 137: I don't see any unit tests for the partial handler you added (?). Either remove the partial handler or, preferably, add a short sanity-check unit test

hcgatewood (changes requested):
Looks about there! Main thing is the unit tests for the partial handler. Then let's get it merged

1. Reach out to the assigned codeowners for any required reviews
2. Reach out to @themarwhal (for a final Orc8r approval and) to merge your PR once everything's green

hcgatewood:
P.S. here's the info for e.g. the failing insync-checkin https://magma.github.io/magma/docs/next/orc8r/dev_testing

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -191,3 +198,17 @@ func getVpnMconfig(gatewayConfig *models.MagmadGatewayConfigs) *mconfig_protos.O
...
+		SyncInterval: 60,
+		LogLevel:     protos.LogLevel_INFO,
+	}
+	netConfig := net.Configs["state_config"]
+	if netConfig != nil {
+		nsConfig := netConfig.(*models.StateConfig)
+		syncInterval := nsConfig.SyncInterval
+		mconfigProto.SyncInterval = math.JitterUint32(syncInterval, gwKey, 0.25)
+	}
+	return mconfigProto
```
hcgatewood, line 214: Not quite right -- we want to jitter the value regardless of whether we use the default proto or augment it from the netconfig

Also add a regression test for this affordance

orc8r/cloud/go/services/orchestrator/obsidian/handlers/network_handlers_test.go
```diff
@@ -214,6 +221,19 @@ func Test_GetNetworkHandlers(t *testing.T) {
 		ExpectedResult: tests.JSONMarshaler([]string{"n1", networkID2}),
 	}
 	tests.RunUnitTest(t, e, tc)
+
+	// get network state
+	tc = tests.Test{
+		Method:         "GET",
+		URL:            fmt.Sprintf("%s/%s/", testURLRoot, "n1"),
```
hcgatewood: URL seems wrong

orc8r/cloud/go/services/orchestrator/obsidian/handlers/network_handlers_test.go
```diff
@@ -214,6 +221,19 @@ func Test_GetNetworkHandlers(t *testing.T) {
 		ExpectedResult: tests.JSONMarshaler([]string{"n1", networkID2}),
 	}
 	tests.RunUnitTest(t, e, tc)
+
+	// get network state
```
hcgatewood: Sentence-case your comments -- just do this as a default please 🙂

hcgatewood (changes requested):
Need to unconditionally jitter the sync interval

hcgatewood (approved):
Lgtm! Fix insync-checkin etc then msg me to merge

cwf/gateway/integ_tests/gateway.mconfig
```diff
@@ -128,6 +128,7 @@
   "state": {
    "@type": "type.googleapis.com/magma.mconfig.State",
    "logLevel": "INFO"
+   "sync_interval": 60
```
ardzoht, line 131: missing trailing comma on previous key

cwf/gateway/integ_tests/gateway.mconfig.idle_timer
```diff
@@ -122,6 +122,7 @@
   "state": {
    "@type": "type.googleapis.com/magma.mconfig.State",
    "logLevel": "INFO"
+   "sync_interval": 60
```
ardzoht, line 125: same here, add trailing comma after logLevel

orc8r/gateway/python/magma/state/state_replicator.py
```diff
@@ -55,9 +53,7 @@ def __init__(
         garbage_collector: GarbageCollector,
         grpc_client_manager: GRPCClientManager,
     ):
-        sync_interval = service.config.get(
-            'sync_interval', DEFAULT_SYNC_INTERVAL,
-        )
+        sync_interval = service.mconfig.sync_interval
```
ardzoht, line 57: minor nit: can you add a comment here specifying this is on seconds?

ardzoht (changes requested):

cwf/gateway/integ_tests/gateway.mconfig
```diff
@@ -128,6 +128,7 @@
   "state": {
    "@type": "type.googleapis.com/magma.mconfig.State",
    "logLevel": "INFO"
```
ardzoht: comma should be after logLevel, similar as your change on `lte/gateway/configs/gateway.mconfig`

cwf/gateway/integ_tests/gateway.mconfig.idle_timer
```diff
@@ -122,6 +122,7 @@
   "state": {
    "@type": "type.googleapis.com/magma.mconfig.State",
    "logLevel": "INFO"
```
ardzoht: same here

ardzoht (approved):
lgtm, just fix the comma on the cwf mconfigs

uri200 (approved):

cwf/gateway/integ_tests/gateway.mconfig
uri200: this is marked as resolved but I dont see the comma after "INFO"

uri200 (approved):
Note in teravm we have to use 4s. The reason is some process require the information to be ready as soon as possible in the orc8r so FEG can find it. Otherwise we have issues like we can't do a RAR right after attach (because the information takes 60s to be on the cloud)

So it would be good to either consider reduce it to 4s or change how state sync works (sync should be immediate when change or addition happened)

lte/gateway/configs/state.yml
```diff
@@ -13,8 +13,6 @@
 
 # log_level is set in mconfig. it can be overridden here
 
-sync_interval: 60
```
uri200, line 16: maybe we can comment that line so it can be use it if the client wants to (like line 14). Otherwise we will forget this setting even exists.

orc8r/cloud/go/services/orchestrator/obsidian/models/swagger.v1.yml
```diff
@@ -1426,6 +1458,17 @@ definitions:
...
 
+  state_config:
+    type: object
+    description: State configuration
+    properties:
+      sync_interval:
+        description: Sync interval in seconds
+        type: number
+        format: uint32
+        minimum: 60
```
uri200: thi is not going to work in real network. We need to make this at least 4 or we need to make the information available as son as the UE attaches.

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -72,6 +73,11 @@ func (b *baseOrchestratorBuilder) Build(network *storage.Network, graph *storage
 		return nil, err
 	}
 
+	net, err := (configurator.Network{}).FromProto(network, serdes.Network)
+	if err != nil {
+		return nil, err
```
uri200: I would recommend to wrap the error with some message (like line 84) so we know it comes from the builder.

uri200 (changes requested):

cwf/gateway/integ_tests/gateway.mconfig
mstre123: oops the diff didn't get pushed up - did it now

lte/gateway/configs/state.yml
mstre123: commenting this out and leaving it in the code seems a bit messy.
another approach would be a config on the gateway that turns on and off accepting a sync_interval from orc8r but maybe that's overkill.

EDIT: Spoke with. @uri200 offline - clarified that he was looking for an additional comment - added

orc8r/cloud/go/services/orchestrator/obsidian/models/swagger.v1.yml
mstre123: upped the minimum to 5 minutes - will need to do this for all the other defaults though.

uri200:
Unfortunately after talking to some of the providers, we can't increase this value or some requests on federated case will stop working (RARs won;t work and Bearer Request on inbound roaming for example). We need to make it 4s to make this work until we find other solution.

mstre123 (commented):
> Unfortunately after talking to some of the providers, we can't increase this value or some requests on federated case will stop working (RARs won;t work and Bearer Request on inbound roaming for example). We need to make it 4s to make this work until we find other solution.

4 seconds? @hcgatewood i think we chatted offline about this needing to be more that 60s or we would overload the service with requests.

uri200:
Right now we don't have much of an option: either we set it to a small value to provide some kind of real time, or Core initiated requests done before 60 will fail.

My suggestion is lower it for now not to block current federated deployments and plan a change on how state service works.

Overwhelming orc8r will not have much of an impact until we have many AGW talking to the cloud. But request coming from core is something currently requiered.

orc8r/cloud/go/services/orchestrator/obsidian/models/swagger.v1.yml
```diff
@@ -1426,6 +1458,17 @@ definitions:
...
 
+  state_config:
+    type: object
+    description: State configuration
+    properties:
+      sync_interval:
+        description: Sync interval in seconds
+        type: number
+        format: uint32
+        minimum: 300
```
emakeev: Isn't 5 minute state reporting interval a bit too high?

emakeev:
It looks like we are trying to address the state design/implementation shortcomings by drastically increasing the sync interval. Which in turn defeats the purpose of the cloud based state persistence. Maybe we should leave the minimum/default time at lower setting until we find a viable long term solution & meanwhile rely on ability to configure (increase) it to address the service overload issues.

mstre123:
> Right now we don't have much of an option: either we set it to a small value to provide some kind of real time, or Core initiated requests done before 60 will fail.
> 
> My suggestion is lower it for now not to block current federated deployments and plan a change on how state service works.
> 
> Overwhelming orc8r will not have much of an impact until we have many AGW talking to the cloud. But request coming from core is something currently requiered.

Wasn't the original default 60 seconds? 
https://github.com/magma/magma/pull/8482/files#diff-324ea256c5aab8aca44483e5f455fa248f80a5e262e6c4f9a5c3055d6a781008L16
why would we need to lower that to 4 seconds?

just parsing the above is the concern about a restart of the state service triggered by a config change?

orc8r/cloud/go/services/orchestrator/obsidian/models/swagger.v1.yml
mstre123: dropped the min to 3 seconds.

mstre123:
Ok to prevent breaking existing use cases and to try and move us towards a better place i split the baby.
Service configs for the sync interval will be honored if there under 10 seconds and less than the passed in interval from orc8r.
This is a temporary fix and i added a todo to remove this when we feel comfortable that no partners are relying on the service config.

uri200 (approved):

orc8r/gateway/python/magma/state/state_replicator.py
```diff
@@ -55,9 +53,18 @@ def __init__(
         garbage_collector: GarbageCollector,
         grpc_client_manager: GRPCClientManager,
     ):
-        sync_interval = service.config.get(
-            'sync_interval', DEFAULT_SYNC_INTERVAL,
-        )
+        # Sync_interval is in seconds
+        sync_interval = service.mconfig.sync_interval
+        service_config_sync_interval = service.config.get('sync_interval')
+        # TODO remove this once we confirm partners no longer set sync_interval service configs.
```
hcgatewood: Always give future readers a recourse with the todo. So either attach an issue number or put your github handle. In this case, go ahead and create a github issue, then do `TODO(#xyz): remove ...`

uri200:
note we created an issue defining current problems with how we synchronize state - directoryd

https://github.com/magma/magma/issues/8925

PR #8483, opened 2021-08-04
feat(orc8r): EKS Fargate via Cloudformation

arunuke:

## Summary

Experimental feature to deploy alternate compute backends to compare against EC2 for price and performance.

## Test Plan

aws cloudformation create-stack --stack-name <Stack Name> --template-body file://./cfnMagmaEksFargate.json --profile _Profile Name_ --region _Region_ --capabilities CAPABILITY_NAMED_IAM

The aforesaid command will deploy an EKS cluster backed by a Fargate compute backend.

### DIFF
--- a/experimental/cloudstrapper/playbooks/roles/cfn/cfnMagmaEksFargate.json
+++ b/experimental/cloudstrapper/playbooks/roles/cfn/cfnMagmaEksFargate.json
@@ -1,0 +1,346 @@
+{
+  "Parameters": {
+    "paramEksClusterName" : {
+      "Type": "String",
+      "Default": "eksMagmaCluster",
+      "Description": "EKS Cluster Name"
+    },
+    "paramCidr" : {
+      "Type": "String",
+      "Default": "10.19.0.0/16",
+      "Description": "CIDR for EKS VPC"
+    },
+    "paramPub1Cidr" : {
+      "Type": "String",
+      "Default": "10.19.6.0/24",
+      "Description": "CIDR for EKS Public Public 1"
+    },
+    "paramPub2Cidr" : {
+      "Type": "String",
+      "Default": "10.19.8.0/24",
+      "Description": "CIDR for EKS Public Network 2"
+    },
+    "paramPriv1Cidr" : {
+      "Type": "String",
+      "Default": "10.19.4.0/24",
+      "Description": "CIDR for EKS Private Network 1"
+    },
+    "paramPriv2Cidr" : {
+      "Type": "String",
+      "Default": "10.19.2.0/24",
+      "Description": "CIDR for EKS Private Network 2"
+    },
+    "paramAvlZone1" : {
+      "Type": "AWS::EC2::AvailabilityZone::Name",
+      "Default": "ap-northeast-1a",
+      "Description": "Availability Zone 1"
+    },
+    "paramAvlZone2" : {
+      "Type": "AWS::EC2::AvailabilityZone::Name",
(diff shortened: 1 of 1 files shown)


experimental/cloudstrapper/playbooks/roles/cfn/cfnMagmaEksFargate.json
```diff
@@ -0,0 +1,294 @@
...
+    },
+    "sgVpc": {
+      "Type" : "AWS::EC2::SecurityGroup",
+      "Properties" : {
+        "GroupDescription" : "Allow all TCP to client host",
+	"GroupName": { "Ref": "paramSecGroupName" },
+        "VpcId" : {"Ref" : "vpcMagmaEks"},
+        "SecurityGroupIngress" : [{
+          "IpProtocol" : "-1",
+          "FromPort" : 0,
```
krramasw, line 234: I understand it is experimental for now.  I guess,  at some point we need to list the specific ports (originating from AGW) so that it is not wide open.  Perhaps we can document all ports used by AGW -> Orchestrator

experimental/cloudstrapper/playbooks/roles/cfn/cfnMagmaEksFargate.json
```diff
@@ -0,0 +1,294 @@
...
+          "ToPort" : 65535,
+          "CidrIp" : "0.0.0.0/0"
+        }]
+      }
+    },
+    "roleMagmaEksCluster": {
+      "Type": "AWS::IAM::Role",
+      "Properties": {
```
krramasw, line 248: If NMS/Orch wants to utilize AWS IAM for auth rather than the embedded one comes with the magma, wonder we need to change anything here.

experimental/cloudstrapper/playbooks/roles/cfn/cfnMagmaEksFargate.json
arunuke: Agree. Individual implementations would have to choose which ports need to be locked and which ones need to open based on the services that are run and hopefully, this experimental feature provides a placeholder for all such rules. The container build script includes a list of all ports needed by the various services that run on the AGW and could provide the first template.

experimental/cloudstrapper/playbooks/roles/cfn/cfnMagmaEksFargate.json
arunuke: The two key use-cases this experimental feature is hoping to address are (a) Can we move away from a static, EC2 based compute backend to a more flexible Fargate backend and reduce the Orc8r cost footprint and (b) Can a Fargate based environment for AGWs and Orc8r provide an alternate dev environment that can be used for scale (over EC2 which has some limitations). When Orc8r moves to using Fargate, it will use the embedded policies that come with Magma.

quentinDERORY (approved):
So with the fargate backed EKS cluster, we could apply directly terraform or8r-aws-helm part with this eks cluster id ? Did you already tested it out? Or are you manually deploying the helm charts there?

arunuke:
> So with the fargate backed EKS cluster, we could apply directly terraform or8r-aws-helm part with this eks cluster id ? Did you already tested it out? Or are you manually deploying the helm charts there?

Correct - we can use this template under terraform to swap out the eks elements, auto scaling groups and the ec2 instances while keeping the rest of the elements the same (rds, secrets etc.) and do a helm deployment of the magma services. That might be best way to minimize disruption.

PR #8496, opened 2021-08-05
fix(mme): Fixed mme crash occurring while updating stats

rsarwad:
fix(mme): Fixed mme crash occurring while updating stats
## Summary
Issue description: While executing roaming attach, s8 task was crashing while updating stats.
Fix: increment_counter() is defined now only in file, MetricsHelpers.cpp, earlier it was defined in service303.cpp also.
so inclusion of file MetricsHelpers.h resolved the crash.
## Test Plan


uri200 (approved):

waqaraqeel (approved):
Thank you for the fix!

ardzoht (approved):

ulaskozat (approved):
LGTM

PR #8513, opened 2021-08-05
refactor(nms): Hooked feg gateway context provider to existing gateway page

ato-dani:
## Summary
The existing federation gateway page used to fetch gateway information data from the API using a useEffect hook. However, during the previous weeks, since a gateway context provider was created, the context provider was hooked to the page. This way the data source(i.e the context provider) will be the same on all pages and wouldn't create any data difference or confusion.

## Test Plan
A test named FEGGatewaysTest.js was created to test that changes didn't break previous implementation.

## Additional Information
Screen Shot 2021-08-05 at 4 56 03 PM

### DIFF
--- a/nms/app/packages/magmalte/app/components/feg/FEGGateways.js
+++ b/nms/app/packages/magmalte/app/components/feg/FEGGateways.js
@@ -18,13 +18,12 @@ import type {WithAlert} from '@fbcnms/ui/components/Alert/withAlert';
 import type {federation_gateway} from '@fbcnms/magma-api';
 
 import Button from '@material-ui/core/Button';
-import CircularProgress from '@material-ui/core/CircularProgress';
 import DeleteIcon from '@material-ui/icons/Delete';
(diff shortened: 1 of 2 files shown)


nms/app/packages/magmalte/app/components/feg/FEGGateways.js
```diff
@@ -86,44 +84,22 @@ const useStyles = makeStyles(theme => ({
...
-        setGateways(
-          Object.keys(response)
-            .map(k => response[k])
-            .filter(g => g.id),
-        ),
-      [],
-    ),
+  const ctx = useContext(FEGGatewayContext);
+  const [gateways, setGateways] = useState<federation_gateway[]>(
+    Object.keys(ctx.state).map(gatewayId => ctx.state[gatewayId]),
```
ato-dani, line 89: Used Object.keys + map to get an array of the values here instead of Object.values because Object.values returns an array of mixed type which gives a flow error. https://stackoverflow.com/questions/47187242/flow-object-type-incompatible-with-arraymixed

andreilee (approved):

PR #8514, opened 2021-08-05
refactor: migrate subscriberdb to generic digests pattern

wangyyt1013:

## Summary

1. Migrate cloud services `subscriberdb` and `subscriberdb_cache` to use `syncstore` for storage
2. Use more generic naming for `subscriberdb`-related cloud-AGW gRPC endpoints 
3. Adapt AGW-side client to use the new endpoints

## Test Plan

2. Verified e2e workflow in `subscriberdb` via local docker

## Additional Information


lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -83,53 +84,52 @@ func GetPerSubscriberDigests(network string) ([]*lte_protos.SubscriberDigestWith
...
-// GetPerSubscriberDigestsDiff computes the changeset between two lists of per-subscriber digests,
-// ordered by their subscriber IDs (unique within a network). It returns
-// 1. A set of subscribers that have been added/modified, with the new digests.
-// 2. An ordered list of subscribers that have been removed.
-func GetPerSubscriberDigestsDiff(prev []*lte_protos.SubscriberDigestWithID, next []*lte_protos.SubscriberDigestWithID) (map[string]string, []string) {
+// GetLeafDigestsDiff computes the data changeset according to two lists of leaf digests,
+// ordered by their IDs. It returns
+// 1. The set of objects that have been added/modified, with the new digests.
+// 2. An ordered list of the objects that have been removed.
+func GetLeafDigestsDiff(prev []*orc8r_protos.LeafDigest, next []*orc8r_protos.LeafDigest) (map[string]string, []string) {
```
hcgatewood: This is now generic / not unique to subscriberdb, so we can export it to somewhere in the syncstore pkg

lte/cloud/go/services/subscriberdb/digest_test.go
```diff
@@ -369,89 +369,90 @@ func TestGetPerSubscriberDigests(t *testing.T) {
...
-	assert.Equal(t, "00001", perSubDigests[0].Sid.Id)
+	assert.Equal(t, "IMSI00001", perSubDigests[0].Id)
 	assert.NotEqual(t, digestSub1, perSubDigests[0].Digest.Md5Base64Digest)
-	assert.Equal(t, "00003", perSubDigests[1].Sid.Id)
+	assert.Equal(t, "IMSI00003", perSubDigests[1].Id)
 	assert.NotEqual(t, digestSub3, perSubDigests[1].Digest.Md5Base64Digest)
 }
 
-func TestGetPerSubscriberDigestsDiff(t *testing.T) {
+func TestGetLeafDigestsDiff(t *testing.T) {
```
hcgatewood: Same idea, move this test along with its impl

lte/cloud/go/services/subscriberdb/load.go
```diff
@@ -63,6 +64,32 @@ func LoadSubProtosPage(
 	return subProtos, nextToken, nil
 }
 
+func SerializeSubProtos(subProtos []*lte_protos.SubscriberData) (map[string][]byte, error) {
```
hcgatewood: "SerializeSubscribers" and "DeserializeSubscribers"

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -19,59 +19,52 @@ import (
...
 func NewSubscriberdbServicer(
 	config subscriberdb.Config,
-	digestStore storage.DigestStore,
-	perSubDigestStore *storage.PerSubDigestStore,
-	subStore *storage.SubStore,
-	lastResyncTimeStore *storage.LastResyncTimeStore,
+	store syncstore.SyncStoreReader,
 ) lte_protos.SubscriberDBCloudServer {
 	servicer := &subscriberdbServicer{
 		digestsEnabled:         config.DigestsEnabled,
```
hcgatewood: Rather than copying the config values over one-by-one, you *can* use type embedding, if you want https://play.golang.org/p/FNzNYbGgQc6

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -254,32 +262,32 @@ func (s *subscriberdbServicer) loadSubscribersPageFromCache(networkID string, re
...
 	// If check last resync time in store fails, swallow the error and stick to the original callpath
 	if err != nil {
 		glog.Errorf("check last resync time of gateway %+v of network %+v: %+v", gateway, network, err)
 		return false
 	}
 	// Add a deterministic jitter to AGW sync intervals in the range of
 	// [0, 0.5 * resyncIntervalSecs] to ameliorate the thundering herd effect
-	resyncIntervalJitter := math.JitterUint32(l.resyncIntervalSecs, gateway, 0.5)
-	shouldResync := uint32(time.Now().Unix())-lastResyncTime > l.resyncIntervalSecs+resyncIntervalJitter
+	resyncIntervalJitter := math.JitterInt64(l.resyncIntervalSecs, gateway, 0.5)
```
hcgatewood: Why Int64 instead of Uint64?

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer_test.go
```diff
@@ -849,46 +848,26 @@ func assertEqualSubscriberData(t *testing.T, expectedProtos []*lte_protos.Subscr
 	}
 }
 
-func initializeDigestStore(t *testing.T) subscriberdb_storage.DigestStore {
-	db, err := sqorc.Open("sqlite3", ":memory:")
-	assert.NoError(t, err)
-	store := subscriberdb_storage.NewDigestStore(db, sqorc.GetSqlBuilder())
-	assert.NoError(t, store.Initialize())
-	return store
+func assertEqualAnyData(t *testing.T, expected []*any.Any, got []*any.Any) {
```
hcgatewood, line 852: I think you might be able to, either as a replacement for this fn, or within this fn, use https://github.com/gogo/protobuf/blob/master/proto/equal.go#L72

lte/cloud/go/services/subscriberdb_cache/subscriberdb_cache/main.go
```diff
@@ -32,30 +32,31 @@ func main() {
...
 	if err != nil {
 		glog.Fatalf("Error opening db connection: %+v", err)
 	}
-	digestStore := subscriberdb_storage.NewDigestStore(db, sqorc.GetSqlBuilder())
-	if err := digestStore.Initialize(); err != nil {
-		glog.Fatalf("Error initializing digest storage: %+v", err)
-	}
-
-	fact := blobstore.NewEntStorage(subscriberdb.PerSubDigestTableBlobstore, db, sqorc.GetSqlBuilder())
+	fact := blobstore.NewEntStorage(subscriberdb.SyncstoreBlobstore, db, sqorc.GetSqlBuilder())
```
hcgatewood: "SyncstoreBlobstoreType" do we use that "Type" suffix elsewhere? Seems like we should

lte/cloud/go/services/subscriberdb_cache/subscriberdb_cache/main.go
```diff
@@ -32,30 +32,31 @@ func main() {
...
-	perSubDigestStore := subscriberdb_storage.NewPerSubDigestStore(fact)
-
-	subStore := subscriberdb_storage.NewSubStore(db, sqorc.GetSqlBuilder())
-	if err := subStore.Initialize(); err != nil {
-		glog.Fatalf("Error initializing subscriber proto storage: %+v", err)
+	// Garbage collection interval for syncstore cache writers is enforced to be half the time for the service worker's
+	// update interval, to prevent cache writers from outliving update cycles
+	store, err := syncstore.NewSyncStore(db, sqorc.GetSqlBuilder(), fact, syncstore.Config{
+		TableNamePrefix:              subscriberdb.SyncstoreTableNamePrefix,
+		CacheWriterValidIntervalSecs: int64(serviceConfig.SleepIntervalSecs / 2),
```
hcgatewood, line 55: Make sure we validate somewhere that the sleepinterval is at least 1 min

lte/cloud/go/services/subscriberdb_cache/worker.go
```diff
@@ -16,28 +16,28 @@ package subscriberdb_cache
...
+	"magma/orc8r/lib/go/protos"
 
 	"github.com/golang/glog"
 	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/thoas/go-funk"
 )
 
-func MonitorDigests(config Config, digestStore storage.DigestStore, perSubDigestStore *storage.PerSubDigestStore, subStore *storage.SubStore) {
+func MonitorDigests(config Config, store syncstore.SyncStore) {
```
hcgatewood, line 32: So what's the verdict: you think worker.go can be at least somewhat generalized? If yes I'll have my next intern work on it 🙂

orc8r/cloud/go/orc8r/math/math.go
```diff
@@ -28,3 +28,17 @@ func JitterUint32(n uint32, key string, maxMultiplier float32) uint32 {
 	maxJitter := float32(n) * maxMultiplier
 	return uint32(multiplier * maxJitter)
 }
+
+// JitterInt64 returns a jitter of the given int64 value that is deterministic
+// based on the given key.
+func JitterInt64(n int64, key string, maxMultiplier float64) int64 {
+	// FNV-1 is a non-cryptographic hash function that is fast and very simple to implement
```
hcgatewood, line 47: If we do keep this JitterInt64 around, pull out the part where we generate a deterministic int from the key

orc8r/cloud/go/orc8r/math/math.go
```diff
@@ -28,3 +28,17 @@ func JitterUint32(n uint32, key string, maxMultiplier float32) uint32 {
...
+	h := fnv.New32a()
+	_, err := h.Write([]byte(key))
+	if err != nil {
+		return 0
+	}
+	multiplier := float64(h.Sum32()%100) / 100.0
+	maxJitter := float64(n) * maxMultiplier
+	return int64(multiplier * maxJitter)
```
hcgatewood: Also I'm realizing: to both, add a check that "output >= input", and if false, just return input

Not relevant now, but we don't know how this fn might be used in the future, so we want to account for overflowing the int type

hcgatewood (commented):
Looks great! Have you been able to validate the AGW changes?

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
wangyyt1013: Because oftentimes `lastResyncTime` is used arithmetically with `time.Now().Unix()` (which is `int64`), so using `int64` saves more type casting @hcgatewood

hcgatewood: Cool, makes sense

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer_test.go
wangyyt1013: I have tried using `proto.Equal`, unfortunately it doesn't seem to work with the `any.Any` type @hcgatewood

hcgatewood: Hm must be because we're using an outdated version. Will you add a comment then about why you're not using proto.Equal

lte/cloud/go/services/subscriberdb_cache/subscriberdb_cache/main.go
wangyyt1013: I think this is supposed to be the blobstore table name; the lookup table in `subscriberdb` uses `subscriberdb.LookupTableBlobstore`, so I'll change `SyncStoreBlobstore` to `SyncstoreTableBlobstore` for consistency? @hcgatewood

hcgatewood: 👍

lte/cloud/go/services/subscriberdb/servicers/subscriberdb_servicer.go
```diff
@@ -254,32 +262,32 @@ func (s *subscriberdbServicer) loadSubscribersPageFromCache(networkID string, re
...
+	shouldResync := time.Now().Unix()-lastResyncTime > l.resyncIntervalSecs+resyncIntervalJitter
 	return shouldResync
 }
```
krramasw, line 274: time.Now().Unix() is zone independent. Just wondering whether we need to normalize lastResyncTime to match with time.Now().Unix default zone.

lte/cloud/go/services/subscriberdb_cache/worker.go
wangyyt1013: I think definitely with the use of a worker object that accepts custom functions as initializing arguments. Currently, the `subscriberdb` worker follows the workflow of
- Every <loop_interval>, Initiate loop where
  - garbage collection (takes in `trackedNetworks`, rest is left to `syncstore`)
  - find networks to renew (takes in `trackedNetworks`, rest depends on networks in `syncstore`)
  - renew digests (uses custom function to generate new digest trees)
  - renew cached objects (uses custom function to generate new cached objects)

All these steps are generic enough to be propagated for other service workers as well, as long as we take in the service functions to generate new digests & cached objects @hcgatewood

wangyyt1013:
> Looks great! Have you been able to validate the AGW changes?

As far as validating AGW changes, I've run the python unit tests and they did all pass; should I also try to run `make integ_tests` again / validate the workflow manually via local docker instances? @hcgatewood

lte/cloud/go/services/subscriberdb/digest.go
```diff
@@ -22,6 +22,7 @@ import (
 	"magma/orc8r/cloud/go/mproto"
 	mproto_protos "magma/orc8r/cloud/go/mproto/protos"
 	"magma/orc8r/cloud/go/services/configurator"
+	orc8r_protos "magma/orc8r/lib/go/protos"
```
hcgatewood, line 25: No need to alias this one (?)

lte/cloud/go/services/subscriberdb/load.go
```diff
@@ -63,6 +64,32 @@ func LoadSubProtosPage(
 	return subProtos, nextToken, nil
 }
 
+func SerializeSubscribers(subProtos []*lte_protos.SubscriberData) (map[string][]byte, error) {
+	subProtosSerialized := map[string][]byte{}
```
hcgatewood: Nit here and below: I would just call these "subsSerialized" (serialized) and subs (deserialized)

lte/cloud/go/services/subscriberdb_cache/subscriberdb_cache/main.go
```diff
@@ -32,30 +32,34 @@ func main() {
 		glog.Fatalf("Error creating service: %+v", err)
 	}
 
+	serviceConfig := subscriberdb_cache.MustGetServiceConfig()
+	if serviceConfig.Validate() != nil {
+		glog.Fatalf("Invalid service configs for subscriberdb_cache")
```
hcgatewood: Include the generated error in the fatal for debuggability 

Also, move the validation inside the MustGetServiceConfig for simplicity

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -233,23 +239,42 @@ async def _get_all_subscribers(self) -> Optional[CloudSubscribersInfo]:
...
-        if Digest is None:
+    def _process_changeset(self, changeset: Optional[Changeset]) -> ProcessedChangeset:
+        if changeset is None:
+            return ProcessedChangeset(to_renew=[], deleted=[])
+
+        to_renew, deleted = [], []
+        if changeset.deleted is not None:
+            deleted = changeset.deleted
+        if changeset.to_renew is not None:
+            for anyVal in changeset.to_renew:
```
hcgatewood: Nit: `any_val` in python

orc8r/cloud/go/orc8r/math/math.go
```diff
@@ -15,16 +15,34 @@ package math
 
 import "hash/fnv"
 
-// JitterUint32 returns a jitter of the given uint32 value that is deterministic
-// based on the given key.
-func JitterUint32(n uint32, key string, maxMultiplier float32) uint32 {
+func getFNV1Hash(key string) uint32 {
```
hcgatewood: Nit: move this to bottom of file, we want the public fns to be at the top

orc8r/cloud/go/orc8r/math/math.go
```diff
@@ -15,16 +15,34 @@ package math
...
-	multiplier := float32(h.Sum32()%100) / 100.0
-	maxJitter := float32(n) * maxMultiplier
-	return uint32(multiplier * maxJitter)
+	return h.Sum32()
+}
+
+// JitterUint32 returns a jitter of the given uint32 value that is deterministic
+// based on the given key.
+func JitterUint32(n uint32, key string, maxMultiplier float32) uint32 {
+	fnv1Hash := getFNV1Hash(key)
```
hcgatewood: Nit here and below: just call this `hash := ...`

orc8r/cloud/go/orc8r/math/math.go
```diff
@@ -15,16 +15,34 @@ package math
...
+	multiplier := maxMultiplier * float32(fnv1Hash%100) / 100.0
+	if multiplier >= 1 {
+		return n
+	}
+	return uint32(multiplier * float32(n))
```
hcgatewood: Here and below: this isn't quite what we want, actually should be something like

```suggestion
	multiplier := maxMultiplier * float32(hash%100) / 100.0
	jittered := n + (n * multiplier)

	// Check for integer overflow
	if jittered < n {
		return n
	}
	
	return jittered
```

orc8r/cloud/go/syncstore/digest_utils.go
```diff
@@ -0,0 +1,54 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package syncstore
+
+import orc8r_protos "magma/orc8r/lib/go/protos"
```
hcgatewood: No need to alias

orc8r/cloud/go/syncstore/digest_utils.go
```diff
@@ -0,0 +1,54 @@
+/*
```
hcgatewood: Nit: "utils" is an unproductive naming pattern. Just call this digest.go and digest_test.go

orc8r/cloud/go/syncstore/store.go
```diff
@@ -65,3 +67,9 @@ type CacheWriter interface {
 	// Apply completes the batch cache update.
 	Apply() error
 }
+
+type DigestTrees map[string]*protos.DigestTree
+
+func (digestTrees DigestTrees) Networks() []string {
+	return funk.Keys(digestTrees).([]string)
```
hcgatewood: Nit: just construct this manually with a for loop, I've slowly turned away from over-using funk 🙂

orc8r/protos/digest.proto
```diff
@@ -15,9 +15,10 @@ syntax = "proto3";
 
 package magma.orc8r;
 option go_package = "magma/orc8r/lib/go/protos";
+import "google/protobuf/any.proto";
 
 // DigestTree contains the full set of digest information for a particular network.
-// DigestTree is similar to a depth=2 Merkle tree.
+// DigestTree is seimilar to a depth=2 Merkle tree.
```
hcgatewood: "similar"

hcgatewood (approved):
Everything looks good! Main thing is just verify that this works e2e (or I can do it 🙂), then I'm ready to get this merged

themarwhal (approved):

lte/cloud/go/services/subscriberdb/digest.go
wangyyt1013: I think this is aliased because the `protos` in this file refers to the `subscriberdb` service-specific protos @hcgatewood

orc8r/cloud/go/orc8r/math/math.go
wangyyt1013: I thought the function is supposed to return a jitter (a fraction of `n`) instead of `n + jitter`, according to its original implementation? @hcgatewood

hcgatewood: AH yes, I see. So you're correct, it's currently implemented+used as `return jitter`. That said, I'd originally envisioned it would be `return n + jitter`). And I think the latter is preferable because, like I mention, we can handle the case where there's integer overflow

Up to you though, if you wanna leave it that's fine with me

hcgatewood (approved):

ardzoht (approved):
lgtm

PR #8517, opened 2021-08-06
fix(mme): Free bstring field for ICS response

ardzoht:

## Summary

- Backporting to v1.5 the according changes on #8481 

## Test Plan

- ran s1ap integ tests on magma VM from v1.5 branch, validating there's no memory leak 

## Additional Information


ulaskozat (approved):
LGTM

PR #8535, opened 2021-08-08
feat(orc8r): Context propagation for all network-related functions in configurator client API

xjtian:
## Summary

- Propagate contexts in all network functions in configurator client_api.go

## Test Plan

- make test

## Additional Information

Signed-off-by: Jacky Tian <jacky@xjtian.com>

### DIFF
--- a/cwf/cloud/go/services/cwf/obsidian/handlers/handlers.go
+++ b/cwf/cloud/go/services/cwf/obsidian/handlers/handlers.go
@@ -236,7 +236,7 @@ func getSubscriberDirectoryHandler(c echo.Context) error {
 	}
 
 	reqCtx := c.Request().Context()
-	configuratorNetwork, err := configurator.LoadNetwork(networkID, false, false, serdes.Network)
+	configuratorNetwork, err := configurator.LoadNetwork(reqCtx, networkID, false, false, serdes.Network)
 	if err != nil {
 		return obsidian.HttpError(err, http.StatusNotFound)
 	}
@@ -285,7 +285,7 @@ func getHAPairStatusHandler(c echo.Context) error {
 	}
 
 	reqCtx := c.Request().Context()
-	network, err := configurator.LoadNetwork(nid, true, true, serdes.Network)
+	network, err := configurator.LoadNetwork(reqCtx, nid, true, true, serdes.Network)
 	if err == merrors.ErrNotFound {
 		return c.NoContent(http.StatusNotFound)
 	}
@@ -356,7 +356,7 @@ func createHAPairHandler(c echo.Context) error {
 	if err := c.Bind(haPair); err != nil {
 		return obsidian.HttpError(err, http.StatusBadRequest)
 	}
-	if err := haPair.ValidateModel(); err != nil {
+	if err := haPair.ValidateModel(context.Background()); err != nil {
 		return obsidian.HttpError(err, http.StatusBadRequest)
 	}
 	_, err := configurator.CreateEntity(networkID, haPair.ToEntity(), serdes.Entity)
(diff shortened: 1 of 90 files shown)


cwf/cloud/go/services/cwf/obsidian/handlers/handlers.go
```diff
@@ -356,7 +356,7 @@ func createHAPairHandler(c echo.Context) error {
 	if err := c.Bind(haPair); err != nil {
 		return obsidian.HttpError(err, http.StatusBadRequest)
 	}
-	if err := haPair.ValidateModel(); err != nil {
+	if err := haPair.ValidateModel(context.Background()); err != nil {
```
hcgatewood, line 359: Why context.Background here? Assuming we can grab the reqCtx from the echo.Context and pass it in

lte/cloud/go/services/lte/analytics/calculations/state.go
```diff
@@ -48,7 +48,7 @@ func (x *UserMetricsCalculation) Calculate(prometheusClient query_api.Prometheus
 	glog.V(1).Info("Calculate User Metrics")
 
 	var results []*protos.CalculationResult
-	networks, err := configurator.ListNetworkIDs()
+	networks, err := configurator.ListNetworkIDs(context.Background())
```
hcgatewood, line 51: All calculations can take a ctx from https://github.com/magma/magma/blob/4676cc0b7897a219473f41f6b6e02d64da3d141a/orc8r/cloud/go/services/analytics/collector_servicer.go#L49-L50

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
```diff
@@ -327,7 +328,7 @@ func createEnodeb(c echo.Context) error {
 	if err := c.Bind(payload); err != nil {
 		return obsidian.HttpError(err, http.StatusBadRequest)
 	}
-	if err := payload.ValidateModel(); err != nil {
+	if err := payload.ValidateModel(context.Background()); err != nil {
```
hcgatewood, line 331: Same here and below: seems like most of these can take a reqCtx

lte/cloud/go/services/lte/obsidian/models/validate.go
```diff
@@ -27,23 +28,23 @@ import (
...
 
-func (m *LteNetwork) ValidateModel() error {
+func (m *LteNetwork) ValidateModel(context.Context) error {
 	if err := m.Validate(strfmt.Default); err != nil {
 		return err
 	}
 
 	var res []error
-	if err := m.Cellular.ValidateModel(); err != nil {
+	if err := m.Cellular.ValidateModel(context.Background()); err != nil {
```
hcgatewood, line 37: These can be the passed context param

hcgatewood (changes requested):
Looking good! Got about halfway through, main thing I notice is we're passing context.Background() as args in a fair number of places where we actually wanna forward a context from the scope's params

Take a skim through when you have a chance, then msg me when you're ready and I'll give it another look over 🙂

xjtian:
> Looking good! Got about halfway through, main thing I notice is we're passing context.Background() as args in a fair number of places where we actually wanna forward a context from the scope's params
> 
> Take a skim through when you have a chance, then msg me when you're ready and I'll give it another look over 🙂

My plan at the moment is to tackle the top-level callsites of `ValidateModel` along the way as I refactor the entity-related functions - that diff will hit all of those. Then I can follow that up with a cleanup diff that addresses the recursive `ValidateModel` calls in the model definitions.

Thoughts?

hcgatewood:
> > Looking good! Got about halfway through, main thing I notice is we're passing context.Background() as args in a fair number of places where we actually wanna forward a context from the scope's params
> > Take a skim through when you have a chance, then msg me when you're ready and I'll give it another look over 🙂
> 
> My plan at the moment is to tackle the top-level callsites of `ValidateModel` along the way as I refactor the entity-related functions - that diff will hit all of those. Then I can follow that up with a cleanup diff that addresses the recursive `ValidateModel` calls in the model definitions.
> 
> Thoughts?

Yeah I think what we can do then is I'll just go ahead and stamp all your PRs going forward, then at the end we can have ~2 cleanup PRs

1. Remove all the `x/.../context` in favor of the stdlib `context`
2. Comb through all non-test `context.Background()` declarations and determine whether they should be removed in favor of passing a received context

hcgatewood (approved):

themarwhal (approved):

uri200 (approved):

PR #8580, opened 2021-08-10
fix(mme): Modified the proto structure, PacketFilter defined in file, ../protos/oai/spgw_state.proto to take care of multiple occurrences of packet filter contents within each packet filter

### LINKED ISSUE #5449: [Inbound Roaming][MME] Dedicated bearer support for Roaming
AGW to support dedicated bearers towards PGW over s5/s8.

rsarwad:
fix(mme): Modified the proto structure, PacketFilter defined in file, ../protos/oai/spgw_state.proto to take care of multiple occurrences of packet filter contents within each packet filter
## Summary
Proto structure for PacketFilter is used in two scenarios, 
Scenario-1: The structure is used while storing tft contents to Redis DB. 
Scenario-2: The structure is used over s8 interface while sending tft info from roaming network's PGW to MME through FedGW.
In Scenario-1, the structure contains flag, where each bit in flag was used to indicate presence of specific IP flow within packet filter content. But as per 24.008 spec, flag here is "packet_filter_component_type_identifier" that holds various values as per spec, so each bit in flag can not be used, as there could be multiple bits are set for 3gpp defined values.

To address this, we changed proto to have multiple occurrences of packet filter contents, flag holds the 3gpp defined values
## Test Plan


uri200:
As discussed with @rsarwad , the checks on this PR will not pass until I do the changes on FEG side.

Please, ignore cwag-build and feg-precommit errors

uri200 (approved):

ardzoht (approved):
thanks for this change

PR #8583, opened 2021-08-10
feat(mme): 5G Traffic test related fixes

### LINKED ISSUE #8708: Magma- SA TeraVM | AGW sends unknown "Transport layer Addr" in PDU session setup Request
AGW sends unknown "Transport layer Addr (49.48.46.50)" in PDU session setup Request , This address is not the eth1 or any interface IP of AGW, Need to investigate the source of this IP selection.

Here is the snap of the traces

Screen Shot 2021-08-20 at 5 27 06 PM

Enclosed the pcap

[SA_01_02_03_04_Hyper_Suci_210818_050204-1.pcapng.zip](https://github.com/magma/magma/files/7024577/SA_01_02_03_04_Hyper_Suci_210818_050204-1.pcapng.zip)

panyogesh:
Fix :
  1. Changed the local IP Address recieved from sessiond to network
     format and relaying to GNB.
  2. Changing the default bandwidth to 1 Gigabits

Test:
  1. Tested using UERANSIM. Packets are reaching table=0

cookie=0x0, duration=224.372s, table=0, n_packets=17, n_bytes=1428
cookie=0x0, duration=224.354s, table=0, n_packets=17, n_bytes=1666

## Additional Information


panyogesh:
[Test-Logs-8583.txt](https://github.com/magma/magma/files/6963669/Test-Logs-8583.txt)

lte/gateway/c/core/oai/include/n11_messages_types.h
```diff
@@ -179,12 +179,43 @@ typedef struct AggregatedMaximumBitrate_respose_t {
...
+  Tbps_64  = Tbps_0 + 4,
+  Tbps_256 = Tbps_0 + 5,
+
+  Pbps_0   = Tbps_256,
+  Pbps_1   = Pbps_0 + 1,
+  Pbps_4   = Pbps_0 + 2,
+  Pbps_16  = Pbps_0 + 3,
+  Pbps_64  = Pbps_0 + 4,
+  Pbps_256 = Pbps_0 + 5
+
```
uri200, line 218: instead of defining every single element as an enumm, can't we just use the raw value?

lte/protos/session_manager.proto
```diff
@@ -935,11 +935,36 @@ message RatSpecificContextAccess {
...
+   Tbps_4   = 17;
+   Tbps_16  = 18;
+   Tbps_64  = 19;
+   Tbps_256 = 20;
+
+   Pbps_1   = 21;
+   Pbps_4   = 22;
+   Pbps_16  = 23;
+   Pbps_64  = 24;
+   Pbps_256 = 25;
```
uri200: Instead of defining every single constant like on 3gpp, can't we just pass the raw value and do the conversion at `cpp` level if required?

I have the feeling we are defining three times these constants just to convert them back and forth across the AMF, SMF, Pipelined. I think the only place it really matters to know what those value mean is on the system applying the Qos (I guess pipelined). For the rest of the system, the raw value is enough. I don't think thre is any difference for sessiond to use `3` int vs `Kbps_16` enum

This way we keep the logic in the services. Proto shouldn't really care what the value means. Plus may force the sender to do an extra conversion from raw value into an enum



The system that needs to understand what raw values need it will have to have a function that matches raw to a actual value: 
```
0 to int(0)
1 to int(1)
3 to int(16)
10 to int(256000000)
...
```

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
```diff
@@ -576,12 +576,15 @@ void SessionStateEnforcer::prepare_response_to_access(
   /* AMBR value need to compared from AMF and PCF, then fill the required
    * values and sent to AMF.
    */
-  // For now its default QOS, AMBR has default values
+  // For now its default QOS, AMBR is 1 Gbps downlink
   rsp->mutable_downlink_unit_type()->set_ambr_unit_type(
-      magma::AmbrUnit::Kbps_64);
-  rsp->mutable_uplink_unit_type()->set_ambr_unit_type(magma::AmbrUnit::Kbps_64);
+      magma::AmbrUnit::Gbps_1);
```
uri200: see my comment at the [bottom](https://github.com/magma/magma/pull/8583/files#r686339624) first:

in this case if we want to set any ambrunit we will need to create a switch function that transforms the  QOS value (for example 11) into `magma::AmbrUnit::Gbps_1`.  If you don't define any enum you can just do 

```
 rsp->mutable_downlink_unit_type()->set_ambr_unit_type(config.rat_specific_context.m5gsm_session_context().qos.get_ambr_unit_type());

```
(note qos access through config may be wrong in my example)

lte/protos/session_manager.proto
panyogesh: Comments addressed.

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
panyogesh: Right now for 5G context,  sessiond is using a default value. As part of QoS changes this value will be derived from either default subscriber profile or values stored in policydb.  The qos infrastructure like "qos.get_ambr_unit_type" still needs to be updated for 5G Context in sessiond.   Have tried to keep the changes in line based on the comments.

lte/protos/session_manager.proto
```diff
@@ -933,17 +933,6 @@ message RatSpecificContextAccess {
...
-  enum AmbrUnitType {
-   Kbps_0 = 0;
-   Kbps_1 = 1;
-   Kbps_4 = 2;
-   Kbps_16 = 3;
-   Kbps_64 = 4;
-  }
-  AmbrUnitType ambr_unit_type = 1;
-}
-
```
uri200, line 946: removing this looks good. 

For future cases, remember we can pass around the raw value. Let the service that uses that value to decode it without the need to go through 3gpp structures

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
```diff
@@ -576,12 +576,11 @@ void SessionStateEnforcer::prepare_response_to_access(
...
-  rsp->mutable_downlink_unit_type()->set_ambr_unit_type(
-      magma::AmbrUnit::Kbps_64);
-  rsp->mutable_uplink_unit_type()->set_ambr_unit_type(magma::AmbrUnit::Kbps_64);
-  rsp->set_downlink_units(DEFAULT_AMBR_UNITS);
-  rsp->set_uplink_units(DEFAULT_AMBR_UNITS);
+  // For now its default QOS, AMBR is 1 Gbps downlink
+  rsp->mutable_session_ambr()->set_br_unit(AggregatedMaximumBitrate::KBPS);
+  rsp->mutable_session_ambr()->set_max_bandwidth_ul(DEFAULT_AMBR_UNITS);
+  rsp->mutable_session_ambr()->set_max_bandwidth_dl(DEFAULT_AMBR_UNITS);
+
```
uri200, line 583: can you please point me to any unit test that tests this function `prepare_response_to_access`

ulaskozat:
Based on my conversation with @panyogesh, session AMBR is temporarily hardwired in sessiond. Since this is the same as APN AMBR in 4G, we will reuse APN profile and AMF will get the value and will fetch it to sessiond in a separate PR.

panyogesh:
Test Logs
---------------
vagrant@ue-ran-sim:~$ ping 192.168.129.42 -I  192.168.128.254 -c 5
PING 192.168.129.42 (192.168.129.42) from 192.168.128.254 : 56(84) bytes of data.
64 bytes from 192.168.129.42: icmp_seq=1 ttl=63 time=5.92 ms


      Start 17: test_set_session_manager_handler
17/17 Test #17: test_set_session_manager_handler .....   Passed    0.23 sec

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
panyogesh: prepare_response_to_access : No direct UT for it.  Extended coverage for m5g_move_to_active_state() which calls the function internally.

ulaskozat (approved):
LGTM

ulaskozat:
@uri200 could you approve if your concerns are addressed?

uri200 (approved):

PR #8586, opened 2021-08-10
fix(orc8r): Fix Orc8r bug causing subscriber sync to fail

hcgatewood:
## Summary

Subscriber sync is failing due to "read only" being incorrectly assigned. Also cleaned up other usages.

Refs

- Error: https://gist.github.com/themarwhal/fa1a93fd0d6a03b584cfc130c4ba69c4
- Reported: https://magmacore.slack.com/archives/C01RENQQZB4/p1628623693054900

## Additional Information


andreilee (approved):

hcgatewood:
Force-merging to unblock. All CircleCI checks passed, seems to be an issue with GH Actions

PR #8592, opened 2021-08-10
test(s8_proxy): temporary hide tft filters so we can change definition on MME

### LINKED ISSUE #6219: [Inbound Roaming][FEG] Dedicated bearer support for Roaming

uri200:

## Summary

tft filter on oai protos is not properly (fix here https://github.com/magma/magma/pull/8580)

But since tft filters are used both in AGW and feg that fix can't land without having both sides fixed.

Since tft filters at feg are used by a feature yet to be completed, this PR disables all the changes at s8_proxy to allow #8580 to land and let us test AGW independently from the feg changes

## Test Plan
./build.py c

## Additional Information

### DIFF
--- a/feg/gateway/services/s8_proxy/servicers/mock_pgw/cb.go
+++ b/feg/gateway/services/s8_proxy/servicers/mock_pgw/cb.go
@@ -13,6 +13,7 @@ limitations under the License.
 
 package mock_pgw
 
+/*
 import (
 	"fmt"
 	"net"
@@ -173,3 +174,4 @@ func (mPgw *MockPgw) getHandleCreateBearerRequest() gtpv2.HandlerFunc {
 		return nil
 	}
 }
+*/
--- a/feg/gateway/services/s8_proxy/servicers/s8_proxy_test.go
+++ b/feg/gateway/services/s8_proxy/servicers/s8_proxy_test.go
@@ -15,7 +15,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"os"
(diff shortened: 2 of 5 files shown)


feg/gateway/services/s8_proxy/servicers/mock_pgw/cb.go
```diff
@@ -173,3 +174,4 @@ func (mPgw *MockPgw) getHandleCreateBearerRequest() gtpv2.HandlerFunc {
 		return nil
 	}
 }
+*/
```
uri200, line 177: commented the whole file

feg/gateway/services/s8_proxy/servicers/tft_proto_conversion.go
```diff
@@ -166,3 +167,4 @@ func handlePacketFilterComponent(packetComponentIE *ie.TFTPFComponent) (*oaiprot
 
 	return content, nil
 }
+*/
```
uri200, line 170: commented the whole file

feg/gateway/services/s8_proxy/servicers/s8_proxy_test.go
```diff
@@ -826,9 +823,9 @@ func TestCreateBearerRequest(t *testing.T) {
 		}
 	}
 }
-
+*/
```
uri200, line 826: commented `TestCreateBearerRequest`

themarwhal (approved):

PR #8601, opened 2021-08-11
feat(orc8r): Monitor NMS cert lifespan with alerts

andreilee:

## Summary

Added `nms_operator.pem` certificate to `certifier.yml` to be monitored for certificate lifespan. This is so that alerts can fire when NMS operator certificates are going to expire soon.

## Test Plan

Stood up a local test development environment using Docker. Then checked that the metric for cert lifespan of the `nms_operator.pem` cert would be propagated to Prometheus. The screenshot below shows the metric.

Screen Shot 2021-08-11 at 12 51 15 PM

The alert functionality was not tested, as it has not changed.

## Additional Information

### DIFF
--- a/orc8r/cloud/configs/certifier.yml
+++ b/orc8r/cloud/configs/certifier.yml
@@ -18,6 +18,7 @@ certsDirectory: "/var/opt/magma/certs/"
 orchestratorCerts:
   - "rootCA.pem"
   - "admin_operator.pem"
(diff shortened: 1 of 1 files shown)


hcgatewood (changes requested):
In "normal" deployments this is the same thing as the admin_operator. Double check me but I think the fact that staging has nms_operator separate is legacy

Issue #8494, opened 2021-08-05
techdebt: SessionD unit tests should use full path to access SessionD files


themarwhal:
Subtask of https://github.com/magma/magma/issues/8299

See how this project imports files that it's testing :https://github.com/RobotLocomotion/drake/blob/master/math/test/autodiff_test.cc

PR #8616, opened 2021-08-12
feat(agw): bazel-ify SessionD

themarwhal:
## Summary
* This PR includes all changes necessary to build SessionD and its tests. Now all SessionD unit tests have been bazel-ified.

## Issues and workarounds
* Same issue experienced in the previous PR, SessionD test files do not use the full include path. In order to get around this temporarily, I had to add `strip_include_prefix` to most libraries in SessionD. (https://github.com/magma/magma/issues/8494)
* Folly - Figuring out how to get Folly to build was the hardest part of this PR and I'm giving up slightly for now. Ideally, I would like all of Folly's dependencies to be built with Bazel. For some dependencies like glog and gflags, this was easy to do. For others (boost, libevent, libliberty, etc.), not so much. For now, I've had to result to just installing the Folly static library into the container. This is definitely not ideal, but I'll have to think harder on different strategies. (I've tried using Folly's build script directly in a `genrule`, but that has similar issues. I've tried the build everything from scratch method used in https://github.com/bazelment/trunk/tree/master/third_party , but having issues there as well.)

## Cool improvements / functionalities
1. The build is done inside the bazel-base docker container but the cache lives in the host machine. This means that cache doesn't go away across container restarts and docker restarts. (TODO: Share stats on number of seconds after Docker restart vs vagrant VM restart with the current setup)
2. Tagging tests - bazel supports a way to tag unit tests as flaky. [This allows bazel to try flaky tests 3 times before failing](https://docs.bazel.build/versions/main/be/common-definitions.html#common-attributes-tests). 

## Pending actions for SessionD
1. Adding configs for enabling ASAN, LSAN, etc.
2. Figure out a plan for how the Bazel build would integrate into our existing VM based workflow
3. Verify CLion remote debugger workflow
4. Get information on VSCode / CLion integrations

## Testing with S1AP
I plugged in the SessionD binary built into the local S1AP test setup and all tests passed. 

## Test Plan
Running all tests...
Screen Shot 2021-08-10 at 9 41 43 AM

```
root@08cfd1d8300c:/magma# bazel run //:buildifier
INFO: Analyzed target //:buildifier (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:buildifier up-to-date:
  bazel-bin/buildifier.bash
INFO: Elapsed time: 2.754s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
root@08cfd1d8300c:/magma#
```

## Additional Information


lte/gateway/c/session_manager/test/SessiondMocks.h
```diff
@@ -12,7 +12,6 @@
  */
 #pragma once
 
-#include <folly/io/async/EventBase.h>
```
themarwhal, line 15: Removed as there's no folly dependency here. Another TODO is to maybe integrate IWYU into the bazel configs

WORKSPACE.bazel
```diff
@@ -125,6 +129,28 @@ nlohmann_json()
...
+git_repository(
+    name = "rules_foreign_cc",
+    commit = "a7f1c9e6046a98b6006e183f0112f4815bf2ebc0",
+    remote = "https://github.com/bazelbuild/rules_foreign_cc",
+)
+
+load("@rules_foreign_cc//foreign_cc:repositories.bzl", "rules_foreign_cc_dependencies")
+
+# Don't use preinstalled tools to ensure builds are as hermetic as possible
+rules_foreign_cc_dependencies(register_preinstalled_tools = False)
```
electronjoe: Can you provide more context on this in PR description and or a link in comment or something? I'm not sure what this is doing =D

lte/gateway/c/BUILD.bazel
```diff
@@ -0,0 +1,46 @@
...
+        "-ldl",
+        "-levent",
+        "-ldouble-conversion",
+        "-liberty",
+    ],
+
+    # Unfortunately, aside from glog, gflags, and libfmt we pull all Folly dependencies from the Docker container.
+    # See experimental/bazel-base/Dockerfile for the list of dependencies installed for this purpose - GH8577
+    out_static_libs = ["libfolly.a"],
+    deps = [
```
electronjoe: Do these deps actually get used by the cmake() call? I would be surprised if they do. Does this folly build work if you drop them? Don't want misleading deps if they aren't actually listened to by the cmake handoff.

lte/gateway/c/session_manager/BUILD.bazel
```diff
@@ -0,0 +1,494 @@
...
+)
+
+cc_binary(
+    name = "sessiond",
+    srcs = ["sessiond_main.cpp"],
+    copts = [
+        "-DLOG_WITH_GLOG",
+    ],
+    # Make the binary as static as possible to allow for more reliable behavior when moving binaries around
+    # From bazel doc: this flag makes it so all user libraries are linked statically (if a static version is available),
```
electronjoe, line 483: Thanks for the clarification on what this does! Yeah I think we would always want `linkstatic` on, too bad this is not true by default?

lte/gateway/c/session_manager/test/BUILD.bazel
```diff
@@ -0,0 +1,280 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_cc//cc:defs.bzl", "cc_test")
+
+test_suite(
+    name = "non_flaky_test",
+    tags = ["-flaky"],
```
electronjoe: Flaky-ness at Google was established using the cc_test `flaky` directive documented [here](https://docs.bazel.build/versions/main/be/common-definitions.html#test.flaky). `If set, executes the test up to three times, marking it as failed only if it fails each time. By default, this attribute is set to False and the test is executed only once. Note, that use of this attribute is generally discouraged - tests should pass reliably when their assertions are upheld.`

I would say that if a unit test is very likely to pass with three attempts (per cc_test `flaky` flag above) we should include it in CI.  If not, then maybe a test suite is merited... or we should discard an exceptionally flaky test =D Or fix it.

With this test_suite approach, how does this work? Will Bazel execute all non-flaky test suites across the `bazel test <targets>` so long as the tag is identical? How do we selectively add unit tests in this bazel target?

If we go with this approach, any opinion on `-flaky` versus `-noflakes` or something? With `-flaky` I could see this being either "only run flaky tests" or "no flaky by virtue of the minus sign in front". Not totally clear to the user. WDYT?

lte/gateway/c/session_manager/test/SessiondMocks.h
electronjoe: Something that Bazel tends to help you detect / enforce is actual include dependencies.  It's not bulletproof but ensures you at least re-validate your deps (or it won't build). Maybe worth noting in your prezzo! Not useful for Golang for which this is a non-issue =D Dunno about Python.

third_party/fmtlib.BUILD
```diff
@@ -0,0 +1,16 @@
+# Taken from https://github.com/envoyproxy/envoy/blob/f066fdc261429a1a00d2073fd76fd4d4f58f618e/bazel/external/fmtlib.BUILD
```
electronjoe: Maybe we need a PSC policy to make it easier for reviewers to know what license applies to copied code.

Though this particular segment of Bazel code seems pretty standard issue for anybody who wants to just `build this glob of files`.

electronjoe (approved):
LGTM % comments!

lte/gateway/c/BUILD.bazel
themarwhal: Yes they are actually used. The build doesn't work without these.The libraries get placed in `CMAKE_INCLUDE_PATH` and `CMAKE_LIBRARY_PATH ` so that it can be used in the CMake command

WORKSPACE.bazel
themarwhal: Yeah I think I must have copied and pasted this from an example - will look into what it does and either comment or remove :D

lte/gateway/c/session_manager/test/BUILD.bazel
themarwhal: I completely missed the actual flaky attribute! The running three times is very useful. Will try that out!

themarwhal: resolved - now we use the built in flaky tag

WORKSPACE.bazel
markjen: https://github.com/bazelbuild/rules_foreign_cc/blob/main/docs/README.md#rules_foreign_cc_dependencies

i think setting this flag to False is correct. AFAICT, it tells bazel not to look for pre-installed cmake/ninja on the path. `register_built_tools` (default True) should take over and download/build cmake/ninja inside the bazel sandbox for use.

do we want to set cmake_version or ninja_version here? otherwise, rules_foreign_cc has some defaults (3.21.0/1.10.2)

WORKSPACE.bazel
```diff
@@ -125,6 +129,28 @@ nlohmann_json()
...
+
+### Folly dependencies ###
```
markjen: nit: extra block start/end?

feg/gateway/services/aaa/protos/BUILD.bazel
```diff
@@ -0,0 +1,40 @@
...
+cpp_proto_library(
+    name = "context_cpp_proto",
+    protos = [":context_proto"],
+)
+
+proto_library(
+    name = "context_proto",
+    srcs = ["context.proto"],
+    # TODO(@themarwhal): Clean up proto directory structures to be consistent across the project - GH8625
+    strip_import_prefix = "/feg/gateway/services/aaa/protos",
```
markjen, line 27: do we need this strip? (this proto file doesn't have any imports)

lte/gateway/c/session_manager/BUILD.bazel
```diff
@@ -0,0 +1,494 @@
...
+    # TODO(@themarwhal): Migrate to using full path for includes - GH8494
+    strip_include_prefix = "/lte/gateway/c/session_manager",
+)
+
+cc_library(
+    name = "session_state",
+    srcs = ["SessionState.cpp"],
+    hdrs = ["SessionState.h"],
+    # TODO(@themarwhal): Migrate to using full path for includes - GH8494
+    strip_include_prefix = "/lte/gateway/c/session_manager",
```
markjen, line 199: do these strip_include_prefix args do something?

and should we be preferring full import paths? if so, do we need a strip/includes_prefix or will bazel do the right thing and use paths with workspace dir as the base?

lte/gateway/c/session_manager/test/BUILD.bazel
markjen: would definitely add a comment. I looked at `-flaky` and thought it meant we were asking bazel to pass in a flag and "only run flaky tests"

third_party/fmtlib.BUILD
markjen: PSC? :rofl:

markjen (approved):

third_party/fmtlib.BUILD
electronjoe: Oops, TSC =D Technical Steering Committee (Too much PSC on the brain)

lte/gateway/c/session_manager/test/BUILD.bazel
electronjoe: I think in her new PR (maybe yet unpushed?) This whole test_suite thing is gone and @themarwhal is using Bazel's cc_test builtin for flaky annotations (so no command line flag necessary - will run flaky annotated tests three tries for success).

WORKSPACE.bazel
themarwhal: Yes, I think setting the versions explicitly will be better.

feg/gateway/services/aaa/protos/BUILD.bazel
themarwhal: yes this is for other files importing context.proto. Since accounting.proto includes context.proto without the prefix, we need it here.

lte/gateway/c/session_manager/BUILD.bazel
themarwhal: Yes, so since we don't use full import paths, I've had to add the strip include prefix for now. GH8494 tracks the work for converting paths to be full path. If we fix up all the paths, we shouldn't need any special configuration here and everywhere else.

lte/gateway/c/session_manager/test/BUILD.bazel
themarwhal: Yes, in the updated PR I've removed the `non_flaky_test` target as we might not need it. The built in flaky config that I've added to some of the tests will indicate to Bazel to always run it up to 3 times before indicating failure.

third_party/fmtlib.BUILD
themarwhal: For the third party BUILD files added in this PR, I've simplified them to just be the vanilla include everything + build everything configuration. Now that they are pretty straightforward, I don't think any extra third party licensing comment is necessary here. But this convo still applies to any other BUILD files we may need to borrow from other repos in the future.

hcgatewood (approved):
💯🔥

uri200 (approved):

ardzoht (approved):
lgtm

PR #8626, opened 2021-08-12
fix(agw): GTPV2-C fteid mem align decoding fix

lionelgo:

## Summary

Change the way deserialization of TEID and IPv4 address was done for FTEID IE of GTPV2-C messages.
It was done by assigning a u32 value to a *((uint32_t*) that can lead to a situation where the pointer is misaligned, 
depending of position of these fields in the buffer.
Some undersized GTPV2-C internal arrays have been fixed (possible memory corruption).

## Test Plan

Tested on OAI CUPS deployment with ds Tester Attach/detach scenario.
Wrote Unit tests for testing the de-serialization of some GTPV2-C messages and specifically FTEID IE.

## Additional Information

- The scope of this change is circumvent only in GTPv2c lib and S11 task.


ulaskozat (changes requested):
Unit test is failing.

ulaskozat (approved):
LGTM

PR #8632, opened 2021-08-13
chore: Set default build type of magma-build.sh to be RelWithDebInfo

### LINKED ISSUE #8617: Default build type in `build-magma.sh` is Debug
See https://github.com/magma/magma/blob/d425940a2176343cd57a6072aa3fa3f9bebed234/lte/gateway/release/build-magma.sh#L27

Given that the Debug mode is only intended for light testing / debugging and has major performance issues for SessionD/MME, we should change the default type to RelWithDebInfo

themarwhal:

## Summary
Debug type has (which has ASAN enabled) has performance issues. It would make more sense for RelWithDebInfo to be the default as most people using this script wants a full AGW package that is functional. 

Also clean up debian mentions.

## Additional Information


ardzoht (approved):
lgtm

PR #5381, opened 2021-03-09
[lte][orc8r] Add subscriber API endpoint for paginated loads

mpgermano:
Signed-off-by: Michael Germano <mgermano@fb.com>

## Summary

This PR follows up on configurator entity pagination support by extending that pagination support through the subscriber API. Since the current endpoint response is not sufficient for this support, a new endpoint with version `v2` has been created to allow backwards compatibility and flexibility on the NMS changes that are needed to support pagination.

The main idea here is that a page size and token can be specified to paginate the subscriber load request. Each response will contain the next page token to be used for the subsequent request sent. We also update the subscriber state fetching process to limit the state requested for only the subscribers in the page.

## Test Plan

Added unit test
Tested locally via docker
Tested helm changes via minikube

PR #8640, opened 2021-08-13
feat(agw): paginate enodeb northbound api

mstre123:
feat(agw) paginate listenodeb endpoint

## Summary

Paginated the enodeb endpoint
#8404 
Endpoint now optionally accepts a page size, and page token
Returns a map of enodebs, a page token if there are more to be returned, and an estimated count of the total number of page tokens.

## Test Plan

Tested locally
Tested calls using the swagger ui.
Tested NMS by bringing up the enodeBs page, adding a new enodeb and seeing it populate on the page.


lte/cloud/go/services/lte/obsidian/handlers/handlers.go
```diff
@@ -300,21 +304,41 @@ func listEnodebs(c echo.Context) error {
 	if nerr != nil {
 		return nerr
 	}
+	var pageSize uint64 = 0
+	var err error
```
hcgatewood: No need to explicitly define this here

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
```diff
@@ -300,21 +304,41 @@ func listEnodebs(c echo.Context) error {
...
 
-	ents, _, err := configurator.LoadAllEntitiesOfType(
+	if pageSizeParam := c.QueryParam(ParamPageSize); pageSizeParam != "" {
```
hcgatewood: We're probably going to be doing a lot of parsing page size and page token. Create a helper fn somewhere under orc8r/cloud/go/obsidian, or something like that, that takes a c and returns page size, page token, error

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
```diff
@@ -2145,7 +2145,7 @@ func TestListAndGetEnodebs(t *testing.T) {
 	)
 	assert.NoError(t, err)
 
-	expected := map[string]*lteModels.Enodeb{
+	enodebs := &map[string]*lteModels.Enodeb{
```
hcgatewood: Why are you taking the addr of a map ??

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -1372,13 +1372,15 @@ paths:
         - EnodeBs
       parameters:
         - $ref: './orc8r-swagger-common.yml#/parameters/network_id'
+        - $ref: './orc8r-swagger-common.yml#/parameters/page_size'
```
hcgatewood, line 1375: While we're here, go ahead and refactor all the existing swagger instances/definitions of pagesize/pagetoken to use the common.yml definitions

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -1372,13 +1372,15 @@ paths:
...
+        - $ref: './orc8r-swagger-common.yml#/parameters/page_token'
       responses:
         '200':
-          description: All enodeBs registered in the network
+          description: All enodeBs registered in the network with pagination support
```
hcgatewood: It doesn't actually return "all" enodebs in each response. So I would say "List eNodeBs registered..."

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -1372,13 +1372,15 @@ paths:
...
           schema:
             type: object
             additionalProperties:
-              $ref: '#/definitions/enodeb'
+              $ref: '#/definitions/paginated_enodebs'
```
hcgatewood: This isn't right. The response type is specifically paginated_enodebs. Right now you have ~ `anonymous_object{implicit: map[string]paginated_enodebs}`

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -2611,3 +2613,25 @@ definitions:
...
+    properties:
+      next_page_token:
+        $ref: './orc8r-swagger-common.yml#/definitions/next_page_token'
+      enodebs:
+        additionalProperties:
+          $ref: '#/definitions/enodeb'
+          x-nullable: true
+        type: object
+      total_count:
+        description: estimated total number of enodebs
```
hcgatewood: Sentence case

orc8r/cloud/go/models/swagger-common.yml
```diff
@@ -121,6 +121,10 @@ definitions:
     description: The type of the network
     type: string
     example: LTE
+  next_page_token:
```
hcgatewood: "next" page token is just a page token, so no need to include the `next_`

orc8r/cloud/go/models/swagger-common.yml
```diff
@@ -121,6 +121,10 @@ definitions:
...
+    type: string
+    description: Base 64 encoded page token for subsequent paginated API requests
```
hcgatewood:
```suggestion
    description: Base64-encoded page token for subsequent paginated API requests
```

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -2611,3 +2613,25 @@ definitions:
...
+
+  paginated_enodebs:
+    description: Page of enodebs
+    properties:
+      next_page_token:
+        $ref: './orc8r-swagger-common.yml#/definitions/next_page_token'
+      enodebs:
+        additionalProperties:
+          $ref: '#/definitions/enodeb'
+          x-nullable: true
```
hcgatewood, line 2629: 1. Nullable=true is the default
2. Idk why we would need to make this nullable. I believe we can just remove this line to no effect (and also remove it from subs_v2)

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -2611,3 +2613,25 @@ definitions:
         pattern: '^(\d{6})$'
         example: '176148'
         x-nullable: false
+
+  paginated_enodebs:
+    description: Page of enodebs
```
hcgatewood: Nit: be consistent about whether you're calling them "eNodeBs" (full/correct capitalization) or "enodebs" (simple/economical capitalization)

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
```diff
@@ -2219,14 +2225,45 @@ func TestListAndGetEnodebs(t *testing.T) {
 	}
 	tests.RunUnitTest(t, e, tc)
 
+	paginatedExpectation := &lteModels.PaginatedEnodebs{
+		Enodebs:       map[string]*lteModels.Enodeb{"abcdefg": expected.Enodebs["abcdefg"]},
+		NextPageToken: "CgdhYmNkZWZn",
```
hcgatewood: Any time we have values like this that need to be manually computed out-of-band (e.g. the jittered state sync values), place them into a const with descriptive name and likely a short comment about how you generated the value

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
```diff
@@ -300,21 +304,41 @@ func listEnodebs(c echo.Context) error {
...
+	enodebs := make(map[string]*lte_models.Enodeb, len(ents))
 	for _, ent := range ents {
-		ret[ent.Key] = (&lte_models.Enodeb{}).FromBackendModels(ent)
+		enodebs[ent.Key] = (&lte_models.Enodeb{}).FromBackendModels(ent)
 	}
-	return c.JSON(http.StatusOK, ret)
+	paginatedEnodebs := &lte_models.PaginatedEnodebs{
+		Enodebs:       enodebs,
+		NextPageToken: lte_models.NextPageToken(nextPageToken),
+		TotalCount:    int64(len(enodebs)),
```
hcgatewood: This isn't quite right. TotalCount is meant to be an estimate of how many of the object exist "in total", not just how many objects were returned on this page. Fix this and then add a regression test to display the proper behavior

lte/cloud/helm/lte-orc8r/Chart.yaml
```diff
@@ -13,7 +13,7 @@ apiVersion: v2
 appVersion: "1.0"
 description: A Helm chart for magma orchestrator's lte module
 name: lte-orc8r
-version: 0.2.5
+version: 0.2.6
```
hcgatewood: No need to bump the helm chart version -- we didn't change anything that would require redeploying/upgrading the charts, so all we would need to do is upgrade the container image version

Side note, this change is meant to be backwards-compatible (up to the default page size)

hcgatewood (changes requested):
Cool shape is looking correct. The other thing we need to do here is verify whether this is actually backwards compatible

To do this, you'll want to spin up a local NMS instance and try to list subscribers etc. Connect with @andreilee for pointers on getting that set up, as well as which NMS page to visit to verify this

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
mstre123: this was a pattern used in the reference pr
https://github.com/magma/magma/pull/5381/files#diff-ccdd661b3d08b3d70276e71041e28714a15c9e8c1e10cf260b162a84af292b43R211-R219

im fine with refactoring this but i think we should conform to a pattern throughout the apis when calculating page sizes.
that said will pull this out into its own helper to make that easier going forward

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
mstre123: agreed will do

orc8r/cloud/go/models/swagger-common.yml
mstre123: was following convention in subscriber api
https://github.com/magma/magma/pull/5381/files#diff-68151344ac75147d32963d9707aa36d7b701da45e0c65b90a8931aa4d7975249R378
I'm fine with removing the next prefix though

hcgatewood: So there's a difference between the an object's field name vs. an object definition

When we're defining what a page token is (object definition), it's irrelevant whether it's the "next" page token. When we're adding a "next page token" field to e.g. a return object (object's field name), then it is indeed relevant that the page token is the "next" one

The line you linked is the field name, while the line I commented on here is the object definition

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
```diff
@@ -300,21 +300,35 @@ func listEnodebs(c echo.Context) error {
 	if nerr != nil {
 		return nerr
 	}
+	pageSize, pageToken, err := obsidian.GetPageSizeAndTokenParams(c)
```
hcgatewood: Nit: functions with the word "and" in them are clunky. Try `GetPaginationParams`

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
```diff
@@ -300,21 +300,35 @@ func listEnodebs(c echo.Context) error {
...
-	ents, _, err := configurator.LoadAllEntitiesOfType(
+	ents, nextPageToken, err := configurator.LoadAllEntitiesOfType(
 		nid, lte.CellularEnodebEntityType,
-		configurator.EntityLoadCriteria{LoadMetadata: true, LoadConfig: true, LoadAssocsToThis: true},
+		configurator.EntityLoadCriteria{
+			LoadMetadata:     true,
+			LoadConfig:       true,
+			LoadAssocsToThis: true,
+			PageSize:         uint32(pageSize),
+			PageToken:        pageToken},
```
hcgatewood: General: move this closing brace to its own line

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
```diff
@@ -300,21 +300,35 @@ func listEnodebs(c echo.Context) error {
...
+	enodebs := make(map[string]*lte_models.Enodeb, len(ents))
 	for _, ent := range ents {
-		ret[ent.Key] = (&lte_models.Enodeb{}).FromBackendModels(ent)
+		enodebs[ent.Key] = (&lte_models.Enodeb{}).FromBackendModels(ent)
 	}
-	return c.JSON(http.StatusOK, ret)
+	paginatedEnodebs := &lte_models.PaginatedEnodebs{
+		Enodebs:    enodebs,
+		PageToken:  lte_models.PageToken(nextPageToken),
+		TotalCount: int64(len(enodebs)),
```
hcgatewood: As mentioned, this is the wrong count

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
```diff
@@ -300,21 +300,35 @@ func listEnodebs(c echo.Context) error {
...
 	)
 	if err != nil {
 		return obsidian.HttpError(err, http.StatusInternalServerError)
 	}
 
-	ret := make(map[string]*lte_models.Enodeb, len(ents))
+	enodebs := make(map[string]*lte_models.Enodeb, len(ents))
 	for _, ent := range ents {
-		ret[ent.Key] = (&lte_models.Enodeb{}).FromBackendModels(ent)
+		enodebs[ent.Key] = (&lte_models.Enodeb{}).FromBackendModels(ent)
```
hcgatewood, line 335: From offline: it looks like we use map of key->ptr everywhere. If that's the case, then it's fine to leave the "xnullable: true" in the IDL defn, just add a comment to that IDL line with why we have it there

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -271,9 +271,9 @@ func listSubscribersV2Handler(c echo.Context) error {
 		return c.JSON(http.StatusOK, nil)
 	}
 	paginatedSubs := subscribermodels.PaginatedSubscribers{
-		TotalCount:    int64(count),
-		NextPageToken: subscribermodels.NextPageToken(nextPageToken),
-		Subscribers:   subs,
+		TotalCount:  int64(count),
+		PageToken:   subscribermodels.PageToken(nextPageToken),
```
hcgatewood: Following up on my comment from elsewhere

```suggestion
		NextPageToken:   subscribermodels.PageToken(nextPageToken),
```

orc8r/cloud/go/obsidian/handler.go
```diff
@@ -322,3 +327,18 @@ func NetworkIdHttpErr() *echo.HTTPError {
...
+// GetPageSizeAndTokenParams takes in context and returns page size and token params.
+func GetPageSizeAndTokenParams(c echo.Context) (uint64, string, error) {
+	pageSizeParam := c.QueryParam(ParamPageSize)
+	pageTokenParam := c.QueryParam(ParamPageToken)
+	if pageSizeParam == "" {
+		return 0, pageTokenParam, nil
+	}
+	pageSize, err := strconv.ParseUint(pageSizeParam, 10, 32)
+	if err != nil {
+		err := fmt.Errorf("invalid page size parameter: %s", err)
```
hcgatewood: errors.Wrap

orc8r/cloud/go/obsidian/handler.go
```diff
@@ -322,3 +327,18 @@ func NetworkIdHttpErr() *echo.HTTPError {
 func TenantIdHttpErr() *echo.HTTPError {
 	return HttpError(fmt.Errorf("Missing Tenant ID"), http.StatusBadRequest)
 }
+
+// GetPageSizeAndTokenParams takes in context and returns page size and token params.
+func GetPageSizeAndTokenParams(c echo.Context) (uint64, string, error) {
+	pageSizeParam := c.QueryParam(ParamPageSize)
```
hcgatewood: "pageSizeStr"

orc8r/cloud/go/obsidian/handler.go
```diff
@@ -322,3 +327,18 @@ func NetworkIdHttpErr() *echo.HTTPError {
...
+	pageTokenParam := c.QueryParam(ParamPageToken)
```
hcgatewood: "pageToken"

hcgatewood (changes requested):
Nearly there. Also from offline: give this a manual inspection from the NMS to ensure it's backward compatible. cc @andreilee

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
mstre123: Do we have a team formatter/linter for stuff like this?

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
```diff
@@ -300,21 +300,46 @@ func listEnodebs(c echo.Context) error {
...
 	}
+	pageSize, pageToken, err := obsidian.GetPaginationParams(c)
+	if err != nil {
+		return err
+	}
 
-	ents, _, err := configurator.LoadAllEntitiesOfType(
-		nid, lte.CellularEnodebEntityType,
-		configurator.EntityLoadCriteria{LoadMetadata: true, LoadConfig: true, LoadAssocsToThis: true},
+	count, err := configurator.CountEntitiesOfType(
```
hcgatewood: While we're here: in an immediately subsequent PR, let's fix CountEntitiesOfType to not take a load criteria or serdes -- and push that change down through the servicer as well

Since those fields aren't necessary to count

hcgatewood: Nit: do the load first, then the count

mstre123:
> Nearly there. Also from offline: give this a manual inspection from the NMS to ensure it's backward compatible. cc @andreilee

@andreilee is going to land a pr on top of this one before merging to handle changes in the NMS to support this change.
Tested swagger endpoints to confirm desired behavior.

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -2611,3 +2611,26 @@ definitions:
...
+
+  paginated_enodebs:
+    description: Page of eNodeBs
+    properties:
+      page_token:
+        $ref: './orc8r-swagger-common.yml#/definitions/page_token'
+      enodebs:
+        additionalProperties:
+          $ref: '#/definitions/enodeb'
+          # The x-nullable flag is set to true to generate a map of pointers
```
hcgatewood, line 2628: Add this comment to the subsv2 paginated_subs as well

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -2611,3 +2611,26 @@ definitions:
...
+          x-nullable: true
+        type: object
+      total_count:
+        description: Estimated total number of eNodeBs
+        example: 10
+        format: int64
```
hcgatewood: Feels like this should be uint64

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -2611,3 +2611,26 @@ definitions:
...
+        type: integer
+        x-nullable: false
+    required:
```
hcgatewood: Move these fields up -- "properties" should be the last property at that level

orc8r/cloud/go/obsidian/handler.go
```diff
@@ -322,3 +327,18 @@ func NetworkIdHttpErr() *echo.HTTPError {
 func TenantIdHttpErr() *echo.HTTPError {
 	return HttpError(fmt.Errorf("Missing Tenant ID"), http.StatusBadRequest)
 }
+
+// GetPaginationParams takes in context and returns page size and token params.
```
hcgatewood: Nit: "takes in a context" is redundant because it's clear from the function signature

hcgatewood (approved):
Lgtm! From offline: rc the nms issue, then msg me to merge

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
hcgatewood: Yes, but a quick glance doesn't reveal any way to lint for this specific issue. Seems like maybe https://github.com/mvdan/gofumpt would support but unclear, you can try enabling this via our golangci linter

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
mstre123: looks like the param we take in is a uint so probably the right call
subscriberv2 uses int64 though
https://github.com/magma/magma/blob/fad4a1e4a6f28c699f741e06b0fd5b886662cc6c/lte/cloud/go/services/subscriberdb/obsidian/models/swagger.v1.yml#L403

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
mstre123: > While we're here: in an immediately subsequent PR, let's fix CountEntitiesOfType to not take a load criteria or serdes -- and push that change down through the servicer as well
> 
> Since those fields aren't necessary to count

isn't their value to load criteria if future callers want to add filters to their count call?

mstre123: ah never mind took a closer look it just adds things to the response that gets dropped - we can remove it in the next pr

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
hcgatewood: Feel free to fix both 🙂

mstre123: i'll put that up in its own pr - adding comments or cleaning typos i come across is one thing but i'd prefer not to make logical changes to subscriber here.

andreilee (approved):
Thanks for getting this working on NMS as well

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -1372,13 +1372,13 @@ paths:
         - EnodeBs
       parameters:
         - $ref: './orc8r-swagger-common.yml#/parameters/network_id'
+        - $ref: './orc8r-swagger-common.yml#/parameters/page_size'
+        - $ref: './orc8r-swagger-common.yml#/parameters/page_token'
       responses:
         '200':
-          description: All enodeBs registered in the network
+          description: List eNodeBs registered registered in the network with pagination support
```
hcgatewood: "registered registered"

PR #8641, opened 2021-08-13
feat(s8_proxy): add delete bearer request support

### LINKED ISSUE #8445: [feg] add Delete Bearer Request

uri200:

## Summary

Add support for delete bearer request

## Test Plan
Extended unit test 
./build.py -c

## Additional Information


feg/gateway/services/s8_proxy/servicers/senders.go
```diff
@@ -93,3 +93,18 @@ func (s *S8Proxy) sendAndReceiveCreateBearerResponse(res *protos.CreateBearerRes
...
 	return &orc8r_protos.Void{}, err
 }
+
+// sendAndReceiveDeleteBearerResponse
+func (s *S8Proxy) sendAndReceiveDeleteBearerResponse(res *protos.DeleteBearerResponsePgw,
+	cPgwUDPAddr *net.UDPAddr,
+	cbResMsg message.Message) (*orc8r_protos.Void, error) {
+	glog.V(2).Infof("Send Delete Bearer Response (grpc) to %s:\n%s", cPgwUDPAddr,
+		res)
+	glog.V(2).Infof("Send Delete Bearer Response (gtp) to %s:\n%s",
```
themarwhal, line 103: Why do we need both logs?

themarwhal (approved):

feg/gateway/services/s8_proxy/servicers/senders.go
uri200: this is because go-gtp doesn't show all the content. Since those messages are new I prefer to have the plain value (grpc) and the encoded value (gtp) visible for easy debugging. 

Note this is only visible on level 2

PR #8656, opened 2021-08-17
feat: Domain proxy helm chart

m-trojanowski:
Signed-off-by: Marcin Trojanowski <marcin.trojanowski@freedomfi.com>

## Summary

- Create dp directory to host helm charts for domain-proxy component.
- Domain proxy chart requires ingress controller to be present on Cluster.

## Test Plan

 Install chart using one of provided example values.
 All tests should pass.

## Additional Information

### DIFF
--- a/dp/cloud/helm/dp/Chart.yaml
+++ b/dp/cloud/helm/dp/Chart.yaml
@@ -1,0 +1,67 @@
+---
+# Copyright 2020 The Magma Authors.
+
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v2
+appVersion: "0.1.0"
+version: 0.1.0
+description: A Helm chart for magma orchestrator's domain-proxy module.
+name: domain-proxy
+engine: gotpl
+sources:
+  - https://github.com/magma/magma
+keywords:
+  - magma
+  - or8cr
+  - dp
+type: application
+
+long_description: |
+  This Chart will deploy the following:
+
+  - 1 x Configuration controller.
+  - 1 x protocol controller with 80/443 TCP port exposed with either ingress/contour ingress resource.
+  - 1 x radio controller with GRPC 50053/TCP port.
+  - All using Kubernetes Deployment
+  - 1 x job for Postgres database migration.
+
+  ## Installation.
+
+  ### Preprequsites.
+
+  Domain proxy requires an ingress controller to be present on targer kubernetes cluster.
+  Standard ingress resource is supported as well as countour httpproxy CRD.
+  (https://projectcontour.io/docs/v1.16.0/config/fundamentals/)
+
+  ### From Source.
+
+  ```bash
+  git clone
+  cd dp/cloud/helm/dp
+  helm dep update domain-proxy
+  helm install --name myname --namespace mynamespace domain-proxy
+  ```
+
+  ### Certificates.
+
+  In order to work properly Domain proxy requires set of certificates. To enable chart to consume them place them inside.
+  `certificates` directory and setup proper certificate paths in your `values.yaml` file.
+
+  ## Development.
+
+  ### Local Deployment using Minikube.
+  If you're running locally in Minikube, see the `examples/minikube_values.yml` file.
+
+  To run local development environment:
+
+  ```bash
+  make
+  ```
--- a/dp/cloud/helm/dp/README.md
+++ b/dp/cloud/helm/dp/README.md
@@ -1,0 +1,186 @@
+
+Domain-proxy
+===========
+
+A Helm chart for magma orchestrator's domain-proxy module.
+
+This Chart will deploy the following:
+
+- 1 x Configuration controller.
+- 1 x protocol controller with 80/443 TCP port exposed with either ingress/contour ingress resource.
+- 1 x radio controller with GRPC 50053/TCP port.
+- All using Kubernetes Deployment
+- 1 x job for Postgres database migration.
+
+## Installation.
+
+### Preprequsites.
+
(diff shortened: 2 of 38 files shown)


dp/cloud/helm/dp/Chart.yaml
```diff
@@ -0,0 +1,67 @@
...
+  - 1 x protocol controller with 80/443 TCP port exposed with either ingress/contour ingress resource.
+  - 1 x radio controller with GRPC 50053/TCP port.
+  - All using Kubernetes Deployment
+  - 1 x job for Postgres database migration.
+
+  ## Installation.
+
+  ### Preprequsites.
+
+  Domain proxy requires an ingress controller to be present on targer kubernetes cluster.
```
hcgatewood, line 40: Orc8r currently doesn't have an ingress controller. This is something I've been wanting to move toward, specifically I'm interested in trying Ambassador since it will enable our module-style service mesh architecture

What's your current plan for adding an ingress controller? I see nginx and contour -- are those the choices, and why?

dp/cloud/helm/dp/README.md
```diff
@@ -0,0 +1,186 @@
...
+`certificates` directory and setup proper certificate paths in your `values.yaml` file.
+
+## Development.
+
+### Local Deployment using Minikube.
+If you're running locally in Minikube, see the `examples/minikube_values.yml` file.
+
+To run local development environment:
+
+```console
```
hcgatewood: General:
```suggestion
```bash
```

dp/cloud/helm/dp/README.md
```diff
@@ -0,0 +1,186 @@
...
+$ make
```
hcgatewood: General:
```suggestion
make
```

dp/cloud/helm/dp/README.md
```diff
@@ -0,0 +1,186 @@
...
+| `db_service.image.tag` | Overrides the image tag whose default is the chart appVersion. | `""` |
+| `db_service.imagePullSecrets` | Name of the secret that contains container image registry keys | `[]` |
+| `db_service.serviceAccount.create` | Specifies whether a service account should be created. | `false` |
+| `db_service.serviceAccount.annotations` | Annotations to add to the service account. | `{}` |
+| `db_service.serviceAccount.name` | The name of the service account to use,If not set and create is true, a name is generated using the fullname template. | `""` |
+
+
+
+---
+_Documentation generated by [Frigate](https://frigate.readthedocs.io)._
```
hcgatewood, line 185: 👍 

Ideally this would be generated automatically as part of the `make gen` cloud job

dp/cloud/helm/dp/README.md
```diff
@@ -0,0 +1,186 @@
...
+
```
hcgatewood, line 186: Why are we adding the .gitkeep files? Can you point to where the directories' existence needs to exist a priori

dp/cloud/helm/dp/examples/minikube_values_nginx.yaml
```diff
@@ -0,0 +1,248 @@
+---
```
hcgatewood, line 1: These example values.yaml files feel quite redundant. Is there a way to reduce the redundancy across the examples?

dp/cloud/helm/dp/README.md
```diff
@@ -0,0 +1,186 @@
...
+A Helm chart for magma orchestrator's domain-proxy module.
+
+This Chart will deploy the following:
+
+- 1 x Configuration controller
+- 1 x protocol controller with 80/443 TCP port exposed with either ingress/contour ingress resource.
+- 1 x radio controller with GRPC 50053/TCP port.
+- All using Kubernetes Deployment
+- 1 x job for Postgres database migration.
```
hcgatewood, line 13: Can you point me to the migration / what specifically is occurring -- is it just initializing the relevant tables?

hcgatewood (changes requested):
Two main things here

1. The core Orc8r Helm charts are already a mess, so for now I say y'all get to own this, and I don't have much input on style etc -- so this is a stamp from me
2. This PR could be much slimmer if we followed the monolith approach -- i.e. there's no reason to split these components into separate physical services. However, since they currently are split, we can go with this topology for now

At some point I'll get around to revamping the cloud's Helm charts, at which point we can take a look at the DP charts as well. But for now, I'll treat this as domain-specific functionality, so lgtm

---

Post note: the main thing here I'd like request discussion on is the choice of ingress controller -- please set up a sync with me so we can discuss. After that I'll be ready to stamp 🙂

dp/cloud/helm/dp/Chart.yaml
m-trojanowski: Hi @hcgatewood ,
nginx was picked because of it's widespread adoption, good documentation, and big community. Contour on the other hand underneath uses envoy proxy, which is also used by ambassador, which we can later move on. DP expects ingress controller to be present, and I think that whether it should be enabled for whole magma project is a different, much bigger discussion. That being said, adding it on fresh kubernetes cluster is a matter of update existing terraform code, which can be done in separate PR. Of course I can prepare also config for current nginx pod solution.

dp/cloud/helm/dp/README.md
m-trojanowski: Fixed

dp/cloud/helm/dp/README.md
m-trojanowski: Fixed

dp/cloud/helm/dp/README.md
m-trojanowski: Fixed.
.gitkeeps are added to commit empty directories,so user don't have to create them by himself,but of course I can remove them and include this step in documentation instead.

dp/cloud/helm/dp/examples/minikube_values_nginx.yaml
m-trojanowski: So we're using those files in our CI, to test both ingress controllers. I do realize that they are redundant,and I think that I can reduce them to single file with proper section commented out.

dp/cloud/helm/dp/README.md
m-trojanowski: Sure, 
we have specific service which is responsible for running database migrations. You can find it here:
https://github.com/magma/domain-proxy/tree/master/db_service
Basically it's just initialize tables required by DP.

PR #8664, opened 2021-08-17
feat(agw): Paginate list gateways api.

mstre123:
feat(agw): Paginate list gateways api.
## Summary
Updated the List Gateways API to support pagination
#8403 

## Test Plan
Tested Locally
Tested Swagger Endpoints
Tested Local NMS


hcgatewood (changes requested):
From offline: looks like this can be implemented using the existing LoadAllEntitiesOfType, which itself is already paginated. Try that and lmk

hcgatewood (approved):
💯💯

andreilee (approved):

BiffaloBuff:
Did this change break the `register_vm` fabric command used in the Quick Start Guide? [dev_utils.py line 220](https://github.com/magma/magma/blob/6d359769e360182cd54a0e6e84d6c7f10ad3e49c/orc8r/tools/fab/dev_utils.py#L220) is still expecting the old json structure.

mstre123:
> Did this change break the `register_vm` fabric command used in the Quick Start Guide? [dev_utils.py line 220](https://github.com/magma/magma/blob/6d359769e360182cd54a0e6e84d6c7f10ad3e49c/orc8r/tools/fab/dev_utils.py#L220) is still expecting the old json structure.

good catch - will put up a pr to fix that
#9278

PR #8669, opened 2021-08-18
feat(nms): Add subscriber bulk delete

HannaFar:
## Summary

Bulk delete subscribers

https://user-images.githubusercontent.com/26038920/130644863-bf2cd59f-6c16-40fc-ad8c-f153e15fcfbd.mov

## Additional Information

### DIFF
--- a/nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
+++ b/nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
@@ -13,6 +13,7 @@
  * @flow strict-local
  * @format
  */
+import type {WithAlert} from '@fbcnms/ui/components/Alert/withAlert';
 import type {mutable_subscribers, subscriber} from '@fbcnms/magma-api';
 
 import ActionTable from '../../components/ActionTable';
@@ -41,6 +42,7 @@ import Tab from '@material-ui/core/Tab';
 import Tabs from '@material-ui/core/Tabs';
 import TypedSelect from '@fbcnms/ui/components/TypedSelect';
 import nullthrows from '@fbcnms/util/nullthrows';
+import withAlert from '@fbcnms/ui/components/Alert/withAlert';
 
 import {
   AltFormField,
@@ -154,7 +156,7 @@ function parseSubscriber(line: string) {
     authOpc: items[SUB_AUTH_OPC_OFFSET],
     state: items[SUB_STATE_OFFSET] === 'ACTIVE' ? 'ACTIVE' : 'INACTIVE',
     dataPlan: items[SUB_DATAPLAN_OFFSET],
-    apns: items[SUB_APN_OFFSET].split('|')
+    apns: items[SUB_APN_OFFSET]?.split('|')
       .map(item => item.trim())
       .filter(Boolean),
     policies: items?.[SUB_POLICY_OFFSET]?.split('|')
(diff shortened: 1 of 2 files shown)


nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -163,11 +163,11 @@ function parseSubscriber(line: string) {
   };
 }
 
-function parseSubscriberFile(fileObj: File) {
+function parseSubscriberFile(props: {fileObj: File, delete: boolean}) {
```
andreilee:
```suggestion
function parseSubscriberFile(fileObj: File, delete: boolean) {
```

Don't use an object as parameter to simulate named parameters - this is a pattern that we haven't used throughout the NMS, and this is more specific to React components.

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -243,6 +246,75 @@ export function EditSubscriberButton(props: EditProps) {
   );
 }
 
+export function BulkDeleteSubscriberButton(props: {onClose: () => void}) {
```
andreilee: nit: Create a type for the `Props` of this component

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -243,6 +246,75 @@ export function EditSubscriberButton(props: EditProps) {
...
+        enqueueSnackbar('error deleting ' + subscriber.imsi + ' : ' + errMsg, {
+          variant: 'error',
+        });
+        return;
+      }
+    }
+    enqueueSnackbar(
+      ` Subscriber${
+        successCountRef.current > 1 ? 's ' : ''
+      } deleted successfully`,
```
andreilee:
```suggestion
      'Subscriber(s) deleted successfully',
```

nit: This is just an opinion of mine, but I don't think the extra complexity is worth it here for things like this.

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -243,6 +246,75 @@ export function EditSubscriberButton(props: EditProps) {
...
+            try {
+              const newSubscribers = await parseSubscriberFile({
+                fileObj: fileInput.current?.files[0],
+                delete: true,
+              });
+              deleteSubscribers(newSubscribers);
+            } catch (e) {
+              enqueueSnackbar(e, {
+                variant: 'error',
+              });
```
andreilee, line 306: We have a lot of problems with our snackbar error messages being rather cryptic to non-developers. If you can provide some easy to understand error messages for different failure modes here, that would improve the usability quite a bit.

andreilee (changes requested):
Looks good on this first-draft pass. Something that I think we should emphasize more is making sure that our error messages are more helpful to users.

Make sure to include enough screenshots in the description so the entire user flow can be understood.

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
HannaFar: I think we should do this in another PR because it seems that we don't handle all the errors (for example deleting a non existing subscriber)

nms/app/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -864,16 +947,10 @@ function AddSubscriberDetails(props: DialogProps) {
   );
 }
 
-function validateSubscriberInfo(
-  info: SubscriberInfo,
-  subscribers: {[string]: subscriber},
```
HannaFar, line 869: This is not useful with paginated subscribers

Issue #8589, opened 2021-08-10
[NMS] Modify make gen to update NMS MagmaAPIBindings.js


andreilee:
Currently `MagmaAPIBindings.js` is updated through a shell script under the `nms` subdirectory. This file should be updated whenever the `Orc8r` is built. Additionally, CI should block merging if this is not updated when the swagger files are updated.

PR #8681, opened 2021-08-19
chore(nms): Trigger cwag gh actions on nms changes

andreilee:
## Summary

Opening this PR more as a point of discussion - I noticed that neither `cwag-build` nor the `cwag-precommit` Github checks are ever reported on pull requests that only include NMS changes. See #8589 and #8666 for reference.

Not sure if this is the best way to fix this though.

## Test Plan

Absolutely none

## Additional Information


andreilee:
Closing this because I have no idea why this includes four commits

PR #8718, opened 2021-08-23
fix: Remove as_dict override for spyne

amarpad:
## Summary
The underlying memory leak has been fixed in spyne so remove
override in the subclass

## Test Plan

make python_test

## Additional Information


ardzoht (approved):
lgtm

ardzoht:
Looks like `lte_test` is failing on pylint tests for enodebd, I think this function https://github.com/magma/magma/blob/master/lte/gateway/python/magma/enodebd/enodeb_status.py#L555 is missing encoding parameter, not sure how it got triggered till now though.

PR #8724, opened 2021-08-23
fix(orc8r): Fix sub sync compatibility issue with v1.6

hcgatewood:
## Summary

v1.6 only uses the ListSubscribers endpoint for checking whether subs are in/out of date. Master uses a 3-tiered approach (CheckInSync, Sync, ListSubscribers). In order to retain backward compatibility, the ListSubscribers endpoint must be backward compatible.

This PR fixes an oversight which caused a compatibility breakage.

v1.6 definition of ListSubscribers:

```proto
message ListSubscribersResponse {
  repeated SubscriberData subscribers = 1;
  // next_page_token is a serialized entity page token for subsequent paginated
  // loads.
  string next_page_token = 2;
  // flat_digest is the server-defined, deterministic digest of the full set of
  // returned subscriber objects.
  Digest flat_digest = 3;
  // per_sub_digests is a set of per-subscriber digests, indexed by the
  // subscriber IDs (IMSI).
  map<string, Digest> per_sub_digests = 4;
}
```

## Additional Information


uri200 (approved):

lte/protos/subscriberdb.proto
```diff
@@ -255,5 +255,8 @@ message ListSubscribersResponse {
   // loads.
   string next_page_token = 2;
   // digests contains all digests for the network.
-  magma.orc8r.DigestTree digests = 3;
+  magma.orc8r.DigestTree digests = 5;
+
+  // Deprecated
+  reserved 3, 4;
```
ulaskozat, line 261: I think this pattern does not preserve the backward compatibility. Can we instead use deprecated annotation?

hcgatewood: From offline: I believe it does preserve backward compat, using reserved because the previous type no longer exists

PR #8728, opened 2021-08-23
fix(mme): Fix enable_converged_core flag for headless ops

ulaskozat:

## Summary

- Local MME override configuration for `enable_converged_core` was not working when older builds of orc8r are used.
- Similarly the existing default flag for congestion control in mme.yml file has precedence negating the orc8r pushed configs if any. Removed it from mme.yml.

## Test Plan

Test on SIT with older orc8r with various local overrides.

## Additional Information


pshelar (approved):
LGTM

ardzoht (approved):
lgtm

PR #8737, opened 2021-08-24
feat(amf): T3592 Timer support

Kaleem-Wavelabs:
Signed-off-by: Kaleemullah Mohammed <kaleemullah.mohammed@wavelabs.ai>

Support for PDU session release timer 3592

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/amf/amf_app_timer_management.cpp
+++ b/lte/gateway/c/core/oai/tasks/amf/amf_app_timer_management.cpp
@@ -10,8 +10,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-
-#include "amf_app_timer_management.h"
 //--C includes -----------------------------------------------------------------
 extern "C" {
 #include "log.h"
@@ -19,6 +17,7 @@ extern "C" {
 #include "intertask_interface.h"
 #include "common_types.h"
 }
+#include "amf_app_timer_management.h"
 //--C++ includes ---------------------------------------------------------------
 #include <stdexcept>
 //--Other includes -------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


lte/gateway/c/core/oai/tasks/amf/amf_app_timer_management.cpp
```diff
@@ -10,15 +10,14 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-
-#include "amf_app_timer_management.h"
 //--C includes -----------------------------------------------------------------
 extern "C" {
 #include "log.h"
 #include "conversions.h"
 #include "intertask_interface.h"
 #include "common_types.h"
 }
+#include "amf_app_timer_management.h"
```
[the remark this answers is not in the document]
ulaskozat: how is this different than mme_app_timer_management ?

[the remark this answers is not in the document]
Kaleem-Wavelabs: mme_app_timer_management defined class is placed as part of MME library, not in the common library. Also the class MmeUeContext makes use of mme_app_task_zmq_ctx zmq task, not in generic nature.       
This amf specific utility amf_app_timer_management, apart from ue context lookup also includes map for pdu session lookup:
 std::map<int, ue_pdu_id_t> amf_pdu_timers; 
typedef struct ue_pdu_id {
  uint8_t ue_id;
  uint8_t pdu_id;
} ue_pdu_id_t;

lte/gateway/c/core/oai/tasks/amf/amf_app_timer_management.cpp
```diff
@@ -72,4 +71,52 @@ bool AmfUeContext::GetTimerArg(const int timer_id, timer_arg_t* arg) const {
...
+
+//------------------------------------------------------------------------------
+int AmfUeContext::StartPduTimer(
+    size_t msec, timer_repeat_t repeat, zloop_timer_fn handler,
+    ue_pdu_id_t arg) {
+  int timer_id = -1;
+  if ((timer_id = start_timer(
+           &amf_app_task_zmq_ctx, msec, repeat, handler, nullptr)) != -1) {
+    // amf_pdu_timers.insert(timer_id, std::make_pair(ue_id, pdu_id));
+    // amf_pdu_timers.insert(timer_id, arg);
```
ulaskozat: remove if not used

lte/gateway/c/core/oai/tasks/amf/amf_app_timer_management.cpp
```diff
@@ -72,4 +71,52 @@ bool AmfUeContext::GetTimerArg(const int timer_id, timer_arg_t* arg) const {
...
+    amf_pdu_timers[timer_id] = arg;
```
ulaskozat, line 99: This pattern was copied from 4G, but I think it was an lack of proper oversight by me. We do not need to have a separate data structure to pack the arguments for the handler. `start_timer` has the last parameter for packing the arguments for the handler function. Instead use that facility.

Kaleem-Wavelabs: In our debug session with timers, we have observed that underlying zmq library function "zloop_timer" is not reliable for the last argument void * when it expires, meaning we have seen when the timer expires it points to a invalid data (sometimes seen coredumps when trying to fetch void * arg when timer expires). So we added internal mapping structure based on the timer-id to be able to lookup the correct context when timer expires. Let me know if this approach is ok.

lte/gateway/c/core/oai/tasks/amf/amf_app_timer_management.cpp
Kaleem-Wavelabs: ok, will remove the same.

lte/gateway/c/core/oai/tasks/amf/amf_app_timer_management.cpp
ulaskozat: are you sure that it is the zmq issue? You need to calloc for the args and deep-copy content into it. when timer expires, you need to free its content. If you have some data passed by reference to args where the owner is someone else (not the handler), then you would face such issues.

Kaleem-Wavelabs: Yes @ulaskozat with this approach we were able to fetch the timer arg properly when expires, but we had apprehension that it might lead to memory leaks in the cases where timer do not expires as the zmq API zloop_timer_end do not mention any thing about arg parameter.

ulaskozat: this is a valid point. It is better to assume the worst and not rely on proper behavior on zloop_timer_end side. Though, you can write a unit test and check this.

ulaskozat: anyhow, for now let's leave this as is...

lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
```diff
@@ -196,6 +200,147 @@ void clear_amf_smf_context(smf_context_t* smf_ctx) {
...
+    OAILOG_ERROR(
+        LOG_AMF_APP,
+        "T3592: Maximum retires:%d, for PDU_SESSION_RELEASE_COMPELETE done "
+        "hence Abort "
+        "the pdu sesssion release "
+        "procedure\n",
+        smf_ctx->retransmission_count);
+    // To abort the registration procedure
+    // amf_proc_registration_abort(amf_ctx, ue_amf_context);
+    // pdu_session_resource_release_abort()
```
ulaskozat: will this be uncommented in another PR?

ulaskozat (approved):
LGTM

lte/gateway/c/core/oai/tasks/amf/amf_app_timer_management.cpp
Kaleem-Wavelabs: ok sure @ulaskozat, will try UT for this timer lib APIs.

lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
Kaleem-Wavelabs: As of now, these are not needed, we will remove.

Kaleem-Wavelabs: done

lte/gateway/c/core/oai/tasks/amf/amf_app_timer_management.cpp
Kaleem-Wavelabs: done

PR #8740, opened 2021-08-24
fix(agw): Rename Reboot state to be generic

amarpad:
## Summary
The reboot command is the same across eNBs so rename it to reflect
the same

## Test Plan

Validated reboot against sercomm small cell.

## Additional Information


ardzoht (approved):
lgtm

PR #8745, opened 2021-08-24
chore(nms): Remove circular imports of magmalte package

andreilee:

## Summary

The NMS was still importing from package `@fbcnms/magmalte', which is just importing from the same package. Reworked these imports to be more consistent and not circular.

## Test Plan

Sanity checked by bringing up a local development instance of the NMS and clicked through all the LTE and FeG network pages to make sure everything still works.

## Additional Information


HannaFar (approved):
LGTM (there's some eslint errors)

PR #8746, opened 2021-08-24
feat(orc8r): orc8r cost reduction

quentinDERORY:
## Summary

Create two new sizes of orc8r deployment:
 - poc size ~50 % cheaper (instance price per hour) to deploy
 - production size, same as regular but updated rds instance size to m5.large 

## Test Plan

Successfuly deployed orc8r in aws

## Additional Information

Could not choose smaller eks instance size as there is a limit of pods per instance size here: https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt
The ratio of pod/price was more interesting on t3.small than on t3.micro.

t4g instances types are not an option right now as they are arm based and most of helm charts does not behave well on this type of architecture

### DIFF
--- a/experimental/cloudstrapper/playbooks/roles/control/templates/main.tf.j2.poc
+++ b/experimental/cloudstrapper/playbooks/roles/control/templates/main.tf.j2.poc
@@ -13,10 +13,11 @@
 
 # instances price $0.2564/hour
 # 55 % cheaper than production
+# eks instance costs: $0.1664/hour 33 % cheaper than 3 t3.large
+# db instance cost: $0.018/hour 89 % cheaper than m4/m5 large
+# elastic search cost: $0.072/hour  51 % cheaper
 module "orc8r" {
 
-  # 33 % cheaper than 3 t3.large
-  # cost $0.1664/hour
   eks_worker_groups =[ {
         name                 = "wg-1"
         instance_type        = "t3.small"
@@ -33,8 +34,6 @@ module "orc8r" {
 
   region = "{{ awsOrc8rRegion }}"
 
-  # $0.018/hour
-  # 89 % cheaper than m4/m5 large
   orc8r_db_password           = "{{ orc8rDbPassword }}" # must be at least 8 characters
   orc8r_db_instance_class     = "db.t3.micro"
 
@@ -47,8 +46,6 @@ module "orc8r" {
   cluster_name = "{{ orc8rTfCluster }}"
   cluster_version = "1.17"
 
-  # 51 % cheaper
-  # $0.072/hour
   deploy_elasticsearch          = true
   elasticsearch_domain_name     = "{{ orc8rTfEs }}"
   elasticsearch_version         = "7.7"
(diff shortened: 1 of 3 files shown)


hcgatewood (approved):
Awesome. For the tf files -- can you move all the price-related comments to a single block comment near the top of the tf file. I think it will read better that way

Excited for this!

I think the next step would be experimenting with using Postgres/ES K8s operators to remove the RDS/ES deployments altogether. Is that something you're interested in taking on? Down to sync if so, ping me on Slack 🙂

experimental/cloudstrapper/playbooks/roles/control/templates/main.tf.j2.poc
```diff
@@ -0,0 +1,112 @@
...
+module "orc8r" {
+
+  # 33 % cheaper than 3 t3.large
+  # cost $0.1664/hour
+  eks_worker_groups =[ {
+        name                 = "wg-1"
+        instance_type        = "t3.small"
+        asg_desired_capacity = 8
+        asg_min_size         = 1
+        asg_max_size         = 8
```
arunuke, line 26: Considering we only launch about two instances of each type, is there a reason why we would need to scale to a 8-node instance?

arunuke (approved):
Is it possible to get some screenshots of both configs and their respective actual costs? This would help us close the Magma Program Goal of creating an Orc8r deployment that costs 50% less than the current prod version?

experimental/cloudstrapper/playbooks/roles/control/templates/main.tf.j2.poc
quentinDERORY: We deploy 57 pods in the orc8r deployment and the t3.small have 11 pod per instances.
The orc8r deployment requires minimum 6 instances of t3.small but with only 6 and 7 instances, the k8 cluster has issues.

quentinDERORY:
> Is it possible to get some screenshots of both configs and their respective actual costs? This would help us close the Magma Program Goal of creating an Orc8r deployment that costs 50% less than the current prod version?

I will deploy it in a new region and track costs from there for one day. AWS cost tracker makes it hard when there are resources already in that region.
There will be some uncertainty as there is always the refund that happens at the end of the month.

arunuke:
> > Is it possible to get some screenshots of both configs and their respective actual costs? This would help us close the Magma Program Goal of creating an Orc8r deployment that costs 50% less than the current prod version?
> 
> I will deploy it in a new region and track costs from there for one day. AWS cost tracker makes it hard when there are resources already in that region.
> There will be some uncertainty as there is always the refund that happens at the end of the month.

You could consider using two unused regions just to eliminate cost for any other resources. The same config file with awsOrc8rRegion changed should help in setting it up.

quentinDERORY:
Screen Shot 2021-09-01 at 2 43 19 PM
I first deployed a production size orc8r for a couple of days which resulted in $0.83/hour in canada.
Then I deployed the poc size for two days now for $0.53/hour.
The cost that I did not reduce are the ec2 elbs and eks fee which are roughly $0.24/hour.

The cost reduction is approximately 36 %.

NB: It took one day to wipe out remaining resources of the former production deployment. Will let the poc deployment live a couple days more to see if AWS lower the price again.

Issue #8752, opened 2021-08-25
[FEG] Host-IP-Address AVP in CER is hardcoded to 127.0.0.1


sourabh-nanoti:
According to Diameter RFC https://datatracker.ietf.org/doc/html/rfc6733 
the AVP Host-IP-Address in CER should be set to the IP address which will be used to send diameter messages. As of now its  hardcoded to localhost ip  address 127.0.0.1.
During a recent interop with a 3rdParty PCRF, this  caused an issue as the PCRF was expecting the IP address of the transport endpoint sending CER. 

After we changed the Host-IP-Address AVP to contain valid IP address, the  CER-CEA exchange was successful. (Please see screenshots
<img width="1666" alt="Screen Shot 2021-08-25 at 10 27 00 AM" src="https://user-images.githubusercontent.com/48265190/130716471-78a86a62-45d2-47c1-b1b9-528918931c92.png">
) 

### Your Environment

AGW-Feg. Orc8r in AWS. 
Operator PCRF from Huawei.  
<img width="548" alt="Screen Shot 2021-08-23 at 3 55 53 PM" src="https://user-images.githubusercontent.com/48265190/130716075-41c790a3-1804-4ac0-86f4-124dd4094b2e.png">

MAAHamid:
Faced same issue while integrating with OCS

uri200:
this parameter is configured here:  adding the IP aaddressof your client to the locall_address on your diameter server configuration


![image](https://user-images.githubusercontent.com/16157139/150453926-de89c389-fd1a-4a89-9d38-82449aac8c2c.png)

From https://github.com/magma/magma/issues/10001


![image](https://user-images.githubusercontent.com/16157139/150454099-78ae49d0-8faf-4c77-8de3-3ddb6e38a847.png)

uri200:
Hey @sourabh-nanoti  This Pr https://github.com/magma/magma/pull/11252 may fix the issue

PR #8753, opened 2021-08-25
fix(feg) Changed Host-IP-Address in CER to be local address configured in diameter client.

sourabh-nanoti:
https://github.com/magma/magma/issues/8752 

## Test Plan

Verified the changes by establishing connections with the PCRF as mentioned in the PR description,.

## Additional Information

- [ ] This change is backwards-breaking -- No
Screen Shot 2021-08-25 at 10 27 00 AM

### DIFF
--- a/feg/gateway/diameter/diameter_client.go
+++ b/feg/gateway/diameter/diameter_client.go
@@ -96,7 +96,7 @@ func NewClient(clientCfg *DiameterClientConfig) *Client {
 		ProductName:      datatype.UTF8String(clientCfg.ProductName),
 		OriginStateID:    datatype.Unsigned32(originStateID),
 		FirmwareRevision: 1,
-		HostIPAddress:    datatype.Address(net.ParseIP("127.0.0.1")),
+		HostIPAddress:    datatype.Address(net.ParseIP(clientCfg.Host)),
 	})
 
 	appIdAvp := diam.NewAVP(avp.AuthApplicationID, avp.Mbit, 0, datatype.Unsigned32(clientCfg.AppID))


feg/gateway/diameter/diameter_client.go
```diff
@@ -96,7 +96,7 @@ func NewClient(clientCfg *DiameterClientConfig) *Client {
 		ProductName:      datatype.UTF8String(clientCfg.ProductName),
 		OriginStateID:    datatype.Unsigned32(originStateID),
 		FirmwareRevision: 1,
-		HostIPAddress:    datatype.Address(net.ParseIP("127.0.0.1")),
+		HostIPAddress:    datatype.Address(net.ParseIP(clientCfg.Host)),
```
emakeev, line 99: It's only going to work in case of clientCfg.Host being an IP. Host is DiamIdent, HostIPAddress is Address.
How would it work when you need to set Origin-Host to 'feg.magma.com' and Host-IP-Address to '10.16.2.204'?

How was it tested?

You may need a new config parameter added to DiameterClientConfig to set Host-IP-Address independently from Origin-Host.

emakeev (changes requested):
please see the inline comment.

feg/gateway/diameter/diameter_client.go
sourabh-nanoti: "How would it work when you need to set Origin-Host to 'feg.magma.com' and Host-IP-Address to '10.16.2.204'?"
@emakeev - Thanks for pointing this out. It worked because due to the error introduced by the above change, the HostIPAddress was set  to a nil string. 
Upon examination of the  code in Fiorix, (https://github.com/fiorix/go-diameter/blob/master/diam/sm/client.go) Line#196, if the HostIPAddress is nil, the module calls getLocalAddresses() API which sets it to the correct IP address. Hence the test was successful as you see in the Github issue https://github.com/magma/magma/issues/8752
Do you think we should leave the HostIPAddress as nil for Fiorix to pick up the correct IP address?

sourabh-nanoti: @emakeev Although when SCTP multihoming  is supported we would need a configuration to specify multiple ip addresses here as the Requests can  go via different ip addresses for SCTP. We can attempt that in the future.

emakeev: I would just add a single optional parameter (multihoming support can be done later if needed) to allow setting the IP & make default nil to let diameter to pick local addr if the parameter is not provided.

ekowtaylor:
@sourabh-nanoti please advise if you'd like to proceed with this.

PR #8720, opened 2021-08-23
fix(agw): Fixing pylint tests for open function missing encoding param

ardzoht:
Signed-off-by: Alex Rodriguez <ardzoht@gmail.com>

## Summary

- As title

## Test Plan

- make test 

## Additional Information

- [ ] This change is backwards-breaking

PR #8754, opened 2021-08-25
fix(agw): Fix bootstrap failure due to bad Python open args

hcgatewood:
## Summary

Current master's AGW is failing to connect to the controller at all, including to sync subscribers. Root cause is service registry code is broken due to the light refactor introduced in #8720.

This fix reverts the subset of #8720's changes that introduced the bug.

## Test Plan

Before, the root cause log was

```
Aug 24 20:56:56 mpk-dogfooding magmad[3250809]: ERROR:root:Failed to get rpc channel: binary mode doesn't take an encoding argument
```

After, the fix, this log no longer appears, and gateway is able to successfully bootstrap and make its usual requests to the controller.

## Additional Information

### DIFF
--- a/orc8r/gateway/python/magma/common/service_registry.py
+++ b/orc8r/gateway/python/magma/common/service_registry.py
@@ -100,7 +100,7 @@ class ServiceRegistry:
         authority = proxy_config['bootstrap_address']
 
         try:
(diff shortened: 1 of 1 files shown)


ardzoht (approved):
lgtm, thanks for this fix

themarwhal (approved):

PR #8771, opened 2021-08-25
feat(agw): support containerized agw installation

ymasmoudi:

## Summary

Support containerised AGW installation over Ubuntu focal.
- Install ovs and configure interfaces.
- Install docker and set the environment
- pull and start containers.

## Test Plan

Tested over AWS

Screenshot 2021-08-26 at 12 45 27 AM

```
root@ip-172-31-9-95:/var/opt/magma/docker# docker-compose ps
    Name                   Command                       State           Ports
------------------------------------------------------------------------------
connectiond     /usr/local/bin/connectiond       Up                           
control_proxy   sh -c /usr/local/bin/gener ...   Up                           
ctraced         /usr/bin/env python3 -m ma ...   Up                           
directoryd      /usr/bin/env python3 -m ma ...   Up                           
enodebd         /usr/bin/env python3 -m ma ...   Up                           
eventd          /usr/bin/env python3 -m ma ...   Up                           
health          /usr/bin/env python3 -m ma ...   Up                           
liagentd        /usr/local/bin/liagentd          Exit 0                       
magmad          /usr/bin/env python3 -m ma ...   Up                           
mobilityd       /usr/bin/env python3 -m ma ...   Up                           
monitord        /usr/bin/env python3 -m ma ...   Up                           
oai_mme         sh -c sed -i "s~10\.0\.2\. ...   Up                           
pipelined       bash -c /usr/bin/ovs-vsctl ...   Up                           
policydb        /usr/bin/env python3 -m ma ...   Up                           
redirectd       /usr/bin/env python3 -m ma ...   Up                           
redis           /bin/bash -c /usr/local/bi ...   Up                           
sctpd           /usr/local/bin/sctpd             Up                           
sessiond        sh -c mkdir -p /var/opt/ma ...   Up (health: starting)        
smsd            /usr/bin/env python3 -m ma ...   Up                           
state           /usr/bin/env python3 -m ma ...   Up                           
subscriberdb    /usr/bin/env python3 -m ma ...   Up                           
td-agent-bit    /bin/bash -c /usr/local/bi ...   Up                                        
```


lte/gateway/deploy/agw_install_docker.yml
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
arunuke: Is this a bash file and not a yaml file (as the name indicates)?

lte/gateway/deploy/roles/agw_docker/files/magma_ifaces_gtp
```diff
@@ -0,0 +1,77 @@
+auto gtp_br0
```
arunuke: Is there a way to include_role these from other roles? This is just to make sure changes in the baremetal AGW files are carried over here as applicable. We can always refactor this later, but something to consider.

lte/gateway/deploy/roles/agw_docker/files/magma_interfaces
```diff
@@ -0,0 +1,34 @@
+# interfaces(5) file used by ifup(8) and ifdown(8)
```
arunuke: Same as before. If we can create a symlink or an include_role, this might help in keeping future changes consistent.

lte/gateway/deploy/roles/agw_docker/tasks/main.yml
```diff
@@ -0,0 +1,195 @@
...
+  become: yes
+  apt:
+    update_cache: yes
+
+- name: Install runtime dependencies.
+  become: yes
+  apt:
+    name: "{{ packages }}"
+  vars:
+    packages:
```
arunuke: Are all these packages required for Openvswitch (since that's the only element used on the baremetal host?)? In other words, are the versions we used custom or can be picked up from focal repos directly? @pshelar for inputs.

lte/gateway/deploy/roles/containers/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+- name: copy the config files for magma
+  shell: cp -r  /opt/magma/lte/gateway/configs/* /etc/magma/
+
+- name: copy the config files for orchestrator
+  shell: cp -r  /opt/magma/orc8r/gateway/configs/templates/* /etc/magma/templates/
+
+- name: copy docker compose config file
+  shell: cp -r  /opt/magma/lte/gateway/docker/docker-compose.yml /var/opt/magma/docker/
+
+- name: copy the rootca
```
arunuke: Perhaps a check earlier as an assert to ensure rootCA.pem exists?

arunuke (changes requested):
General feedback on addressing relevant lint warnings (the double quotes warnings seem to be relevant)

lte/gateway/deploy/roles/agw_docker/tasks/main.yml
```diff
@@ -0,0 +1,195 @@
...
+    - "/etc/magma"
+    - "/etc/magma/templates/"
+    - "/var/opt/magma/docker"
+
+- name: Enable IP forwarding
+  become: true
+  sysctl: name="net.ipv4.ip_forward" value=1 sysctl_set=yes state=present reload=yes
+
+- name: touch snowflake
+  shell: touch /etc/snowflake
```
arunuke, line 93: SNOWFLAKE_PATH perhaps? It appears to be referenced in the docker-compose file as well.

lte/gateway/deploy/roles/containers/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+  copy: content={{ rootca }} dest=/var/opt/magma/certs/rootCA.pem mode='400'
```
arunuke: https://magma.github.io/magma/docs/lte/config_agw seems to indicate the file be in /var/opt/magma/tmp/certs/rootCA.pem. Is there any historical reason behind this or it doesn't matter as long as the volume, control_proxy.yml entry match?

lte/gateway/deploy/roles/containers/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+- name: Clone the git repo
+  git:
+   repo: "https://github.com/magma/magma.git"
+   clone: yes
+   force: yes
+   dest: "/opt/magma"
+   version: "{{ image_version }}"
+
+- name: copy the config files for magma
+  shell: cp -r  /opt/magma/lte/gateway/configs/* /etc/magma/
```
arunuke: Would the config files be modified for the containerization use-case? Ex: magmad is not required to monitor any services running on the node anymore.

lte/gateway/deploy/agw_install_docker.yml
```diff
@@ -0,0 +1,61 @@
...
+DEPLOY_PATH="/opt/magma/lte/gateway/deploy"
+
+echo "Checking if the script has been executed by root user"
+if [ "$WHOAMI" != "root" ]; then
+  echo "You're executing the script as $WHOAMI instead of root.. exiting"
+  exit 1
+fi
+
+echo "Checking if Ubuntu is installed"
+if ! grep -q 'Ubuntu' /etc/issue; then
```
pshelar: do we care about base OS for AGW installation? It would be nice document the reasoning here.

lte/gateway/deploy/roles/agw_docker/files/magma_ifaces_gtp
pshelar: or just create a symbolic link. Lets not add any dup config.

lte/gateway/deploy/roles/agw_docker/tasks/main.yml
```diff
@@ -0,0 +1,195 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Include vars of all.yaml
```
pshelar, line 13: can as above this is introducing code duplication. can you refactor existing ansible roles and fork deployment for using debian package and docker containers.

pshelar (changes requested):
we need to restructure the ansible code.

lte/gateway/deploy/agw_install_docker.yml
arunuke: Some host-specific packages are needed to get OVS up. If they are standard upstream packages of a certain version with no specific customization, they could be pulled directly from the host repo. However, there are some platform specific commands (ex: installing sudo, adding user etc., please see below) to prep the host.

The current approach is to address them (along with the reuse of code) in refactor before the code goes from alpha to beta.

lte/gateway/deploy/roles/agw_docker/files/magma_ifaces_gtp
arunuke: @ymasmoudi and I talked about this the other day in our review. The reason the code is repeated here is because some tasks on the baremetal need to be skipped and a decision was made to leave the current working code as-is while the feature is alpha. Before going to beta, a refactor is planned which would unify these streams where some tasks can be skipped for the containerized AGW.

lte/gateway/deploy/roles/containers/tasks/main.yml
ymasmoudi: Yes, moved the verification to the installation script

lte/gateway/deploy/roles/containers/tasks/main.yml
ymasmoudi: Actually, feg and cwag, xwfm expect the rootca here /var/opt/magma/certs/rootCA.pem
But yes, it does not matter as long as the control proxy matches.

lte/gateway/deploy/roles/containers/tasks/main.yml
ymasmoudi: magmad is still able to monitor all the other services.
The only change for now is nat interface in pipelined which is set to a default eth2 (moved to eth0)

lte/gateway/deploy/roles/agw_docker/tasks/main.yml
ymasmoudi: there will be no customisation required for snowflake path.
removed SNOWFLAKE_PATH.

ymasmoudi:
@arun-magma @pshelar reworked the PR to re-use existing tasks.
[Correction] It's @arunuke

lte/gateway/deploy/agw_install_docker.sh
```diff
@@ -0,0 +1,78 @@
...
+if ! grep -q "$MAGMA_USER ALL=(ALL) NOPASSWD:ALL" /etc/sudoers; then
+  apt install -y sudo
+  adduser --disabled-password --gecos "" $MAGMA_USER
+  adduser $MAGMA_USER sudo
+  echo "$MAGMA_USER ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
+fi
+
+# unlink resolv.conf to avoid overrides by systemd-resolved
+cp /etc/resolv.conf /etc/resolv.conf.bak
+rm -f /etc/resolv.conf && mv /etc/resolv.conf.bak /etc/resolv.conf
```
pshelar: why is this required? 
I dont see this hack on bare metal agw.

lte/gateway/deploy/agw_install_docker.sh
```diff
@@ -0,0 +1,78 @@
...
+ROOTCA="/var/opt/magma/certs/rootCA.pem"
+if [ ! -f "$ROOTCA" ]; then
+
+  echo "Upload rootCA to $ROOTCA"
+  exit 1
+fi
+
+echo "Install Magma"
+apt-get update
+apt-get -y install curl make virtualenv zip rsync git software-properties-common python3-pip python-dev apt-transport-https
```
pshelar: given this is dockerized agw, its odd to see these packages getting installed.

lte/gateway/deploy/agw_install_docker.sh
```diff
@@ -0,0 +1,78 @@
...
+echo "[agw_docker]
+127.0.0.1 ansible_connection=local
+[magma_deploy]
+127.0.0.1 ansible_connection=local" > $DEPLOY_PATH/agw_hosts
+
+# install magma and its dependencies including OVS.
+su - $MAGMA_USER -c "sudo ansible-playbook -e \"MAGMA_ROOT='/opt/magma' OUTPUT_DIR='/tmp'\" -i $DEPLOY_PATH/agw_hosts --tags agwc $DEPLOY_PATH/magma_deploy.yml"
+su - $MAGMA_USER -c "sudo ansible-playbook -e \"MAGMA_ROOT='/opt/magma' OUTPUT_DIR='/tmp'\" -i $DEPLOY_PATH/agw_hosts $DEPLOY_PATH/magma_docker.yml"
+
+echo "Cleanup temp files"
```
pshelar: how is interface renaming handled ?
is that step not automated for dockerized AGW?

lte/gateway/deploy/roles/agw_docker/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
...
+    name: "{{ packages }}"
+    state: present
+    update_cache: yes
+  vars:
+    packages:
+    - apt-transport-https
+    - ca-certificates
+    - curl
+    - software-properties-common
+    - gnupg-agent
```
pshelar: it would be nice to move all of these packages to the docker container that needs them. can we comment which containers needs these packages?

ymasmoudi: removed.

lte/gateway/deploy/agw_install_docker.sh
ymasmoudi: removed.

lte/gateway/deploy/agw_install_docker.sh
ymasmoudi: Yes, It is not automated yet. 
For now, we use the existing scripts for standard installation to do it.

lte/gateway/deploy/agw_install_docker.sh
ymasmoudi: this is solution for an ec2 issue we were facing during the setup.
I removed it for now and we will see how to include the fix separately

lte/gateway/deploy/agw_install_docker.sh
```diff
@@ -0,0 +1,68 @@
...
+pip3 install ansible
+
+rm -rf /opt/magma/
+git clone "${GIT_URL}" /opt/magma
+cd /opt/magma || exit
+git checkout "$MAGMA_VERSION"
+
+cp -f $DEPLOY_PATH/roles/magma/files/magma_ifaces_gtp /etc/network/interfaces.d/gtp
+cp -f $DEPLOY_PATH/roles/magma/files/ovs-kmod-upgrade.sh /usr/local/bin/
+cp -f $DEPLOY_PATH/roles/magma/files/magma_modules_load /etc/modules-load.d/magma.conf
```
pshelar: Lets move the `cp`  to ansible role. 
so this script should be only installing prerequisite for the ansible script.

lte/gateway/deploy/roles/magma_deploy/tasks/main.yml
```diff
@@ -81,13 +94,30 @@
...
+    name: "{{ packages }}"
+  vars:
+    packages:
+      - libopenvswitch
+      - openvswitch-datapath-dkms
+      - openvswitch-datapath-source
+      - openvswitch-common
+      - openvswitch-switch
+  tags:
+    - never
```
pshelar, line 111: This is interesting tag, is it never used ? then why do we need it?

lte/gateway/deploy/roles/trfserver/files/util/traffic_messages.py
```diff
@@ -1 +0,0 @@
-../../../../../python/integ_tests/s1aptests/util/traffic_messages.py
```
pshelar: This looks like unrelated change.

pshelar (changes requested):
I dont see changed done yet.

lte/gateway/deploy/roles/magma_deploy/tasks/main.yml
ymasmoudi: "never" prevents a task from running unless a tag is specifically requested. In this case agwc.

pshelar: I see... I think it fine for now.
But with so many tags it becoming hard to keep track of all steps are in a ansible role. it would be better if there is separate ansible role to deploy containerized AGW.

arunuke (approved):

lte/gateway/deploy/agw_install_docker.sh
ymasmoudi: @pshelar I moved it to ansible.

lte/gateway/deploy/roles/agw_docker/vars/all.yaml
```diff
@@ -0,0 +1,32 @@
...
+magma_use_pkgrepo: yes
+
+docker_compose_version: 1.29.2
+
+docker_registry: ""
+docker_user: ""
+docker_pass: ""
+image_version: "latest"
+
+sgi_interface_ip: "172.31.9.95/24"
```
pshelar: This looks new change,
Why deployment role has hardcoded IP address configured ?

pshelar (approved):
LGTM

arunuke (approved):
Changes and plan LGTM.

PR #8772, opened 2021-08-25
feat(agw): Pass mconfig and service config to post_processing

amarpad:
## Summary
The PostProcessor class is the class where model specific customization
happens. The current interface to this class is too simple and assumes the
post processing is always the same. Extend it to include configs passed into
the eNB. This allows other eNB models (like Sercomm/FreedomFiOne) to
customize the eNB settings based on the config.

## Test Plan
make test_python

## Additional Information


ardzoht (approved):
lgtm, you're gonna have to update the commit title to `feat(...` for the semantic check to pass

PR #8774, opened 2021-08-25
fix(agw): Add support for ParameterKey

amarpad:
## Summary

ParameterKey is a field that can be used to communicate the generation of
config that is being set on the enodeb. This is useful for debugging and
some models of eNB (e.g. Sercomm) require this to be set. It looks like
the field was only partially implemented (only in model).
Also fix the type of ParameterKey to be correct in the model definition
(per standard) and updates tr-69 tests to validate Key and NS.
Since there is no way to validate that ParamKey is supported by other
eNBs add the field through an opt in flag.

## Test Plan

test_python

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/enodebd/state_machines/enb_acs.py
+++ b/lte/gateway/python/magma/enodebd/state_machines/enb_acs.py
@@ -12,6 +12,7 @@ limitations under the License.
 """
 from abc import ABC, abstractmethod
 from asyncio import BaseEventLoop
+from time import time
 from typing import Any, Type
 
 from magma.common.service import MagmaService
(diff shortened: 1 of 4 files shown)


lte/gateway/python/magma/enodebd/state_machines/enb_acs.py
```diff
@@ -40,12 +41,15 @@ class EnodebAcsStateMachine(ABC):
     This ABC is more of an interface definition.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, use_param_key=False) -> None:
```
ardzoht:
```suggestion
    def __init__(self, use_param_key=False: bool) -> None:
```
minor nit to add typing

ardzoht (approved):
lgtm, just one nit on python typing

PR #8848, opened 2021-08-27
fix(mme): Default Allow all Packet Filters & QOS Rule seems not adding in right QF-id 5 on DL 5G NAS

ganeshg87:
fix(mme): Default Allow all Packet Filters & QOS Rule seems not adding in right QF-id 5 on DL 5G NAS

Test case:
   - verified registration with PDU Session setup.

Signed-off-by: ganeshg87 <ganesh.gedela@wavelabs.ai>

## Additional Information


ganeshg87:
verified basic registration with PDUSession and Qfi value is setting correctly.
Please find attached log & pcap snap
image
[mme_pdu_qfi_fix.log](https://github.com/magma/magma/files/7066113/mme_pdu_qfi_fix.log)

ulaskozat (approved):
LGTM

PR #8850, opened 2021-08-27
fix(agw): Handle WaitSetParameterValue correctly

amarpad:
## Summary
ACS clients can return non zero values for set calls on success (per std).
Make an optional flag to allow for SetParams to return non zero values.
Validated against FreedomFiOne and Baicells which return 1, 0 for
SetParameterValueResponse message.Status respectively.
Make change backwards compatible by using optional flag.

## Test Plan

make test_python
Sercomm integration

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/enodebd/state_machines/enb_acs_states.py
+++ b/lte/gateway/python/magma/enodebd/state_machines/enb_acs_states.py
@@ -10,7 +10,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
-
 from abc import ABC, abstractmethod
 from collections import namedtuple
 from typing import Any, Optional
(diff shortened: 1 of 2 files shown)


ardzoht (approved):
lgtm

lte/gateway/python/magma/enodebd/state_machines/enb_acs_states.py
```diff
@@ -918,19 +917,26 @@ def __init__(
...
+        self.status_non_zero_allowed = status_non_zero_allowed
 
     def read_msg(self, message: Any) -> AcsReadMsgResult:
         if type(message) == models.SetParameterValuesResponse:
-            if message.Status != 0:
-                raise Tr069Error(
-                    'Received SetParameterValuesResponse with '
-                    'Status=%d' % message.Status,
-                )
+            if not self.status_non_zero_allowed:
```
ardzoht, line 934: it may be worth to add a unit test for this case?

PR #8860, opened 2021-08-30
feat(orc8r): Context propagation in mutating configurator entity functions

xjtian:
## Summary

- Propagate context args in all configurator client API functions that write entities

## Test Plan

- make test

## Additional Information


uri200:
Hey @xjtian , what was the behaviour before this change? Were we just handling wrong ctx?

cwf/cloud/go/services/cwf/obsidian/handlers/handlers_test.go
```diff
@@ -985,12 +985,8 @@ func seedCwfGateway(t *testing.T, id string, hwId string) {
...
 func seedCwfTier(t *testing.T, networkID string) {
 	// setup fixtures in backend
-	_, err := configurator.CreateEntities(
-		networkID,
-		[]configurator.NetworkEntity{
-			{Type: orc8r.UpgradeTierEntityType, Key: "t1"},
-		},
-		serdes.Entity,
-	)
+	_, err := configurator.CreateEntities(context2.Background(), networkID, []configurator.NetworkEntity{
```
uri200, line 988: I think 
```
context2 "context"
"golang.org/x/net/context"
```
are the same. What about removing "golang.org/x/net/context" which is the old reference to the library? We already did something similar for feg

xjtian: I'm going to be replacing all usages of the deprecated `golang.org/x/net/context` package in a followup PR - this is just something that the refactoring engine in IDEA did.

xjtian:
> Hey @xjtian , what was the behaviour before this change? Were we just handling wrong ctx?

We would just pass in a `context.Background()`, which means that any tracing headers/timeouts/etc would be lost after the first network hop.

lte/cloud/go/services/lte/servicers/indexer_servicer_test.go
```diff
@@ -111,29 +111,21 @@ func seedNetwork(t *testing.T, networkID string) {
...
-		configurator.NetworkEntity{
-			Type:         orc8r.MagmadGatewayType,
-			Key:          gatewayID,
-			Config:       &models2.MagmadGatewayConfigs{},
-			PhysicalID:   hwID,
-			Associations: []storage.TypeAndKey{{Type: orc8r.UpgradeTierEntityType, Key: "t0"}},
-		},
-		serdes.Entity,
-	)
+	_, err := configurator.CreateEntity(context2.Background(), networkID, configurator.NetworkEntity{
```
hcgatewood, line 114: Not quite the desired formatting but also not huge deal since we're trying to automate this

hcgatewood (approved):
Skimmed, lgtm

uri200 (approved):

Issue #8609, opened 2021-08-11
[FeG] Cancel Location Answer with a Result Code 3002 UNABLE TO DELIVER


emersonacuna:
### Your Environment

- **Version:** 1.6 
- **Affected Component:**  Fed Gateway
- **Affected Subcomponent:** Fed Gateway
- **Deployment Environment:** local docker-compose (orc8r), Openstack AGW/FeG)

### Describe the Issue
FeG receives a Command Code: 317 3GPP-Cancel-Location from HSS. FeG answers with a result code 3002 DIAMETER_UNABLE_TO_DELIVER in frame # 703
![image](https://user-images.githubusercontent.com/79710920/129115645-3e1dd302-8d55-4171-bf2f-7ea0d776fa62.png)

Logs of the s6a_proxy shows this failure:
```
[36ms6a_proxy        |[0m I0810 20:36:20.814825       1 service_config.go:169] Successfully loaded '::control_proxy' service configs from '/etc/magma/control_proxy.yml'
[36ms6a_proxy        |[0m I0810 20:36:20.815072       1 service_config.go:180] Successfully loaded Override configs for service :control_proxy from '/var/opt/magma/configs/control_proxy.yml'
[36ms6a_proxy        |[0m I0810 20:36:20.815116       1 cloud_connection.go:116] connecting to: 127.0.0.1:8443, authority: feg_relay-controller.magma.test
[36ms6a_proxy        |[0m E0810 20:36:20.851575       1 cl.go:49] Failed to forward CLR to gateway. err: rpc error: code = Unknown desc = OK: HTTP status code 200; transport: missing content-type field. Retries left: 3
[36ms6a_proxy        |[0m I0810 20:36:20.851659       1 cloud_connection.go:116] connecting to: 127.0.0.1:8443, authority: feg_relay-controller.magma.test
[36ms6a_proxy        |[0m E0810 20:36:20.874176       1 cl.go:49] Failed to forward CLR to gateway. err: rpc error: code = Unknown desc = OK: HTTP status code 200; transport: missing content-type field. Retries left: 2
[36ms6a_proxy        |[0m I0810 20:36:20.874305       1 cloud_connection.go:116] connecting to: 127.0.0.1:8443, authority: feg_relay-controller.magma.test
[...]

PR #8877, opened 2021-08-30
fix(agw): Provided fix for issue 8609

rsarwad:
fix(agw): Provided fix for issue, https://github.com/magma/magma/issues/8609
## Summary
As part of fixing issue, https://github.com/magma/magma/issues/8609, it's required to support asynchronous server, create a task to handle messages received on asynchronous server.
## Test Plan
1. Executed s1ap sanity test suite on local setup
2. Executed CLR and reset test case on Tera VM


uri200:
Is this supposed to be closed?

PR #8909, opened 2021-08-31
feat(sctpd): flagged configuration of upstream / downstream sock

### LINKED ISSUE #8828: [AGW][vN] MME build flag to enable SCTPD proxy via accessd
Our preference here is to enable over-ride of the Unix Domain Socket for uplink / downlink SCTP traffic through sctpd.

The configuration will take place in the `MME`, allowing it to optionally be redirected away from the standard SCTP uplink / downlink Unix Domain Sockets offered by `sctpd`, and instead to the set offered by `accessd` - the golang proxy.

Configuration based change (e.g. through Magma's `mconfigs`) is preferred, though a gflag fallback could be acceptable.

electronjoe:
## Summary

Added linking with gflags library, and added two flags to the sctpd binary:

```
  Flags from /magma/lte/gateway/c/sctpd/src/sctpd.cpp:
    -downstream_sock (Path to unix domain socket for SCTPD downstream traffic.)
      type: string default: "unix:///tmp/sctpd_downstream.sock"
    -upstream_sock (Path to unix domain socket for SCTPD upstream traffic.)
      type: string default: "unix:///tmp/sctpd_upstream.sock"
```

Closes #8828.

## Test Plan

- ran LTE integ_test to valdiate existing behavior
- verified command line flags help display

### DIFF
--- a/lte/gateway/c/core/oai/tasks/sctp/sctpd_uplink_server.cpp
+++ b/lte/gateway/c/core/oai/tasks/sctp/sctpd_uplink_server.cpp
@@ -136,18 +136,6 @@ using magma::mme::SctpdUplinkImpl;
 std::shared_ptr<SctpdUplinkImpl> service = nullptr;
 std::unique_ptr<Server> server           = nullptr;
 
-int start_sctpd_uplink_server(void) {
-  service = std::make_shared<SctpdUplinkImpl>();
-
-  ServerBuilder builder;
-  builder.AddListeningPort(UPSTREAM_SOCK, grpc::InsecureServerCredentials());
-  builder.RegisterService(service.get());
-
-  server = builder.BuildAndStart();
-
-  return 0;
-}
-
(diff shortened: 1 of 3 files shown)


electronjoe:
I've discovered that the configuration of these linux FD paths crosses over into MME land (which makes sense, in retrospect).  What I did not anticipate is that the MME code calls into scptd functions to achieve this.  Likely as an awkward step to enable more C++ use in MME.

I am not clear that we have any support for flag use in our C code.

I'm considering whether we do something awful like call extern C++ functions to enable gflags use from MME... Or if I should instead use the json based configuration files that I believe many components read for configuration [magma::mconfig::MME](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/magma/magma%24+magma::mconfig+file:%5Elte/gateway/c/core/oai/&patternType=literal).

Issue #8927, opened 2021-09-01
[agw] Experiment w/ Golang for main gateway functionality


markjen:
# Experiment w/ Golang for main gateway functionality

## Problem

1. Our current lte/gateway codebase suffers from stability, performance, and maintenance issues.
  i. `lte/gateway/c` code is a c/c++ mix, much of it having poor testability. This has resulted in a dearth of tests; efforts to add tests lead to requiring major refactoring.
  ii. `lte/gateway/python` has significant portions without type-hinting/checking and has performance issues.
2. Developer tooling support in the `lte` codebase is poor. Builds are slow and not hermetic (see #8338), and IDEs/debuggers do not work reliably.
3. Onboarding developers into the `lte` codebase is slow and error-prone.
4. Rate of regression in the `lte` codebase is high, even for developers experienced in the codebase. 
5. Mobility across the codebase (e.g. between `orc8r`, `feg`, and `lte`) is high friction.

Combined, these issues are an impediment to Magma ecosystem growth, a continuous drag on productivity, and a major contributing factor to ongoing issues in deployments.

However, we also have significant unknowns with using a different programming language and toolchain for `lte` services. This expirement aims to answer the following:

1. Will Golang (or other alternatives) be able to achieve the functional and performance requirements?
2. Will the developer experience actually improve? Or will we trade one set of issues for another?
3. Will improvements to developer experience increase velocity, decrease issues, and promote mobility across the codebase?

## Solution

Experiment with reimplementing `lte` components in Golang, starting with the S1AP task in MME and Pipelined.GetStats. Experimental code will not be used in any release until we achieve exit criteria; if the experiment is a success, another proposal will be submitted to decide future development direction, using results from this experiment as an input.

Goals of this experiment are:

1. Experiment should pass S1APTest when integrated with existing services on a magma vargrant VM.
2. New implementation should achieve the same performance metrics as our published v1.6 benchmark.
3. Test code coverage should be >= 75%. Execution of tests to achieve this coverage should be < 15 minutes.
4. Improved developer experience:
  i. Decreased onboarding time / dev environment setup time
  ii. All IDE / dev tools working for all developers
  iii. Decreased time to first PR
  iv. Code is buildable / unit tests are runnable in the developers' native environment w/ IDE integration

## Non-goals

1. Interruption of any existing workstreams, e.g. 5G

hcgatewood:
@markjen are you desiring a specific vote on this from the TSC, or is this more for communication purposes?

[...]

PR #8928, opened 2021-09-01
feat(agw): Golang experiment initial commit

### LINKED ISSUE #8824: [AGW][vN] Setup lte/gateway/go/accessd main binary

### LINKED ISSUE #8845: [AGW][vN] lte/gateway/go/log abstraction and first implementation with zap for local logging

markjen:
## Summary

Set up new polyglot/monorepo Go module for AGW Golang experiment #8927

## Test Plan

`github.com/magma/magma/log` package is fully unit tested
`github.com/magma/magma/protos` package is codegen, no tests
`github.com/magma/magma/accessd` and `github.com/magma/magma/service` packages are WIP and will be undergoing heavy changes, no tests added yet.

## Additional Information


src/go/protos/protos.go
```diff
@@ -0,0 +1,18 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package protos
+
+//go:generate go install google.golang.org/protobuf/cmd/protoc-gen-go
+//go:generate go install google.golang.org/grpc/cmd/protoc-gen-go-grpc
+
+//go:generate protoc --go_out=. --go_opt=Msctpd.proto=magma/sctpd;sctpd -I ../../../lte/protos sctpd.proto
```
mstre123, line 17: Might be worth looking into using prototool to handle our proto generation.
It takes a config based approach which might make it a bit cleaner than the go:generate commands
Other fun part is supports linting your protos
https://github.com/uber/prototool#prototool-generate

mstre123: it also pretty nifty for making grpc calls using your protos against your services

markjen: Ah cool. Looks like there is also https://github.com/bufbuild/buf

Also, this is a temp/stop-gap while we figure out our bazel stuff cc: @themarwhal

src/go/accessd/main.go
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	uber_zap "go.uber.org/zap"
```
themarwhal, line 15: Do we prefer this over glog?

electronjoe: I looked into this when Mark mentioned it - glog is "frozen" for Go if I recall correctly, and this is a forked glog with fixes that the go community needed for e.g. cloud generalization (again, if I recall correctly).  Choosing this over glog made sense from what I saw.  I think Mark mentioned possibly using our own interface definition for logging, allowing us to swap out the back end implementation?

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
```
electronjoe, line 12: Here and elsewhere, I thought every go package was supposed to have a package level comment explaining the purpose of the package? Will a go linter complain here (or will it be fooled by the copyright notice, perhaps)? Or are you not for package level comments? Or am I confused (possibly).

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
```
electronjoe, line 14: Related topic, what Go linter would you like CI to be running? I'm confident we are using one on Orc8r code side, but would like to hear if we are aligned there. @hcgatewood FYI

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+	"net"
+	"time"
+
+	"github.com/pkg/errors"
```
electronjoe, line 19: Dropping comments without doing much legwork myself - so apologies if I'm asking questions that are answered elsewhere.  Does our e.g. Golang Best Practices document call out our preference for github's errors package? Related to Marie's question above, do we want / need a list of these common infrastructure packages that we are importing - and why / how they were selected? Seems useful (as I go google each one of them to understand the decision). @hcgatewood @themarwhal

src/go/accessd/main.go
electronjoe: Below I see we are using `github.com/magma/magma/log` for e.g. `src/go/accessd/server/server.go` so now I am curious how these interact - I'll dig as I review but I think our logging strategy needs to be documented somewhere (I just reviewed our as-of-yet-not-released-publicly Golang Best Practices doc and see nothing about logging). @hcgatewood

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+	"google.golang.org/grpc"
+
+	"github.com/magma/magma/log"
+	"github.com/magma/magma/protos/magma/sctpd"
```
electronjoe: Is this a protocol buffer compiled output? Inside Google I'm accustomed to an alias for any proto import with suffix _proto or maybe it was _pb (or at least that is what I recall).  WDYT?

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+	service_sctpd "github.com/magma/magma/service/sctpd"
+)
+
+const (
```
electronjoe, line 27: One or more of these - might they be desirable to be flag supported? Particularly timeouts for debug / testing? Comment / question applies to other modules below with Constants that might want to be accessible at command line for debug / iteration?

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+	mmeUpstreamTarget  = "unix:///tmp/mme_sctpd_upstream.sock"
+	mmeGrpcDialTimeout = time.Second
+)
+
+func newMmeGrpcConn() (*grpc.ClientConn, error) {
+	ctx, _ := context.WithTimeout(context.Background(), mmeGrpcDialTimeout)
+	return grpc.DialContext(ctx, mmeUpstreamTarget, grpc.WithInsecure(), grpc.WithBlock())
```
electronjoe, line 35: `WithBlock()` makes `newMmeGrpcConn()` a blocking function call pending bring-up of the MME endpoint, I think? That is the type of implication I might want in a comment on the function? I realize this a private function, but the use of it chains up to `Start()` below, which is public (and lacks comments).

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func newServer(logger log.Logger) *grpc.Server {
+	grpcServer := grpc.NewServer()
+
+	mmeGrpcConn, err := newMmeGrpcConn()
+	if err != nil {
+		panic(err)
```
electronjoe, line 43: Do we generally prefer `panic`, or a logger call like `~logger.Fatal`? I cannot actually recall in Golang / glog if there is a fatal logging option. Might again be a topic for golang best practices? With panic do we get logs in our logging infra?

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+func Start(logger log.Logger) error {
+	listener, err := net.Listen(sctpdUpstreamNetwork, sctpdUpstreamPath)
+	if err != nil {
+		return errors.Wrapf(
+			err,
+			"net.Listen(network=%s, address=%s)",
+			sctpdUpstreamNetwork,
+			sctpdUpstreamPath)
+	}
+	return errors.WithStack(newServer(logger).Serve(listener))
```
electronjoe, line 65: Aah, this appears to be one of the special sauces of github errors... I think probably this belongs in our Golang Best practices? @hcgatewood

src/go/log/log.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+var levelNames = map[Level]string{
+	DebugLevel: "DEBUG",
+	InfoLevel:  "INFO",
+	WarnLevel:  "WARN",
+	ErrorLevel: "ERROR",
+}
+
+func (l Level) String() string {
```
electronjoe, line 110: Looking at the code it's super obvious what is happening here, but if somebody were just reading Go docs they wouldn't have anything to work with.  Broadly I'm seeing a lot of Public Functions without comments. What's our stance on that?  Will linters complain (linters we want to apply for public contributions)?

src/go/log/log.go
```diff
@@ -0,0 +1,136 @@
...
+	Error() Printer
+	Warning() Printer
+	Info() Printer
+	Debug() Printer
+
+	Level() Level
+	SetLevel(level Level)
+
+	Named(name string) Logger
+	With(field string, value interface{}) Logger
```
electronjoe, line 147: I understand Named generally (this will be a log stream with a prefix / channel concept?). But what is "With"? Is this also a stream differentiator based upon some dynamic field? A sub-stream of Named? I'll keep reviewing and maybe this will be answered in code, but this being a common API for consumers I think comments would be helpful.

src/go/accessd/main.go
electronjoe: Also I see an implementation of our logger interface in `src/go/log/zap/zap.go` below. So why the reference here to `uber_zap`? And why our own zap implementation? Still figuring this out...

src/go/log/log.go
```diff
@@ -0,0 +1,136 @@
...
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package log
```
electronjoe, line 77: I really like your logger interface implementation here!  Super clear / clean.

src/go/log/log.go
```diff
@@ -0,0 +1,136 @@
...
+		l, ok := m.loggers[fullName]
+		if ok {
+			prev = l
+			continue
+		}
+
+		for j := i; j < len(names); j++ {
+			newFullName := FullName(names[0 : j+1])
+			if _, ok := m.loggers[newFullName]; ok {
+				panic(fmt.Sprintf("sublogger %s exists, but %s didn't", newFullName, FullName(names[0:j])))
```
electronjoe: Same question as above, panic versus a log fatal? What's our plan here. I guess Panic results will be captured by our crash detection infra? E.g. Sentry.io?

src/go/log/log_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLevel_String(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
```
electronjoe, line 24: I <3 Golang table based testing. So readable! And easily extended with new cases.

src/go/log/log_test.go
```diff
@@ -0,0 +1,180 @@
...
+			want:  "ERROR",
+		},
+		{
+			level: 100,
+			want:  "INVALID (100)",
+		},
+	}
+
+	for _, test := range tests {
+		have := test.level.String()
```
electronjoe, line 51: nit / question: I was trained to use `got` not `have`. Do you see preference in open source for have?

src/go/log/log_test.go
```diff
@@ -0,0 +1,180 @@
...
+		},
+		"duplicate names": {
+			names: []string{"a", "a", "a"},
+			want:  "a.a.a",
+		},
+	}
+
+	for desc, test := range tests {
+		have := FullName(test.names)
+		assert.Equal(t, test.want, have, desc)
```
electronjoe, line 80: Particularly in table driven testing: I was trained to log the function call / parameters that led to failure. E.g. here something like `FullName([]string{"a"}) = "b", want "a"` - the goal I believe was to enable as much context and debug as possible exclusively from logs without having to re-run the test to output context on the function called / its inputs?

src/go/log/log_test.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+}
+
+//go:generate go run github.com/golang/mock/mockgen -package log -destination mock_logger_test.go . Logger
```
electronjoe, line 84: Is this accidentally left over from dev?

src/go/log/log_test.go
```diff
@@ -0,0 +1,180 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package log
+
+import (
+	"testing"
+
+	"github.com/golang/mock/gomock"
```
electronjoe, line 17: This is another best practice maybe? that I don't see in our internal (so far) `Golang Best Practices` authored by @hcgatewood

src/go/log/log_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/stretchr/testify/assert"
```
electronjoe, line 18: Ditto best practice here. Allows asserts in tests to continue test execution / summary output? I think maybe? Generally explanation is helpful somewhere in best practices maybe? @hcgatewood

src/go/accessd/main.go
electronjoe: OK, so now I think I get it. `src/go/log/zap/zap.go` is our implementation of our new `src/go/log/log.go` generalized logging interface, backed by `go.uber.org/zap`.  I'm not entirely sure how to improve the understandability here (it's not bad when you look at all the source code, but a newcomer may not do that regularly - and in a linear review of this PR the picture comes out of order). But I suppose so long as they copy pasta the logging set-up and whatnot, maybe they don't need to get the plumbing. I still think a doc on how we are doing logging in e.g. Golang Best Practices may be quite helpful.

src/go/log/zap/zap.go
```diff
@@ -0,0 +1,156 @@
...
+
+type Logger struct {
+	names []string
+	*zap.Logger
+	zap.Config
+
+	printers map[log.Level]*printer
+}
+
+var _ log.Logger = (*Logger)(nil)
```
electronjoe, line 56: I am not piecing together at a glance what is happening here. Will try to reconstruct by looking at references to log.Logger.

electronjoe: Nope, not getting it =D I see no log package .Logger object? I see a Logger within the `log.ManagedLogger` type. How is it that `log.Logger` here is setting the logger pointer inside the ManagedLogger? I'm missing something.

electronjoe: Secondary question is how does this presently-nil pointer get initialized to our zap logger? I see nothing in `log/zap/zap.go` doing so.

src/go/protos/protos.go
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package protos
+
+//go:generate go install google.golang.org/protobuf/cmd/protoc-gen-go
```
electronjoe, line 14: I've not used `go generate` - are these commented out lines actually acted upon by go tooling? Or are they here to describe what needs to be executed by a shell script / Makefile / by hand for notekeeping purposes?

src/go/service/sctpd/proxy_uplink_server.go
```diff
@@ -0,0 +1,63 @@
...
+
+	log.Logger
+	*sctpd.UnimplementedSctpdUplinkServer
+}
+
+func NewProxyUplinkServer(logger log.Logger, cc *grpc.ClientConn) *ProxyUplinkServer {
+	return &ProxyUplinkServer{Logger: logger, client: sctpd.NewSctpdUplinkClient(cc)}
+}
+
+func (p *ProxyUplinkServer) SendUl(ctx context.Context, req *sctpd.SendUlReq) (*sctpd.SendUlRes, error) {
```
electronjoe: My first question(s) are whether this is a blocking call or not - would be helpful for SendUl godoc to specify blocking or nonblocking nature.  Second question generally is the thread safety of the ProxyUplinkServer generally (can anybody bang on SendUl from whatever context they chose?). This topic might belong as godoc on ProxyUplinkServer?

src/go/tools.go
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build tools
+// +build tools
+
+package tools
+
+import (
+	_ "github.com/golang/mock/mockgen"
```
electronjoe, line 20: Not sure what this file is doing generally `src/go/tools.go` - can you briefly explain for me? Do these tree modules require initialization for general use, and so you group them here to simplify import by other? or..

electronjoe (approved):
Context: You are clearly a far more experienced Go programmmer than I am. But I attempted to regurgitate review feedback I had during my Golang readability at Google. I may mis-recall some of it here (or best practice in Google may not apply in our environment).

Unfortunately, GH Review doesn't allow annotations to indicate "question / suggestion" versus "I think this needs to be done".  I tend to dump a lot of questions and thoughts that may not meet the bar of acting upon. I am hoping for education / discussion / future plans that may not be clear presently.  I hope my review comments make it clear what a newcomer to the code might be asking / thinking.

src/go/accessd/main.go
markjen: two features we want: 1. nested/scoped named loggers, 2. parameterized/k=v logging

the first we want because we want to be able to control log-level for various internal components. the second we want because we will one day (today?) need to elide PII in logs shared with us (e.g. IMSI)

note: we want log level changable at runtime for a particular scope, without restarting the service. glog/klog don't have the idea of a logger tree, so only provide global log level. if using glog/klog, we'd have to rebuild these functions.

lastly, zap claims to have very fast performance, 98% code coverage, and is an active project :)

src/go/accessd/server/server.go
markjen: yes i am a fan of package level comments; I'll add them!

src/go/accessd/server/server.go
markjen: I think we've got golangci-lint going, which I've had an OK experience with :)

src/go/accessd/main.go
themarwhal: Awesome! Thanks for that. The PII thing was the biggest hurdle I faced when trying to enable sentry for our Python services. We don't scope most of our logs over there, so it made it very hard to filter to reduce load on the error level logs and scrape PII. :/

src/go/accessd/server/server.go
markjen: Re: errors pkg, I'm happy with whatever; I'm not sure which one github's error package is though. For even more serious business, check out https://github.com/cockroachdb/errors

Re: thinking behind dependencies, should I add it to a README.md? is that the right place for it?

src/go/accessd/server/server.go
markjen: this is upcoming :)

WIP here https://github.com/markjen/magma/commit/b33680c0b2989963a8c0ac18486ba121a4ec488a

src/go/accessd/server/server.go
markjen: yeah, I removed this in the next commit (not yet put up for PR)

https://github.com/markjen/magma/commit/5b83ca0825f62bcfbb4c1a77a7ea125dbf94fe31

src/go/log/log_test.go
electronjoe: Or is this actually acted on by go tooling. I am pre-noob on `go generate`...

src/go/accessd/server/server.go
electronjoe: I think that README.md is a good starting place to rationalize your dependency selection - get it down on paper at least. I like code-adjascent documentation like that, though I'm not sure everybody here finds it as discoverable (preference towards Magma website it seems).

src/go/accessd/server/server.go
markjen: I removed the log.fatal func because 1. it should be used rarely, 2. IMO explicit panic is more clear, 3. panic already dumps a stack and prints the values; usually no need for more machinery.

Once we have crash handling, sentry.io should grab these and send them up; I don't think adding a log package in the middle helps too much :)

src/go/log/log.go
markjen: I believe in following effective go, just didn't get around to it :)

I'll add comments on all public items

src/go/log/log.go
markjen: there are two cases for panic:

1. un-recoverable initialization error
2. logically impossible condition / assert

this is an example of the latter.

src/go/log/log_test.go
markjen: not sure where I picked up `have`, will go for `got`

src/go/log/log_test.go
markjen: ah no; it's there so that running `go generate` will re-gen the mock interface if necessary

I'm thinking of moving all generate directives to either a separate file (`gen.go`, `log_gen.go`) or the package level file, e.g. `log.go`, (next to the package-level documentation)

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+	service_sctpd "github.com/magma/magma/service/sctpd"
+)
+
+const (
+	mmeUpstreamTarget  = "unix:///tmp/mme_sctpd_upstream.sock"
+	mmeGrpcDialTimeout = time.Second
+)
+
+func newMmeGrpcConn() (*grpc.ClientConn, error) {
+	ctx, _ := context.WithTimeout(context.Background(), mmeGrpcDialTimeout)
```
mstre123: looking to the future if we use fx one of the nice things will be to use the injected context rather than context.Background so that fx can control the lifecycle

src/go/log/log_test.go
markjen: I think there's currently use of `mockery`, but that lib is more relaxed on expectations and type checking:

https://blog.codecentric.de/2019/07/gomock-vs-testify/#expectations

Thus, I lean towards gomock, but am open to sticking with testify/mockery if we want

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/pkg/errors"
+	"google.golang.org/grpc"
+
+	"github.com/magma/magma/log"
+	"github.com/magma/magma/protos/magma/sctpd"
+	service_sctpd "github.com/magma/magma/service/sctpd"
+)
+
+const (
+	mmeUpstreamTarget  = "unix:///tmp/mme_sctpd_upstream.sock"
```
mstre123, line 28: another plug for fx - these can all be defined in an injected config

src/go/protos/protos.go
mstre123: these aren't comments, they're instructions for go generate
https://go.dev/blog/generate

src/go/log/log_test.go
markjen: ah, `assert` is a misnomer here.

the `testify/assert` library is a reflection/syntactic sugar library to help turn this:

```
if  got != want {
    t.Errorf("X(%s) = %s, want=%s", input, got, want)
}
```

into this:

```
assert.Equal(t, want, got, "X(%s)", input)
```

The first construction uses the compiler to type-check `got` and `want`, but the latter is one line.

Years ago, I used to be a stickler for the first construction ("if golang authors haven't added it to the lang/stdlib, then we don't need it either!"), but the overwhelming majority of folks prefer the latter and it's just for tests so I've relented.

To be fair, a huge block of `if X != Y { ... }` is much harder on the eyes than a set of `assert.Equal(...)` one-liners.

src/go/log/log.go
```diff
@@ -0,0 +1,136 @@
...
+	// production.
+	DebugLevel Level = iota - 1
+	// InfoLevel is the default logging priority.
+	InfoLevel
+	// WarnLevel logs are more important than Info, but don't need individual
+	// human review.
+	WarnLevel
+	// ErrorLevel logs are high-priority. If an application is running smoothly,
+	// it shouldn't generate any error-level logs.
+	ErrorLevel
```
mstre123, line 99: out of interest is there a reason we're not introducing fatals, panics etc.
https://github.com/uber-go/zap/blob/master/zapcore/level.go#L35-L52

src/go/log/zap/zap.go
markjen: this is a golang idiosyncratic compile-type interface assertion.

https://commaok.xyz/post/compile-time-assertions/

note that not everyone likes these because some prefer conformance to an interface to be optional by design. i.e. if it's necessary for a struct to conform to an interface, then it'll be used as such somewhere else in the code. If you "accidentally" stop conforming, but your tests pass, then there really isn't a problem.

src/go/protos/protos.go
markjen: https://go.dev/blog/generate

yeah, this is a `go` tool built-in that does simple one-liners to re-gen code since that's the preference of the original golang authors.

src/go/service/sctpd/proxy_uplink_server.go
markjen: ah, that's kind of a grpc server impl detail, but by default all unary calls are concurrent/blocking.

the server spins a goroutine for each inbound request and executes the provided implementation.

this "simple" pattern is the recommended default for all our stuff going foward: do your work in a linear/blocking fashion and count on the underlying server to spin goroutines to allow for parallel processing (vs. MME's current strategy of attempting to serialize work in queues; this is somewhat like concurrent vs. co-operative threading patterns)

src/go/tools.go
markjen: https://github.com/golang/go/issues/25922

this is the convention for getting `go mod tidy` to work when you have dev tools that are not used by code.

src/go/accessd/server/server.go
markjen: yeah, playing with fx is on my to-do list. will take a look at that next after fixing up a few diffs

src/go/log/log.go
markjen: we definitely do not want `os.Exit` since that will not allow a crash reporter to intercept.

I think we should use panic for unrecoverable or "logically impossible" assert conditions.

src/go/accessd/server/server.go
electronjoe: Will people notice easily in our log outputs - that a panic occurred? I worry they may need to synthesize multiple streams of data (e.g. sentry.io or syslog + logs) to know what happened when? Or no.

src/go/log/log.go
electronjoe: Agree with the scenarios in which immediate halt is necessary, but was unsure on panic() versus having logging do it - but I already asked this question above so lets drop this dupe =D Your point here is a good one regardless.

src/go/log/log_test.go
electronjoe: To confirm this I actually found documentation with this phrasing on the [golang github repo](https://github.com/golang/go/issues/34384) ctrl-f `Got Before Want`.

src/go/log/log_test.go
electronjoe: I just spotted that this advice is also on the [golang github repo](https://github.com/golang/go/issues/34384) FWIW (ctrl-f for `Identify the Function` and `Identify the Input`).

src/go/log/log_test.go
electronjoe: Hopefully others are less noob-y than I on go generate and wouldn't be confused =D But moving them sounds good too. I don't know the pros and cons.

src/go/log/log_test.go
electronjoe: Leaving discussion to you and @hcgatewood  =D

src/go/log/log_test.go
electronjoe: Assert library use is discussed also in [golang github repo](https://github.com/golang/go/issues/34384) - they don't seem to like it but that seems to align with what you say here about the `Go way of doing things the long way` - I have no horse in this "race" =D Just thought I would note.

src/go/log/zap/zap.go
electronjoe: I need a `far out, dude` emoji =D

src/go/accessd/server/server.go
electronjoe: @mstre123 discussion here related to your question on the lack of Fatal on logging interface below.

src/go/log/log.go
electronjoe: So in a reply elsewhere, Mark indicated that this is to allow ... substream? differentiation by key for e.g. PII filtering? Am I parroting that right? I'm not totally clear I guess. Would it be a log stream per PII sensitive emitter? Allowing you to fetch just the log stream for one user? Or..

src/go/accessd/server/server.go
uri200: just in case you wanted to add linter. Here you have how it is at feg

http://github.com/magma/magma/blob/cdc9dd678ccb35e9dd24c8b9a4aefc2894444dba/feg/gateway/Makefile#L58-L58

src/go/protos/protos.go
uri200: maybe it would be good to follow the same system we use in orc8r. It will save time and will make the system homogeneous between gateways.

We trigger generation of all protos with `./build.py -g` at `~/magma/orc8r/cloud/docker`. That command runs the following steps 
- Creates a build pod and run a command [here](http://github.com/magma/magma/blob/cdc9dd678ccb35e9dd24c8b9a4aefc2894444dba/orc8r/cloud/docker/build.py#L88-L88) 
- Then we run `make fullgen` command inside the a pod we use for building or8cr [here](http://github.com/magma/magma/blob/cdc9dd678ccb35e9dd24c8b9a4aefc2894444dba/orc8r/cloud/Makefile#L117-L117)
- That command will run a `make gen` on all the `cloud/go` folders in our system. For example for feg is [this](http://github.com/magma/magma/blob/cdc9dd678ccb35e9dd24c8b9a4aefc2894444dba/feg/cloud/go/Makefile#L2-L2)
- which finaly runs [gp generate ./...](http://github.com/magma/magma/blob/cdc9dd678ccb35e9dd24c8b9a4aefc2894444dba/orc8r/cloud/go/module.mk#L32-L32)
- The protos ends up under the cloud/go/proto folder

That command generates all the protos used in magma for go. FEG feeds from those. So if you create a `cloud/go` folder in AGW We could do the same for AGW and centralize.

Note that protos are shared between agw feg and orc8c so it makes sense to have one single place to generates then than 3. 

Note that this command also generates swagger modules. So once we have swagger on AGW that will come for free

markjen: a few things to note here:

1. we're trying hard to not have local go.mod rewrites that go outside of `src/go`
2. generate/build/etc. should be fast, even from scratch. with go, there's no reason to need a VM/docker build appliance; developers should be able to do most of their development in their native environment if desired (macos, linux, windows, etc.). this yields the best/fastest dev loops and reduces "dev infra" setup/maintenance time
3. the go namespace directives in our proto files appear to be all fubar (although this would be more a point towards moving everything to src/go instead of spread out across <box>/<area>/<deployment env>/<programming language> directories)

and also, 4. I think this entire area is temporary while we figure out bazel :)

src/go/accessd/server/server.go
markjen: sweet, @electronjoe has this PR about to land to get things going https://github.com/magma/magma/pull/8942

seeing if we can get by without a makefile for now, but a few aliases sure would be nice

src/go/protos/protos.go
uri200: with the rest of magma we generate the protos using `build.py -g` command on  `~/magma/orc8r/cloud/docker`. The process it follows is 

- It runs make `make fullgen` inside a build pod [here](http://github.com/magma/magma/blob/cdc9dd678ccb35e9dd24c8b9a4aefc2894444dba/orc8r/cloud/docker/build.py#L88-L88)
- `fullgen` runs `gen_protos` which runs `make gen_protoss` in all the Makefiles under `/cloud/go` [here](http://github.com/magma/magma/blob/cdc9dd678ccb35e9dd24c8b9a4aefc2894444dba/orc8r/cloud/Makefile#L117-L117)
-  In feg this Maklefile imports a common library [here](http://github.com/magma/magma/blob/cdc9dd678ccb35e9dd24c8b9a4aefc2894444dba/feg/cloud/go/Makefile#L2-L2)
- The `make gen` [here](http://github.com/magma/magma/blob/cdc9dd678ccb35e9dd24c8b9a4aefc2894444dba/orc8r/cloud/go/module.mk#L32-L32) just do a `go generate ./...` and drops the pb files under the proto folder 


Note protos are shared between gateways, so it would be good we had a single point to generate them as we do right now

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+
+	mmeGrpcConn, err := newMmeGrpcConn()
+	if err != nil {
+		panic(err)
+	}
+	sctpdUplinkServer := service_sctpd.NewProxyUplinkServer(logger, mmeGrpcConn)
+	sctpd.RegisterSctpdUplinkServer(grpcServer, sctpdUplinkServer)
+
+	return grpcServer
+}
```
uri200, line 49: there are too many comments so it is difficult for me to understand what it is here, but it looks like you are creating a GRPC server.

Why don't we use/extend the orc8r framework to create services? We don't have to worry for config files or how to identify each service. it is all included on the framework see [here](http://github.com/magma/magma/blob/cdc9dd678ccb35e9dd24c8b9a4aefc2894444dba/feg/gateway/services/s8_proxy/s8_proxy/main.go#L33-L33) how we use it in feg to create s8_proxy grpc server.

This framework also allows us to create GRPC test services. This way we don't need to inject mocks inside the code. We just build a complete mock service. For example that is what we did [here](http://github.com/magma/magma/blob/cdc9dd678ccb35e9dd24c8b9a4aefc2894444dba/feg/gateway/services/aaa/servicers/accounting_test.go#L41-L41) were the AAA server communicates with pipelined through grpc. In that case, instead of having to make the client injectable for testing and polluting the code, we just created our *own mock pipelined grpc service* and make it work like it would work on production.

src/go/accessd/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+
+const (
+	sctpdUpstreamNetwork = "unix"
+	sctpdUpstreamPath    = "/tmp/sctpd_upstream.sock"
+)
```
uri200, line 54: we normally place const at the top of the file.

src/go/accessd/server/server.go
electronjoe: Regarding the too many comments @uri200, hit the `i` key and it will toggle all comments!

src/go/accessd/server/server.go
markjen: these strings are all gone in an upcoming PR (moving to config)

src/go/accessd/server/server.go
markjen: will look at the orc8r code for creating services, but after we figure out the DI strategy (currently doing some manual DI, but exploring Fx next)

src/go/accessd/server/server.go
markjen: added README and some package level comments; if we graduate from expiremental status, then we'll put up magma website docs also

PR #8930, opened 2021-09-01
fix(pipelined): Handle ovs poll multi reply

koolzz:

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/enforcement_stats.py
+++ b/lte/gateway/python/magma/pipelined/app/enforcement_stats.py
@@ -727,19 +727,23 @@ class EnforcementStatsController(PolicyMixin, RestartMixin, MagmaController):
         try:
             response = ofctl_api.send_msg(
                 self, message, reply_cls=parser.OFPFlowStatsReply,
-                reply_multi=False,
+                reply_multi=True,
             )
-            if response == None:
+            if not response:
                 self.logger.error("No rule records match the specified cookie and cookie mask")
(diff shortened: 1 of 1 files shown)


lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -727,19 +727,23 @@ def get_stats(self, cookie: int = 0, cookie_mask: int = 0):
...
-                self.loop.call_soon_threadsafe(self._delete_old_flows, usage.values())
-                record_table = RuleRecordTable(
-                    records=usage.values(),
-                    epoch=global_epoch,
-                )
-                return record_table
+
+            aggregated_msgs = []
+            for r in response:
+                aggregated_msgs += r.body
```
pshelar, line 738: can you iterate over the response list in the _get_usage_from_flow_stat() function.
Just to avoid one extra iteration.

koolzz: I'd rather not because it makes that function a bit too verbose(too many indents) I could add a wrapper that calls the function but then I have to pass around a dict which I don't like

pshelar: ok, what is typical size of the response list

koolzz: should be single element most of the time

pshelar (approved):
LGTM

PR #8949, opened 2021-09-03
chore(mme): Migrate bearer activation and deactivation timers to zmq

ulaskozat:
## Summary

As the final part of migrating to zmq for NAS timers, network initiated default and dedicated bearer activation/deactivation timers are completed. Since these timers need ebi in addition to the s1ap mme id, a new struct for holding timer argument and methods to retrieve it are also added. 

There were couple of clean ups as we no longer pass calloced arguments to the timers.

## Test Plan

S1AP integration tests that already covers these procedures with expired timers and service restarts.
Unit tests will be added in the next couple of sprints for all the procedures including timer expiration cases.

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/include/mme_app_ue_context.h
+++ b/lte/gateway/c/core/oai/include/mme_app_ue_context.h
@@ -48,6 +48,7 @@
 #include "emm_data.h"
 #include "esm_data.h"
 #include "emm_cnDef.h"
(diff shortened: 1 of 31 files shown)


lionelgo (approved):
We may change the name of nas_timer_t type to something else, now this timer type is used by other layers.

PR #8972, opened 2021-09-07
feat(agw): Unit test cases for roaming UE

rsarwad:
feat(agw): Unit test cases for roaming UE

## Summary
1. Added unit test cases for roaming ue attach 
2. Fixed linking error inconsistently occurring in common module due to dependency on openssl and crypto modules
## Test Plan


lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
```diff
@@ -0,0 +1,354 @@
...
+  EXPECT_EQ(
+      sgw_update_teid_in_ue_context(
+          sgw_state, imsi64, sgw_s11_tunnel.local_teid),
+      RETURNok);
+  EXPECT_EQ(
+      hashtable_ts_get(
+          sgw_state->imsi_ue_context_htbl, (const hash_key_t) imsi64,
+          reinterpret_cast<void**>(&ue_context_p)),
+      HASH_TABLE_OK);
+
```
uri200: is it possible to check the value inserted on `ue_context_p->sgw_s11_teid_list`  is equal to `sgw_s11_tunnel.local_teid`?

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
```diff
@@ -0,0 +1,354 @@
...
```
uri200, line 175: line 159 to 175 seems to be the same in all the test. What about creating a function so all the test can reuse it?

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
```diff
@@ -0,0 +1,354 @@
...
+  sgw_eps_bearer_context_information_t* sgw_pdn_session = NULL;
+  sgw_pdn_session = sgw_create_bearer_context_information_in_collection(
+      sgw_s11_tunnel.local_teid);
+  EXPECT_TRUE(sgw_pdn_session != nullptr);
```
uri200, line 183: since those test run only one single function, it may be good to check the values of the hashtables that modifies and the content of the `sgw_eps_bearer_context_information_t` list

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
```diff
@@ -0,0 +1,354 @@
...
+  itti_s11_create_session_request_t session_req = {0};
+  fill_itti_csreq(&session_req);
+  sgw_s11_tunnel.remote_teid = session_req.sender_fteid_for_cp.teid;
+  EXPECT_EQ(
+      sgw_update_bearer_context_information_on_csreq(
+          sgw_state, sgw_pdn_session, sgw_s11_tunnel, &session_req, imsi64),
+      RETURNok);
+  sgw_state_exit();
+}
```
uri200, line 228: Lets discuss this next week but I think in order to reduce the amount of code/tests and put more effort on more complex functional test, lets keep only `create_bearer_within_pdn_session_cs_req` which includes, `create_pdn_session_cs_req` and `create_ue_context_cs_req`. We can set it as `create session on cs rsp`.

I know that in that case we may not be doing a Unit test per se. But to do a real unit test to test 
`create_ue_context_cs_req` we should not depend on any previous function (we should create a mock object with the information already on it)
 
Lets spend more  creating functional  complex tests that simulate the call flow of those functions.
The reason is we want to take advantage of Unit test to run some kind of small integration test too. We do the same with sessiond where our unit test are more small integ test between features

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
```diff
@@ -0,0 +1,354 @@
...
+  EXPECT_EQ(
+      hashtable_ts_get(
+          sgw_state->imsi_ue_context_htbl, (const hash_key_t) imsi64,
+          reinterpret_cast<void**>(&ue_context_p)),
+      HASH_TABLE_OK);
+
+  sgw_eps_bearer_context_information_t* sgw_pdn_session = NULL;
+  sgw_pdn_session = sgw_create_bearer_context_information_in_collection(
+      sgw_s11_tunnel.local_teid);
+  EXPECT_TRUE(sgw_pdn_session != nullptr);
```
uri200: can you check the content of `sgw_pdn_session`? 
I think it is a hash table so we should check local_teid is there

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
```diff
@@ -0,0 +1,354 @@
...
+  itti_s11_create_session_request_t session_req = {0};
+  fill_itti_csreq(&session_req);
+  sgw_s11_tunnel.remote_teid = session_req.sender_fteid_for_cp.teid;
+  EXPECT_EQ(
+      sgw_update_bearer_context_information_on_csreq(
+          sgw_state, sgw_pdn_session, sgw_s11_tunnel, &session_req, imsi64),
+      RETURNok);
```
uri200, line 200: can you check some of the fields that have changed for `new_sgw_eps_context`. 
For example you could check if `new_sgw_eps_context->pdn_connection.apn_in_use == strdup("NO APN")`

Checking those fields will help to make sure if we ever change something and break it, the test will fail

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
```diff
@@ -0,0 +1,354 @@
...
+  fill_itti_csreq(&session_req);
+
+  sgw_s11_tunnel.remote_teid = session_req.sender_fteid_for_cp.teid;
+  EXPECT_EQ(
+      sgw_update_bearer_context_information_on_csreq(
+          sgw_state, sgw_pdn_session, sgw_s11_tunnel, &session_req, imsi64),
+      RETURNok);
+
+  EXPECT_TRUE(
+      (sgw_get_sgw_eps_bearer_context(sgw_s11_tunnel.local_teid)) != nullptr);
```
uri200, line 247: same here: can we check the value `sgw_s11_tunnel.local_teid` 

For the rest of the test, we should be checking more than just the object not being nullptr. We need to make sure the information is proper

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
```diff
@@ -0,0 +1,354 @@
...
+  sgw_state_init(false, config);
+  sgw_state_t* sgw_state          = get_sgw_state(false);
+  spgw_ue_context_t* ue_context_p = NULL;
+
+  mme_sgw_tunnel_t sgw_s11_tunnel = {0};
+  sgw_s11_tunnel.local_teid       = sgw_s8_generate_new_cp_teid();
+  EXPECT_EQ(
+      sgw_update_teid_in_ue_context(
+          sgw_state, imsi64, sgw_s11_tunnel.local_teid),
+      RETURNok);
```
uri200: for those cases where we are reusing already tested functions, we can remove all those `EXPECT_EQ`

Since the code is exaclty the same we can consider this is already testd. It will reduce the amount of code and make the test easy to be modified

so in this case you could have 

```

  sgw_state_init(false, config);
  sgw_state_t* sgw_state          = get_sgw_state(false);
  spgw_ue_context_t* ue_context_p = NULL;

  mme_sgw_tunnel_t sgw_s11_tunnel = {0};
  sgw_s11_tunnel.local_teid       = sgw_s8_generate_new_cp_teid();       
  sgw_update_teid_in_ue_context(sgw_state, imsi64, sgw_s11_tunnel.local_teid);
  hashtable_ts_get(sgw_state->imsi_ue_context_htbl, (const hash_key_t) imsi64,       
 reinterpret_cast<void**>(&ue_context_p));

  sgw_eps_bearer_context_information_t* sgw_pdn_session = NULL;
  sgw_pdn_session = sgw_create_bearer_context_information_in_collection(
      sgw_s11_tunnel.local_teid);

  itti_s11_create_session_request_t session_req = {0};
  fill_itti_csreq(&session_req);

  sgw_s11_tunnel.remote_teid = session_req.sender_fteid_for_cp.teid;

  sgw_update_bearer_context_information_on_csreq(sgw_state, sgw_pdn_session, sgw_s11_tunnel, &session_req, imsi64);
....
```

uri200 (approved):

ulaskozat (approved):
LGTM

PR #8979, opened 2021-09-08
fix(agw): Misc minor eNB fixes

amarpad:
## Summary
Some minor eNBd fixes
1 - Fix log file location
2 - Only type convert GPS Status if it is not already bool
3 - Handle PUTs and POST, sercomm eNB reports stats as PUTS and Baicells as POST.

## Test Plan

- python tests
- Managing sercomm radios

## Additional Information


ardzoht (approved):
lgtm

PR #9018, opened 2021-09-09
fix(mme): removed a few compilation warnings

rdefosse:

## Summary

5G Pull Requests are incorporating a lot of compilation warnings.

I trimmed:

- `Ubuntu18` image --> 22 warnings downto 6
- `RHEL8` image --> 8 warnings downto 3

## Test Plan

None.

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
+++ b/lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
@@ -662,7 +662,7 @@ void amf_app_handle_pdu_session_response(
     amf_sap.u.amf_as.u.establish.guti = ue_context->amf_context.m5_guti;
     rc                                = amf_sap_send(&amf_sap);
     if (RETURNok == rc) {
-      ue_context->mm_state == REGISTERED_CONNECTED;
+      ue_context->mm_state = REGISTERED_CONNECTED;
     }
   } else {
     OAILOG_DEBUG(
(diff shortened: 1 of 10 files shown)


lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
```diff
@@ -1215,9 +1215,8 @@ static int paging_t3513_handler(zloop_t* loop, int timer_id, void* arg) {
         LOG_AMF_APP,
         "T3513: Maximum retires done hence Abort the Paging Request "
         "procedure\n");
-    OAILOG_FUNC_RETURN(LOG_NAS_AMF, RETURNok);
   }
-  OAILOG_FUNC_RETURN(LOG_NAS_AMF, RETURNok);
+  OAILOG_FUNC_RETURN(LOG_NAS_AMF, rc);
```
lionelgo: rc is initialized to RETURNerror, if you do not enter the if, here your modification change the returned value of function, I did not dig into the consequence.

lionelgo (commented):
Just left a comment

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
ulaskozat: This should always return okay otherwise zmq loop will exist treating this as the intent. I suggest that we remove rc completely and return Ok always.

ulaskozat (changes requested):
see zmq comment.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
rdefosse: @ulaskozat 

I agree w/ your comment. My point I wanted to make w/ this PR is that we are accepting code in w/o paying attention to compilation warnings.

One of them was quite important --> `ue_context->mm_state == REGISTERED_CONNECTED;`

I was away for a few weeks and the number rises to `22`....

For me, this is not acceptable that every 2 months, I have to do such a PR to fix other's people work.

Raphael

ulaskozat:
@panyogesh @sureshgorijavolu, can you take care of compiler warnings on the 5G path?

ulaskozat (approved):
LGTM

PR #9022, opened 2021-09-09
refactor(ci): move to uniform GH workflow naming for linters

electronjoe:
As requested by @quentinDERORY.

Changed workflow NAME for linters to drop the ReviewDog references and just name the tool being run.

While here, re-named all the workflow files to match the workflow NAME.


quentinDERORY (approved):
Thanks for doing that :)

quentinDERORY:
Let's use #9032 instead of renaming the jobs.

Issue #8820, opened 2021-08-27
[AGW][vN] Setup CI for src/go


markjen:

electronjoe:
Basic functionality is complete. May generate a follow-up sprint task for ARM support.

PR #9030, opened 2021-09-09
ci: turn up go test / build coverage for src/go

electronjoe:
Working towards #8820.

Separate jobs to break apart `go build ./...` and `go test -race ./...`.

Additionally, test coverage is generated and uploaded to codecov (see logs from Checks for the final Codecov.io URL after analysis - or view results post merge to master).

The workflow will always execute, which is required by GitHub in order to label any tasks as Required for PR submission.

The job determinator cribbed from @themarwhal's #9032 enables the job to no-op when no go code is touched.

### DIFF
--- a/.github/workflows/golang-build-test.yml
+++ b/.github/workflows/golang-build-test.yml
@@ -1,0 +1,92 @@
+---
+name: Golang Build & Test
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+
+jobs:
+  pre_job_src_go_determinator:
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          # Only run the main job for changes including the following paths
+          paths: '[".github/workflows/golang-build-test.yml", "src/go/**"]'
+
+  build_src_go:
+    needs: pre_job_src_go_determinator
+    if: ${{ needs.pre_job_src_go_determinator.outputs.should_skip == 'false' }}
+    strategy:
+      matrix:
+        go-version: [1.17.x]
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Test
+        run: |
+          cd src/go/
+          go build ./...
+
+  test_src_go:
+    needs: pre_job_src_go_determinator
+    if: ${{ needs.pre_job_src_go_determinator.outputs.should_skip == 'false' }}
+    strategy:
+      matrix:
+        go-version: [1.17.x]
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Setup gotestsum
+        uses: autero1/action-gotestsum@v1.0.0
+        with:
+          gotestsum_version: 1.7.0
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Test
+        run: |
+          cd src/go/
+          gotestsum --format=testname --junitfile test_src_go.xml -- -race ./...
+      - name: Upload Test Results
+        uses: actions/upload-artifact@v2
+        with:
+          name: Unit Test Results
+          path: "${{ github.workspace }}/src/go/test_src_go.xml"
+
+  codecov_src_go:
+    needs: pre_job_src_go_determinator
+    if: ${{ needs.pre_job_src_go_determinator.outputs.should_skip == 'false' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Test
+        run: |
+          cd src/go/
+          go test -race -coverprofile=coverage.txt -covermode=atomic ./...
+      - name: Codecov.io Upload
+        uses: codecov/codecov-action@v2
+        with:
+          flags: src_go
+          fail_ci_if_error: true
+          verbose: true
--- a/.github/workflows/unit-test-workflow.yml
+++ b/.github/workflows/unit-test-workflow.yml
@@ -6,6 +6,7 @@ on:  # yamllint disable-line rule:truthy
       - cloud-workflow
       - feg-workflow
       - agw-workflow
+      - golang-build-test
     types:
       - completed
 


.github/workflows/src-go-test.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Test
+        run: |
+          cd src/go/
+          go test -race ./...
```
themarwhal: In other Go components, we use [testsum](https://github.com/gotestyourself/gotestsum) to run tests and nicely generate test summaries. Might be worth considering if we want to generate test summaries and upload it to GitHub. :)

.github/workflows/src-go-test.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Test
+        run: |
+          cd src/go/
+          go test -race -coverprofile=coverage.txt -covermode=atomic ./...
+      - name: Codecov.io Upload
+        uses: codecov/codecov-action@v2
+        with:
+          flags: src_go
+          fail_ci_if_error: true
+          verbose: true
```
markjen: nit: missing newline :)

.github/workflows/src-go-test.yml
```diff
@@ -0,0 +1,84 @@
...
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          # Only run the main job for changes including the following paths
+          paths: '[".github/workflows/golang-build-test.yml", "src/go/**"]'
```
markjen: does `golang-build-test.yml` need to be `src-go-test.yml` to match this file?

markjen (approved):

.github/workflows/src-go-test.yml
electronjoe: Thanks for the pointer!

.github/workflows/src-go-test.yml
electronjoe: I believe YAMLlint wants no hanging newline, let me check!

.github/workflows/src-go-test.yml
electronjoe: Thanks for catching that!

.github/workflows/src-go-test.yml
electronjoe: Strange, no YAMLLint warnings with the final newline. Or without it. Wonder what I've been seeing...

.github/workflows/golang-build-test.yml
```diff
@@ -0,0 +1,88 @@
...
+      - name: Setup gotestsum
+        uses: autero1/action-gotestsum@v1.0.0
+        with:
+          gotestsum_version: 1.7.0
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Test
+        run: |
+          cd src/go/
+          gotestsum --format=testname -- -race ./...
```
themarwhal: You can add `--junitfile test_src_go.xml` and then upload that result like this:
https://github.com/magma/magma/blob/master/.github/workflows/cloud-workflow.yml#L151

.github/workflows/golang-build-test.yml
```diff
@@ -0,0 +1,88 @@
...
+          go-version: ${{ matrix.go-version }}
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Test
+        run: |
+          cd src/go/
+          go build ./...
+
+  test_src_go:
+    if: ${{ needs.pre_job_src_go_determinator.outputs.should_skip != 'true' }}
```
themarwhal: `if: ${{ needs.pre_job_src_go_determinator.outputs.should_skip == 'false' }}` ? :p

.github/workflows/golang-build-test.yml
themarwhal: Screen Shot 2021-09-10 at 6 00 34 AM

That will trigger these auto-updating comments

.github/workflows/golang-build-test.yml
```diff
@@ -0,0 +1,93 @@
...
+    strategy:
+      matrix:
+        go-version: [1.17.x]
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
themarwhal, line 49: this is awesome! :D didn't realize this was supported!

themarwhal (approved):
Thanks for adding these! Especially cool that it tests on multiple platforms.
Left some comments about how to upload test results to GitHub (it's entirely up to you whether we want to do this or not)

Some general questions regarding how we name workflows. We currently seem to have a mix of snake case? and regular title-like names. It might be good stylistically to merge on one pattern. @quentinDERORY any thoughts?

.github/workflows/golang-build-test.yml
electronjoe: Yeah! And in fact we found a test-code bug thanks to the windows-latest! Sadly though they don't yet support ARM out of the box simply (GitHub doesn't). We may add ourselves...

.github/workflows/golang-build-test.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Upload Test Results
+        uses: actions/upload-artifact@v2
+        with:
+          name: Unit Test Results
+          path: "${{env.MAGMA_ROOT}}/src/go/test_src_go.xml"
+
+  codecov_src_go:
+    if: ${{ needs.pre_job_src_go_determinator.outputs.should_skip == 'false' }}
+    strategy:
+      matrix:
```
electronjoe: Don't actually want matrix here, I need to drop it.

electronjoe:
> Some general questions regarding how we name workflows. We currently seem to have a mix of snake case? and regular title-like names. It might be good stylistically to merge on one pattern. @quentinDERORY any thoughts?

I believe @hcgatewood had an opinion that we prefer `-` to `_` due to e.g. Python reasons? So I was aiming for `-` in workflow names.  But yeah, we need to standardize. Edit: oops, opposite! [Prefer underscores over hyphens](https://docs.magmacore.org/docs/contributing/contribute_conventions).

themarwhal:
> > Some general questions regarding how we name workflows. We currently seem to have a mix of snake case? and regular title-like names. It might be good stylistically to merge on one pattern. @quentinDERORY any thoughts?
> 
> I believe @hcgatewood had an opinion that we prefer `-` to `_` due to e.g. Python reasons? So I was aiming for `-` in workflow names. But yeah, we need to standardize. Edit: oops, opposite! [Prefer underscores over hyphens](https://docs.magmacore.org/docs/contributing/contribute_conventions).

My question was more about the workflow name: in this PR it's `name: Golang Build & Test`
It seems like all the newer workflows that Quentin's team added looks like `agw-workflow` or `feg-workflow`.

.github/workflows/golang-build-test.yml
electronjoe: Added upload, don't yet see the auto-updated comments though? Upload log shows success.

themarwhal: Hmm that's weird..... @quentinDERORY might know more on how this actually works :D 

On another note, how does this work the test being run multiple times in different platforms? I wonder if there's a collision of test results if the .xml files are all named the same?

electronjoe: I wondered if the multiple runs would cause a problem, yes. I had thought it might be smart enough but wasn't sure.

quentinDERORY: You need to add the name of your workflow there: https://github.com/magma/magma/blob/1861aeef6c348f845fb76bc3a62d6835d3757946/.github/workflows/unit-test-workflow.yml#L6
The upload artifact only upload github action artifacts, this second job does the actual comment on the PR.
This has to do again with github token permission on pull request on forked PR.

themarwhal: ah there you go! it is the pull_request_target magic again! :D

quentinDERORY:
> > > Some general questions regarding how we name workflows. We currently seem to have a mix of snake case? and regular title-like names. It might be good stylistically to merge on one pattern. @quentinDERORY any thoughts?
> > 
> > 
> > I believe @hcgatewood had an opinion that we prefer `-` to `_` due to e.g. Python reasons? So I was aiming for `-` in workflow names. But yeah, we need to standardize. Edit: oops, opposite! [Prefer underscores over hyphens](https://docs.magmacore.org/docs/contributing/contribute_conventions).
> 
> My question was more about the workflow name: in this PR it's `name: Golang Build & Test`
> It seems like all the newer workflows that Quentin's team added looks like `agw-workflow` or `feg-workflow`.

I personally like the workflow name to be one word (with hyphens or underscore) as we reference them in other places. It might work with spaces but feels weird to reference phrases.

tmdzk (approved):

PR #9032, opened 2021-09-09
chore: cleanup reviewdog github workflows

themarwhal:

## Summary
This is up for discussion, but I think consolidating all Reviewdog jobs into a single workflow is nicer stylistically. From looking at all the jobs, the current naming / workflow makes it hard to tell which jobs do assertions/tests and which jobs are meant for review aid. 

Putting all of them into a single reviewdog workflow makes it easy to tell that these jobs are for helping PR reviews and are not mandatory.

Another followup action is to make sure that reviewdog jobs do not fail. For example, the python linter job fails when all lint warnings are not addressed. I think this is confusing for contributors and reviewers as they are not marked as mandatory jobs.

## Additional Information


electronjoe:
This is counter to https://github.com/magma/magma/pull/9022 - I pitched your argument for binning these under one heading for readability of PR authors (as you cite) but @quentinDERORY pointed out that for the GitHub workflow exploration UI (not tied to any PR) you lose all visibility with colliding names. Lemme see if I can find the URL. He convinced me - but maybe we can re-open discussion =D

[This is the dashboard](https://github.com/magma/magma/actions) that completely fails navigation if we use the same Name all over the place.... This is very unfortunate because I would prefer to make things clean for PR authors understanding (as you point out). But ...

themarwhal:
> This is counter to #9022 - I pitched your argument for binning these under one heading for readability of PR authors (as you cite) but @quentinDERORY pointed out that for the GitHub workflow exploration UI (not tied to any PR) you lose all visibility with colliding names. Lemme see if I can find the URL. He convinced me - but maybe we can re-open discussion =D

ah I see, yeah I was wondering if this had been discussed once before. My main motivation for this change is that there's a lot of confusion between the python reviewdog linting job (not mandatory) and the python formatting job (required). 

Trying to see if there's a nice way to show reviewdog jobs are suggestions, unless they are explicitly marked as required.

quentinDERORY:
> This is counter to #9022 - I pitched your argument for binning these under one heading for readability of PR authors (as you cite) but @quentinDERORY pointed out that for the GitHub workflow exploration UI (not tied to any PR) you lose all visibility with colliding names. Lemme see if I can find the URL. He convinced me - but maybe we can re-open discussion =D
> 
> [This is the dashboard](https://github.com/magma/magma/actions) that completely fails navigation if we use the same Name all over the place.... This is very unfortunate because I would prefer to make things clean for PR authors understanding (as you point out). But ...

I was talking about having multiple workflows file with the same name but this PR is grouping all jobs under one workflow and this will have the behavior we want: Grouping every linter job under one header.
Thanks @themarwhal for doing this.

.github/workflows/reviewdog-workflow.yml
```diff
@@ -0,0 +1,163 @@
...
+    #  suppresses errors of all categories on that line.
+    ##
+    #  For details on cpplint optinos see the detailed comments in
+    #  https://github.com/google/styleguide/blob/gh-pages/cpplint/cpplint.py
+    ##
+    runs-on: ubuntu-latest
+    steps:
+      -
+        uses: actions/checkout@master
+      -
```
quentinDERORY: weird indentation here.

quentinDERORY (approved):

themarwhal:
> This is counter to #9022 - I pitched your argument for binning these under one heading for readability of PR authors (as you cite) but @quentinDERORY pointed out that for the GitHub workflow exploration UI (not tied to any PR) you lose all visibility with colliding names. Lemme see if I can find the URL. He convinced me - but maybe we can re-open discussion =D
> 
> [This is the dashboard](https://github.com/magma/magma/actions) that completely fails navigation if we use the same Name all over the place.... This is very unfortunate because I would prefer to make things clean for PR authors understanding (as you point out). But ...

I can't really make sense out of that dashboard and can't tell if I'm supposed to be able to :(

PR #9033, opened 2021-09-09
fix(ci): lint-clang-format always runs

electronjoe:
Any Github check that we want to make Required for PR merge, unfortunately, must execute for every PR. Github is not presently smart enough to realize that un-executed workflows are not required, even if annotated required.  This was discovered by @themarwal when working on Python linting infrastructure.

### DIFF
--- a/.github/workflows/lint-clang-format.yml
+++ b/.github/workflows/lint-clang-format.yml
@@ -6,13 +6,23 @@ on:  # yamllint disable-line rule:truthy
       - opened
       - reopened
       - synchronize
-    paths:
-      - .github/workflows/lint-clang-format.yml
-      - lte/gateway/c/**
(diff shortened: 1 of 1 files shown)


quentinDERORY (changes requested):
Let's use this module fkirc/skip-duplicate-actions@master instead to avoid noisy lint issues not coming from the PR itself.
we need to put a pre job that will skip the linter if files did not change in the specified path.
@themarwhal did it there #9032

themarwhal (approved):

electronjoe:
@quentinDERORY thanks for the pointer! I agree this is much better - given GH's limitations on path triggers and required flows. Please take a look at the revised PR!

quentinDERORY (approved):
Perfect :)

PR #9038, opened 2021-09-09
feat(agw): SctpdDownlink proxy

markjen:
## Summary

#8821 #8830

- Start magma/service.Router abstraction to provide injectable gRPC clients
- SctpdDownlink service proxy
- Add tests to proxy sctpd services

## Test Plan

```
[markjen@markjen-x1c8 go]$ go test -cover ./...
?   	github.com/magma/magma/accessd	[no test files]
?   	github.com/magma/magma/accessd/server	[no test files]
ok  	github.com/magma/magma/log	(cached)	coverage: 100.0% of statements
ok  	github.com/magma/magma/log/zap	(cached)	coverage: 100.0% of statements
?   	github.com/magma/magma/service	[no test files]
ok  	github.com/magma/magma/service/sctpd	(cached)	coverage: 100.0% of statements
```

`magma/accessd` and `magma/service` tests are pending reworked w/ a config system, DI library, and accessd -> agwd service rename.

## Additional Information


src/go/service/sctpd/sctpd.go
```diff
@@ -0,0 +1,14 @@
...
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package sctpd contains implementations of magma.sctpd.SctpdDownlink and
```
mstre123: is the plan to fill this out or is this for package level comments?

src/go/internal/testutil/fs.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+// MustTempDir returns a randomly generated temp dir from ioutil.TempDir and
+// a cleanup function. If the test failed, the temp dir is not cleaned up to
+// aid in debugging. Any encountered err results in a panic.
```
mstre123: should we clean temp directories before the next run?
maybe add prefix so that we can find them

src/go/accessd/main.go
```diff
@@ -21,8 +21,8 @@ import (
 
 func main() {
 	lm := log.NewManager(zap.NewLogger(uber_zap.NewDevelopmentConfig()))
+	server.Start(lm.LoggerFor("server"))
 
-	if err := server.Start(lm.LoggerFor("server")); err != nil {
-		panic(err)
-	}
+	stopper := make(chan struct{})
+	<-stopper
```
mstre123, line 25: is this being used for anything right now?

src/go/service/sctpd/sctpd.go
markjen: currently just for package level comments; there might be some code in here at some point.

I looked around the base libs and there aren't really any `doc.go` files, even though that appears to be popular in some libraries. Most of the time, the package level comment is in the file that matches the package name, but sometimes it's in a random file (which seems bad to me).

src/go/accessd/main.go
markjen: no use right now, but needs to be there since i made the grpc server start in a goroutine, so main() will exit without a stopper

src/go/internal/testutil/fs.go
markjen: we can't really safely clean the system-wide temp dir (e.g. magma uses /tmp for sctpd <-> mme unix domain sockets). Also, CI envs should start clean on each run in an ideal world :)

markjen:
Reorganized this PR into two [commits](https://github.com/magma/magma/pull/9038/commits):

3d71ad6d215f62251f6b0be75482415b9d585aa7 is #9067 and will be merged first.
ec0d847236e24590ca3cbf5c153097ba8651755e is for review here and will be rebased on top of #9067 before merging.

Please take a look at #9067 first. To see this PR's changes, please look at [ec0d847236e24590ca3cbf5c153097ba8651755e](https://github.com/magma/magma/pull/9038/commits/ec0d847236e24590ca3cbf5c153097ba8651755e).

src/go/accessd/server/server.go
```diff
@@ -21,46 +21,82 @@ import (
...
+	mmeUpstreamTarget     = "unix:///tmp/mme_sctpd_upstream.sock"
+	mmeGrpcDialTimeout    = time.Second
 )
 
-func newMmeGrpcConn() (*grpc.ClientConn, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), mmeGrpcDialTimeout)
-	defer cancel()
-	return grpc.DialContext(ctx, mmeUpstreamTarget, grpc.WithInsecure(), grpc.WithBlock())
+func newServiceRouter() service.Router {
+	sctpdDownstreamClient, err := newGrpcConn(sctpdDownstreamTarget, sctpdGrpcDialTimeout)
```
electronjoe: Nit: the naming here is a bit confusing to me - `Client` suffix here, versus `Conn` suffix for `mmeGrpcCon`.

src/go/accessd/server/server.go
```diff
@@ -21,46 +21,82 @@ import (
...
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	return grpc.DialContext(ctx, target, grpc.WithInsecure())
+}
+
 const (
-	mmeUpstreamTarget  = "unix:///tmp/mme_sctpd_upstream.sock"
-	mmeGrpcDialTimeout = time.Second
+	sctpdDownstreamTarget = "unix:///tmp/mme_sctpd_downstream.sock"
+	sctpdGrpcDialTimeout  = time.Second
```
electronjoe, line 36: Sounds like this context [In the non-blocking case, the ctx does not act against the connection. It only controls the setup steps.](https://pkg.go.dev/google.golang.org/grpc#DialContext). In our setting, what are the implications for this Timeout (and the associated `mmeGrpcDialTimeout`) and the sequencing of binary bring-up / establishment of these unix domain sockets?  I'll go see how you handle timeouts firing - but wondering how the system stabilizes initially with race conditions on startup of binaries (unless we intend to sequence via systemd... and in my experience this is hard to guarantee also and one should be robust to arbitrary ordering).

src/go/accessd/server/server.go
```diff
@@ -21,46 +21,82 @@ import (
...
+	return service.NewRouter(
+		sctpdpb.NewSctpdDownlinkClient(sctpdDownstreamClient),
+		sctpdpb.NewSctpdUplinkClient(mmeGrpcConn),
+	)
 }
 
-func newServer(logger log.Logger) *grpc.Server {
-	grpcServer := grpc.NewServer()
+const (
+	sctpdDownstreamNetwork = "unix"
```
electronjoe, line 57: Question: Is it typical to have more than one const () block? I feel like collecting these in one place would aid me personally - is that not standard?

src/go/accessd/server/server.go
```diff
@@ -21,46 +21,82 @@ import (
...
+	sctpdDownstreamPath    = "/tmp/sctpd_downstream.sock"
```
electronjoe, line 58: These const feel like duplication of the above const of lines 35-38?

src/go/internal/testutil/fs.go
electronjoe: Have you previously used this paradigm Mark? Like Eli I do worry about tmp refuse building up (but I also see the strong value of seeing what went wrong after a test fails).  What if we default to tmpdir cleanup but a flag allows preservation for debug? It sounds like nominal workflow is to clean up (both per ioutil.TempDir Godocs and for example when I explore [Bazel's recommended handling of TempDirs](https://docs.bazel.build/versions/main/test-encyclopedia.html#test-interaction-with-the-filesystem).

src/go/internal/testutil/log.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/magma/magma/log"
+	"github.com/magma/magma/log/zap"
+)
+
+// NewTestLogger returns a logger that logs to the returned buffer. Timestamps,
+// caller, and stacktraces are disabled to make it easier to validate log
+// output. Test coverage for timestamp, caller, and stacktrace output can be
+// found in github.com/magma/magma/log/zap/zap_integ_test.go.
+func NewTestLogger() (log.Logger, *bytes.Buffer) {
```
electronjoe: This is a really nice ability! Clean / simple logging for test validation. Hadn't seen that done.

src/go/service/router.go
```diff
@@ -0,0 +1,52 @@
...
+package service
+
+import (
+	"github.com/magma/magma/protos/magma/sctpd"
+)
+
+// A Router is responsible for providing clients, where the connection to
+// server implementations can be configured to be in-process or grpc.ClientConn
+// targets.
+type Router interface {
```
electronjoe, line 21: I expected src/go/service/router.go to define general interfaces for routers, though perhaps that's not possible in Go due to lack of Generics?  Regardless, it's strange to have service.Router tied to just sctpd behavior? Maybe this is a prototype that will eventually be moved / renamed as our Go implementation / need for routers grows beyond the present work?

src/go/service/router.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+//go:generate go run github.com/golang/mock/mockgen -destination mock_service/mock_router.go . Router
+
+type router struct {
+	sctpdDownlinkClient sctpd.SctpdDownlinkClient
+	sctpdUplinkClient   sctpd.SctpdUplinkClient
+}
+
+// NewRouter returns a Router that provides the provided clients.
```
electronjoe: nit: wording of `provides the provided`. Maybe just `returns a Router with the provided clients.`?

src/go/service/sctpd/proxy_downlink_server.go
```diff
@@ -0,0 +1,59 @@
...
+		With("ppid", req.GetPpid()).
+		With("force_restart", req.GetForceRestart()).
+		Debug().Print("Init")
+	return p.SctpdDownlinkClient().Init(ctx, req)
+}
+
+// SendDl proxies calls to SctpdDownlink.SendDl.
+func (p *ProxyDownlinkServer) SendDl(ctx context.Context, req *pb.SendDlReq) (*pb.SendDlRes, error) {
+	p.Logger.
+		With("assoc_id", req.GetAssocId()).
```
electronjoe, line 53: Question: I don't think I got an explanation on the `With` in your prior PR with the first logging implementation (or I missed it).  Is this enabling structured logging of fields within the log message? So it's well formatted / parsable?

electronjoe (approved):

src/go/accessd/server/server.go
markjen: oh yeah, these timeouts and the blocking option are going away in the next diff :)

src/go/accessd/server/server.go
markjen: in general, `var` / `const` are defined next to / close to usage, not all up top.

this is also true for variable declaration within blocks

src/go/accessd/server/server.go
markjen: yeah, it's all going away in #9069

src/go/service/router.go
markjen: yeah, the Router will end up having a bunch of <X>Client accessors on it; probably just hand coded (codegen is possible, but probably low value here)

To be fair, even if go supported generics, unclear if it's actually helpful to have `func Router.GetClient(T Type) T` in cases like this; it means all calls to GetClient might get indexed together which will just make a big jumble of things?

src/go/service/sctpd/proxy_downlink_server.go
markjen: Yeah, it adds the key=value as a structured log field, separate from the message. Take a look at TestLogger_With in src/go/log/zap/zap_integ_test.go for an example

src/go/accessd/server/server.go
markjen: oops typo

src/go/internal/testutil/fs.go
markjen: Good points; I'll remove this "enhancement". Tests should be repeatable anyways, so folks should be able to re-run anywhere to debug failures

PR #9041, opened 2021-09-09
chore: default all reviewdog annotations to be github-pr-check except…

themarwhal:
… golint

## Summary
In my previous commit https://github.com/magma/magma/pull/9032/, I accidentally reverted the change @electronjoe added to enable lint annotation that uses GitHub comments. 

Changing the configuration so that all reviewdog jobs use github-pr-check EXCEPT for the golangci check.

## Additional Information


quentinDERORY (approved):

electronjoe (commented):
Thanks @themarwhal ! Nice simplification of the logic statements too!

PR #9042, opened 2021-09-09
chore: federated integ test part 1

### LINKED ISSUE #9039: [test] federated integ test including all gateways (AGW, orc8r and FEG)
Currently we do not have a way to run a complete Magma setup (AGW + orc8r + feg) on our dev environments. That prevents us to experiment with our code and even test before landing. 

In order to allow developers to run a complete setup, we purpose the creation of a similar setup to lte_integ test but which include all 3 gateways.

This setup must me easy to build and to run. We have to provide a way to wipe all the changes so we can switch between federated integ_test and lte integ_test.

The setup should use:
- AGW running on magma vm
- FEG running on local host or on magma VM 
- Orc8r running on local host
- s1ap running from magma_test
- HSS running as part of the FEG
- Traffic server running from magma_trfserver

The steps to develop this feature will be:
- Develop/Improve fab files to configure network and local gateways in orc8r, making sure we can add and remove the configuration easily
- Modify AGW and FEG configuration on orc8r to support federated case. gateway.mconfig files must be proerly generated by orc8r and stream down to the gateways
- Run FEG either on magma vagrant vm or on local vm. We need to reconfigure feg so it can run simultaneously with AGW and Orc8r
- Modify `build.py`, `run.py` on orc8r and `dev_tools.py`, `fabfile.py` on to support new commands to automate the build and run of the federated case
- Create a new repository for the `federated_integ_test` which includes a new `fabfile.py` to automate the whole setup and any specific configuration files from the setup.
- Add some specific test to s1apTester

uri200:

## Summary

This is the first PR which tries to add some code to configure a full local federated environment running AGW, Orc8r and FEG all at the same time.

Note this PR is not final and more testing will be needed to make sure the setup works properly. More PR will come to try to set up a full working environment.

This PR includes new fab file to control this environment
You can try it either running
```
cd ~/magma/l/g/python/integ_tests/federated_tests

fab build_all
# if it fails you may need to build MAGMA , feg and Orc8r manually using
fab_build_or8r
fab build_feg
fab build_agw
fab start_all
fab configure_orc8r

```

## Test Plan

created new fab file 

## Additional Information

### DIFF
--- a/feg/gateway/fabfile.py
+++ b/feg/gateway/fabfile.py
@@ -22,18 +22,31 @@ import tools.fab.types as types
 
 SNOWFLAKE_FEG_FILE = '../../.cache/feg/snowflake'
 NETWORK_ID = 'feg_test'
+FEG_DOCKER_LOCATION = 'docker/'
 
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
 
-def register_feg():
+def register_feg_gw(location_docker_compose: str = FEG_DOCKER_LOCATION):
+    """
+    Add FEG gateway to orc8r
+    Args:
+        location_docker_compose: location of docker compose. Default set to
+        FEG_DOCKER_LOCATION
+    """
     _register_federation_network()
-    _register_feg()
+    _register_feg(location_docker_compose)
 
 
-def deregister_feg_gw(vm_name='feg'):
-    dev_utils.delete_feg_certs(vm_name)
-    _deregister_feg_gw(vm_name)
+def deregister_feg_gw(location_docker_compose: str = FEG_DOCKER_LOCATION):
+    """
+    Remove FEG gateway from orc8r and remove certs from FEG gateway
+    Args:
+        location_docker_compose: location of docker compose. Default set to
+        FEG_DOCKER_LOCATION
+    """
+    _deregister_feg_gw(location_docker_compose)
+    dev_utils.delete_gateway_certs_from_docker(location_docker_compose)
 
 
 class RadiusConfig:
@@ -225,7 +238,7 @@ class HssConfigs:
 class S6aConfigs:
     def __init__(
         self,
-        plmn_ids: List[str] = [],
+        plmn_ids: List[str] = list(),
         server: DiamServerConfig = DiamServerConfig(),
     ):
         if plmn_ids is None:
@@ -340,22 +353,23 @@ def _register_federation_network(payload: FederationNetwork = FederationNetwork(
     dev_utils.create_tier_if_not_exists(nid, 'default')
 
 
-def _register_feg():
-
+def _register_feg(location_docker_compose: str):
     with open(SNOWFLAKE_FEG_FILE) as f:
         hw_id = f.read().rstrip('\n')
     if not hw_id:
         print(f'Could not open test feg snowflake {SNOWFLAKE_FEG_FILE}')
-        hw_id = dev_utils.get_hardware_id_from_docker_on_vagrant(vm_name='feg')
+        hw_id = dev_utils.get_gateway_hardware_id_from_docker(
+            location_docker_compose=location_docker_compose,
+        )
 
     already_registered, registered_as = dev_utils.is_hw_id_registered(
         NETWORK_ID, hw_id,
     )
     if already_registered:
         print()
-        print(f'============================================')
+        print('============================================')
         print(f'Feg is already registered as {registered_as}')
-        print(f'============================================')
+        print('============================================')
         return
 
     gw_id = dev_utils.get_next_available_gateway_id(NETWORK_ID)
@@ -400,17 +414,19 @@ def _register_feg():
     )
     dev_utils.cloud_post(f'feg/{NETWORK_ID}/gateways', gw_payload)
     print()
-    print(f'=====================================')
+    print('=====================================')
     print(f'Feg {gw_id} successfully provisioned!')
-    print(f'=====================================')
+    print('=====================================')
 
 
-def _deregister_feg_gw(vm_name: str):
+def _deregister_feg_gw(location_docker_compose: str):
     with open(SNOWFLAKE_FEG_FILE) as f:
         hw_id = f.read().rstrip('\n')
     if not hw_id:
         print(f'Could not open test feg snowflake {SNOWFLAKE_FEG_FILE}')
-        hw_id = dev_utils.get_hardware_id_from_docker_on_vagrant(vm_name=vm_name)
+        hw_id = dev_utils.get_gateway_hardware_id_from_docker(
+            location_docker_compose=location_docker_compose,
+        )
 
     already_registered, registered_as = dev_utils.is_hw_id_registered(
         NETWORK_ID, hw_id,
@@ -418,14 +434,13 @@ def _deregister_feg_gw(vm_name: str):
 
     if not already_registered:
         print()
-        print(f'===========================================')
-        print(f'VM is not registered')
-        print(f'===========================================')
+        print('===========================================')
+        print('VM is not registered')
+        print('===========================================')
         return
 
     dev_utils.cloud_delete(f'feg/{NETWORK_ID}/gateways/{registered_as}')
     print()
-    print(f'=========================================')
+    print('=========================================')
     print(f'Feg Gateway {registered_as} successfully removed!')
-    print(f'(restart docker FEG on {vm_name} vm)')
-    print(f'=========================================')
+    print('=========================================')
--- a/feg/gateway/fabfile.py
+++ b/feg/gateway/fabfile.py
@@ -20,6 +20,7 @@ sys.path.append('../../orc8r')
 import tools.fab.dev_utils as dev_utils  # NOQA
 import tools.fab.types as types
 
+SNOWFLAKE_FEG_FILE = '../../.cache/feg/snowflake'
 NETWORK_ID = 'feg_test'
 
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
@@ -28,6 +29,11 @@ urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 def register_feg():
     _register_federation_network()
     _register_feg()
+
+
+def deregister_feg_gw(vm_name='feg'):
+    dev_utils.delete_feg_certs(vm_name)
+    _deregister_feg_gw(vm_name)
 
 
 class RadiusConfig:
@@ -219,7 +225,7 @@ class HssConfigs:
 class S6aConfigs:
     def __init__(
         self,
-        plmn_ids: List[str] = None,
+        plmn_ids: List[str] = [],
         server: DiamServerConfig = DiamServerConfig(),
     ):
         if plmn_ids is None:
@@ -335,11 +341,13 @@ def _register_federation_network(payload: FederationNetwork = FederationNetwork(
 
 
 def _register_feg():
-    with lcd('docker'), hide('output', 'running', 'warnings'):
-        hw_id = local(
-            'docker-compose exec magmad bash -c "cat /etc/snowflake"',
-            capture=True,
-        )
+
+    with open(SNOWFLAKE_FEG_FILE) as f:
+        hw_id = f.read().rstrip('\n')
+    if not hw_id:
+        print(f'Could not open test feg snowflake {SNOWFLAKE_FEG_FILE}')
+        hw_id = dev_utils.get_hardware_id_from_docker_on_vagrant(vm_name='feg')
+
     already_registered, registered_as = dev_utils.is_hw_id_registered(
         NETWORK_ID, hw_id,
     )
@@ -353,12 +361,71 @@ def _register_feg():
     gw_id = dev_utils.get_next_available_gateway_id(NETWORK_ID)
     md_gw = dev_utils.construct_magmad_gateway_payload(gw_id, hw_id)
     gw_payload = FederationGateway(
-        id=md_gw.id, name=md_gw.name, description=md_gw.description,
+        id=md_gw.id,
+        name=md_gw.name,
+        description=md_gw.description,
         device=md_gw.device,
-        magmad=md_gw.magmad, tier=md_gw.tier,
+        magmad=md_gw.magmad,
+        tier=md_gw.tier,
+        federation=FederationNetworkConfigs(
+            hss=HssConfigs(
+                server=HssServer(
+                    local_address='localhost:3767',
+                    address='localhost:3768',
+                ),
+            ),
+            s6a=S6aConfigs(
+                plmn_ids=[],
+                server=DiamServerConfig(
+                    local_address='localhost:3767',
+                    address='localhost:3768',
+                ),
+            ),
+            gx=GxConfig(
+                servers=[
+                    DiamServerConfig(
+                        address='localhost:3868',
+                    ),
+                ],
+            ),
+            gy=GyConfig(
+                servers=[
+                    DiamServerConfig(
+                        address='localhost:3968',
+                    ),
+                ],
+            ),
+        ),
+
     )
     dev_utils.cloud_post(f'feg/{NETWORK_ID}/gateways', gw_payload)
     print()
     print(f'=====================================')
     print(f'Feg {gw_id} successfully provisioned!')
     print(f'=====================================')
+
+
+def _deregister_feg_gw(vm_name: str):
+    with open(SNOWFLAKE_FEG_FILE) as f:
+        hw_id = f.read().rstrip('\n')
+    if not hw_id:
+        print(f'Could not open test feg snowflake {SNOWFLAKE_FEG_FILE}')
+        hw_id = dev_utils.get_hardware_id_from_docker_on_vagrant(vm_name=vm_name)
+
+    already_registered, registered_as = dev_utils.is_hw_id_registered(
+        NETWORK_ID, hw_id,
+    )
+
+    if not already_registered:
+        print()
+        print(f'===========================================')
+        print(f'VM is not registered')
+        print(f'===========================================')
+        return
+
+    dev_utils.cloud_delete(f'feg/{NETWORK_ID}/gateways/{registered_as}')
+    print()
+    print(f'=========================================')
+    print(f'Feg Gateway {registered_as} successfully removed!')
+    print(f'(restart docker FEG on {vm_name} vm)')
+    print(f'=========================================')
--- a/lte/gateway/configs/subscriberdb.yml
+++ b/lte/gateway/configs/subscriberdb.yml
@@ -12,9 +12,9 @@
 # limitations under the License.
 
 # log_level is set in mconfig. it can be overridden here
-
+log_level: DEBUG
 # Append GRPC content to the log
-print_grpc_payload: False
+print_grpc_payload: True
 
 # Size of subscriber pages returned from cloud
 subscriber_page_size: 5000
--- a/lte/gateway/configs/subscriberdb.yml
+++ b/lte/gateway/configs/subscriberdb.yml
@@ -12,9 +12,9 @@
 # limitations under the License.
 
 # log_level is set in mconfig. it can be overridden here
-log_level: DEBUG
+
 # Append GRPC content to the log
-print_grpc_payload: True
+print_grpc_payload: false
 
 # Size of subscriber pages returned from cloud
 subscriber_page_size: 5000
--- a/lte/gateway/dev_tools.py
+++ b/lte/gateway/dev_tools.py
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-import os
+import subprocess
 import sys
 from typing import Any, List
 
@@ -27,6 +27,8 @@ NIDS_BY_TYPE = {
     LTE_NETWORK_TYPE: 'test',
     FEG_LTE_NETWORK_TYPE: 'feg_lte_test',
 }
+
+FEG_FAB_PATH = '../../feg/gateway/'
 
 # Disable warnings about SSL verification since its a local VM
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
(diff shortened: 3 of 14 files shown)


lte/gateway/configs/subscriberdb.yml
```diff
@@ -12,9 +12,9 @@
 # limitations under the License.
 
 # log_level is set in mconfig. it can be overridden here
-
+log_level: DEBUG
```
themarwhal: should we revert these? or keep them

lte/gateway/dev_tools.py
```diff
@@ -119,13 +155,34 @@ def _register_agw(network_type: str):
...
+        print(f'===========================================')
+        return
+
+    dev_utils.cloud_delete(f'lte/{network_id}/gateways/{registered_as}')
+    print()
+    print(f'=========================================')
+    print(f'AGW Gateway {registered_as} successfully removed!')
+    print(f'(restart AGW services on magma vm)')
+    print(f'=========================================')
+
```
themarwhal, line 202: i'm tempted to silence these f' related lints.... They can be controlled here: https://github.com/magma/magma/blob/master/setup.cfg

lte/gateway/configs/subscriberdb.yml
uri200: my bad!😭

lte/gateway/dev_tools.py
uri200: let me remove them. I think f-string will not make sense here if htere is no format to be applied at all

uri200: all done now

lte/gateway/python/integ_tests/federated_tests/README.md
```diff
@@ -0,0 +1,95 @@
...
+vagrant up magma_test
+vagrannt ssh magma_test
+# inside vagrant vm
+cd magma/lte/gateway/python
+make
+```
+Traffic vm
+```
+cd magma/lte/gateway
+vagrant up magma_trf
```
themarwhal:
```suggestion
vagrant up magma_trfserver
```

lte/gateway/python/integ_tests/federated_tests/README.md
```diff
@@ -0,0 +1,95 @@
...
+- AGW will run on **magma** vm.
+- FEG and Orc8r will run on our **local docker**.
+- Traffic server will run on **magma_trf** vm
+- Test will be started from **magma_test** vm
+
+## Build environment
+### Automatic build
+On your host machine go to
+`cd  magma/lte/gateway/python/integ_tests/federated_tests` and run
+```
```
themarwhal:
```suggestion
```bash
```

lte/gateway/python/integ_tests/federated_tests/README.md
```diff
@@ -0,0 +1,95 @@
...
+fab build_all
+```
+Once it is built, start the traffic and test vm
```
themarwhal: might be helpful to specify what will be built here. Is it just all AGW services? or everything including FeG containers?

lte/gateway/python/integ_tests/federated_tests/README.md
```diff
@@ -0,0 +1,95 @@
...
+cd magma/lte/gateway
+vagrant up magma
+vagrant ssh magma
+# inside vagrant vm
+cd magma/lte/gateway
+make run
+```
+- FEG:
+```
+cd magma/lte/gateway/python/integ_tests/federated_tests/docker
```
themarwhal, line 94: You mentioned that feg is run on the host machine. but I think people will still be confused, so it might be good to say here again to run this on the host machine.

themarwhal: Or say above that the instructions are all for the host machine unless explicitly marked otherwise?

lte/gateway/python/integ_tests/federated_tests/README.md
uri200: I think bash is the same as ```, but i added it just in case to all the boxes

themarwhal: ohhh I was wondering if it adds any useful syntax coloring, but for bash I suppose that is minimal.

lte/gateway/python/integ_tests/federated_tests/README.md
uri200: Sure, the script shows it, but may no be clear

lte/gateway/python/integ_tests/federated_tests/README.md
uri200: Thanks

I added a comment and include the command 

```
# exit from vagrant vm
exit
```

to all the actions done on AGW

themarwhal (approved):

orc8r/cloud/docker/docker-compose.yml
```diff
@@ -7,6 +7,7 @@ services:
   controller:
     volumes:
       - $PWD/../../../.cache/test_certs:/var/opt/magma/certs
+      - $PWD/controller/supervisord.conf:/etc/supervisor/conf.d/supervisord.conf
```
hcgatewood, line 11: If this is intentional, can you add a comment explaining why we're mounting instead of copying during image build

orc8r/cloud/docker/run.py
```diff
@@ -127,6 +137,12 @@ def _parse_args() -> argparse.Namespace:
         help='Stop running containers',
     )
 
+    parser.add_argument(
+        '--clear_db', '-b',
```
hcgatewood: Nit: convention seems to be `--clear-db`, e.g. from `man grep`. I'm assuming that still translates to `args.clear_db` in python code (?)

orc8r/cloud/docker/run.py
```diff
@@ -127,6 +137,12 @@ def _parse_args() -> argparse.Namespace:
...
+        action='store_true',
+        help='Clear contents of pgdata database (swagger api)',
```
hcgatewood: This clears the entire db, not just the data for the swagger api

orc8r/cloud/docker/run.py
```diff
@@ -50,6 +50,16 @@ def main() -> None:
     if args.print:
         return
 
+    if args.clear_db:
+        try:
+            cmd = ['docker-compose', 'down']
+            subprocess.run(cmd, check=True)
+            cmd = ['docker', 'volume', 'rm', '-f', 'orc8r_pgdata']
```
hcgatewood: Nit here and elsewhere: default to long-form of flags, since it's self-documenting

orc8r/cloud/docker/run.py
hcgatewood: I would almost *not* give this a short-form flag, just to protect against annoying accidents

orc8r/tools/fab/dev_utils.py
```diff
@@ -192,6 +198,70 @@ def get_hardware_id_from_vagrant(vm_name: str) -> str:
     return str(hardware_id)
 
 
+def get_hardware_id_from_docker(
```
hcgatewood: Which hwid is this grabbing? We already have a fab cmd somewhere for grabbing the hwid from an agw vm

orc8r/tools/fab/dev_utils.py
```diff
@@ -192,6 +198,70 @@ def get_hardware_id_from_vagrant(vm_name: str) -> str:
...
+        sudo('rm gw_challenge.key')
+
+
+def delete_feg_certs(location_docker_compose: str):
+    """
+    Get the magmad hardware ID from magma running on docker from Vagrant
+
+    Args:
+        location_docker_compose: location of docker compose used to run FEG
+        by default lte/gateway/python/integ_tests/federated_tests/docker
```
hcgatewood: Copied?

orc8r/tools/fab/dev_utils.py
```diff
@@ -192,6 +198,70 @@ def get_hardware_id_from_vagrant(vm_name: str) -> str:
...
+    with lcd('docker'), hide('output', 'running', 'warnings'), \
+            cd(location_docker_compose):
+        hardware_id = local(
+            'docker-compose exec magmad bash -c "cat /etc/snowflake"',
+        capture=True,
+        )
+    return str(hardware_id)
+
+
+def delete_agw_certs(vm_name: str):
```
hcgatewood: These delete_* functions can probably be run on the host -- they're deleting files that exist on the host, are owned/used by the agw, and only happen to be created (for ease of dev bootstrapping) by the orc8r

orc8r/tools/fab/dev_utils.py
```diff
@@ -192,6 +198,70 @@ def get_hardware_id_from_vagrant(vm_name: str) -> str:
...
+    Args:
+        location_docker_compose: location of docker compose used to run FEG
+        by default lte/gateway/python/integ_tests/federated_tests/docker
+    """
+    print("delete_feg_certs is running on directory %s" % os.getcwd())
+
+    subprocess.check_call(
+        [
+            'docker-compose exec magmad bash -c '
+            '"rm -f /var/opt/magma/certs/gateway.*"',
```
hcgatewood, line 246: This looks like it deletes *all* certs, not just for feg. Correct me if I'm wrong

orc8r/tools/fab/dev_utils.py
```diff
@@ -303,7 +373,35 @@ def cloud_post(
         cert=admin_cert,
     )
     if resp.status_code not in [200, 201, 204]:
+        parsed = json.loads(jsonpickle.pickler.encode(data))
```
hcgatewood, line 373: For my understanding: why are we pickling this first?

hcgatewood (changes requested):
1-2 small fixes and a couple questions, otherwise looks good. Excited for this!

orc8r/cloud/docker/run.py
uri200: Do you know if is there any way to clear just the swagger tables?

orc8r/cloud/docker/docker-compose.yml
uri200: `supervisord.conf` has flags for logs. If we mount it here we can change the verbosity on the fly while debugging instead of having to build the whole thing again

orc8r/tools/fab/dev_utils.py
uri200: almost. On an initial version I was running docker inside a Vagrant VM for FEG because I was getting some issues with ports.

Once I fixed I changed all the names, but forgot this description. Thanks for finding it!

orc8r/tools/fab/dev_utils.py
uri200: This gets the hwid from magmad that runs in docker (this is for feg)

orc8r/cloud/docker/run.py
hcgatewood: I would default to deleting everything -- it will have a similar effect, except that the gateways will have to re-bootstrap

But that said, (a) the first-best solution is implementing https://app.zenhub.com/workspaces/magma-5fac75d3e2cd890011f1677a/issues/magma/magma/8365, and (b) the good-enough-for-now solution is dropping specific tables. In particular, dropping all cfg_* tables should (*should* lol) have the desired effect

Tables ref: https://github.com/magma/magma/discussions/7845

orc8r/tools/fab/dev_utils.py
hcgatewood: Ah okay. Based on your previous comment, I see that we're using this for utils across magma, not just for orc8r. In which case, I think we can coalesce these two delete functions and note that it's for gws

Similar idea for my other fabfile comments -- ignore them, and instead: I think "dev utils" file should be agnostic to the particular gateway they are on (feg vs agw) and instead can indicate whether they're for a vm or docker (if that matters for the particular fn)

orc8r/cloud/docker/run.py
uri200: Yes bootstrapping again is a desired side effect. The reason for doing that is to have the ability to start from scratch. So deleting all may be the best option here.

orc8r/tools/fab/dev_utils.py
uri200: my bad, yes copy issue

orc8r/tools/fab/dev_utils.py
uri200: Certs on gateways are stored on `/var/opt/magma/certs` 
That folder is only visible outside Vagrant VM
The only shared folder is `/home/vagrant/magma` which is your local magma folder

orc8r/tools/fab/dev_utils.py
uri200: Note a gateway has just one set of certificates. 
So in that folder we have `root.CA` that we need to keep. The rest (gateway.crt gateway.key gw_challenge.key) can be deleted

This deletes the certs inside magmad container (which is FEG container)

orc8r/tools/fab/dev_utils.py
uri200: data is of type `Any` so we that function we make sure that data is json object

```
jsonpickle.pickler.encode: Return a JSON formatted representation of value, a Python object
```

Note we already do that on line [367](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=orc8r%2Ftools%2Ffab%2Fdev_utils.py&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=366&start_col=17&end_row=366&end_col=17)

orc8r/tools/fab/dev_utils.py
hcgatewood: Yeah I misread the purpose of these fns -- I thought it was only for running in orc8r. Yes it makes total sense to do it this way given that these would be run from a gw

orc8r/cloud/docker/docker-compose.yml
hcgatewood: Cool makes sense. Can you add a comment in the docker-compose file with that rationale

orc8r/tools/fab/dev_utils.py
hcgatewood: Agreed about the set of certs present and which to delete. I think the input I'm trying to give is

1. dev_utils should contain gateway-agnostic functionality. So either this fn should be in a feg fabfile, or rename it to e.g. `delete_docker_gateway_certs` (my preference is the latter)
2. The `delete_docker_gateway_certs` and `delete_vm_gateway_certs` can each ssh/cd to the correct directory, then they call `delete_gateway_certs` as the single entrypoint for how to delete the certs you mentioned -- rather than duplicating that `rm` logic across both fns

uri200: Yes I see. THe dev_utils was unknown to me when I started this PR

Let me check if I can move them around where they may make more sense

uri200: My initial idea was to have a generic function. But that didnt work

Accessing to a vagrant vm and docker container with ssh is not trivial. Vagrant changes the ssh port every time it is run (depending if you have other vagrant vm running). Accessing a docker container through ssh means you have to have ssh enabled on the container. 

Best option is to keep it simple and make a function for each case.

uri200: After checking the file I saw why I added those functions there:  `dev_utils` already include `get_hardware_id_from_vagrant` which is tight to vagrant. That function is used by `register_generic_gateway` which is not accurate since the function requires a vagrant vm.

Maybe `register_generic_gateway` can be refactored in future PRs to remove that function from inside of register geneirc gateway

uri200: In fact after having reviewed `dev_utils` and seeing we already have a dependence on vagrant with `get_hardware_id_from_vagrant`, it make sense to keep those functions in there. 

I will rename all those functions to include the provider and `gateway` so they are not confused with orc8r functions. This way we will be able to reuse them the day we have containerized AGW
```
get_gateway_hardware_id_on_vagrant
get_gateway_hardware_id_on_docker
delete_gateway_certs_on_vagrant 
delete_gateway_certs_on_docker
```

In the future I may create a functions for
```
register_generic_gateway_on_vagrant
register_generic_gateway_on_docker

```

feg/gateway/fabfile.py
```diff
@@ -219,7 +238,7 @@ def __init__(
 class S6aConfigs:
     def __init__(
         self,
-        plmn_ids: List[str] = None,
+        plmn_ids: List[str] = list(),
```
ardzoht, line 241: we should not create data structs as a default param, let's keep it as None as if there's no param passed we can create an empty list on the function itself

lte/gateway/dev_tools.py
```diff
@@ -75,14 +94,50 @@ def register_federated_vm():
...
+    subprocess.check_call(
+        'fab register_feg_gw', shell=True, cwd=FEG_FAB_PATH,
+    )
+
+
+def deregister_feg_gw():
+    """
+    Remove FEG gateway from orc8r and remove certs from FEG gateway
+    """
+    subprocess.check_call(
```
ardzoht, line 135: let's try to stay away from `shell=True` usages, you can omit this param by instead passing a list of whitespaces separated parameters

lte/gateway/dev_tools.py
```diff
@@ -75,14 +94,50 @@ def register_federated_vm():
...
+    dev_utils.delete_gateway_certs_from_vagrant('magma')
+    _deregister_agw(FEG_LTE_NETWORK_TYPE)
+
+
+def register_feg_gw():
+    """
+    Registers FEG AGW gateway on orc8r
+    """
+    subprocess.check_call(
+        'fab register_feg_gw', shell=True, cwd=FEG_FAB_PATH,
```
ardzoht, line 127: same here, let's not use `shell=True` calls

lte/gateway/dev_tools.py
```diff
@@ -11,6 +11,8 @@
 limitations under the License.
 """
 
+import os
```
ardzoht: remove?

lte/gateway/fabfile.py
```diff
@@ -462,6 +463,58 @@ def load_test(gateway_host=None, destroy_vm=True):
         env.hosts = [gateway_host]
 
 
+def build_and_start_magma(gateway_host=None, destroy_vm='False', provision_vm='False'):
```
ardzoht, line 466: add type hints for the params and the functions below

lte/gateway/python/integ_tests/federated_tests/fabfile.py
```diff
@@ -0,0 +1,190 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import subprocess
+import sys
+from distutils.util import strtobool
+
+from fabric.api import cd, env, execute, hide, run
```
ardzoht: remove unused imports

lte/gateway/python/integ_tests/federated_tests/fabfile.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def clear_gateways():
+    """
+    delete AGW and FEG gateways from orc8r
+    """
+    print('#### Removing federated agw from orc8r and deleting certs ####')
+    subprocess.check_call(
+        'fab --fabfile=dev_tools.py deregister_federated_agw',
+        shell=True, cwd=agw_path,
```
ardzoht, line 93: omit `shell=True` usages for all of check_call functions

ardzoht (changes requested):

feg/gateway/fabfile.py
uri200: Note that if it is set to None there, the code default PLMN to 123456.
That is why I was forced to use an empty list there.

Not sure where this come from and I was not able to see if this was really used somewhere else.

ardzoht: Then should we do `plmn_ids = []` ? It would achieve the same as your current change without initiating the list on the param

lte/gateway/dev_tools.py
uri200: `shell=True` is intentional. The reason is with shell=true we exit immediately if return is different than 0 allowing us to stop the scrip.

There are different ways to do capture exit code as shown [here](https://csatlas.com/python-subprocess-run-exit-code/). I chose that solution because is clearer.

Note the only draw back of using `shell=true` is security. On a production code (magmad, subscriberdb, etc) shell should not be used, but on a test script security is not a concern.
https://docs.python.org/2/library/subprocess.html

If you think there are other concerns of using shell let me know and I will change it

feg/gateway/fabfile.py
uri200: Same issue. Before `list()` i had `[]` and it complain the same.

But not this is not a real problem. All the object will point to the same empty list. The important is that is empty. That object will never be modified because once it is used up to call the API it is gone.

I can try to figure out in the future why we are using 123456 as a default in case of none

lte/gateway/fabfile.py
uri200: Note those are fab commands. None of our fab commands have hints (at least at that file)

The reason, I guess, is that for example `destroy_vm='False'` is a string  tyoe but in fact it represents a boolean.

orc8r/tools/fab/dev_utils.py
hcgatewood: Awesome! Lgtm

hcgatewood (approved):
Orc8r changes lgtm! Thanks Oriol, excited for this

lte/gateway/dev_tools.py
ardzoht: check_call can still exit immediately without `shell=True`, we can check for `CalledProcessError` exception raised and exit, my reasoning more than the security aspect is that if `shell=True` is not needed (we're not using shell pipes or other shell features) let's not use it at all

feg/gateway/fabfile.py
ardzoht: Sorry I was meaning `plmn_ids` should stay None, and if that's the case, initialize the list in the function (change `plmn_ids = ["123456"]` to `plmn_ids = []` (since anyways the current change is already overwriting that)

lte/gateway/dev_tools.py
uri200: Hey @ardzoht , you are right, check_call raises an exception
```
check_call(*popenargs, **kwargs):
    """Run command with arguments.  Wait for command to complete.  If
    the exit code was zero then return, otherwise raise
    CalledProcessError.  The CalledProcessError object will have the
    return code in the returncode attribute.
```

uri200: Ok , I finally remembered why I was using shell=true

Shell=true loads all envars from the shell before running the script. Since those scripts are calling other scripts that are supposed to be run through the command line directly, we need those envars (like ROOT_MAGMA) to be loaded.

For example, not having shell=true prevents tu runn build.py and run.py from orc8r

stackoverflow.com/questions/3172470/actual-meaning-of-shell-true-in-subprocess

ardzoht (approved):
lgtm

PR #9043, opened 2021-09-10
chore(td-agent-bit): change local port for td-agent-bit client

### LINKED ISSUE #9039: [test] federated integ test including all gateways (AGW, orc8r and FEG)

uri200:

## Summary

td-agent-bit client and server use the same local port to run 24224. That prevents running td-agent-bit client and server on the same host (for example feg and orc8r can run at the same time)

This PR changes td-agent-bit client template to use 24221 as a local port at `orc8r/gateway/configs/templates/td-agent-bit.conf.template`

This PR **does not** change the server local port at `orc8r/cloud/docker/fluentd/conf/fluent.conf`

## Test Plan

ran both orc8r and feg on the same host

## Additional Information

### DIFF
--- a/orc8r/gateway/configs/templates/td-agent-bit.conf.template
+++ b/orc8r/gateway/configs/templates/td-agent-bit.conf.template
@@ -49,7 +49,7 @@
     Name     forward
     Tag      docker.log
     Listen   0.0.0.0
-    port     24224
+    port     24221
 
 # Input from the EventD service
 [INPUT]


orc8r/gateway/configs/templates/td-agent-bit.conf.template
```diff
@@ -49,7 +49,7 @@
     Name     forward
     Tag      docker.log
     Listen   0.0.0.0
-    port     24224
+    port     24221
```
themarwhal, line 52: Are FeG dockers still able to export docker service logs if we change this value?

uri200: you mean show `docker-compose log`?  If that is what you meant, then yes.

Or you may mean this 

http://github.com/magma/magma/blob/f21837cd2427b024f1258d55b7b7705c29bbc919/xwf/gateway/docker/docker-compose.yml#L123-L123

emakeev (approved):
lgtm

orc8r/gateway/configs/templates/td-agent-bit.conf.template
themarwhal: I was wondering if docker logs still get forwarded to fluent propoerly if we change this.

Does this input logic mean that it tags any input received from port 24221 as `docker.log`? If it is, I was wondering if there was any other configuration that specifies 24221 as the forwarding port to stream logs coming out of docker containers. 

From a quick glance in the repo I didn't seem to find anything, but just making sure :)

PR #9056, opened 2021-09-10
feat(agw): Cherry pick FreedomFiOne enodebd changes

amarpad:
## Summary

Cherry pick adding support for FreedomFiOne CBRS enodebs

## Test Plan

Running for a ~ month without issues on a 1.6 AGW

## Additional Information

### DIFF
--- a/lte/gateway/configs/enodebd.yml
+++ b/lte/gateway/configs/enodebd.yml
@@ -21,10 +21,28 @@ tr069:
   #       if this is ever changed in dnsd.yml, this needs to be updated too
   public_ip: 192.88.99.142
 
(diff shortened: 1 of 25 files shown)


ardzoht (approved):
lgtm, was this tested on v1.6 branch? if so can you please add a comment on your summary just to note that, thanks!

amarpad:
> lgtm, was this tested on v1.6 branch? if so can you please add a comment on your summary just to note that, thanks!

Yeah sorry, we are shipping 1.6 AGWs, so all of this was validated against 1.6. Will update the summary. As a corollary any field issues that we see will also need to be back ported. Will update the summary, good point.

PR #9057, opened 2021-09-10
refactor(agw): refactor MConfigLoader to enable test, add tests

electronjoe:
# Summary

This refactoring PR was necessary to enable testing of e.g. #9047.

The general points of refactoring were as follows.

- Removed the Class definition
  - As it was unused and extraneous
- Moved the bulk of logic to a public `load_service_mconfig` function that takes in an istream
  - To allow tests that do not require filesystem read/write and associated complexities
- Moved non-public functions to anonymous namespace
  - Preferred over static declaration that (sometimes) existed

# Outstanding Work

When we have access to Google's abseil library, we should make use of a couple helpers there.

- Consider use of the absl::Status return type instead of Boolean
  - Enables more fidelity in error reporting, error joining, etc
- Shall use absl::Cleanup to ensure closure of opened resources (e.g. the file.close())

# Behavioral Change

In authoring tests, it was discovered that many mal-formed json mconfigs would actually result in an uncaught exception being thrown by our third party json library.  I now catch this exception in the MConfigLoader library and pass back a status instead (failure notice via the boolean return value).

# Bazel Build based Test

Our CMake project architecture seems to make it ~impossible to access our protobuf definitions for mconfig (in /lte/gateway/..) from the MConfigLoader library that uses them (in /orc8r/gateway/c/..).

This PR includes a Bazel BUILD file which has no issue with this. Hopefully we'll go the way of Bazel.

# Test Strategy

Authored new unit tests to confirm behavior for the following. Note this omits testing for the file i/o portions of the library.

- Empty input ("")
  - Should return failure status (false) on function call
- Input with proper MME mconfig shape and a couple fields set
  - Validate fields are parsed and in expected proto result

The PR has also passed [lte-integ-test](https://app.circleci.com/pipelines/github/magma/magma/32696/workflows/d9495b91-538b-4ed3-ab98-92fa8596441f/jobs/380159).


orc8r/gateway/c/common/config/test/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
-include_directories("/usr/src/googletest/googlemock/include/")
+include_directories("/usr/src/googletest/googlemock/include/" .)
```
themarwhal: what does this do?

orc8r/gateway/c/common/config/test/CMakeLists.txt
```diff
@@ -26,3 +26,14 @@ target_link_libraries(yaml_utils_test
     pthread
     ${GCOV_LIB})
 add_test(test_yaml_utils yaml_utils_test)
+
+add_executable(mconfig_loader_test test_mconfig_loader.cpp
+    lte/protos/mconfig/mconfigs.pb.cc
+    orc8r/protos/common.pb.cc
+    )
```
themarwhal: nit: but should this ) be on line 32?

orc8r/gateway/c/common/config/test/CMakeLists.txt
electronjoe: So I learned some things about compiling protobufs in my manual effort to get this working.

Our mconfig.proto protobuf includes common.proto, and so that means that when the compiler gets to mconfig.pb.cc and tries to compile it, includes have to be sucked in that indicate a relative path of `orc8r/protos`.  But the protos aren't at that relative path, I want them instead back at `.../c/ommon/config/test/orc8r/protos`. So what we really want is that the compiler includes the local path (.) in every search it does.

I would say this was trial and error. But I think it makes sense in retrospect. So I'm not at all certain this is the only way - only that this is a way (exactly what I don't like about CMake... and what I like about Bazel's opinionated methods).

electronjoe:
I would like to run include-what-you-use on this PR, but our CMake config is so broken that I cannot output compile_commands.json despite my work-arounds. Bazel supports easy export of compile_commands.json - I will wait until we formally adopt Bazel to do this.

orc8r/gateway/c/common/config/test/CMakeLists.txt
electronjoe: This didn't work in CI - aborting test execution by CMake - tossing it in this PR =(

themarwhal (approved):

ulaskozat (approved):
LGTM

Issue #9053, opened 2021-09-10
orc8/nms R1.6 reporting cert expiring when cert show expirers in 2031


pkapatsos-blinq:
I see this related ticket Add alerts about gateway certificates expiring soon #5960 and perhaps is the cause??

### Your Environment

- **Version:R.16
- **Affected Component:** Orchestrator, Access Gateway, or NMS all R1.6
-
- **Deployment Environment:** e.g. AWS (orc8r), R1.6

### Describe the Issue
We are getting certificates expire soon on some networks not all, 
![image](https://user-images.githubusercontent.com/61711855/132887194-f32f9cb9-6329-4c28-a371-dbc41ebfdbee.png)

![image](https://user-images.githubusercontent.com/61711855/132887920-600aee90-de0c-440a-a9a0-71779b061ec6.png)

the cert output shows 2031 from the AGW.
![image](https://user-images.githubusercontent.com/61711855/132887405-0d6edffc-a10f-465a-a0e4-deb318368f31.png)

The timer in the alert ruls shows 720hrs
![image](https://user-images.githubusercontent.com/61711855/132888004-e1e9f98d-042b-4af9-b813-49ec22650623.png)

**Expected behavior**
This should not be just on some networks but all and it should not show up unlill 720Hrs before expiring in 2031.
What more can we provide to help , think this is a false positive so how can we check , clear them and ensure that the functions to and from NMS continue to work until 2031?

joary:
@pkapatsos-blinq , can you share the data you see in your grafana view for the `cert_expires_in_hours` value?

Something like the one below:

![image](https://user-images.githubusercontent.com/1166157/132894775-242f5727-db64-4510-9b37-ec857d99a871.png)


Also please include the logs from your certifier k8s pods, from:
```
kubectl logs orc8r-certifier-aaaaaaaa-bbbbb -n orc8
kubectl logs orc8r-certifier-aaaaaaaa-ccccc -n orc8
```

Use `kubectl get pods -n orc8r` to find the correct name.

pkapatsos-blinq:
Hi Ok show 86428 hrs that a lot better, 
![image](https://user-images.githubusercontent.com/61711855/132896771-13f01c9d-5dc9-4640-aa14-87487edfd03e.png)

will upload the kube side commands also.

joary:
Can you please show the table below the graph as well? including the other parameters collected for this metric.

pkapatsos-blinq:
Yup sorry. 
[...]

PR #9058, opened 2021-09-10
fix: Fix certificate expiration alert

joary:
From the [lifespan.go](https://github.com/magma/magma/blob/master/orc8r/cloud/go/services/certifier/analytics/calculations/lifespan.go#L80):
the metric is calculated as the remaining hours (will decrease over time).

Them alert shuold just trigger if cert_expires_in_hours is smaller than 720

Solves issue #9053

## Summary

Changed alert rules to be triggered only when certificate remaining hours is smaller than 720

## Test Plan

Observe cert_expires_in_hours is higher than threshold and reducing over time, see issue #9053 

## Additional Information


andreilee (approved):

andreilee:
Duplicate PR #9803 was just merged, closing this pull request.

PR #9063, opened 2021-09-10
fix(agw): Fix freedomfiOne unit tests

amarpad:
## Summary
The order in which the desired configuration is set is based on a
dictionary. This results in the call order not being stable.
Sort the list of calls before comparing with existing results.
Since the call to the eNB itself is a single call the order in which
each param is set is not relevant, so it doesn't make sense to use
the more expensive OrderedDict datastructure.

## Test Plan

- Ran unit test in multiple python versions and verified it passes.

## Additional Information


ardzoht (approved):
lgtm

PR #9067, opened 2021-09-12
feat(agw): Refactor log package and test fixtures

markjen:
#8845 

Cleaning up tests and fixing failure on Windows (e.g. https://github.com/magma/magma/runs/3580999355)

Signed-off-by: Mark Jen <markjen@fb.com>

## Summary

- Internal/testutil package for shared test fixtures
- Move gomock codegen to mock_* packages
- Rework zap logger adapter to better fit our usages
- Clean up zap_test and zap_integ_test
- Add windows file workaround to log/zap package
- Fix zap_integ_test on windows
- Update go.mod via `go mod tidy`

## Test Plan

```
$ go test -cover ./...
ok  	github.com/magma/magma/log	(cached)	coverage: 100.0% of statements
ok  	github.com/magma/magma/log/zap	0.003s	coverage: 100.0% of statements
```

## Additional Information


electronjoe (approved):
So something went wrong with the PR stacking @markjen , and this PR was globbed into #9038 which I reviewed first this morning... So to the extent I have comments, they are located on that PR.

src/go/log/mock_log/mock_logger.go
```diff
@@ -0,0 +1,145 @@
...
+// Source: github.com/magma/magma/log (interfaces: Logger)
+
+// Package mock_log is a generated GoMock package.
+package mock_log
+
+import (
+	reflect "reflect"
+
+	gomock "github.com/golang/mock/gomock"
+	log "github.com/magma/magma/log"
```
themarwhal, line 11: go question..
```suggestion
    "reflect"

	"github.com/golang/mock/gomock"
	"github.com/magma/magma/log"
```
does that work too? or are we doing something special here?

themarwhal (approved):

src/go/log/mock_log/mock_logger.go
markjen: your suggestion definitely works and would be preferred, but this is codegen'd output from `gomock` (explicit package names can be seen in generated pb.go files as well). I didn't bother fixing up any generated code since it would require updating the gomock/protoc-gen-go tooling to make it stable.

hopefully files like this should be collapsed/hidden by default in the GH UI.

themarwhal: oh! I see. It makes sense if it's generated. Weird that GitHub didn't detect this. We could always add the file to .gitattributes at the root of our repo.

PR #9070, opened 2021-09-13
docs: fixed semantic-pr.yml link on magma documentation

uri200:

## Summary

Fixed missing link for new location of `semantic.yml`  (now `semantic-pr.yml` under `.githug/workflow`)

## Additional Information


themarwhal (approved):
Are we just fixing this for the 1.6 doc? Also I think it's helpful to leave in the example, but I guess it doesn't matter too much for 1.6.

uri200:
> Are we just fixing this for the 1.6 doc? Also I think it's helpful to leave in the example, but I guess it doesn't matter too much for 1.6.

Problem was it was changed in master and not 1.6.

Currently, 1.6 is our default public documentation, so that will help new contributuor with a semantic check failing that doesn't know about our doc structure.

PR #9071, opened 2021-09-13
feat(agw): Add the enb params in enb data model for agw and nms, the version contains pr…

hakeen0618:
Signed-off-by: hakeen <hanhufeng@baicells.com>

## Summary

- Add new parameters to eNB, including the following paramters:
- Reference_Signal_power,PA,PB,PowerClass
- X2 Enable/Disable
- We can set the eNB detail config from NMS eNB config, if the eNB support the params.

-  On NMS , we can config the eNB details use the PA,PB,Reference_Signal_power, as following test pic.
## Additional Information

This add enb parameters as optional for managed configs. We already test it on NMS , AGW , it should be good.


lte/protos/mconfig/mconfigs.proto
```diff
@@ -71,6 +71,14 @@ message EnodebD {
         int32 tac = 8;
         int32 cell_id = 9;
         string ip_address = 10;
+        PowerControl power_control = 34;
+        bool x2_enable_disable = 112;
```
uri200, line 75: can you please use sequential numbers

hakeen0618: In order to solve the conflict between parameters supported by different versions and coding by different people, we specify the number submitted by each person. If it must modify, I will modify at present. 
number

PR #9115, opened 2021-09-15
chore(subscriberdb): replace print_grpc to use rpc utils function

uri200:

## Summary

Replace _print_grpc for the new print_grpc from rpc utils

## Test Plan

make test

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/subscriberdb/protocols/m5g_auth_servicer.py
+++ b/lte/gateway/python/magma/subscriberdb/protocols/m5g_auth_servicer.py
@@ -13,8 +13,8 @@ limitations under the License.
 
 import logging
 
-from google.protobuf.json_format import MessageToJson
 from lte.protos import subscriberauth_pb2, subscriberauth_pb2_grpc
+from magma.common.rpc_utils import print_grpc
 from magma.subscriberdb import metrics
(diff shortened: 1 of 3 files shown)


lte/gateway/python/magma/subscriberdb/protocols/m5g_auth_servicer.py
```diff
@@ -67,6 +72,10 @@ def M5GAuthenticationInformation(self, request, context):
             m5gauth_vector.xres_star = m5g_ran_auth_vectors.xres_star[16:]
             m5gauth_vector.autn = m5g_ran_auth_vectors.autn
             m5gauth_vector.kseaf = m5g_ran_auth_vectors.kseaf
+            print_grpc(
```
ardzoht: nit: we could also put all these `print_grpc` calls into a `finally` block, either way it's fine for me

ardzoht (approved):
lgtm, left a minor comment

pshelar (approved):
LGTM

lte/gateway/python/magma/subscriberdb/protocols/m5g_auth_servicer.py
uri200: You are right! Didn't remember Finally was executed even after return!

will refactor that

PR #6987, opened 2021-05-18
docs(agw): Initial MME C++ migration proposal

lionelgo:
Signed-off-by: lionelgo <29477918+lionelgo@users.noreply.github.com>

## Summary

First draft of the MME code migration proposal in C++.
Need more work to better evaluate, describe, schedule tasks to be done.

PR #9118, opened 2021-09-15
docs(agw): MME C++ migration proposal

### LINKED ISSUE #9136: [AGW] LTE C++ Migration
# C++ Migration

## Problem

Initially and historically, the development of MME magma was done only in C.

Many third party libraries and features that were integrated and or developed around the MME core are written in C++.
The integration of third party C++ libraries necessary to support the evolution of the MME is each time not straightforward due to the significant remaining C code.

## Solution

A migration of an identified part of the C code in C++ is proposed to facilitate the maintenance and the integration of third party libraries.

See PR #9118

### LINKED ISSUE #11199: C++ migration of Stateless MME
# C++ migration of Stateless MME

## Problem

As part of the larger effort to migrate MME service to C++, this sub-proposal outlines the design to migrate the state management logic in MME service. Specifically, the MME service maintains per UE and per task state through:
- Hashtables implemented in C through [in-house library](https://github.com/magma/magma/tree/master/lte/gateway/c/core/oai/lib/hashtable) in 4G core
- Hashtables implemented with [std::unordered_map](https://github.com/magma/magma/blob/master/lte/gateway/c/core/oai/include/map.h) in 5G core 
- Redis <key,value> pairs on disk

There are multiple performance issues in the current implementation:
1. Hashtable C implementation is inefficient
2. The task-level state for Mme_app and S1ap tasks requires conversion from protobuf::map in Redis to std::unorderd_map in memory, and vice versa.
3. Since protobuf::map only accepts native types as keys, hashtables which use structures as keys (Guti to Mme_UE_S1ap_Id map) need another step of conversion from Guti struct to string and vice versa.
4. Every state read/write requires conversion from protobuf object to C struct and vice versa
5. Further, the task-level state for Mme_app and S1ap tasks bundles multiple hashmaps into one. Every time any of these hashmaps are modified, the entire task-level state object is serialized and written to Redis.

## Solution

Hashtable improvements:

- [ ]  https://github.com/magma/magma/issues/11189 
- [ ]  https://github.com/magma/magma/issues/10869 
- [x] https://github.com/magma/magma/issues/11190 

C to C++ conversions:
- [x] https://github.com/magma/magma/issues/11191 
- [ ] https://github.com/magma/magma/issues/12785
- [x] https://github.com/magma/magma/issues/13096

Improving write performance:
- [x] https://github.com/magma/magma/issues/5571 

Simplify state managers:
- [x] https://github.com/magma/magma/issues/11192

lionelgo:

## Summary

First draft of the MME code migration proposal in C++.
It replaces former PR #6987

## Additional Information

### DIFF
--- a/docs/readmes/proposals/p021_mme_migrate_to_c++.md
+++ b/docs/readmes/proposals/p021_mme_migrate_to_c++.md
@@ -1,0 +1,143 @@
+---
+id: p021_mme_migrate_to_c++
+title: C++ Migration
+hide_title: true
+---
+
+
+# Proposal: C++ Migration
+
+Author(s): [@alexrod, @electronjoe, @kozat, @lionelgo]
+
+Last updated: 09/15/2021
+
+Discussion at
+[https://github.com/magma/magma/issues/4074](https://github.com/magma/magma/issues/4074).
+
+## Abstract
+
+This document describes the steps to migrate progressively, from C to C++, the MME core.
+
+## Background
+
+Initially and historically, the development of MME magma was done only in C.
+
+Many third party libraries and features that were integrated and or developed around the MME core are written in C++.
+
+We are now to the point where, for an easier and cleaner way of integrating new C++ libraries, for the ease of the maintenance of the code, a C to C++ migration seems unavoidable.
+
+## Proposal
+
+The proposal is constrained by a lack of performance of the current context serialization library (protobuf).
+ 
+Because the replacement libraries may have a big impact on the MME core code, this library migration will be studied first.
+
+
+### State Serialization
+The serialization performance of protobuf is insufficient, so a serialization library that allows better performance must be selected.
+
+The bad performance is due to the global serialization process (translation, memory allocation, copy).
+
+The replacement libraries considered are "cap'n proto" and "Flat buffers”, they support C/C++ standard API, binary output format and do not need memory copy (zero copy).
+
+
+The states that have to be serialized are:
+- S1AP, S1AP_IMSI_MAP, S1AP_UE
+- MME_UE, MME_NAS
+- SGW, SGW_UE, SPGW, SPGW_UE
+- S11/GTPV2-C
+- NGAP, NGAP_IMSI_MAP, NGAP_UE
+
+#### S1AP, MME_APP, NAS migration
+
+ S1AP seems to have no state dependency with other tasks.
+ 
+ Should be an opportunity to start doing a first migration with S1AP.
+
+ MME_APP and NAS tasks have contexts in common.
+ 
+
+### 3GPP protocols (OAI)
+This work is necessary only for the deployment of Magma MME with 3GPP compliant S11 over GTPV2-C and S6a over freeDiameter interfaces.
+
+S6a Interface implemented in C should be migrated to C++.
+A C++ port of S6a interface is already available in OAI HSS repo (https://github.com/OPENAIRINTERFACE/openair-hss) could serve as a basis.
+
+S11 Interface and its GTPV2-C protocol implemented in C should be migrated to C++.
+A C++ port of S11 and GTPV2-C interface done in OAI SPGW-C repo (https://github.com/OPENAIRINTERFACE/openair-spgwc) could serve as a basis.
+
+
+## Rationale
+
+### Get/Set state attributes
+
+For a progressive migration, and a decoupling of MME core code from the serialization library, macros, inline, helpers functions/C++ patterns may be used instead of directly using serialization library API in MME core code.
+
+Some MM and NAS IEs, for the ease of their manipulation (comparison operators, etc) could be migrated with wrapper classes.
+
+[A discussion of alternate approaches and the trade offs, advantages, and
+disadvantages of the specified approach.]
+
+## Compatibility
+
+Endianness: Whatever is the HW platform, endianness is fixed.
+
+Backward compatibility of the new serialization library with protobuf is not envisioned.
+
+## Implementation
+
+  Description of steps in the C++ migration:
+ 
+* Serialization library selection
+    * Build 2 realistics Prototypes with MME_APP, NAS protos (cap’n proto, Flat buffers)
+        * Migrate hashtables (find functions)
+    * Get measurements and evaluate performances, maintainability.
+    * Select library
+* MME_APP/NAS Serialization migration
+    * Write proto
+    * Write redis client libraries
+    * Design classes for wrapping generated POD classes (?)
+    	Not all context attribute need to be serialized
+    * Migrate states (update core code with C++ patterns)
+    * Write Unit tests
+* Migrate MME_APP/NAS hash tables (find functions, evaluate performance)
+
+  List of hash tables with (Key,Value) types:
+    * state_ue_ht            (mme_ue_s1ap_id_t, ue_mm_context_t*)
+    * UeIpImsiMap            (std::string/ue_ip, vector<imsi64_t>)
+    * guti_ue_context_htbl   (guti_t, mme_ue_s1ap_id_t)
+    * enb_ue_s1ap_id_ue_context_htbl (enb_s1ap_id_key_t, mme_ue_s1ap_id_t)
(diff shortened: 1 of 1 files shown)


docs/readmes/proposals/p021_mme_migrate_to_c++.md
```diff
@@ -0,0 +1,143 @@
...
+Endianness: Whatever is the HW platform, endianness is fixed.
+
+Backward compatibility of the new serialization library with protobuf is not envisioned.
+
+## Implementation
+
+  Description of steps in the C++ migration:
+ 
+* Serialization library selection
+    * Build 2 realistics Prototypes with MME_APP, NAS protos (cap’n proto, Flat buffers)
```
markjen, line 92: Strongly recommend storing state directly in protobuf-generated objects themselves. They will then be ready for serialization. No need for hashtable -> cap'n proto / flat buffer translation (performance should be roughly equivalent if state is already a proto in memory and just getting serialized)

docs/readmes/proposals/p021_mme_migrate_to_c++.md
```diff
@@ -0,0 +1,143 @@
+---
+id: p021_mme_migrate_to_c++
+title: C++ Migration
+hide_title: true
+---
+
+
+# Proposal: C++ Migration
```
markjen, line 8: since this proposal is centered around studying state representation and serialization performance, perhaps we should rename it as such?

docs/readmes/proposals/p021_mme_migrate_to_c++.md
```diff
@@ -0,0 +1,143 @@
...
+ 
+Because the replacement libraries may have a big impact on the MME core code, this library migration will be studied first.
+
+
+### State Serialization
+The serialization performance of protobuf is insufficient, so a serialization library that allows better performance must be selected.
+
+The bad performance is due to the global serialization process (translation, memory allocation, copy).
+
+The replacement libraries considered are "cap'n proto" and "Flat buffers”, they support C/C++ standard API, binary output format and do not need memory copy (zero copy).
```
markjen, line 41: the way cap'n proto and flat buffers achieve 0copy is making their structures read only. this yields great performance benefit during read operations. we do not do many state deserialization/reads in the MME codebase.

once mutation of state and then re-serialization is necessary, cap'n proto and flat buffers loses most of its performance advantage.

while our current pattern of hashtable -> protobuf obj -> serialization yields a lot of wasted cycles, storing the state in protobuf objs directly will alleviate most of this (specifically, a dramatic decrease in heap alloc/free). going from an in-memory protobuf -> serialized bytes should be close to the same performance of cap'n proto and flat buffers.

sticking with protobuf if possible is a significantly superior strategy. protobuf has dramatically more support compared to cap'n proto and flat buffers. it also matches model/codegen and ser/des choices in grpc and other services working with MME. while one could consider MME state serialization to be a "private" concern, it's still a slight advantage for all other sections of the code to be able to share models and make dev tooling universal.

electronjoe: > going from an in-memory protobuf -> serialized bytes should be close to the same performance of cap'n proto and flat buffers.

I don't buy this without some data - protobuf is not stored in memory compressed is it? That would be a serious penalty for any accesses.  So serialization involves all the encoding overhead of Protobuf (which is substantial - given the objective is to minimize size on the wire).

> sticking with protobuf if possible is a significantly superior strategy.

Agreed. Should only modify the strategy if the performance impact (system wide, not MME specifically) is enabling / make or break.

docs/readmes/proposals/p021_mme_migrate_to_c++.md
electronjoe: > They will then be ready for serialization

This is the part that incurs a large chunk of overhead. See the link I added above (protobuf encoding to wire format ~60x slower than flatbuffers).

docs/readmes/proposals/p021_mme_migrate_to_c++.md
electronjoe: [Here](https://google.github.io/flatbuffers/flatbuffers_benchmarks.html) is direct data from Google.

I believe we would be comparing `FlatBuffers (binary)` with `Protocol Buffers Lite` (actually not Lite for us, presumably worse for full fledge Protobuf).  Encoding speed for 1 million encodes in the FlatBuffers benchmark (by Google) yields **~60x speedup for FlatBuffers vs PB Lite**.

I went to look for their source code for the benchmark to confirm that they were not measuring the population time of the fields in the buffers, just encoding time - but the benchmarks are long gone and I've not rediscovered them in binary search of the git repo. Still - the encoding effort delta makes sense to me (even once in memory) - Protobuf does a lot of work to encode with minimal size.

docs/readmes/proposals/p021_mme_migrate_to_c++.md
lionelgo: I would say it is the first task to address, since it may add some serious constraints to the data model. Sure it is a big subject, that seems to shadow the C++ migration.

electronjoe:
The current Proposal Process commits design docs with only review of ~formatting or typos, and then discusses the proposal details on the associated GH Issue.  We've started to review the issue in this PR - can we move discussion to the GH Issue (which needs to be tagged in the PR @lionelgo).  See the [Magma Proposal Process](https://docs.magmacore.org/docs/contributing/contribute_proposals).

electronjoe (approved):

docs/readmes/proposals/p021_mme_migrate_to_c++.md
markjen: > [Here](https://google.github.io/flatbuffers/flatbuffers_benchmarks.html) is direct data from Google.
> 
> I believe we would be comparing `FlatBuffers (binary)` with `Protocol Buffers Lite` (actually not Lite for us, presumably worse for full fledge Protobuf). Encoding speed for 1 million encodes in the FlatBuffers benchmark (by Google) yields **~60x speedup for FlatBuffers vs PB Lite**.
> 
> I went to look for their source code for the benchmark to confirm that they were not measuring the population time of the fields in the buffers, just encoding time - but the benchmarks are long gone and I've not rediscovered them in binary search of the git repo. Still - the encoding effort delta makes sense to me (even once in memory) - Protobuf does a lot of work to encode with minimal size.

This benchmark is not representative of our scenario: update + encode.

Comparing pre-canned immutable flatbuf serialization to mutable protobuf serialization is not representative of our use case. Here are the scenarios to consider:

- State stored directly in mutable protobuf objs, update field value in protobuf obj (essentially free), serialize out
- State stored directly in flatbuf, rebuilding flatbuf on each update, serialize out
- State stored in hashtable (or other datastructure), convert to flatbuf (this part is effectively the same as serializing), serialize out

Thus:

1. Doing a straight flatbuf vs protobuf serialization benchmark is a waste of time, we already know the results.
2. Replacing the current convert to protobuf + serialize with convert to flatbuf + serialize will yield better performance. However, it will introduce a new dependency and be hard to migrate away from.
3. Replacing current, in-memory state representation w/ protobuf is the best case if it provides acceptable performance. I don't want us to get stuck on step 2; if we want to benchmark things, then let's compare 2 vs. 3, not 1 vs. 2.

electronjoe: Comparing 2 vs 3 sounds good - and agree that a reasonable first step is in memory protobuf representation / mutation in place.  Removes a whole lot of non-proto hash map iteration and copies (which may be plenty to resolve the performance bottleneck.

lionelgo: > we do not do many state de-serialization/reads in the MME code base.

Actually I agree, depends on constraints and strategy and deployments envisioned.
If you have tight constraints on occupied memory, you could flush your context to persistent storage and free it after the received MODIFY_BEARER_RESPONSE or with less aggressiveness when releasing UE_CONTEXT based on number of UE attached threshold.
If you want to deploy magma in a cluster you may become even more aggressive by choosing to follow the stateless paradigm.

ardzoht (approved):
lgtm

PR #9135, opened 2021-09-16
feat: add exclusive flags to SessionD unit tests that cannot be run i…

themarwhal:
…n parallel

## Summary
Some SessionD unit tests are sad when they are run together. I believe this is because the two unit tests both try to open up a test service at the same port. Will have to refactor the two tests eventually, but disabling parallel test execution for those two.

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/test/BUILD.bazel
+++ b/lte/gateway/c/session_manager/test/BUILD.bazel
@@ -150,6 +150,7 @@ cc_test(
     size = "small",
     srcs = ["test_cloud_reporter.cpp"],
     flaky = True,
+    tags = ["exclusive"],
     deps = [
         ":sessiond_mocks",
         "//lte/gateway/c/session_manager:session_reporter",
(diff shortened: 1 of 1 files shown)


lte/gateway/c/session_manager/test/BUILD.bazel
```diff
@@ -176,6 +177,10 @@ cc_test(
     size = "small",
     srcs = ["test_sessiond_integ.cpp"],
     flaky = True,
+    # This test unfortunately brings up a test server that collides with another test (cloud_reporter)
+    # Marking it as exclusive so that the test is not run in parallel with other tests
+    # See https://docs.bazel.build/versions/4.2.1/test-encyclopedia.html#tag-conventions
+    tags = ["exclusive"],
```
uri200, line 183: In golang to avoid that we set the port to 0. That forces golang to choose any available port.

I guess cpp does not have that feature, but maybe we could try to come up with something like that in the future.

uri200 (approved):

lte/gateway/c/session_manager/test/BUILD.bazel
themarwhal: yeah I tried to look into this at some point but didn't find anything too easy unfortunately. :(

PR #9137, opened 2021-09-16
[REVIEW NOT NECESSARY] refactor(agw): refactor MConfigLoader to enable test, add tests

themarwhal:
Modified https://github.com/magma/magma/pull/9057 to remove Bazel changes for a demo


orc8r/gateway/c/common/config/includes/MConfigLoader.h
```diff
@@ -12,41 +12,28 @@
...
-   *          passed is not the right type
-   */
-  bool load_service_mconfig(
-      const std::string& service_name, google::protobuf::Message* message);
-
- private:
-  static constexpr const char* DYNAMIC_MCONFIG_PATH =
-      "/var/opt/magma/configs/gateway.mconfig";
-  static constexpr const char* CONFIG_DIR        = "/etc/magma";
-  static constexpr const char* MCONFIG_FILE_NAME = "gateway.mconfig";
```
uri200, line 46: just curious, why are those paths migrated to .cpp instead of being at header?

orc8r/gateway/c/common/config/MConfigLoader.cpp
```diff
@@ -18,33 +18,63 @@
...
 }
 
-bool MConfigLoader::load_service_mconfig(
+void open_mconfig_file(std::ifstream* file) {
+  // Load from /var/opt/magma if config exists, else read from /etc/magma
+  if (check_file_exists(DYNAMIC_MCONFIG_PATH)) {
+    file->open(DYNAMIC_MCONFIG_PATH);
+    return;
+  }
+  const char* cfg_dir = std::getenv("MAGMA_CONFIG_LOCATION");
```
uri200, line 39: shouldn't we use  a variable to define `MAGMA_CONFIG_LOCATION`?

orc8r/gateway/c/common/config/includes/MConfigLoader.h
themarwhal: not my PR :) I just replicated it for a demo - might want to comment on Scott's original PR

PR #9149, opened 2021-09-16
style: Add check for correct Go imports order to build.py -g

emakeev:

## Summary

Add check for correct Go imports order to build.py -g

There is currently no enforcement of corect goimports order outside of a single import block (set of imports surrounded by empty lines). This PR adds such check to build -g to be run during insync check.

## Test Plan
build.py -g
CI tests

## Additional Information

### DIFF
--- a/orc8r/cloud/docker/controller/Dockerfile
+++ b/orc8r/cloud/docker/controller/Dockerfile
@@ -27,6 +27,9 @@ RUN curl https://facebookconnectivity.jfrog.io/artifactory/generic/go1.13.4.linu
     tar xf go1.13.4.linux-amd64.tar.gz && \
     cp -r go/bin/* /usr/local/bin/
 
+# Install goimports
+# RUN go get golang.org/x/tools/cmd/goimports
+
 # Protobuf compiler
 # Apt has 2.x but we need 3.x
 # See: https://grpc.io/docs/protoc-installation/
--- a/orc8r/cloud/order_go_imports.sh
+++ b/orc8r/cloud/order_go_imports.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# remove all blank lines in go 'imports' statements,
+# then sort with goimports
+
+if [ $# != 1 ] ; then
+  echo "usage: $0 <filename>"
+  exit 1
+fi
+
+sed -i '
+  /^import/,/)/ {
+    /^$/ d
+  }
+' "$1"
+
+goimports -w -local magma/ "$1"
--- a/cwf/cloud/go/services/cwf/analytics/analytics.go
+++ b/cwf/cloud/go/services/cwf/analytics/analytics.go
@@ -17,12 +17,12 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+
 	cwf_calculations "magma/cwf/cloud/go/services/cwf/analytics/calculations"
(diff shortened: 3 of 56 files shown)


orc8r/cloud/order_go_imports.sh
```diff
@@ -0,0 +1,17 @@
...
+  exit 1                                                                         
+fi                                                                               
+                                                                                 
+sed -i ''  '
+  /^import/,/)/ {
+    /^$/ d
+  }
+' "$1"
+
+goimports -w -local magma/ $1
```
uri200: How is the sort supposed to be in this case?  go library modules,  3rd party, magma modules (?)

like here ?

image

emakeev: Yes, if you remove empty lines between import blocks first, goimports sorts them correctly in this order and then re-inserts the empty lines.

uri200: ohh , it is going to be   goLibs, magma then 3rd party libs 

because the `magma` prefix at ```goimports -w -local magma/```

orc8r/cloud/docker/controller/Dockerfile
```diff
@@ -27,6 +27,9 @@ RUN curl https://facebookconnectivity.jfrog.io/artifactory/generic/go1.13.4.linu
     tar xf go1.13.4.linux-amd64.tar.gz && \
     cp -r go/bin/* /usr/local/bin/
 
+# Install goimports
+RUN go get golang.org/x/tools/cmd/goimports
```
uri200: if you add `golang.org/x/tools/cmd/goimports` to the `TOOL_DEPS` below you will get the same result i think


https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Auri200%2Fmagma.git&branch=master&file=orc8r%2Fcloud%2Fgo%2FMakefile&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=1&start_col=0&end_row=1&end_col=9

uri200 (commented):
should we add the same for feg?

orc8r/cloud/docker/controller/Dockerfile
uri200: we have the same for feg btw

https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Auri200%2Fmagma.git&branch=master&file=feg%2Fgateway%2FMakefile&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=98&start_col=0&end_row=98&end_col=0

uri200 (approved):

orc8r/cloud/docker/controller/Dockerfile
emakeev: added it to make as well, we can remove redundant part later

orc8r/cloud/docker/controller/Dockerfile
```diff
@@ -27,6 +27,9 @@ RUN curl https://facebookconnectivity.jfrog.io/artifactory/generic/go1.13.4.linu
     tar xf go1.13.4.linux-amd64.tar.gz && \
     cp -r go/bin/* /usr/local/bin/
 
+# Install goimports
+# RUN go get golang.org/x/tools/cmd/goimports
```
hcgatewood, line 31: @emakeev is this intended to stay commented out?

Issue #8904, opened 2021-08-31
Extension to Orc8r SubscriberDB


anusha-konda:
# Extension to Orc8r SubscriberDB
## Problem

UE doesn't have access restrictions to "EPC" or "5GC" network currently. 
This proposal is to extend below support for 5G subscribers, 
 1. Enabling options for operators, to map subscribers to the respective network type.

## Solution

CoreNetwork Type restriction is enabled during subscriber addition, defining the access restrictions. AMF/MME validates the details from SubscriberDB and take an action for NAS signaling procedures.

## Non-goals

This proposal only includes the detailed changes in Orc8r.  AGW changes are not specified in this proposal, it is referenced wherever required.

[Design doc](https://docs.google.com/document/d/1V7wK5OxVxv9fmDiF1_6IbCXr5hnyLDT0/edit)

hcgatewood:
Hi @anusha-konda! Thanks for opening this. Based on the design doc, I don't think this type of change needs to be a proposal -- since it doesn't need top-down community alignment or blessing

Instead, for forward-compatible changes such as this, feel free to work directly with the relevant tech lead (@ulaskozat or @uri200, for example) on the shape of the change, then put up a draft PR and I can give you more actionable Orc8r-related feedback

Let me know if you have any questions, happy to answer here or have a quick chat 🙂

anusha-konda:
> Hi @anusha-konda! Thanks for opening this. Based on the design doc, I don't think this type of change needs to be a proposal -- since it doesn't need top-down community alignment or blessing
> 
> Instead, for forward-compatible changes such as this, feel free to work directly with the relevant tech lead (@ulaskozat or @uri200, for example) on the shape of the change, then put up a draft PR and I can give you more actionable Orc8r-related feedback
> 
> Let me know if you have any questions, happy to answer here or have a quick chat 🙂

Hi @hcgatewood . can close the PR as said by you. Anyways PR 9161 (Orc8r changes) and PR 9165 (NMS Changes) are under review by Andrei Lee.

PR #9161, opened 2021-09-17
feat(orc8r): SubscriberDB Extensions - Updation to Swagger API doc, callback functions, subscriberDB service.

Abhishek-R15:
## Summary

This PR deals with the changes as mentioned in the proposal 
https://github.com/magma/magma/issues/8904

Updated Swagger API to accept the list of allowed network types for Subscriber.
By default ["NT_EPC", NT_5GC"] is configured as allowed network types for Subscriber from NMS.
Updated subscriberdb.proto to populate core network type restriction parameter to AGW.
Updated subscriber_cli.py to configure allowed network types for subscriber.
This PR depends on https://github.com/magma/magma/pull/9165 (for NMS changes)

## Order of Merge
Please merge the current PR as per the below specified order
PR #9161 - PR #9165 - PR #9775 

## Test Plan

Added Unit test cases for all the modified functions.
Verified end-to-end by configuring network type from NMS UI and listed the same using subscriber_cli.py script in AGW.

## Additional Information

Signed-off-by: Abhishek-R15 <abhishek.r@wavelabs.ai>


Abhishek-R15:
/subscriber_cli.py
image

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
```diff
@@ -7475,6 +7475,14 @@ definitions:
         format: uint64
         type: integer
     type: object
+  core_network_type_restriction:
+    items:
+      enum:
+      - NT_EPC
```
andreilee: What's the purpose of the `'NT'` prefix here? Why not just `'EPC', '5GC'`

andreilee (commented):
How will subscribers without existing values of 'Allowed Network Types' be handled?

hcgatewood (approved):
Deferring to @ulaskozat / @uri200 for domain-specific functionality

anusha-konda:
> How will subscribers without existing values of 'Allowed Network Types' be handled?

@andreilee , Existing Subscribers without any value for "Allowed Network Types" will be by default assumed to have access to both "EPC" and "5GC" networks.

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
anusha-konda: @andreilee , Will remove "NT_" and do the changes in Orc8r and NMS accordingly. This is somehow mistook with the constraint on the variables starting with number. Will correct it. Thanks.

andreilee (changes requested):
@anusha-konda Could you update the swagger.v1.yml file so that it's clear just from reading it?

@hcgatewood @electronjoe Not sure if a migration to update the subscribers with the `core network restriction` is necessary. But as this PR is stands, there's a bit of an oddity here where a subscriber can access `5GC` and `EPC` either if it has a `core network restriction` with both those types, or no `core network restriction`, which is a strange way to structure our API.

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
Abhishek-R15: @andreilee We have removed "NT_" from the swagger file, now its just 'EPC' and '5GC'.

anusha-konda:
> @anusha-konda Could you update the swagger.v1.yml file so that it's clear just from reading it?
> 
> @hcgatewood @electronjoe Not sure if a migration to update the subscribers with the `core network restriction` is necessary. But as this PR is stands, there's a bit of an oddity here where a subscriber can access `5GC` and `EPC` either if it has a `core network restriction` with both those types, or no `core network restriction`, which is a strange way to structure our API.

@hcgatewood @electronjoe @ulaskozat , could you please provide your comments here.

ulaskozat:
> > @anusha-konda Could you update the swagger.v1.yml file so that it's clear just from reading it?
> > @hcgatewood @electronjoe Not sure if a migration to update the subscribers with the `core network restriction` is necessary. But as this PR is stands, there's a bit of an oddity here where a subscriber can access `5GC` and `EPC` either if it has a `core network restriction` with both those types, or no `core network restriction`, which is a strange way to structure our API.
> 
> @hcgatewood @electronjoe @ulaskozat , could you please provide your comments here.

I tend to agree with @andreilee here. It would be better to invert the logic and instead of specifying "allowed core networks" we should have "forbidden core networks". Otherwise (i.e., if we continue to use allowed core networks) we should not let a subscriber to attach when this information is empty. This then would require a db migration for existing subscribers. @anusha-konda

anusha-konda:
> > > @anusha-konda Could you update the swagger.v1.yml file so that it's clear just from reading it?
> > > @hcgatewood @electronjoe Not sure if a migration to update the subscribers with the `core network restriction` is necessary. But as this PR is stands, there's a bit of an oddity here where a subscriber can access `5GC` and `EPC` either if it has a `core network restriction` with both those types, or no `core network restriction`, which is a strange way to structure our API.
> > 
> > 
> > @hcgatewood @electronjoe @ulaskozat , could you please provide your comments here.
> 
> I tend to agree with @andreilee here. It would be better to invert the logic and instead of specifying "allowed core networks" we should have "forbidden core networks". Otherwise (i.e., if we continue to use allowed core networks) we should not let a subscriber to attach when this information is empty. This then would require a db migration for existing subscribers. @anusha-konda

@ulaskozat , thank you for the reply. Agree on what you said. Initially, allowed network types is chosen from TS 23-501 
image
So forbidden core networks can accept list of network types to be restricted and will not enable any default value from NMS for this option, so it will be assumed that subscriber has access to all Network Types.

hcgatewood:
Supporting the input from @ulaskozat and @andreilee: let's go ahead and use "forbidden core networks" as the new field, so that its zero-value is backwards-compatible cc @anusha-konda

hcgatewood (changes requested):
As noted: let's swap to using the "forbidden core networks" pattern

Abhishek-R15:
output of subscriber_cli.py
image

anusha-konda:
@ulaskozat , @andreilee , @hcgatewood , Could you please review the changes

1. Swapped allowed_nw_types with forbidden_nw_types.
2. If forbidden network types is not configured, subscriber will have access to all Network Types by default.
3. From NMS UI, no default value will be populated for Forbidden Network per Subscriber.
4. Verified Add/Edit/Get/Delete subscriber from NMS UI

Please find corresponding NMS code changes, verified UI screenshots in PR #9165 .

lte/cloud/go/services/subscriberdb/load_test.go
```diff
@@ -39,10 +39,12 @@ func TestConvertSubEntsToProtos(t *testing.T) {
 			ApnName: "apple",
 		},
 	}
+
+	sub_nt := models.CoreNetworkType{"5GC", "EPC"}
```
hcgatewood: Nit: use camelCase in golang

lte/cloud/go/services/subscriberdb/obsidian/models/swagger.v1.yml
```diff
@@ -383,6 +387,19 @@ definitions:
         $ref: './lte-policydb-swagger.yml#/definitions/policy_ids_by_apn'
       active_apns:
         $ref: '#/definitions/apn_list'
+      forbidden_nw_types:
+        description: List of Network Types to be restricted per subscriber. If not configured, subscriber will have access to all Network Types by default.
+        $ref: '#/definitions/core_network_type'
+
+  core_network_type:
```
hcgatewood: Semantically, this is actually `core_network_types` (plural) since it's an array

lte/protos/subscriberdb.proto
```diff
@@ -159,6 +159,16 @@ message SubscriberData {
...
+
+  CoreNetworkType sub_network = 8;
+}
+
+message CoreNetworkType {
+  enum CoreNetworkTypes {
+      NT_EPC = 0;
+      NT_5GC = 1;
+  }
+  repeated CoreNetworkTypes forbidden_nw_types = 1;
```
hcgatewood: Here and throughout: I'm not a fan of abbreviating when it's not an abundantly clear abbreviation, especially for IDLs such as protos or swagger. So e.g. "nw" isn't 90%+ clear that it means "network", so let's go ahead and spell it out as "network"

lte/cloud/go/services/subscriberdb/load.go
```diff
@@ -169,6 +169,14 @@ func ConvertSubEntsToProtos(ent configurator.NetworkEntity, apnConfigs map[strin
 		AuthOpc:  cfg.Lte.AuthOpc,
 	}
 
+	const coreNwTypePrefix = "NT_"
+	subNetwork := &lte_protos.CoreNetworkType{}
+	subNetwork.ForbiddenNwTypes = make([]lte_protos.CoreNetworkType_CoreNetworkTypes, len(cfg.ForbiddenNwTypes))
+	for i, nwType := range cfg.ForbiddenNwTypes {
+		subNetwork.ForbiddenNwTypes[i] = lte_protos.CoreNetworkType_CoreNetworkTypes(lte_protos.CoreNetworkType_CoreNetworkTypes_value[coreNwTypePrefix+nwType])
```
hcgatewood: Nit: prefer sprintf over concatenation (for clarity)

hcgatewood (changes requested):
Cool, shape looks good for Orc8r. Handle the couple nits then am ready to approve!

lte/cloud/go/services/subscriberdb/load_test.go
Abhishek-R15: Updated _sub_nt_ to _subNt_ .

lte/protos/subscriberdb.proto
Abhishek-R15: Updated _forbidden_nw_types_ to _forbidden_network_types_ .

lte/cloud/go/services/subscriberdb/load.go
Abhishek-R15: Used sprintf instead of concatenation.

hcgatewood (approved):
Orc8r changes look reasonable to me 👍

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
Abhishek-R15: @andreilee We've made the necessary changes, please review this PR.

lte/protos/subscriberdb.proto
```diff
@@ -159,6 +159,16 @@ message SubscriberData {
...
+
+  CoreNetworkType sub_network = 8;
+}
+
+message CoreNetworkType {
+  enum CoreNetworkTypes {
+      NT_EPC = 0;
+      NT_5GC = 1;
+  }
+  repeated CoreNetworkTypes forbidden_network_types = 1;
```
andreilee, line 171: In general, I think that our `.proto` files can have more documentation. Please add some detailed info to our RPC/API definitions when possible, otherwise people reading these need to poke around into a lot of code to figure it out.
```suggestion
  // Network Types to be restricted per subscriber. If not configured, subscriber will have access to all Network Types by default.
  repeated CoreNetworkTypes forbidden_network_types = 1;
```

lte/cloud/go/services/subscriberdb/load.go
```diff
@@ -169,6 +170,14 @@ func ConvertSubEntsToProtos(ent configurator.NetworkEntity, apnConfigs map[strin
 		AuthOpc:  cfg.Lte.AuthOpc,
 	}
 
+	const coreNwTypePrefix = "NT_"
```
andreilee, line 173: Is there any particular reason we need this prefix? The values are set inside the field `ForbiddenNetworkTypes`, so even without the prefix, it seems clear to me that the values are network types.

andreilee (commented):
Looks good. I just have two nits.

lte/protos/subscriberdb.proto
Abhishek-R15: @andreilee Added description to the proto file

lte/cloud/go/services/subscriberdb/load.go
Abhishek-R15: This prefix is used to map swagger and proto enum values for Core Network Type. In swagger enum values are _5GC,EPC_, but in proto enum values are _NT_5GC,NT_EPC_. Hence the prefix is added here.

orc8r/cloud/go/obsidian/swagger/v1/swagger.yml
Abhishek-R15: @andreilee Can you please review the changes.

lte/cloud/go/services/subscriberdb/load.go
ulaskozat: why not then use `NT_` prefix also in swagger enums?

ulaskozat (approved):
LGTM

lte/cloud/go/services/subscriberdb/load.go
anusha-konda: Hi @ulaskozat , @andreilee suggested to remove NT_ prefix in swagger, same has been reflected in UI as well, as options in UI is referred from swagger. As it is not possible to remove NT_ in proto, we continued with.

ulaskozat: okay, I leave it up to @andreilee then.

andreilee (approved):

PR #9167, opened 2021-09-17
feat: modify devcontainer to support bazel builds

themarwhal:

## Summary
See https://github.com/magma/magma/pull/9172 for some fixes done via the Clang Tidy linting enabled by this PR. :) 

Still trying to figure out a way to add all the debug / build configurations to the devcontainer setup. Will work on those next. 

1. Added a devcontainer bazel configuration to get bazel builds to work in this env
2. Added a similar extension configuration to `vscode-workspaces/workspace.magma-vm-workspace.code-workspace` (most non-core bazel functionalities are turned off by default, as they require subscription or they are still experimental)
3. turning off cpptools extension as it collides with clangd

## Additional Information

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,14 +3,18 @@
 		"CoenraadS.bracket-pair-colorizer-2",
 		"xaver.clang-format",
 		"notskm.clang-tidy",
-		"ms-vscode.cpptools",
 		"eamodio.gitlens",
 		"yzhang.markdown-all-in-one",
 		"christian-kohler.path-intellisense",
 		"vscodevim.vim",
 		"vscode-icons-team.vscode-icons",
 		"ms-vsliveshare.vsliveshare",
-		"ms-azuretools.vscode-docker"
+		"ms-azuretools.vscode-docker",
+		"llvm-vs-code-extensions.vscode-clangd",
+		"mitaki28.vscode-clang",
+		"stackbuild.bazel-stack-vscode",
+		"coolchyni.beyond-debug",
+		"stackbuild.bazel-stack-vscode-cc",
 	],
 	"image": "ghcr.io/magma/devcontainer:sha-b435783",
 	"settings": {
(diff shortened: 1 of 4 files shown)


.devcontainer/devcontainer.json
```diff
@@ -3,14 +3,18 @@
...
 		"eamodio.gitlens",
 		"yzhang.markdown-all-in-one",
 		"christian-kohler.path-intellisense",
 		"vscodevim.vim",
 		"vscode-icons-team.vscode-icons",
 		"ms-vsliveshare.vsliveshare",
-		"ms-azuretools.vscode-docker"
+		"ms-azuretools.vscode-docker",
+		"llvm-vs-code-extensions.vscode-clangd",
+        "mitaki28.vscode-clang",
```
electronjoe: nit: whitspace alignment changed? Don't have an opinion what whitespace is nominal here, but lets matchy matchy one way or the other (maybe you have to fix up my own whitespace and didn't want the PR polluted by diff lines?).

.devcontainer/devcontainer.json
```diff
@@ -47,6 +51,35 @@
...
+			"**/.bazel-cache-repo/**": true,
+		},		
+		"bsv.bazel.buildFlags": [
+			"--config=devcontainer",
+		],
+		"bsv.bazel.testFlags": [
+			"--compilation_mode=dbg",
+		],
+		"bsv.bes.enabled": false,
+		"bsv.bzl.codesearch.enabled": false,
```
electronjoe, line 66: what does codesearch.enabled do? Sounds interesting.

.devcontainer/devcontainer.json
```diff
@@ -47,6 +51,35 @@
...
+		"bsv.bzl.invocation.buildEventPublishAllActions": false,
+		"bsv.bzl.invocation.enabled": false,
+		"bsv.bzl.invocation.invokeWithBuildEventStreaming": false,
```
electronjoe, line 69: Do you generally have a pointer for how you determined all these configuration parameters? Or did the author pass them to you?

Anything here that may change down the road? Would be nice to have comments in PR description.

.devcontainer/devcontainer.json
```diff
@@ -47,6 +51,35 @@
...
+		"bsv.bzl.remoteCache.enabled": false,
+		"bsv.bzl.starlarkDebugger.enabled": false,
+		"clangd.path": "clangd-12",
+		"clangd.arguments": [
+			"-log=verbose",
+			"-pretty",
+			"--background-index",
+		],
+		"clangd.onConfigChanged": "restart",
+		"bsv.cc.compdb.targets": [
```
electronjoe, line 82: Does this need to be updated for ... debug symbol support? Or what. If new targets need to be added here down the road lets highlight that in the PR description.

electronjoe (approved):
Fantastic! Have a few asks in PR review.

.devcontainer/devcontainer.json
themarwhal: It's code search within a Bazel target. So it's like a fancier search functionality within a scope. It might require a subscription, though. Should double check :) I was just turning off all non-core features when testing the settings.

.devcontainer/devcontainer.json
themarwhal: I'm not sure if it changes for debug symbol support. (It takes the default build configuration, so it is always debug build by default for our case).

Yes, if we add new targets, we need to edit the configurations. Will add in comments.

PR #9173, opened 2021-09-17
feat(agw): Inject state sync interval from orc8r into agw

mstre123:

## Summary

Fix Sync Interval

## Test Plan

Test AGW locally on its own
Test Connecting to orc8r and setting through rest api
Unit tests


orc8r/cloud/go/services/orchestrator/obsidian/models/defaults.go
```diff
@@ -44,6 +44,12 @@ func NewDefaultSentryConfig() *NetworkSentryConfig {
 	}
 }
 
+func NewDefaultStateConfig() *StateConfig {
```
markjen, line 47: new unit test here?

orc8r/cloud/go/services/orchestrator/obsidian/models/defaults.go
```diff
@@ -57,6 +63,7 @@ func NewDefaultNetwork(networkID string, name string, description string) *Netwo
 		DNS:          NewDefaultDNSConfig(),
 		Features:     NewDefaultFeaturesConfig(),
 		SentryConfig: NewDefaultSentryConfig(),
+		StateConfig:  NewDefaultStateConfig(),
```
markjen, line 66: is there already a unit test for this func? If not, we can add a simple one here too

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -191,3 +198,18 @@ func getVpnMconfig(gatewayConfig *models.MagmadGatewayConfigs) *mconfig_protos.O
 
 	return ret
 }
+
+func getStateMconfig(net configurator.Network, gwKey string) *mconfig_protos.State {
```
markjen, line 202: new unit test here? i.e. a test that specifically runs different scenarios on `getStateMconfig`, not as part of a larger integration test on `baseOrchestratorBuilder.Build`

orc8r/gateway/python/magma/state/state_replicator.py
```diff
@@ -265,3 +264,24 @@ async def _cleanup_deleted_keys(self):
         for key in deleted_keys:
             del self._state_versions[key]
         self._state_keys_from_current_iteration = set()
+
+    # TODO(#8806): update this once we confirm partners no longer set sync_interval service configs.
+    def resolve_sync_interval(self, service):
```
markjen: if `self` used here? if not, this can be a pure function, not defined in the class

markjen: also minor: can we add a pyre type hint here?

mstre123: yup can update - wasn't sure what the teams stance on type hints was since i didn't see them elsewhere

orc8r/cloud/go/services/orchestrator/obsidian/models/defaults.go
mstre123: i think these are just helper functions for unit tests, looking at code usages i don't see it used anywhere outside of tests

orc8r/cloud/go/services/orchestrator/obsidian/models/defaults.go
mstre123: i think these are just helper functions for unit tests, looking at code usages i don't see it used anywhere outside of tests

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
mstre123: getStateMconfig isn't exported - thus we should really be exercising all the paths through .Build which i believe we are.
Re separating out test cases for detail vs passed in values through .Build there was push back on this on the original PR
https://github.com/magma/magma/pull/8482#discussion_r687209638

orc8r/cloud/go/services/orchestrator/obsidian/models/defaults.go
markjen: oh interesting. I see it's used across packages; probably belongs in an `internal/testutil` or `internal/test/fixtures` but fine for now :)

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
markjen: strongly disagree here: targeted unit tests allow us to exercise more scenarios without a lot of overhead (all the surrounding inputs/outputs in .Build when we want to test specific scenarios in getStateMconfig)

not a make/break in this particular case, but I think we should have table-driven unit tests for private funcs (with edge condition cases) as well as typical/expected external usage unit tests for package public items.

cc @hcgatewood

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -191,3 +198,18 @@ func getVpnMconfig(gatewayConfig *models.MagmadGatewayConfigs) *mconfig_protos.O
...
+	mconfigProto := &mconfig_protos.State{
+		SyncInterval: 60,
+		LogLevel:     protos.LogLevel_INFO,
+	}
+	netConfig := net.Configs["state_config"]
+	if netConfig != nil {
+		nsConfig := netConfig.(*models.StateConfig)
```
markjen, line 209: is it possible for this cast to fail? what should happen in that case?

(right now, line 210 will panic because nsConfig will be nil if it's not a *models.StateConfig)

orc8r/gateway/python/magma/state/state_replicator.py
```diff
@@ -37,28 +37,51 @@
...
 GARBAGE_COLLECTION_ITERATION_INTERVAL = 2
 
 
+def _resolve_sync_interval(service: MagmaService) -> int:
+    # TODO(#8806): update this once we confirm partners no longer set sync_interval service configs.
+    # Sync_interval is in seconds
+    mconfig_sync_interval = service.mconfig.sync_interval
+    if mconfig_sync_interval is None or mconfig_sync_interval <= 0:
+        logging.info(
+            "mconfig_sync_interval is none, resetting to default interval %s",
```
markjen: this message is not correct when `mconfig_sync_interval <= 0`

orc8r/gateway/python/magma/state/tests/state_replicator_test.py
```diff
@@ -112,6 +114,7 @@ def setUp(self):
         asyncio.set_event_loop(self.loop)
 
         service = MagicMock()
+        service.mconfig.sync_interval = 75
```
markjen, line 116: should this value be verified somewhere?

orc8r/gateway/python/magma/state/state_replicator.py
mstre123: good catch updating

orc8r/gateway/python/magma/state/tests/state_replicator_test.py
mstre123: This is the setup function called by all the tests in the class, put this in to prevent a type error and was handling all verification in the resolver table tests below
`   if mconfig_sync_interval is None or mconfig_sync_interval <= 0: TypeError: '<=' not supported between instances of 'MagicMock' and 'int' `

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
mstre123: There are definitely strong arguments on both sides of testing private/unexported functions but if the team wants to commit to doing that we should add it to the style guide. right now though given the fact that we had to roll this change back i'll cover getStateMconfig directly

orc8r/cloud/go/services/orchestrator/obsidian/models/defaults.go
mstre123: yeah lets save that for a refactor, want to limit scope of this pr to handling the sync interval change

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
mstre123: added a nil check after the cast - and a test case for it

andreilee (approved):
Stamping for NMS auto-generated changes - this is not an Orc8r-side review

koolzz (approved):
stamping /gateway related changes

orc8r/cloud/go/services/orchestrator/obsidian/models/conversion.go
```diff
@@ -55,6 +56,9 @@ func (m *Network) FromConfiguratorNetwork(n configurator.Network) *Network {
 	if cfg, exists := n.Configs[orc8r.NetworkSentryConfig]; exists && cfg != nil {
 		m.SentryConfig = cfg.(*NetworkSentryConfig)
 	}
+	if cfg, exists := n.Configs[orc8r.StateConfig]; exists && cfg != nil {
+		m.StateConfig = cfg.(*StateConfig)
+	}
```
emakeev: this check is a bit redundant, if StateConfig key is missing, cfg will be set at 'zero value' - nil anyway.
https://golang.org/doc/effective_go#maps
It's just a style preference, up to you...

emakeev (approved):
LGTM, one stylistic suggestion inline

orc8r/cloud/go/services/orchestrator/obsidian/models/conversion.go
markjen: perhaps more interesting is the next line:

`m.StateConfig = cfg.(*StateConfig)`

should it be:

```
sc, ok := cfg.(*StateConfig)
if !ok {
  panic("wtf") // replace with useful message
}
m.StateConfig = sc
```

this would be along the lines of: "strict checks/patterns that strongly discourage the use of `map[string]interface{}`"

mstre123: makes sense - I followed the pattern the line above for NetworkSentry
```	
if cfg, exists := n.Configs[orc8r.NetworkSentryConfig]; exists && cfg != nil {
		m.SentryConfig = cfg.(*NetworkSentryConfig)
	}
```
but i agree we should be more defensive in how we approach casting.

mstre123: @emakeev removed the exists check
@markjen added a nil check to mappings in the  FromConfiguratorNetwork function.
I didn't add a panic since the code right now does not fail if only a partial mapping is made - introducing that can lead to unexpected outcomes

Issue #9178, opened 2021-09-20
devexp: turn up registry based devcontainer image


electronjoe:
Now that GitHub Code Spaces have been turned up for all members of Mamga repo - we should accelerate the startup time.  Presently the Dockerfile of `.devcontainer/Dockerfile` takes ~30+ minutes to build.  We should set up a GH Action to build this target on a schedule and push to a registry, then reconfigure our devcontainer to use it.

PR #9182, opened 2021-09-20
ci: adds scheduled GH Action docker builds for dev / CI

electronjoe:
## Summary

Working toward #9178.

The Action builds the vscode devcontainer Dockerfile and pushes it to the GH Container Registry. This allows GH Code Spaces to rapidly spin up by local registry access of the devcontainer.  This PR may be extended in follow-up to include other CI related docker images (e.g. the MME build container).

The GitHub Action is triggered on `Push` that modifies the associated Dockerfile, and also executes on a fixed schedule four times a day.

## Testing

Pushed to a novel branch of `magma/magma` and confirmed build and push to Github Container Registry, then used the container in GitHub Code Spaces with #9177. Validated successful codespace loading / operation.

### DIFF
--- a/.github/workflows/docker-push.yml
+++ b/.github/workflows/docker-push.yml
@@ -1,0 +1,64 @@
+---
+name: "Dockerfile Push"
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches:
+      -master
+    paths:
+      - .devcontainer/Dockerfile
+      - .github/workflows/docker-push.yml
+      - lte/gateway/docker/mme/Dockerfile.ubuntu20.04
+  schedule:
+    # Run four times a day (hours zero, six, twelve and eighteen)
+    - cron:  '0 0,6,12,18 * * *'
+
+jobs:
+  build_devcontainer_dockerfile:
+    env:
+      DOCKERFILE: .devcontainer/Dockerfile
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Out Repo
+        uses: actions/checkout@v2
+      - name: Docker meta
+        id: meta
+        uses: crazy-max/ghaction-docker-meta@v2
+        with:
+          # list of Docker images to use as base name for tags
+          images: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-push.yml
```diff
@@ -0,0 +1,68 @@
...
+jobs:
+  build_devcontainer_dockerfile:
+    env:
+      DOCKERFILE: .devcontainer/Dockerfile
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Check Out Repo
+        uses: actions/checkout@v2
+      -
```
quentinDERORY: indentation is a bit weird.
I personally use hyphen then definition, not hyphen line break and then something.
if we want to have this style it would be better to be the same in the entire file

.github/workflows/docker-push.yml
```diff
@@ -0,0 +1,68 @@
+---
+name: "Dockerfile Push"
+on:  # yamllint disable-line rule:truthy
+  push:
```
quentinDERORY, line 4: we might want to only execute it in the master branch.
add here:
```
branches:
  -master
```

quentinDERORY (changes requested):
LGTM but just specify the branch for the push trigger

.github/workflows/docker-push.yml
electronjoe: Thanks for catching this - good point!

.github/workflows/docker-push.yml
electronjoe: Yeah I think we have a mix across our workflow files. I'll move this file to hyphen and then def.

electronjoe: Frankly I tend not to care about these things and just want consistency - love me some auto-formatters for that reason! Wonder if there is one for GH Workflow YAML files.. Must be an opinionated YAML formatter out there somewhere.

quentinDERORY (approved):

PR #9185, opened 2021-09-20
ci: adds 386/arm/arm64 unit test validation for src/go

electronjoe:
## Summary

Add QEMU backed `386/arm/arm64` test support for src/go.

## Test

To validate the PR is actually executing on ARM - I first ran the PR without the QEMU Install step.

Without QEMU installation, we get an error https://github.com/magma/magma/pull/9185/commits/6c80759a25f050a7b9535b765f3c7d8783f3aeb1

[fork/exec /tmp/go-build1105311754/b238/config.test: exec format error](https://github.com/magma/magma/runs/3653393335?check_suite_focus=true)

### DIFF
--- a/.github/workflows/golang-build-test.yml
+++ b/.github/workflows/golang-build-test.yml
@@ -69,6 +69,40 @@ jobs:
           name: Unit Test Results
           path: "${{ github.workspace }}/src/go/test_src_go.xml"
 
+  test_src_go_qemu:
+    needs: pre_job_src_go_determinator
+    if: ${{ needs.pre_job_src_go_determinator.outputs.should_skip == 'false' }}
+    strategy:
+      matrix:
+        go-version: [1.17.x]
+        arch: [386, arm, arm64]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Install QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Setup gotestsum
+        uses: autero1/action-gotestsum@v1.0.0
+        with:
+          gotestsum_version: 1.7.0
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Run tests via qemu/binfmt
+        run: |
+          cd src/go/
+          gotestsum --format=testname --junitfile test_src_go.xml -- ./...
+        env:
+          GOARCH: ${{ matrix.arch }}
+      - name: Upload Test Results
+        uses: actions/upload-artifact@v2
+        with:
+          name: Unit Test Results arm64
+          path: "${{ github.workspace }}/src/go/test_src_go.xml"
+
+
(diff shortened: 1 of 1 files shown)


quentinDERORY (changes requested):
Note on the matrix strategy, It will not appear as one job with different parameters in the github interface but it will be looking like 3 different jobs.
I know @hcgatewood did not like that. I ended up simply having multiple similar steps where it was using the matrix variable.
If you are fine with it add a name to the job with the matrix variable so we can easily differentiate them in the UI.

electronjoe:
> I ended up simply having multiple similar steps where it was using the matrix variable.

Can you show me where you did this @quentinDERORY ? I'm not super for avoiding GitHub's "way of doing things" for the sake of UI visibility - not necessarily... GitHub may well fix the UI issue down the road - since this is their intended manner for matrix testing.

Edit with screen shot of how this looks. To me this is pretty clear and when a sub-matrix test fails it's easier to see what is failing? Versus having to parse through a string-of-tests-across-matrix that presents only a single UI failure? 
Screen Shot 2021-09-20 at 2 27 46 PM

.github/workflows/golang-build-test.yml
```diff
@@ -69,6 +69,40 @@ jobs:
           name: Unit Test Results
           path: "${{ github.workspace }}/src/go/test_src_go.xml"
 
+  test_src_go_qemu:
+    needs: pre_job_src_go_determinator
+    if: ${{ needs.pre_job_src_go_determinator.outputs.should_skip == 'false' }}
+    strategy:
+      matrix:
+        go-version: [1.17.x]
+        arch: [386, arm, arm64]
```
markjen, line 78: not sure if 386 is necessary, but i guess it doesn't hurt! :)

markjen (approved):

quentinDERORY (approved):

quentinDERORY:
> > I ended up simply having multiple similar steps where it was using the matrix variable.
> 
> Can you show me where you did this @quentinDERORY ? I'm not super for avoiding GitHub's "way of doing things" for the sake of UI visibility - not necessarily... GitHub may well fix the UI issue down the road - since this is their intended manner for matrix testing.
> 
> Edit with screen shot of how this looks. To me this is pretty clear and when a sub-matrix test fails it's easier to see what is failing? Versus having to parse through a string-of-tests-across-matrix that presents only a single UI failure?
> Screen Shot 2021-09-20 at 2 27 46 PM

okay that's good for me :) .
I personally like the matrix better but just wanted to point this behavior out.
Here is the PR I made to not use matrix on the helm check job: https://github.com/magma/magma/pull/7762

.github/workflows/golang-build-test.yml
electronjoe: My logic was that if something failed in the `arm` / `arm64` jobs, it would be nice to have a sanity check x86 build to confirm it's not QEMU or infra related. The `386` was arbitrary... Don't mind dropping though! *shruggie*

PR #9187, opened 2021-09-20
chore: update clang-tidy / oai paths in devoncatiner settings

themarwhal:

## Summary
The OAI paths are outdated, updating them to reflect the current directory structure.

## Test Plan
Not exactly sure how to test this. I'm assuming something was broken before this fix?

## Additional Information


.devcontainer/devcontainer.json
```diff
@@ -21,36 +21,36 @@
 		"terminal.integrated.shell.linux": "/bin/bash",
 		"clang-tidy.compilerArgs": [
 			// TODO: see if there is a way to use environment variables for workspace root (I tried)
-			"-I/workspaces/magma/build/c/oai/tasks/s1ap/r15/",  // Requires `make build_oai` to generate
```
electronjoe, line 24: I should have commented on this in your Bazel PR - but now with your tooling I think all of this is useless and should be removed? It was an incomplete list already.  The downside will be that CMake clang-tidy won't work but it wasn't really functional anyway.

themarwhal: ok will remove then :)

PR #9217, opened 2021-09-21
fix(mme): Removing unused files.

Vyas20:
fix(mme): Removing unused files

Removing some of the unused files from nas directory.

This PR is tested with S1AP-Tester.


ulaskozat (approved):
LGTM

PR #9224, opened 2021-09-21
feat(agw): adds li_agent Bazel build and test support

electronjoe:
## Summary

Adds support for building and testing li_agent code via Bazel.

Note the auxiliary change to 

### Validated health in devcontainer

From a Magma GH Code Space (which uses our devcontainer).

```
bazel test ... --config=devcontainer
```

### Validated health in experimental/bazel-baze/Dockerfile

From a Magma GH Code Space (which uses our devcontainer).

```
sudo docker build -t bazel-base -f experimental/bazel-base/Dockerfile .
sudo docker run -v /var/lib/docker/codespacemount/workspace/magma:/workspaces/magma -v /var/lib/docker/codespacemount/workspace/magma/lte/gateway/configs:/etc/magma -it bazel-base /bin/bash
bazel test ... --config=docker
```

### DIFF
--- a/experimental/bazel-base/Dockerfile
+++ b/experimental/bazel-base/Dockerfile
@@ -21,6 +21,7 @@ RUN apt-get update && \
         python-dev \
         zip \
         unzip \
+        uuid-dev \
         vim \
         wget \
         libssl-dev
--- a/lte/gateway/c/li_agent/src/BUILD.bazel
+++ b/lte/gateway/c/li_agent/src/BUILD.bazel
@@ -1,0 +1,86 @@
+# Copyright 2021 The Magma Authors.
+
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//lte/gateway/c/li_agent/src:__subpackages__"])
+
+cc_binary(
+    name = "main",
+    srcs = ["main.cpp"],
+    deps = [
+        ":interface_monitor",
+        "//orc8r/gateway/c/common/config:mconfig_loader",
+        "//orc8r/gateway/c/common/service303",
+    ],
+)
+
+cc_library(
+    name = "interface_monitor",
+    srcs = ["InterfaceMonitor.cpp"],
+    hdrs = ["InterfaceMonitor.h"],
+    deps = [
+        ":pdu_generator",
+        "@system_libraries//:libpcap",
+    ],
+)
+
+cc_library(
+    name = "pdu_generator",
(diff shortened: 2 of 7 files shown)


themarwhal (approved):
looks good!

third_party/system_libraries.BUILD
```diff
@@ -46,12 +46,20 @@ cc_library(
...
 
 cc_library(
     name = "libmnl",
-    srcs = ["usr/lib/x86_64-linux-gnu/libmnl.so"],
     linkopts = ["-lmnl"],
 )
+
+cc_library(
+    name = "libpcap",
+    linkopts = ["-lpcap"],
```
themarwhal, line 59: does this approach work for folly too?

electronjoe: I anticipate it does so long as you are using shared lib linking

experimental/bazel-base/Dockerfile
```diff
@@ -21,6 +21,7 @@ RUN apt-get update && \
         python-dev \
         zip \
         unzip \
+        uuid-dev \
```
themarwhal, line 24: is it possible to build these with Bazel? is it this one? https://github.com/google/uuid

electronjoe: This particular package is a linux optional one: https://en.wikipedia.org/wiki/Util-linux

themarwhal: ok it's probably not that one :D

themarwhal: also i'm surprpsied we don't need pcap here

lte/gateway/c/li_agent/src/BUILD.bazel
```diff
@@ -0,0 +1,86 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//lte/gateway/c/li_agent/src:__subpackages__"])
+
+cc_binary(
+    name = "main",
```
themarwhal, line 15: do we want the binary to be called main? For sessiond I named it sessiond so that it's consistent with our current naming.

electronjoe: Good call, will change

lte/gateway/c/li_agent/src/BUILD.bazel
```diff
@@ -0,0 +1,86 @@
...
+)
+
+cc_library(
+    name = "proxy_connector",
+    srcs = ["ProxyConnector.cpp"],
+    hdrs = ["ProxyConnector.h"],
+    # TODO(@smoeller): Migrate to using full path for includes - GH8494
+    strip_include_prefix = "/lte/gateway/c/li_agent/src",
+    deps = [
+        "//orc8r/gateway/c/common/logging",
```
themarwhal, line 58: I needed to add `"@openssl//:ssl",` to proxy_connector as a dependency for it to work in the Magma VM. Seems to have a dependency on it.

PR #9239, opened 2021-09-22
feat(sctpd): enables Bazel build and test

electronjoe:
## Test Plan

From GH Code Space devcontainer:

```
bazel test ... --config=devcontainer
```

Passes all repo tests.

### DIFF
--- a/lte/gateway/c/sctpd/src/BUILD.bazel
+++ b/lte/gateway/c/sctpd/src/BUILD.bazel
@@ -1,0 +1,107 @@
+# Copyright 2021 The Magma Authors.
+
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//lte/gateway/c/sctpd/src:__subpackages__"])
+
+cc_binary(
+    name = "sctpd",
+    srcs = ["sctpd.cpp"],
+    deps = [
+        ":config",
+        ":sctpd_downlink_impl",
+        ":sctpd_event_handler",
+    ],
+)
+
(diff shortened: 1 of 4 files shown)


lte/gateway/c/sctpd/src/BUILD.bazel
```diff
@@ -0,0 +1,107 @@
...
+
+cc_library(
+    name = "sctp_desc",
+    srcs = ["sctp_desc.cpp"],
+    hdrs = ["sctp_desc.h"],
+    # TODO(@smoeller): Migrate to using full path for includes - GH8494
+    strip_include_prefix = "/lte/gateway/c/sctpd/src",
+    deps = [
+        ":sctp_assoc",
+    ],
```
themarwhal, line 78: I've been doing single line for single item, FYI.
```suggestion
    deps = [":sctp_assoc"],
```

third_party/system_libraries.BUILD
```diff
@@ -43,3 +43,32 @@ cc_library(
         ],
     }),
 )
+
+cc_library(
```
themarwhal, line 47: Can we add these dependencies to the bazel-base Dockerfile too?

lte/gateway/c/sctpd/src/BUILD.bazel
electronjoe: We can do that, at Loon we defaulted to multi-line because the assumption was that future edits may well enlarge the dependency count and it's *slightly* easier to add when the formatting is already prepared for that =D But I don't care either way. I feel like if Buildifier accepts this then it's a don't-care?

third_party/system_libraries.BUILD
electronjoe: We can do that - but does bazel-base have a purpose any longer?

themarwhal: I still like it for building out new bazel files since it only has dependencies that we need. I think eventually we want our devcontainer docker image to be as clean as the bazel base one.

lte/gateway/c/sctpd/src/BUILD.bazel
themarwhal: Yeah either is OK I thinK!

PR #9253, opened 2021-09-22
chore: Prep for new Go build constraints syntax transition

emakeev:

## Summary

Prepare magma codebase for new Go build constraints syntax transition.

Some magma modules are using Go build constraints (build tags) of legacy style. Since, magma build system is still using old version of Go, we cannot fully upgrade to the new build constraints syntax introduced in 1.17, but we can prepare the codebase for transition as suggested in:  https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md

## Test Plan
run builds, unit tests, CI

## Additional Information


uri200 (approved):
so that will not have any effect until we move to 1.16

What will happen when in version 16 go sees both tags (same tag but defined differently)?

uri200:
> so that will not have any effect until we move to 1.16
> 
> What will happen when in version 16 go sees both tags (same tag but defined differently)?

Oh I [answer](https://www.jetbrains.com/help/go/configuring-build-constraints-and-vendoring.html) to myself 

> According to the Bug-resistant build constraints proposal, //+build will be replaced by //go:build. A transition period from //+build to //go:build syntax will last from Go version 1.16 through version 1.18. In the 1.16 version of Go, you can use either the old syntax or both syntaxes at the same time.

PR #9261, opened 2021-09-23
ci: remove docker buildx caches

electronjoe:
## Summary

Github cache use is limited to 5 GB and docker layer caching easily consumes that allocation, yet is not a build priority w.r.t. speed.  This PR drops the docker build caching so that limited cache space can be reserved for e.g. build tooling that runs regularly on commits (such as Bazel build cache).

Bazel limits cache use to 2.5 GB by default.


quentinDERORY (approved):
Is there a requirement to use the ghcr.io docker registry?
Could we use another container registry to take advantage of caching?

PR #9266, opened 2021-09-23
fix(ci): Automate space cleaning in lab nodes

quentinDERORY:

## Summary

Automate the cleaning of the node to avoid the issue where the node is lacking space

## Test Plan

integ tests run fine after the command

## Additional Information

### DIFF
--- a/circleci/fabfile.py
+++ b/circleci/fabfile.py
@@ -104,7 +104,10 @@ def integ_test(
             "awk '/virtualbox/{print $1}' | "
             "xargs -I {} vagrant destroy -f {}",
         )
-
+        # Clean some space
+        run(
+            "docker system prune -f",
+        )
         if should_test:
             if env.stack == LTE_STACK:
                 _run_remote_lte_integ_test(repo, magma_root)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ aliases:
           - master
           - develop
   - &lte_build_verify
-    paths: "orc8r lte"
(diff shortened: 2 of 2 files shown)


circleci/fabfile.py
```diff
@@ -104,7 +104,10 @@ def integ_test(
             "awk '/virtualbox/{print $1}' | "
             "xargs -I {} vagrant destroy -f {}",
         )
-
+        # Clean some space
+        run(
```
themarwhal, line 108: is it too aggresive? maybe not. I guess it only affects cwf integ test

themarwhal (approved):

circleci/fabfile.py
xjtian: This is gonna make your integ tests take way longer, if you care. In the ansible playbook in the fbcode folder where the provisioning code for these things lives, there should be a commented out task that adds this to the crontab. Can't entirely remember why I commented it out in the first place though

themarwhal: Was it to avoid removing images while they were being built / used? (I wouldn't think Docker would remove images tagged with names, but not sure if it's smart enough to recognize untagged build images that are being used)

We could also be less aggressive and remove only old**er** images, that should be fine too.

quentinDERORY: Looks like the job is still scheduled on the the first of each month, I will run it more often to be twice a month.
The command is only deleting build/images/networks that are not being referenced and used so we should be fine.
If the image is not being used but referenced by a tagged image, it will not be deleted.

PR #9288, opened 2021-09-25
fix(agw): Check for unset params while extracting status

amarpad:
## Summary

When the eNB just comes up some of the eNB status nodes might not
be populated, check before dereferencing attribute.

Seen in the field right after eNB boots:
Sep 24 02:56:03 FreedomFi-Gateway enodebd[889]:     if name_to_val[cls.ENB_STATUS].upper() == success_str:
Sep 24 02:56:03 FreedomFi-Gateway enodebd[889]: AttributeError: 'NoneType' object has no attribute 'upper'

## Test Plan

Wrote a unit test

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/enodebd/devices/freedomfi_one.py
+++ b/lte/gateway/python/magma/enodebd/devices/freedomfi_one.py
@@ -233,7 +233,8 @@ class StatusParameters(object):
         success_str = "SUCCESS"  # String constant returned by radio
         insync_str = "INSYNC"
 
-        if name_to_val[cls.DEFAULT_GW].upper() != success_str:
+        if name_to_val[cls.DEFAULT_GW] \
+                and name_to_val[cls.DEFAULT_GW].upper() != success_str:
             # Nothing will proceed if the eNB doesn't have an IP on the WAN
             serial_num = "unknown"
             if device_cfg.has_parameter(ParameterName.SERIAL_NUMBER):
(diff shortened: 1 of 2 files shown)


lte/gateway/python/magma/enodebd/devices/freedomfi_one.py
```diff
@@ -233,7 +233,8 @@ def set_magma_device_cfg(
         success_str = "SUCCESS"  # String constant returned by radio
         insync_str = "INSYNC"
 
-        if name_to_val[cls.DEFAULT_GW].upper() != success_str:
+        if name_to_val[cls.DEFAULT_GW] \
+                and name_to_val[cls.DEFAULT_GW].upper() != success_str:
```
ardzoht: One potential alternative to also do the lookup (and also avoid any `KeyError` exception) it's
```suggestion
        if name_to_val.get(cls.DEFAULT_GW, "").upper() != success_str:
```

However I'm not sure how often could we hit the said exception here (given on how this dictionary is constructed), so I'm fine with the current way of handling this.

ardzoht (approved):
lgtm, left minor question

lte/gateway/python/magma/enodebd/devices/freedomfi_one.py
amarpad: probably a better suggestion as it helps with keyerror will update

PR #9290, opened 2021-09-27
[agw] Add PowerControl and FactoryReset function parameter for Baicells eNB configs

hakeen0618:
Signed-off-by: hakeen <hanhufeng@baicells.com>

## Summary

- Add new parameters to eNB, including the following paramters:
- Reference_Signal_power,PA,PB,PowerClass
- X2 Enable/Disable
- Factory Reset
- We can set the eNB detail config from NMS eNB config, if the eNB support the params.

-  On NMS , we can config the eNB details use the PA,PB,Reference_Signal_power, as following test pic.
## Additional Information

This add enb parameters as optional for managed configs. We already test it on NMS , AGW , it should be good.
swagger
agw_data
log_enodeb
magma_pb


hakeen0618:
this is error

PR #9291, opened 2021-09-27
fix(mme): pgw api for dedicated bearer

pshelar:
This patch removes dependency of flow-dl for
dedicated bearer of inbound roaming UEs. This results
in simplified inbound roaming flows by matching on
PGW TEIDs.

## Test Plan
`make test_oai`
Ran `make integ_test` on branch and master. I got same result for both test result.

## Additional Information


ardzoht (approved):
lgtm, looks like you have to update commit tile to fit the semantic PR check

ardzoht:
Also, can you please run the according dedicated bearer s1ap tests for sanity validation? Thanks

PR #9307, opened 2021-09-27
feat: Add Sentry to Bazelified SessionD

themarwhal:

## Summary
Adding sentry-native for Bazel-built SessionD. 

### Note
1. I did try to build Sentry without relying on CMake at all, but this was extremely difficult. The main difficulty was that even if sentry-native is relatively simple to build, we also need to figure out how to build breakpad, which is significantly more complicated. This PR contains my incomplete attempt at this approach: https://github.com/magma/magma/pull/9311
2. Instead of building with pure Bazel files, I added a genrule that runs cmake and outputs .so files. This essentially just means we have a command inside the bazel build command to run cmake. The positive of this approach is that, their cmake configurations is probably well tested. So it's harder to go wrong with the build. The **potential** negative of this approach might be bad caching? I'm not noticing any significant slowdown, but I wonder if Bazel can optimize less as it's more of a black box.
3. Moved out CPP only repositories into cpp_repositories. 

## Blocked on
- [x] adding a dependency to the devcontainer Docker: https://github.com/magma/magma/pull/9321
- [x] updating the docker version on the bazel build job (and others): https://github.com/magma/magma/pull/9335

## Test Plan
Building @sentry_native//:sentry
Screen Shot 2021-09-28 at 12 38 01 PM

Built sessiond with faulty logic to see crashes get reported -> they do!

## Additional Information


experimental/bazel-base/Dockerfile
```diff
@@ -37,7 +39,7 @@ RUN apt-get install -y --no-install-recommends cmake
 ## Install Fmt (Folly Dep)
 RUN git clone https://github.com/fmtlib/fmt.git && cd fmt && \
     mkdir _build && cd _build && \
-    cmake .. && \
+    cmake .. -DFMT_TEST=0 && \
```
themarwhal, line 42: We don't need to build the tests

experimental/bazel-base/Dockerfile
```diff
@@ -14,6 +14,8 @@ RUN apt-get update && \
         build-essential \
         ca-certificates \
         curl \
+        # dependency of @sentry_native//:sentry
+        libcurl4-openssl-dev \
```
markjen, line 18: should we bazelfy transitive dependencies? (a quick googling shows that this might be a large task, so most likely not something to be done in this PR)

markjen (approved):

experimental/bazel-base/Dockerfile
themarwhal: yes we could. Theoretically, we should be able to just build curl with bazel, and point the paths out so that it's useable in sentry's cmake. probably, a lot of work though. 

I am trying to keep the dependencies in this dockerfile as small as possible so that we have an easier time bazelifying the leftover dependencies.

uri200 (approved):

PR #9327, opened 2021-09-28
feat(pipelined): Support usage reporting for 5G ovs-flows

prabinakpattnaik:

## Summary

Support Usage reporting corresponding rule and IMSI for 5G ovs-flows.
1) Based on Local_f_teid value,decided the flow should be 5G or 4G.
2) The _prepare_ruleRecord_report is used for sending 5G ovs-flows usage.
3) For configure parameters are same for 4G/5G.
4) Sessiond is only handled for storing tx/rx values and sessiond is cleanup all dangling flows.

Next PR will handle credit/Charging functionality for sessiond.

## Test Plan
UT written for sessiond and pipelined.

## Additional Information
Test logs using stub:
```
Sep 28 19:19:59 magma-dev-focal sessiond[13113]: I0928 19:19:59.654162 13132 SessionStateEnforcer.cpp:50]  Async Response received from UPF: imsi IMSI123456789012 local fteid : 2147483647
Sep 28 19:19:59 magma-dev-focal sessiond[13113]: I0928 19:19:59.656744 13113 SessionState.cpp:1587] Session: IMSI123456789012-522541 Teid: 2147483647 FSM state change from SESSION_CREATED to SESSION_ACTIVE of imsi: IMSI123456789012
Sep 28 19:19:59 magma-dev-focal sessiond[13113]: I0928 19:19:59.656790 13113 SessionState.cpp:1593]  5G specific-PDU Id: 1
Sep 28 19:19:59 magma-dev-focal sessiond[13113]: I0928 19:19:59.661947 13113 SessionStateEnforcer.cpp:560] Updated SessionStore SessionState based on UPF message  with session_idIMSI123456789012-522541
Sep 28 19:19:59 magma-dev-focal sessiond[13113]: I0928 19:19:59.662052 13113 AmfServiceClient.cpp:40] Sending Set SM Session Notification from SMF
Sep 28 19:19:59 magma-dev-focal td-agent-bit[12983]: [2021/09/28 19:19:59] [error] [io] TCP connection failed: fluentd.magma.test:24224 (Connection refused)
Sep 28 19:19:59 magma-dev-focal td-agent-bit[12983]: [2021/09/28 19:19:59] [error] [output:forward:forward.1] no upstream connections available
Sep 28 19:19:59 magma-dev-focal td-agent-bit[12983]: [2021/09/28 19:19:59] [ warn] [engine] failed to flush chunk '12983-1632856799.662898672.flb', retry in 11 seconds: task_id=0, input=tail.3 > output=forward.1
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: INFO:root:ofp_event.EventOFPFlowStatsReply
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: INFO:EnforcementStatsController:Sending RPC payload: RuleRecordTable{
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: epoch: 1632856782
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: }
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: INFO:EnforcementStatsController:Sending RPC payload: RuleRecordTable{
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: records {
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   sid: "IMSI123456789012"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   rule_id: "internal_default_drop_flow_rule"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   bytes_tx: 28
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   ue_ipv4: "192.168.128.11"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   rule_version: 1
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   teid: 2147483647
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: }
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: records {
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   sid: "IMSI123456789012"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   rule_id: "internal_default_drop_flow_rule"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   ue_ipv4: "192.168.128.11"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   teid: 2147483647
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: }
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: epoch: 1632856782
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: }
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: INFO:root:_prepare_ruleRecord_report
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: INFO:root:records {
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   sid: "IMSI123456789012"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   rule_id: "internal_default_drop_flow_rule"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   bytes_tx: 28
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   ue_ipv4: "192.168.128.11"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   rule_version: 1
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   teid: 2147483647
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: }
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: records {
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   sid: "IMSI123456789012"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   rule_id: "internal_default_drop_flow_rule"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   ue_ipv4: "192.168.128.11"
Sep 28 19:19:59 magma-dev-focal pipelined[13631]:   teid: 2147483647
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: }
Sep 28 19:19:59 magma-dev-focal pipelined[13631]: epoch: 1632856782
Sep 28 19:19:59 magma-dev-focal sessiond[13113]: I0928 19:19:59.907781 13154 UpfMsgManageHandler.cpp:90] SendReportRuleStats
Sep 28 19:19:59 magma-dev-focal sessiond[13113]: I0928 19:19:59.907925 13154 UpfMsgManageHandler.cpp:91] Size of record ::2
Sep 28 19:19:59 magma-dev-focal sessiond[13113]: I0928 19:19:59.909451 13113 LocalSessionManagerHandler.cpp:70] Aggregating 0 records
```


lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -472,13 +468,13 @@ def _handle_flow_stats(self, stats_msgs):
             return
         # Send report even if usage is empty. Sessiond uses empty reports to
         # recognize when flows have ended
-        self._report_usage(current_usage)
+        self._report_usage(current_usage[1])
```
koolzz: ?

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -472,13 +468,13 @@ def _handle_flow_stats(self, stats_msgs):
...
 
         # This is done primarily for CWF integration tests, TODO rm
         self.total_usage = current_usage
         # Report only if their is no change in version
         if self.ng_config.ng_service_enabled == True:
-            self._prepare_session_config_report(stats_msgs)
+            self._prepare_ruleRecord_report(current_usage[0])
```
koolzz: why are you taking only the first element

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -544,7 +540,7 @@ def _get_usage_from_flow_stat(self, flow_stats):
         Update the rule record map with the flow stat and return the
         updated map.
         """
-        current_usage = defaultdict(RuleRecord)
+        current_usage = defaultdict(lambda: defaultdict(RuleRecord))
```
koolzz: what does this do?

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -572,55 +568,107 @@ def _get_usage_from_flow_stat(self, flow_stats):
...
+                elif ipv6_addr:
+                    key += "|" + ipv6_addr
+
+                rule_version = _get_version(flow_stat)
+                if not rule_version:
+                    rule_version = 0
+
+                key += "|" + str(rule_version)
+
+                current_usage[1][key].rule_id = rule_id
```
koolzz: why is this indexed by 1?

koolzz (changes requested):
Please refactor the enforcement_stats.py logic. I don't see any major differences between the logic in the ` if local_f_teid_ng: \ else ` branches, they're virtually identical? keep the logic shared, whats the difference in terms of 5g processing? Also a lot of weird indexing going on, why do we have current_usage[1]?

lte/gateway/python/magma/pipelined/tests/test_enforcement_pull_stats_5g.py
```diff
@@ -0,0 +1,219 @@
...
+            pkt_sender.send(packet2)
+        
+        enf_stat_name = [
+            imsi + '|rule1' + '|' + sub_ip + '|' + "1",
+        ]
+        stats = get_enforcement_stats(
+            self.enforcement_stats_controller._prepare_ruleRecord_report.call_args_list,
+        )
+        #self.assertEqual(stats[enf_stat_name[0]].sid, imsi)
+        #self.assertEqual(stats[enf_stat_name[0]].rule_id, "tx_match")
```
koolzz: clean up the linter warnings

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
```diff
@@ -857,4 +860,225 @@ void SessionStateEnforcer::set_pdr_attributes(
           .redirect_server_address());
 }
 
+void SessionStateEnforcer::aggregate_records(
```
uri200, line 931: All the functions here are copy&paste from our current code [here](sourcegraph.com/-/editor?remote_url=git%40github.com%3Auri200%2Fmagma.git&branch=feg_merge_credit_reports_into_single_crru&file=lte%2Fgateway%2Fc%2Fsession_manager%2FLocalEnforcer.cpp&editor=JetBrains&version=v1.2.1&utm_product_name=CLion&utm_product_version=2021.1.2&start_row=375&start_col=25&end_row=375&end_col=25). We need to reuse the code if possible.
This will have to be refactored so current functions are able to handle reporting of 5g

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
```diff
@@ -857,4 +860,225 @@ void SessionStateEnforcer::set_pdr_attributes(
...
+              << " dangling sessions in PipelineD";
+  for (const RuleRecord& record : dead_sessions_to_cleanup) {
+    deactivate_flows_for_termination(
+        record.sid(), record.ue_ipv4(), record.ue_ipv6(), record.teid());
+  }
+
+  return;
+}
+
+void SessionStateEnforcer::deactivate_flows_for_termination(
```
uri200, line 1033: This is a very similar from [pipelined client](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Auri200%2Fmagma.git&branch=chore_federated_lte_test_1&file=lte%2Fgateway%2Fc%2Fsession_manager%2FPipelinedClient.cpp&editor=JetBrains&version=v1.2.1&utm_product_name=CLion&utm_product_version=2021.1.2&start_row=328&start_col=35&end_row=328&end_col=35) 

This code, if different than existing, should go on pipelined client

uri200 (changes requested):

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
prabinakpattnaik: For current implementation, local_f_teid_ng value is only distinguish the 4G or 5G flows.
The usage report is sending to sessiond using grpc methods are different for 4G and 5G.
The class object are different for 4G and 5G for session creation/storing/termination.

If we are sending all flows to both then sessiond of 4g object will start for cleanup all dangling flows. That all belongs to 5g flows.
and vice-versa.

current_usage = defaultdict(lambda: defaultdict(RuleRecord))
Here, We defined list of defaultdict for index 0 belongs to 5g flows and for index 1 belongs to 4g flows.

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
prabinakpattnaik: I explained in above. please suggest if will have better approach.

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
prabinakpattnaik: Defined list of defaultdict for index 0 belongs to 5g flows and for index 1 belongs to 4g flows

lte/gateway/python/magma/pipelined/tests/test_enforcement_pull_stats_5g.py
prabinakpattnaik: I will do fix it.

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
prabinakpattnaik: Yes @uri200 almost all functions are similar. 
LocalEnforcer object is used for storing for 4g sessions.
SessionStateEnforcer object is used for storing for 5g sessions.

We can not use local defined functions of 4g for 5g. Will not get 5g session info.

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
prabinakpattnaik: The functionality is different.
The "SessionStateEnforcer::deactivate_flows_for_termination" is filling up the SessionState::SessionInfo structure and calling to "pipelined_client_->set_upf_session" as similar to create/delete/INACTIVE.

lte/gateway/python/magma/pipelined/tests/test_enforcement_pull_stats_5g.py
prabinakpattnaik: Fixed it.

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
koolzz: This seems hacky, we can either 
 - tag the RuleRecord with 5g flag if that helps?
 - do 2 separate report calls but lets rename it to current_usage/current_usage_5g or something.  
 
 @uri200 @themarwhal  what would you prefer on sessiond side?

prabinakpattnaik: @koolzz
1.
```
message RuleRecord {
  // IMSI. This value along with teid uniquely identifies a session.
  string sid = 1;
  string rule_id = 2;
  uint64 bytes_tx = 3;
  uint64 bytes_rx = 4;
  // Used for HE
  string ue_ipv4 = 5;
  // Used for HE
  string ue_ipv6 = 6;
  uint64 dropped_tx = 7;
  uint64 dropped_rx = 8;
  // Rule version with which the flow was installed with
  uint64 rule_version = 9;
  // Uplink TEID
  uint32 teid = 10;
  //Indicates that the report is final and the flow will be deleted
  bool final_record = 11;
}
```

teid parameter is only used by 5G as of now, We can use this variable. But 
We need to put the check for teid based RuleRecord will be ignored in LocalEnforcer for 4G, 
Otherwise 4g object will start for cleanup all dangling flows.

If we will put the below check in LocalEnforcer.cpp then not required above changes in pipeliend.
``` 
const uint32 teid = record.teid();
if teid:
   continue;
```   

2. We start off a thread to periodically poll stats from Pipelined in sessiond level if "enable_pull_stats" is true.
    periodic_stats_requester->start_loop(local_enforcer, interval);
   So, Not a good approaches for start new loop for 5G.

@uri200 , @themarwhal , please suggest your view.

uri200: Hey @prabinakpattnaik  please use github format on your comments as suggested previously. It make the comments readable. I fixed the comment for you so you can see the difference now (github syntax description [here](https://docs.github.com/en/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax)

koolzz: Also let's not differentiate 5g vs not just based on teid as we might soon add that to the 4g flows. Let's explicitly add the 5g flag if needed

uri200: I do not have enough context on pipelined side. But the question is: are 4g or 5g flows different? 

Why is the usage report sent using different GRPC call?  To my understanding reporting traffic is the same, no matter if this is 4g, 5g or wifi.  Pipelined should not have leaks of 5g or 4g. The distinction, if necessary should be done at sessiond level

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
uri200: it looks like the link I included was bad. I fixed it and include more hre. Please see how pipelined client works and see how we host there the code to build the request too.

See how we have different functions creating different request. For 5g I only see `set_upf_session` but for 4g i see different functions that build specific request based on different arguments


For example [deactivate_flows_for_rules](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Auri200%2Fmagma.git&branch=chore_federated_lte_test_1&file=lte%2Fgateway%2Fc%2Fsession_manager%2FPipelinedClient.cpp&editor=JetBrains&version=v1.2.1&utm_product_name=CLion&utm_product_version=2021.1.2&start_row=344&start_col=45&end_row=344&end_col=45) and [deactivate_flows_for_rules_for_termination](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Auri200%2Fmagma.git&branch=chore_federated_lte_test_1&file=lte%2Fgateway%2Fc%2Fsession_manager%2FPipelinedClient.cpp&editor=JetBrains&version=v1.2.1&utm_product_name=CLion&utm_product_version=2021.1.2&start_row=328&start_col=35&end_row=328&end_col=35) they do similar steps, but create different types of request:
1. build a request based on some inputs
2. run `deactivate_flows(req)`

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
uri200: please elaborate with `Will not get 5g session info.` 

Sessiond should be able to know the type of record without us copying&pasting code

Please, point at the code below the lines you think will not work on 4g. I am seeing exactly the same code

prabinakpattnaik: From record, we are trying to find the session information for corresponding IMSI and IP. [here](https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/LocalEnforcer.cpp#L390)

The `session_store_` is a private object of SessionStore for `LocalEnforcer` (4g class ) and for `SessionStateEnforcer` (5g class) for storing all sessions information respective to active calls whether it is 4g or 5g. 
For 4g: 
[session_store_](https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/LocalEnforcer.h#L362)
[definition here](https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/LocalEnforcer.cpp#L71)

For 5g:
[session_store_](https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/SessionStateEnforcer.h#L190)
[definition here ](https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/SessionStateEnforcer.cpp#L69)
 
Similar [SessionState](https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/SessionStateEnforcer.cpp#L100) object created for different context for 4g and 5g during initializing the session.

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
prabinakpattnaik: For 4g, activate/deactivate flow requests are going by grpc (sessiond-pipelined) and gtp tunnel creation/deletion requests are going by OAI interface (MME-Pipelined).

For 5g, gtp tunnel + activate/deactivate flow requests are going by grpc (sessiond-pipelined).

As per the design proposed by Ulas, the only one grpc method will be required to implement for sending all active sessions information to UPF (Just like snapshot of all active sessions).

Please find the attached ppt for reference.
[Discussion Material for Meeting 06102020 UPF-SMF GRPC Interface (1).pptx](https://github.com/magma/magma/files/7293596/Discussion.Material.for.Meeting.06102020.UPF-SMF.GRPC.Interface.1.pptx)

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
uri200: The problem is from the initial design. We should have never had have a different store for 4g sessions and 5g sessions. Their content is the same. Now we have 2 classes doing exactly the same

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
uri200: The actions `deactivate_flows_for_termination` are of the same nature as `deactivate_flows_for_rules`

1. build a request based on some inputs
2. run deactivate_flows(req)

You should create a function similar to `deactivate_flows_for_rules` that intakes some arguments and create your `sess_info` (is your request) and runs `set_upf_session`

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
prabinakpattnaik: Yes, ACL team is implemented the separate class for 5g. The new class (5g) is doing little bit different from 4g based on grpc message request coming from AMF.

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
prabinakpattnaik: The `deactivate_flows_for_termination` function is created a default PDRs (`SetGroupPDR`) from private object of rulestore of 5g `ConvergedRuleStore` then push the `SetGroupPDR` into `sess_info` and runs `set_upf_session`.

`
 sess_info.pdr_rules.push_back(reqpdr2); 
 pipelined_client_->set_upf_session(sess_info, call_back_upf);
`

uri200: cant we use `CopyFrom`

https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Auri200%2Fmagma.git&branch=master&file=lte%2Fgateway%2Fc%2Fsession_manager%2FSessionState.cpp&editor=JetBrains&version=v1.2.1&utm_product_name=CLion&utm_product_version=2021.1.2&start_row=813&start_col=31&end_row=813&end_col=39

prabinakpattnaik: We cann't use `copyFrom` because we are suppose to terminating the sessions those are not in a sessiond store (`dangling sessions`) and will not get session information from sessiond store.
These sessions are mismatch between `pipelined` and `sessiond`.

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
prabinakpattnaik: @koolzz @uri200 Can we add 5g flag in `RuleRecord` message and will ignore all 5g flows in `Localenforce` (4g class). Then not required to change in `pipelined`.

uri200: Per my understanding Pipelined does not have 4g or 5g concept, but @koolzz  @pshelar  can confirm. 

In my opinion that design is not proper. Why should pipelined care a flow is 4g or 5g? Sessiond should know what this flow belongs too. We can not leak needs of one system to other systems around

Note, the root cause of our issues like copy&pasted functions code and leaking parameters is having separate objects in sessiond to handle 4g and 5g.

pshelar: agree with @uri200 , why sessiond can not differentiate between 4G and 5G flow stats?

prabinakpattnaik: @pshelar , @uri200 Please find the below points for `why should pipelined care a flow is 4g or 5g`

1.  At the onset it was decided that 5G will be based on set interface. UPF will periodically send SMF,  information about `5G flows + version`. SMF will validate it based on its local structure.

2.   We got the input from the sessiond team that they need a differentiating parameter (at that time we were having only UPF) both for validation and reporting. So we introduced the `reg11` register variable for storing the `local_f_teid` value for distinguish whether it is 4G call flow or 5G call flow.

3.   We have defined two class of objects which are storing 4g sessions info and 5g session info respectively in sessiond.

4.  Also as the tunnel ID generator was different for 4G and 5G it was decided to have a range division between them (5G teid starts with 2^32/2)

With the current design there are 2 alternates to solve the issue :
1.  To have  a flag/differentiatior from UPF to Sessiond where sessiond knows which data structure to look for (4G or 5G)
2.  Or `Sessiond` takes a decision based on the `teid` ( 4G == < 2^32/2 ,,, 5G == > 2^32/2 + 1)

Please provide your inputs.

pshelar: I dont think we can segment the TEID space get sessionD level, that would conflict with orc8r based TEID allocation. 
we need to have the state in sessionD. is there any existing map that would help here?

prabinakpattnaik: I have added `flag5g` in `RuleRecord` message and pipelined set this flag as a true for 5g flow.
Sessiond used the `RuleRecord` for 4g or 5g depends the `flag5g` value.
For 4g:
`
if (record.flag5g() == true) continue;
`
for 5g:
`
if (record.flag5g() == false || teid == 0) continue;
`
Now I removed all hacky type code in piepelined.
Please review it.

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
```diff
@@ -597,6 +601,7 @@ def _get_usage_from_flow_stat(self, flow_stats):
             elif ipv6_addr:
                 current_usage[key].ue_ipv6 = ipv6_addr
             if local_f_teid_ng:
+                current_usage[key].flag5g = True
```
pshelar, line 604: our plan is to converge flow table for 4G and 5G, in such case how to we set this flag?

prabinakpattnaik: @pshelar We are using `NG_SESSION_ID_REG = 'reg11'` for set the value of local_f_teid_ng for 5g flows. As below:
```
def _get_ng_local_f_id(flow):
    if NG_SESSION_ID_REG not in flow.match:
        return None
    return flow.match[NG_SESSION_ID_REG]
```
`local_f_teid_ng = _get_ng_local_f_id(flow_stat)`

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
uri200: When I say `copyFrom` I meant
```
reqpdr2->CopyFrom(reqpdr1);

```

prabinakpattnaik: Ok. But We can't do ` reqpdr2->CopyFrom(reqpdr1);`  because `reqpdr1` is for uplink pdr and few parameters are different from `reqpdr2` which is used for downlink pdr. Below few parameters different like as:
```
reqpdr1.mutable_pdi()->set_net_instance("uplink");
reqpdr1.mutable_pdi()->set_src_interface(ACCESS);
reqpdr1.mutable_pdi()->set_local_f_teid(teid);
reqpdr1.mutable_activate_flow_req()->set_uplink_tunnel(teid);

reqpdr2.mutable_pdi()->set_src_interface(CORE);
reqpdr2.mutable_pdi()->set_net_instance("downlink");

```

uri200: Can you then label `reqpdr` with better naming like `reqpdr_uplink` `reqpdr_downlink`

lte/gateway/c/session_manager/test/test_set_session_manager_handler.cpp
```diff
@@ -661,6 +669,139 @@ TEST_F(SessionManagerHandlerTest, test_SetAmfSessionAmbr) {
...
+  create_rule_record(
+      IMSI1, cfg.common_context.ue_ipv4(), "rule2", 10, 20, record_list->Add());
+
+  auto update_1 = SessionStore::get_default_session_update(session_map);
+  auto uc       = get_default_update_criteria();
+  // session_map[IMSI1][0]->increment_rule_stats("rule1", &uc);
+  session_enforcer->aggregate_records(session_map, table, update_1);
+  EXPECT_EQ(table.records_size(), 2);
+}
+
```
uri200, line 820: Can you please add any missing mock call to all your test?. Use `EXPECT_CALL` like in the example below

https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Auri200%2Fmagma.git&branch=orc8r_add_demap_endpoints_to_directoryd&file=lte%2Fgateway%2Fc%2Fsession_manager%2Ftest%2Ftest_local_enforcer.cpp&editor=JetBrains&version=v1.2.1&utm_product_name=CLion&utm_product_version=2021.1.2&start_row=2158&start_col=20&end_row=2158&end_col=20

prabinakpattnaik: Ok.Sure. I will look into it. If something is missing then will add it.

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
prabinakpattnaik: Ok. I will fix it.

lte/gateway/c/session_manager/test/test_set_session_manager_handler.cpp
prabinakpattnaik: I have added mock call.

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
prabinakpattnaik: Fixed it.

lte/gateway/c/session_manager/test/test_set_session_manager_handler.cpp
```diff
@@ -672,6 +683,229 @@ TEST_F(SessionManagerHandlerTest, test_SetAmfSessionAmbr) {
...
+  auto record_list = table.mutable_records();
+  create_rule_record(IMSI1, IP1, "rule1", 512, 512, record_list->Add());
+
+  upf_session_manager->SendReportRuleStats(
+      &server_context, &table,
+      [this](grpc::Status status, SmContextVoid Void) {});
+  evb->loopOnce();
+
+  session_map_2 = session_store->read_sessions(SessionRead{IMSI1});
+  EXPECT_EQ(session_map_2[IMSI1].front()->get_request_number(), 1);
```
uri200, line 750: This test is the same as an existing test in our current code base [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Auri200%2Fmagma.git&branch=orc8r_add_demap_endpoints_to_directoryd&file=lte%2Fgateway%2Fc%2Fsession_manager%2Ftest%2Ftest_session_manager_handler.cpp&editor=JetBrains&version=v1.2.1&utm_product_name=CLion&utm_product_version=2021.1.2&start_row=399&start_col=28&end_row=399&end_col=28) but adapted for 5G which is fine. But, why is the reason to remove all the `EXPECT_CALL`? 

If  for some reason `SendReportRuleStats`  do not implement the action of sending a message to the PCF yet, please set a `TODO` on that test so we can add the EXPECT_CALL in the future

lte/gateway/c/session_manager/test/test_set_session_manager_handler.cpp
```diff
@@ -672,6 +683,229 @@ TEST_F(SessionManagerHandlerTest, test_SetAmfSessionAmbr) {
...
+  // Check the request number
+  auto session_map_2 = session_store->read_sessions(SessionRead{IMSI1});
+  EXPECT_EQ(session_map_2[IMSI1].front()->get_request_number(), 1);
+
+  // ReportRuleStats
+  RuleRecordTable table;
+  auto record_list = table.mutable_records();
+  create_rule_record(IMSI1, IP1, "rule1", 512, 512, record_list->Add());
+
+  upf_session_manager->SendReportRuleStats(
```
uri200, line 744: missing `EXPECT_CALL` that checks if rules are sent and if sent rules are proper?

uri200, line 744: shouldnt `SendReportRuleStats` check if the report is sent with an `EXPECT_CALL`

lte/gateway/c/session_manager/test/test_set_session_manager_handler.cpp
prabinakpattnaik: I have removed few EXPECT_CALL because Charging, Credit and event report functionality not yet support in 5G class. It will add all functionality soon then will add UT for corresponding functionality.

I have set a ```TODO``` for same.

lte/gateway/c/session_manager/test/test_set_session_manager_handler.cpp
prabinakpattnaik: The ```SendReportRuleStats``` is a grpc method received in a sessiond for usage info from pipelined.

lte/gateway/c/session_manager/test/test_set_session_manager_handler.cpp
prabinakpattnaik: The sending report towards PCF functionality not yet supported completely. Will add soon.

lte/gateway/c/session_manager/test/test_set_session_manager_handler.cpp
prabinakpattnaik: I have added one missing ```EXPECT_CALL``` for ```set_upf_session```.

lte/gateway/c/session_manager/test/test_set_session_manager_handler.cpp
uri200: Can you please add a TODO to the test so we know it needs to be added?

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
uri200: Hey @prabinakpattnaik why sessiond can't infer if this is a 4g or 5g traffic using the name of the rule?

lte/gateway/c/session_manager/test/test_set_session_manager_handler.cpp
prabinakpattnaik: Yes, I have added the TODO for both `report_updates` & `events_reporter`.

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
prabinakpattnaik: @uri200 pipelined is sending all stats information (sid, rx, tx) towards sessiond of grpc call of both 4g/5g.
So, sessiond has to required either flag or other mechanism to decide that this `sid` belongs to 4g or 5g.

uri200: Hey @prabinakpattnaik this is what pipelined sends to sessiond 

```
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]: I1026 12:36:08.077128 3072151 GrpcMagmaUtils.cpp:58]
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:   magma.lte.RuleRecordTable {
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:       records {
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:         sid: "IMSI999991234567810"
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:         rule_id: "PCC104-QCI9-DYNAMIC"
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:         bytes_tx: 1000
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:         ue_ipv4: "172.17.141.77"
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:         rule_version: 1
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:       }
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:       records {
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:         sid: "IMSI999991234567810"
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:         rule_id: "internal_default_drop_flow_rule"
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:         ue_ipv4: "172.17.141.77"
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:       }
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:       epoch: 1634315650
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:       update_rule_versions: true
Oct 26 12:36:08 ubuntuAGW sessiond[3071932]:   }
```

As you can see you have IMSI and rule id. With those two elements you can find if this rule is 4g or 5g. No need to spill 5g concepts to pipelined

You can build a function that can search first in 4g session for that imsi and go over all the  sessions and rules, otherwise search find that imsi on 5g and go over all the sessions and rules

Fortunately each UE will have few sessions and each sessions will have few rules so the search is not going to be inpactful.

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
uri200: In any case, having a hardcoded index is really bad practice. This can crash anytime.

In case this were be necessary, please find a better way (wrap it on an object that handles corner cases like empty string). But first, lets make sure why is this really needed

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
uri200: same here. Hardcoded index are the best way to crash an app. Let's find a better way

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
prabinakpattnaik: @uri200 , already I removed all hacky types of code in pipelined.
And I have added `flag5g` in RuleRecord message and pipelined set this flag as a true for 5g flow.
Sessiond used the RuleRecord for 4g or 5g depends the flag5g value.
For 4g:
``` if (record.flag5g() == true) continue; ```
for 5g:
``` if (record.flag5g() == false || teid == 0) continue;  ```

File: lte/gateway/python/magma/pipelined/app/enforcement_stats.py (hunk 1 above)
Comment (prabinakpattnaik) [reply lte/gateway/python/magma/pipelined/app/enforcement_stats.py]: Yes, already i removed all hacky codes.

File: lte/gateway/python/magma/pipelined/app/enforcement_stats.py (hunk 4 above)
Comment (prabinakpattnaik) [reply lte/gateway/python/magma/pipelined/app/enforcement_stats.py]: This code not present.

File: lte/gateway/python/magma/pipelined/app/enforcement_stats.py (hunk 2 above)
Comment (prabinakpattnaik) [reply lte/gateway/python/magma/pipelined/app/enforcement_stats.py]: Yes, This code is not present.

File: lte/gateway/python/magma/pipelined/app/enforcement_stats.py (hunk 1 above)
Comment (uri200) [reply lte/gateway/python/magma/pipelined/app/enforcement_stats.py]: What about the need of adding any flag.  Can you avoid that please. Pipelined should not see the difference. Sessiond should be the one handling the routing to the proper session/flow for the rule to be propcessed.

RuleRecordTable should be the same for 4g or 5

File: lte/protos/session_manager.proto [hunk 12]
```diff
@@ -40,6 +40,7 @@ message RuleRecord {
   uint32 teid = 10;
   //Indicates that the report is final and the flow will be deleted
   bool final_record = 11;
+  bool flag5g = 12;
```
uri200, line 43: we need to find a better way than that.
We can not spill 4g/5g concepts outside sessiond

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
uri200: If we need a flag, this system will only work when `enable_pull_stats` is true. Having this forces us to work on a single mode which is untested and it is not even reliable



Can you point out when this decision was made?

> We start off a thread to periodically poll stats from Pipelined in sessiond level if "enable_pull_stats" is true.
periodic_stats_requester->start_loop(local_enforcer, interval);
So, Not a good approaches for start new loop for 5G.

prabinakpattnaik: If we are sending RuleRecords to sessiond without flag, then sessiond of 4g object or 5g object will start for cleanup all dangling flows which are not present to that object.

prabinakpattnaik: It's working for both mode. Verified it. It's not dependent with flag

uri200: The problem starts when having duplicated code for both 4g and . That always indicates something is wrong. 5g sessions should have never been stored in a different object than 4g sessions. We should have extended 4g session store to handle all 4g and 5g. Same way we store wifi sessions.

uri200: `void LocalSessionManagerHandlerImpl::ReportRuleStats(`  should be able to divide the request into 4g and 5g before running `aggregate_records`

But again, if the design were proper that would be taken transparently by session store itself.  We have to move thowards that kind of desing where sessiond is just a store of stats and policies, no matter if they are 3g, 4g, 5g

lte/protos/session_manager.proto
```diff
@@ -1124,7 +1125,7 @@ message UPFPagingInfo {
 // Set Interface from Pipelined
 service SetInterfaceForUserPlane {
     rpc SetUPFNodeState(UPFNodeState) returns (SmContextVoid) {}
-    rpc SetUPFSessionsConfig(UPFSessionConfigState) returns (SmContextVoid) {}
+    rpc SendReportRuleStats(RuleRecordTable) returns (SmContextVoid) {}
```
uri200, line 1128: does it mean pipelined now have to know to report either to this function or `SendReportRuleStats` ?  

This causes a spillage of concepts that should be contained in sessiond into other systems.

uri200: I see the idea is to move all 4g content into 5g in the future. In other words, once 5g works the same as 4g, the effort will be to move 4g functionalities into 5g code. This way we will remove all the duplicity Can you confirm that this is the objective with your team?

uri200:
Hey @prabinakpattnaik  per our discussion we are going to try to use the same session store. This way all sessions will be on the same place.

You will have to extend the find_session [function](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=lte%2Fgateway%2Fc%2Fsession_manager%2FSessionStore.cpp&editor=JetBrains&version=v1.2.1&utm_product_name=CLion&utm_product_version=2021.1.2&start_row=217&start_col=29&end_row=217&end_col=29). The options are 
- create a new case `IMSI_AND_5gTEID`
- Just extend the current `IMSI_AND_TEID` so the function searches either on 4g or 5g sessions. Per my understanding we can't have same TEID for 4g and 5g so whatever teid is reported by pipelined is unique

With this we will allow 4g and 5g to report using the same exact function and maybe the code on this PR can be reduced drastically:

- Please, create a separate PR with the same session store so you can test it properly.  You may need to modify/add some test cases

- Once it is working fine create another PR with the changes on `ReportRuleStats` to handle both cases (reporting using ip, reporting using teid)

prabinakpattnaik:
@uri200 @pshelar @koolzz , Raised the new PR with small modification of sessiond for re-use session store of 4G for usage report. Hence there was no changes required in pipelined for usage report, monitoring etc functionality for 5G.
The gRPC ReportRuleStats method handled for 4g and 5g usage report from pipelined.
Below is the new PR:
https://github.com/magma/magma/pull/10239.
So will mark this PR as a close.

PR #9328, opened 2021-09-28
feat(s8_proxy): add multiple bearers on Create Bearer request and Delete

### LINKED ISSUE #8445: [feg] add Delete Bearer Request

uri200:

## Summary

After issue with parsing multiple Bearer IEs was fixed at go-gtp https://github.com/wmnsk/go-gtp/issues/192 now we add support for multiple Delete Bearer Request

Note this PR do not include support for multiple Create Bearer Request. That will come later since it may need more 

## Test Plan

`./build.py -g` at orc8r
`./build.py -c` at feg

## Additional Information


feg/protos/s8_proxy.proto
```diff
@@ -238,7 +238,7 @@ message DeleteBearerResponsePgw {
     uint32 c_pgw_teid = 4;
     // control plane TEID given by PGW during CreateSession
     uint32 linked_bearer_id = 5;
-    BearerContext bearer_context = 6;
+    repeated BearerContext bearer_context = 6;
```
themarwhal, line 241: is this safe?

themarwhal (approved):

feg/protos/s8_proxy.proto
uri200: yep, we haven't even finished the feature on AGW 😭  , so nobody is using it but my tests.
So lets change it here.

PR #9329, opened 2021-09-28
feat: add hostname tag to sentry events

### LINKED ISSUE #9324: Sentry C++: Add hostname to sentry events
Read the environment variable HOSTNAME and set it as an event tag.

edit https://github.com/magma/magma/blob/master/orc8r/gateway/c/common/sentry/SentryWrapper.cpp#L110 to add probably something like
```
    if (const char* hostname = std::getenv(HOSTNAME_ENV)) {
      sentry_set_tag(HOSTNAME, hostname));
    }
```

milapjoshi:
Signed-off-by: Milap Joshi <milapjoshi@gmail.com>

## Summary
This PR adds hostname tag to be sent to sentry events. This will help identify from where sentry issue came from.

## Test Plan

tested on current HIL GW,

`magma@hil-1:~$ ./hello
ENV VAR : hil-1
magma@hil-1:~$ cat hello.cpp
#include <unistd.h>
#include <iostream>
int main() {
  char node_name[50];
  if (gethostname(node_name, sizeof(node_name)) == 0) {
    std::cout << "ENV VAR : " << node_name << std::endl;`

Tested by Marie with local sentry instance:
Screen Shot 2021-09-28 at 6 05 01 PM

Screen Shot 2021-09-28 at 6 04 51 PM

## Additional Information

- [ ] This change is backwards-breaking - NO

### DIFF
--- a/orc8r/gateway/c/common/sentry/SentryWrapper.cpp
+++ b/orc8r/gateway/c/common/sentry/SentryWrapper.cpp
@@ -20,6 +20,7 @@
 #include <cstdlib>
 #include <fstream>
 #include <cstring>
+#include <unistd.h>
 
 #include "sentry.h"
 #include "includes/ServiceConfigLoader.h"
@@ -84,6 +85,7 @@ std::string get_snowflake() {
 }
 
 void initialize_sentry(
+    char node_name[50]
     const char* service_tag, const sentry_config_t* sentry_config) {
   auto control_proxy_config = magma::ServiceConfigLoader{}.load_service_config(
       CONTROL_PROXY_SERVICE_NAME);
@@ -105,7 +107,9 @@ void initialize_sentry(
     }
 
     sentry_init(options);
-
+    if (gethostname(node_name, sizeof(node_name)) == 0) {
+      sentry_set_tag(HOSTNAME, node_name);
+    }
     sentry_set_tag(SERVICE_NAME, service_tag);
     sentry_set_tag(HWID, get_snowflake().c_str());
   }
--- a/orc8r/gateway/c/common/sentry/SentryWrapper.cpp
+++ b/orc8r/gateway/c/common/sentry/SentryWrapper.cpp
@@ -107,7 +107,7 @@ void initialize_sentry(
     }
 
     sentry_init(options);
-    char node_name[HOST_NAME_MAX]
+    char node_name[HOST_NAME_MAX];
     if (gethostname(node_name, HOST_NAME_MAX) == 0) {
       sentry_set_tag(HOSTNAME, node_name);
     }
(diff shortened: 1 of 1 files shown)


orc8r/gateway/c/common/sentry/SentryWrapper.cpp
```diff
@@ -84,6 +85,7 @@ std::string get_snowflake() {
 }
 
 void initialize_sentry(
+    char node_name[50]
```
themarwhal: where do we get this node_name from?

themarwhal: oh i see, are you just trying to make space? in that case, can we move this down to line ~110 right before we call gethostname?

ardzoht: For portability we can use `HOST_NAME_MAX` constant for specifying the node_name length

themarwhal:
Also for the semantic PR check failure you can fix the PR title to be: `feat: add hostname tag to sentry events` and also amend your commit message to be the same.
```
git commit --amend --signoff --allow-empty
# change name
git push -f origin <branch_name>
```

themarwhal:
As for the testplan, do you have a screen shot of the reported sentry event? Would be nice to have in the test plan, but if you don't have it, no worries.

orc8r/gateway/c/common/sentry/SentryWrapper.cpp
milapjoshi: yes, that would be clear.

orc8r/gateway/c/common/sentry/SentryWrapper.cpp
```diff
@@ -105,7 +107,9 @@ void initialize_sentry(
     }
 
     sentry_init(options);
-
+    if (gethostname(node_name, sizeof(node_name)) == 0) {
+      sentry_set_tag(HOSTNAME, node_name);
+    }
```
themarwhal: does this work?
```suggestion
    char node_name[HOST_NAME_MAX];
    if (gethostname(node_name, HOST_NAME_MAX) == 0) {
      sentry_set_tag(HOSTNAME, node_name);
    }
```

orc8r/gateway/c/common/sentry/SentryWrapper.cpp
milapjoshi: yep make sense. thanks !!!

orc8r/gateway/c/common/sentry/SentryWrapper.cpp
milapjoshi: sure it needs limits.h file let me add and push

milapjoshi:
> As for the testplan, do you have a screen shot of the reported sentry event? Would be nice to have in the test plan, but if you don't have it, no worries.

@themarwhal how do i add that? i do not have full dev env setup to trigger dummy sentry event.

orc8r/gateway/c/common/sentry/SentryWrapper.cpp
```diff
@@ -105,7 +107,10 @@ void initialize_sentry(
     }
 
     sentry_init(options);
-
+    char node_name[HOST_NAME_MAX]
```
themarwhal: missing a semicolon here

themarwhal (approved):
awesome work!! this will make the events a lot easier to work with. Thanks milap!

PR #9002, opened 2021-09-08
fix(mme): Fixing SEGV on implicit detach of unknown GUTI attach

ardzoht:
Signed-off-by: Alex Rodriguez <ardzoht@gmail.com>

## Summary

- Closes #8987 
- As part of #7916, part of implicit detach on unknown GUTI attach a double free segfault can be raised, which can be caused due to performing implicit detach (which will later free the MME ue context), and continue with the procedure, which later will try to access the Identification common procedure:
```
      emm_sap.primitive                      = EMMREG_COMMON_PROC_CNF;
      emm_sap.u.emm_reg.ue_id                = ue_id;
      emm_sap.u.emm_reg.ctx                  = emm_ctx;
      emm_sap.u.emm_reg.notify               = notify;
      emm_sap.u.emm_reg.free_proc            = true;
      emm_sap.u.emm_reg.u.common.common_proc = &ident_proc->emm_com_proc;
```

## Test Plan

- Running `make integ_test` and `make precommit` as sanity testing

## Additional Information

- [ ] This change is backwards-breaking

PR #9331, opened 2021-09-28
fix(mme): Fixing guti unknown attach and adding s1ap test case

ardzoht:

## Summary

- Fixes regression on #9002 (which fixed a sefgault that breaks `test_guti_attach_with_zero_mtmsi` TC on s1ap testing
- Adds `test_guti_attach_with_zero_mtmsi` to EXTENDED_TESTS tests list 

## Test Plan

- make integ_test
- verified `test_guti_attach_with_zero_mtmsi` passes successfully

## Additional Information


ulaskozat (approved):
LGTM

PR #9336, opened 2021-09-28
docs: Cloudstrapper Installation Instructions

jaredmullane:

## Summary

This provides instructions for how to deploy Magma via AWS Cloudstrapper. 

## Test Plan

create_docusaurus_website.sh
Screen Shot 2021-09-28 at 3 17 07 PM

### DIFF
--- a/docs/docusaurus/i18n/en.json
+++ b/docs/docusaurus/i18n/en.json
@@ -5,6 +5,9 @@
     "previous": "Previous",
     "tagline": "Bring more people online by enabling operators with open, flexible, and extensible network solutions",
     "docs": {
+      "basics/aws_cloudstrapper": {
+        "title": "AWS Cloudstrapper Install"
+      },
       "basics/introduction": {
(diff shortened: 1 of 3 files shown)


hcgatewood (changes requested):
Thanks for adding this Jared! I'll let @arunuke review for the cloudstrapper content

My docs input is please move this from "basics" to "deploy > orc8r" section+subsection, as that's a better fit

jaredmullane:
Thanks as always @hcgatewood ! Cloudstrapper contains installs both AGW and the Orc8r with minimum setup/commands so that is why I was thinking of putting it in the "Getting Started" section. Open to discuss more about breaking it up.

hcgatewood:
Ah true. Then deploy > general probably makes the most sense

jaredmullane:
@hcgatewood @arunuke created a new PR for this here with it under "Deploy > General": https://github.com/magma/magma/pull/9389

PR #9145, opened 2021-09-16
fix(mme): Updating S1AP UE state IMSI write on connection_establishment_cnf

ardzoht:
Signed-off-by: Alex Rodriguez <ardzoht@gmail.com>

## Summary

- Closes #8657 
- There was an issue on UE idle cases, where S1AP UE state was not recovered / written to redis after restart. This was due to a missing update on IMSI map after MME sent connection_establishment_cnf message to S1AP, after service request was processed. 

## Test Plan

- make integ_test for sanity
- Ran `test_idle_mode_with_mme_restart.py` and ensure there's entry on redis for S1AP UE state 

## Additional Information

- [ ] This change is backwards-breaking

PR #9338, opened 2021-09-28
fix(mme): Updating S1AP UE state IMSI write on connection_establishment_cnf

ardzoht:

## Summary

- Backport changes to v1.6 branch from #9145

## Test Plan

- validated with s1ap non sanity tests

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/s1ap/s1ap_mme.c
+++ b/lte/gateway/c/core/oai/tasks/s1ap/s1ap_mme.c
@@ -219,10 +219,11 @@ static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
     } break;
 
     case MME_APP_CONNECTION_ESTABLISHMENT_CNF: {
-      is_task_state_same = true;  // the following handler does not modify state
-      is_ue_state_same   = true;
+      is_task_state_same = false;
+      is_ue_state_same = false;
       s1ap_handle_conn_est_cnf(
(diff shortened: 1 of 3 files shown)


ulaskozat (approved):
LGTM

lte/gateway/c/core/oai/tasks/s1ap/s1ap_mme.c
```diff
@@ -219,10 +219,12 @@ static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
     } break;
 
     case MME_APP_CONNECTION_ESTABLISHMENT_CNF: {
-      is_task_state_same = true;  // the following handler does not modify state
-      is_ue_state_same   = true;
+      is_task_state_same =
+          false;  // the following handler does not modify state
```
ulaskozat: is the comment still valid here?

ardzoht: yes, it's still valid on v1.6

ulaskozat: hmm, the comment did not look correct on master either. if the state does not remain the same, then the handler should be modifying the state, no?

ardzoht: you're right, the comment was not updated on master as well, i will update it here for v1.6 and on master on a separate PR

ulaskozat (approved):
LGTM

PR #9342, opened 2021-09-29
chore(mme): Remove flakiness in mme_app unit tests

ulaskozat:
## Summary

- In unit testing for mme procedures, there is no explicit observation point in steady states about whether mme_app has completed state updates or not. This PR repurposes the existing health check mechanism to trigger such explicit observation points at mocked service303 task, that in turn is used to verify the expected mme_app_task states.

- Enabled stateless config and sandwiched critical parts with `if` directive to prevent an actual connection and I/O to the redis server.

- Some minor reordering of statements is also done to improve clang formatting output. 

## Test Plan

Run `make test_oai` and check execution time of unit tests. Repeat it 500 times and observe no flakes.

Run integ tests.
 
## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/mme_app/mme_app_state_manager.cpp
+++ b/lte/gateway/c/core/oai/tasks/mme_app/mme_app_state_manager.cpp
@@ -69,8 +69,12 @@ int MmeNasStateManager::initialize_state(const mme_config_t* mme_config_p) {
 
   // Allocate the local mme state
   create_state();
-
+#if !MME_UNIT_TEST
+  OAILOG_DEBUG(LOG_MME_APP, "MME_UNIT_TEST Flag is Disabled");
   redis_client = std::make_unique<RedisClient>(persist_state_enabled);
+#else
+  redis_client = std::make_unique<RedisClient>(false);
+#endif
   int rc       = read_state_from_db();
   read_ue_state_from_db();
   create_mme_ueip_imsi_map();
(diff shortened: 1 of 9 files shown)


lte/gateway/c/core/oai/tasks/mme_app/mme_app_state_manager.cpp
```diff
@@ -69,8 +69,12 @@ int MmeNasStateManager::initialize_state(const mme_config_t* mme_config_p) {
 
   // Allocate the local mme state
   create_state();
-
+#if !MME_UNIT_TEST
```
ardzoht, line 72: Wouldn't be better for abstraction purposes to change instead of passing entire `mme_config_p` on `initialize_state()` to just pass 2 args, `persist_state_enabled` and `max_ues`, this will translate into a simpler dependency injection from unit testing point of view? This way we can also avoid using the #if directives to control this.

ardzoht (changes requested):
Left comment about mme app state initialization

lte/gateway/c/core/oai/tasks/mme_app/mme_app_state_manager.cpp
ulaskozat: well, I do not see how this would prevent using `#if` directives. Note that the idea is to have persist_state_enabled as true and `#if` directive here is needed for preventing a connection set up. Also, note that we are passing the pointer to the configs, which I think is better than passing each used config separately.

ardzoht (approved):
lgtm

PR #9346, opened 2021-09-29
ci: adds packer build of magma_dev

electronjoe:
As json based configurations are now unsupported, converted
magma-focal-virtualbox.json by use of:

```
curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo apt-key add -
sudo apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main"
sudo apt-get update && sudo apt-get install packer
packer hcl2_upgrade orc8r/tools/packer/magma-focal-virtualbox.json
```

- Removed `../../../lte/gateway/deploy/roles/focal_hacks` from recipe as it seems to be gone now


quentinDERORY:
Stalking your draft PRs ^^'
But this will not work in github actions as this needs nested virtualization.
And GHA does not support it yet.
FYI: I am currently fixing the focal vm build #9366

EDIT: Actually  we can run VMs in macos so could work with that!

electronjoe:
> EDIT: Actually we can run VMs in macos so could work with that!

Thanks! Yeah I think that might be what I do with this PR when I get back to it? I didn't know that was possible but since you paved the way.. =D

PR #9349, opened 2021-09-29
chore: clean up folly lib def for bazel builds

themarwhal:

## Summary
Got inspired by  https://github.com/magma/magma/pull/9224 and tried cleaning up the folly def

## Additional Information

### DIFF
--- a/third_party/system_libraries.BUILD
+++ b/third_party/system_libraries.BUILD
@@ -21,7 +21,7 @@ config_setting(
 cc_library(
     name = "folly",
     srcs = select({
-        ":use_folly_so": ["usr/local/lib/libfolly.so"],
+        ":use_folly_so": [],
         "//conditions:default": [
             "usr/local/lib/libfolly.a",
             "usr/local/lib/libfmt.a",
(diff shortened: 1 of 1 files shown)


third_party/system_libraries.BUILD
```diff
@@ -21,14 +21,16 @@ config_setting(
...
-        ":use_folly_so": ["usr/local/lib/libfolly.so"],
+        ":use_folly_so": [],
         "//conditions:default": [
             "usr/local/lib/libfolly.a",
             "usr/local/lib/libfmt.a",
         ],
     }),
     linkopts = select({
         ":use_folly_so": [
+            "-L/usr/local/lib",
```
electronjoe, line 32: Can you explain why this was necessary? Or was it trial and error? I would have thought my `ldconfig` would fix this.. maybe.

electronjoe (approved):

third_party/system_libraries.BUILD
themarwhal: Trial and error. I ran ldconfig but that didn't seem to help

electronjoe: Can you show me the error you get when you lack -L/usr/locallib?

themarwhal: so just checked that the devcontainer works without "-L/usr..."
in the Magma VM, let me try agian..

themarwhal: scratched that this is the error even on the devcontainer: `/usr/bin/ld.gold: error: cannot find -lfolly`

PR #9362, opened 2021-09-29
fix(agw): lte-test fix 

tmdzk:

## Summary

fix(agw): lte-test fix

## Test Plan

lte-test using CI


quentinDERORY (approved):

quentinDERORY:
Would need to do the same changes for the github action workflow

tmdzk:
> Would need to do the same changes for the github action workflow

yes sir

PR #9363, opened 2021-09-29
chore(agw): enable default TSO config

pshelar:
The underlying perf issue is fixes, so its safe
to keep default TSO setting.

## Test Plan
validated on magma-dev vm.

## Additional Information

### DIFF
--- a/lte/gateway/configs/pipelined.yml
+++ b/lte/gateway/configs/pipelined.yml
@@ -209,5 +209,8 @@ sgi_tunnel:
    peer_pub_key: 'VQT+tLY6/xF+k1WqrXeQzlfb8hWMVLcPdtCPvwIUNU0='
    peer_pub_ip: 1.2.3.4
 
-dp_tso:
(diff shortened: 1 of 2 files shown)


koolzz (approved):

crbertoldo:
@pshelar, please: may use this config instead of `ethtool -K gtpu_sys_2152 tx off`?

pshelar:
> @pshelar, please: may use this config instead of `ethtool -K gtpu_sys_2152 tx off`?

This perf issue is seen only on specific setup, so we dont need to turn off tx by default.
I will investigate the issue this week.

crbertoldo:
@alledpaiva @obensouda, FYI

crbertoldo:
> > @pshelar, please: may use this config instead of `ethtool -K gtpu_sys_2152 tx off`?
> 
> This perf issue is seen only on specific setup, so we dont need to turn off tx by default. I will investigate the issue this week.

@alledpaiva @obensouda, FYI

PR #9365, opened 2021-09-29
fix(agw) Check for unset params while extracting status (#9288)

amarpad:
## Summary

fix(agw): Check for unset params while extracting status
When the eNB just comes up some of the eNB status nodes might not
be populated, check before dereferencing attribute.

fix(agw): Use get insteard of operator []

Address review comments, get is safer than [] and
the code is cleaner as explicitly set to None values are
handled identifical to the key itself being missing.

## Test Plan
unit tests

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/enodebd/devices/freedomfi_one.py
+++ b/lte/gateway/python/magma/enodebd/devices/freedomfi_one.py
@@ -233,7 +233,8 @@ class StatusParameters(object):
         success_str = "SUCCESS"  # String constant returned by radio
         insync_str = "INSYNC"
 
-        if name_to_val[cls.DEFAULT_GW].upper() != success_str:
+        if name_to_val.get(cls.DEFAULT_GW) \
+                and name_to_val[cls.DEFAULT_GW].upper() != success_str:
             # Nothing will proceed if the eNB doesn't have an IP on the WAN
             serial_num = "unknown"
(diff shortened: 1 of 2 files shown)


lte/gateway/python/magma/enodebd/devices/freedomfi_one.py
```diff
@@ -233,7 +233,8 @@ def set_magma_device_cfg(
         success_str = "SUCCESS"  # String constant returned by radio
         insync_str = "INSYNC"
 
-        if name_to_val[cls.DEFAULT_GW].upper() != success_str:
+        if name_to_val.get(cls.DEFAULT_GW) \
+                and name_to_val[cls.DEFAULT_GW].upper() != success_str:
```
ardzoht, line 237:
```suggestion
        if (name_to_val.get(cls.DEFAULT_GW) and 
             name_to_val[cls.DEFAULT_GW].upper()) != success_str:
```
wrap this (and the other usages) in parentheses to get rid of the linting warnings

ardzoht (changes requested):

ardzoht (approved):
lgtm

lte/gateway/python/magma/enodebd/devices/freedomfi_one.py
amarpad: As discussed I will fix this on master.

PR #9394, opened 2021-09-30
fix(agw): remove fluentbit registry using ansible(hotfix)

tmdzk:

fix(agw): remove fluentbit registry using ansible(hotfix)

## Summary

fix(agw): remove fluentbit registry using ansible(hotfix)

## Test Plan

lte-integ-test should be happy (tested locally)

### DIFF
--- a/orc8r/tools/ansible/roles/pkgrepo/tasks/main.yml
+++ b/orc8r/tools/ansible/roles/pkgrepo/tasks/main.yml
@@ -14,6 +14,10 @@
   fail: msg="distribution is undefined for the pkgrepo role"
   when: distribution is undefined
 
+- name: remove fluentbit repo as its broken
+  shell: rm -rf /etc/apt/sources.list.d/packages_fluentbit_io_ubuntu_focal.list
+  become: yes
+
 - name: Set up debian style repository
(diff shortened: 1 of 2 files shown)


orc8r/tools/ansible/roles/pkgrepo/tasks/main.yml
```diff
@@ -14,6 +14,10 @@
   fail: msg="distribution is undefined for the pkgrepo role"
   when: distribution is undefined
 
+- name: remove fluentbit repo as its broken
+  shell: rm -rf /etc/apt/sources.list.d/packages_fluentbit_io_ubuntu_focal.list
+  become: yes
```
quentinDERORY, line 19: can you add ignore_error, when we will rebuild the VM, it will fail

quentinDERORY (changes requested):
see comment

orc8r/tools/ansible/roles/pkgrepo/tasks/main.yml
tmdzk: yup

quentinDERORY (approved):

PR #9395, opened 2021-10-01
chore(mme): Adding unit test for ULA failure

ulaskozat:
## Summary

PR adds unit test (`TestImsiAttachEpsOnlyUlaFailure`) for failed Update Location Request towards S6a task. 

## Test Plan

Run the unit test suit multiple times and observe that no flakes exist.

## Additional Information

While running this test, it is discovered that Attach Reject and Context Release Requests are being sent twice rather than once. After this issue is fixed, the unit test need to be updated to the correct values. I.e., instead of:
```
  EXPECT_CALL(*s1ap_handler, s1ap_generate_downlink_nas_transport()).Times(4);
  ...
  EXPECT_CALL(*s1ap_handler, s1ap_handle_ue_context_release_command()).Times(2);
  ...
```

we expect after the fix:

```
  EXPECT_CALL(*s1ap_handler, s1ap_generate_downlink_nas_transport()).Times(3);
  ...
  EXPECT_CALL(*s1ap_handler, s1ap_handle_ue_context_release_command()).Times(1);
  ...

<!--
    If this is a backwards-breaking change, document the upgrade instructions.
    All upgrade instructions for backwards-breaking changes will be aggregated
    in the next release's changelog so this is very important to fill out.
-->

### DIFF
--- a/lte/gateway/c/core/oai/test/mme_app_task/mme_app_test_util.cpp
+++ b/lte/gateway/c/core/oai/test/mme_app_task/mme_app_test_util.cpp
@@ -110,34 +110,38 @@ void send_authentication_info_resp(const std::string& imsi, bool success) {
   return;
 }
 
-void send_s6a_ula(const std::string& imsi) {
+void send_s6a_ula(const std::string& imsi, bool success) {
   MessageDef* message_p =
       itti_alloc_new_message(TASK_S6A, S6A_UPDATE_LOCATION_ANS);
   s6a_update_location_ans_t* itti_msg =
       &message_p->ittiMsg.s6a_update_location_ans;
   strncpy(itti_msg->imsi, imsi.c_str(), imsi.size());
-  itti_msg->imsi_length        = imsi.size();
-  itti_msg->result.present     = S6A_RESULT_BASE;
-  itti_msg->result.choice.base = DIAMETER_SUCCESS;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: lte/gateway/c/core/oai/test/mme_app_task/mme_app_test_util.cpp [hunk 1]
```diff
@@ -110,34 +110,38 @@ void send_authentication_info_resp(const std::string& imsi, bool success) {
...
-  apn_ambr->set_max_bandwidth_ul(10000000);
-  apn_ambr->set_max_bandwidth_dl(75000000);
-  apnconfig.set_pdn(magma::feg::UpdateLocationAnswer::APNConfiguration::IPV4);
-  auto apns = ula.mutable_apn();
-  apns->Add()->CopyFrom(apnconfig);
-  magma::convert_proto_msg_to_itti_s6a_update_location_ans(ula, itti_msg);
+  itti_msg->imsi_length    = imsi.size();
+  itti_msg->result.present = S6A_RESULT_BASE;
+  if (success) {
+    itti_msg->result.choice.base = DIAMETER_SUCCESS;
```
ardzoht, line 122: minor nit: let's separate these between protobuf objects setup, for readability

ardzoht (approved):
lgtm

PR #9088, opened 2021-09-14
feat(agw): add sctp up/downstream sock to mme service config

electronjoe:
## Summary

Closes #8828.

Informed by https://github.com/magma/magma/commit/7c54225801bcd75a8d4ab6c9bf4190100745372c

### Changes

- Removed unused sctp configurations from MME service config
  - in_streams (SCTP_INSTREAMS in config)
  - out_streams (SCTP_OUTSTREAMS in config)
- Adds sctp configurations for MME service config
  - SCTP_UPSTREAM_SOCK (default unix:///tmp/sctpd_upstream.sock)
  - SCTP_DOWNSTREAM_SOCK (default unix:///tmp/sctpd_downstream.sock)

## Test Plan

- Ran `lte integration test` to ensure no regression in nominal parameter
- Added unit test to verify configuration parsing of new fields
- In magma VM, ensured config change results in new unix domain socket path
  - vagrant ssh magma
  - cd `/magma/lte/gateway`
  - vim `/vagrant/configs/templates/mme.conf.template`
    - update `SCTP_UPSTREAM_SOCK = unix:///tmp/sctpd_upstream_test.sock`
    - update `SCTP_DOWNSTREAM_SOCK = unix:///tmp/sctpd_downstream_test.sock`
  - make run
  - vim `/var/opt/magma/tmp/mme.conf` (validate config)
    - validate that the template was filled correctly (should be no-op for new parameters)
  - ls -lh `/tmp/sctpd_*.sock`
    - should contain
      - sctpd_downstream.sock (generated by sctpd)
      - /tmp/sctpd_upstream_2.sock (generated by mme with new config)

Signed-off-by: Scott Harrison Moeller <smoeller@fb.com>

PR #9400, opened 2021-10-01
ci: points agw-workflow to existing devcontainer image

electronjoe:
## Summary

Build of our dockerfile is causing flakes (actually at this very moment, `li agent test` is broken by upstream fetch of grpc as part of container build) and slowing the test down
(e.g. see [failure](https://github.com/magma/magma/pull/9088/checks?check_run_id=3766565267)).

This PR re-uses the devcontainer which supports all c/c++ tests in AGW.

Note that this change does not impact the compiler used - we still use `gcc` for these builds:

```shell
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
```

```shell
@electronjoe ➜ /workspaces/magma (pr-ci-agw-to-devcontainer) $ ls -lh /usr/bin/cc
lrwxrwxrwx 1 root root 20 Sep 21 09:11 /usr/bin/cc -> /etc/alternatives/cc
@electronjoe ➜ /workspaces/magma (pr-ci-agw-to-devcontainer) $ ls -lh /etc/alternatives/cc
lrwxrwxrwx 1 root root 12 Sep 21 09:11 /etc/alternatives/cc -> /usr/bin/gcc
@electronjoe ➜ /workspaces/magma (pr-ci-agw-to-devcontainer) $ ls -lh /usr/bin/c++
lrwxrwxrwx 1 root root 21 Sep 21 09:11 /usr/bin/c++ -> /etc/alternatives/c++
@electronjoe ➜ /workspaces/magma (pr-ci-agw-to-devcontainer) $ ls -lh /etc/alternatives/c++
lrwxrwxrwx 1 root root 12 Sep 21 09:11 /etc/alternatives/c++ -> /usr/bin/g++
```

## Pitfalls

This PR does couple our devcontainer environment with our CI test environment, which eventually we may want to break. But changing what container is executed from the registry is a simple change relative to the delta of this PR.  Tying these means that moving the container forward (see ##Maintenance below) will require additional testing in the container (e.g. running these workflows).

## Maintenance

Periodically we will move the container image forward after testing it's health, by updating the container tag / hash:

```yaml
env:
  DEVCONTAINER_IMAGE: "ghcr.io/magma/devcontainer:sha-375f28a"
```

## Misc

While we are here, add a 10 minute timeout to `session manager tests` which in one of my pushes ran for 44 minutes without failure.

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -11,6 +11,9 @@ on:  # yamllint disable-line rule:truthy
       - 'v1.*'
     types: [opened, reopened, synchronize]
 
+env:
+  DEVCONTAINER_IMAGE: "ghcr.io/magma/devcontainer:sha-375f28a"
+
 jobs:
   path_filter:
     runs-on: ubuntu-latest
@@ -123,6 +126,7 @@ jobs:
         run: |
             docker run --volume ${{ env.MAGMA_ROOT }}:/magma --volume ${{ env.MAGMA_ROOT }}/lte/gateway/configs:/etc/magma --interactive magma/c_cpp_build:latest /bin/bash -c "cd /magma/lte/gateway;make test_common;"
       - name: Run session_manager tests
+        timeout-minutes: 10
         run: |
             docker run --volume ${{ env.MAGMA_ROOT }}:/magma --volume ${{ env.MAGMA_ROOT }}/lte/gateway/configs:/etc/magma --interactive magma/c_cpp_build:latest /bin/bash -c "cd /magma/lte/gateway;make test_session_manager;"
       - name: Extract commit title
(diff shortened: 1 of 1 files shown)


themarwhal (approved):

quentinDERORY (approved):
The test for the dev container are automated right?
If yes we could export the hash of container in the gha job and grab this hash in the job using this dev container.
We can achieve that using github artifacts, we will export it like here https://github.com/magma/magma/blob/e7622b4360483c65b574c4c8681865408c72b606/.github/workflows/dco-check.yml#L22 . The extracting part would require playing with the github API to get the latest run of the testing workflow but should be feasible.

electronjoe:
@quentinDERORY had auto-merge on and just saw your comment.

We should talk! I think we want to move to automation in the hash updates on these - but we only want to move forward on hashes that have been tested against the various workflows that use the container. Not sure how to efficiently include those workflows as part of the docker-push workflow that generates these containers on a cron schedule? But I think that's the future goal.

PR #9401, opened 2021-10-01
fix(ci): packaging agw for ubuntu

quentinDERORY:

## Summary

Ubuntu need a different package file than debian.
This will fix the lte-agw-deploy job building and publishing agw packages.

## Test Plan

Tested in v1.5-ubuntu

## Additional Information

### DIFF
--- a/lte/gateway/fabfile.py
+++ b/lte/gateway/fabfile.py
@@ -135,6 +135,8 @@ def package(
 
         with cd('release'):
             mirrored_packages_file = 'mirrored_packages'
+            if os == "ubuntu":
+                mirrored_packages_file += '_focal'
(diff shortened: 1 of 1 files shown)


lte/gateway/fabfile.py
```diff
@@ -135,6 +135,8 @@ def package(
 
         with cd('release'):
             mirrored_packages_file = 'mirrored_packages'
+            if os == "ubuntu":
```
themarwhal, line 138: do we support any other os type?

quentinDERORY: the mirrored_packages file is for stretch

themarwhal (approved):

PR #9403, opened 2021-10-01
feat(agw): Add interface for crash reporting, implement interface aga…

mstre123:
…inst sentry-go.

## Summary

Interface for crash reporter for agw go
implement sentry-go

## Test Plan
test locally


src/go/reporter/reporter.go
```diff
@@ -0,0 +1,43 @@
...
+	Category  string                 `json:"category,omitempty"`
+	Message   string                 `json:"message,omitempty"`
+	Data      map[string]interface{} `json:"data,omitempty"`
+	Level     Level                  `json:"level,omitempty"`
+	Timestamp time.Time              `json:"timestamp"`
+}
+
+type Reporter interface {
+	AddBreadcrumb(breadcrumb *Breadcrumb)
+	CaptureException(exception error) *EventID
```
markjen: the return here can probably be a non-pointer/value

src/go/reporter/reporter.go
```diff
@@ -0,0 +1,43 @@
...
+// An EventID must be 32 characters long, lowercase and not have any dashes.
+type EventID string
+
+// Level marks the severity of the event.
+type Level string
+
+// Breadcrumb specifies an application event that occurred before a reported event.
+// An event may contain one or more breadcrumbs.
+type Breadcrumb struct {
+	Type      string                 `json:"type,omitempty"`
```
markjen: do we need these json annotations?

src/go/reporter/reporter.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"time"
+)
+
+// EventID is a hexadecimal string representing a unique uuid4 for an Event.
+// An EventID must be 32 characters long, lowercase and not have any dashes.
+type EventID string
+
+// Level marks the severity of the event.
+type Level string
```
markjen: let's grab the enum values from sentry's SDK:

```
const (
	LevelDebug   Level = "debug"
	LevelInfo    Level = "info"
	LevelWarning Level = "warning"
	LevelError   Level = "error"
	LevelFatal   Level = "fatal"
)
```

don't want random ones showing up nor do we want people putting strings in everywhere

src/go/reporter/reporter.go
```diff
@@ -0,0 +1,43 @@
...
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package reporter
```
markjen: nit: maybe we can name this package `crash` or something else; `reporter` is somewhat generic and means a lot of different things around magma

src/go/go.mod
```diff
@@ -25,6 +25,7 @@ require (
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/getsentry/sentry-go v0.11.0 // indirect
```
markjen: my `go mod tidy` run moves this into the direct block (we do reference it directly in `src/go/reporter/sentry/sentry.go`)

src/go/reporter/reporter.go
```diff
@@ -0,0 +1,43 @@
...
+type Breadcrumb struct {
+	Type      string                 `json:"type,omitempty"`
+	Category  string                 `json:"category,omitempty"`
+	Message   string                 `json:"message,omitempty"`
+	Data      map[string]interface{} `json:"data,omitempty"`
+	Level     Level                  `json:"level,omitempty"`
+	Timestamp time.Time              `json:"timestamp"`
+}
+
+type Reporter interface {
```
markjen: I wonder if we really need CaptureMessage and CaptureException

Btw, what do pkg/errors w/ stack look like when we send to sentry? if they get processed well, maybe we only have `Capture(err error)`?

markjen (changes requested):

src/go/go.mod
mstre123: ah that's on me - to implement sentry i go getted it will update

src/go/reporter/reporter.go
mstre123: yeah will rename

src/go/reporter/reporter.go
mstre123: looks like you added something similar to log.go will follow that pattern

src/go/reporter/reporter.go
mstre123: it was a copy and paste from sentry - can remove

src/go/reporter/reporter.go
mstre123: looks like it's really just for typing. https://github.com/getsentry/sentry-go/blob/8fbb027528f9464e0233afa756cef0f99bd70906/client.go#L323-L341
can probably get away with just a capture

src/go/crash/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
```
markjen, line 1: nit: copyright header

src/go/crash/crash.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// EventID is a hexadecimal string representing a unique uuid4 for an Event.
+// An EventID must be 32 characters long, lowercase and not have any dashes.
+type EventID string
+
+// Level marks the severity of the event.
+type Level string
+
+// Describes the severity of the event.
```
markjen, line 26: nit: can we leave a comment here to let people know these were based on the sentry.io values?

src/go/crash/crash.go
```diff
@@ -0,0 +1,53 @@
...
+	Type      string
+	Category  string
+	Message   string
+	Data      map[string]interface{}
+	Level     Level
+	Timestamp time.Time
+}
+
+// Crash provides an interface for interacting with crash reporters.
+type Crash interface {
```
markjen, line 49: could be `crash.Reporter`; this is also fine though up to you :)

src/go/crash/sentry/sentry.go
```diff
@@ -0,0 +1,84 @@
...
+func NewCrash(options sentry.ClientOptions) crash.Crash {
+	client, err := sentry.NewClient(options)
+	if err != nil {
+		panic(errors.Wrapf(err, "sentry.ClientOptions=%+v", options))
+	}
+	newHub := sentry.NewHub(client, sentry.NewScope())
+	newHub.BindClient(client)
+
+	return &Crash{
+		newHub,
```
markjen: nit: explicit field names

src/go/crash/sentry/sentry_test.go
```diff
@@ -0,0 +1,141 @@
...
+	assert.NotNil(t, crash.hub)
+	assert.Equal(t, options.Dsn, crash.hub.Client().Options().Dsn)
+}
+
+func TestNewCrash_Error(t *testing.T) {
+	options := sentry.ClientOptions{
+		Dsn: "\"bad_input",
+	}
+
+	// we expect error from sentry.NewCrash() to panic
```
markjen: there is an `assert.Panic` helper if you prefer

src/go/crash/sentry/sentry_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	c.Capture(exception)
+}
+
+func TestCrash_Capture_Error(t *testing.T) {
+	exception := errors.New("error")
+
+	options := sentry.ClientOptions{}
+	c := NewCrash(options)
+	crash, _ := c.(*Crash)
```
markjen: nit: handle the `!ok` case and `t.Fatal`

src/go/crash/sentry/sentry_test.go
```diff
@@ -0,0 +1,141 @@
...
+	c.Capture(exception)
+}
+
+func TestCrash_AddBreadcrumb(t *testing.T) {
+	breadcrumb := crash.Breadcrumb{}
+
+	options := sentry.ClientOptions{}
+	r := NewCrash(options)
+
+	r.AddBreadcrumb(breadcrumb)
```
markjen: is there something to verify here?

src/go/crash/sentry/sentry_test.go
```diff
@@ -0,0 +1,141 @@
...
+	}()
+
+	c.AddBreadcrumb(breadcrumb)
+}
+
+func TestCrash_Recover(t *testing.T) {
+	options := sentry.ClientOptions{}
+	c := NewCrash(options)
+
+	c.Recover()
```
markjen: what is the actual use case supposed to look like? can we construct some tests around that?

src/go/crash/sentry/sentry.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"time"
+
+	"github.com/getsentry/sentry-go"
+	"github.com/magma/magma/src/go/crash"
+	"github.com/pkg/errors"
+)
+
+type Crash struct {
+	hub *sentry.Hub
```
markjen: for testing purposes, consider creating a sentry interface that sentry.Hub implements:

```
type sentryHub interface {
  AddBreadcrumb(breadcrumb *sentry.Breadcrumb, hint *sentry.BreadcrumbHint)
  Recover(err error)
  Flush(timeout time.Duration) bool 
}
```

then you can mock it out for testing and verify your adapter `type Crash struct` actually calls sentry.Hub with the expected args

markjen (commented):
also, let's see if we can get more code cov :)

src/go/crash/BUILD.bazel
mstre123: Looking through the codebase we don't append copyright headers to makefiles or bazel files.

fixed and setup a copyright header in intellij so this stops coming up on new files though.
Will add a line in the intellij setup on docusaurus to add the copyright profile to there settings - might be worth a linter check.

src/go/crash/sentry/sentry.go
mstre123: updated

src/go/crash/crash.go
mstre123: i'm kind of a fan of keeping the package name in line with the interface name.
id call it crash_report.CrashReporter but effective go is against underscores in package names (https://golang.org/doc/effective_go#package-names)
I know that effective go suggests adding an -er suffix (https://golang.org/doc/effective_go#interface-names) but crasher seems pretty bad and we ruled out reporter in an earlier comment on this diff.

crash.Crash might be the least worst option

src/go/crash/sentry/sentry_test.go
mstre123: added

src/go/crash/sentry/sentry_test.go
mstre123: refactored

src/go/crash/sentry/sentry_test.go
mstre123: verified the underlying function is called

src/go/crash/sentry/sentry_test.go
mstre123: this one is a bit tricky - to test recover on a panic we need to generate a panic - that the function will be able to pull the error out of.

src/go/crash/sentry/sentry.go
mstre123: added

src/go/crash/sentry/sentry.go
```diff
@@ -0,0 +1,102 @@
...
+		sentryHub: newHub,
+	}
+}
+
+type SentryHub struct {
+	hub *sentry.Hub
+}
+
+// SentryHub defines a subset sentry.Hub functions that we use.
+type sentryHub interface {
```
markjen, line 25: nit: prefer this definiton closer to the //go:generate that uses it

src/go/crash/sentry/sentry.go
```diff
@@ -0,0 +1,102 @@
...
+		panic(errors.Wrapf(err, "sentry.ClientOptions=%+v", options))
+	}
+	newHub := sentry.NewHub(client, sentry.NewScope())
+	newHub.BindClient(client)
+	return &Crash{
+		sentryHub: newHub,
+	}
+}
+
+type SentryHub struct {
```
markjen: is this additional wrapper struct necessary?

```
var _ sentryHub = (*sentry.Hub)(nil)
```

should already be valid

src/go/crash/sentry/sentry.go
```diff
@@ -0,0 +1,102 @@
...
+
+// NewCrash returns a Crash with a singleton sentry Hub.
+// All calls on that Crash will be preformed on that Hub.
+func NewCrash(options sentry.ClientOptions) crash.Crash {
+	client, err := sentry.NewClient(options)
+	if err != nil {
+		panic(errors.Wrapf(err, "sentry.ClientOptions=%+v", options))
+	}
+	newHub := sentry.NewHub(client, sentry.NewScope())
+	newHub.BindClient(client)
```
markjen: is this additional `BindClient` necessary?

src/go/crash/sentry/sentry.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/getsentry/sentry-go"
+	"github.com/magma/magma/src/go/crash"
+	"github.com/pkg/errors"
+)
+
+//go:generate go run github.com/golang/mock/mockgen -source=sentry.go -package mock_sentry -destination ../mock_sentry/mock_sentry_hub.go sentryHub
+
+type Crash struct {
+	sentryHub sentryHub
```
markjen: nit: you don't need to specify the name if you intend for `type Crash struct` to effectively embed a sentryHub

src/go/crash/sentry/sentry_test.go
```diff
@@ -0,0 +1,135 @@
...
+	mockHub.EXPECT().AddBreadcrumb(&sentry.Breadcrumb{}, nil)
+
+	c := &Crash{
+		sentryHub: mockHub,
+	}
+
+	c.AddBreadcrumb(breadcrumb)
+}
+
+func TestCrash_Recover_NoError(t *testing.T) {
```
markjen, line 101: you should be able to test recover like this:

```
mockHub := mock_sentry.NewMocksentryHub(ctrl)
mockHub.EXPECT().Recover("hi").Return(mockedEventID)

var wg sync.WaitGroup
wg.Add(1)
go func() {
  defer wg.Done()
  defer func() {
    eventID := c.Recover()
    assert.Equal(t, mockedEventID, eventID)
  }
  panic("hi")
}()

wg.Wait()
```

src/go/crash/sentry/sentry.go
mstre123: can move this up but prefer generates at the top of a file - so you know whats being generated without having to scan the full file.

src/go/crash/sentry/sentry.go
mstre123: its duplicative was following the pattern of sentry init - but also created my own hub so not necessary

src/go/crash/sentry/sentry_test.go
mstre123: the function recover gets overloaded alot :)

the problem with this is crash.Recover() depends on builtin.recover() to pull out an error or it returns an empty event id
https://github.com/magma/magma/pull/9403/files/93fe4c1e36f8d09cab3295aead6ef5bd9aae8aa5#diff-2bf049a689e3807e460ca8e8649732931d9f8b19941e3d473d1ba366444bb2e4R81
what im trying to figure out is how we can mock builtin.recover()

src/go/crash/sentry/sentry.go
mstre123: the struct is used for aliasing the implementation functions of the sentryHub interface.

markjen: I think `type SentryHub struct` is actually not used at all.

src/go/crash/sentry/sentry_test.go
mstre123: resolving since we have a thread going on the actual test below

src/go/crash/crash_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+package crash
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+//go:generate go run github.com/golang/mock/mockgen -package crash -destination mock_crash_test.go . Crash
```
markjen: I think I found it was more useful to create a `crash/mock_crash` package than put it in a _test.go file in the main package. IIRC, tests in _test.go/+test builds cannot access _test.go files in other packages.

See `log/mock_log` for an example

src/go/crash/crash.go
```diff
@@ -0,0 +1,55 @@
+/*
```
markjen: double comment block?

src/go/crash/sentry/sentry.go
```diff
@@ -0,0 +1,101 @@
+/*
```
markjen: double comment block?

src/go/crash/sentry/sentry.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"time"
+
+	"github.com/getsentry/sentry-go"
+	"github.com/magma/magma/src/go/crash"
+	"github.com/pkg/errors"
+)
+
+//go:generate go run github.com/golang/mock/mockgen -source=sentry.go -package mock_sentry -destination ../mock_sentry/mock_sentry_hub.go sentryHub
```
markjen: this is used only by the sentry package, so probably goes in `crash/sentry/mock_sentry_hub` or something

src/go/crash/sentry/sentry.go
```diff
@@ -0,0 +1,101 @@
...
+		Level:     sentry.Level(breadcrumb.Level),
+		Timestamp: time.Time{},
+	}
+	c.sentryHub.AddBreadcrumb(sentryBreadCrumb, nil)
+}
+
+// Recover calls Sentry's Recover function.
+// Returns EventID if successfully, or empty Event ID if Sentry returns nil.
+func (c *Crash) Recover() crash.EventID {
+	if err := recover(); err != nil {
```
markjen: turns out this doesn't work.

if you look at the docs for the `recover()` built-in (https://pkg.go.dev/builtin#recover) it says that recover() only works in the defer func scope (and not other called funcs deeper in the stack).

src/go/crash/sentry/sentry_test.go
markjen: see the test case i wrote in https://github.com/markjen/magma/commit/044933095b987845a44da78d599c07b22268a66e

we do not want/need to mock `builtin.recover()`

markjen (changes requested):
please take a look at this commit I wrote on top of this branch

https://github.com/markjen/magma/commit/044933095b987845a44da78d599c07b22268a66e

it includes:

- fix for broken test `TestCrash_AddBreadcrumb`
- update to crash.Crash.Recover
- test case for Crash.Recover

src/go/crash/crash.go
mstre123: weird this was the intellij auto copyright - will check my settings.

src/go/crash/sentry/sentry.go
mstre123: weird this was the intellij auto copyright - will check my settings.

src/go/crash/sentry/sentry.go
mstre123: got it, was trying to replicate this but your right it wont work.
https://github.com/getsentry/sentry-go/blob/master/hub.go#L316
swapping to taking in the err.

src/go/crash/sentry/sentry_test.go
mstre123: yup once we take in the err this works.

src/go/crash/sentry/sentry.go
mstre123: moved it to crash/sentry/mocks/mock_sentry_hub

src/go/crash/sentry/sentry.go
mstre123: removed

src/go/crash/crash_test.go
mstre123: refactored

src/go/crash/crash_test.go
```diff
@@ -0,0 +1,55 @@
+/*
```
markjen: nit: it looks like most golang code uses `//` instead of `/* */` comment blocks (See source files in golang itself:  https://github.com/golang/go/tree/master/src)

src/go/crash/sentry/sentry.go
```diff
@@ -0,0 +1,87 @@
...
+// It returns false if the timeout was reached. In that case, some events may not have been sent.
+// Flush should be called before terminating the program to avoid unintentionally dropping events.
+// Do not call Flush indiscriminately after every call to Capture.
+func (c *Crash) Flush(timeout time.Duration) bool {
+	return c.sentryHub.Flush(timeout)
+}
+
+// convertSentryEventID is a helper function to convert EventIDs
+func convertSentryEventID(sEID *sentry.EventID) crash.EventID {
+	if sEID == nil {
```
markjen, line 80: pls add unit test for coverage

src/go/crash/sentry/sentry.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"time"
+
+	"github.com/getsentry/sentry-go"
+	"github.com/magma/magma/src/go/crash"
+	"github.com/pkg/errors"
+)
+
+//go:generate go run github.com/golang/mock/mockgen -source=sentry.go -package mock_sentry -destination mocks/mock_sentry_hub.go sentryHub
```
markjen: nit: do we usually match the dir name to the package name, in this case the pkg is `mock_sentry` so the path would be `src/go/crash/sentry/mock_sentry`?

markjen (approved):

src/go/crash/crash_test.go
mstre123: yup will play with the copyright profile auto formatter a bit more - once i get it will try and export it and stick it somewhere.

src/go/crash/sentry/sentry.go
mstre123: yeah the other way we can go is -package mocks, seems cleaner but might cause us headaches in the future.
will refactor the dir

mstre123:
@andreilee looks the nms-elint test isn't working. it has an issue loading a module.
This should not impact this pr since it doesn't touch the nms at all and no code was written in python but wanted to alert you to the failing test.

PR #9409, opened 2021-10-01
fix(agw): Fix default logging level

pshelar:
In scale test, services like state or directoryD can log huge
amount of data on certain events. That results in CPU spikes.
Following PR sets the default value to INFO to avoid unnecessary
logging.

## Additional Information

### DIFF
--- a/lte/gateway/configs/directoryd.yml
+++ b/lte/gateway/configs/directoryd.yml
@@ -13,3 +13,5 @@
 
 # log_level is set in mconfig. It can be overridden here
 print_grpc_payload: false
+
+log_level: INFO
--- a/lte/gateway/configs/state.yml
+++ b/lte/gateway/configs/state.yml
@@ -15,6 +15,7 @@
 # sync_interval is set in mconfig. it can be overridden here
 
 print_grpc_payload: false
+log_level: INFO
 
 #state_protos:
 #  - proto_file:  - file to load proto from


koolzz (approved):

uri200:
Hey @pshelar note the idea is log level is set through mconfig 
Note the comment on the yaml
```
# log_level is set in mconfig. It can be overridden here
```
See all the other services don't use a hardcoded log_level anymore. If you set it here, mconfig will never take effect anymore.

Default level should be set to INFO by the app itself when both log_level or mconfig is missing

uri200 (changes requested):
sorry, see comment above ^^^^^

pshelar:
> Hey @pshelar note the idea is log level is set through mconfig Note the comment on the yaml
> 
> ```
> # log_level is set in mconfig. It can be overridden here
> ```
> 
> See all the other services don't use a hardcoded log_level anymore. Default level should be set to INFO by the app itself when both log_level or mconfig is missing

I could not find any way to change the logging level in orc8r API. It looks like its at mconfig API only.
anyways, I think we can improve this, but I wanted to get fix specifically for 1.6.1 release, so dont want to complicate the PR.

uri200:
Here is how the python services get the logging value. If both are missing it gets INFO

https://github.com/magma/magma/blob/master/orc8r/gateway/python/magma/common/service.py#L288-L288

So if you don't have it on swagger nor on the .yml the log level should be INFO
Can you make sure log info is not set somehow on your mconfig (i doubt it is there)?

```
vim /var/opt/magma/configs/gateway.mconfig
```

uri200:
Thanks @pshelar for your comments during our call. As described by you, it looks like we may have an issue due to the enum value set on our protos for Debug level.

The change in this PR temporary fixes that. Thanks!

https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Auri200%2Fmagma.git&branch=feg_merge_credit_reports_into_single_crru&file=orc8r%2Fprotos%2Fcommon.proto&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=46&start_col=7&end_row=46&end_col=7

uri200 (approved):

pshelar:
create issue #9410 for logging level improvement.

PR #9411, opened 2021-10-01
refactor: Removing WiFi module

reginawang3495:
## Summary
- removed wifi module
- removed `wifi` from `nms`
- modified `docs`

## Test Plan

- removed wifi module
    - [x] Ran `build.py --all` (and `generate`) under `${MAGMA_ROOT}/orc8r/cloud/docker`
- removed `wifi` from `nms`
    - [x] Validated works by running commands as per https://docs.magmacore.org/docs/next/basics/quick_start_guide#using-the-nms-ui
- modified `docs`
    - [x] Viewed local changes as per https://docs.magmacore.org/docs/next/docs/docs_overview#view-local-changes
image

## Additional Information


lte/gateway/Vagrantfile
```diff
@@ -26,7 +26,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     # - linux kernel from debian jessie backports
     # - updated vbguest-tool
     config.vm.box = "magmacore/magma_dev"
-    config.vm.box_version = "1.1.20210618"
+    config.vm.box_version = "1.1.20210928"
```
uri200: why are we changing the box version ?

hcgatewood (commented):
First pass looks promising. Unclear on the purpose of the VM changes -- did you intentionally add those? If not, maybe need to update to latest master and it will remove that part of the diff

.run/go test all.run.xml
```diff
@@ -1,12 +1,11 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="go test all" type="Multirun" factoryName="Multirun" singleton="false" separateTabs="true" reuseTabsWithFailures="false" startOneByOne="true" markFailedProcess="true" hideSuccessProcess="true" delayTime="4.0">
```
reginawang3495, line 2: Should I revert unrelated changes like these?

docs/docusaurus/i18n/en.json
```diff
@@ -441,6 +441,9 @@
       "proposals/p020_sgi_tunnel_transport": {
         "title": "L3 transport for AGW"
       },
+      "proposals/p021_mme_migrate_to_c++": {
```
reginawang3495, line 444: ^^

feg/gateway/go.sum
```diff
@@ -97,6 +97,7 @@ github.com/dgrijalva/jwt-go v3.2.0+incompatible h1:7qlOGliEKZXTDg6OTjfoBKDXWrumC
 github.com/dgrijalva/jwt-go v3.2.0+incompatible/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
 github.com/dgryski/go-farm v0.0.0-20190423205320-6a90982ecee2/go.mod h1:SqUrOPUnsFjfmXRMNPybcSiG0BgUW2AuFH8PAnS2iTw=
 github.com/dgryski/go-sip13 v0.0.0-20181026042036-e10d5fee7954/go.mod h1:vAd38F8PWV+bWy6jNmig1y/TA+kYO4g3RSRF0IAv0no=
+github.com/dnephin/pflag v1.0.7/go.mod h1:uxE91IoWURlOiTUIA8Mq5ZZkAv3dPUfZNaT80Zm7OQE=
```
reginawang3495, line 100: I have made no `feg` changes other than `./build.py` under `orch8r`, I can revert this, or I have also run tests under `feg` which works (from https://docs.magmacore.org/docs/next/feg/dev_testing#test-feg)

.run/go test all.run.xml
hcgatewood: Generally yes. But in this case it's unclear whether this is a "correct" update from IntelliJ, so I'm fine with leaving it in and if it gets reverted in a subsequent PR then fine

nms/README.md
```diff
@@ -164,7 +164,7 @@ Run `yarn test --coverage`
 ## Updating NMS for API Changes
 
 ### How to re-generate MagmaAPIBindings.js
-Run `./build.py -g` in `magma/orc8r/cloud`
+Run `./build.py -g` in `magma/orc8r/cloud/docker`
```
hcgatewood: Nit: use long-form of flags to self-document

```suggestion
Run `./build.py --generate` in `magma/orc8r/cloud/docker`
```

hcgatewood (commented):
LGTM. During our sync let's overview 1-2 manual verifications, then I think we can merge

uri200:
Just curious but where does the wifi removal need comes from?

reginawang3495:
@uri200 IIRC, I think `wifi` is a misnomer for this module and it was previously used to internally "connect to SOMA / Devmand". 

This module is no longer being used anymore so it can be safely removed <to prevent code bloat/excise unnecessary code>

Someone else can probably validate but this is what I've picked up

uri200:
Thanks @reginawang3495 

Hey @hcgatewood can you elaborate on this need. We still have xwfm-test on our CI which, per my understanding, tests wifi path, so that may mean we still need/use wifi modules somewhere (?)

quentinDERORY (changes requested):
Overall LGTM, but failing some NMS tests specifying wifi: https://github.com/magma/magma/pull/9411/checks?check_run_id=3807844451

hcgatewood:
> Thanks @reginawang3495
> 
> Hey @hcgatewood can you elaborate on this need. We still have xwfm-test on our CI which, per my understanding, tests wifi path, so that may mean we still need/use wifi modules somewhere (?)

@uri200 To my understanding, xwf-m is unrelated to the wifi module. Instead, it looks to rely on the cwf module. As an additional input to this: here's where we set the "filepath dependencies" for the xwf-m job, which does not include the wifi module: https://github.com/magma/magma/pull/9411#issuecomment-934817763. Please correct me if I'm wrong!

As for why this change: this is a P3 backlog item for the benefits @reginawang3495 mentioned

Let me know if you have additional concerns and am happy to discuss

uri200:
Thanks @hcgatewood.  The concern came of removing a module on NMS/Orc8r but not removing its support on gateways. But @themarwhal provided some insight: it looks like SOMA had its own gateways so we do not have code on ours.

quentinDERORY (changes requested):
LGTM, but these tests (https://github.com/magma/magma/actions/runs/1315072029) are failing due to lack of support of files with spaces...

Could you remove the 'Fetch list of changed files' step, to make the test run & pass and then revert the change.

EDIT: we need to fix the issue in the workflow with path including spaces, disabling/enabling it will not permit to run the workflow

quentinDERORY (approved):

hcgatewood:
Lgtm. Once Andrei approves, msg me and I'll merge!

andreilee (approved):

hcgatewood (approved):
💯💯

Issue #63, opened 2019-03-13
mme crash with wrong apn 


sarathbrp:
MME crashes when UE is configured with wrong APN.

Attach -> Detach -> Attach -> Detach... in loop

until the MME crashes. 
[mme_crash_wrong_apn.zip](https://github.com/facebookincubator/magma/files/2963548/mme_crash_wrong_apn.zip)

UE is Essential PH1

anooptomar29:
We reproduced the issue here in our lab ( ESM Information response with incorrect APN). It does not crash  but it does not handle it gracefully in the sense it does not send attach reject followed by UE Context release command. After this when we tried attaching UE with correct APN , it worked fine. Attached are the logs and pcap.
[incorrectapn.zip](https://github.com/facebookincubator/magma/files/2980610/incorrectapn.zip)

ssanadhya:
Resolved with commit eb1df5dcd19eb9c862f3791af48a00051a0f1b2a: the MME sends Attach Reject when UE sends wrong APN in ESM info response.

PR #9412, opened 2021-10-01
feat: (vN) Initial commit for creation of metrics framework

### LINKED ISSUE #8843: [AGW][vN] src/go/metrics
- abstraction of metrics data model and collector for use across new golang code
  - potentially use models defined in orc8r/protos/metricsd.proto
  - define `type Collector interface` that will be injected into code, i.e. src/go/service/... implementations
  - helper funcs make it easy for code to create PushedMetric models into a Collector
  - goal is to make sure src/go/metrics does not import github.com/prometheus/...
- implementation that integrates with metricsd
  - maybe have a Collector implementation that periodically pushes to MetricsController.Push? (not sure on the details here yet)

andreilee:

## Summary

Added a new metrics framework, which should allow removal of dependencies on the exact time-series database that we are using. Prometheus-specific code should be moved away from the AGW, and later left only at the Orc8r level. To complete this migration, a new `MetricSender` is provided, which is expected to eventually have an implementation that sends metrics in a platform-agnostic way to Orc8r.

See `const.go` for a bit more details on usage.

## Test Plan

- Added a number of unit tests
- Tested new `go:generate` annotations for generation of protobuf and gomock files

## Additional Information


andreilee:
Hmm will look into the build failures - is there a guide for vN build?

src/go/coverage.txt
```diff
@@ -0,0 +1,122 @@
+mode: atomic
```
markjen: extraneous file I think? (usually most folks make this coverage.out and I think it's ignored)

src/go/metrics/sender.go
```diff
@@ -0,0 +1,99 @@
...
+// limitations under the License.
+
+package metrics
+
+import (
+	"context"
+
+	io_prometheus_client "github.com/prometheus/client_model/go"
+
+	"github.com/go-openapi/swag"
```
markjen: can we avoid the use of swag?

also, protobuf also includes pointer-gen methods like `proto.String`

src/go/go.mod
```diff
@@ -25,7 +28,9 @@ require (
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
-	github.com/golang/protobuf v1.5.0 // indirect
+	github.com/go-openapi/swag v0.19.15 // indirect
```
markjen: running `go mod tidy` changed this file somewhat; not sure why there's a mismatch

(also, see later comment about removing swag; which would also change go.mod/go.sum)

src/go/metrics/collector.go
```diff
@@ -0,0 +1,99 @@
...
+	stop   chan bool
+}
+
+type MetricsCollectionOptions struct {
+	Delay time.Duration
+	Clock clock.Clock
+}
+
+var collectorOnce sync.Once
+var collectorInstance *metricsCollector
```
markjen: no package globals pls; we can make the Collector a public type and return it

src/go/metrics/collector.go
```diff
@@ -0,0 +1,99 @@
...
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/benbjohnson/clock"
+)
+
+type metricsCollector struct {
+	sender MetricSender
+	stop   chan bool
```
markjen: do we ever stop a collector?

"clean shutdown" often adds complexity for not much gain. I'd recommend YAGNI here

src/go/metrics/collector.go
```diff
@@ -0,0 +1,99 @@
...
+var collectorInstance *metricsCollector
+
+func DefaultMetricsCollectionOpts() *MetricsCollectionOptions {
+	return &MetricsCollectionOptions{
+		Delay: time.Minute,
+		Clock: clock.New(),
+	}
+}
+
+func StartMetricsCollection(options *MetricsCollectionOptions) error {
```
markjen: I would turn this into:

`func NewCollector(c clock.Clock, period time.Duration, sender Sender) *Collector`

Collector would have two funcs:

```
func (c *Collector) RegisterGauge(Gauge) { ... }
func (c *Collector) Start() { ... }
```

No more `Registry` (another global)

Lastly, `Sender` becomes:

```
type Sender interface {
  Send(*metricsd.MetricsContainer) error
}
```

`Collector` is responsible for processing all `Gauge` samples every `period` and calling `Sender`

src/go/metrics/collector.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return nil
+}
+
+func StopMetricsCollection() {
+	collector := getMetricsCollector()
+	collector.stop <- true
+}
+
+func RegisterSender(sender MetricSender) error {
```
markjen: do you expect `Sender` to change? if so, move `Sender` out of `NewCollector` and make it a `SetSender` func

but it looks like no, given that you're erroring when someone tries to set the Sender when it already exists

src/go/metrics/gauge.go
```diff
@@ -0,0 +1,76 @@
...
+// NewGauge creates a new Gauge.
+// It is expected that values will be populated not more than once a minute.
+func NewGauge(name string, description string, labelNames []string) Gauge {
+	gauge := &gaugeImpl{name, description, labelNames, make(chan *GaugeSample, 10), clock.New()}
+	RegisterGauge(gauge)
+	return gauge
+}
+
+// Set sets the Gauge to an arbitrary value.
+func (g *gaugeImpl) Set(value float64, labels map[string]string) {
```
markjen, line 82: if `g.samples` is a `make(chan *GaugeSample, 10)`, it means after the 10th call to `Set` without draining the channel, the caller will block (almost definitely unintended)

we are probably better off with a straight mutex array here:

```
type gaugeImpl struct {
  ... other metadata ...
  samples []GaugeSample  // note, a plain struct is fine here; explicit pointer is not necessary since these are not nil-able and copies should be small/light
  sync.Mutex
}

func (g *gaugeImpl ) Set(value float64, labels map[string]string) {
  g.Lock()
  defer g.Unlock()

  // note: labels is a pointer to a map, so if the caller is sharing the labels map and changing it, the samples will all get the same copy. probably ok to note that in the documentation somewhere; making defensive copies would be safer, but incur a large perf penalty
  g.samples = append(s.samples, GaugeSample{...})
}

func (g *gaugeImpl) GetMetric() *io_prometheus_client.Metric {
  samples := g.samples
  g.Lock()
  g.samples = nil
  g.Unlock()

  // create *io_prometheus_client.Metric and return
}
```

---

if we really want to do channels that pipeline into a result, then we'd want to feed the data directly into a `*metricsd.MetricsContainer` via a goroutine inside `Collector`. Then every `period`, a mutex pauses the collector goroutine, swaps out the `*metricsd.MetricsContainer` for a new one, and sends the previous filled-up `*metricsd.MetricsContainer` to `Sender`

this is a much more advanced design and probably overkill; it also risks creating pauses in callers to `Gauage.Set` since it'll be governed by the collector goroutine.

src/go/metrics/sender.go
```diff
@@ -0,0 +1,99 @@
...
+	gatewayId        string
+	client           metricsd.MetricsControllerClient
+}
+
+func NewPrometheusMetricSender(client metricsd.MetricsControllerClient, gatewayId string) MetricSender {
+	// return &prometheusMetricSender{&metricsd.MetricsContainer{}}
+	// container := &metricsd.MetricsContainer{GatewayId: "asdf"}
+	container := &metricsd.MetricsContainer{
+		GatewayId: gatewayId,
+		Family:    []*io_prometheus_client.MetricFamily{},
```
markjen, line 49: dang, it would be nice if `metricsd.MetricsContainer` didn't use `io_prometheus_client.MetricFamily` 

not worth addressing at the moment, and this new `metrics` package should make it so that no one outside of `metrics` should need to use `io_prometheus_client`

markjen (dismissed):

src/go/metrics/collector.go
andreilee: I think somebody might eventually want to send metrics to multiple places (and register multiple `Sender`s), so the naming is reflecting that, though it isn't possible to do at the moment.

andreilee: ...Though given the current limitation, I'll just remove `RegisterSender` and just have it included as part of `NewCollector`

src/go/metrics/sender.go
andreilee: Still planning on removing this from the `metrics` package, but the Orc8r side needs to change first - I think I put some notes about this somewhere in the PR.

src/go/metrics/collector.go
andreilee: sure

src/go/metrics/gauge.go
andreilee: Oops. Thought I forgot something. Intended for the following:
- `Set` to just delete old values if they aren't drained - though perhaps this should be a configurable parameter
- channel size to be adjustable, with defaults based on how frequently metrics are collected (every minute), and the expectation that each metric is set at most once per minute, per label combination

I'll rework this bit.

andreilee:
Updated, but not all suggestions have been taken. Of note is that I'll be trying to remove the io_prometheus_client dependency, and moving that up from AGW side to the Orc8r only. This is a bit of an intermediate step to that.

src/go/metrics/collector.go
```diff
@@ -0,0 +1,59 @@
...
+package metrics
+
+import (
+	"sync"
+	"time"
+
+	"github.com/benbjohnson/clock"
+)
+
+type Collector struct {
```
markjen, line 24: minor: let's add comments on all package-level public items (i know we don't do that in other parts of the code, but I think we should start)

src/go/metrics/collector.go
```diff
@@ -0,0 +1,59 @@
...
+	clock  clock.Clock
+	period time.Duration
+	sender MetricSender
+	gauges []Gauge
+}
+
+func NewCollector(c clock.Clock, period time.Duration, sender MetricSender) *Collector {
+	return &Collector{c, period, sender, []Gauge{}}
```
markjen: nits:

1. prefer specifying field names when initializing structs. this allows us to change the field ordering later and is more clear
2. no need to init `[]Gauage{}`; nil slices are appendable.

src/go/metrics/collector.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (c *Collector) RegisterGauge(gauge Gauge) {
+	c.gauges = append(c.gauges, gauge)
```
markjen, line 44: consider making this threadsafe (via sync.Mutex) or at least documenting that it's not threadsafe.

src/go/metrics/collector.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (c *Collector) Start() {
+	go func() {
```
markjen: consider protecting against double-start?

src/go/metrics/collector.go
```diff
@@ -0,0 +1,59 @@
...
+		for {
+			c.collectMetrics()
+			select {
+			case <-c.clock.After(c.period):
```
markjen: consider creating a NewTicker:

```
ticker := c.clock.Ticker(c.period)
for {
  c.collectMetrics()
  <- ticker.C
}
```

src/go/metrics/gauge.go
```diff
@@ -0,0 +1,82 @@
...
+	name        string
+	description string
+	labelNames  []string
+	clock       clock.Clock
+	samples     []*GaugeSample
+	sync.Mutex
+}
+
+func NewGauge(name string, description string, labelNames []string) Gauge {
+	return &gaugeImpl{name, description, labelNames, clock.New(), []*GaugeSample{}, sync.Mutex{}}
```
markjen: same nit about being explicit on member field names

clock should be injected

don't need to init labelnames, samples, and Mutex

src/go/metrics/const.go
```diff
@@ -0,0 +1,57 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import "time"
+
+// Package metrics provides a generic metrics abstraction for Magma.
```
markjen, line 12: package comment goes above the package definition (i.e. on line 11)

src/go/metrics/collector_test.go
```diff
@@ -0,0 +1,34 @@
...
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
```
markjen, line 12: more cases here to exercise gauges?

src/go/metrics/gauge.go
```diff
@@ -0,0 +1,82 @@
...
+
+	g.samples = append(g.samples, &GaugeSample{value, labels, int64(time.Nanosecond) * g.clock.Now().UnixNano() / int64(time.Millisecond)})
+}
+
+// GetSamples returns all values set since it was last called.
+func (g *gaugeImpl) GetSamples() []*GaugeSample {
+	g.Lock()
+	defer g.Unlock()
+	samples := g.samples
+	g.samples = []*GaugeSample{}
```
markjen: nit: nil is fine

src/go/metrics/gauge.go
```diff
@@ -0,0 +1,82 @@
...
+	LabelValues map[string]string
+	TimestampMs int64
+}
+
+type gaugeImpl struct {
+	name        string
+	description string
+	labelNames  []string
+	clock       clock.Clock
+	samples     []*GaugeSample
```
markjen: do you need GaugeSample to be a reference? or will a value be ok?

src/go/metrics/gauge.go
```diff
@@ -0,0 +1,82 @@
...
+package metrics
+
+import (
+	"sync"
+	"time"
+
+	"github.com/benbjohnson/clock"
+)
+
+type Gauge interface {
```
markjen, line 49: is it advantageous to have Gauge be an interface?

current impl suggests it could just be a direct struct:

```
type Gauge struct {
  Name string
  Description string
  samples []GaugeSample
  sync.Mutex
}
```

(I notice there's no need for a MockGauge for testing)

src/go/metrics/gauge.go
```diff
@@ -0,0 +1,82 @@
...
+type GaugeSample struct {
+	Value       float64
+	LabelValues map[string]string
+	TimestampMs int64
+}
+
+type gaugeImpl struct {
+	name        string
+	description string
+	labelNames  []string
```
markjen, line 70: is this field used anywhere?

markjen (commented):
note: there is a build err; we need `src/go` as part of the include path in `github.com/magma/magma/src/go/...`

src/go/metrics/sender.go
```diff
@@ -0,0 +1,95 @@
...
+	io_prometheus_client "github.com/prometheus/client_model/go"
+
+	"github.com/magma/magma/protos/magma/metricsd"
+	"github.com/pkg/errors"
+	"google.golang.org/protobuf/proto"
+)
+
+type MetricSender interface {
+	// ProcessGauge processes a single gauge. All metrics need to be processed before sending.
+	ProcessGauge(gauge Gauge)
```
markjen: splitting the process / send means that the sender impl has to keep internal state.

i think we should just do this:

```
type Sender interface {
  Send(gauges []Gauge) error
}
```

what do you think?

src/go/metrics/sender.go
```diff
@@ -0,0 +1,95 @@
...
+func NewPrometheusMetricSender(client metricsd.MetricsControllerClient, gatewayId string) MetricSender {
+	container := &metricsd.MetricsContainer{
+		GatewayId: gatewayId,
+		Family:    []*io_prometheus_client.MetricFamily{},
+	}
+	return &prometheusMetricSender{container, gatewayId, client}
+}
+
+// ProcessGauge pops all recent samples from a Gauge and adds a MetricFamily to be sent
+func (s *prometheusMetricSender) ProcessGauge(gauge Gauge) {
```
markjen: in the current implementation, are you calling this func in parallel? if so, I think #63 is a data race

see my other comment on a revised Sender interface that would make the entire collect/send internal to Sender

src/go/metrics/gauge.go
andreilee: Yeah, it's a key part of metrics in Prometheus, and there's equivalent features in other cloud native time series databases.

src/go/metrics/sender.go
andreilee: Hmm... yeah good point

src/go/metrics/sender.go
andreilee: Oh nice catch. Fixed now.

src/go/metrics/collector.go
andreilee: Sure, it's been added to all.

src/go/metrics/gauge.go
markjen: what I mean is, it looks like this field is set in `NewGauge`, but I don't see it being used anywhere

is it supposed to be merged with the labels in `Set` and put into the `GaugeSample`?

src/go/go.mod
```diff
@@ -14,9 +14,12 @@ module github.com/magma/magma/src/go
 go 1.17
 
 require (
-	github.com/getsentry/sentry-go v0.11.0
+	github.com/benbjohnson/clock v1.1.0
+    github.com/getsentry/sentry-go v0.11.0
```
markjen, line 18: nit: file format off by a space?

src/go/metrics/collector_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/benbjohnson/clock"
+	"github.com/golang/mock/gomock"
+)
+
+func TestStartStop(t *testing.T) {
```
markjen, line 22: nit: fn name should be `TestCollector_Start`?

src/go/metrics/collector.go
```diff
@@ -0,0 +1,65 @@
...
+	period    time.Duration
+	sender    MetricSender
+	gauges    []Gauge
+	startOnce sync.Once
+	// gates access to gauges
+	sync.Mutex
+}
+
+func NewCollector(c clock.Clock, period time.Duration, sender MetricSender) *Collector {
+	return &Collector{clock: c, period: period, sender: sender, startOnce: sync.Once{}}
```
markjen, line 35: nit: is there a minimum period we want to enforce?

PR #9418, opened 2021-10-04
fix(mme): Fix default bearer timer expiry handler

ulaskozat:
## Summary

- Upon expiry of T3485, the timer id was not reinitialized to invalid id leading to mishandling later in `esm_ebr_start_timer` where timer arguments are populated as if the timer was not expired.
- Although not functionally creating an issue, for readability and more uniformity, `dedicated_eps_bearer_activate_t3485_handler` is modified in how the bearer context is fetched.

## Test Plan

Run the flaky test `test_3485_timer_for_default_bearer_with_mme_restart.py` multiple times consecutively. Run another integration test that checks multiple expirations.

Run s1ap integ tests.

## Additional Information


ardzoht (approved):
lgtm, thanks for the fix

PR #9426, opened 2021-10-04
feat(amf): Support for default-static QoS configuration

### LINKED ISSUE #8966: [AMF]: To support default QoS profile from subscriberd
AMF fetch the default QoS attributes such as pDUSessionAggregateMaximumBitRateDL and pDUSessionAggregateMaximumBitRateUL from Subscriberdb using s6a_proxy grpc

Kaleem-Wavelabs:
Fix: 
1) To fetch default QoS on S6a interface
2) Transfer default QoS to sessiond
3) Set the latest QoS from sessiond

Test:
1) With UE simulator
2) Orchestrator

## Additional Information


lte/protos/session_manager.proto
```diff
@@ -852,6 +852,21 @@ enum SMSessionFSMState {
...
+   uint64 br_ul = 1;
+   uint64 br_dl = 2;
+
+   enum BitrateUnitsAMBR {
+         BPS = 0;
+         KBPS = 1;
+         }
+   // Unit (either bps or Kbps)
+   BitrateUnitsAMBR br_unit = 3;
+
```
themarwhal: spacing looks weird here.

Don't we already have an ambr struct somewhere else?

Kaleem-Wavelabs: > spacing looks weird here.
> 
> Don't we already have an ambr struct somewhere else?
Fixed the unwanted spaces in proto file
Similar structure Ambr is present under oai package, not able to include oai package under session_manager.proto, so separately defined it

lte/protos/session_manager.proto
```diff
@@ -852,6 +852,17 @@ enum SMSessionFSMState {
...
+message Ambr {
+  uint64 br_ul = 1;
+  uint64 br_dl = 2;
+  enum BitrateUnitsAMBR {
+        BPS = 0;
+        KBPS = 1;
+  }
+  //Unit (either bps or Kbps)
+  BitrateUnitsAMBR br_unit = 3;
+}
```
ulaskozat: isn't this the same definition as `message AggregatedMaximumBitrate` in apn.proto? Please reuse existing proto definitions.

Kaleem-Wavelabs: Fixed the same. Thanks

ulaskozat (approved):
LGTM

PR #9429, opened 2021-10-04
feat(nms): Add subscriber bulk add new workflow

### LINKED ISSUE #8565: [NMS] Rework bulk subscriber add workflow
### Summary
Design update to the 'bulk subscriber add' workflow, to make the process easier to understand and use
[Design Mockup](https://www.figma.com/file/eCFQciASKwI6ydKsTXTafw/Subscribers-Bulk-Edit?node-id=127%3A35040)
- IMSI column is always shown when using horizontal scrolling
- Upload CSV and 'Add Row' buttons are restyled
- Add a new 'Upload CSV' modal workflow

HannaFar:
## Summary

Add subscriber bulk add

https://user-images.githubusercontent.com/26038920/137506750-1ecfcc2f-d56b-46fc-bc3a-60ea969d8ab4.mov

https://user-images.githubusercontent.com/26038920/137506774-58b621ca-b607-483a-bd1e-e9a75e23710c.mov

## Additional Information


nms/packages/magmalte/app/views/subscriber/SubscriberUtils.js
```diff
@@ -62,3 +62,10 @@ export const SUBSCRIBER_EXPORT_COLUMNS = [
   {title: 'Data Plan', field: 'sub_profile'},
   {title: 'Active APNs', field: 'active_apns'},
 ];
+export const SUBSCRIBER_ADD_ERRORS = [
+  "imsi invalid, should match '^(IMSId{10,15})$'",
+  'auth key is not a valid hex',
+  'auth opc is not a valid hex',
+  'Please select a data plan',
+  'The IMSI is duplicated',
+];
```
andreilee: Change this from an array to a map

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -45,11 +50,14 @@ import {
 } from '../../state/lte/SubscriberState';
 import {JsonDialog} from './SubscriberOverview';
 import {RenderLink} from './SubscriberOverview';
-import {base64ToHex} from '@fbcnms/util/strings';
+// import {base64ToHex} from '@fbcnms/util/strings';
```
andreilee: remove - looks like an accident?

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +169,174 @@ async function exportSubscribers(props: ExportProps) {
...
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
 
-function SubscribersTable(props: WithAlert & {refresh: boolean}) {
+  const saveSubscribers = async (
+    subscribers: Array<SubscriberInfo>,
+    edit: boolean,
+  ) => {
+    let addedSubscribers = [];
```
andreilee: Add type annotations

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +169,174 @@ async function exportSubscribers(props: ExportProps) {
...
+        const authKey =
+          subscriber.authKey && isValidHex(subscriber.authKey)
+            ? hexToBase64(subscriber.authKey)
+            : '';
+
+        const authOpc =
+          subscriber.authOpc !== undefined && isValidHex(subscriber.authOpc)
+            ? hexToBase64(subscriber.authOpc)
+            : '';
+        const newSubscriber = {
```
andreilee: add type annotations

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +169,174 @@ async function exportSubscribers(props: ExportProps) {
...
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
 
-function SubscribersTable(props: WithAlert & {refresh: boolean}) {
+  const saveSubscribers = async (
+    subscribers: Array<SubscriberInfo>,
+    edit: boolean,
```
andreilee, line 219: Give a more descriptive name for this parameter, and add comments explaining its purpose if that isn't enough

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +169,174 @@ async function exportSubscribers(props: ExportProps) {
...
+        onClose={handleClose}>
+        <MenuItem data-testid="" onClick={() => setOpen(true)}>
+          <Text variant="subtitle2">Add Subscribers</Text>
+        </MenuItem>
+        <MenuItem>
+          <Text variant="subtitle2">Update Subscribers</Text>
+        </MenuItem>
+        <MenuItem>
+          <Text variant="subtitle2">Delete Subscribers</Text>
+        </MenuItem>
```
andreilee, line 335: `onClick` property?

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +169,174 @@ async function exportSubscribers(props: ExportProps) {
...
+      }
+    }
+    enqueueSnackbar(
+      ` Subscriber${successCountRef.current > 1 ? 's' : ''} saved successfully`,
+      {
+        variant: 'success',
+      },
+    );
+    setOpen(false);
+    // props.refreshContext();
```
andreilee: Remove

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +169,174 @@ async function exportSubscribers(props: ExportProps) {
...
+                variant: 'error',
+              });
+              return;
+            }
+          }
+        }
+      } catch (e) {
+        const errMsg = e.response?.data?.message ?? e.message ?? e;
+        subscriberErrors +=
+          'error saving ' + subscriber.imsi + ' : ' + errMsg + '\n';
```
andreilee: Capitalization - these errors are displayed to the user
```suggestion
          'Error saving ' + subscriber.imsi + ' : ' + errMsg + '\n';
```

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +169,174 @@ async function exportSubscribers(props: ExportProps) {
...
+            addedSubscribers.length == SUBSCRIBERS_CHUNK_SIZE ||
+            i == subscribers.length - 1
+          ) {
+            const success = await bulkAdd(addedSubscribers, subscriberErrors);
+            if (success) {
+              successCountRef.current =
+                successCountRef.current + addedSubscribers.length;
+              addedSubscribers = [];
+            } else {
+              enqueueSnackbar('Saving subscribers to the api failed1: ', {
```
andreilee:
```suggestion
              enqueueSnackbar('Saving subscribers failed: ', {
```

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +169,174 @@ async function exportSubscribers(props: ExportProps) {
...
+        variant: 'success',
+      },
+    );
+    setOpen(false);
+    // props.refreshContext();
+  };
+  const bulkAdd = async (
+    addedSubscribers: mutable_subscribers,
+    subscriberErrors: string,
+  ) => {
```
andreilee: Add function annotations - what does this function do, what are the definitions of its parameters, what is its return value type. For example, I can't tell at a glance here why `subscriberErrors` is a required parameter. That sort of info should be obvious at a glance.

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +169,174 @@ async function exportSubscribers(props: ExportProps) {
...
+    try {
+      if (subscriberErrors.length > 0) {
+        setError(subscriberErrors);
+        return false;
+      }
+      await ctx.setState?.('', addedSubscribers);
+      return success;
+    } catch (e) {
+      const errMsg = e.response?.data?.message ?? e.message ?? e;
+      setError('error saving subscribers: ' + errMsg);
```
andreilee:
```suggestion
      setError('Error saving subscribers: ' + errMsg);
```

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
andreilee: This is a bit of a strange function, and I would suggest reorganizing the functions in this file. The caller of this function can `setError` if `subscriberErrors` is `len(..) > 0`.

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +169,174 @@ async function exportSubscribers(props: ExportProps) {
...
+      },
+    );
+    setOpen(false);
+    // props.refreshContext();
+  };
+  const bulkAdd = async (
+    addedSubscribers: mutable_subscribers,
+    subscriberErrors: string,
+  ) => {
+    let success = true;
```
andreilee: This is unnecessary - just return `true` or `false` directly

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
andreilee: The user may not be clear on what we're referring to by "the api" here.

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +169,174 @@ async function exportSubscribers(props: ExportProps) {
...
+  const handleClick = event => {
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
 
-function SubscribersTable(props: WithAlert & {refresh: boolean}) {
+  const saveSubscribers = async (
```
andreilee, line 217: I would suggest restructuring this function in a few ways.

First, chunk the `subscribers` array, and operate on individual chunks, let's call each one `subscribersChunk`. Here's some options to do so: https://stackoverflow.com/questions/8495687/split-array-into-chunks

Next, rather than use a for loop to process each subscriber individually, use `map`, see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map).

Also, make sure that `try..catch` only contains the statements that require it. Currently it's wrapping almost everything in the function.

nms/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -110,6 +113,16 @@ const useStyles = makeStyles(theme => ({
   dialog: {
     height: '750px',
   },
+  ellipsis: {
+    textOverflow: 'ellipsis',
+    overflow: 'hidden',
+    width: '160px',
+    height: '1.2em',
```
andreilee: Don't mix `em` and `px`

andreilee (changes requested):

nms/packages/magmalte/app/views/subscriber/SubscriberUtils.js
```diff
@@ -62,3 +62,11 @@ export const SUBSCRIBER_EXPORT_COLUMNS = [
...
   {title: 'Active APNs', field: 'active_apns'},
 ];
+export const SUBSCRIBER_ADD_ERRORS = [
+  "imsi invalid, should match '^(IMSId{10,15})$'",
+  'auth key is not a valid hex',
+  'auth opc is not a valid hex',
+  'Please select a data plan',
+  'The IMSI is duplicated',
+  'Auth key is required',
+];
```
andreilee: Make this a map and use Object.freeze

nms/packages/magmalte/app/views/subscriber/SubscriberUtils.js
```diff
@@ -62,3 +62,11 @@ export const SUBSCRIBER_EXPORT_COLUMNS = [
   {title: 'Data Plan', field: 'sub_profile'},
   {title: 'Active APNs', field: 'active_apns'},
 ];
+export const SUBSCRIBER_ADD_ERRORS = Object.freeze({
+  IMSI_ERROR: "imsi invalid, should match '^(IMSId{10,15})$'",
```
andreilee: Change this error message, so that somebody who has no clue what regex is, will be able to understand and fix the error.

nms/packages/magmalte/app/theme/default.js
```diff
@@ -242,6 +242,11 @@ export default createMuiTheme({
         margin: '0px',
       },
     },
+    MuiAlert: {
+      root: {
+        marginBottom: '15px',
```
andreilee: nit: Typically spacing is done in multiples of pixels divisible by two, often a multiple of 4px. Just convention.
```suggestion
        marginBottom: '16px',
```

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +172,172 @@ async function exportSubscribers(props: ExportProps) {
...
+      vertical: 'bottom',
+      horizontal: 'center',
+    }}
+    transformOrigin={{
+      vertical: 'top',
+      horizontal: 'center',
+    }}
+    {...props}
+  />
+));
```
andreilee, line 195: nit: newline?
```suggestion
));

```

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +172,172 @@ async function exportSubscribers(props: ExportProps) {
     });
   }
 }
+const StyledMenu = withStyles({
```
andreilee, line 177: nit: newline?
```suggestion

const StyledMenu = withStyles({
```

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +172,172 @@ async function exportSubscribers(props: ExportProps) {
...
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
+
+  /**
+   * Save subscribers
+   *
+   * @param {Array<SubscriberInfo>} subscribers Array of subscribers to Add, Edit or Delete.
+   * @param {boolean} edit Update existing subscriber(s).
+   */
```
andreilee: Thanks for adding these comments to the function. This still needs another iteration though - an external contributor reading just this comment block and the function signature would not know how to use this function, and with what arguments.

nms/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -361,24 +350,29 @@ type DialogProps = {
   onClose: () => void,
   editProps?: EditProps,
   onSave?: (subscribers: Array<SubscriberInfo>) => void,
+  error?: string,
 };
 
-function AddSubscriberDialog(props: DialogProps) {
+export function AddSubscriberDialog(props: DialogProps) {
```
andreilee, line 359: Since you're already making some significant changes to this file, please add some documentation to any components here (short description of what the component is, and any explanation of Props), as well as documenting any non-trivial functions. Also add typing annotations where it's missing, of return types for functions.

nms/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -898,6 +988,56 @@ export function validateSubscriberInfo(
   }
   return '';
 }
+
+export function validateSubscribers(subscribers: Array<SubscriberInfo>) {
```
andreilee, line 1075: Annotate with return type, and add documentation for this function.

nms/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -898,6 +988,56 @@ export function validateSubscriberInfo(
...
+  const errors = {};
```
andreilee: This needs a type annotation, especially considering it's non-trivial.

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -45,11 +53,14 @@ import {
...
-import {base64ToHex} from '@fbcnms/util/strings';
+import {base64ToHex, hexToBase64, isValidHex} from '@fbcnms/util/strings';
 import {makeStyles} from '@material-ui/styles';
-import {useContext, useEffect, useState} from 'react';
+import {useContext, useEffect, useRef, useState} from 'react';
 import {useEnqueueSnackbar} from '@fbcnms/ui/hooks/useSnackbar';
 import {useRouter} from '@fbcnms/ui/hooks';
+import {withStyles} from '@material-ui/core/styles';
+
+const SUBSCRIBERS_CHUNK_SIZE = 2;
```
andreilee: Define what this means with a comment

nms/packages/magmalte/app/views/subscriber/SubscriberTable.js
```diff
@@ -161,8 +172,172 @@ async function exportSubscribers(props: ExportProps) {
...
+        }
+      } catch (e) {
+        const errMsg = e.response?.data?.message ?? e.message ?? e;
+        enqueueSnackbar('Saving subscribers failed : ' + errMsg, {
+          variant: 'error',
+        });
+      }
+    }
+    enqueueSnackbar(
+      ` Subscriber${successCountRef.current > 1 ? 's' : ''} saved successfully`,
```
andreilee: I don't think we're gaining much here with the added complexity. Let's keep things simple where possible.
```suggestion
      'Subscriber(s) saved successfully',
```

andreilee: Looks like `enqueueSnackbar` will be called for a successful/failed save of a chunk, and also after the `for` loop runs as well. This could mean that possibly, a lot of different messages are enqueued onto the snackbar, and some of them will be success messages, and some of them will be failure messages. This is almost definitely going to be a very confusing experience for the user if we leave it this way, so I would suggest reworking this so that only one message max is enqueued onto the snackbar. When there's a list of errors, we can expand on `error`, `setError` to store a list of errors, and have the users correct each individual row.

nms/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -624,257 +615,356 @@ function AddSubscriberDetails(props: DialogProps) {
 
   return (
     <>
-      <DialogContent>
-        {successCountRef.current > 0 && subscribers.length > 0 && (
-          <LinearProgressWithLabel
-            value={Math.round(
-              (successCountRef.current * 100) / subscribers.length,
+      {!upload ? (
```
andreilee: Based on this flag, one of two entirely different components can be rendered. Have two subcomponents - `SubscriberDetailsUpload`, and `SubscriberDetailsTable`, or something similar, and have `AddSubscriberDetails` render either of those.

andreilee (changes requested):

andreilee:
Change the default rows shown per page in the dialog to something higher than just 5. Maybe 100?

nms/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
andreilee: This still needs documentation

nms/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -599,18 +593,88 @@ export function SubscriberEditDialog(props: DialogProps) {
...
+  setSubscribers: (Array<SubscriberInfo>) => void,
+  setAddError: (Array<string>) => void,
+  addError: Array<string>,
+  setUpload: boolean => void,
+  onClose: () => void,
+  onSave?: (Array<SubscriberInfo>) => void,
+  upload: boolean,
+  subscribers: Array<SubscriberInfo>,
+  error?: string,
+};
```
andreilee: Document this type. What do the different fields mean? Ideally somebody should be able to use the component without reading all of its code.

nms/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -621,56 +685,38 @@ function AddSubscriberDetails(props: DialogProps) {
...
-                  placeholder="Enter Name"
-                  type="text"
-                  value={props.value}
-                  onChange={e => {
-                    props.onChange(e.target.value);
-                  }}
-                />
+              title: '',
+              field: '',
+              width: '5%',
```
andreilee: Set this to a pixel count - mixing our units between percentages, pixel counts, `em`, `rem`, etc., is going to lead to a lot of confusion and complexity. Let's stick to pixel counts for dimensions and spacing.

nms/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -621,56 +685,38 @@ function AddSubscriberDetails(props: DialogProps) {
...
+              editable: 'never',
+              render: rowData => (
+                <Text variant="subtitle3">
+                  {rowData.tableData?.id + 1 || ''}
```
andreilee: Is there a valid case when the id is missing? If that shouldn't happen, we should just show an error, rather than hiding it and dealing with it.

nms/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
andreilee: This still needs documentation

nms/packages/magmalte/app/views/subscriber/SubscriberAddDialog.js
```diff
@@ -898,6 +1016,59 @@ export function validateSubscriberInfo(
...
+  Object.keys(SUBSCRIBER_ADD_ERRORS).map(error => {
+    const subscriberError = SUBSCRIBER_ADD_ERRORS[error];
+    errors[subscriberError] = [];
+  });
+  subscribers.forEach((info, i) => {
+    if (!info.authKey) {
+      errors[SUBSCRIBER_ADD_ERRORS['AUTH_KEY_REQUIRED_ERROR']].push(i + 1);
+    }
+    if (!info?.imsi?.match(/^(IMSI\d{10,15})$/)) {
+      // imsi invalid, should match '^(IMSId{10,15})$'
```
andreilee: I think we can remove the comments on these lines - the same info is clear from reading the conditionals. To help make it clearer, renaming the keys in `SUBSCRIBER_ADD_ERRORS` could help - eg. instead of `AUTH_KEY_ERROR`, name it `INVALID_AUTH_KEY`. The suffix `_ERROR` is redundant - it's a key of `SUBSCRIBER_ADD_ERRORS`, after all.

andreilee (changes requested):

andreilee (approved):

PR #9441, opened 2021-10-04
feat: add system level bazelrcs to make building easier in various sy…

themarwhal:
…stems

## Summary
It gets annoying to always add the --config=docker/devcontainer/vm depending on different underlying build environments. Add a system level bazelrc at /etc/bazelrc so that this can be done implicitly.

When running bazel build ... in the magma VM , bazel will import the --config=vm flag from the system bazelrc. 

## Additional Information

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,3 +1,5 @@
+.bazel-cache/
+.bazel-cache-repo/
 .cache
 orc8r/cloud/docker/metrics-configs/alert_rules
 nms/app/node_modules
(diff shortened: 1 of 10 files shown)


uri200 (approved):
not sure if `vm-bazelrc` is a folder, a file or a command. But if it is a command, wouldnt be better to place it to `/local/sbin`?

themarwhal:
> not sure if `vm-bazelrc` is a folder, a file or a command. But if it is a command, wouldnt be better to place it to `/local/sbin`?

It's just a configuration file. So when bazel is booting up, it reads some configs from the file.

Edit: @uri200  I renamed the files to be vm.bazelrc , docker.bazelrc, etc. so hopefully it's more clear that it's a config file

uri200:
Yes that makes more sense. Thanks

PR #9444, opened 2021-10-05
fix(pipelined): ipv6 neighbor discovery handling.

pshelar:
Following patch fixes neighbour discovery request from
SGi interface.

## Test Plan
`make test`

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/ipv6_solicitation.py
+++ b/lte/gateway/python/magma/pipelined/app/ipv6_solicitation.py
@@ -260,6 +260,8 @@ class IPV6SolicitationController(MagmaController):
                 ),
             ),
         )
+        self.logger.debug("NS pkt response %s", pkt)
+
         pkt.serialize()
 
         # For NS from SGI response doesn't need tunnel information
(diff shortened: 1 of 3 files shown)


lte/gateway/python/magma/pipelined/app/ipv6_solicitation.py
```diff
@@ -21,7 +21,7 @@
 from ryu.controller.handler import MAIN_DISPATCHER, set_ev_cls
 from ryu.lib.packet import ether_types, ethernet, icmpv6, in_proto, ipv6, packet
 from ryu.ofproto.inet import IPPROTO_ICMPV6
-
+import logging
```
koolzz: rm

lte/gateway/python/magma/pipelined/app/ipv6_solicitation.py
```diff
@@ -260,6 +260,8 @@ def _send_neighbor_advertisement(
                 ),
             ),
         )
+        logging.debug("NS pkt response %s", pkt)
```
koolzz: just use self.logger.debug, its included for all MagmaControllers

koolzz (approved):

PR #9456, opened 2021-10-05
fix: Fix show-tech tool checksum errors

wallyrb:
## Summary

- Change copy module with fetch module. Copy module always do checksum validation, for logs that changes very fast(syslogs) the task kept failing due to checksum errors. Fetch module has the option to set `validate_checksum: no`
- There was an error in getting the /etc/magma/ configuration. Fixing the path
- Renaming "debian" with "ubuntu" in some commands.

## Test Plan

- Updated playbook has been tested in my local environment and one production AGW.

### DIFF
--- a/show-tech/constants.yml
+++ b/show-tech/constants.yml
@@ -11,7 +11,7 @@ general_commands:
   - "uname -a"
   - "top -b -n 1"
   - "df -kh"
-debian_commands:
+ubuntu_commands:
   - "dpkg -l magma*"
   - "dpkg -l *openvswitch*"
   - "ovs-vsctl show"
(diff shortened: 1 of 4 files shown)


mstre123:
is there any value to just creating a new config file for ubuntu.yml while leaving the debian.yml file?

wallyrb:
> is there any value to just creating a new config file for ubuntu.yml while leaving the debian.yml file?

@mstre123 IMO there is no value. From the commands in that list, there are no specific commands applicable for one distribution only. 

Reason of changing is only for consistency on logs and file names. Since we are using ubuntu only since 1.5+, it get some confusion from users when they see is running debian commands.

mstre123:
> > is there any value to just creating a new config file for ubuntu.yml while leaving the debian.yml file?
> 
> @mstre123 IMO there is no value. From the commands in that list, there are no specific commands applicable for one distribution only.
> 
> Reason of changing is only for consistency on logs and file names. Since we are using ubuntu only since 1.5+, it get some confusion from users when they see is running debian commands.

got it makes sense.

pshelar (approved):
LGTM

PR #9460, opened 2021-10-05
fix(mme): Fix for handling ULA failure

pruthvihebbani:
fix(mme): Fix for handling ULA failure

## Summary

In case of ULA failure, mme was triggering attach reject and ue context release twice, once as part of emm_proc_attach_reject() and then in fsm handling. This PR fixes the issue.

## Test Plan

- Verified that attach reject and ue context release messages are sent only once in case of ULA failure by negating if (status.ok()) in s6a_handle_update_location_ans and executing test_attach_detach.py testcase
- Verified s1ap tester sanity suite


ulaskozat:
You need to update the mme unit test added and as suggested here: https://github.com/magma/magma/pull/9395

ulaskozat (changes requested):
modify the failing unit test as indicated in https://github.com/magma/magma/pull/9395

ulaskozat (approved):
LGTM

PR #9490, opened 2021-10-06
feat(orc8r): [Terraform][helm] improve HA and reliability of EKS based orc8r deployment

119Vik:
## Summary
At the moment default EKS deployment have following issues, which make it not truly production ready:
 - magmalte deployment is not HA
 - magmalte deployment does not have PDB  (which kinda makes sense cause it's not HA)
 - non of the deployment has affinity / anti-affinity specified

Following was done to make default EKS based deployment a bit more production ready:
 - orc8r-app terraform module was extended with magmalte replicas count configuration
 - orc8r-app terraform module was extended with affinity configuration for magmalte and controller
 - orc8r helm chart was extended with PDB for magmalte

### DIFF
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/main.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/main.tf
@@ -182,7 +182,10 @@ data "template_file" "orc8r_values" {
     orc8r_db_user    = var.orc8r_db_user
     orc8r_db_pass    = var.orc8r_db_pass
 
-    deploy_nms  = var.deploy_nms
+    deploy_nms          = var.deploy_nms
+    magmalte_replicas   = var.nms_replicas
+    ui_nginx_replicas   = var.nginx_replicas
+    
 
     metrics_pvc_promcfg  = kubernetes_persistent_volume_claim.storage["promcfg"].metadata.0.name
     metrics_pvc_promdata = kubernetes_persistent_volume_claim.storage["promdata"].metadata.0.name
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/main.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/main.tf
@@ -182,9 +182,9 @@ data "template_file" "orc8r_values" {
     orc8r_db_user    = var.orc8r_db_user
     orc8r_db_pass    = var.orc8r_db_pass
 
-    deploy_nms          = var.deploy_nms
-    magmalte_replicas   = var.nms_replicas
-    ui_nginx_replicas   = var.nginx_replicas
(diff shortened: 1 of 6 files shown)


hcgatewood (commented):
Hey @119Vik! This looks like the right way to achieve this. For my understanding, can you describe the value-add for having a multi-pod NMS? My minimal understanding is that the NMS app expects to run as a singleton, but @andreilee please educate me if I'm incorrect

119Vik:
@hcgatewood few things :
 - for maintenance we need at least 2 pods so we can do EKS maintenance and recreate node even when NMS pod was there
 - ensure that we have a way to scale pod when more  clients coming

119Vik:
@hcgatewood maybe it make sense to add some policy that there should be at least one replica per node so we can do safe EKS maintenance

orc8r/cloud/deploy/terraform/orc8r-helm-aws/main.tf
```diff
@@ -182,7 +182,10 @@ data "template_file" "orc8r_values" {
     orc8r_db_user    = var.orc8r_db_user
     orc8r_db_pass    = var.orc8r_db_pass
 
-    deploy_nms  = var.deploy_nms
+    deploy_nms          = var.deploy_nms
+    magmalte_replicas   = var.nms_replicas
+    ui_nginx_replicas   = var.nginx_replicas
```
mattymo: I would call this nms_nginx_replicas

orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
```diff
@@ -85,6 +85,18 @@ variable "deploy_nms" {
   default     = true
 }
 
+variable "nms_replicas" {
+  description = "Replica count for magmalte pods."
+  type        = number
+  default     = 1
+}
+
+variable "nginx_replicas" {
```
mattymo: This is a bad var name because there's nginx both for orc8r and nms

orc8r/cloud/deploy/terraform/orc8r-helm-aws/main.tf
119Vik: good point, fixed

119Vik:
@mattymo @hcgatewood @andreilee  - as a result of testing and improving things this patch grew up so i updated title and description

hcgatewood:
Waiting on @andreilee for input on the architecture update of NMS pod from singleton to 2+

hcgatewood:
> Waiting on @andreilee for input on the architecture update of NMS pod from singleton to 2+

Friendly bump @andreilee -- also happy to dive into the code a bit myself if you have any entry points

PR #9468, opened 2021-10-05
feat(agw): Add interface for tracing, implement interface against sen…

mstre123:
…try-go.

Signed-off-by: mstre123 <mstre123@gmail.com>

## Summary

Interface for tracing in AGW

## Test Plan

test locally

## Additional Information

Will fill out additional tests, comments etc. Wanted to get eyes on interface first.

PR #9504, opened 2021-10-06
feat(agw): Update README with bazel clean command.

mstre123:

## Summary

Small update to readme

## Test Plan

n/a

### DIFF
--- a/src/go/README.md
+++ b/src/go/README.md
@@ -124,3 +124,11 @@ package. In this context, "clean" means the interface package is defined
 entirely with Golang built-ins and other clean magma interface packages. i.e.
 there are no 3rd party library dependencies, except for test helpers (e.g.
 gomock).
+
+### Troubleshooting
+
+If you have issues running bazel after switching between branches you may need
+to clean your bazel workspace
+```sh
+bazel clean --expunge
+```


src/go/README.md
```diff
@@ -124,3 +124,11 @@ package. In this context, "clean" means the interface package is defined
 entirely with Golang built-ins and other clean magma interface packages. i.e.
 there are no 3rd party library dependencies, except for test helpers (e.g.
 gomock).
+
+### Troubleshooting
+
+If you have issues running bazel after switching between branches you may need
+to clean your bazel workspace
+```sh
+bazel clean --expunge
```
markjen, line 133: @themarwhal @electronjoe ideas why this would be necessary?

markjen (approved):

src/go/README.md
themarwhal: that's odd, it shouldn't be necessary. @mstre123 what error were you seeing?

themarwhal: and how were you running bazel? in the magma VM? or devcontainer

mstre123: I was getting this error 
```Every .bzl file must have a corresponding package, but X does not have one```
Running locally on my mac

got the fix from this issue
https://github.com/bazelbuild/bazel/issues/12630

themarwhal: How did you run bazel? Like this? `bazel test src/go/...:all`. Note that all other directories apart from src/go cannot be run on macs. 

Also did you rebase from a really old branch maybe? The only times I've had to run clean --expunge is when I changed the output_base parameter, and I did that maybe like a month ago.

mstre123: I ran bazel run //:gazelle from src/go. I was jumping between branches for #9403 and #9468 and both stemmed from master as of last week.
Possible it was a fluke

themarwhal: Hmm yeah that should work. Hmm I'm not sure what happened here. Can we hold off on landing this until we see this issue again? If something is making it so that we need to run expunge periodically, we're doing something very wrong, I think  :D

mstre123: makes sense i'll close out this commit without pushing, will ping you if it comes up again.

mstre123:
closing for now re conversation here https://github.com/magma/magma/pull/9504#discussion_r725053920

PR #9488, opened 2021-10-06
fix: corrects turn-up of gopls in devcontainer

electronjoe:
This PR corrects a previous attempt (#9399) to add `gopls` to our devcontainer for vscode use.

The prior PR mistakenly installed `gopls` to the root home directory and was inaccessible to our vscode enviornment which is configured to run as user `vscode`.

To avoid making a mistake again, this time I tested more thoroughly.

## Test Plan

- Modified `.devcontainer/dev.container.json
  - removed `"image":` line
  - replaced with
 ```json
	"dockerFile": "Dockerfile",
	"context": "..",
```
- Updated `.devcontainer/Dockerfile` per this PR
- Pressed `F1` and selected `Remote-Containers: Rebuild Container` on a local vs code instance of our repo
- Verified healthy `gopls` behavior

Note that `gopls` was not happy about lack of a `go.mod` in our root repository - but various functions like code discovery and autocomplete were functional.

Signed-off-by: Scott Moeller <electronjoe@gmail.com>

PR #9507, opened 2021-10-06
fix: fix system wide bazelrc path in devcontainer

themarwhal:

## Summary
Bad copy paste in this Dockerfile. The magma directory lives under /workspaces, not /

## Test Plan
Tested inside devcontainer

## Additional Information

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -313,6 +313,6 @@ RUN go get -v golang.org/x/tools/gopls
 #### Update shared library configuration
 RUN ldconfig -v
 
-RUN ln -s /magma/experimental/bazel-base/bazelrcs/devcontainer.bazelrc /etc/bazelrc
+RUN ln -s "$MAGMA_ROOT"/experimental/bazel-base/bazelrcs/devcontainer.bazelrc /etc/bazelrc
 
-WORKDIR /workspaces/magma
+WORKDIR $MAGMA_ROOT


.devcontainer/Dockerfile
```diff
@@ -313,6 +313,6 @@ RUN go get -v golang.org/x/tools/gopls
 #### Update shared library configuration
 RUN ldconfig -v
 
-RUN ln -s /magma/experimental/bazel-base/bazelrcs/devcontainer.bazelrc /etc/bazelrc
+RUN ln -s /workspaces/magma/experimental/bazel-base/bazelrcs/devcontainer.bazelrc /etc/bazelrc
```
quentinDERORY: Use the MAGMA_ROOT environment variable instead

quentinDERORY (changes requested):
See comment

.devcontainer/Dockerfile
themarwhal: ah good point!

quentinDERORY (approved):

.devcontainer/Dockerfile
```diff
@@ -313,6 +313,6 @@ RUN go get -v golang.org/x/tools/gopls
 #### Update shared library configuration
 RUN ldconfig -v
 
-RUN ln -s /magma/experimental/bazel-base/bazelrcs/devcontainer.bazelrc /etc/bazelrc
+RUN ln -s "$MAMGA_ROOT"/experimental/bazel-base/bazelrcs/devcontainer.bazelrc /etc/bazelrc
```
electronjoe: Hadolint thinks this is mis-spelled, I think so too? Did you test this change? I have a recommended process (after my attempts in this area failed also...) detailed in test plan https://github.com/magma/magma/pull/9488

themarwhal: I was waiting for the Dockerfile health job to finish. Was hoping that would catch any errors.

themarwhal: but yea totally it is mispelled!

electronjoe: Hadolint works! I was impressed.

themarwhal: yeah goo job hadolint! nice to be able to catch it in CI

PR #9521, opened 2021-10-07
fix(agw): [Non_Sanity] Fixed standalone pdn conn req with apn correction failing

VinashakAnkitAman:
## Title
[Non_Sanity] Fixed standalone pdn conn req with apn correction and ESM information with APN correction

## Summary
The non sanity test case s1aptests/test_standalone_pdn_conn_req_with_apn_correction.py and s1aptests/test_attach_esm_info_with_apn_correction.py were failing because MME was not reachable after the small sleep time set in the test script. This PR fixes the issue by increasing the sleep time to 30 seconds for MME to come up properly

## Test Plan
Tested individually and as part of sanity


ardzoht (approved):
lgtm

PR #9524, opened 2021-10-07
feat(feg): Services Based Interface code generation for N7 interface

GANESH-WAVELABS:

## Summary
 - 3GPP SBI Code generated based on https://github.com/deepmap/oapi-codegen. SBI openapis are defined in OpenAPI 3.0 specification and go-swagger will not work.
 - Though only N7 inteface openapi is required, it has depencies on other 3GPP openapi files. Hence other files are also included.
 - The OpenAPI specifications are taken from 3GPP specification v15.8. The  following modifications are made to the specifications
    - Resolved cycling dependencies by moving the definitions to common openapi file
    - Changed few names that were causing compilation error on the generated code
 - Added go:generate for regenerating the go code, when the openapi yaml files are changed.
 - Each generated go file is maintained in its own package, because of the way the code is generated and to avoid conflicting definitions.

Files obtained from 3GPP are in the directory - feg/gateway/sbi/defs/TS*.yaml

Files generated - feg/gateway/sbi/specs/TS*/TS*.gen.go

Files modified or newly added
feg/gateway/sbi/sbi_test.go
feg/gateway/sbi/sbi.go
feg/gateway/docker/go/Dockerfile
feg/gateway/go.mod
feg/gateway/go.sum
feg/gateway/Makefile

## Test Plan
Unit test code tests client request to a mock server.


uri200:
Can we try to shrink this to the files N7 really needs only? 

We should not maintain unused code

feg/gateway/sbi/n7_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+type MockPcf struct {
+	Policies map[string]n7.SmPolicyDecision
+}
+
+func TestCreateSMPolicy(t *testing.T) {
+
+	startServer()
+
+	client, err := n7_client.NewClient("http://localhost:8080")
```
uri200: Try to use `http://localhost:0` 
We should not use hardcoded ports for testing
Try to see if there are functions to get the running port once started.

feg/gateway/sbi/n7_test.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
```
uri200: Use assert library.

You can also build a `TestCase` struct if there are many repeated test to be done like [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=cwf%2Fcloud%2Fgo%2Fservices%2Fcwf%2Fanalytics%2Fcalculations%2Fcwf_calculations_test.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=127&start_col=4&end_row=127&end_col=4)


For examples, please check other test files in magma

feg/gateway/sbi/n7_test.go
```diff
@@ -0,0 +1,118 @@
...
+		log.Fatalf("Failed to create new client %s", err)
+	}
+
```
uri200: please add some comments to the test cases
Examples in magma like [this](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fs8_proxy%2Fservicers%2Fs8_proxy_test.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=59&start_col=9&end_row=59&end_col=9)

feg/gateway/sbi/n7_test.go
```diff
@@ -0,0 +1,118 @@
...
+	ipAddr := sbi.Ipv4Addr("10.1.2.3")
+	imsi := sbi.Supi("12345678901234")
+	body := n7_client.PostSmPoliciesJSONRequestBody{
+		Ipv4Address:  &ipAddr,
+		PduSessionId: 10,
+		Supi:         imsi,
+	}
+
+	log.Println("Sending PostSmPolicies")
```
uri200: why do you need to log stuff on a test?

uri200: Note you should use your debugger in go, no prints. 
Golang is easy to run on an IDE.
Furthermore, any test added should ran on an IDE to make sure test runs on IDE environment

image

feg/gateway/sbi/n7_test.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	_ = resp
+
+	log.Println("Sending GetSmPoliciesSmPolicyId")
+
+	resp, err = client.GetSmPoliciesSmPolicyId(context.Background(), "1234")
+	if err != nil {
+		log.Fatalf("Failed to get SM Policy %s", err)
+	}
+	_ = resp
```
uri200: ?

feg/gateway/sbi/n7_test.go
```diff
@@ -0,0 +1,118 @@
...
+		Supi:         imsi,
+	}
+
+	log.Println("Sending PostSmPolicies")
+
+	resp, err := client.PostSmPolicies(context.Background(), body)
+	if err != nil {
+		log.Fatalf("Failed to get SM Policy %s", err)
+	}
+	_ = resp
```
uri200: not sure what is this

feg/gateway/sbi/n7_test.go
```diff
@@ -0,0 +1,118 @@
...
+		Policies: make(map[string]n7.SmPolicyDecision),
+	}
+}
+
+func (pcf *MockPcf) PostSmPolicies(ctx echo.Context) error {
+	log.Println("PostSmPolicies")
+	return ctx.NoContent(http.StatusOK)
+}
+
+// (GET /sm-policies/{smPolicyId})
```
uri200: Comments in go must start with the name of the function

```
// GetSmPoliciesSmPolicyId handles POST /sm-policies/{smPolicyId}/delete

```

feg/gateway/sbi/n7_test.go
```diff
@@ -0,0 +1,118 @@
...
+	log.Println("Sending GetSmPoliciesSmPolicyId")
+
+	resp, err = client.GetSmPoliciesSmPolicyId(context.Background(), "1234")
+	if err != nil {
+		log.Fatalf("Failed to get SM Policy %s", err)
+	}
+	_ = resp
+}
+
+func NewMockPcf() *MockPcf {
```
uri200: structs and objects definition should be at the top of the file

Please compare it with other go files at magma

feg/gateway/sbi/n7_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+		n7.RegisterHandlers(eserver, mockPcf)
+		serverStarted <- struct{}{}
+
+		eserver.Start("localhost:8080")
+	}()
+
+	<-serverStarted
+
+	time.Sleep(1 * time.Second)
```
uri200: Imagine have 1k test with 1s sleep each. This is a very bad practice and needs to be avoided. In future we will not add any time sleep unless is strictly necessary. Specially if it is 1 second long.  

The reason you are forced to ad a sleep here is because your go routine with the server will not be started until your main routine has given up control. The sleep causes that.

feg/gateway/sbi/n7_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	mockPcf := NewMockPcf()
+	go func() {
+		eserver := echo.New()
+
+		n7.RegisterHandlers(eserver, mockPcf)
+		serverStarted <- struct{}{}
+
+		eserver.Start("localhost:8080")
+	}()
```
uri200: Why do you need a go routine to start the test server?

feg/gateway/sbi/n7_test.go
```diff
@@ -0,0 +1,118 @@
...
+	serverStarted := make(chan struct{})
+
+	mockPcf := NewMockPcf()
+	go func() {
+		eserver := echo.New()
+
+		n7.RegisterHandlers(eserver, mockPcf)
+		serverStarted <- struct{}{}
+
+		eserver.Start("localhost:8080")
```
uri200: you server is started AFTER you have informed through the channel. That will cause issues because your main routing thinks your server is started but it is not

uri200 (changes requested):

uri200:
Hey @GANESH-WAVELABS , what IDE are you using for GO. We do need to take advantage of how easy is to debug with IDE on go.

feg/gateway/sbi/n7_test.go
GANESH-WAVELABS: Moved the definitions to top.

feg/gateway/sbi/n7_test.go
GANESH-WAVELABS: The Start method is a blocking call and hence run on a goroutine.

GANESH-WAVELABS:
> Hey @GANESH-WAVELABS , what IDE are you using for GO. We do need to take advantage of how easy is to debug with IDE on go.

Have been using vscode and it also has IDE debugging. Started with goland.

GANESH-WAVELABS:
> Can we try to shrink this to the files N7 really needs only?
> 
> We should not maintain unused code

@uri200 The 3GPP openapi definitions introduces such dependencies that required the inclusion of other openapi files.

uri200:
Thanks @GANESH-WAVELABS , VSCode sounds good. We use intellij but VScode will have almost every basic feature than Intellij. I want to try to avoid of using plain VIM for coding. You can still use VIM in vscode, but let's try to take advantage of all automation we can. 

Can you try to run your test using VSCode debugger and tell us if it works? Can you paste a screenshot with the debugger working.  We may switch from intellij to VScode in the near future and it would be good to start comparing them.

lte/gateway/Vagrantfile
```diff
@@ -40,7 +40,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     magma.vm.network "private_network", ip: "192.168.129.1", nic_type: "82540EM"
 
     # IPv6 interface
-    config.vm.provision :shell, inline: "ip -6 a add  2020::10/64 dev eth0 && ip -6 r add default via 2020::1", run: 'always'
```
uri200: is this legit

uri200:
Not sure of this, but should `openapi/sbi_15.8` be part of the `sbi` folder? In my view it may make sense to put all sbi related under sbi folder (yaml source files and go generated files)?  But please provide your view, maybe it make sense to keep them separate

Also I think all TS folders under sbi could be part of a somethinc called `sbi/TS` or `sbi/specs` folder. 

image

GANESH-WAVELABS:
> Can you try to run your test using VSCode debugger and tell us if it works? Can you paste a screenshot with the debugger working. We may switch from intellij to VScode in the near future and it would be good to start comparing them.

@uri200 Here is a screenshot of debugging a unit test case with VS Code.
Screenshot from 2021-11-01 16-31-00

GANESH-WAVELABS:
> Not sure of this, but should `openapi/sbi_15.8` be part of the `sbi` folder? In my view it may make sense to put all sbi related under sbi folder (yaml source files and go generated files)? But please provide your view, maybe it make sense to keep them separate

@uri200 
Was trying to align with the existing directory structure where the proto and swagger files were in the magma/feg directory and the source files were not in the same directory.

uri200:
> > Not sure of this, but should `openapi/sbi_15.8` be part of the `sbi` folder? In my view it may make sense to put all sbi related under sbi folder (yaml source files and go generated files)? But please provide your view, maybe it make sense to keep them separate
> 
> @uri200 Was trying to align with the existing directory structure where the proto and swagger files were in the magma/feg directory and the source files were not in the same directory.

Hey @GANESH-WAVELABS  here you have an example of where were store those files.

Note we have yml definitions and generated files. But not any test file. Lets try to build a similar structure where only definition and generated files exist. If you put them on a subfolder you can import them as a module

https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fcloud%2Fgo%2Fservices%2Ffeg%2Fobsidian%2Fmodels%2Fswagger.v1.yml&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=31&start_col=22&end_row=31&end_col=22

feg/gateway/sbi/sbi.go
```diff
@@ -0,0 +1,63 @@
...
+//go:generate oapi-codegen -generate types,skip-prune -o ../../openapi/sbi_15.8/TS29122ReportingNetworkStatus/TS29122ReportingNetworkStatus.gen.go --import-mapping TS29122_CommonData.yaml:magma/feg/gateway/sbi/TS29122CommonData,TS29571_CommonData.yaml:magma/feg/gateway/sbi/TS29571CommonData TS29122_ReportingNetworkStatus.yaml
+//go:generate oapi-codegen -generate types,skip-prune -o ../../openapi/sbi_15.8/TS29122GMDviaMBMSbyxMB/TS29122GMDviaMBMSbyxMB.gen.go --import-mapping TS29572_Nlmf_Location.yaml:magma/feg/gateway/sbi/TS29572NlmfLocation,TS29122_CommonData.yaml:magma/feg/gateway/sbi/TS29122CommonData,TS29571_CommonData.yaml:magma/feg/gateway/sbi/TS29571CommonData TS29122_GMDviaMBMSbyxMB.yaml
+//go:generate oapi-codegen -generate types,skip-prune -o ../../openapi/sbi_15.8/TS29510NnrfNFDiscovery/TS29510NnrfNFDiscovery.gen.go --import-mapping TS29510_Nnrf_NFManagement.yaml:magma/feg/gateway/sbi/TS29510NnrfNFManagement,TS29503_Nudm_SDM.yaml:magma/feg/gateway/sbi/TS29503NudmSDM,TS29571_CommonData.yaml:magma/feg/gateway/sbi/TS29571CommonData TS29510_Nnrf_NFDiscovery.yaml
```
uri200: i think oapi-codegen can grab the specs from github directly. Wouldn't it make more sense to get them from there instead of having to maintain them in magma?

Would it be possible to use filters to generate any spec dropped at sbi folder?


https://github.com/deepmap/oapi-codegen

uri200:
plan to add a Makefile command to generate the files

Note we do the same for orc8r when we run `build.py -g`

feg/gateway/sbi/sbi.go
GANESH-WAVELABS: > i think oapi-codegen can grab the specs from github directly. Wouldn't it make more sense to get them from there instead of having to maintain them in magma?

When you say "specs from github directly" where (which repo) would these yaml files be kept? Also some of the yaml files were modified by us to resolve cyclic dependencies for type definitions. 

> Would it be possible to use filters to generate any spec dropped at sbi folder?

The main issue for not going this way is the --import-mapping flag, which needs to be provided to resolve the dependencies for type defined in other specs. These import mappings for each spec file were determined manually.

GANESH-WAVELABS:
> plan to add a Makefile command to generate the files

@uri200 
There is already a Makefile target "gen" in mamga/feg/gateway/Makefile, that invokes "go gen ./...". Should this be fine or did you mean something else?

GANESH-WAVELABS:
> Note we have yml definitions and generated files. But not any test file. Lets try to build a similar structure where only definition and generated files exist. If you put them on a subfolder you can import them as a module

@uri200 Please let me know if this directory structure is fine

magma/feg/gateway/sbi - all the yaml files goes here
magma/feg/gateway/sbi/sbi.go - will be used for "go generate"
magma/feg/gateway/sbi/TS* - will have the generated definitions files

Should the sbi directory be under magma/feg/gateway or under mamga/feg/gateway/services?

The test file was added for the sake of completion of the PR to check if the code generation works fine.

uri200:
Hey @GANESH-WAVELABS why do we need to throw it all in the same folder. We can separate it to make it clearer.   
```
magma/feg/gateway/sbi- will be used for "go generate" (files like sbi.go  and sbi_test.go 
magma/feg/gateway/sbi/defs - all the yaml files goes here
magma/feg/gateway/sbi/specs/TS* - will have the generated go specs (coming from the yamls in def folder definitions files)
```


Related to the other question (make command to gen specs)
We should have a make command just to generate the specs this way if we ever change it we can generate only them and see the changes before coding it.

emakeev (commented):
I'd recommend splitting this PR, maybe to at least 3 PRs (yaml files in a dedicated directory; generated files in another dedicated dir & the implementation for review)

GANESH-WAVELABS:
> I'd recommend splitting this PR, maybe to at least 3 PRs (yaml files in a dedicated directory; generated files in another dedicated dir & the implementation for review)

@emakeev 
Since the yaml files (from 3GPP) and the generated code would not be needing a review, the files that are required to be reviewed is mentioned in the Summary (along with the directory structure), these were raised as a single PR.

GANESH-WAVELABS:
> Hey @GANESH-WAVELABS why do we need to throw it all in the same folder. We can separate it to make it clearer.
> 
> ```
> magma/feg/gateway/sbi- will be used for "go generate" (files like sbi.go  and sbi_test.go 
> magma/feg/gateway/sbi/defs - all the yaml files goes here
> magma/feg/gateway/sbi/specs/TS* - will have the generated go specs (coming from the yamls in def folder definitions files)
> ```
> 
> Related to the other question (make command to gen specs) We should have a make command just to generate the specs this way if we ever change it we can generate only them and see the changes before coding it.

@uri200 
These changes are done.

uri200:
Thanks @GANESH-WAVELABS  that looks more organized

image

feg/gateway/Makefile
```diff
@@ -53,6 +53,9 @@ fmt:
 gen:
 	go generate ./...
 
+gen_sbi:
+	go generate feg/gateway/sbi
+
```
uri200: have you tried to run this command from its current path?

shouldn't this be just `go generate sbi`

uri200: by the way i see we already have `gen` endpoint. Does running `make gen` generates sbi too?

If so, then we dont need `gen_sbi` as an endpont

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// Get SM Policy given the policyId
+	policyResp, err := client.GetSmPoliciesSmPolicyIdWithResponse(context.Background(), "1234")
+	assert.NoError(t, err, "Failed to get SM Policies")
+	assert.Equal(t, http.StatusOK, policyResp.StatusCode())
+	assert.Equal(t, &ipAddr, policyResp.JSON200.Context.Ipv4Address)
+	assert.Equal(t, imsi, policyResp.JSON200.Context.Supi)
+}
+
+func startServer(t *testing.T) *MockPcf {
```
uri200, line 113: We have something similar in our codebase [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=orc8r%2Fcloud%2Fgo%2Fobsidian%2Freverse_proxy%2Freverse_proxy_test.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=73&start_col=17&end_row=73&end_col=17). But i think your way looks proper. 

```
	handler := NewReverseProxyHandler()
	e, err := startTestServer(handler)					 <-----
	assert.NoError(t, err)

	listener := tests.WaitForTestServer(t, e)			<-----
	if listener == nil {
		return // WaitForTestServer should have 'logged' error already
	}
	urlPrefix := "http://" + listener.Addr().String()
```

Try maybe compare it with your code to see if something can be done better (i haven't, so this is just a suggestion)

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,158 @@
...
+	addr, err := waitForServerStart(eserver, errChan)
+
+	assert.NoError(t, err)
+
+	mockPcf.ListenAddr = addr
+
+	return mockPcf
+}
+
+func waitForServerStart(e *echo.Echo, errChan <-chan error) (string, error) {
```
uri200: waitForTestServer may make more sense

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}()
+
+	addr, err := waitForServerStart(eserver, errChan)
+
+	assert.NoError(t, err)
+
+	mockPcf.ListenAddr = addr
+
+	return mockPcf
```
uri200, line 130: try not to use so many blank lines 

This may look better
```
	mockPcf.ListenAddr, err := waitForServerStart(eserver, errChan)
	assert.NoError(t, err)
	return mockPcf
```

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/labstack/echo/v4"
+	"github.com/stretchr/testify/assert"
+
+	n7_client "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
+	n7 "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControlServer"
+	sbi "magma/feg/gateway/sbi/specs/TS29571CommonData"
+)
+
+type MockPcf struct {
```
uri200, line 33: maybe it would be better to have the echo server exposed here instead of `ListenAddr`. This way you can access any propery. Note this is just a mock object

uri200 (changes requested):

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,158 @@
...
+	var newPolicy n7.SmPolicyContextData
+	err := ctx.Bind(&newPolicy)
+	if err != nil {
+		return ctx.NoContent(http.StatusBadRequest)
+	}
+
+	var policy n7.SmPolicyControl
+	policy.Context = newPolicy
+
+	pcf.Policies["1234"] = policy
```
uri200: if 1234 is something reused by other tests, please create a constant

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,158 @@
...
+	mockPcf := startServer(t)
+
+	// Create new N7 client object
+	client, err := n7_client.NewClientWithResponses(fmt.Sprintf("http://%s", mockPcf.ListenAddr))
+
+	assert.NoError(t, err, "Client creation failed")
+
+	// Create a new SM Policy
+	ipAddr := sbi.Ipv4Addr("10.1.2.3")
+	imsi := sbi.Supi("12345678901234")
```
uri200: Define constants at the top of this files to define those values  like`12345678901234`

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,158 @@
...
+	resp, err := client.PostSmPolicies(context.Background(), body)
+	assert.NoError(t, err, "Failed to create SM Policies")
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+	// Get SM Policy given the policyId
+	policyResp, err := client.GetSmPoliciesSmPolicyIdWithResponse(context.Background(), "1234")
+	assert.NoError(t, err, "Failed to get SM Policies")
+	assert.Equal(t, http.StatusOK, policyResp.StatusCode())
+	assert.Equal(t, &ipAddr, policyResp.JSON200.Context.Ipv4Address)
+	assert.Equal(t, imsi, policyResp.JSON200.Context.Supi)
```
uri200: in my opinion is way better to have something  like this

```
const IMSI1 =  "12345678901234"

assert.Equal(t, sbi.Supi(IMSI1), policyResp.JSON200.Context.Supi)

or 

assert.Equal(t, IMSI1 ,  fmt.Sprint(policyResp.JSON200.Context.Supi)
....


```

Otherwise you have to look in the test for what was imsi.

Same for the other test cases

uri200 (changes requested):

feg/gateway/Makefile
GANESH-WAVELABS: make gen does generate sbi files. This was mentioned previously, see the quote below. Will remove gen_sbi target.

> There is already a Makefile target "gen" in mamga/feg/gateway/Makefile, that invokes "go gen ./...". Should this be fine or did you mean something else?

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: Tried reusing WaitForTestServer from tests package. It does not work as the Echo server version is different. The generated code requires Echo v4, however the one used in the tests package appears to be a older one.

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: This is now changed.

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,165 @@
...
+		n7.RegisterHandlers(eserver, mockPcf)
+
+		err := eserver.Start("localhost:0")
+		if err != nil {
+			errChan <- err
+		}
+	}()
+
+	addr, err := waitForTestServer(eserver, errChan)
+	assert.NoError(t, err)
```
uri200: you can use `assert.Required(t, err) ` so this way the test fails imediately. 

In case mock pgw has an error doesn't make much sense to go on. And that will help to pinpoint the issue

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,165 @@
...
+	mockPcf.ListenAddr = addr
+
+	return mockPcf
+}
+
+// waitForTestServer waits for the Echo server to be launched and returns the Listener Address.
+// There is a similar one in magma/orc8r/cloud/go/obsidian/access/tests but that uses a older
+// version of Echo server.
```
uri200: Plese, remove that comment . You can use github comments to mention where your code came from

Remember that comments before functions must start with the name of the function
// waitForTestServer

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,165 @@
...
+			return "", ctx.Err()
+		case <-ticker.C:
+			addr := e.ListenerAddr()
+			if addr != nil && strings.Contains(addr.String(), ":") {
+				// server started
+				return addr.String(), nil
+			}
+		case err := <-errChan:
+			if err == http.ErrServerClosed {
+				return "", nil
```
uri200, line 153: Why is this function taking an error and not returning it? This seems to be hiding an error

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,165 @@
...
+	mockPcf.ListenAddr = addr
+
+	return mockPcf
+}
+
+// waitForTestServer waits for the Echo server to be launched and returns the Listener Address.
+// There is a similar one in magma/orc8r/cloud/go/obsidian/access/tests but that uses a older
+// version of Echo server.
+func waitForTestServer(e *echo.Echo, errChan <-chan error) (string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
```
uri200, line 135: why do we need the context here.
Normally context it is used as a way to communicate between go routines or through stack of functions.

https://dev.to/hekonsek/using-context-in-go-timeout-hg7

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,165 @@
...
+	assert.NoError(t, err)
+	mockPcf.ListenAddr = addr
+
+	return mockPcf
+}
+
+// waitForTestServer waits for the Echo server to be launched and returns the Listener Address.
+// There is a similar one in magma/orc8r/cloud/go/obsidian/access/tests but that uses a older
+// version of Echo server.
+func waitForTestServer(e *echo.Echo, errChan <-chan error) (string, error) {
```
uri200: This function is small enough to [do only one thing](https://alvinalexander.com/programming/clean-code-quotes-robert-c-martin/#:~:text=As%20you've%20probably%20read,that%20object%2C%20but%20not%20both.): wait

The caller should just use that function to wait for the server to be ready

Once it is ready the server can do any other stuff like checking `e.ListenerAddr()`

So lets make it to return just error
```
func waitForTestServer(e *echo.Echo, errChan <-chan error) error
```

feg/gateway/sbi/sbi_test.go
```diff
@@ -0,0 +1,165 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+	defer cancel()
+
+	ticker := time.NewTicker(5 * time.Millisecond)
+	defer ticker.Stop()
+
+	for {
```
uri200, line 148: see below how can you easily implement this



https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fsession_proxy%2Fservicers%2Fcredits.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=39&start_col=4&end_row=39&end_col=4

uri200, line 141: see [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fsession_proxy%2Fservicers%2Fcredits.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=39&start_col=4&end_row=39&end_col=4) how can you implement this

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: This is the error that will be returned when there is a graceful shutdown of the server. For example when the test case shuts down the server due to some other error.

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: Only with the help of e.ListernerAddr() we are determining that the server has come up successfully. So it is required within the waitForTestServer.

The link you have provides is for mixing Set and Get. Here it is Wait and Get. Wait and Get is a common coding practice. Even go channels use that pattern. May be I can change the method signature to waitForTestServerAndGetAddr() to make it more clear.

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: The context here provides a deadline for the loop. This code is taken from the test code of Echo server. See [here](https://github.com/labstack/echo/blob/0c4ad8646ad949d13a8d4ebe073185c1741b4d71/echo_test.go#L735).

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: We will still need a condition for ending the loop (setting a deadline). We will have to use a counter, to set a forced exit condition. I just see the use of context as another way of implementing this. As mentioned before this is the logic followed in the Echo server test/example code.

GANESH-WAVELABS:
@uri200 Please let me know which all comments are not resolved. I had pushed fixes for the comments expect for the last four, which I had asked for clarification or given clarification for the implementation. Older comments I had marked them as resolved as they are marked outdated and I couldn't update the comment expect for marking it as resolved.

uri200:
Hey @GANESH-WAVELABS  we normally let the reviewer mark comments as resolved. That give reviewer a chance for a final review. Sometimes we don't even close it

For example [this ](https://github.com/magma/magma/pull/9524/files#r745188667) comment below you marked it as resolved but it has not had any change. 

There are also other comments like [this ](https://github.com/magma/magma/pull/9524/files#r746975377) without a fix

Below you have the pictures of what I see on those comments. I don't see your comments.  Maybe you did the comments and never hit COMMENT on the top left corner at github. Remember that you need to hit comment at the end of the review or your comments will not be published


image



image

uri200:
Also here you have what I see when I try to comment an outdated comment. As you can see there is a `REPLY` bar where you can type your replies. You can also use the emoji with the thumb up to say you agree with that. Then we know you have gone through over it. See for example one review they did to me and check how we go through over it [here](https://github.com/magma/magma/pull/10182/files#r748559629)




image

GANESH-WAVELABS:
@uri Sorry, my bad I had not submitted the comments. I've now submitted the pending comments.

> For example this comment below you marked it as resolved but it has not had any change

I did make a change and reduced the number of blank lines. But left a blank space before the return
Screenshot from 2021-11-14 16-03-51
.

feg/gateway/sbi/sbi_test.go
uri200: Lets use  
```
	mockPcf.ListenAddr, err := waitForServerStart(eserver, errChan)
	assert.NoError(t, err)
	return mockPcf
```

There is no need of having `mockPcf.ListenAddr = addr` there.  

Note you will find this construction in go a lot.

(3 lines better than 4 lines  )

feg/gateway/sbi/sbi_test.go
uri200: Hey @GANESH-WAVELABS I see the problem: `e.ListenerAddr()` is not blocking so we need to retry it.

On the example I sent you the call was blocking. So we didn't need a loop

feg/gateway/sbi/sbi_test.go
uri200: Ok looks fine, lets leave it as it is.
I am still not sure we need that case. It looks like that case may be useful for echo testing but not for us. But lets leave it as it is

uri200 (changes requested):
make sure you do 
`./build.py -c` on your `/magma/feg./gateway` folder

That will possible correct format issues

You also need to rebase for that conflict

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: This results in the following error.
`sbi/sbi_test.go:127:2: expected identifier on left side of :=`

The variable err was not defined before.

uri200: remove the `:`


```
mockPcf.ListenAddr, err = waitForServerStart(eserver, errChan) 
assert.NoError(t, err) 
return mockPcf
```

uri200 (commented):
were you able to run `./build.py -c` on feg and commit any change with it?

uri200 (approved):

themarwhal:
Sorry about the delayed comments, just saw the changes made from this PR. We seem to have added a lot of generated files. I have a few asks:
1. Can we mark these files as generated in this file? https://github.com/magma/magma/blob/master/.gitattributes This will help with GitHub PR reviews as it will clearly state which files are generated. You might have noticed that PRs with large file changes cause a slowdown in the GitHub UI. :'(  
2. We should have a CI check to make sure the source template and the generated files are synched up. (Maybe we have this already, if we do just ignore this comment :) )

cc @GANESH-WAVELABS, @uri200

GANESH-WAVELABS:
> were you able to run `./build.py -c` on feg and commit any change with it?

Yes and there were no errors reported.

uri200:
Hey @GANESH-WAVELABS  can you raise s PR to include that directory as generated please?

GANESH-WAVELABS:
> Hey @GANESH-WAVELABS can you raise s PR to include that directory as generated please?

Raised this #10375

GANESH-WAVELABS:
@themarwhal 
> 2. We should have a CI check to make sure the source template and the generated files are synched up. (Maybe we have this already, if we do just ignore this comment :) )

The files are generated using ./orc8r/cloud/docker/build.py -g and hence I think we have that covered.

themarwhal:
> @themarwhal
> 
> > 1. We should have a CI check to make sure the source template and the generated files are synched up. (Maybe we have this already, if we do just ignore this comment :) )
> 
> The files are generated using ./orc8r/cloud/docker/build.py -g and hence I think we have that covered.

Great! thanks for checking :D

PR #9541, opened 2021-10-07
test(orc8r): add tests for checking entBlobStore -> sqlBlobStore migration

reginawang3495:
## Summary
Tests that SQL→ent application change won’t require a data migration

## Test Plan
Ran stated test in IntelliJ


orc8r/cloud/go/blobstore/ent_test.go
```diff
@@ -24,89 +25,231 @@ import (
...
+}
+
+func testAllFunctionalityGivenBlob1Saved(t *testing.T, fact blobstore.BlobStorageFactory) {
+	checkBlobs1SimpleFunctionality(t, fact)
+	checkRollbackOnBlobs1(t, fact)
+	checkModifyingFunctionsBlobs1(t, fact)
+	checkBlobs1SimpleFunctionality(t, fact)
+}
+
+func testMigrationsToDifferentBlobstores(t *testing.T, fact1 blobstore.BlobStorageFactory, fact2 blobstore.BlobStorageFactory) {
```
hcgatewood: This is short enough, let's just inline it

orc8r/cloud/go/blobstore/ent_test.go
```diff
@@ -24,89 +25,231 @@ import (
...
+	createOrUpdateBlobs1(t, fact1)
+	testAllFunctionalityGivenBlob1Saved(t, fact1)
+	testAllFunctionalityGivenBlob1Saved(t, fact2)
+}
+
+func TestSQLToEntMigration(t *testing.T) {
```
hcgatewood: General input: structure your file such that the "most important" or "top-level" functions are at the top -- this will generally mean putting the public functions first. Then, organize the remaining functions (often private) in roughly the order in which they're called

Another way of thinking about this is, you have some rough approximation of a call tree -- the order of your functions should correlate with that call tree, meaning layer 0 functions at the top, layer 1 functions in the middle, layer 2 functions near the bottom, etc

There's no way to be perfect about this, but approximating this flow will make reading your code feel more natural, like reading a book

orc8r/cloud/go/blobstore/ent_test.go
```diff
@@ -24,89 +25,231 @@ import (
...
+	// Spin up DB
```
hcgatewood: These types of "topic sentence" comments are a bad code smell. Instead, either omit them, or move the block into a function and use the comment as the function name. In this particular case, you can just omit the comment

Caveat 1: it is helpful to use these comments *during* development. But the final, checked-in code should generally be free of them

Caveat 2: in test code, providing a description of "what you're testing" can be a useful application of these topic sentences. So e.g. `// Ensure GetMany works` is fine, since it summarizes the intent of the block, but `// Spin up DB` is unnecessary since it's clear from skimming the next two lines

orc8r/cloud/go/blobstore/ent_test.go
```diff
@@ -24,89 +25,231 @@ import (
...
+	assert.ElementsMatch(t, []string{blobPair5.tk.Key}, keys)
+	err = store.Commit()
 
-	blob, err = storev2.Get("id1", storage.TK{Type: "type2", Key: "key2"})
+	// Check Rollback on committed transaction causes error
+	err = store.Rollback()
+	require.Error(t, err)
+}
+
+/**
```
hcgatewood: Style preference is single-line comments (`//`) rather than java-style multiline

orc8r/cloud/go/blobstore/ent_test.go
```diff
@@ -24,89 +25,231 @@ import (
...
 
-func TestMigration(t *testing.T) {
-	db, err := sqorc.Open("sqlite3", ":memory:")
+// Setup Blob Details
+type Pair struct {
+	tk   storage.TK
+	blob blobstore.Blob
+}
+
+var blobPair1 = Pair{
```
hcgatewood: Personally I don't like declaring package-level variables in your test files. This causes issues with variable name clashes when you add other test files in the same pkg. Instead, declare these directly in the function where they're needed

hcgatewood (changes requested):
I think we're moving in the right direction. Here's what I had in mind for the top-level

- TestEntToToSQLMigration, TestSQLToEntMigration
- writeBlobs, readExistingBlobs

Where each migration test does the following
- writeBlobs with impl A
- readExistingBlobs with impl A
- readExistingBlobs with impl B

And writeBlobs tests all the write-based interface methods, while readExistingBlobs tests all the read-based interface methods

I think this way of structuring the tests, plus the comments I added below, will make for a more concise migration test

Also down to talk through any of the individual points, either in Slack or in vc, just lmk!

hcgatewood (changes requested):
Edit: github being buggy. Ignore this requested change in favor of the above^

hcgatewood:
P.S. looks like the cloud linter job is failing. To fix this, you'll want to add gofmt as an on-save job to intellij https://stackoverflow.com/questions/33774950/execute-gofmt-on-file-save-in-intellij

The easy solution here is copy the watcher file from .intellij.shared to .intellij https://github.com/magma/magma/tree/ce40d2585f9de0a7a7b66a4c88aa9d4b2815a2ed/.idea.shared

orc8r/cloud/go/blobstore/ent_test.go
```diff
@@ -24,94 +25,208 @@ import (
 	magmaerrors "magma/orc8r/lib/go/errors"
 )
 
-func TestMigration(t *testing.T) {
+func TestSQLToEntMigration(t *testing.T) {
```
hcgatewood: I think we can actually just have 1 `TestBlobstoreImplMigrations`, then run `checkBlobstoreMigrations` twice in that func (or use a sub-test with `t.Run`)

orc8r/cloud/go/blobstore/ent_test.go
```diff
@@ -24,94 +25,208 @@ import (
...
+	var blobNotInFact = blobstore.Blob{Type: "notSaved", Key: "notSavedKey", Value: []byte("notSavedValue")}
+
+	createBlobs(t, fact1, networkID, blobs)
+	checkReadBlobs(t, fact1, networkID, blobs, blobNotInFact)
+	checkReadBlobs(t, fact2, networkID, blobs, blobNotInFact)
+	checkWriteBlobs(t, fact2, networkID, blobs, blobNotInFact)
+}
 
-	err = storev1.Commit()
+// Assumes that expectedBlobs length is at least 2.
```
hcgatewood, line 76: Docstrings should be prefixed with the identifier -- so e.g.
```suggestion
// checkReadBlobs does xyz.
// Assumes that expectedBlobs length is at least 2.
```

orc8r/cloud/go/blobstore/ent_test.go
```diff
@@ -24,94 +25,208 @@ import (
...
+	}
+
+	blobSlice := expectedBlobs[0:2]
+	blobs, err := store.GetMany(networkID, getTKsFromBlobs(blobSlice))
 	require.NoError(t, err)
-	blobs, err = storev2.GetMany("id1", storage.TKs{
-		{Type: "type1", Key: "key1"},
-		{Type: "type2", Key: "key2"},
-	})
+	assert.ElementsMatch(t, blobs, blobSlice)
```
hcgatewood: Throughout: use `want`/`got` style, either as want/got or as e.g. `expectedBlobs`/`actualBlobs`

orc8r/cloud/go/blobstore/ent_test.go
```diff
@@ -24,94 +25,208 @@ import (
...
-		{Type: "type3", Key: "key1", Value: []byte("value")},
+	err = store.CreateOrUpdate(networkID, blobstore.Blobs{
+		expectedBlobs[1],
+		blobNotInFact,
 	})
 	require.NoError(t, err)
-	blob, err = storev2.Get("id1", storage.TK{Type: "type3", Key: "key1"})
+
+	checkBlobInStore(t, store, networkID, expectedBlobs[1])
+	checkBlobInStore(t, store, networkID, blobNotInFact)
```
hcgatewood, line 153: Feels like this should fail ?

orc8r/cloud/go/blobstore/ent_test.go
```diff
@@ -24,94 +25,208 @@ import (
...
+}
+
+func checkBlobInStore(
+	t *testing.T,
+	store blobstore.TransactionalBlobStorage,
+	networkID string,
+	expectedBlob blobstore.Blob,
+) {
+	blob, err := store.Get(networkID, getTKFromBlob(expectedBlob))
+	require.NoError(t, err)
```
hcgatewood: I know this file already used require, but an opportunity to improve the file is to use assert rather than require. Difference is require fails immediately. Assert is preferred cause it's easier to understand what happened in CI

orc8r/cloud/go/blobstore/ent_test.go
```diff
@@ -24,94 +25,208 @@ import (
...
+	require.Equal(t, expectedBlob, blob)
+}
+
+func getTKsFromBlobs(blobs blobstore.Blobs) storage.TKs {
```
hcgatewood: I think these 2 helper funcs already exist in the codebase

hcgatewood (changes requested):
Okay this is looking great 👍 Couple nits and one q on the checkBlobInStore seems like it should fail (?), other than that I think we're gtg

orc8r/cloud/go/blobstore/ent_test.go
reginawang3495: Not really, can see L152-155. This function checks that writing works, so we first check that `CreateOrUpdate` wrote `blobNotInFact` properly

hcgatewood (changes requested):
Edit: disregard, misread the test 🙂 

---

Okay, I see where I was confused for checkBlobInStore

1. checkWriteBlobs *does* have a side effect -- namely, it adds blobNotInFact into the fact. So let's fix that doc comment
2. Add a comment above where blobNotInFact gets added in, to draw a bit of extra attention to the fact that we are, in fact, mutating the store to make sure writes still succeed

Other than that lgtm, msg when ready and I'll accept + merge

hcgatewood (approved):

hcgatewood:
Force-merging to override borked unrelated test

Issue #8834, opened 2021-08-27
[AGW][vN][S1AP] SCTP / eNB association


markjen:
- Tee SCTP traffic from the Proxy{Uplink,Downlink}Servers to a SCTP manager
- Create S1APStateManager that manages a S1apState in lte/protos/oai/s1ap_state.proto
  - We will likely want to create a GRPC service for this
- Reimplement s1ap_handle_new_association and s1ap_handle_sctp_disconnection from s1ap_mme_handlers.c
- enb state can be represented as EnbDescription from lte/protos/oai/s1ap_state.proto

PR #9566, opened 2021-10-07
feat(agw): Scaffold S1AP state manager proto service.

mstre123:

## Summary
Implement and generate protos for s1ap state manager grpc service.
#8834


lte/protos/s1ap_state_manager.proto
```diff
@@ -0,0 +1,41 @@
+/*
+Copyright 2020 The Magma Authors.
```
markjen: nit: 2021 :)

lte/protos/s1ap_state_manager.proto
```diff
@@ -0,0 +1,41 @@
...
+import "lte/protos/oai/s1ap_state.proto";
+
+package magma.lte;
+
+option go_package = "magma/lte/cloud/go/protos";
+
+service S1apStateManagerService {
+  // Connect handles an S1AP Connection Request
+  rpc Connect (ConnectRequest) returns (ConnectResponse) {}
+  // Disconnect handles an S1AP DisconnectRequest
```
mstre123: typo

lte/protos/s1ap_state_manager.proto
```diff
@@ -0,0 +1,41 @@
...
+  rpc Disconnect (DisconnectRequest) returns (DisconnectResponse) {}
+}
+
+message ConnectRequest {
+  oai.S1apState state = 1;
+}
+
+message ConnectResponse{
+  uint32 associd = 1; // s1ap association id
```
themarwhal: this is more just my preference but can we place the comments above the field? That's what the uber style guide for proto seems to use: https://docs.magmacore.org/docs/next/contributing/contribute_conventions#grpc :) 
```suggestion
  // s1ap association id
  uint32 associd = 1;
```

mstre123: Haha I actually prefer them that way!
I was following the pattern in the proto i was importing https://github.com/magma/magma/blob/6e862525c76c143ea07dbf18d1e1c23655c4d64b/lte/protos/oai/s1ap_state.proto#L58
but will move the comments up.

themarwhal: yeah I think we're definitely not consistent as ususal :D hopefully we can push to converge 😄

lte/protos/s1ap_state_manager.proto
```diff
@@ -0,0 +1,41 @@
...
+
+service S1apStateManagerService {
+  // Connect handles an S1AP Connection Request
+  rpc Connect (ConnectRequest) returns (ConnectResponse) {}
+  // Disconnect handles an S1AP DisconnectRequest
+  rpc Disconnect (DisconnectRequest) returns (DisconnectResponse) {}
+}
+
+message ConnectRequest {
+  oai.S1apState state = 1;
```
markjen: i don't think we'll want to send this entire thing to setup a single enb assoc

lte/protos/s1ap_state_manager.proto
markjen: i think assoc_id is usually snake_cased

lte/protos/s1ap_state_manager.proto
```diff
@@ -0,0 +1,41 @@
...
+limitations under the License.
+*/
+
+syntax = "proto3";
+
+import "lte/protos/oai/s1ap_state.proto";
+
+package magma.lte;
+
+option go_package = "magma/lte/cloud/go/protos";
```
markjen: do we need this option? can we go without?

markjen (changes requested):
consider putting this in /src/protos since it's a new service and unlikely to be used by other stuff in /lte/gateway

lte/protos/s1ap_state_manager.proto
mstre123: good catch - been copy and pasting that from other files - will set auto heading in my ide

lte/protos/s1ap_state_manager.proto
mstre123: yeah i copied that from from S1apState
https://github.com/magma/magma/blob/6e862525c76c143ea07dbf18d1e1c23655c4d64b/lte/protos/oai/s1ap_state.proto#L59
but ill fix move it to camel case

lte/protos/s1ap_state_manager.proto
mstre123: was following the pattern we've used in other proto files.
i think the style guide prefers this to just defining it in the protoc call
https://developers.google.com/protocol-buffers/docs/reference/go-generated#package

lte/protos/s1ap_state_manager.proto
mstre123: changed to just store the id

PR #9592, opened 2021-10-08
fix(agw): Fix bootstrapper when certs rejected at app level

### LINKED ISSUE #8369: stability: When receive error that client certificate not recognized, should re-bootstrap
Context

* Seems like AGW bootstrapper service can enter into an unhealthy state under certain network conditions
* Related issue: AGW bootstrapper only determines the need to re-bootstrap based on the contents of its session cert. However, we also want to support revoking that cert at the Orc8r (not least for testing purposes, but also for e.g. for blowing away certs at orc8r during support sessions) and have the AGW bootstrapper gracefully handle this situation

Resolution

1. Clean up existing AGW bootstrapper service: instead of sprinkling “schedule next bootstrap” throughout orc8r/gateway/python/magma/magmad/bootstrap_manager.py, let’s just always perform a “check and renew if necessary” once every ~30 seconds. This will drastically simplify the logic, and allow us to manually verify, from the `_run` function, that we are always re-scheduling the next bootstrap attempt
2. Lightly refactor `checkin_cli.py` such that its functionality can be used from the AGW bootstrapper service. Then, instead of only checking that the gw’s session certs are valid, we also check that the Orc8r accepts the session certs. So if (a) the Orc8r is reachable but (b) a TLS connection is rejected, then AGW bootstrapper initiates a re-bootstrap
3. AGW should accept certs from Orc8r even if they’re “not valid” based on the AGW’s view of the time — yes Orc8r could send bad certs, but rejecting these bad certs gains AGW a couple extra hours of connectivity; accepting these bad certs keeps us from AGWs failing under extreme time skew / NTP issues. Ref: https://github.com/magma/magma/blob/1dd65f2a335332c7356aa45012eabb0e75d3db8e/orc8r/gateway/python/magma/magmad/bootstrap_manager.py#L375-L399

hcgatewood:
## Summary

Fixes https://github.com/magma/magma/issues/8369

Note: started down the path of a large refactor, but testing in Python -- especially the way we set up our dev vm -- is needlessly difficult. And since this is an essential service, we'll be safe and default to just a small change.

## Test Plan

Manual

- AGW connects to controller
- Delete that AGW's cert from controller's cert store
- Observe AGW requests getting rejected
- Observe AGW re-bootstrapping and requests subsequently succeeding

## Additional Information

### DIFF
--- a/orc8r/gateway/python/magma/magmad/bootstrap_manager.py
+++ b/orc8r/gateway/python/magma/magmad/bootstrap_manager.py
@@ -115,6 +115,7 @@ class BootstrapManager(SDWatchdogTask):
         """
         if self._state is BootstrapState.BOOTSTRAPPING:
             return
+        self._state = BootstrapState.SCHEDULED_BOOTSTRAP
         await self.wake_up()
 
     def _maybe_create_challenge_key(self):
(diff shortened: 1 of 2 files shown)


ardzoht (approved):
lgtm, thanks for the fix

andreilee (approved):

PR #9599, opened 2021-10-09
docs(agw): Updated AGW Install to remove username & password info

jaredmullane:

## Summary

Remove recommended configuration information from AGW install steps (username, password, etc.) as a security concern. Users recommended to enter parameters consistent with their network and best practices. 

## Test Plan

Screen Shot 2021-10-08 at 5 18 48 PM

### DIFF
--- a/docs/readmes/lte/deploy_install.md
+++ b/docs/readmes/lte/deploy_install.md
@@ -25,26 +25,14 @@ satisfies the following requirements:
 
 ### 1. Create boot USB stick and install Ubuntu on your AGW host
 
-- Download Ubuntu Server 20.04 LTS .iso image from Ubuntu website
+- Download the Ubuntu Server 20.04 LTS .iso image from the Ubuntu website
 - Create bootable usb using etcher [tutorial here](https://tutorials.ubuntu.com/tutorial/tutorial-create-a-usb-stick-on-macos#0)
 - Boot your AGW host from USB
   (Press F11 to select boot sequence, :warning: This might be different for
   your machine). If you see 2 options to boot from USB, select the non-UEFI
   option.
-- Select "Install" option.
-- Network missing firmware "No"
-- Primary network interface "enp1s0"
-- Configuration.
-    - hostname: `<choose host name>`. Default is "magma". You need to select unique
-    host name in case of multiple AGW in network.
-    - Domain name : ""
-    - Root password: "magma"
-    - Username : "magma"
-    - Password: "magma"
-    - Partition disk: "Use entire disk"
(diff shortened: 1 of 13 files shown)


docs/readmes/lte/deploy_install.md
```diff
@@ -25,26 +25,14 @@ satisfies the following requirements:
...
-    - Username : "magma"
-    - Password: "magma"
-    - Partition disk: "Use entire disk"
-    - Select disk to partition: "sda"
-    - Partitioning scheme: "All files in one partition"
-    - Only tick ssh server and utilities (untick every other)
+- Install and configure you access gateway according to your network defaults.
+    - Make sure to enable ssh server and utilities (untick every other)
 - Connect your SGi interface to the internet and select this port during the
 installation process to get an IP using DHCP.
```
pshelar, line 38: I see couple of instances remaining:
docs/docusaurus/versioned_docs/version-1.0.0/lte/setup_baremetal.md
docs/readmes/lte/setup_baremetal.md

otherwise looks good.

pshelar (approved):

docs/readmes/lte/deploy_install.md
jaredmullane: @pshelar Good catch! Updated.

hcgatewood (commented):
@jaredmullane why are we updating the versioned docs -- is this something that broke for the debian installs of 1.0.0, for example? If no, then we only need to update for master and maybe 1.6.0

Lmk 🙂

jaredmullane:
@hcgatewood I was just thinking since it was a security risk to just fix it for all of them in case a partner was on a previous version or setting up Magma on another version for that edge case.

hcgatewood:
@jaredmullane gotcha. I guess the question is: will the described change work for all previous versions of Magma, both on ubuntu and debian? Maybe @ardzoht can clue me in / I'll defer to his review

hcgatewood (approved):
Or, I suppose the idea is we're telling installers to choose their own usernames/passwords etc. If that's the case, ez accept. Will approve under that assumption

ardzoht:
> @jaredmullane gotcha. I guess the question is: will the described change work for all previous versions of Magma, both on ubuntu and debian? Maybe @ardzoht can clue me in / I'll defer to his review

There should not be any dependency on previous versions for this change too, so I think it's good to update all.

jaredmullane:
@hcgatewood yeah, the intent was just for users to pick their own usernames/passwords rather than the default magma/magma which can be a security risk.

PR #9635, opened 2021-10-12
fix(agw): adds cleanup of stray unix domain socket to agwd

electronjoe:
## Summary

This PR adds more graceful handling of dangling / leftover unix domain socket handling by the Golang `agwd` binary.

- If an existing socket exists at the specified path
  - But a Dial attempt indicates no Listener -> Log warning and delete old sock
  - But a Dial attempt indicates a Listener -> Return error (ultimately panic)

## Test Plan

Authored new unit tests to cover the majority of the code path for socket creation.

### DIFF
--- a/src/go/agwd/server/server.go
+++ b/src/go/agwd/server/server.go
@@ -12,7 +12,11 @@
 package server
 
 import (
+	"fmt"
 	"net"
+	"os"
+	"runtime"
+	"time"
 
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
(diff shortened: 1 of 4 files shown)


electronjoe:
@markjen first pass looking at this failing unit test I'm concerned about portability of the "Dial to see if existing Listener is on the sock" approach. See https://github.com/magma/magma/pull/9635/checks?check_run_id=3868996715. Thoughts?

electronjoe:
@mstre123 FYI

src/go/agwd/server/server.go
```diff
@@ -41,10 +44,40 @@ func newServiceRouter(cfgr config.Configer) service.Router {
 	)
 }
 
+func cleanupUnixSocket(logger log.Logger, path string) error {
+	if _, err := os.Stat(path); err == nil {
```
markjen: nit: i wonder if the construction would be cleaner like this:

```
_, err := os.Stat(Path)
if os.IsNotExist(err) {
  return nil
}
if err != nil {
  return errors.Wrap(...)
}
// do connect check
```

src/go/agwd/server/server.go
```diff
@@ -41,10 +44,40 @@ func newServiceRouter(cfgr config.Configer) service.Router {
 	)
 }
 
+func cleanupUnixSocket(logger log.Logger, path string) error {
```
markjen: if we want to do more controlled testing, we can pass in fn ptrs for os.Stat and net.DialTimeout

markjen (approved):

markjen:
> @markjen first pass looking at this failing unit test I'm concerned about portability of the "Dial to see if existing Listener is on the sock" approach. See https://github.com/magma/magma/pull/9635/checks?check_run_id=3868996715. Thoughts?

windows support might be a bit more involved here due to the way file handling is done.

for now we could just fail if `runtime.GOOS == "windows"` in `startSctpdDownlinkServer` if `target.Scheme == "unix"` and leave a TODO

src/go/agwd/server/server.go
electronjoe: Much better readability, thanks

src/go/agwd/server/server.go
electronjoe: Nice. Done.

src/go/agwd/server/server.go
```diff
@@ -41,10 +45,51 @@ func newServiceRouter(cfgr config.Configer) service.Router {
 	)
 }
 
+func cleanupUnixSocket(
```
markjen, line 48: personal style pref:

```
func cleanupUnixSocket(
	logger log.Logger,
	osStat func(string)(os.FileInfo,error), 
	netDialTimeout func(network, address string, timeout time.Duration) (net.Conn, error),
	path string,
) error {
	_, err := osStat(path)
```

but anything `gofmt` is OK with, I am as well! :)

src/go/agwd/server/server_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+// TestCleanupExistingSock ensures that should an un-used existing socket exist, we clean it up
+// and log a warning.
+func TestCleanupExistingSock(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	if runtime.GOOS == "windows" {
```
markjen: can also have a `//+build !windows` and `//+build windows` test files (the latter allows us to validate the correct err message return for `runtime.GOOS` in server.go#62)

markjen (approved):

src/go/agwd/server/server.go
electronjoe: Turns out gofmt is not working in my vs code devcontainer. It was... but stopped. But that could be because I'm mid way through testing fixups to go tooling in devcontainer.  Ran gofmt manually, thanks.

src/go/agwd/server/server_test.go
electronjoe: Today I learned... Gazelle did some interesting things here but I guess they are OK (tests pass) - lots of select statements based on OS - yet all the results of select are identical.

PR #9653, opened 2021-10-13
Revert "fix(mme): memory leak for congestion based attach rejects"

ulaskozat:
Reverts magma/magma#9641


ardzoht (approved):

ulaskozat:
since this is reverted PR on github directly, I will ignore DCO and semantics checks.

PR #9669, opened 2021-10-14
fix(agw): AGW containerization fixes

rmeleromira:
Signed-off-by: Ramon Melero <ramonmelero@fb.com>

## Summary

These are fixes needed for containerized AGW


lte/gateway/deploy/roles/magma_deploy/tasks/main.yml
```diff
@@ -17,6 +17,18 @@
   tags:
     - agwc
 
+- name: Ensure ca-certificates is up to date
+  become: yes
+  apt:
+    name: "{{ packages }}"
+    only_upgrade: yes
```
electronjoe: Here and elsewhere below - per yamllint, could you use `true` instead of `yes`?

lte/gateway/deploy/roles/magma_deploy/tasks/main.yml
```diff
@@ -175,6 +195,44 @@
...
 
+- name: Get current DNS server from systemd-resolvd
+  shell: grep nameserver /run/systemd/resolve/resolv.conf
+  register: dns
+  tags:
+    - agwc
+
+- name: Disable systemd-resolved
+  systemd:
+    name: systemd-resolved
```
arunuke: Why do we need to stop using systemd-resolved and hardcode the dns value?

lte/gateway/deploy/roles/magma_deploy/tasks/main.yml
ymasmoudi: be careful here, ansible expect yes I think

lte/gateway/deploy/roles/magma_deploy/tasks/main.yml
ymasmoudi: @rmeleromira adding a tag agwc will not prevent these tasks from running when installing normal agw.
in addition, all these changes to systemd-resolved are needed for aws only. We might want to not impact the default installation.

ymasmoudi (changes requested):

lte/gateway/deploy/roles/magma_deploy/tasks/main.yml
rmeleromira: I thought at one point Youssef said that we needed to replace systemd-reolved, so I tried this method. Rather than selectively do these steps based on being on aws, I did a little more digging. It seems that it's the ipfix0 interface that is causing issues. Repo instructions here

https://gist.github.com/rmeleromira/ecac9790cf8fc7c59c07ebd4e983c351#file-test-md

There's a vagrantfile that reproduces the issue. 

```
# start vm using Vagrantfile
vagrant up
vagrant ssh

# test broken dns
ping google.com

# bring down ipfix0 and restart dns
ifdown ipfix0
systemctl restart systemd-resolved

# test and it works
ping google.com
```
Issue with this is I'm not sure how to fix this, so I'll remove the DNS changes I made and leave the rest.

Here's a run through of the whole repro

https://gist.github.com/rmeleromira/163b4ff6ecb31b16685b9b4c39ee8713#file-gistfile1-txt-L755-L768

lte/gateway/deploy/roles/magma_deploy/tasks/main.yml
rmeleromira: Fixed. Seems on with true

```
ok: [127.0.0.1] => {
    "cache_update_time": 1634612673,
    "cache_updated": false,
    "changed": false,
    "invocation": {
        "module_args": {
            "name": [
                "ca-certificates"
            ],
            "only_upgrade": true,
```

ymasmoudi (approved):
please rebase and fix the conflicts

arunuke (approved):
LGTM. Let's track the DNS issue as a separate one since there is a manual workaround to it if the problem occurs.

lte/gateway/docker/readme.md
```diff
@@ -60,7 +60,24 @@ growpart /dev/xvda 1; resize2fs /dev/xvda1
 
 Create/deploy an ubuntu 20.04 (latest version) server/instance with 2 interfaces (S1, Sgi) and the appropriate resources for your use case.
 
-Your interfaces should not be named eth0 or eth1, they will be changed to that. eth0 should be your Sgi interface, and eth1 should be S1 interface.
+Your interfaces should be named eth0 and eth1. eth0 should be your Sgi interface, and eth1 should be S1 interface.
+
+If your interfaces are not named eth0 and eth1, you will have to use netplan to rename your interfaces and reboot
+```
+cat /etc/netplan/50-cloud-init.yaml
```
pshelar: we don't support netplan, it would confict with OVS bridge interface definitions. Can you use interface file for physical inerfaces/

pshelar (changes requested):

lte/gateway/docker/readme.md
rmeleromira: The current deploy tasks for containerized AGW leaves netplan in place if it's there and netpan comes with ubuntu 20. Should we add some tasks to disable netplan? Or I can just change this part of the documentation to udev rules which will accomplish the renaming without using netplan. Or leave out the example all together. I just put it in there cause I figured almost everyone will need to change their interface names and would be nice to supply an example.

pshelar: its done during agw deployment:
https://github.com/magma/magma/blob/2635f924633f0245b47420294257f80bc8007c4f/lte/gateway/deploy/agw_install_ubuntu.sh#L118

rmeleromira: I see. Package agw script enables networking service and purges netplan, but not containerized atm. I'll bring this up to Youssef to see how we can handle it.

why is the eth1 hard coded in [agw_install_ubuntu.sh](https://github.com/magma/magma/blob/master/lte/gateway/deploy/agw_install_ubuntu.sh#L113-L116)? 

It figures out the ip info for eth0 but not eth1, not sure why

arunuke: A key point to remember that the containerized AGW is different from the packaged AGW because of it's expected independence of the host. The packaged AGW is built to run on Ubuntu while the objective of the containerized AGW is to run across hosts which is why the configuration of the host itself is separated from the starting up of the containers.

lte/gateway/deploy/roles/magma_deploy/tasks/main.yml
```diff
@@ -17,6 +17,18 @@
   tags:
     - agwc
 
+- name: Ensure ca-certificates is up to date
+  become: yes
```
electronjoe: Can we use the yamlint recommendation of `become: true`?

lte/gateway/deploy/roles/magma_deploy/tasks/main.yml
electronjoe: Can we use the yamlint advice of `only_upgrade: true`?

lte/gateway/docker/publish.sh
```diff
@@ -14,8 +14,8 @@ REGISTRY=$1
 VERSION=latest
 
 publish (){
-  docker tag agw_gateway_$1:latest "${REGISTRY}"gateway_$1:"${VERSION}"
-  docker push "${REGISTRY}"gateway_$1:"${VERSION}"
+  docker tag agw_gateway_$1:latest "${REGISTRY}"agw_gateway_$1:"${VERSION}"
```
electronjoe: Is it possible to address the Shellcheck findings here and below?

electronjoe (approved):

lte/gateway/docker/readme.md
pshelar: > I see. Package agw script enables networking service and purges netplan, but not containerized atm. I'll bring this up to Youssef to see how we can handle it.
> 
> why is the eth1 hard coded in [agw_install_ubuntu.sh](https://github.com/magma/magma/blob/master/lte/gateway/deploy/agw_install_ubuntu.sh#L113-L116)?
> 
> It figures out the ip info for eth0 but not eth1, not sure why

eth1 (S1) is typically private network, so we recommend this specific network.

pshelar: > A key point to remember that the containerized AGW is different from the packaged AGW because of it's expected independence of the host. The packaged AGW is built to run on Ubuntu while the objective of the containerized AGW is to run across hosts which is why the configuration of the host itself is separated from the starting up of the containers.

I agree with this in medium term, but today we do make use of host networking for containers, that creates dependency on the host network configuration.
For now we need to move our OVS bridge configuration to netplan based config to have single/consistent configuration across all host devices.

arunuke: Agree. Although the ideal goal is full independence, since we have some tight dependencies today (as you highlight above) the hope is to keep them as minimal as possible to essential elements (like OVS running on the base host). If it works, we can experimentally pilot the addition of netplan support for OVS in the container AGW effort since this is still in alpha without impacting the package based version which is the currently prod supported version.

pshelar (approved):
as Arun mentioned we will continue to make progress to netplan based config for containerized agw.

lte/gateway/docker/readme.md
rmeleromira: I've removed the section mentioning netplan, this should resolve this for this PR.

PR #9671, opened 2021-10-14
feat(orc8r): add indexRemove to indexers interface

### LINKED ISSUE #8925: feat(orc8r): reconcile secondary indexes with state

```
  count
---------
 2872606
```

```
 count
-------
  2977
```

**Possible solutions**

uri200:

## Summary

This PR modifies protos to include `IndexRemove` and includes end points on all the required implementation of Indexer interface. 

It also modify the function that handles Index to include IndexRemove by the Indexer.

## Test Plan

```
./build.py -g
./build.py -c
```

## Additional Information


orc8r/cloud/go/services/state/indexer/indexer.go
```diff
@@ -11,6 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+//go:generate bash -c "mockery --name=Indexer --note='Run ./build -g at Orc8r to re-generate'"
```
hcgatewood: Nit
```suggestion
//go:generate bash -c "mockery --name=Indexer --note='Run ./build --generate at Orc8r to re-generate'"
```

orc8r/cloud/go/services/state/protos/indexer.proto
```diff
@@ -24,6 +24,9 @@ service Indexer {
   // Index a set of states by forwarding to locally-registered indexers.
   rpc Index(IndexRequest) returns (IndexResponse) {}
 
+  // IndexRemove removes a set of states by forwarding to locally-registered indexers.
+  rpc IndexRemove(IndexRequest) returns (IndexResponse) {}
```
hcgatewood: Preferred pattern is to have a unique Request, Response per rpc -- so `IndexRemove(IndexRemoveRequest) ...`

orc8r/cloud/go/services/state/test_init/test_indexer_servicer.go
```diff
@@ -5,7 +5,7 @@ This source code is licensed under the BSD-style license found in the
 LICENSE file in the root directory of this source tree.
 
 Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
+an
```
hcgatewood: Typo?

orc8r/cloud/go/services/state/indexer/index/index.go
```diff
@@ -105,25 +140,43 @@ func Index(networkID string, states state_types.SerializedStatesByID) ([]error,
 	return indexErrs, nil
 }
 
-func indexOne(networkID string, idx indexer.Indexer, states state_types.SerializedStatesByID) error {
-	filtered := states.Filter(idx.GetTypes()...)
-	if len(filtered) == 0 {
+func indexerSingleAction(networkID string, idx indexer.Indexer, states state_types.SerializedStatesByID, actionType int) error {
```
hcgatewood: Nit: function names should start with a verb. In this case, I'm fine with leaving it as `indexOne` and having the action type in the fn signature make it clear that this could be either index or deindex

hcgatewood (changes requested):
Lgtm, couple nits then good to go

As an optional aside: I actually like `DeIndex` better than `IndexRemove`, to me it's more self-descriptive. But up to you 🙂

orc8r/cloud/go/services/state/protos/indexer.proto
uri200: yep forgot that

orc8r/cloud/go/services/state/test_init/test_indexer_servicer.go
uri200: vim command 😅

lte/gateway/python/integ_tests/federated_tests/configs/service_registry.yml
```diff
@@ -44,7 +44,7 @@ services:
     port: 9101
   feg_hello:
     ip_address: 127.0.0.1
-    port: 9093
+    port: 9094
```
uri200, line 47: changed port on feg_hello service to avoid collision with metrics service on orc8r when running feg and or8cr on the same machine

ardzoht (approved):
agw changes lgtm

docs/readmes/orc8r/dev_security.md
```diff
@@ -171,6 +171,9 @@ Debug southbound interface (gRPC)
 # Places generated session secrets at /tmp/magma_protos/gateway.{key,crt}
 ${MAGMA_ROOT}/orc8r/tools/scripts/bootstrap.bash YOUR_PROVISIONED_GATEWAY_HWID
 
+# You may need to Generate compiled proto definitions (see Miscellaneous
+commands)
```
hcgatewood, line 175: Comment?

orc8r/cloud/go/services/state/indexer/indexer.go
```diff
@@ -11,6 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+//go:generate bash -c "mockery --name=Indexer --note='Run ./build --generate at Orc8r to re-generate'"
```
hcgatewood, line 14: Nit 
```suggestion
//go:generate bash -c "mockery --name=Indexer --note='Run build --generate at Orc8r to re-generate'"
```

hcgatewood (approved):
Lgtm!

orc8r/cloud/go/services/state/indexer/index/index_test.go
```diff
@@ -91,6 +94,16 @@ func TestIndexImpl_HappyPath(t *testing.T) {
 	assert.Contains(t, e, iid1)
 	assert.Contains(t, e, index.ErrIndex)
 	assert.Contains(t, e, someErr.Error())
+
```
electronjoe, line 97: Drive-by nit: this unit test is getting pretty large - a failure may increasingly become harder to differentiate the cause (which part of happy path is broken).  At some point we probably want to break the happy path up into step-focused happy path sub-tests?

electronjoe:
Thanks for implementing this functionality, @uri200

PR #9682, opened 2021-10-14
feat(ci): new ci for pre and post merge

quentinDERORY:

## Summary

Pre merge pipeline:
- deploy build to test registries when there is a change (currently pointing to CI, will change when test registries are created) 
- Test executes in parallel of build
- Integ test (CWF & LTE) on demand

Post merge pipeline:
- Always deploy build to ci registries  (no path filtering as makes it harder to deploy latest of master to an environment if only one part of magma is on ci-registries)
- Test executes in parallel of build
- Integ test (CWF & LTE) always run

Build of agw has been migrated to github actions.
CWF integ tests run on two runners due to poor performance of docker-compose on MacOS.

Will later improve tests result display when building the full post build heavy testing pipeline with firebase.

## Test Plan

Tested on Forked repo

## Additional Information

Will later improve the way the workflows file hierarchy, but we will no be able to achieve one file per pre/post merge due to permissions in forked repo.
And current reusability of workflow is showing limitations.

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -5,10 +5,11 @@ on:  # yamllint disable-line rule:truthy
   push:
     branches:
       - master
+      - v1.*
   pull_request:
     branches:
       - master
-      - 'v1.*'
+      - v1.*
     types: [opened, reopened, synchronize]
 
 env:
@@ -29,6 +30,17 @@ jobs:
           filters: |
             filesChanged:
               - [".github/workflows/agw-workflow.yml", "orc8r/**", "lte/**"]
+      # Need to save metadata for deploy from PR job
+      - name: Save PR number and trigger type
+        run: |
+          mkdir -p ./pr
+          echo ${{ github.event.number }} > ./pr/NR
+          echo ${{ !steps.changes.outputs.filesChanged }} > ./pr/skipped
+      - name: upload metadata
+        uses: actions/upload-artifact@v2
+        with:
+          name: pr
+          path: pr/
   lte-test:
     needs: path_filter
     if: ${{ needs.path_filter.outputs.should_not_skip == 'true' }}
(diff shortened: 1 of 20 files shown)


.github/workflows/agw-workflow.yml
```diff
@@ -5,10 +5,11 @@ on:  # yamllint disable-line rule:truthy
   push:
     branches:
       - master
+      - v1.*
   pull_request:
     branches:
       - master
-      - 'v1.*'
+      - v1.*
```
tmdzk, line 12: at some point this won't work anymore, we'll need to come up with a solution, lat time I tried `v*.*` it didnt' work

.github/workflows/agw-workflow.yml
```diff
@@ -29,6 +30,17 @@ jobs:
           filters: |
             filesChanged:
               - [".github/workflows/agw-workflow.yml", "orc8r/**", "lte/**"]
+      # Need to save metadata for deploy from PR job
+      - name: Save PR number and trigger type
+        run: |
+          mkdir -p ./pr
+          echo ${{ github.event.number }} > ./pr/NR
```
tmdzk, line 37: let's store them in /tmp 
on top of that let's just do one file with all env var, so that we can just source the file and have it 

echo "PR_NUMBER=${{ github.event.number }}" >> /tmp/ci_vars

something like that 

then when needed we do `source /tmp/ci_vars`

tmdzk (changes requested):
overall looks good just 2 tiny comments

.github/workflows/agw-workflow.yml
quentinDERORY: I understand the intent but, the way artifacts are retrieved today is with javascript code being executed on the triggered workflow.
Changing it to a source file will make the workflow needing the information longer(still need js code to fetch artifacts and download it) and exporting variable in gha is a litlle bit trickier than regular export of variable.
All in all, it's possible but will imply a not tiny refactor for a not so better code.

.github/workflows/agw-workflow.yml
quentinDERORY: we can always add v2.* when needed and figure out a solution by then.

tmdzk (approved):

PR #9683, opened 2021-10-14
feat: add cc_asn1_library rule to generate .asn1 changes

### LINKED ISSUE #9715: bazel: Investigate how to generate files from .asn1 configurations with asn1c

themarwhal:

## Summary
This PR adds a custom logic for Magma to generate ASN1 files and make a library out of it. 
All flags were copied from https://github.com/magma/magma/blob/4ad87780c543b05baaf6511e987a225e4b5e1a97/lte/gateway/c/core/oai/tasks/ngap/CMakeLists.txt

Some notes on how I ended up with this solution.

### Why genrule doesn't work
using genrule to run asn1c does not work as genrule requires an explicit list of output files. As ASN1C generates ~ 100 files, I did not want to go this route. (which would involve listing out every single output file)

### Why I chose to write a custom rule
I found two main ways that people seem to get around this. The first is zipping up the generated files and declaring that as an output. Wasn't exactly sure how those files ended up being used though. The other option is to create a custom rule that uses `ctx.actions.declare_directory`. (See https://github.com/envoyproxy/envoy/blob/main/bazel/envoy_build_system.bzl#L74) I ended up going with the second approach as I could find more examples on sourcegraph that used `ctx.actions.declare_directory`.

This thread helped me figure out how to use CCInfo - https://groups.google.com/g/bazel-discuss/c/iH_Id2sCTe4/m/BvuHW7PVBAAJ

## How to use the new library definition
```
# in any BUILD file
cc_asn1_library(
     name = "asn1_r16",
     asn1_file = "messages/asn1/r16/r16.asn1",
     prefix = "Ngap_",
 )

cc_library(
   name = "...",
   srcs = [...],
   hdrs = [...],
   deps = [":asn1_r16"],
}
```

## Pending
1. figure out if compiling asn1c with bazel is an option, in order to increase determinism of asn1c use and reduce environmental dependencies by Bazel.

## Test Plan
After running `bazel clean` (not the full clean with --expunge)
Screen Shot 2021-10-14 at 3 19 52 PM

## Additional Information

### DIFF
--- a/experimental/bazel-base/Dockerfile
+++ b/experimental/bazel-base/Dockerfile
@@ -57,6 +57,15 @@ RUN git clone --branch v4.3 https://github.com/mfontanini/libtins.git && \
     # symlink to /usr/lib to match Magma VM
     ln -s /usr/local/lib/libtins.so /usr/lib/libtins.so
 
+# TODO(GH9710): Generate asn1c with Bazel - also this repo is really old :o 
+RUN git clone https://gitlab.eurecom.fr/oai/asn1c.git && \
+    cd asn1c && \
+    git checkout f12568d617dbf48497588f8e227d70388fa217c9 && \
+    autoreconf -iv && \
+    ./configure && \
+    make -j"$(nproc)" && \
+    make install
+
 # Update shared library configuration
 RUN ldconfig -v
 
--- a/tools/build_defs/asn1c.bzl
+++ b/tools/build_defs/asn1c.bzl
@@ -1,0 +1,185 @@
+# Copyright 2021 The Magma Authors.
+
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Custom library and macros to generate files with asn1c"""
+
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+def _get_dir_name(name):
+    # we need to postfix the directory name with .cc to trick Bazel into thinking this is a valid input
+    # Related GH issue: https://github.com/bazelbuild/bazel/issues/10552
+    return name + ".cc"
+
+def _contruct_substitution_commands(ctx, dir_path):
+    """Returns a list of commands that replaces certain strings with another. 
+
+    The replacement configuration is taken from ctx.attr.substitutions.
+    """
+    substitution_commands = []
+    substitution_template = "egrep -lRZ \'{before}\' {dir} | xargs --no-run-if-empty -0 -l sed -i -e \'s/{before}/{after}/g\'"
+    for before in ctx.attr.substitutions:
+        after = ctx.attr.substitutions[before]
+        substitution_commands.append(
+            substitution_template.format(
+                dir = dir_path,
+                before = before,
+                after = after,
+            ),
+        )
+    return substitution_commands
+
+def _constructor_file_filter_commands(ctx, dir_path):
+    """Returns a list of commands that removes any files that fit the condition described below.
+
+    1. any file that does not end with .c or .h
+    2. any file that is specified in ctx.attr.files_to_remove
+    """
+    filter_commands = []
+
+    # cc_library will complain if there are files that are not one of (.c, .cc, .cpp, .h, ...)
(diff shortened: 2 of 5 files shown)


experimental/bazel-base/Dockerfile
```diff
@@ -84,6 +84,19 @@ RUN git clone --branch v4.3 https://github.com/mfontanini/libtins.git && \
     # symlink to /usr/lib to match Magma VM
     ln -s /usr/local/lib/libtins.so /usr/lib/libtins.so
 
+RUN apt-get install -y --no-install-recommends \ 
+    libtool
+
+#####  asn1c
+RUN git clone https://gitlab.eurecom.fr/oai/asn1c.git && \
```
themarwhal, line 61: look into generating this with Bazel

electronjoe:
I modified the PR Description text mildly for outstanding tasks - to `figure out if compiling asn1c with bazel is an option, in order to increase determinism of asn1c use and reduce environmental dependencies by Bazel.`

tools/build_defs/asn1c.bzl
```diff
@@ -0,0 +1,163 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Custom library and macros to generate files with asn1c"""
+
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+def _get_dir_name(name):
+    # we need to postfix the directory name with .cc to trick Bazel into thinking this is a valid input
```
electronjoe, line 17: Is there any GH Issue we can cite about this approach (e.g. future fix / feature to Bazel that enables what you need here)? Or was this found in Google Groups?

tools/build_defs/asn1c.bzl
```diff
@@ -0,0 +1,163 @@
...
+    return name + ".cc"
+
+def _get_substitution_commands(ctx, dir_path):
```
electronjoe: In my read of this file, I'm unable to piece together effectively what the various `_get_` private functions are doing for you.  Would it be possible to give a top level strategy comment / description in the file? To explain what is broadly going on in this library? I know these are private, this is not for public use but for future editors or individuals who want to copy your work here for another generation task.  On that note you might highlight that other approaches are preferred if you know in advance your output file naming (either in comment in this file or in PR desc, maybe both).  I can just see this being copy-pasted in the future (maybe).

tools/build_defs/asn1c.bzl
```diff
@@ -0,0 +1,163 @@
...
+    Args:
+        name: the name of rule
+        asn1_file: relative path to the .asn1 file that will be passed to asn1c
+        prefix: value that is set to ASN1C_PREFIX
+    """
+    gen_name = name + "_genrule"
+
+    flags = "-pdu=all -fcompound-names -fno-include-deps -gen-PER"
+    substitutions = {
+        "18446744073709551615": "18446744073709551615u",
```
electronjoe, line 167: Wondering what these substitutions are

electronjoe (approved):

tools/build_defs/asn1c.bzl
themarwhal: yeah top level comment for what this rule does would be useful to have.

the _get functions I made were to keep the main command construction cleaner looking as it involves a lot of string formatting. Maybe _construct_X_ commands is a better name?

electronjoe: `_construct_` may help yeah.  And then some explanation of what you are constructing for (which I suppose is specific to asn1c compiler command line needs?).

tools/build_defs/asn1c.bzl
themarwhal: If I don't have the right extension, I just get an error message saying that the file has to have a certain extension. I added a comment with a related GH issue for more context.

tools/build_defs/asn1c.bzl
themarwhal: added a comment for each helper function and a bigger picture description in the main library definition

tools/build_defs/asn1c.bzl
themarwhal: I took it from here: https://github.com/magma/magma/blob/14c1cf643a61d576b3d24642e17ed3911d19210d/lte/gateway/c/core/oai/tasks/s1ap/CMakeLists.txt#L35

But I think it is to suppress warnings about the constant being too large.

Added a comment on it in the library def.

electronjoe: Oh my lol. OK well good to know the origin at least.

electronjoe (approved):

ardzoht (approved):
lgtm

PR #9695, opened 2021-10-15
fix(self-sign-cert): Add explicit list of altnames into controller cert

119Vik:
Signed-off-by: Vitalii Kostenko <vitalii@freedomfi.com>

## Summary
Some of the libraries, specifically Java based, required to automate interactions with API are not happy to work with wildcard certificates. At least alt names are required, when specific Alt names are missing it's treated as security issue.


hcgatewood (approved):
Looks reasonable to me

hcgatewood:
Hey @119Vik! Looks like the Semantic PR check is failing -- here's how to fix that https://github.com/magma/magma/pull/9695#issuecomment-944209424

PR #9697, opened 2021-10-15
feat: Migrate Bazel folly usage to https://github.com/storypku/rules_folly 

themarwhal:

## Summary
I was browsing the google discuss for bazel (https://groups.google.com/g/bazel-discuss) for an unrelated topic this morning and found that someone implemented a bazel library for folly!

It seems to work great and all unit tests pass. It does make building `bazel build ...`  from scratch (clean cache) longer as we now have to build folly instead of pulling it in from the system. (But this is OK, because we were cheating by building Folly in Docker anyways)

Additional work in this PR:
1. cleaning up the Dockerfile to remove unused `apt install ...`
2. pinning versions of `apt install ...`

## Test Plan
Building `bazel build @folly//:folly` after `bazel clean --expunge`
Screen Shot 2021-10-15 at 12 58 44 PM

Running ` bazel test lte/gateway/c/...:*` after ^
Screen Shot 2021-10-15 at 1 02 59 PM

## Additional Information

### DIFF
--- a/experimental/bazel-base/Dockerfile
+++ b/experimental/bazel-base/Dockerfile
@@ -6,26 +6,21 @@ FROM ubuntu:focal as bazel_builder
 ENV TZ=America/New_York \
     DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && \
-    apt-get upgrade -y && \
-    apt-get install -y --no-install-recommends \
+RUN apt-get update && apt-get upgrade -y 
+
+RUN apt-get install -y --no-install-recommends \
         apt-transport-https \
         apt-utils \
         build-essential \
         ca-certificates \
         curl \
-        # dependency of @sentry_native//:sentry
-        libcurl4-openssl-dev \
         gcc \
         git \
         gnupg2 \
         g++ \
         python-dev \
-        zip \
-        unzip \
         vim \
-        wget \
-        libssl-dev
+        wget
 
 # Install bazel
 WORKDIR /usr/sbin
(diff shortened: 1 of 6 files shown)


uri200 (approved):

experimental/bazel-base/Dockerfile
```diff
@@ -6,70 +6,42 @@ FROM ubuntu:focal as bazel_builder
...
         ca-certificates \
         curl \
-        # dependency of @sentry_native//:sentry
-        libcurl4-openssl-dev \
         gcc \
         git \
         gnupg2 \
         g++ \
         python-dev \
-        zip \
```
electronjoe, line 24: So I *think* the best way to achieve this PR would be to actually commit these changes to the Dockerfile first - then once it has been built to registry use this PR to migrate the devcontainer hash to the one with these omissions. That way your PR demonstrates health on the full suite of changes (otherwise the environmental is contaminanting our CI validation of the bazel changes?).

themarwhal: ah you mean remove Folly from the devcontainer dockerfile first?

themarwhal: So I can't remove the folly installation from the devcontainer dockerfile just yet as that would make it so that SessionD cannot be built with CMake in the devcontainer. (Which is what some of our CI tests rely on). The bazel-base/Dockefile container is not used for anything aside from me testing bazel changes.

electronjoe: Aah shoot, and this is why my coupling of our devcontainer with CI phases / CMake use is bad... Or at least prevents best practices.

electronjoe (approved):

experimental/bazel-base/Dockerfile
themarwhal: Yes, that is why I'm keeping around bazel-base/Dockerfile :) Once everything is migrated to Bazel, the devcontainer Dockerfie should look exactly like this file. And then dev tools on top of it

Issue #9508, opened 2021-10-06
Refactor Orc8r mconfig streaming of Sentry config


themarwhal:
## Changes to be made
Modify the MConfig structure so that have a shared_config (name pending) key in the streamed down Mconfig.

Current structure of Mconfig is a map of service name to service specific config proto. 
![current-config](https://user-images.githubusercontent.com/37634144/136258501-24345b7a-b8bf-41eb-897b-093a4df0df0f.png)

Up until now, we've limited the sentry config to MME and SessionD only, but now that we are going ahead with Sentry to enable for the rest of AGW services, we need the sentry configurations to be available to all AGW services.
This config structure has served us well for now as most configs are specific to a service, but with the addition of sentry configs we need a place to store configs that are shared by all services. 

Suggested structure of Mconfig keeps the same format for all configs except ones that are shared by all services, like sentry config.
![suggested-config](https://user-images.githubusercontent.com/37634144/136259018-5515ea16-5036-42a2-8af3-67ac3816f52e.png)

## Some things to be considered
1. AGW services will now have to read two keys. it's own service config and the shared config.
2. Do we want all services to restart when a shared config is updated? If so, we will need to modify the service restart logic in magmad.
3. Will there be any changes needed to make sure the mconfig digest mechanism does not break?

## Follow up tasks
- [x] https://github.com/magma/magma/issues/10170
- [x] https://github.com/magma/magma/issues/9907
- [x] https://github.com/magma/magma/issues/10760
- [x] #11092
- [x] #11091
- [x] #11399

PR #9698, opened 2021-10-15
feat(agw): Configure Sentry per Python service

sebathomas:
## Summary

Performance issues were reported after enabling Sentry for Python gateway services. As a first step towards Sentry integration, this PR makes it possible to enable Sentry per Python service. By default, it is only enabled for mobilityd.

(Sentry remains disabled for all services unless a sentry URL is given in control_proxy.yml, this is not changed here.)

## Test Plan

Executed loadtest_mobilityd.py with and without Sentry enabled, including fake error messages that were written for each request. No performance degradation was observed.

### DIFF
--- a/orc8r/gateway/python/magma/common/sentry.py
+++ b/orc8r/gateway/python/magma/common/sentry.py
@@ -18,6 +18,7 @@ import snowflake
 from magma.configuration.service_configs import get_service_config_value
 
 CONTROL_PROXY = 'control_proxy'
+SENTRY_ENABLED = 'sentry_enabled'
 SENTRY_URL = 'sentry_url_python'
 SENTRY_SAMPLE_RATE = 'sentry_sample_rate'
 COMMIT_HASH = 'COMMIT_HASH'
@@ -27,6 +28,14 @@ SERVICE_NAME = 'service_name'
 
 def sentry_init(service_name: str):
     """Initialize connection and start piping errors to sentry.io."""
+    sentry_enabled = get_service_config_value(
+        service_name,
+        SENTRY_ENABLED,
+        default=False,
+    )
+    if not sentry_enabled:
+        return
+
     sentry_url = get_service_config_value(
         CONTROL_PROXY,
         SENTRY_URL,
--- a/lte/gateway/configs/ctraced.yml
+++ b/lte/gateway/configs/ctraced.yml
@@ -23,3 +23,6 @@ trace_interfaces:
 #   - tshark
 # tshark has more capabilities - see command_builder.py
 trace_tool: tshark
+
+# [Experimental] Enable Sentry for this service
+sentry_enabled: false
--- a/lte/gateway/configs/directoryd.yml
+++ b/lte/gateway/configs/directoryd.yml
@@ -15,3 +15,6 @@
 print_grpc_payload: false
 
 log_level: INFO
+
+# [Experimental] Enable Sentry for this service
+sentry_enabled: false
--- a/lte/gateway/configs/enodebd.yml
+++ b/lte/gateway/configs/enodebd.yml
@@ -46,3 +46,6 @@ web_ui_enable_list: []
 
 # Network interface to terminate S1
 s1_interface: eth1
+
+# [Experimental] Enable Sentry for this service
+sentry_enabled: false
--- a/lte/gateway/configs/eventd.yml
+++ b/lte/gateway/configs/eventd.yml
@@ -60,3 +60,6 @@ event_registry:
   s1_setup_success:
     module: lte
     filename: mme_events.v1.yml
+
+# [Experimental] Enable Sentry for this service
+sentry_enabled: false
--- a/lte/gateway/configs/health.yml
+++ b/lte/gateway/configs/health.yml
@@ -25,3 +25,6 @@ state_recovery:
   interval_check_mins: 3
   # Destination path to save redis RDB temp snapshots
   snapshots_dir: /tmp/redis_snapshots
+
+# [Experimental] Enable Sentry for this service
+sentry_enabled: false
--- a/lte/gateway/configs/kernsnoopd.yml
+++ b/lte/gateway/configs/kernsnoopd.yml
@@ -20,3 +20,6 @@ collect_interval: 10
 ebpf_programs:
   ~
 #  - byte_count
+
+# [Experimental] Enable Sentry for this service
+sentry_enabled: false
--- a/lte/gateway/configs/magmad.yml
+++ b/lte/gateway/configs/magmad.yml
@@ -172,3 +172,6 @@ services_to_restart:
 
 # How many times can grpc status check fail before restarting the services above
 restart_timeout_threshold: 15
+
+# [Experimental] Enable Sentry for this service
+sentry_enabled: false
--- a/lte/gateway/configs/mobilityd.yml
+++ b/lte/gateway/configs/mobilityd.yml
@@ -15,3 +15,6 @@
 persist_to_redis: true
 redis_port: 6380
 print_grpc_payload: false
+
+# [Experimental] Enable Sentry for this service
+sentry_enabled: true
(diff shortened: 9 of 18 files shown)


electronjoe (commented):
Thanks for the PR @sebathomas!

In PR Description Test plan: `Executed loadtest_mobilityd.py with and without Sentry enabled, including fake error messages that were written for each request. No performance degradation was observed.` Could we get a little more detail on the following?

- What error rate of fake errors was generated in the load test?
- What were the load test output results (e.g. json output if not too awful to visually parse)
- What delta was that load test result relative to master (presume at least some jitter in measurement)

electronjoe:
Aside: all GH Workflow checks look good. the `nms-eslint...` job has been failing for days and is not required (fix is in progress by @andreilee). And the Codecov warning is due to an ongoing coverage analysis challenge given we have such a large code base and do not run all tests on every PR presubmit (a nuance that CodeCov does support but it's not easily accomplished in practice).

orc8r/gateway/python/magma/common/sentry.py
```diff
@@ -27,6 +28,14 @@
 
 def sentry_init(service_name: str):
     """Initialize connection and start piping errors to sentry.io."""
+    sentry_enabled = get_service_config_value(
+        service_name,
+        SENTRY_ENABLED,
+        default=False,
```
themarwhal, line 34: Since we are planning on all the sentry logs to be opt-in only, I wonder if it makes sense for the default value to be True. 

We already have an AGW wide `sentry_url_python` value that determines whether Sentry can be enabled or not. If we want all services to enable Sentry eventually, it would be annoying to have to enable Sentry in control_proxy.yml (and orc8r eventually) and per-service yamls. 

But if the intention is to remove this per-service sentry enable/disable configuration once we sanitize sentry outputs for all services, then the default being False makes sense to me.

sebathomas:
Thank you for the review! The load test results vary a lot, with typical rates of 30-90 IP allocation requests per second (rps). For directly successive runs of the load tests the rps were relatively stable, running them a few minutes apart showed larger differences.

This was unchanged independent of whether Sentry was enabled or not (either for mobilityd or all services).


## The raw numbers

We extracted the most important fields from the result jsons. Unfortunately, we only recorded three results without Sentry enabled but we actually ran more of these tests.

### Sentry disabled:

| slowest (ns) | average (ns) | fastest (ns) | total (ns) | rps | count (number of requests) |
| -- | -- | -- | -- | -- | -- |
| 2001564677 | 829587934 | 15152491 | 33449235559 | **59** | 2000 |
| 2761811440 | 1261932126 | 19580185 | 50765568867 | **39** | 2000 |
| 2251804070 | 779607697 | 38923832 | 31462311827 | **63** | 2000 |

### Sentry enabled for mobilityd, no errors:

| slowest (ns) | average (ns) | fastest (ns) | total (ns) | rps | count (number of requests) |
| -- | -- | -- | -- | -- | -- |
| 699774951 | 567317268 | 19712021 | 22956395684 | **87** | 2000 |
| 2403008103 | 871337176 | 25797381 | 35126123042 | **56** | 2000 |
| 1130951875 | 723729686 | 25866845 | 29222904762 | **68** | 2000 |
| 973374505 | 713638749 | 29594374 | 7460408428 | **67** | 500 |
| 1247997865 | 951059901 | 28206858 | 4327131018 | **46** | 200 |
| 1119424531 | 934027053 | 33253210 | 4203257839 | **47** | 200 |
| 1098205666 | 933978103 | 30752760 | 4235286691 | **47** | 200 |
| 756555979 | 607117568 | 16585943 | 2751665446 | **72** | 200 |
| 954310342 | 625839296 | 23051592 | 25328930391 | **78** | 2000 |

### Sentry enabled for mobilityd, one error log message per request:

| slowest (ns) | average (ns) | fastest (ns) | total (ns) | rps | count (number of requests) |
| -- | -- | -- | -- | -- | -- |
| 2094596199 | 1299393269 | 79661398 | 5974935428 | **33** | 200 |
| 1746449215 | 1333042372 | 73151225 | 5931546724 | **33** | 200 |
| 1655324284 | 1344652743 | 128402362 | 6161859192 | **32** | 200 |
| 943498484 | 785870390 | 24925691 | 3528208763 | **56** | 200 |
| 648469062 | 537667053 | 29762453 | 2415173465 | **82** | 200 |
| 1219043356 | 1017220822 | 76298562 | 4596527111 | **43** | 200 |

orc8r/gateway/python/magma/common/sentry.py
sebathomas: Since we currently couldn't measure any performance issues (but as you see below, the jitter is massive), we are not sure if Sentry needs be made opt-in only in the end. Of course it might still be an option for clean-up and filtering of what exceptions to send.

So it's possible these config values will be removed again in the end if we can enable Sentry for all services.

themarwhal: As for opt-in vs opt-out, I think opt-in reporting of logs can still be beneficial for filtering / cleanup as you said. So we can go with assuming opt-in for now.

If the goal is to remove the per-service configuration after this migration/experimentation phase, then I'm happy with this change.

ardzoht:
@sebathomas Thanks for parsing the output, those JSON files are usually large and seeing them on table is great. From the load test results shared, I agree that the RPS and avg time of request seems to fluctuate more than we would want to (this from a general run, not taking into account sentry enabled / disabled type of runs). From my past runs I also see RPS being in a expected range from 50-80 RPS, and from the `Sentry enabled` results I don't see any major hit on the performance of the runs.  cc @electronjoe

orc8r/gateway/python/magma/common/sentry.py
themarwhal: One last note, if the default value for non existing config is False already, should we just enable it for mobilityd in this PR? (And just leave the other config files untouched).

orc8r/gateway/python/magma/common/sentry.py
```diff
@@ -27,6 +28,14 @@
 
 def sentry_init(service_name: str):
     """Initialize connection and start piping errors to sentry.io."""
+    sentry_enabled = get_service_config_value(
```
ardzoht, line 31: I think it makes more sense to do this config retrieval on each of the services/main() functions, this way the config can be accessed through the `service.config` object and also would help if we're looking to add Orc8r/mconfig support as well. However this depends if we're aiming through a wide enablement of Sentry on the AGW this probably is good enough for reading the configuration. @themarwhal thoughts?

themarwhal: Not sure if you've seen the thread here: https://github.com/magma/magma/pull/9698#discussion_r729864971

So just to make the goals clear:
1. we do want to enable sertry for AGW services eventually
2. we do want to move the configuration to mconfig eventually. This is actually tracked here: https://github.com/magma/magma/issues/9508. 

What I wanted to figure out in the discussion I linked was whether we want to support per-service configuration on sentry enablement. I think that if would be great to just do a simple all or nothing configuration for Sentry. (Some assumptions I'm making here as I mention in the thread is that all logs will be exported on an opt-in basis. So I'm assuming here that performance will not be a concern, which may or may not be true).

But not to get too sidetracked, I think the placing the config reading here is fine, if it's temporary :) . If we want to support per-service decision on whether to enable sentry or not long term, I agree with @ardzoht that reading the config out in services/main is cleaner.

ardzoht (approved):
lgtm

orc8r/gateway/python/magma/common/sentry.py
sebathomas: That's actually what we wanted to do at first. But if you try to access a config value that's not in the file, the service will log an error, so we thought it would be cleaner to put it into all configs.

themarwhal (approved):

sebathomas:
It's ready to ship but I can't set the label.

PR #9712, opened 2021-10-15
fix(mobilityd): ipv6 block size

pshelar:
mobilityD expect operator to allocate '/48' or
bigger range. That is not practical or required for
many deployments. Following PR relaxes it to
'/60' and makes it configurable.

## Test Plan
`make test`

## Additional Information

### DIFF
--- a/lte/gateway/configs/mobilityd.yml
+++ b/lte/gateway/configs/mobilityd.yml
@@ -18,3 +18,4 @@ print_grpc_payload: false
 
 # [Experimental] Enable Sentry for this service
 sentry_enabled: true
+ipv6_prefixlen: 58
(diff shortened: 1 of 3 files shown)


ardzoht (approved):
lgtm

ulaskozat:
@pshelar, @ardzoht currently prefix allocation to a session happens randomly. To prevent too many collisions, I believe Alex have left originally at least 16 bits.  Can you guys add a non-random strategy for prefix assignment to the sessions if we are shrinking this to as small as 4 bits?

ulaskozat (changes requested):
add a non-random assignment strategy first as each session must have a unique IPv6 prefix.

pshelar:
> @pshelar, @ardzoht currently prefix allocation to a session happens randomly. To prevent too many collisions, I believe Alex have left originally at least 16 bits. Can you guys add a non-random strategy for prefix assignment to the sessions if we are shrinking this to as small as 4 bits?

@ulaskozat I have increased number of tries for now. we can discuss improvement in separate ticket: #9735

ulaskozat (approved):
LGTM. Let's not have the follow up issue to fall through the cracks.

PR #9719, opened 2021-10-16
fix(ci): fix path filter with new module

### LINKED ISSUE #9563: GHA path filter module is not best suited for our usecase
The current used path filter module (fkirc/skip-duplicate-actions@master) works best on workflow run and not on pull_request.
The backtracking algorithm does not work good in our use case

The documentation of this module recommend using dorny/paths-filter@v2 for pull_request trigger
I replaced it but it was failing on push event : https://github.com/magma/magma/pull/9256

From module documentation it should not fail on push.
Investigate what's happening or find another module

quentinDERORY:
## Summary

Old path filter module was not best suited for pull_request and push event.
The backtracking algorithm was suited for workflow triggers.

## Test Plan

On PR and on my fork for push trigger

## Additional Information

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -17,18 +17,21 @@ env:
 jobs:
   path_filter:
     runs-on: ubuntu-latest
-    # Map a step output to a job output
     outputs:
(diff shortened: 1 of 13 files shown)


electronjoe (approved):

electronjoe:
Thanks Quentin!

PR #9723, opened 2021-10-18
chore: run agw-workflow for new devcontainer

maxhbr:
## Summary

try to use [reusing-workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows) to run the agw-workflow on change of the devcontainer dockerfile.

This currently depends on https://github.com/magma/magma/pull/9787

## Test Plan

tbd.

### DIFF
--- a/.github/workflows/docker-push.yml
+++ b/.github/workflows/docker-push.yml
@@ -85,3 +85,13 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+
+  agw_test_devcontainer:
+    # DO NOT COMMIT:
+    # before merging, this needs to point to magma/magma/.github/workflows/agw-workflow.yml@master
+    # and https://github.com/magma/magma/pull/9787 needs to be merged
+    uses: maxhbr/magma/.github/workflows/agw-workflow.yml@maxhbr/makeAgwWorkflowCallable
+    needs: build_devcontainer_dockerfile
+    with:
+      devcontainer_image: ${{ needs.build_devcontainer_dockerfile.outputs.tags[0] }}
+      force_run: true
--- a/.github/workflows/dockerfile-health.yml
+++ b/.github/workflows/dockerfile-health.yml
@@ -83,3 +83,13 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           outputs: type=docker,dest=/tmp/mme_builder.tar
+
+  agw_test_devcontainer:
+    # DO NOT COMMIT:
+    # before merging, this needs to point to magma/magma/.github/workflows/agw-workflow.yml@master
+    # and https://github.com/magma/magma/pull/9787 needs to be merged
+    uses: maxhbr/magma/.github/workflows/agw-workflow.yml@maxhbr/makeAgwWorkflowCallable
+    needs: build_devcontainer_dockerfile
+    with:
+      devcontainer_image: ${{ needs.build_devcontainer_dockerfile.outputs.tags[0] }}
+      force_run: true


maxhbr:
Making progress:

.github/workflows/agw-workflow.yml
```diff
@@ -126,7 +135,7 @@ jobs:
       - name: Run common tests
         uses: addnab/docker-run-action@v2
         with:
-          image: ${{ env.DEVCONTAINER_IMAGE }}
+          image: ${{ inputs.devcontainer_image }}
```
quentinDERORY: Can you default it to the previous variable if the input is not present.
Something like `${{ inputs.devcontainer_image ||  env.DEVCONTAINER_IMAGE  }}`
I would advise testing it... Otherwise you could always default this variable in a bash step.
Otherwise it will break this workflow.

.github/workflows/dockerfile-health.yml
```diff
@@ -12,7 +12,21 @@ on:  # yamllint disable-line rule:truthy
...
 
 jobs:
+  build_ubuntu20_04_dockerfile_filter:
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
```
quentinDERORY: use the latest dorny module for path filtering.
And why you need two filter job? One should be enough

.github/workflows/agw-workflow.yml
maxhbr: yes, I am aware of that. Will test before setting it to ready for review.

.github/workflows/dockerfile-health.yml
maxhbr: My goal was, to not run the the both health checks, if only one Dockerfile was touched.

.github/workflows/dockerfile-health.yml
```diff
@@ -82,4 +110,11 @@ jobs:
           push: false
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          outputs: type=docker,dest=/tmp/mme_builder.tar
\ No newline at end of file
+          outputs: type=docker,dest=/tmp/mme_builder.tar
+
+  agw_test_devcontainer:
+    uses: maxhbr/magma/.github/workflows/agw-workflow.yml@maxhbr/runAgwWorkflowForNewDevcontainers
```
maxhbr: still my biggest issue: have not yet found a way to reference the `agw-workflow.yml` of the current revision.

.github/workflows/dockerfile-health.yml
quentinDERORY: `.github/workflows/agw-gateway.yml` is not present in the original trigger path filtering, you should add it there too.
To simplify this you can put both path filtering in same job and have two output variables

.github/workflows/dockerfile-health.yml
quentinDERORY: YOu should be able to suffix the current commit id at the end like it's done here:
`https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#calling-a-reusable-workflow`
THis will look something like 
`maxhbr/magma/.github/workflows/agw-workflow.yml@maxhbr/runAgwWorkflowForNewDevcontainers@FULL_COMMIT_ID`

maxhbr: Yes, but I have no access to variables at that point. It is not possible to write something like
```suggestion
    uses: ${{ github.repository }}/.github/workflows/agw-workflow.yml@${{ github.sha }}
```

.github/workflows/dockerfile-health.yml
maxhbr: It is not a trigger, but if it was changed it should result in a rerun. I also thought about dropping that part again, since it caches way more aggressively. If the listed files did not change since the last run, it also skips.

.github/workflows/agw-workflow.yml
maxhbr: I will try to do these tests in https://github.com/magma/magma/pull/9787

.github/workflows/dockerfile-health.yml
quentinDERORY: I would say it works but better test it ^^'

.devcontainer/Dockerfile
```diff
@@ -119,8 +119,7 @@ RUN echo "Install 3rd party dependencies" && \
     echo "Install ZeroMQ" && \
     apt-get install -y libczmq-dev && \
     echo "Install dependencies for Sentry Native" && \
-    apt-get install -y libcurl4-openssl-dev && \
-    ln -s /usr/bin/python2.7 /usr/local/bin/python
```
electronjoe: Was the removal of python here technically part of this PR? Or was that the pre-work that triggered the value of these PRs and it slipped in here.

.github/workflows/docker-push.yml
```diff
@@ -85,3 +85,12 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+
+  agw_test_devcontainer:
+    # TODO:
```
electronjoe: Not a request for change, just a comment in case you find it helpful.  I personally use `# DO NOT COMMIT:`  style comments when I want to make it abundantly clear that this is to be fixed before merge (makes it all the more obvious to reviewer or myself at a glance).  Some tooling might prevent merge with this style of string.. Although as I write this I am failing to find any such plug-in and wonder if this was just a cultural thing in some teams I was in.

.github/workflows/dockerfile-health.yml
```diff
@@ -82,4 +82,13 @@ jobs:
           push: false
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          outputs: type=docker,dest=/tmp/mme_builder.tar
\ No newline at end of file
+          outputs: type=docker,dest=/tmp/mme_builder.tar
+
+  agw_test_devcontainer:
```
electronjoe, line 87: I'm a bit surprised that this stage (`agw_test_devcontainer`) works, because I didn't think that e.g. `/tmp/mme_build.tar` which you added to the prior stage is accessible to this stage (which can run on a different resource)? I think historically to achieve this I had to export the desired binary artifact, then pull it in in the subsequent phase and import the docker image for use?

If you have to go this route, let me know and I'll dig up my example GH workflow that used to do this (have to find an old commit).

.github/workflows/dockerfile-health.yml
```diff
@@ -82,4 +82,13 @@ jobs:
...
+    # TODO:
+    # before merging, this needs to point to magma/magma/.github/workflows/agw-workflow.yml@master
+    # and https://github.com/magma/magma/pull/9787 needs to be merged
+    uses: maxhbr/magma/.github/workflows/agw-workflow.yml@maxhbr/makeAgwWorkflowCallable
+    needs: build_devcontainer_dockerfile
+    with:
+      devcontainer_image: ${{ needs.build_devcontainer_dockerfile.outputs.tags[0] }}
```
electronjoe, line 94: Does this need your `force_run`? I'm not seeing the workflow execute in e.g. [workflow test results](https://github.com/magma/magma/actions/runs/1363988275)

.github/workflows/agw-workflow.yml
electronjoe: This answers my question in #9787 - I was unaware you could do this =D You can close my question comment in that PR @maxhbr

.github/workflows/dockerfile-health.yml
electronjoe: `have not yet found a way to reference the agw-workflow.yml of the current revision.` Would you like us to ask our friends at the Linux Foundation to relay this question to Github? Is there a documentation / feasibility hole in the new re-use?

.devcontainer/Dockerfile
maxhbr: that was pre-work that is still here to force a rebuild (although the modification of the `dockerfile-health.yml` file is sufficient as I learned later).

I will drop that change.

.github/workflows/dockerfile-health.yml
maxhbr: yes, you are absolutely right

maxhbr:
similar to #9787, this is not the best way of composing workflows. composite-actions are the better, but also more complex, solution.

PR #9724, opened 2021-10-18
fix(mme): Unsafe delete_wrapper

### LINKED ISSUE #8894: Unsafe delete_wrapper

shashidhar-patil:
Signed-off-by: shashidhar-patil <shashidhar.patil@wavelabs.ai>

## Summary

Earlier implementation: Pointer to be freed is cast to void** and passed to delete_wrapper. The delete wrapper calls delete on the void pointer.

Issue: Casting the pointers into void** and then calling delete on them may cause defects as the delete operator will not call the destructor for the object of type "Example_type" pointed to by 'example_ptr'. This will lead to undefined behavior and is unsafe.

Code Change: The delete wrapper is implemented as a template function and now takes into account the data type of the pointer while deleting it. This performs a safe delete.
 

## Test Plan
- Verified basic sanity with UERANSIM.
- Please find attached logs and pcap snap for basic registration.
- Verified UT cases.

Logs:
[mme.log](https://github.com/magma/magma/files/7364678/mme.log)

pcap snapshot:
pcap_delete_wrapper

Unit test snapshot:
UT_delete_wrapper

## Additional Information

<!--
    If this is a backwards-breaking change, document the upgrade instructions.
    All upgrade instructions for backwards-breaking changes will be aggregated
    in the next release's changelog so this is very important to fill out.
--

>

### DIFF
--- a/lte/gateway/c/core/oai/tasks/amf/amf_app_ue_context_and_proc.h
+++ b/lte/gateway/c/core/oai/tasks/amf/amf_app_ue_context_and_proc.h
@@ -41,6 +41,7 @@ extern "C" {
 #include "M5GRegistrationRequest.h"
 #include "ngap_messages_types.h"
 #include "amf_common.h"
+#include "assertions.h"
 
 // NAS messages
 #include "M5GDLNASTransport.h"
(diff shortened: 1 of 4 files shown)


lte/gateway/c/core/oai/tasks/amf/amf_app_ue_context_and_proc.h
```diff
@@ -850,4 +849,29 @@ ue_m5gmm_context_s* ue_context_lookup_by_gnb_ue_id(
 
 int amf_idle_mode_procedure(amf_context_t* amf_ctx);
 void amf_free_ue_context(ue_m5gmm_context_s* ue_context_p);
+
+/*******************  ***************************************************
```
ulaskozat: fix the gap in `**  ***`

ulaskozat (approved):
LGTM with a nit comment. please label as ready to merge when ready.

lte/gateway/c/core/oai/tasks/amf/amf_app_ue_context_and_proc.h
shashidhar-patil: Okay, fixed it in the same commit.

PR #9736, opened 2021-10-18
feat(mobilityd): add support for IPv6 default gw

pshelar:
This would allow Non-NAT IPV6 traffic over SGi.

## Test Plan
`make test`
lab setup test on bare metal agw.

## Additional Information


ardzoht (approved):
lgtm

PR #9738, opened 2021-10-18
chore: remove OAILOG_* from assertions.h

### LINKED ISSUE #9732: [MME bazelification] Remove cyclic dependency between Logging, ITTI and Assertions
@themarwhal noticed the following cyclic dependency across MME libraries and common utils

mme-dependency-cycle-2

themarwhal:
## Summary
See discussion here: https://github.com/magma/magma/issues/9732#issuecomment-947055779

This PR removes OAI_LOG macros from assertions.h

## Test Plan
make test_oai
make integ_test

## Additional Information


lionelgo (approved):
Fine if we don't care about logging assert messages with app messages.

PR #9739, opened 2021-10-18
feat(orc8r): add DeMap endpoints to directoryd

### LINKED ISSUE #8925: feat(orc8r): reconcile secondary indexes with state

uri200:

## Summary

Added endpoints to handle DeMap an entry (remove from a map) on directoryd. 

Those DeMap functions impact directoryd_blobstore table with type
```
sessionid_to_imsi
sgwCteid_to_hwid
hwid_to_hostname
```

## Test Plan

```
./build.py -g
./build.pu -c
```
## Additional Information

### DIFF
--- a/orc8r/cloud/go/services/directoryd/servicers/servicer.go
+++ b/orc8r/cloud/go/services/directoryd/servicers/servicer.go
@@ -56,13 +56,13 @@ func (d *directoryLookupServicer) MapHWIDsToHostnames(ctx context.Context, req *
 	return &protos.Void{}, err
 }
 
-func (d *directoryLookupServicer) DeMapHWIDsToHostnames(ctx context.Context, req *protos.DeMapHWIDToHostnameRequest) (*protos.Void, error) {
+func (d *directoryLookupServicer) UnmapHWIDsToHostnames(ctx context.Context, req *protos.UnmapHWIDToHostnameRequest) (*protos.Void, error) {
 	err := req.Validate()
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to validate DeMapHWIDToHostnameRequest")
+		return nil, errors.Wrap(err, "failed to validate UnmapHWIDToHostnameRequest")
 	}
 
-	err = d.store.DeMapHWIDsToHostnames(req.Hwid)
+	err = d.store.UnmapHWIDsToHostnames(req.Hwids)
 
 	return &protos.Void{}, err
 }
@@ -93,13 +93,13 @@ func (d *directoryLookupServicer) MapSessionIDsToIMSIs(ctx context.Context, req 
 	return &protos.Void{}, err
 }
 
-func (d *directoryLookupServicer) DeMapSessionIDsToIMSIs(ctx context.Context, req *protos.DeMapSessionIDToIMSIRequest) (*protos.Void, error) {
+func (d *directoryLookupServicer) UnmapSessionIDsToIMSIs(ctx context.Context, req *protos.UnmapSessionIDToIMSIRequest) (*protos.Void, error) {
 	err := req.Validate()
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to validate DeMapSessionIDToIMSIRequest")
+		return nil, errors.Wrap(err, "failed to validate UnmapSessionIDToIMSIRequest")
 	}
 
-	err = d.store.DeMapSessionIDsToIMSIs(req.NetworkID, req.SessionID)
+	err = d.store.UnmapSessionIDsToIMSIs(req.NetworkID, req.SessionIDs)
 
 	return &protos.Void{}, err
 }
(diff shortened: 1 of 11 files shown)


orc8r/cloud/go/services/directoryd/storage/storage_blobstore.go
```diff
@@ -51,113 +51,112 @@ type directorydBlobstore struct {
 	factory blobstore.BlobStorageFactory
 }
 
+////////////////
+// Get functions
```
hcgatewood: Nit: block comments like this are generally discouraged. Instead, for cases like this, it can make sense to separate into e.g. `storage_blobstore.go`, `storage_blobstore_getters.go`, and `storage_blobstore_setters.go`

Or can split it by the logical piece of data under operation. But either way, preference is separate files before block comments

orc8r/cloud/go/services/directoryd/storage/storage_blobstore.go
```diff
@@ -51,113 +51,112 @@ type directorydBlobstore struct {
...
-		storage.TK{Type: DirectorydTypeHWIDToHostname, Key: hwid},
-	)
-	if err == merrors.ErrNotFound {
-		return "", err
-	}
-	if err != nil {
-		return "", errors.Wrap(err, "failed to get hostname")
-	}
+	res, err := d.getFromStore(placeholderNetworkID, DirectorydTypeHWIDToHostname, hwid)
+	printIfError(err, "Error GetHostnameForHWID: %s", err)
```
hcgatewood: No need to decorate the err with the function name -- instead, can print the err with `%+v` to add the stack trace

orc8r/cloud/go/services/directoryd/servicers/servicer.go
```diff
@@ -48,20 +48,32 @@ func (d *directoryLookupServicer) GetHostnameForHWID(
...
-		return nil, errors.Wrap(err, "failed to validate request")
+		return nil, errors.Wrap(err, "failed to validate MapHWIDToHostnameRequest")
 	}
 
 	err = d.store.MapHWIDsToHostnames(req.HwidToHostname)
 
 	return &protos.Void{}, err
 }
 
+func (d *directoryLookupServicer) DeMapHWIDsToHostnames(ctx context.Context, req *protos.DeMapHWIDToHostnameRequest) (*protos.Void, error) {
```
hcgatewood: General nit: seems "Unmap" is the preferred terminology, rather than "Demap"

```
~/Desktop/go/src master                                                                                                                   
$ ag '\bdemap' | wc -l
       0

~/Desktop/go/src master                                                                                                              
$ ag '\bunmap' | wc -l
      60
```

hcgatewood (approved):
Lgtm. Couple nits, otherwise looks 💯💯

orc8r/cloud/go/services/directoryd/storage/storage_blobstore.go
uri200: opps forgot to remove them

orc8r/cloud/go/services/directoryd/storage/storage_blobstore.go
uri200: Not sure I got this right, but `printIfError` is just a helper function to print something if there is an error.  It is a conditional print.


```
// printIfError prints in case of errors.
// Args:
// 	err: error to check
//  msg: message to print
//  a: Arguments are handled in the manner of fmt.Printf; a newline is appended if missing.
func printIfError(err error, msg string, a ...interface{}) {
	if err != nil {
		glog.Errorf(msg, a...)
	}
}
```

hcgatewood: Ah right so the point was not that we shouldn't use `printIfError` (I usually do the `if err != nil { glog... }`, but up to you), but rather that you don't need to put `GetHostnameForHWID` in the printed err since you can instead use `%+v` rather than `%s`, where the former will include the stack trace

uri200: Oh I see what you mean. The reason not to include the stack trace is to reduce noise. Since this is a function on orc8r it may be called by many AGW. Errors can be as simple as requests are missing some needed values. Having the stack trace may mask the obvious.

From my experience on sessiond debugging (many subscribers are creating logs), the best is to have clear and minimal error messages better than the whole trace.

hcgatewood: My input is I've had the opposite experience with Orc8r: generally we don't have a lot of err logs, so full context on each err is super helpful for nailing down the rc without having to try to replicate the bug in a lab

In any case: current convention is to use `%+v`. I'm happy to revisit the convention if desired, but let's leave it as outside the scope of this PR

uri200: Lets do this to get the best of two worlds 
```
printIfError(err, "Error GetHostnameForHWID: %+v", err)
```

Another reason I forgot to mention about custom errors is to be able to grep easily on a long log. So this solution will give us the chance to do that without missing any context provided by the error

hcgatewood: Cool, definitely works for me 👍👍

PR #9749, opened 2021-10-19
chore(mme): Remove unused code on bstring lib

ardzoht:
## Summary

- Removes all unused code related to bstring library
- Applied clang-tidy recommendations to bstring code

## Test Plan

- `make build_oai FEATURES=mme` `make build_oai`
- ran make integ_test
- built docker MME and ensured it passes successfully

## Additional Information


ssanadhya (approved):

ulaskozat (approved):
Thanks for pruning the unused code/libs...

themarwhal:
Just a heads up that this PR breaks the bstrlib unit tests. Wasn't sure if you were aware. But the unit tests aren't currently run in the MME test unit suite, so we don't see anything.  (See https://github.com/magma/magma/pull/9584)

If we are planning on removing bstr anyways, then this probably doesn't matter. But if we plan to keep it around, it might be good to run the unit tests as we seemed to have changed the lib a little bit.

(and if we don't care about the unit tests, we can just delete those too :) )

ulaskozat:
@themarwhal we are trying to remove unused code to eliminate the need for adding unit test coverage for them. does codecov take these unit tests into account?

ulaskozat:
and which workflow triggers bstrlib unit tests?

themarwhal:
> @themarwhal we are trying to remove unused code to eliminate the need for adding unit test coverage for them. does codecov take these unit tests into account?

Ah I see. No these unit tests are currently not run in CI. So they are not tracked. 

They are runnable, at least they were before some functions were removed in this PR, but they're not run. 

So the options are
1. Run the unit tests to add coverage for these files
2 remove the unit tests as they are also dead code 

1 is probably better if we want bstrlib to stick around

ardzoht:
I would advocate for removing these unit tests (at least the ones that are related with the files we are removing here) since we will be moving away from this lib once we start migrating to C++ @ulaskozat @ssanadhya  thoughts?

ulaskozat:
let's prune the parts of the unit tests that are no longer relevant due to deletion.

ssanadhya:
+1 on removing the unit tests which are not relevant after this PR. No point keeping unused code just because it has test coverage.

ardzoht:
I will put a PR to take these out @themarwhal I will need your help / guidance validating the changes on bazel build process

themarwhal:
So it is looking like all unit tests (there are 2 test files) are not runnable after the initial pruning (this PR). 

I'm also seeing that the following files are not buildable:
* bstrwrap.cpp / bstrwrap.h

These files don't seem to be used anywhere:
* bsafe.c / bsafe.h
* bstraux.h

And these unit tests can be deleted:
* test.cpp
* bstest.c

cc @ardzoht

PR #9751, opened 2021-10-19
chore(agw): Enable pylint test on mobilityd

### LINKED ISSUE #6099: [Pylint] Enable pylint for `mobilityd` service
POC: @themarwhal (GH) Marie Bremner (Slack)

We currently only do `pylint` unit test checks for the following services:
```
# orc8r/gateway/python
'common',
'configuration',
'ctraced',
'directoryd',
'eventd',
'magmad',
'state',
# lte/gateway/python
'enodebd',
'pipelined',
'policydb',
'subscriberdb',
```

In the orc8r directory, we should enable pylint checks for the following services:
```
'ctraced',
```

In the lte directory, we should enable pylint checks for the following services:
```
'mobilityd',
'smsd',
'health',
'monitord',
'pkt_tester',
'redirectd',
```

This issue tracks the work for adding back `mobilityd`.
Please uncomment the line containing `mobilityd` in
`magma/lte/gateway/python/magma/tests/pylint_tests.py` and fix all pylint issues.

To run pylint test: 
```
# in vm
cd magma/lte/gateway
make test_python
# or as a shortcut
cd magma/lte/gateway
make build_python
. /home/vagrant/build/python/bin/activate; /home/vagrant/build/python/bin/nosetests -s magma/tests
```

sebathomas:
## Summary

Address pylint issues in mobilityd and enable automatic pylint runs. Most of the issues were related to string formatting in log messages and exceptions, or to differences in parameters between abstract base class methods and their implementations.

Fixes #6099.

## Test Plan

Executed unit tests in magma vm (`make test_python`).

### DIFF
--- a/lte/gateway/python/magma/tests/pylint_tests.py
+++ b/lte/gateway/python/magma/tests/pylint_tests.py
@@ -42,20 +42,15 @@ class MagmaPyLintTest(unittest.TestCase):
             ],
             show_categories=["warning", "error", "fatal"],
         )
-        # TODO look up directories in magma/lte/gateway/python/magma
-        directories = [
-            'enodebd',
-            'health',
-            # 'mobilityd',
-            'monitord',
-            'pipelined',
-            'pkt_tester',
-            'policydb',
-            'redirectd',
-            'smsd',
-            'subscriberdb',
+        excluded_directories = [
+            'kernsnoopd',
+            'tests',
         ]
         parent_path = os.path.dirname(os.path.dirname(__file__))
+        directories = [
+            d.name for d in os.scandir(parent_path)
+            if d.is_dir() and d.name not in excluded_directories
+        ]
         for directory in directories:
             path = os.path.join(parent_path, directory)
             py_wrap.assertNoLintErrors(path)
--- a/lte/gateway/python/magma/mobilityd/dhcp_client.py
+++ b/lte/gateway/python/magma/mobilityd/dhcp_client.py
@@ -233,7 +233,7 @@ class DHCPClient:
 
             # default in wait is 30 sec
             wait_time = max(wait_time, self._lease_renew_wait_min)
-            logging.debug("lease renewal check after: %s sec" % wait_time)
+            logging.debug("lease renewal check after: %s sec", wait_time)
             self._monitor_thread_event.wait(wait_time)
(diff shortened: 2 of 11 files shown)


lte/gateway/python/magma/tests/pylint_tests.py
```diff
@@ -42,11 +42,10 @@ def test_pylint(self):
             ],
             show_categories=["warning", "error", "fatal"],
         )
-        # TODO look up directories in magma/lte/gateway/python/magma
         directories = [
             'enodebd',
             'health',
-            # 'mobilityd',
+            'mobilityd',
```
themarwhal: I think it might be good to default to enabling pylint for all services instead of having to add any new services here. 

We can look at the directory structure to decide which subdirectories to run pylint for. For any directories that need fixups, we can opt out of the check.

themarwhal (approved):
thanks for cleaning up the service :)

lte/gateway/python/magma/tests/pylint_tests.py
sebathomas: Changed it so there is now a denylist.

lte/gateway/python/magma/tests/pylint_tests.py
```diff
@@ -13,6 +13,7 @@
 
 import os
 import unittest
+from glob import glob
```
themarwhal: looks like this is an actual lint

lte/gateway/python/magma/tests/pylint_tests.py
themarwhal: awesome thanks! this can be another PR, but can we do something similar for the same test in `orc8r/gateway/python/` ? (would be great if you could file an issue to track that as well.)

lte/gateway/python/magma/tests/pylint_tests.py
sebathomas: Oops

lte/gateway/python/magma/tests/pylint_tests.py
sebathomas: Added #9811 to track it.

sebathomas:
Rebased again, waiting for second review.

pshelar (approved):
Looks good. Thanks!

PR #9767, opened 2021-10-19
refactor(orc8r): Add obsidian middleware to properly log and report non-2xx responses

### LINKED ISSUE #9766: [orc8r] Add obsidian middleware to properly log and report non-2xx responses
### Summary 
* Currently, we use the go lang native `HttpError`, which is a terrible way to do this, because it relies on every Orc8r to use `HttpError`
* Instead, replicate the logging logic from the original error handler [`obsidian.HttpError`]( https://github.com/magma/magma/blob/5ad38a568320003c927aaf84e137022d6f3d737d/orc8r/cloud/go/obsidian/handler.go#L179-L193) and add into the middleware function [`Collect Stats`](https://github.com/magma/magma/blob/13b485a9b6da5874761919c840e4c0a8be81641d/orc8r/cloud/go/obsidian/server/server.go#L45-L46). 
* Here is more on error handling in the web framework echo [docs](https://echo.labstack.com/guide/error-handling/).
* Finally, remove the logging/reporting from HttpError.

### Rationale
Using middleware to log errors is more generalizable and allows for centralized HTTP error handling.

christinewang5:
## Summary
* Created Makefile in orc8r for better dev experience
* Added obsidian middleware to each service to properly log and report non-2xx responses 
* Removed old `obsidian.HttpError` pattern 

## Test Plan
1. Ran unit tests that evoke the middleware function `Logger` with various verbosity levels
newLogs

2. Checked `docker-compose logs -f controller` and sent erroneous requests from swagger ui 
service


orc8r/cloud/go/obsidian/server/metrics_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"magma/orc8r/cloud/go/services/configurator"
+	"magma/orc8r/cloud/go/services/configurator/test_init"
+	deviceTestInit "magma/orc8r/cloud/go/services/device/test_init"
+	"magma/orc8r/cloud/go/services/orchestrator/obsidian/handlers"
+	stateTestInit "magma/orc8r/cloud/go/services/state/test_init"
+)
+
+// TestLogger tests the error logging through the echo middleware.
+// Run `go test -v ./... --args --logtostderr -v=<verbosity>` to see logs
+func TestLogger(t *testing.T) {
```
hcgatewood: Hm so, this doesn't *programmatically* test anything (i.e. there aren't any asserts)

As an aside, this is why dependency injection is so useful -- you can assert that the logger actually did log what you want, programmatically

But anyways, I don't think there's a good way to make this test programmatically useful, so for now I think we should omit it. Lmk if you feel strongly though

orc8r/cloud/go/obsidian/server/metrics_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"magma/orc8r/cloud/go/serdes"
+	"magma/orc8r/cloud/go/services/configurator"
+	"magma/orc8r/cloud/go/services/configurator/test_init"
+	deviceTestInit "magma/orc8r/cloud/go/services/device/test_init"
+	"magma/orc8r/cloud/go/services/orchestrator/obsidian/handlers"
+	stateTestInit "magma/orc8r/cloud/go/services/state/test_init"
+)
+
+// TestLogger tests the error logging through the echo middleware.
+// Run `go test -v ./... --args --logtostderr -v=<verbosity>` to see logs
```
hcgatewood: Side note: you can achieve this programmatically e.g. https://github.com/magma/magma/blob/13b485a9b6da5874761919c840e4c0a8be81641d/orc8r/cloud/go/services/state/indexer/reindex/queue_sql_test.go#L39-L41

orc8r/cloud/go/obsidian/server/metrics_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/labstack/echo"
+	"github.com/stretchr/testify/assert"
+
+	"magma/orc8r/cloud/go/obsidian"
+	"magma/orc8r/cloud/go/obsidian/tests"
+	"magma/orc8r/cloud/go/serdes"
+	"magma/orc8r/cloud/go/services/configurator"
+	"magma/orc8r/cloud/go/services/configurator/test_init"
+	deviceTestInit "magma/orc8r/cloud/go/services/device/test_init"
```
hcgatewood: Nit: I feel mixed, but at the end of the day we've settled on doing `device_test_init` rather than `deviceTestInit` -- thought there are plenty of places in the codebase where that's not the case yet

orc8r/cloud/go/obsidian/server/server.go
```diff
@@ -99,8 +99,8 @@ func Start() {
 				tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
 				tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, // 4 HTTP2 support
 				tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
-				//tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
-				//tls.TLS_RSA_WITH_AES_256_CBC_SHA,
+				// tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
+				// tls.TLS_RSA_WITH_AES_256_CBC_SHA,
```
hcgatewood: Let's just go ahead and delete these, they've been commented-out for over 2 years

orc8r/cloud/go/service/service.go
```diff
@@ -165,9 +166,32 @@ func getEchoServerForOrchestratorService(serviceName string) (*echo.Echo, error)
...
+	return code >= http.StatusInternalServerError && code <= http.StatusNetworkAuthenticationRequired
+}
+
+// Logger is the middleware function
+func Logger(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		err := next(c)
+		if err != nil {
+			c.Error(err)
+			he, _ := err.(*echo.HTTPError)
```
hcgatewood: I don't think this is guaranteed -- i.e. you *can* return any type of error, we just happen to always return an echo.HTTPError

I would just pass the error directly to the logger -- and the `%s` directive should coerce it into its stringified form by calling `Error`

If that doesn't work, you can conditionally convert the err directly -- `errMsg := err.Error()`

orc8r/cloud/go/service/service.go
```diff
@@ -165,9 +166,32 @@ func getEchoServerForOrchestratorService(serviceName string) (*echo.Echo, error)
...
+
+			if isServerErrCode(he.Code) {
+				glog.Infof("REST HTTP Error: %s, Status: %d", he.Message, he.Code)
```
hcgatewood: Put the code first, then the msg -- will be easier to grep for and visually skim

orc8r/cloud/go/service/service_test.go
```diff
@@ -79,3 +93,44 @@ func TestServiceRun(t *testing.T) {
 	assert.Equal(t, protos.ServiceInfo_STOPPING, srv.State)
 	assert.Equal(t, protos.ServiceInfo_APP_UNHEALTHY, srv.Health)
 }
+
+// TestLogger tests the error logging through the echo middleware.
+// Run `go test -v ./... --args --logtostderr -v=<verbosity>` to see logs
+func TestLogger(t *testing.T) {
```
hcgatewood: Same idea for this test -- we don't have great DI right now, so this doesn't actually test anything programmatically, so let's default to leaving it out for now

hcgatewood (changes requested):
Looks great! Couple small things then we'll be ready to merge 💯🌟

orc8r/cloud/go/obsidian/server/metrics_test.go
christinewang5: Yup I agree, deleting it in the next commit....

orc8r/cloud/go/service/service.go
christinewang5: Sounds good, here are the new error logs:
newLogs

christinewang5:
Ok also I decided to just delete the logging from `obsidian.HttpError` rather than replace it completely, because it did some error formatting for certain errors and a bunch of tests were failing without them...

orc8r/cloud/docker/Makefile
```diff
@@ -0,0 +1,6 @@
+build:
```
hcgatewood: Nit: let's make this a separate PR. For that PR: can you change these build.py calls to be build.py --all --parallel

orc8r/cloud/go/obsidian/handler.go
```diff
@@ -176,18 +176,17 @@ func AttachHandlers(e *echo.Echo, handlers []Handler, m ...echo.MiddlewareFunc)
 	}
 }
 
-// HttpError wraps the passed error as an HTTP error.
+// HTTPError wraps the passed error as an HTTP error and reformats rpc errors.
 // Code is optional, defaulting to http.StatusInternalServerError (500).
-func HttpError(err error, code ...int) *echo.HTTPError {
+func HTTPError(err error, code ...int) *echo.HTTPError {
```
hcgatewood: Couple nits

- From offline discussion: let's make this `MakeHTTPError` or `NewHTTPError`
- "...wraps the passed error as an echo HTTP error..." (add "echo")
- Add a not in the doc comment that echo.NewHTTPError should be preferred where possible

orc8r/cloud/go/service/service.go
```diff
@@ -165,9 +166,31 @@ func getEchoServerForOrchestratorService(serviceName string) (*echo.Echo, error)
...
+
+// Logger is the middleware function
+func Logger(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		err := next(c)
+		if err != nil {
+			c.Error(err)
+			status := c.Response().Status
+			if isServerErrCode(status) {
+				glog.Infof("Error: %s", err)
```
hcgatewood:
```suggestion
				glog.Infof("HTTP server error: %s", err)
```

orc8r/cloud/go/service/service.go
```diff
@@ -165,9 +166,31 @@ func getEchoServerForOrchestratorService(serviceName string) (*echo.Echo, error)
...
+			} else {
+				glog.V(1).Infof("Error: %s", err)
```
hcgatewood:
```suggestion
				glog.V(1).Infof("HTTP error: %s", err)
```

orc8r/cloud/go/service/service.go
```diff
@@ -165,9 +166,31 @@ func getEchoServerForOrchestratorService(serviceName string) (*echo.Echo, error)
...
 	e.HideBanner = true
+	e.Use(Logger)
 	return e, nil
 }
 
+func isServerErrCode(code int) bool {
+	return code >= http.StatusInternalServerError && code <= http.StatusNetworkAuthenticationRequired
+}
+
+// Logger is the middleware function
```
hcgatewood:
```suggestion
// Logger is a middleware function that intelligently logs HTTP errors.
```

orc8r/cloud/go/obsidian/handler.go
```diff
@@ -176,18 +176,17 @@ func AttachHandlers(e *echo.Echo, handlers []Handler, m ...echo.MiddlewareFunc)
...
 		status = code[0]
 	}
-	// TODO(hcgatewood): we should be handling REST error logging and metrics via middleware
 	if isServerErrCode(status) {
-		glog.Infof("REST HTTP Error: %s, Status: %d", err, status)
+		glog.Infof("REST HTTP Error Status: %d, Message: %s", status, err)
 	} else {
-		glog.V(1).Infof("REST HTTP Error: %s, Status: %d", err, status)
+		glog.V(1).Infof("REST HTTP Error Status: %d, Message: %s", status, err)
 	}
```
hcgatewood: From offline: we don't want to do any logging from this func at all -- just coerce the error to echo.HTTPError (the final line of the fn)

hcgatewood (changes requested):
Looking good, couple nits, main thing is obsidian.MakeHTTPError should just be a one-liner

hcgatewood (approved):
Awesome! Excited for this, thanks for the fix @christinewang5 💯

hcgatewood:
Looks like cloud lint may be failing

Also, looks like CWF operator is failing... not super certain how that's happening. Try rebasing onto latest master first. If that doesn't work, try running those tests locally on your branch and see if they fail. If yes, try to debug / lmk what you find. If no, lmk and we can look into it together

themarwhal (approved):

PR #9782, opened 2021-10-20
ci(nms): Publish Testim reports to Slack

andreilee:

## Summary

Publishes reports to Magma's Slack whenever the daily Testim test plan finishes running.

**Test example**:
Screen Shot 2021-10-20 at 12 13 42 AM

First, a header is posted to channel `testim-reports` showing whether the daily tests passed or failed, and shows a count of how many passed. Then, a breakdown of the failed tests is shown. Finally, and not shown in the screenshot, a generated `.html` test report of the results is posted for more detailed viewing if desired.

## Test Plan

Used [`nektos/act`](https://github.com/nektos/act) to test the modified `testim-workflow.yml` locally:
```
➜  magma git:(1-10-testim-ci2) ✗ act -j testim_check_job --secret-file .secrets --insecure-secrets
...
[Testim Check/Testim Check]   ✅  Success - Run Daily NMS Tests
[Testim Check/Testim Check] ⭐  Run Post Test Report
...
[Testim Check/Testim Check]   ✅  Success - Post Test Report
[Testim Check/Testim Check] ⭐  Run Generate HTML Report
[Testim Check/Testim Check]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /Users/andreilee/magma/workflow/8] user=
[Testim Check/Testim Check]   ✅  Success - Generate HTML Report
[Testim Check/Testim Check] ⭐  Run Post HTML Report File
...
```
- Secrets were copied to a local file, `.secrets`, to be able to run with `nektos/act`
- A simplified test plan from the one that will be used in production was chosen for local testing

## Additional Information

### DIFF
--- a/.github/workflows/testim-workflow.yml
+++ b/.github/workflows/testim-workflow.yml
@@ -10,12 +10,37 @@ jobs:
   testim_check_job:
     runs-on: ubuntu-latest
     name: Testim Check
+    env:
+      MAGMA_ROOT: "${{ github.workspace }}"
     steps:
+      - uses: actions/checkout@v2
       - name: Setup Node v12
         uses: actions/setup-node@v2
         with:
           node-version: '12'
       - name: Install Testim CLI
         run: npm install -g @testim/testim-cli
+      - name: Setup pyenv
+        uses: "gabrielfalcao/pyenv-action@v8"
+        with:
+          default: 3.8.5
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.8.5'
+      - name: Install Python Dependencies
+        run: |
+          python3 -m pip install junit2html requests xmltodict
       - name: Run Daily NMS Tests
-        run: testim --token ${{ secrets.TESTIM_TOKEN }} --project ${{ secrets.TESTIM_PROJECT_ID }} --test-plan "Daily Test Plan"
+        run: testim --token ${{ secrets.TESTIM_TOKEN }} --project ${{ secrets.TESTIM_PROJECT_ID }} --test-plan "Daily Test Plan" --report-file ./tests.xml
+      - name: Post Test Report
+        if: always()
+        run: |
+          ${{ env.MAGMA_ROOT }}/ci-scripts/post_testim_report.py tests.xml "${{ secrets.TESTIM_SLACK_TOKEN }}" C02EU9X4DNV
+      - name: Generate HTML Report
+        if: always()
+        run: |
+          python3 -m junit2htmlreport tests.xml tests.html
+      - name: Post HTML Report File
+        if: always()
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/testim-workflow.yml
```diff
@@ -10,12 +10,34 @@ jobs:
...
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.8.5'
+      - name: Install Python Dependencies
+        run: |
+          python3 -m pip install junit2html requests xmltodict
       - name: Run Daily NMS Tests
-        run: testim --token ${{ secrets.TESTIM_TOKEN }} --project ${{ secrets.TESTIM_PROJECT_ID }} --test-plan "Daily Test Plan"
+        run: testim --token ${{ secrets.TESTIM_TOKEN }} --project ${{ secrets.TESTIM_PROJECT_ID }} --test-plan "Daily Test Plan" --report-file ./tests.xml || true
+      - name: Post Test Report
```
quentinDERORY, line 35: add` if: always()` on all steps after running the test so that it always post results if the test step fails

quentinDERORY (changes requested):
Instead of writing our own html generator, would it not be enough to construct a message in the worflow with the module `uses: actions/github-script@v3` (javascript code) and use the existing slack module to publish this message: `rtCamp/action-slack-notify@v2`.
It's currently how all others workflow are set up.

andreilee:
@quentinDERORY The html generator is just a separate downloaded dependency, which just processes the JUnit XML file output from Testim. The message shown in Slack in the PR description is also too rich in detail for `rtCamp/action-slack-notify@v2`. In hindsight, writing that script in `js` instead of `python` would have made this simpler as whole, but I don't think it's worth it to rewrite it now.

edit: I think it's important to include specifically which tests failed, and what the error messages are, as not everybody will have a Testim account to view detailed test results. Having all Magma contributors able to see at a glance what tests are failing can give good signal if you accidentally broke something from a recent PR.

.github/workflows/testim-workflow.yml
andreilee: I tried this and maybe I just couldn't get it working with `nektos/act`, because whenever any tests failed, the steps annotated with `always()` wouldn't run. So, I made it so that the `Run Daily NMS Tests` step always passes. There's that ` || true` at the end of the command that the step runs.

quentinDERORY: We kind of loose the information at a glance that the test failed... If everything else is working, I can ensure using always will work.

andreilee:
Added `if: always()` to steps running after triggering Testim 'test plan'.

quentinDERORY (approved):

PR #9787, opened 2021-10-20
chore: make agw-workflow callable as action

maxhbr:
## Summary

this allows to call the agw-workflow.yml via [reusing-workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows) from another run.

The call than can look like:
```
  agw_test_devcontainer:
    uses: magma/magma/.github/workflows/agw-workflow.yml@master
    needs: build_devcontainer_dockerfile
    with:
      devcontainer_image: ${{ needs.build_devcontainer_dockerfile.outputs.tags[0] }}
      force_run: true
```

This is the first part of https://github.com/magma/magma/pull/9723.

## Test Plan

The respective jobs from agw-workflow should run in CI.

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -2,6 +2,17 @@
 name: agw-workflow
 
 on:  # yamllint disable-line rule:truthy
+  workflow_call:
+    inputs:
+      devcontainer_image:
+        description: 'The tag of the devcontainer image that should be used'
+        required: false
+        type: string
+      force_run:
+        description: 'Whether the run should be forced, even if no changes in agw were found'
+        default: false
+        required: false
+        type: boolean
   push:
     branches:
       - master
@@ -18,7 +29,7 @@ jobs:
   path_filter:
     runs-on: ubuntu-latest
     outputs:
-      should_not_skip: ${{ steps.changes.outputs.filesChanged }}
+      should_not_skip: ${{ steps.changes.outputs.filesChanged || inputs.force_run }}
     steps:
       # Need to get git on push event
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


quentinDERORY (approved):

electronjoe:
As to your test Plan section of the PR Desc, this change triggered re-execution of the agw-workflow and that [worklfow is passing](https://github.com/magma/magma/actions/runs/1364229662).  I would personally consider that sufficient (and as a sanity check the phases are taking reasonable durations..).

.github/workflows/agw-workflow.yml
```diff
@@ -136,7 +147,7 @@ jobs:
         timeout-minutes: 10
         uses: addnab/docker-run-action@v2
         with:
-          image: ${{ env.DEVCONTAINER_IMAGE }}
+          image: ${{ inputs.devcontainer_image || env.DEVCONTAINER_IMAGE }}
```
electronjoe, line 150: For my education, I am assuming here that when `inputs.devcontainer_image` is omitted and `~nil`, this `||` is used to pull in `env.DEVCONTAINER_IMAGE`? That must be a GH Action behavioral thing? Or is this a common YAML processing strategy?  I would not have thought to try this.

maxhbr: That was suggested by @quentinDERORY in slack

maxhbr:
Thank you for the input. I have squased and rebased the PR.

.github/workflows/agw-workflow.yml
quentinDERORY: It's common, I use it in ansible too

jwagantall: In case a value is not provided, will devcontainer_image be treated as unset/undefined is it?

jwagantall (approved):

maxhbr:
Instead of reusing-workflows, the better solution is to use composite workflows, as demonstrated in https://github.com/magma/magma/pull/10232 or https://github.com/magma/magma/pull/10312/files

PR #9788, opened 2021-10-20
feat(agw): Inject crash reporter into agwd

mstre123:

## Summary

Inject crash reporter into agwd
Add wrap function to crash

## Test Plan

Test locally

## Additional Information


lte/protos/mconfig/mconfigs.proto
```diff
@@ -520,4 +520,9 @@ message AgwD {
   string sctpd_upstream_service_target = 3;
   string mme_sctpd_downstream_service_target = 4;
   string mme_sctpd_upstream_service_target = 5;
+
+  // Sentry Client Options.
+  // The DSN to use. If the DSN is not set, the client is effectively
+  // disabled.
+  string dsn = 6;
```
themarwhal: how will this be configured? So we have a plan to migrate all sentry config to be per-mconfig, not per-service inside the mconfig.  https://github.com/magma/magma/issues/9508

src/go/agwd/config/config.go
```diff
@@ -85,6 +85,8 @@ func newDefaultConfig() *mconfig.AgwD {
 		SctpdUpstreamServiceTarget:      "unix:///tmp/sctpd_upstream.sock",
 		MmeSctpdDownstreamServiceTarget: "unix:///tmp/mme_sctpd_downstream.sock",
 		MmeSctpdUpstreamServiceTarget:   "unix:///tmp/mme_sctpd_upstream.sock",
+		//This is essentially a noop - prevents crash reporter from panicing.
+		Dsn: "http://whatever@really.com/1337",
```
themarwhal: In C++ and Python, we don't intialize sentry if the DSN is an empty string. Could we do something like that?

src/go/agwd/server/server.go
```diff
@@ -127,7 +128,13 @@ func startSctpdUplinkServer(
 	go grpcServer.Serve(listener)
 }
 
-func Start(cfgr config.Configer, logger log.Logger) {
+func Start(cfgr config.Configer, logger log.Logger, crash crash.Crash) {
+	defer func() {
```
markjen: why not put this up in main.go?

also, maybe we should add a crash wrapper to `github.com/magma/magma/src/go/crash` like this:

```
func Wrap(crash Crash, fn func()) {
	defer func() {
		if err := recover(); err != nil {
			crash.Recover(err)
			crash.Flush(time.Second * 5)
		}
	}
	fn()
}
```

it would allow more things to be easily wrapped

src/go/agwd/server/server_notwindows_test.go
```diff
@@ -9,7 +9,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//+build !windows
+//go:build !windows
```
markjen: do we need both of these directive statements?

src/go/agwd/server/server_notwindows_test.go
```diff
@@ -21,6 +22,8 @@ import (
 	"time"
 
 	"github.com/golang/mock/gomock"
+	"github.com/magma/magma/src/go/agwd/config"
+	crash2 "github.com/magma/magma/src/go/crash/mock_crash"
```
markjen: do we not like the default `mock_crash` package name?

src/go/agwd/config/config.go
markjen: agreed; default value should be empty str

src/go/agwd/main.go
```diff
@@ -14,8 +14,10 @@ package main
 import (
 	"flag"
 
+	sentry2 "github.com/getsentry/sentry-go"
```
markjen: nit: i might allow this package to use `sentry` and our package to be `crash_sentry` or something

lte/protos/mconfig/mconfigs.proto
markjen: i would at least call this field `sentry_data_soruce_name`

we haven't gotten to figuring out mconfig stuff yet; should all services expect to get a shared config obj alongside their own config? (also will we support allowing services to override the shared config values?)

mstre123: Thanks @themarwhal hadn't seen that doc!

Since the src/go project is still in beta right now i think it's worth keeping its mconfigs separate from agw to allow us to experiment faster and ensure that were not impacting running services. in the future though i do see the value in a single config with per service overrides (for instance we may want to change the sample rate on a per service basis)

mstre123: @markjen agreed the name can be better but Dsn is the variable name that we are setting and to make it clear to the user what the config is for i'd keep it the same.
But prepending it with sentry works.
`sentry_dsn`

src/go/agwd/config/config.go
mstre123: was a placeholder from the unit tests - will move to empty string

src/go/agwd/main.go
mstre123: changed it

src/go/agwd/server/server_notwindows_test.go
mstre123: intelij auto import.... will fix

src/go/agwd/server/server_notwindows_test.go
mstre123: weird not sure how this got in here - will remove

src/go/agwd/server/server.go
mstre123: put it in server.go so that the crash reporter would be setup for anyone calling Start() but its a good idea to move it up to ensure we capture any panics that take place as part main.

I like the wrap function, will add it to crash and call it from main.

src/go/agwd/server/server_notwindows_test.go
mstre123: intelij reformats this everytime i commit - will remove manually

markjen (approved):

themarwhal (approved):

PR #9800, opened 2021-10-20
refactor: move out indexVersioner from jobQueue

reginawang3495:
## Summary
Move out `indexVersioner` from `jobQueue`.
## Test Plan
Run all orch8r tests in intelliJ

## Additional Information


orc8r/cloud/go/services/state/indexer/reindex/index_versioner.go
```diff
@@ -0,0 +1,203 @@
...
+
+	"magma/orc8r/cloud/go/services/state/indexer"
+	"magma/orc8r/cloud/go/sqorc"
+
+	"github.com/Masterminds/squirrel"
+	"github.com/pkg/errors"
+	"github.com/thoas/go-funk"
+)
+
+type indexVersionerImpl struct {
```
hcgatewood: I would just call this `versioner` -- and the interface `Versioner`, for conciseness. The "Index" is implicit from the pkg context, and also add a doc/description to the interface

orc8r/cloud/go/services/state/indexer/reindex/index_versioner.go
```diff
@@ -0,0 +1,203 @@
...
+	builder sqorc.StatementBuilder
+}
+
+func NewIndexVersioner(db *sql.DB, builder sqorc.StatementBuilder) IndexVersioner {
+	return &indexVersionerImpl{db: db, builder: builder}
+}
+
+func (i *indexVersionerImpl) GetIndexerVersions() ([]*indexer.Versions, error) {
+	txFn := func(tx *sql.Tx) (interface{}, error) {
+		return getIndexerVersionsImpl(i.builder, tx)
```
hcgatewood: Here and elsewhere: just call this `getIndexerVersions`, since it being private is sufficient / no need to add "Impl"

orc8r/cloud/go/services/state/indexer/reindex/index_versioner.go
```diff
@@ -0,0 +1,203 @@
...
+	composed, err := getComposedVersions(old)
+	if err != nil {
+		return nil, err
+	}
+	if EqualVersions(composed, old) {
+		return composed, nil
+	}
+
+	// Test hook after first db call so the tx has "officially" started by acquiring some locks
+	TestHookGet()
```
hcgatewood: Do we still need these hooks here? I recall adding them mostly to deal with the complexities of testing the queue

reginawang3495:
We can either move tests for `indexVersioner` out from the `jobQueue` now, or do it later when we completely remove `jobQueue` and leave `indexVersioner` behind. 

I prefer the second, but am not opposed to the first.

orc8r/cloud/go/services/state/indexer/reindex/queue_sql.go
```diff
@@ -94,6 +93,7 @@ type sqlJobQueue struct {
 	maxAttempts uint
 	db          *sql.DB
 	builder     sqorc.StatementBuilder
+	IndexVersioner
```
hcgatewood: Move this to the top of the struct -- is clearer

orc8r/cloud/go/services/state/indexer/reindex/queue_sql.go
```diff
@@ -116,7 +116,8 @@ type sqlJobQueue struct {
 //
 // Only provides Postgres support due to use of the non-standard "FOR UPDATE SKIP LOCKED" clause.
 func NewSQLJobQueue(maxAttempts uint, db *sql.DB, builder sqorc.StatementBuilder) JobQueue {
-	return &sqlJobQueue{maxAttempts: maxAttempts, db: db, builder: builder}
+	i := NewIndexVersioner(db, builder)
```
hcgatewood: `i` is a risky variable to use, cause people usually expect it to contain an integer. Instead, just skip the var entirely and create the object directly in the struct literal

hcgatewood (changes requested):
Looks great. Couple nits, main thing is: let's add some `Versioner`-specific tests. Feel free to use the existing tests as a starting point

orc8r/cloud/go/services/state/indexer/reindex/index_versioner.go
reginawang3495: I believe so, unless we mock out `Versioner` in `queue_sql_integ_test.go`, or move it to something in `queue_sql.go`. <`TestSQLReindexJobQueue_Integration_PopulateJobs` fails otherwise. I can diagnose the issue if you want me to though>

Since it's an integration test, I think it makes sufficient sense to leave as is?

hcgatewood: Ah right, so basically: if we're going to re-use the existing tests, then we need to keep this for now, with the idea that we can remove it once we remove the reindexer_queue impl

Anyways, re-request review / DM when ready for next round of review

orc8r/cloud/go/services/state/indexer/reindex/queue.go
```diff
@@ -55,6 +55,9 @@ type JobInfo struct {
 // State indexers are added to the queue for reindexing at initialization, and removed from the queue after the reindex is complete.
 // ClaimAvailableJob should be polled periodically, as jobs may become available at any time.
 type JobQueue interface {
+	// Tracks version info for all tracked indexers
```
hcgatewood, line 58: General nit: prepend with identifier name
```suggestion
	// Versioner tracks version info for all tracked indexers
```

orc8r/cloud/go/services/state/indexer/reindex/queue.go
```diff
@@ -73,6 +76,13 @@ type JobQueue interface {
 	// GetJobInfos provides full information about job progress, keyed by indexer ID.
 	// A job info only includes an error when its job has been attempted at least the max number of attempts.
 	GetJobInfos() (map[string]JobInfo, error)
+}
+
+// Tracks version info for all tracked indexers
+type Versioner interface {
+	// Initialize the Versioner.
+	// Call before other methods.
+	InitializeVersioner() error
```
hcgatewood: I'm like 60% sure there's a way you can have this be named just `Initialize` (remove the `Initialize` from the `JobQueue` interface, then call `versioner.Initialize` from `queue.Initialize`)

If not, add a todo for yourself to this method to rename it back to `Initialize` once we remove jobqueue

orc8r/cloud/go/services/state/indexer/reindex/versioner_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"magma/orc8r/cloud/go/services/state/indexer"
+	"magma/orc8r/cloud/go/services/state/indexer/mocks"
+	"magma/orc8r/cloud/go/services/state/indexer/reindex"
+	"magma/orc8r/cloud/go/sqorc"
+)
+
+func TestVersioner(t *testing.T) {
+	dbName := "state___versioner_test"
+	versioner := initVersioner(t, dbName)
```
hcgatewood: Since the versioner doesn't rely on postgres-specific syntax (just the job queue), we can just use an in-memory sqlite db, same as how you did for milestone 1

hcgatewood (changes requested):
Looks great, couple small things then I think we're there!

orc8r/cloud/go/services/state/indexer/reindex/queue.go
reginawang3495: I couldn't find one easily from stack overflow (I tried prior to you mentioning it), so I'll just add a TODO

hcgatewood: I think let's at least give it a shot. We can trial together during 1-1 if needed

hcgatewood: Did you try the solution I outlined?

reginawang3495: Now I did, and it looks like it should work if interfaces aren't used as per https://stackoverflow.com/questions/20861786/golang-struct-calling-embedded-type-methods-when-method-has-been-overloaded

In this case, it says `Duplicate method 'Initialize'` probably because interface method overloading

image

hcgatewood: Right, so remove the `Initialize` method from JobQueue, because it's included implicitly as part of the `Versioner` embedded iface

reginawang3495: `Initialize` is needed in JobQueue also though

orc8r/cloud/go/services/state/indexer/reindex/queue.go
hcgatewood: Nit: we would want this to be `// Versioner tracks version info...`

hcgatewood (approved):
Awesome!

PR #9803, opened 2021-10-21
fix(nms): Adjust certificate expiry alert threshold

andreilee:

## Summary

Fixes a small but potentially devastating typo in the alert rule which triggers if a certificate will expire within a month. The typo caused the alert to fire if the expiry was longer than a month away, and to not fire if the expiry was within a month.

## Test Plan

Checked by editing the alert rule in Magma's staging environment.

First found cert expiry data for `certifier.pem`:

Screen Shot 2021-10-20 at 6 41 06 PM

It can be seen that the expiry is still 2562047 hours away.

Here, the certificate expiry alert can be seen firing:
Screen Shot 2021-10-20 at 6 39 45 PM

Here are the new settings which correct the issue:

Screen Shot 2021-10-20 at 6 40 00 PM

Which simply swaps the `> 720` to `< 720`, which means the alert triggers when the expiry is within 720 hours, or, less than 30 days.

After a minute or two, checked the alerts again to see that there were no active alerts:
Screen Shot 2021-10-20 at 6 44 20 PM

Also checked that the alert would fire when the cert life goes under the specified time. Modified the rule to trigger alert if certificate life was < 9999999:
Screen Shot 2021-10-20 at 6 58 41 PM

And checked alerts again to see they were firing again:
Screen Shot 2021-10-20 at 6 59 05 PM

## Additional Information

### DIFF
--- a/nms/packages/magmalte/alerts/cwfAlerts.js
+++ b/nms/packages/magmalte/alerts/cwfAlerts.js
@@ -22,7 +22,7 @@ export default function getCwfAlerts(
   return {
     'Certificate Expiring Soon': {
       alert: 'Certificate Expiring Soon',
-      expr: `cert_expires_in_hours > 720`,
+      expr: `cert_expires_in_hours < 720`,
       labels: {severity: 'major'},
       annotations: {
         description: `Alerts when certificate necessary for Orc8r function is expiring soon`,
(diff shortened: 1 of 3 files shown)


HannaFar (approved):

sudhikan:
can we back port to 1.6 please?

cc: @ulaskozat @aswin-alexander

ulaskozat:
> can we back port to 1.6 please?
> 
> cc: @ulaskozat @aswin-alexander

this is already backported. https://github.com/magma/magma/commit/ef53c22e21ba7309e4c6ad3bd852ea2246a61c8b

PR #9823, opened 2021-10-21
chore: move functions from commoon_types.h/c to conversions.c/h

### LINKED ISSUE #9817: Dependency cycle(?) witthin common_types, log, and itti
Another dependency cycle detected.

1. `common/log` depends on `lib/itti/intertask_interface`
2. `lib/itti/intertask_interface.h` needs `imsi64_t` to be defined. This is not explicitly defined in any of the files that it depends on (maybe there is an implicit dependency somewhere), but I see that it's defined in `common/common_types.h` (https://github.com/magma/magma/blob/master/lte/gateway/c/core/oai/common/common_types.h#L118). This makes `lib/itti/intertask_interface.h` dependent on `common/common_types.h` .
3. `common/common_types.h` declares a few helper functions that are defined in `common/common_types.c`
4. `common/common_types.c` depends on `common/log.h`

This is a light dependency cycle and there are probably a few ways to fix this
1. move the type definition of imsi64_t to some other file
2. `common/common_types.c` seems to implement some functions defined in `common/common_types.h`. We can split `common/common_types.h` into 2 files, one with only type definitions and one with helper function declarations. This way we avoid step 3/4 in the above section.

I don't like option 1 as basic types as `imsi64_t` would make sense to stay in a file called `common_types.h`. I would prefer option 2 as it separates pure type definitions that are used by many files from a few helper functions that are used by a few files.

mme-dependency-cycle-imsi64-2

themarwhal:

## Summary
See discussion tracked here: https://github.com/magma/magma/issues/9817

Changes made
1. move function declarations from `c/core/oai/common/common_types.h` to `c/core/oai/common/conversions.h`
2. move function declarations from `c/core/oai/lib/3gpp/3gpp_23.003.h` to `c/core/oai/common/conversions.h`
3. move function definitions from `c/core/oai/common/common_types.c` to `c/core/oai/common/conversions.c`
4. address syntax error in `c/core/oai/oai_mme/CMakeLists.txt`

## Test Plan
make build_oai
make test_oai

## Additional Information


lte/gateway/c/core/oai/common/CMakeLists.txt
```diff
@@ -84,7 +83,7 @@ if (EXISTS "/etc/os-release")
   endif ()
 endif ()
 
-message("os_name", ${os_name})
+message("os_name" ${os_name})
```
themarwhal, line 86: Resolves the following error/warning

```
CMake Warning (dev) at common/CMakeLists.txt:86:
  Syntax Warning in cmake code at column 18

  Argument not separated from preceding token by whitespace.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

lte/gateway/c/core/oai/oai_mme/CMakeLists.txt
```diff
@@ -31,7 +31,6 @@ endif ()
 
 add_executable(mme
     ${PROJECT_SOURCE_DIR}/oai_mme/oai_mme.c
-    ${PROJECT_SOURCE_DIR}/common/common_types.c
```
themarwhal, line 34: i wonder why we needed this :o

ardzoht (approved):
lgtm

PR #9833, opened 2021-10-22
feat(orc8r): deIndex implementation

### LINKED ISSUE #8925: feat(orc8r): reconcile secondary indexes with state

uri200:

## Summary

Add implementation of deIndex to remove TEID-HwID mapping and SessinId-IMSI mappings

## Test Plan

```
./build.py -g
./build.py -c
```

Done manual test with local setup

Created Index with

```
grpcurl -vv \
    -insecure \
    -key /tmp/magma_protos/gateway.key \
    -cert /tmp/magma_protos/gateway.crt \
    -authority state-controller.magma.test \
    -protoset /tmp/magma_protos/out.protoset \
    -d '{"states": [{"type": "directory_record",
"deviceID": "IMSI001020000000064",
"value":"eyJpZGVudGlmaWVycyI6IHsic2d3X2NfdGVpZCI6ICIxMTY2NDEzNDkwIn0sICJsb2NhdGlvbl9oaXN0b3J5IjogWyJlODhlZThjYy1hYThiLTQxODktOGYxNC00MWQ4M2RlOWNlNGUiXSwgInB5L29iamVjdCI6ICJtYWdtYS5kaXJlY3RvcnlkLnJwY19zZXJ2aWNlci5EaXJlY3RvcnlSZWNvcmQifQ==", "version":"3"}]}' \
    localhost:7443 \
    magma.orc8r.StateService/ReportStates
```

Observed index is created
```
magma_dev=# SELECT network_id, type, key FROM directoryd_blobstore;
  network_id  |       type       |    key
--------------+------------------+------------
 feg_lte_test | sgwCteid_to_hwid | 1166413490
(1 row)
```

Deleted index with

```
grpcurl -vv \
    -insecure \
    -key /tmp/magma_protos/gateway.key \
    -cert /tmp/magma_protos/gateway.crt \
    -authority state-controller.magma.test \
    -protoset /tmp/magma_protos/out.protoset \
    -d '{"ids": [{"type": "directory_record",
"deviceID": "IMSI001020000000064"}]}' \
    localhost:7443 \
    magma.orc8r.StateService/DeleteStates
```

Observed index is gone

```
magma_dev=# SELECT network_id, type, key FROM directoryd_blobstore;
 network_id | type | key
------------+------+-----
(0 rows)
```

## Additional Information

### DIFF
--- a/orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
+++ b/orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
@@ -69,7 +69,7 @@ func (i *indexerServicer) Index(ctx context.Context, req *protos.IndexRequest) (
 	if err != nil {
 		return nil, err
 	}
-	stErrs, err := indexImpl(ctx, req.NetworkId, states)
+	stErrs, err := setSecondaryStates(ctx, req.NetworkId, states)
 	if err != nil {
 		return nil, err
 	}
@@ -78,7 +78,16 @@ func (i *indexerServicer) Index(ctx context.Context, req *protos.IndexRequest) (
 }
 
 func (i *indexerServicer) DeIndex(ctx context.Context, req *protos.DeIndexRequest) (*protos.DeIndexResponse, error) {
-	return &protos.DeIndexResponse{}, nil
+	states, err := state_types.MakeStatesByID(req.States, serdes.State)
+	if err != nil {
+		return &protos.DeIndexResponse{}, err
+	}
+	stErrs, err := unsetSecondaryStates(ctx, req.NetworkId, states)
+	if err != nil {
+		return nil, err
+	}
+	res := &protos.DeIndexResponse{StateErrors: state_types.MakeProtoStateErrors(stErrs)}
+	return res, nil
 }
 
 func (i *indexerServicer) PrepareReindex(ctx context.Context, req *protos.PrepareReindexRequest) (*protos.PrepareReindexResponse, error) {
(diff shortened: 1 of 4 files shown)


orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
```diff
@@ -134,8 +121,64 @@ func setSecondaryStates(ctx context.Context, networkID string, states state_type
 	return stateErrors, multiError.AsError()
 }
 
+// unsetSecondaryStates removes {sessionID -> IMSI} and {teid -> HwId} maps
+func unsetSecondaryStates(ctx context.Context, networkID string, states state_types.StatesByID) (state_types.StateErrors, error) {
+	sessionIDToIMSI, teidoHwId, stateErrors := getMappings(states)
```
hcgatewood, line 126: Typo/nit
```suggestion
	sessionIDToIMSI, teidToHWID, stateErrors := getMappings(states)
```

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
```diff
@@ -134,8 +121,64 @@ func setSecondaryStates(ctx context.Context, networkID string, states state_type
...
+		err := directoryd.UnmapSessionIDsToIMSIs(ctx, networkID, sessionIds)
+		if err != nil {
+			glog.Errorf("UnmapSessionIDsToIMSIs failed: %s", err)
+		}
+		return stateErrors, err
+	}
+	return stateErrors, nil
+}
+
+// getMappings builds SessionID to Imsi and Teid to Hwid maps from state
```
hcgatewood: Nit
```suggestion
// getMappings builds sessionID to IMSI and TEID to HWID maps from state
```

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer_test.go
```diff
@@ -121,15 +121,33 @@ func TestIndexerSessionID(t *testing.T) {
 	assert.Equal(t, imsi1, imsi)
 
 	// Err when can deserialize but is wrong type
-	id.Type = orc8r.GatewayStateType
-	st.ReportedState = &models.GatewayStatus{Meta: map[string]string{"foo": "bar"}}
-	errs, err = idx.Index(nid0, state_types.SerializedStatesByID{id: serialize(t, st, orc8r.GatewayStateType)})
+	st_2 := state_types.State{
```
hcgatewood: General nit:
```suggestion
	st2 := state_types.State{
```

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer_test.go
```diff
@@ -121,15 +121,33 @@ func TestIndexerSessionID(t *testing.T) {
...
 
+	// deIndex sessionID
+	st.ReportedState.(*directoryd_types.DirectoryRecord).Identifiers[directoryd_types.RecordKeySessionID] = sid0
+	errs, err = idx.DeIndex(nid0, state_types.SerializedStatesByID{id: serialize(t, st, orc8r.DirectoryRecordType)})
+	assert.NoError(t, err)
+	assert.Empty(t, errs)
+	_, err = directoryd.GetIMSIForSessionID(context.Background(), nid0, sid0)
+	assert.Error(t, err)
+
 	// ////////////////////////////
```
hcgatewood: General nit: not a fan of these types of block comments, would rather turn this into a sub-test with `t.Run`

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
```diff
@@ -134,8 +121,64 @@ func setSecondaryStates(ctx context.Context, networkID string, states state_type
 	return stateErrors, multiError.AsError()
 }
 
+// unsetSecondaryStates removes {sessionID -> IMSI} and {teid -> HwId} maps
```
hcgatewood: Nit
```suggestion
// unsetSecondaryStates removes {sessionID -> IMSI} and {TEID -> HWID} maps
```

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
```diff
@@ -134,8 +121,64 @@ func setSecondaryStates(ctx context.Context, networkID string, states state_type
...
+	}
+
+	if len(teids) != 0 {
+		err := directoryd.UnmapSgwCTeidToHWID(ctx, networkID, teids)
+		if err != nil {
+			glog.Errorf("UnmapSgwCTeidToHWID failed: %s", err)
+			return stateErrors, err
+		}
+	}
+	if len(sessionIds) != 0 {
```
hcgatewood: General nit:
```suggestion
	if len(sessionIDs) != 0 {
```

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
```diff
@@ -134,8 +121,64 @@ func setSecondaryStates(ctx context.Context, networkID string, states state_type
...
 
+// unsetSecondaryStates removes {sessionID -> IMSI} and {teid -> HwId} maps
+func unsetSecondaryStates(ctx context.Context, networkID string, states state_types.StatesByID) (state_types.StateErrors, error) {
+	sessionIDToIMSI, teidoHwId, stateErrors := getMappings(states)
+
+	teids := []string{}
+	for teid := range teidoHwId {
+		teids = append(teids, teid)
+	}
+	sessionIds := []string{}
```
hcgatewood: General nit: idiom when initializing empty slices is to use e.g. `var sessionIDs []string`

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
```diff
@@ -134,8 +121,64 @@ func setSecondaryStates(ctx context.Context, networkID string, states state_type
...
+	for sessionId := range sessionIDToIMSI {
+		sessionIds = append(sessionIds, sessionId)
+	}
+
+	if len(teids) != 0 {
+		err := directoryd.UnmapSgwCTeidToHWID(ctx, networkID, teids)
+		if err != nil {
+			glog.Errorf("UnmapSgwCTeidToHWID failed: %s", err)
+			return stateErrors, err
```
hcgatewood: I think we should have each type of secondary state be handled in its own sub-function. That way each sub-function can early-return when there's an error, while the unsetSecondaryStates function would instead *not* return early, instead exhaustively calling each sub-function and aggregating the errors

The reason to do this is so that an issue with e.g. unmapping the teids wouldn't cause an issue unmapping the hwids

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
```diff
@@ -134,8 +121,64 @@ func setSecondaryStates(ctx context.Context, networkID string, states state_type
...
+			continue
+		}
+		if params.sessionId != "" {
+			sessionIDToIMSI[params.sessionId] = params.imsi
+		}
+
+		for _, teid := range params.teids {
+			teidoHwId[teid] = params.hwid
+		}
+	}
```
hcgatewood, line 194: Up to you: for clarity, sometimes I like to do `getMappings` as a simple fn that loops over a sub-function called `getMapping` (singular) -- makes skimming `getMappings` easier

hcgatewood (approved):
Couple comments, lgtm though. Main thing is not early-returning during the top-level deindex attempt, and instead finishing the full set of unmap requests then aggregating their errs

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
uri200: We have mixture of HwID and HwID and HWID on orc8r. This file has all its vars as `tedi` and `HwId`

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
uri200: I try to avoid nil slices. You never know how they are going to be used in the future.

https://medium.com/@habibridho/golang-nil-vs-empty-slice-87fd51c0a4d

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
uri200: sure that make sense

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer_test.go
uri200: Sometimes a comment is worth 1000 line of code ;)

Refactored to look nicer

orc8r/cloud/go/services/orchestrator/servicers/indexer_servicer.go
hcgatewood: I personally don't view those as sufficient rationale -- deepequal should generally only be used in test code (and even then is not the current preference -- cmp is preferred), while json encoding is a specific use-case that doesn't affect the vast majority of code

In any case, the de facto preference is nil slice https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

hcgatewood: That said, is not end of the world, is just a nit 🙂

uri200: no problem. Let's make it beautiful then

Issue #9839, opened 2021-10-22
MME better eng: Run IWYU on MME


themarwhal:
## Repro steps (tested inside the devcontainer)
1. run make build_oai
2. while 1 is running install IWYU - https://include-what-you-use.org (checkout clang 11)
3. run `iwyu_tool.py -p /workspaces/magma/build/c/core/oai/compile_commands.json  /workspaces/magma/lte/gateway/c/core/oai` where -p points to the generated compile_commands.json from step 1. (`/workspaces/magma/build/c/core/oai/compile_commands.json`)
4. I filtered the output by running the following regex (`(?=The full include-list for \/workspaces\/magma\/lte)((.|\n)*?)(?=---)`

## IWYU result filtered as of Oct 22
[filtered-iwyu-output.txt](https://github.com/magma/magma/files/7398005/filtered-iwyu-output.txt)

electronjoe:
Related: #4868

electronjoe:
Just a note that iwyu is already installed in our devcontainer (from my prior work here) so I think step 2 isn't necessary if you specify that environment?

Issue #9819, opened 2021-10-21
Co-dependent files - itti/intertask_interface and common/shared_ts_log


themarwhal:
* `itti/intertask_interface.c` includes `common/shared_ts_log.h`
* `common/shared_ts_log.c` includes `itti/intertask_interface.h`

This seems like a weird pattern. 

![itti-shared_ts_log-dep](https://user-images.githubusercontent.com/37634144/138315123-45004b41-9f0b-4175-845e-6f3b550d0a6e.png)

PR #9840, opened 2021-10-22
chore: apply IWYU result to MME - part 1

themarwhal:

## Summary
I've been seeing inaccurate set of includes in MME, this causes confusion when trying to figure out the correct dependency to add for bazel. (For example, I thought intertask_interface.c depended on shared_ts_log.h, but it was just a misplaced #include https://github.com/magma/magma/issues/9819)

 I ran the Include-What-You-Use tool on MME (see result attched and repro steps here https://github.com/magma/magma/issues/9839), and this PR attempts to address the following directories

* `oai/common`
* `oai/lib/itti`
* `oai/lib/hashtable`
* `oai/lib/openflow`
* `oai/lib/pcef`

## Test Plan
make test_oai

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/common/TLVDecoder.c
+++ b/lte/gateway/c/core/oai/common/TLVDecoder.c
@@ -21,11 +21,9 @@
   \company Eurecom
 */
 
-#include <stdio.h>
-#include <stdint.h>
-
-#include "bstrlib.h"
 #include "TLVDecoder.h"
+#include <stdint.h>   // for uint16_t, uint32_t, uint8_t
+#include "bstrlib.h"  // for bstring, bcatcstr, bformat, bformata, bfromcstr
 
 int errorCodeDecoder = 0;
 
--- a/lte/gateway/c/core/oai/common/async_system.c
+++ b/lte/gateway/c/core/oai/common/async_system.c
@@ -34,22 +34,21 @@
    \date 2017
    \email: lionel.gauthier@eurecom.fr
 */
-#include <stdio.h>
-#include <stdlib.h>
-#include <stdbool.h>
-#include <stdarg.h>
 
-#include "bstrlib.h"
-#include "intertask_interface.h"
-#include "log.h"
 #include "async_system.h"
-#include "assertions.h"
-#include "dynamic_memory_check.h"
-#include "itti_free_defined_msg.h"
-#include "common_defs.h"
-#include "async_system_messages_types.h"
-#include "intertask_interface_types.h"
-#include "itti_types.h"
+#include <czmq_library.h>               // for zloop_t, zsock_t
+#include <pthread.h>                    // for pthread_exit
+#include <stdarg.h>                     // for va_end, va_list, va_start
+#include <stdbool.h>                    // for bool
+#include <stdio.h>                      // for NULL, perror
+#include <stdlib.h>                     // for free, exit, system
+#include "bstrlib.h"                    // for bdata, bfromcstralloc, btrunc
+#include "common_defs.h"                // for status_code_e, RETURNerror
+#include "dynamic_memory_check.h"       // for bdestroy_wrapper
+#include "intertask_interface.h"        // for DEPRECATEDitti_alloc_new_me...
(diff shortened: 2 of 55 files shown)


lte/gateway/c/core/oai/common/TLVDecoder.c
```diff
@@ -21,11 +21,9 @@
   \company Eurecom
 */
 
-#include <stdio.h>
-#include <stdint.h>
-
-#include "bstrlib.h"
 #include "TLVDecoder.h"
+#include <stdint.h>   // for uint16_t, uint32_t, uint8_t
+#include "bstrlib.h"  // for bstring, bcatcstr, bformat, bformata, bfromcstr
```
ssanadhya, line 26: Can we preserve the order of includes so that the standard libraries and local libraries are in separate blocks?

lte/gateway/c/core/oai/common/async_system.c
```diff
@@ -34,22 +34,22 @@
...
-#include "itti_types.h"
+#include <czmq_library.h>               // for zloop_t, zsock_t
+#include <pthread.h>                    // for pthread_exit
+#include <stdarg.h>                     // for va_end, va_list, va_start
+#include <stdbool.h>                    // for bool
+#include <stdio.h>                      // for NULL, perror
+#include <stdlib.h>                     // for free, exit, system
+#include "bstrlib.h"                    // for bdata, bfromcstralloc, btrunc
+#include "common_defs.h"                // for status_code_e, RETURNerror
+#include "czmq_library.h"               // for zloop_start
```
ssanadhya: Duplicate include, already included at line 39

lte/gateway/c/core/oai/common/TLVDecoder.c
themarwhal: do you want a space between the different sections?
so in this case it would look like
```
#include "TLVDecoder.h"

#include <stdint.h>   // for uint16_t, uint32_t, uint8_t

#include "bstrlib.h"  // for bstring, bcatcstr, bformat, bformata, bfromcstr
```

lte/gateway/c/core/oai/common/async_system.c
themarwhal: yes sorry about that, addressing the cpplint annotations

lte/gateway/c/core/oai/common/TLVDecoder.c
themarwhal: (@ssanadhya)

themarwhal: actually, it might be difficult to add a space between the include blocks automatically. So it will be a manual process of looking at the output that the tool gives. There is a separate tool that takes a look at the generated output by IWYU and applies it to the source files, but our builds are not reliable enough to handle that. (as in sometimes adding an extra include that seems to make sense breaks the build) (Adding `#include <liblfds710/lfds710_misc.h>` in intertask_interface.c breaks the build, for example)

The ordering that the tool uses is the following
1. corresponding header file for the .c/.cpp
2. c system
3. c++ system
4. other files

ssanadhya: @themarwhal, spacing would be nice, since it improves readability and we have been trying to enforce it in the current code.  Would it be possible to run the tool and then manually verify the extra includes to un-break the build?

themarwhal: Yes I can try using the tool - that might be easier once we apply the path fixups and C++/C header file differentiations. (cc @electronjoe)

electronjoe: Is it possible to upstream this ability to `iwyu` Marie? =D Sure would be nice to have it able to format our files without our intervention

themarwhal: I think the `fix_includes.py` tool supplied with IWYU already has this functionality: https://github.com/include-what-you-use/include-what-you-use/blob/master/README.md#applying-fixes

themarwhal:
closing for now as it conflics with scotts changes

PR #9850, opened 2021-10-24
fix(orc8r): remove min/max from  opc lte key swagger definition

uri200:

## Summary

It looks like previous swagger-ui was not checking the min/max length defined on swagger properly. Since we updated swagger ui lte federated network post method was failing. It looks like the new ui version does check the length. 

Length of the `lte_auth_op` is not properly defined there. Removing it from the definition same we have in the rest of `lte_auth_op` fields on swagger (like in [HSS](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=orc8r%2Fcloud%2Fswagger%2Fspecs%2Fpartial%2Ffeg.swagger.v1.yml&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=848&start_col=17&end_row=848&end_col=17))

## Test Plan
The request was taken properly
image

## Additional Information

### DIFF
--- a/lte/cloud/go/services/lte/obsidian/models/network_epc_configs_swaggergen.go
+++ b/lte/cloud/go/services/lte/obsidian/models/network_epc_configs_swaggergen.go
@@ -47,8 +47,6 @@ type NetworkEpcConfigs struct {
 
 	// lte auth op
 	// Required: true
-	// Max Length: 16
-	// Min Length: 15
 	// Format: byte
 	LteAuthOp strfmt.Base64 `json:"lte_auth_op"`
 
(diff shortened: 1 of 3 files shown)


themarwhal (approved):

themarwhal:
what is the length supposed to be? should we figure this out and add it back somehow?

uri200:
> what is the length supposed to be? should we figure this out and add it back somehow?

we can try to figure out but note none of the other keys in the system has any length set on swagger (like `lte_auth_amf`)

PR #9864, opened 2021-10-25
fix(agw): Enable opt-in based logging to Sentry

sebathomas:
Makes it possible to select only a subset of log messages to be
sent to Sentry instead of everything above a certain log level.

Includes example that use the feature in mobilityd. These should
be removed before merging.

### DIFF
--- a/lte/gateway/python/magma/mobilityd/main.py
+++ b/lte/gateway/python/magma/mobilityd/main.py
@@ -17,7 +17,7 @@ from typing import Any, Optional
 from lte.protos.mconfig import mconfigs_pb2
 from lte.protos.subscriberdb_pb2_grpc import SubscriberDBStub
 from magma.common.redis.client import get_default_client
-from magma.common.sentry import sentry_init
+from magma.common.sentry import SEND_TO_SENTRY, sentry_init
 from magma.common.service import MagmaService
 from magma.common.service_registry import ServiceRegistry
 from magma.mobilityd.ip_address_man import IPAddressManager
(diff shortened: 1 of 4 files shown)


lte/gateway/python/magma/mobilityd/rpc_servicer.py
```diff
@@ -89,6 +89,7 @@ def add_ip_block(self, ip_block):
         self._ip_address_man.add_ip_block(ip_block)
 
     @return_void
+    @log_error_sentry
```
sebathomas, line 92: This decorator should be set for all gRPC requests where uncaught exceptions should be logged in Sentry.

lte/gateway/python/magma/mobilityd/main.py
```diff
@@ -115,8 +115,9 @@ def main():
     service = MagmaService('mobilityd', mconfigs_pb2.MobilityD())
 
     # Optionally pipe errors to Sentry
-    sentry_init(service_name=service.name)
+    sentry_init(service_name=service.name, requires_opt_in=True)
 
+    logging.error("Explicit logging test", extra=SEND_TO_SENTRY)
```
sebathomas, line 120: Just for testing, remove before merging.

lte/gateway/python/magma/mobilityd/main.py
```diff
@@ -115,8 +115,9 @@ def main():
     service = MagmaService('mobilityd', mconfigs_pb2.MobilityD())
 
     # Optionally pipe errors to Sentry
-    sentry_init(service_name=service.name)
+    sentry_init(service_name=service.name, requires_opt_in=True)
```
sebathomas, line 118: Just for testing. Still unclear if opt-in is required for mobilityd.

sebathomas:
Abandoned in favor of #9974

PR #9867, opened 2021-10-25
fix(orc8r): change sgi_management_iface_ipv6_addr type of CIDR

pshelar:
This option needs subnet of SGi interface, so it needs type change.

## Additional Information


hcgatewood (approved):
Lgtm. Need to make the change at https://github.com/magma/magma/blob/6cc284c15c604d4ccc69e1a6423ec1a47a76edb3/lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml#L2031-L2035 though, then run `build.py --generate`

(The v1/swagger.yml file is generated)

PR #9872, opened 2021-10-26
feat(agw): Add service level memory and cpu metrics

### LINKED ISSUE #5378: [lte][agw] Add per-service metrics for CPU and memory usage

wallyrb:
## Summary

1. Add memory usage of magma@*,  stcpd and ovs services. Memory will be taken from systemctl property MemoryCurrent, only for services that have MemoryAccounting enabled.
2. Add memory percent usage for magma@*,  stcpd and ovs services. Calculation is based using MemoryCurrent(from 1) divided my MemoryLimit obtained from systemctl properties.
3.  Add cpu_percent using psutil using the MainPID for each of the service. MainPID is obtained from systemctl property.

## Test Plan

- Tested in master

image

image
image

- service303_cli.py metrics magmad https://www.internalfb.com/phabricator/paste/view/P464745922

## Additional Information


orc8r/gateway/python/magma/magmad/metrics.py
```diff
@@ -243,3 +252,47 @@ def monitor_unattended_upgrade_status():
         logging.debug('Unattended upgrade status is %d', status)
         UNATTENDED_UPGRADE_STATUS.set(status)
         yield from asyncio.sleep(POLL_INTERVAL_SECONDS)
+
+
+@asyncio.coroutine
+def _collect_service_metrics():
+    cmd = ["systemctl", "show", "magma@*", "--property=Names"]
+    output = subprocess.check_output(cmd)
```
ardzoht: rather than using subprocess, you can use 
```
process = psutil.Process(os.getpid())
print(process.memory_info().rss)  # bytes
```
to get the current mem usage of each process

ardzoht (changes requested):

orc8r/gateway/python/magma/magmad/metrics.py
wallyrb: thanks @ardzoht , actually I tried using memory_info().rss and memory_info().vms but the values didn't match in what systemctl was showing. For example below, memory_info is showing 403 MB, but systemctl has accounted 342M


```
vagrant@magma-dev-focal:~$ sudo service magma@mme status
● magma@mme.service - Magma OAI MME service
     Loaded: loaded (/etc/systemd/system/magma@mme.service; disabled; vendor preset: enabled)
     Active: active (running) since Tue 2021-10-26 07:17:58 UTC; 13h ago
    Process: 146425 ExecStartPre=/usr/bin/env python3 /usr/local/bin/generate_oai_config.py (code=exited, status=0/SUCCESS)
    Process: 146439 ExecStartPre=/usr/bin/env python3 /usr/local/bin/config_stateless_agw.py reset_sctpd_for_stateful (code=exited, status=0/SUCCESS)
   Main PID: 146443 (mme)
      Tasks: 28 (limit: 9511)
     Memory: 342.5M (min: 512.0M limit: 1.0G)
     CGroup: /system.slice/system-magma.slice/magma@mme.service
             └─146443 /usr/local/bin/mme -c /var/opt/magma/tmp/mme.conf -s /var/opt/magma/tmp/spgw.conf


>>> process = psutil.Process(pid=146443)
>>> process.memory_info()
pmem(rss=423452672, vms=21990709542912, shared=68476928, text=48164864, lib=0, data=15393784320000, dirty=0)
>>> 423452672/1024/1024
403.8359375
>>> 21990709542912/1024/1024
20971974.890625

```

wallyrb: I wanted to match what systemctl was showing, since the memory limits are set there.

ardzoht: Ok sounds good, for the purposes of unblocking this I think we can start with calling subprocess, however I think we can use `pystemd` (which is already installed) to talk directly to systemd instead of using commands, could you please add a TODO to mark that we plan to migrate this using pystemd library?

ardzoht (approved):
lgtm, let's mark migration to pystemd as a TODO as a follow-up

orc8r/gateway/python/magma/magmad/metrics.py
wallyrb: noted, I added the TODO mark and will look into the psystemd library. Thanks @ardzoht !

orc8r/gateway/python/magma/magmad/metrics.py
```diff
@@ -243,3 +252,49 @@ def monitor_unattended_upgrade_status():
...
         UNATTENDED_UPGRADE_STATUS.set(status)
         yield from asyncio.sleep(POLL_INTERVAL_SECONDS)
+
+
+@asyncio.coroutine
+def _collect_service_metrics():
+    cmd = ["systemctl", "show", "magma@*", "--property=Names"]
+    # TODO(@wallyrb): Move away from subprocess and use psystemd
+    output = subprocess.check_output(cmd)
+    magma_services = str(output, "utf-8").strip().replace("Names=", "").split("\n\n")
```
ardzoht: magma_services can be retrieved from `service.config['magma_services'` and then attaching `magma@<service>` prefix, check line 124 on this file

orc8r/gateway/python/magma/magmad/metrics.py
```diff
@@ -87,6 +90,13 @@
 )
 
 
+SERVICE_METRICS = Gauge(
```
ardzoht: one more nit: I think we should separate these into `SERVICE_CPU_USAGE`, `SERVICE_MEMORY_USAGE`, `SERVICE_MEMORY_PERCENTAGE` from prometheus view is better to keep separate metrics with values rather than separating the metrics using labels

ardzoht (changes requested):
we should separate `SERVICE_METRICS` labels into separate metrics themselves

orc8r/gateway/python/magma/magmad/metrics.py
wallyrb: Thanks @ardzoht for the suggestion. I have updated the PR

ardzoht (approved):
lgtm, thanks for this!

hcgatewood (approved):
Thanks for the addition Wally!

@wallyrb @ardzoht any interest in a ready-made solution for the next iteration of this? https://github.com/povilasv/systemd_exporter

ardzoht:
> Thanks for the addition Wally!
> 
> @wallyrb @ardzoht any interest in a ready-made solution for the next iteration of this? https://github.com/povilasv/systemd_exporter

@hcgatewood Yeah that should be the definitive option to migrate in the future, thanks for bringing that up!

PR #9873, opened 2021-10-26
chore(magmad): Expand magmad service restart error list

koolzz:

## Summary
- Connectiond fails to start on agw reboot due to recieved sigterm, this remediates it
- Sometimes the err code is different and I'm not quite sure why I specified the DEADLINE_EXCEEDED in the initial iteration. 
- This should remediate from critical issues such as service being down in all cases, not just DEADLINE_EXCEEDED.

## Test Plan

- Tested with connectiond

## Additional Information

y I specified the DEADLINE_EXCEEDED in the initial iteration. 
- This should remediate from critical issues such as service being down in all cases, not just DEADLINE_EXCEEDED.

## Test Plan

- Tested with connectiond

## Additional Information

### DIFF
--- a/orc8r/gateway/python/magma/magmad/service_poller.py
+++ b/orc8r/gateway/python/magma/magmad/service_poller.py
@@ -183,5 +183,4 @@ class ServicePoller(Job):
                     err.code(),
                     err.details(),
                 )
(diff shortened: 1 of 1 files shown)


ulaskozat:
can you add to the summary the issue and how this PR fixes it?

PR #9892, opened 2021-10-26
feat(orc8r): add bootstrap metrics to determine which networks are active

### LINKED ISSUE #9906: [orc8r] Add bootstrap metrics to determine which networks are active
### Summary
* Created Prometheus metric to monitor the networks with the labels
  * network_id
  * hardware_id
  * successful 

### Rationale
* We want to know, on a per-network basis, the rate at which gateways are checking in — that way, we can determine whether a network is dead vs. actively used.
* Gateways have to re-bootstrap every ~4 days to receive new session certificates, so we can just increment a per-network counter every time a gateway successfully bootstraps ([bootstrap overview](https://magma.github.io/magma/docs/next/orc8r/dev_security#gateway-bootstrap))([code ref](https://github.com/magma/magma/blob/72b1f20e85dcf83fe2d16ca38408bb52d56ce5c3/orc8r/cloud/go/services/bootstrapper/servicers/bootstrapper.go#L141-L145))

christinewang5:
## Summary
* We want to know, on a per-network basis, the rate at which gateways are checking in — that way, we can determine whether a network is dead vs. actively used.
* Gateways have to re-bootstrap every ~4 days to receive new session certificates, so we can just increment a per-network counter every time a gateway successfully bootstraps ([bootstrap overview](https://magma.github.io/magma/docs/next/orc8r/dev_security#gateway-bootstrap))([code ref](https://github.com/magma/magma/blob/72b1f20e85dcf83fe2d16ca38408bb52d56ce5c3/orc8r/cloud/go/services/bootstrapper/servicers/bootstrapper.go#L141-L145))

## Test Plan
* manually triggering bootstrapping by deleting session certs
  1. `vagrant ssh magma`
  2. `cd lte/gateway`
  3. `sudo rm /var/opt/magma/certs/gateway.{crt,key}`
  4. `make restart`

* checked prometheus to see if logged successfully (deleted twice)
image


orc8r/cloud/go/services/bootstrapper/servicers/bootstrapper.go
```diff
@@ -50,6 +51,20 @@ const TimeLength = 8 // length of time encoded in byte array from int64
 const MinKeyLength = 1024
 const GatewayCertificateDuration = time.Hour * 97 // 4 days, lifetime of GW Certificate
 
+var (
+	bootstrapAttempts = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "network_bootstrap_attempts",
+			Help: "Number of bootstrap attempts for each network",
```
hcgatewood: Currently this is number of *successful* attempts

orc8r/cloud/go/services/bootstrapper/servicers/bootstrapper.go
```diff
@@ -142,6 +157,8 @@ func (srv *BootstrapperServer) RequestSign(ctx context.Context, resp *protos.Res
 	if err != nil {
 		return nil, errorLogger(status.Errorf(codes.Aborted, "Failed to sign csr: %s", err))
 	}
+	entity, _ := configurator.LoadEntityForPhysicalID(strippedIncomingCtx(ctx), hwId, configurator.EntityLoadCriteria{}, serdes.Entity)
+	bootstrapAttempts.WithLabelValues(entity.NetworkID, hwId).Inc()
```
hcgatewood: - We can use the hwid from the `resp` variable, so no need for a configurator call
- Since successful bootstrap happens per-gateway once every ~3 days, let's also add a .Info log

hcgatewood (changes requested):
Looks great! 2 small things, and then: for testing, can you delete the session certs twice, so we see the counter get incremented twice -- just to double check 🙂 

Thanks Christine!

hcgatewood:
cc @talkhasib

orc8r/cloud/go/services/bootstrapper/servicers/bootstrapper.go
christinewang5: I'm already using the `hwid` from the `resp` variable, the configurator call is for network ID and am not sure if there are alternatives to this but was just looking at examples. 


Also I wasn't sure whether to use `glog` or `log`, and bc `errorLogger` in this file seems to use the latter I used it here too...

hcgatewood: - Ah I see, sorry I misread -- this is indeed the correct way to do this 🙂 
- As for log -- `log` is legacy for our codebase, prefer `glog`
- Finally, handle the err case for loading the entity -- probably just glog the error

hcgatewood (changes requested):

orc8r/cloud/go/services/bootstrapper/servicers/bootstrapper.go
```diff
@@ -142,6 +157,12 @@ func (srv *BootstrapperServer) RequestSign(ctx context.Context, resp *protos.Res
 	if err != nil {
 		return nil, errorLogger(status.Errorf(codes.Aborted, "Failed to sign csr: %s", err))
 	}
+	entity, err := configurator.LoadEntityForPhysicalID(strippedIncomingCtx(ctx), hwId, configurator.EntityLoadCriteria{}, serdes.Entity)
+	if err != nil {
+		return nil, errorLogger(status.Errorf(codes.NotFound, "Gateway with hwid %s is not registered: %s", hwId, err))
```
hcgatewood: Nit
```suggestion
		return nil, errorLogger(status.Errorf(codes.NotFound, "Gateway with HWID %s is not registered: %+v", hwId, err))
```

orc8r/cloud/go/services/bootstrapper/servicers/bootstrapper.go
```diff
@@ -24,12 +24,13 @@ import (
 	"crypto/x509"
 	"encoding/binary"
 	"fmt"
-	"log"
 	"math/big"
 	"time"
 
+	"github.com/gogf/gf/os/glog"
```
hcgatewood: Wrong glog 🙂 

```suggestion
	"github.com/golang/glog"
```

hcgatewood (changes requested):

orc8r/cloud/go/services/bootstrapper/servicers/bootstrapper.go
christinewang5: oops thanks

hcgatewood (approved):
💯 🔥

hcgatewood:
@christinewang5 looks like merge conflict

PR #9909, opened 2021-10-26
fix(agw): vagrant init error on second magma vm provision.

pshelar:
Rather than adding ip address replace it to avoid
-EEXIST error.

## Test Plan
`vagrant provision magma`

## Additional Information


themarwhal (approved):

themarwhal:
isn't this now done in ansible?

pshelar:
> isn't this now done in ansible?

yea, but this is kept here to handle `vagrant reload` case.

PR #9921, opened 2021-10-27
fix(mme): Fix for handling NAS authentication failure

pruthvihebbani:
Signed-off-by: Pruthvi Hebbani <pruthvi.hebbani@radisys.com>

fix(mme): Fix for handling NAS authentication failure

## Summary

In case of NAS authentication failure, MME was triggering UE context release command twice. Once as part of handling attach failure and then during implicit detach. This PR fixes the issue.

## Test Plan

- Verified test_no_auth_response.py TC to ensure that MME sends UE context release command only once as part of handling attach failure
- Verified s1ap tester sanity

### DIFF
--- a/lte/gateway/c/core/oai/tasks/nas/emm/Authentication.c
+++ b/lte/gateway/c/core/oai/tasks/nas/emm/Authentication.c
@@ -1133,12 +1133,6 @@ status_code_e mme_app_handle_auth_t3460_expiry(
       // abort ANY ongoing EMM procedure (R10_5_4_2_7_b)
       nas_delete_all_emm_procedures(emm_ctx);
 
-      // Clean up MME APP UE context
-      memset((void*) &emm_sap, 0, sizeof(emm_sap));
-      emm_sap.primitive = EMMCN_IMPLICIT_DETACH_UE;
-      emm_sap.u.emm_cn.u.emm_cn_implicit_detach.ue_id = ue_id;
-      emm_sap_send(&emm_sap);
-      increment_counter("ue_attach", 1, 1, "action", "attach_abort");
     }
   }
   OAILOG_FUNC_RETURN(LOG_NAS_EMM, RETURNok);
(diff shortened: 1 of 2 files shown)


lte/gateway/c/core/oai/tasks/nas/emm/Authentication.c
```diff
@@ -1132,13 +1132,6 @@ status_code_e mme_app_handle_auth_t3460_expiry(
 
       // abort ANY ongoing EMM procedure (R10_5_4_2_7_b)
       nas_delete_all_emm_procedures(emm_ctx);
-
-      // Clean up MME APP UE context
-      memset((void*) &emm_sap, 0, sizeof(emm_sap));
-      emm_sap.primitive = EMMCN_IMPLICIT_DETACH_UE;
```
ssanadhya, line 1138: @pruthvihebbani how is the UE context release command getting triggered without Implicit Detach? The Common Proc Abort message in Deregistered mode will not trigger it. It will just log that this primitive is not valid (https://github.com/magma/magma/blob/99b136382f1f2e4c17d4f6347f59b62558fb5641/lte/gateway/c/core/oai/tasks/nas/emm/sap/EmmDeregistered.c#L118)

Are you seeing the UE context release as a result of the SCTP shutdown at the end of the test case?

pruthvihebbani: @ssanadhya  Its not triggered due to sctp shutdown. emm_as_establish_rej in emm_as.c sends out attach reject message and returns AS_TERMINATED_NAS to indicate that UE context release cmd has to be sent due to attach failure. Based on the error code returned, mme_app_handle_nas_dl_req is invoked which sends UE context release cmd.

ssanadhya: Thanks for the clarification @pruthvihebbani . Can you verify this for other timer handlers for T3460 and T3470 too? You can use the test cases added in https://github.com/magma/magma/pull/9942.

ssanadhya (approved):

PR #9936, opened 2021-10-27
feat(orc8r): add a reindexer singleton

### LINKED ISSUE #10135: Make reindexer a singleton service

reginawang3495:
## Summary
Add in a reindexer that runs as a singleton

### Next PR in Stack
- Save jobs/indexes # times run

## Test Plan
Run created test

## Additional Information


orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	return batches
+}
+
+func (r *reindexerSingleton) reindexJobs(ctx context.Context, indexerID string, batches []reindexBatch, sendUpdate func(string)) error {
+	jobs, err := r.getJobs(indexerID)
+	if err != nil {
+		return wrap(err, ErrDefault, indexerID)
+	}
+	// QUESTION: why return error if no jobs? this is brought over from Run / RunUnsafe of reindexer_queue
```
hcgatewood: Was probably to indicate whether some jobs were successfully reindexed -- if no, sleep longer before trying again

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// Single indexer
+	idx0 := getIndexer(id0, zero, version0, true)
+	idx0.On("GetTypes").Return(allTypes).Once()
+	// Register indexers
+	register(t, idx0)
+
+	// Check
+	recvCh(t, ch)
+	// QUESTION: is this not recommended? We check that nothing else is sent for a whole second.
```
hcgatewood: Hm. It's not the end of the world since tests get parallelized. But I think in this case it's probably not necessary -- just let the subsequent recvCh handle the expectation

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+// TODO(reginawang3495): Remove "Desired" from Indexer.Versions once reindexer_queue and queue are removed
+
+// This Reindexer runs as though it is a singleton
+type reindexerSingleton struct {
+	versioner Versioner
+	store     Store
+	// TODO(reginawang3495) Add some structure to save jobs/indexes # times run
+}
+
+// TODO: to be setup and used in M2 Part C
```
hcgatewood: Nit: always add a gh username in committed todos

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+const reindexLoopInterval = 5 * time.Second
+
+func NewReindexerSingleton(store Store, versioner Versioner) Reindexer {
+	return &reindexerSingleton{store: store, versioner: versioner}
+}
+
+func (r *reindexerSingleton) Run(ctx context.Context) {
+	const indexerID = ""
+	var sendUpdate func(string) = nil
```
hcgatewood, line 49: The reason we have a sendUpdate is for use from the indexers CLI https://github.com/magma/magma/tree/13b485a9b6da5874761919c840e4c0a8be81641d/orc8r/cloud/go/tools/indexers

So add a todo here -- we do want to support a non-nil sendUpdate from Run, but we can wait till we remove the queue impl

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+
+		err := r.reindexJobs(ctx, indexerID, batches, sendUpdate)
+
+		// NOTE: consider sleeping when no err for perf reasons
+		if err != nil {
+			clock.Sleep(failedJobSleep)
+		} else {
+			clock.Sleep(reindexLoopInterval)
+		}
```
hcgatewood:
```suggestion
		err := r.reindexJobs(ctx, indexerID, batches, sendUpdate)
		if err != nil {
			clock.Sleep(failedJobSleep)
		}
		
		clock.Sleep(reindexLoopInterval)
```

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func (r *reindexerSingleton) RunUnsafe(ctx context.Context, indexerID string, sendUpdate func(string)) error {
+	batches := r.getReindexBatches(ctx)
+	return r.reindexJobs(ctx, indexerID, batches, sendUpdate)
+}
+
+func (r *reindexerSingleton) GetIndexerVersions() ([]*indexer.Versions, error) {
+	return r.versioner.GetIndexerVersions()
+}
```
hcgatewood: As I understand it, you can just leave this out and it will work implicitly

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+
+// Get network-segregated reindex batches with capped number of state IDs per batch.
```
hcgatewood: Nit: prefix doc comment with identifier -- so `// getReindexBatches ...`

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+
+// TODO: to be setup and used in M2 Part C
+const reindexLoopInterval = 5 * time.Second
+
+func NewReindexerSingleton(store Store, versioner Versioner) Reindexer {
+	return &reindexerSingleton{store: store, versioner: versioner}
+}
+
+func (r *reindexerSingleton) Run(ctx context.Context) {
+	const indexerID = ""
```
hcgatewood, line 47: Optional: isn't immediately obvious why you'd set this to empty string, so either change var name or add comment to make it clear

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+func (r *reindexerSingleton) reindexJob(job *Job, ctx context.Context, batches []reindexBatch, sendUpdate func(string)) error {
+	defer TestHookReindexDone()
+
+	start := clock.Now()
+
+	jobErr := executeJob(ctx, job, batches)
+
+	// TODO(reginawang3495) Add indexer fail count increase and logging
+	// Set Indexer Version every time for now
+	err := r.versioner.SetIndexerActualVersion(job.Idx.GetID(), job.To)
```
hcgatewood: Only set actual version if there's no jobErr

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	// QUESTION: why return error if no jobs? this is brought over from Run / RunUnsafe of reindexer_queue
+	if len(jobs) == 0 {
+		return merrors.ErrNotFound
+	}
+
+	for _, j := range jobs {
+		err = r.reindexJob(j, ctx, batches, sendUpdate)
+		if err != nil {
+			return wrap(err, ErrDefault, indexerID)
```
hcgatewood: Let's use a multierror here instead of an early return. That is, just because one reindex job fails, that doesn't mean we should give up on the others

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+	idxs, err := getIndexersFromRegistry(indexerID)
+
+	if err != nil {
+		return nil, err
+	}
+
+	var ret []*Job
+	for _, x := range idxs {
+		// Get Indexer Version saved from memory
+		v, err := GetIndexerVersion(r.versioner, x.GetID())
```
hcgatewood: This version is actually the one stored in the db

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+		err = r.reindexJob(j, ctx, batches, sendUpdate)
+		if err != nil {
+			return wrap(err, ErrDefault, indexerID)
+		}
+	}
+	return nil
+}
+
+func (r *reindexerSingleton) reindexJob(job *Job, ctx context.Context, batches []reindexBatch, sendUpdate func(string)) error {
+	defer TestHookReindexDone()
```
hcgatewood, line 116: Okay so from offline: I think we can get rid of all the hooks except this one. Keeping this one though will make it easier to test things e2e, by giving us a clear way to understand when a single reindexJob has completed

We can remove the success hook though, and instead add a `reindexer_singleton_internal_test.go` in the `reindex` pkg that tests `reindexJob` in more detail

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,185 @@
...
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reindex
```
hcgatewood, line 14: https://magma.github.io/magma/docs/next/contributing/contribute_conventions#:~:text=Default%20to%20using%20separate%20_test%20package%20for%20tests

hcgatewood (changes requested):
Okay this is looking real good I'd say. Main thing is let's (1a) remove the success hook and (1b) test reindexJob directly as well

Other than that, looks fantastic and excited to get it merged 🙂

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
reginawang3495: For the sake of not redoing this infinite number of times, I purposefully set actual version even if `jobErr`. 

This is effectively making job retry count = 1. This will be modified with the TODO in the next PR

Thoughts?

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
reginawang3495: Oh since we are resetting the index between each test piece, the next emitted `ch <- nil` from the past test might've normally run and not have been tested. This is something that sometimes happened while testing bad code. 

Kinda confusing, I can explain better if that's not sufficient/any thoughts?

hcgatewood: Not following, can you explain more here or in 1-1

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
hcgatewood: Hm. Correct me if I'm wrong, but this introduces a bug -- yes we won't retry at all, but also it will incorrectly mark jobs as completed even when they're not (?)

If that's the case, then I think we shouldn't commit the code as-is

reginawang3495: Yes it'll mark the jobs as completed even when they aren't -- Last time we had a sync up, I remember we said that this (marking jobs as complete) is the expected behavior once the `numTimesRetries >= maxRetries`. In this case, I am simply pretending that `maxRetries == 0` 

If this isn't correct, then I can change this

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
reginawang3495: ## Example
For the same test case response where `reindexer_singleton` sends `ch -> nil` twice (where it should only send once), stack of code being run could be:
### In case where we have `recvNoCh(t, ch)`
1) `register(t, idx0)`
2) `ch-> nil` (received by `recvCh(t, ch)`)
3) `ch -> nil` (received by `recvNoCh(t, ch)`) ==> this causes error, as expected

### In case where we don't have `recvNoCh(t, ch)`
1) `register(t, idx0)`
2) `ch-> nil` (received by `recvCh(t, ch)`)
...
3)  `register(t, idx0a)`

Results in no error since the second `ch-> nil` would not have time to be sent before `register(t, idx0a)` happened

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
hcgatewood: Ah right. So difference between "job complete" and "job successful"

Since we don't have a job queue anymore, we don't need to distinguish. So we *don't* want to update an indexer's version unless it was successfully reindexed

I think actually, we can go ahead and skip that concept we talked about of tracking numRetries -- it's fine if we endlessly retry to reindex a broken indexer, we just need to add a long sleep between attempts

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
hcgatewood: Ah I see. Okay that makes sense, let's keep it then

P.S. feel free to re-request a review when ready

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
reginawang3495: Now thinking about it, why is `failedJobSleep` even different from `reindexLoopInterval`? Maybe there should be a set cycle of sleeping no matter whether a job failed? EG

```
err := r.reindexJobs(ctx, indexerID, batches, sendUpdate) 
// note: no if err !- nil statement
clock.Sleep(reindexLoopInterval)
```

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
reginawang3495: I changed the struct to composition!

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
reginawang3495: ? I think this isn't the test package? I do name the package of the test `reindex_test`? Don't quite understand what we want to change

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
reginawang3495: I don't think `reindexJob` is complex enough to require such in depth testing? I feel like `TestHookReindexSuccess` sufficiently tests the cases? I understand the benefits of testing out smaller components, but I think that `reindexJobs` is so simple that the benefits are marginal and not worth the much larger effort needed in having specific `reindexJob` tests

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
hcgatewood: Well, I completely misread that LOL. Please disregard 🙂

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,176 @@
...
+
+// TODO(reginawang3495): to be setup and used in M2 Part C
+const reindexLoopInterval = 5 * time.Second
+
+func NewReindexerSingleton(store Store, versioner Versioner) Reindexer {
+	return &reindexerSingleton{store: store, Versioner: versioner}
+}
+
+func (r *reindexerSingleton) Run(ctx context.Context) {
+	// indexerID being "" means that we are retrieving jobs on all indices
```
hcgatewood:
```suggestion
	// indexerID being "" means that we are reindexing all indexers
```

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,176 @@
...
+	return batches
+}
+
+func (r *reindexerSingleton) reindexJobs(ctx context.Context, indexerID string, batches []reindexBatch, sendUpdate func(string)) error {
+	jobs, err := r.getJobs(indexerID)
+	if err != nil {
+		return wrap(err, ErrDefault, indexerID)
+	}
+
+	var errs *multierror.Error
```
hcgatewood: Ah, I thought you needed to initialize multierrors -- but looks like you can just declare them!

For a subsequent side PR outside the project -- you could fix all the initializations in the codebase to be just declarations (so fixing them to follow the way you did it). E.g. https://github.com/magma/magma/blob/ff012313835a8ecb0cd4b9ac2055186468295838/lte/cloud/go/services/subscriberdb_cache/config.go#L29

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
hcgatewood: Oops sorry I meant `executeJob` we should test in isolation

But in any case -- I would disagree that it's too simple to test. I think testing-in-depth is a valuable pattern, and here in particular it would add confidence that executing a job is doing exactly what we think

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (r *reindexerSingleton) reindexJob(job *Job, ctx context.Context, batches []reindexBatch, sendUpdate func(string)) error {
+	defer TestHookReindexDone()
+
+	start := clock.Now()
+
+	jobErr := executeJob(ctx, job, batches)
+
+	if jobErr != nil {
+		return fmt.Errorf("error executing job %+v with job err <%s>", job, jobErr)
```
hcgatewood: We don't really want to early return here -- because then we are missing out on logging and metrics, which we want to propagate on both success and failure

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"magma/orc8r/cloud/go/services/orchestrator/obsidian/models"
+	"magma/orc8r/cloud/go/services/state/indexer"
+	"magma/orc8r/cloud/go/services/state/indexer/reindex"
+	state_test_init "magma/orc8r/cloud/go/services/state/test_init"
+	state_test "magma/orc8r/cloud/go/services/state/test_utils"
+)
+
+func TestSingletonRun(t *testing.T) {
+	// Make nullimpotent calls to handle code coverage indeterminacy
+	reindex.TestHookReindexSuccess()
```
hcgatewood, line 38: Let's do this then: let's leave the success hook in place for this PR, and refactor it out at the end of this milestone by using `sendUpdate` instead of the test hook

hcgatewood (changes requested):
Looking good, main things are 1. early return concern, and 2. let's add tests for executeJob

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
hcgatewood: Right so previously it made sense because you desired some loop timeout when popping jobs off the queue, and want to wait much longer if the job you popped off keeps failing

In this new situation, agreed it makes sense to just have 1 sleep interval

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
reginawang3495: Added a comment!

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
reginawang3495: Why is it better? I think it might not be especially it's not clear that it doesn't need to be initialized + I think you need to initialize it if you are not directly using certain methods on it. EG https://github.com/magma/magma/blob/master/orc8r/cloud/go/obsidian/swagger/spec/combine.go#L68 has to stay same (initialized, not declared)

reginawang3495: I did make a PR on it though! https://github.com/magma/magma/pull/10098

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,172 @@
...
+// TODO(reginawang3495): Remove "Desired" from Indexer.Versions once reindexer_queue and queue are removed
+
+// reindexerSingleton runs as though it is a singleton
+type reindexerSingleton struct {
+	Versioner
+	store Store
+}
+
+// TODO(reginawang3495): to be setup and used in M2 Part C
+const reindexLoopInterval = 5 * time.Second
```
hcgatewood: Let's make this 1m -- e.g. 1m between every reindex attempt, where we only envision having to actually retry anything if an indexer fails

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
hcgatewood: Hm so what's the issue with making the instance you link non-initialized?

Basically 1. we want everything to be uniform, for simplicity + clear communication to future programmers. And then we want to use declaration if possible, initialization if necessary. It's part of the Golang zen, but also there are performance indications at scale (not huge deal here) -- initialization uses memory, declaration doesn't, and with enough of those it can put additional strain on the garbage collector

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -0,0 +1,172 @@
...
+	batches := r.getReindexBatches(ctx)
+	for {
+		if isCanceled(ctx) {
+			glog.Warning("State reindexing async job canceled")
+			return
+		}
+
+		r.reindexJobs(ctx, indexerID, batches, sendUpdate)
+		clock.Sleep(reindexLoopInterval)
+	}
```
hcgatewood, line 60: Hm okay I think we need to reorient this just a bit. Here are the concerns I have as-is

1. We only read in the states once -- so e.g. it works fine at startup, but if an indexer's version gets bumped 1 day later, the reindexer would miss out on all the state that had been reported in that last day
2. Every loop, we read in all state (IDs). Not end of the world, but what we'd rather do is first check whether any indexers need to be reindexed -- if no, return early (and sleep); if yes, then read in state IDs and proceed

We also wanna add a regression test for this pattern

hcgatewood (approved):
Okay cool. This looks good! Couple comments, but let's go ahead and merge this PR (lgtm) and address them in a followup PR!

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
reginawang3495: Isn't the current test already a regression test? But yes I will try to add testing for the new changes in the new PR

PR #9945, opened 2021-10-28
fix(pipelined): Fix IPv6 enforcement rule match.

pshelar:
magma AGW provides IPv6 subnet to each UE. UE is
free to select any IP address in that subnet. Fix
enforcement rule accordingly.

## Test Plan
`make test`
bare metal test.

## Additional Information


lte/gateway/python/magma/pipelined/policy_converters.py
```diff
@@ -57,6 +58,10 @@ def _check_pkt_protocol(match):
     return True
 
 
+def flow_ue_ipv6_match_value(ip_addr) -> str:
```
koolzz: mbe rename to smth like -> get_ipv6_match_value

koolzz (approved):

PR #9953, opened 2021-10-28
feat(ci): Group all builds in on workflow

quentinDERORY:
Will permit to execute integ test upon completion

## Summary

Regroup all build job into one build workflow.
Will permit to trigger full integ test suite upon completion of all builds.

## Test Plan

Currently testing on forked repo

## Additional Information


talkhasib:
Have you considered using reusable call flows for this. make the reusable flows not trigger on any event other than the "workflow_call"
https://docs.github.com/en/actions/learn-github-actions/reusing-workflows

PR #9955, opened 2021-10-28
feat(agw): Capture Service definition and implementation.

mstre123:
feat(agw): Capture Service definition and implementation. 
Test runner for capturing protos.
Add replace rpc call for config service.
Fix bug in config service merge.

## Summary

Initial commit for sctpd capture.

Follow on items:
included one example golden file generated - follow on commit will contain all other golden files to avoid crowding this commit

## Test Plan

Unit testing and ran capture service test runner.

## Additional Information


src/go/agwd/config/config.go
```diff
@@ -87,6 +87,9 @@ func newDefaultConfig() *mconfig.AgwD {
 		MmeSctpdUpstreamServiceTarget:   "unix:///tmp/mme_sctpd_upstream.sock",
 		// Sentry is disabled if DSN is not set.
 		SentryDsn: "",
+		SctpdDownstreamMiddlewareConfig: &mconfig.MiddlewareConfig{
```
mstre123: this default will not be committed - was for visibility of how we can set it.

lte/protos/mconfig/mconfigs.proto
```diff
@@ -525,4 +525,10 @@ message AgwD {
   // The DSN to use. If the data source name is not set, the client is effectively
   // disabled.
   string sentry_dsn = 6;
+
+  MiddlewareConfig sctpd_downstream_middleware_config = 7;
```
markjen:
```
CaptureConfig capture_config = 7;

message CaptureConfig {
  message MatchSpec {
    string service = 1;
    string method = 2;
  }
  repeated MatchSpec match_specs = 1;
}
```

src/go/agwd/server/server.go
```diff
@@ -29,8 +30,10 @@ import (
 )
 
 func newServiceRouter(cfgr config.Configer) service.Router {
-	sctpdDownstreamConn, err := grpc.Dial(
-		cfgr.Config().GetSctpdDownstreamServiceTarget(), grpc.WithInsecure())
+	sctpdDownstreamClientMiddleware := middleware.NewMiddleware(cfgr.Config().GetSctpdDownstreamMiddlewareConfig())
+	dialOptions := sctpdDownstreamClientMiddleware.GetDialOptions()
+	dialOptions = append(dialOptions, grpc.WithInsecure())
+	sctpdDownstreamConn, err := grpc.Dial(cfgr.Config().GetSctpdDownstreamServiceTarget(), dialOptions...)
```
markjen: let's have the middleware always installed on all services and controlled by the config

src/go/middleware/middleware.go
```diff
@@ -0,0 +1,106 @@
...
+func (m *Middleware) GetDialOptions() (a []grpc.DialOption) {
+	var options = []grpc.DialOption{}
+	if m.writePath == "" {
+		return append(options, m.withPassthroughUnaryClientInterceptor())
+	}
+	return append(options, m.withCaptureUnaryClientInterceptor())
+}
+
+// withPassthroughUnaryClientInterceptor returns an interceptor that is a passthrough and takes no action on the intercepted rpc calls.
+func (m *Middleware) withPassthroughUnaryClientInterceptor() grpc.DialOption {
```
markjen: also need a UnaryServerInterceptor

src/go/middleware/middleware.go
```diff
@@ -0,0 +1,106 @@
...
+
+// middleware defines middlewares functions - used for generating mocks.
+type middleware interface {
+	GetDialOptions() []grpc.DialOption
+	withPassthroughUnaryClientInterceptor() grpc.DialOption
+	withCaptureUnaryClientInterceptor() grpc.DialOption
+}
+
+type Middleware struct {
+	writePath string
```
markjen: let's hold a pointer to the config mgr so we can consult it on every call

markjen (commented):
let's see about building a "config grpc service" that runs in agwd/main.go and allows for the caller to update config proto

and also a "capture grpc service" that dump/flushes the captured buffer

src/go/agwd/config/config.go
```diff
@@ -55,23 +58,48 @@ const (
...
 		return resolver.Target{
-			Scheme:   ipv4Scheme,
+			Scheme:   tcp4Scheme,
 			Endpoint: target[len(ipv4Scheme)+1:],
 		}
 	}
 	if strings.HasPrefix(target, ipv6Scheme+":") {
 		return resolver.Target{
-			Scheme:   ipv6Scheme,
+			Scheme:   tcp6Scheme,
```
mstre123: @zbrock we can ignore this right now ill rebase on #10364 before pushing

src/go/agwd/config/config_test.go
```diff
@@ -96,27 +172,43 @@ func TestParseTarget(t *testing.T) {
...
 			want: resolver.Target{
-				Scheme:    "ipv4",
+				Scheme:    "tcp4",
 				Authority: "",
 				Endpoint:  "localhost:1234",
 			},
 		},
 		{
 			target: "ipv6:[2607:f8b0:400e:c00::ef]:443",
 			want: resolver.Target{
```
mstre123: Same re above

src/go/agwd/config/config.go
```diff
@@ -82,15 +110,19 @@ type ConfigManager struct {
...
 		SctpdUpstreamServiceTarget:      "unix:///tmp/sctpd_upstream.sock",
 		MmeSctpdDownstreamServiceTarget: "unix:///tmp/mme_sctpd_downstream.sock",
 		MmeSctpdUpstreamServiceTarget:   "unix:///tmp/mme_sctpd_upstream.sock",
 		// Sentry is disabled if DSN is not set.
-		SentryDsn:           "",
-		ConfigServiceTarget: "127.0.0.1:50090",
+		SentryDsn:               "",
+		VagrantPrivateNetworkIp: "192.168.60.142",
+		ConfigServicePort:       "6000",
+		CaptureServicePort:      "6001",
```
markjen, line 114: in these cases do we want `net.Listen("tcp", ":" + port)`? perhaps we don't bother going through a `resolver.Target`

src/go/agwd/config/config.go
```diff
@@ -156,7 +188,15 @@ func (c *ConfigManager) Merge(update *config.AgwD) {
...
-	c.config = config
+	*c.config = *config
+}
+
+// Replace overwrites the managed config.
+func (c *ConfigManager) Replace(replace *config.AgwD) {
+	c.Lock()
+	defer c.Unlock()
+
+	*c.config = *replace
```
markjen: you probably want to use https://pkg.go.dev/github.com/golang/protobuf/proto#Clone, which creates a deep copy.

(`*x = *y` does a shallow copy)

src/go/capture/buffer/buffer.go
```diff
@@ -0,0 +1,55 @@
...
+func NewBuffer() *Buffer {
+	return &Buffer{}
+}
+
+// Write locks the buffer while writing a UnaryCall to the buffer.
+func (b *Buffer) Write(call capture.UnaryCall) error {
+	b.Lock()
+	defer b.Unlock()
+	b.unaryCalls = append(b.unaryCalls, &call)
+	return nil
```
markjen: if there's no way to error, then change the func signature to not return one :)

src/go/capture/buffer/buffer.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Flush locks the buffer, returns all of its contents and replaces the internal call list with nil.
+func (b *Buffer) Flush() ([]*capture.UnaryCall, error) {
+	b.RLock()
+	defer b.RUnlock()
+	calls := b.unaryCalls
+	b.unaryCalls = nil
+	return calls, nil
```
markjen: same here, no err return necessary if it's not possible

src/go/middleware/middleware.go
```diff
@@ -0,0 +1,145 @@
...
+
+// middleware defines middlewares functions - used for generating mocks.
+type middleware interface {
+	GetDialOptions() []grpc.DialOption
+	GetServerOptions() []grpc.ServerOption
+	withCaptureUnaryClientInterceptor() grpc.DialOption
+	withCaptureUnaryServerInterceptor() grpc.ServerOption
+}
+
+type Middleware struct {
```
markjen: did you mean to define the interface publicly and leave the impl private?

src/go/middleware/middleware.go
```diff
@@ -0,0 +1,145 @@
...
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package middleware
```
markjen: nit: maybe this can go in the `capture` package

markjen (changes requested):

src/go/agwd/config/config.go
mstre123: probably cleaner - since i don't see us accessing these services through sockets, ill repurpose GetLocalHostTarget

src/go/middleware/middleware.go
mstre123: moved

src/go/agwd/config/config.go
mstre123: cool just added

src/go/middleware/middleware.go
mstre123: flipped it

src/go/capture/buffer/buffer.go
```diff
@@ -0,0 +1,54 @@
...
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package buffer
```
markjen: nit: i would probably put this in the `capture` package so it's `capture.Buffer`, but this is fine also.

generally, i'm not a fan of "micropackages", just like I'm not a fan of microservices :)

src/go/capture/buffer/buffer.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"sync"
+
+	"github.com/magma/magma/src/go/protos/magma/capture"
+)
+
+//go:generate go run github.com/golang/mock/mockgen -source=buffer.go -package mock_buffer -destination mock_buffer/mock_buffer.go Buffer
+
+// buffer interface defines functions of Buffer and is used for generating mocks.
+type buffer interface {
```
markjen: this feels reversed, `Buffer` should be the interface and `buffer` the implementation

for a simple struct like this, it's OK to not have an interface (testers can instantiate the concrete impl if necessary; it's entirely in memory and doesn't have any dependencies)

src/go/agwd/server/server.go
```diff
@@ -107,7 +110,8 @@ func startSctpdDownlinkServer(
 			target.Endpoint))
 	}
 
-	grpcServer := grpc.NewServer()
+	mw := middleware.NewMiddleware(cfgr.Config(), buf)
```
markjen: could the middleware be passed in?

i.e. this func could take a `opt ...grpc.ServerOption` param and pass it along to `grpc.NewServer`

src/go/agwd/server/server.go
```diff
@@ -177,21 +182,37 @@ func startPipelinedServer(cfgr config.Configer, logger log.Logger) {
...
 	cfgr config.Configer, logger log.Logger,
 ) {
-	target := config.ParseTarget(cfgr.Config().GetConfigServiceTarget())
-	if target.Scheme == "unix" {
-		cleanupUnixSocketOrDie(logger, target.Endpoint)
-	}
+	address := fmt.Sprintf(":%s", cfgr.Config().GetConfigServicePort())
 
-	listener, err := net.Listen(target.Scheme, target.Endpoint)
+	listener, err := net.Listen(config.Tcp4Scheme, address)
```
markjen: do we want `tcp4` or just `tcp`

markjen (approved):

src/go/agwd/server/server.go
mstre123: yeah thatd actually be safer to make sure ppl use the same buffer

src/go/capture/buffer/buffer.go
mstre123: i can drop the interface but Buffer needs to be exported for use in the capture server

PR #9964, opened 2021-10-28
feat(orc8r): create new shared mconfig with Sentry info

### LINKED ISSUE #9907: [Orc8r] Create new shared mconfig with Sentry info
### Summary 
* Create a new “shared” service [mconfig in Orc8r ](https://github.com/magma/magma/blob/8aa32f8195adb2325a960b307eb0e8bd3d26e647/orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go#L90-L103)
* Should contain the [SentryConfig](https://github.com/magma/magma/blob/ec568fc811feaecd7586333d6feb04e9f70358a7/lte/cloud/go/protos/mconfig/mconfigs.pb.go#L304-L325) and populate it from the [network-wide config](https://github.com/magma/magma/blob/ec568fc811feaecd7586333d6feb04e9f70358a7/lte/cloud/go/services/lte/servicers/builder_servicer.go#L637-L652)
* This will, eventually, be consumed by the AGW. 

### Rationale 
* Right now, each service has its own config, but there are instances where we want to share config info with the services. 
* (e.g.) SentryConfig, which contains the crash reporting should be shared amongst all services

christinewang5:
## Summary
* Created a `shared_mconfig` that contains configs that is shared among services. Currently it only contains `SentryConfig` 

## Test Plan
* Created unit test 
test_unit

* Used swagger api `/networks/{network_id}/sentry` to create sentry config and check that it is in [gateway config](https://docs.magmacore.org/docs/next/lte/dev_notes#:~:text=/var/opt/magma/configs/gateway.mconfig) 
test_e2e


lte/protos/mconfig/mconfigs.proto
```diff
@@ -14,30 +14,11 @@ limitations under the License.
...
-    string url_native = 2;
-    // upload_mme_log decides whether MME service log file (/var/log/mme.log)
-    // is uploaded along with MME crashreports
-    bool upload_mme_log = 3;
-    // sample_rate sets the rate at which Python error events are sampled.
-    // sample_rate should be a number between 0 (0% of errors sent) and 1 (100%
-    // of errors sent)
-    float sample_rate = 4;
-}
-
```
christinewang5: Not sure if I can remove the SentryConfig here to avoid having duplicates, but all the unit tests still pass...

christinewang5: Actually it seems to break the mme_test CI check because can't import this...
image

lte/cloud/go/services/lte/servicers/builder_servicer.go
```diff
@@ -634,23 +633,6 @@ func getNetworkProbeConfig(ctx context.Context, networkID string) ([]*lte_mconfi
 	return npTasks, liUes
 }
 
-func getNetworkSentryConfig(network *configurator.Network) *lte_mconfig.SentryConfig {
```
hcgatewood: Eh tbh I would just leave this duplicated (since the return types are technically different), and add a note in this one that it's a dupe of the orc8r one

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -101,6 +101,7 @@ func (b *baseOrchestratorBuilder) Build(network *storage.Network, graph *storage
 	vals["control_proxy"] = &mconfig_protos.ControlProxy{LogLevel: protos.LogLevel_INFO}
 	vals["metricsd"] = &mconfig_protos.MetricsD{LogLevel: protos.LogLevel_INFO}
 	vals["state"] = getStateMconfig(net, gatewayID)
+	vals["shared_mconfig"] = &mconfig_protos.SharedMConfig{SentryConfig: getNetworkSentryConfig(&net)}
```
hcgatewood: Nit: seems like the prevalent styling is "Mconfig" rather than "MConfig". I'm a fan of the latter, but consistency is preferable for small stuff like this

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
```diff
@@ -39,6 +39,46 @@ func TestBaseOrchestratorMconfigBuilder_Build(t *testing.T) {
...
+			"shared_mconfig": &mconfig_protos.SharedMConfig{
+				SentryConfig: &mconfig_protos.SharedSentryConfig{
+					SampleRate:   0.75,
+					UploadMmeLog: true,
+					UrlPython:    "https://www.example.com/v1/api",
+					UrlNative:    "https://www.example.com/v1/api",
+				},
+			},
+		}
+		actual, err := buildSharedMConfig(&nw, &graph, "gw1")
```
hcgatewood: I'd say let's just forego adding the `buildSharedMConfig` function, and instead use the buildBaseOrchestrator func. You'll have to add in some non-shared_mconfig values into the expected, but that's fine + even improves the test coverage a bit

orc8r/protos/mconfig/mconfigs.proto
```diff
@@ -18,6 +18,38 @@ import "orc8r/protos/common.proto";
...
 
+
+//===========================================================================
+// SharedMConfig is a wrapper around configs that are shared amongst services
+//===========================================================================
+// --------------------------------------------------------------------------
+// SentryConfig stores the network-wide Sentry.io configuration
+// TODO(hcgatewood): rename, duplicated in lte/.../mconfigs
+// --------------------------------------------------------------------------
+message SharedSentryConfig {
```
hcgatewood, line 33: (Same nit: Mconfig rather than MConfig)

orc8r/protos/mconfig/mconfigs.proto
```diff
@@ -18,6 +18,38 @@ import "orc8r/protos/common.proto";
...
+  // sample_rate sets the rate at which Python error events are sampled.
+  // sample_rate should be a number between 0 (0% of errors sent) and 1 (100%
+  // of errors sent)
+  float sample_rate = 4;
+}
+
+// --------------------------------------------------------------------------
+// SharedMConfig is a wrapper around configs that are shared amongst services
+// --------------------------------------------------------------------------
+message SharedMConfig {
```
hcgatewood: Nit: structure files top-down -- i.e. as you jump-to-definition, you should be moving progressively down the file (increasing line number)

So that to say, place sharedmconfig above sharedsentryconfig

hcgatewood (changes requested):
Looks great! Couple of nits, main thing is let's restructure the unit test just a bit. Other than that will be ready to merge 👍

lte/protos/mconfig/mconfigs.proto
```diff
@@ -20,6 +20,7 @@ option go_package = "magma/lte/cloud/go/protos/mconfig";
 
 // --------------------------------------------------------------------------
 // SentryConfig stores the network-wide Sentry.io configuration
+// TODO(hcgatewood): remove this in next release, duplicated in the shared_mconfig
```
hcgatewood:
```suggestion
// TODO(hcgatewood): remove this after 1.7 release, duplicated in the shared_mconfig
```

hcgatewood (approved):
Lgtm! Awesome. One nit, msg me when ready and I'll merge 👍

hcgatewood:
Oh also need a review from @themarwhal when she has time 🙂

themarwhal (approved):
awesome! thanks for doing this.
QQ should this file be committed? `orc8r/gateway/python/core`

christinewang5:
@themarwhal I removed `orc8r/gateway/python/core`, don't think it was there before.

christinewang5:
renamed url -> dsn to match sentry's page, discussed w @themarwhal offline

PR #9523, opened 2021-10-07
feat(agw): [S1_Handover] Added test cases for S1 Handover scenarios

VinashakAnkitAman:
## Title
[S1_Handover] Added test cases for S1 Handover scenarios

## Summary
This PR adds some new test cases to verify handover scenarios supported by Magma. The S1 Handover feature developed in Magma as per the document: https://github.com/magma/magma/blob/master/docs/readmes/proposals/p011_intra_agw_mobility.md.

Signed-off-by: VinashakAnkitAman <ankit.aman@radisys.com>

PR #9966, opened 2021-10-28
fix(agw): Fixed memleak for s1 handover cancel 

ardzoht:
## Summary

- Backporting changes of #9938 into v1.6 branch

## Test Plan

- ran `make integ_test`
- used s1 handover individual cancel test case from https://github.com/magma/magma/pull/9523

## Additional Information


ssanadhya (approved):

ulaskozat (approved):
LGTM

PR #9967, opened 2021-10-29
fix(agw): Compensate for eNB firmware param inversion

amarpad:
## Summary
A regression was introduced into the firmware build for Sercomm where
the flag for cell reserved was inverted. Compensate for that by
setting the cell reserved bit to be True on initial connect.
At a later point when we have a software version where the issue is fixed
we can add this as a conditional check.

## Test Plan

UE traffic was not working before fix, working after fix.
Unit test

## Additional Information


lte/gateway/python/magma/enodebd/tests/freedomfi_one_tests.py
```diff
@@ -804,3 +814,5 @@ def test_status_nodes(self):
         ]
         status.set_magma_device_cfg(n8, device_config)
         self.assertEqual(expected, device_config.mock_calls)
+
```
ardzoht: remove?

ardzoht (approved):
lgtm, left minor comment

PR #9968, opened 2021-10-29
fix(mme): Handle failed teid update by deactivating default bearer

ulaskozat:
## Summary

When mme service updates sessiond with TEID, if sessiond cannot find an existing session, it returns abort response. MME was ignoring such cases. The PR treats such aborts the same as deletion of the default bearer (i.e., session termination) and mme triggers a network initiated detach procedure.

## Test Plan

Modified s1ap test and hardcoded sessiond to return error. 
Tested on spirent where such events are observed when sessiond and mme diverges about the existence of a session.

Unit tests to be added once spgw test suit becomes ready.
## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/lib/pcef/pcef_handlers.cpp
+++ b/lte/gateway/c/core/oai/lib/pcef/pcef_handlers.cpp
@@ -212,7 +212,8 @@ void pcef_update_teids(
               LOG_UTIL,
               "Received grpc::ABORTED for update_teids RPC for %s with error "
               "msg: %s. Deactivating bearer %u.",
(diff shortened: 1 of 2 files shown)


ardzoht (approved):
lgtm

themarwhal:
Screen Shot 2021-10-29 at 2 03 52 PM
^ what are these binaries?

themarwhal (approved):

ulaskozat:
> Screen Shot 2021-10-29 at 2 03 52 PM
> 
> ^ what are these binaries?

Good question! They have no place here :)

PR #9974, opened 2021-10-29
fix(agw): Enable opt-in based logging to Sentry

sebathomas:
## Summary

Makes it possible to select only a subset of log messages to be sent to Sentry instead of everything above a certain log level. In Python services there are now three possible configuration settings for Sentry:
* send all errors
* send only selected errors (opt-in)
* disabled

Also adds a configurable decorator to send all uncaught errors to Sentry, which is useful in the opt-in case. It is used for all gRPC endpoints of mobilityd.

### DIFF
--- a/orc8r/gateway/python/magma/common/tests/sentry_tests.py
+++ b/orc8r/gateway/python/magma/common/tests/sentry_tests.py
@@ -1,0 +1,52 @@
+"""
+Copyright 2021 The Magma Authors.
+
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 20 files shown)


orc8r/gateway/python/magma/common/tests/sentry_tests.py
```diff
@@ -0,0 +1,52 @@
+"""
+Copyright 2020 The Magma Authors.
```
themarwhal: 2021 :)

themarwhal:
Do you have any pictures to indicate what these exceptions might look like on Sentry?

orc8r/gateway/python/magma/common/tests/sentry_tests.py
sebathomas: Right, almost 2022 now. :)

sebathomas:
> Do you have any pictures to indicate what these exceptions might look like on Sentry?

From manual tests:

sentry-error

themarwhal (approved):

pshelar (approved):
LGTM

PR #9978, opened 2021-10-29
fix(amf): Changing the default parameter for amfName in gate

RahulKalsangra:
Signed-off-by: RahulKalsangra rahul.kalsangra@wavelabs.ai

## Summary
Default name of parameter name AMF_NAME is falling in sit test in mconfig file
1. Change the variable name default value with acceptable with all gnb.

## Test Plan
 1. UERANSIM
 
Registration and PDU session establishment successful with UERANSIM.
AMF

Unit Testing Successful:

TEST_CASE


ulaskozat (approved):
LGTM

PR #9982, opened 2021-10-29
fix(agw): Cherry-pick: Compensate for eNB firmware param inversion (#9967)

amarpad:
## Summary
fix(agw): Compensate for eNB firmware param inversion

A regression was introduced into the firmware build for Sercomm where
the flag for cell reserved was inverted. Compensate for that by
setting the cell reserved bit to be True on initial connect.

At a later point when we have a software version where the issue is fixed
we can add this as a conditional check.

fix(agw): Update tests for cell reserved

Update expectations to account for calls to Cell Reserved
Also fixes an issue where in place sort was wrongly used in assertEqual

Co-authored-by: Amar Padmanabhan <amar@freedomfi.com>

## Test Plan
make test

## Additional Information


ardzoht (approved):
lgtm

ulaskozat (approved):
LGTM

PR #9983, opened 2021-10-29
fix(cloud): Adding staging domain SANs to self-signed certificates

arunuke:

fix(cloud): Adding staging domain SANs to self-signed certificates

## Summary

Enhancing the self-signed certificate script to add additional domains to support Orc8r Blue-Green upgrades. 

## Test Plan

After creating certificates, verify if they include the configured domain and staging domains as well.

openssl x509 -in controller.crt -noout -text | grep DNS

 DNS:*.bluegreen.failedwizard.dev, DNS:*.nms.bluegreen.failedwizard.dev, DNS:*.staging.bluegreen.failedwizard.dev, DNS:*.staging.nms.bluegreen.failedwizard.dev

## Additional Information

### DIFF
--- a/orc8r/cloud/deploy/scripts/self_sign_certs.sh
+++ b/orc8r/cloud/deploy/scripts/self_sign_certs.sh
@@ -55,6 +55,8 @@ subjectAltName = @alt_names
 [alt_names]
 DNS.1 = *.$domain
 DNS.2 = *.nms.$domain
+DNS.3 = *.staging.$domain
+DNS.4 = *.nms.staging.$domain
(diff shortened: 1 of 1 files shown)


hcgatewood (approved):
Nit: I'd say let's use `canary` rather than `staging`, since it feels a bit more like "this is different be careful"

Also, we'll want to document the SANs required when operators procure their own certs out of band -- I'll throw up a PR later this week with the current list

Edit: PR documenting SAN requirements: https://github.com/magma/magma/pull/10054

PR #9985, opened 2021-10-29
fix(agw): Agw containerization fixes2

rmeleromira:
Signed-off-by: Ramon Melero <ramonmelero@fb.com>

## Summary

Adds health checks to most services.

Fixes an issue where the "only_upgrade" option causes the ifupdown package to not install. Moved it into it's own task so that it's in the latest state and always installs.

## Test Plan

Tested on a brand new aws cloudstraper ubuntu 20 instance

https://gist.github.com/rmeleromira/6699a8ee3cd1338dc788102455a7cc5d

## Additional Information


lte/gateway/docker/docker-compose.yaml
```diff
@@ -94,45 +110,78 @@ services:
...
     command: /usr/bin/env python3 -m magma.ctraced.main
 
   sctpd:
     <<: *ltecservice
     container_name: sctpd
+    # Figure out sctp test
+    # healthcheck:
+    #   test: ["CMD", "nc", "-zv", "localhost", "50067"]
+    #   timeout: "4s"
+    #   retries: 3
```
ymasmoudi, line 185: why is the sctpd health check not enabled ?

rmeleromira: sctpd afaik doesn't listen on any tcp/udp port we can check. I couldn't figure out a reliable way to test for port 36412 or some other health check for sctpd.

```
root@ip-10-7-4-43:/var/opt/magma/tmp# nc -vz localhost 36412
nc: connect to localhost port 36412 (tcp) failed: Connection refused
root@ip-10-7-4-43:/var/opt/magma/tmp# nc -vz 10.7.2.67 36412
nc: connect to 10.7.2.67 port 36412 (tcp) failed: Connection refused
root@ip-10-7-4-43:/var/opt/magma/tmp# ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
root           1       0  0 Oct14 ?        00:06:47 /usr/local/bin/sctpd
root         199       0  0 19:08 pts/0    00:00:00 bash
root         216     199  0 19:09 pts/0    00:00:00 ps -ef
root@ip-10-7-4-43:/var/opt/magma/tmp# lsof -p 1 | grep LISTEN
root@ip-10-7-4-43:/var/opt/magma/tmp#
```

I thought of checking for the sctp active endpoint, but that's just OS level check, not application level.
```
root@ip-10-7-4-43:/var/opt/magma/tmp# cat /proc/net/sctp/eps
 ENDPT     SOCK   STY SST HBKT LPORT   UID INODE LADDRS
ffff98e4a0dc7c00 ffff98e488408600 2   10  9    36412     0 1779729 10.7.2.67
```

Do you have any tricks for checking sctpd health?

ymasmoudi: @rmeleromira Yes, we are using unix socket for sctpd.
You can try the following,

root@agw:/var/opt/magma/docker# ncat -vz -u --sctp -4 -U /tmp/sctpd_upstream.sock 
Ncat: Version 7.80 ( https://nmap.org/ncat )
Ncat: Connected to /tmp/sctpd_upstream.sock.
Ncat: 0 bytes sent, 0 bytes received in 0.01 seconds.

root@agw:/var/opt/magma/docker# ncat -vz -u --sctp -4 -U /tmp/sctpd_downstream.sock 
Ncat: Version 7.80 ( https://nmap.org/ncat )
Ncat: Connected to /tmp/sctpd_downstream.sock.
Ncat: 0 bytes sent, 0 bytes received in 0.01 seconds.

lte/gateway/docker/docker-compose.yaml
```diff
@@ -94,45 +110,78 @@ services:
...
+      retries: 3
     command: >
       sh -c "/usr/local/bin/generate_nghttpx_config.py && sed -i 's/errorlog-syslog=yes/errorlog-syslog=no/' /var/opt/magma/tmp/nghttpx.conf &&\
              /usr/bin/env nghttpx --conf /var/opt/magma/tmp/nghttpx.conf /var/opt/magma/certs/controller.key /var/opt/magma/certs/controller.crt"
 
   ctraced:
     <<: *pyservice
     container_name: ctraced
+    healthcheck:
+      test: ["CMD", "nc", "-zv", "localhost", "50079"]
```
themarwhal, line 173: Is there a way to get this port value from the service registry config? if not, maybe we should add a note to change the value when one config is modified?

lte/gateway/docker/docker-compose.yaml
arunuke: A future task for the k8s effort in Q1 would be to eliminate some of these dependencies like Unix sockets. Given we will move away from unix sockets soon, this should be something that we can mark as a known issue.

rmeleromira: > 

Reported issue https://github.com/magma/magma/issues/10573 , I'm not familiar with the effort to remove unix sockets so issue is kind of anemic.

lte/gateway/docker/docker-compose.yaml
arunuke: Created a [new issue](https://github.com/magma/magma/issues/10578) to track this. Current thinking is to create a docker-compose template file that will be filled out runtime based on previously set values.

arunuke (commented):
LGTM. If comments from @ymasmoudi and @themarwhal are satisfactorily addressed, we should be good to merge. I will follow up on both tomorrow.

themarwhal (approved):

quentinDERORY (approved):

PR #9986, opened 2021-10-30
fix(pipelined): Set tunnel out id in reg

pshelar:
This value can used to map ingress tunnel id to egress.

## Test Plan
`make test`

## Additional Information


lte/gateway/c/core/oai/lib/openflow/controller/GTPApplication.cpp
```diff
@@ -171,7 +171,7 @@ void GTPApplication::add_uplink_tunnel_flow(
   apply_ul_inst.add_action(set_eth_dst);
 
   of13::SetFieldAction set_tunnel_id(
-      new of13::NXMRegX(TUNNEL_ID_REG, ev.get_in_tei()));
+      new of13::NXMRegX(TUNNEL_ID_REG, ev.get_out_tei()));
```
koolzz, line 174: Mm do we use this anywhere rn?

koolzz (approved):

lte/gateway/c/core/oai/lib/openflow/controller/GTPApplication.cpp
pshelar: nope, plan is to use it in ipv6 solicitation handler app.

ssanadhya (approved):

PR #9989, opened 2021-10-30
feat(orc8r): Support registering dev VM with remote controller

hcgatewood:
## Summary

Update fabfile to support registering local dev AGW VM with remote Orc8r.

## Test Plan

Manual

## Additional Information

### DIFF
--- a/lte/gateway/dev_tools.py
+++ b/lte/gateway/dev_tools.py
@@ -10,10 +10,10 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
-
+import os
 import subprocess
 import sys
-from typing import Any, List
+from typing import Any, List, Optional
 
 import urllib3
 
@@ -65,6 +65,34 @@ def register_vm():
     )
     _register_network(LTE_NETWORK_TYPE, network_payload)
     _register_agw(LTE_NETWORK_TYPE)
+
+
+def register_vm_remote(certs_dir: str, network_id: str, url: str):
+    """
+    Register local VM gateway with remote controller.
+
+    Example usage:
+    fab -f dev_tools.py register_vm_remote:certs_dir=~/certs,network_id=test,url=https://api.stable.magmaeng.org
+    """
+    if None in {certs_dir, url, network_id}:
+        print()
+        print('==============================================================')
+        print('Must provide the following arguments: certs_dir,url,network_id')
+        print('==============================================================')
+        return
+
+    admin_cert = types.ClientCert(
+        cert=os.path.expanduser(f'{certs_dir}/admin_operator.pem'),
+        key=os.path.expanduser(f'{certs_dir}/admin_operator.key.pem'),
+    )
+    full_url = url + '/magma/v1/'
+
+    _register_agw(
+        LTE_NETWORK_TYPE,
+        url=full_url,
+        admin_cert=admin_cert,
+        network_id=network_id,
+    )
 
 
 def register_federated_vm():
@@ -145,11 +173,27 @@ def _register_network(network_type: str, payload: Any):
     dev_utils.create_tier_if_not_exists(network_id, 'default')
 
 
-def _register_agw(network_type: str):
-    network_id = NIDS_BY_TYPE[network_type]
+def _register_agw(
+        network_type: str,
+        url: Optional[str] = None,
+        admin_cert: Optional[types.ClientCert] = None,
+        network_id: Optional[str] = None,
+):
+    network_id = network_id or NIDS_BY_TYPE[network_type]
+
+    dev_utils.create_tier_if_not_exists(
+        network_id,
+        'default',
+        url=url,
+        admin_cert=admin_cert,
+    )
+
     hw_id = dev_utils.get_gateway_hardware_id_from_vagrant(vm_name='magma')
     already_registered, registered_as = dev_utils.is_hw_id_registered(
-        network_id, hw_id,
+        network_id,
+        hw_id,
+        url=url,
+        admin_cert=admin_cert,
     )
     if already_registered:
         print()
@@ -158,7 +202,11 @@ def _register_agw(network_type: str):
         print(f'===========================================')
         return
 
-    gw_id = dev_utils.get_next_available_gateway_id(network_id)
+    gw_id = dev_utils.get_next_available_gateway_id(
+        network_id,
+        url=url,
+        admin_cert=admin_cert,
+    )
     md_gw = dev_utils.construct_magmad_gateway_payload(gw_id, hw_id)
     gw_payload = LTEGateway(
         device=md_gw.device,
@@ -173,7 +221,12 @@ def _register_agw(network_type: str):
         ),
         connected_enodeb_serials=[],
     )
-    dev_utils.cloud_post(f'lte/{network_id}/gateways', gw_payload)
+    dev_utils.cloud_post(
+        f'lte/{network_id}/gateways',
+        gw_payload,
+        url=url,
+        admin_cert=admin_cert,
+    )
     print()
     print(f'=========================================')
     print(f'Gateway {gw_id} successfully provisioned!')
--- a/orc8r/tools/fab/dev_utils.py
+++ b/orc8r/tools/fab/dev_utils.py
@@ -14,7 +14,7 @@ limitations under the License.
 import json
 import os
 import subprocess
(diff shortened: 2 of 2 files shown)


lte/gateway/dev_tools.py
```diff
@@ -67,6 +68,42 @@ def register_vm():
...
+    Example usage:
+    fab --set certs_dir=~/certs,network_id=test,url=https://api.stable.magmaeng.org \
+        -f dev_tools.py register_vm_remote
+    """
+    certs_dir = env.get('certs_dir')
+    url = env.get('url')
+    network_id = env.get('network_id')
+    if None in [certs_dir, url, network_id]:
+        print()
+        print(f'============================================================')
```
uri200: remove all the `f` from any string that doesn't have any variable placeholder

lte/gateway/dev_tools.py
```diff
@@ -67,6 +68,42 @@ def register_vm():
...
 
+def register_vm_remote():
+    """
+    Register local VM gateway with remote controller.
+
+    Example usage:
+    fab --set certs_dir=~/certs,network_id=test,url=https://api.stable.magmaeng.org \
+        -f dev_tools.py register_vm_remote
+    """
+    certs_dir = env.get('certs_dir')
```
uri200: why are you using env_vars instead of using arguments?
```
fab -f dev_tools.py register_vm_remote:url=api.stable.magmaeng.org
```

orc8r/tools/fab/dev_utils.py
```diff
@@ -102,25 +102,31 @@ def construct_magmad_gateway_payload(
 
 def get_next_available_gateway_id(
         network_id: str,
-        admin_cert: types.ClientCert = types.ClientCert(
-            cert='./../../.cache/test_certs/admin_operator.pem',
-            key='./../../.cache/test_certs/admin_operator.key.pem',
-        ),
+        url: str = PORTAL_URL,
+        admin_cert: Optional[types.ClientCert] = None,
```
uri200, line 106: Even that having defaults defined is not the most pythonic way, since this is a library is used by fab we better keep those values there as defaults.

Note that with your change `NONE` does not mean NONE or empty anymore. It means, in fact that we are using a `default` value somewhere in the functio. That forces you to know the implementation of the function. So, since they are default values, they belong to the default definition at the signature of the function.

Note in the past we had to do weird modifications due to using weird values as defaults that caused bugs (like [here](https://github.com/magma/magma/pull/9042/files#diff-5eae7e5883026c0cd6faf3adee5ef8848df8b0fd156b489d3808129490cc619cR241))

lte/gateway/dev_tools.py
```diff
@@ -142,14 +179,28 @@ def _register_network(network_type: str, payload: Any):
...
 
+def _register_agw(
+        network_type: str,
+        url: str,
+        admin_cert: Optional[types.ClientCert],
+        network_id: Optional[str],
+):
+    network_id = network_id or NIDS_BY_TYPE[network_type]
+
+    dev_utils.create_tier_if_not_exists(
```
uri200, line 184: Per my understanding tier should belong to the network, not to the gateway (one tier per network, not one tier per gateway). Otherwise we will create a tier every time we register an agw.

Also note this function is used by non-federated and federated, so unless you have tested this with the other deployments (no federated and federateD) and can say you can switch back and forth them without a problem, we should leave `create_tier_if_not_exists` inside `_register_network`.

uri200 (changes requested):

lte/gateway/dev_tools.py
```diff
@@ -67,6 +68,42 @@ def register_vm():
...
+        print()
+        print(f'============================================================')
+        print(f'Must --set the following variables: certs_dir,url,network_id')
+        print(f'============================================================')
+        return
+
+    admin_cert = types.ClientCert(
+        cert=os.path.expanduser(f'{certs_dir}/admin_operator.pem'),
+        key=os.path.expanduser(f'{certs_dir}/admin_operator.key.pem'),
+    ),
```
uri200: you are creating the tuple adding that `,`

lte/gateway/dev_tools.py
hcgatewood: Ah awesome, I couldn't figure out the right way to do this 🙂

lte/gateway/dev_tools.py
hcgatewood: I literally spent 30 minutes trying to debug that LOL. Thanks for the catch!

lte/gateway/dev_tools.py
hcgatewood: So this function explicitly creates an AGW (`lte/:network_id:/gateway`), independent of the network type

That said, the AGW could be in either network type, but it's the same handler. And the AGW (and FeGs, for that matter) require an underlying tier to exist https://github.com/magma/magma/blob/abab96b287980f7874bc03434cc2fa9795363d7f/orc8r/cloud/go/services/orchestrator/obsidian/handlers/gateway_handlers.go#L175-L182

So in this case, it should be safe to do it this way -- because if the tier didn't exist, registering the gateway would fail

orc8r/tools/fab/dev_utils.py
hcgatewood: The reason for providing implicit defaults is this way we can pass `None` in as an arg and have it be overridden -- makes it easier for intermediate functions to blindly pass in the admin_cert they received, rather than checking whether it's None

The rc of the linked issue is: the default value is created once, rather than on each function call. So if the default value is mutable, it will exhibit unexpected behavior https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

lte/gateway/dev_tools.py
uri200: Taking into consideration Tier is a network property (1 tier N Gateways), and  that a gateway shouldn't exist without a network, those changes seems to allow that.

In other words, trying to create a gateway without a network or a tier should raise an error, and this script is hiding this. 

Eventually you need to create a tier, either you do when you create the network or you create the gateway. Since this is a one time thing, this should be part of the network creation. 

But to me the most important reason is,  his is a script that doesn't have any way to test its reliability. Is this change really achieving something worth that risk?  Do you have the time to test the different permutations of adding and removing gateways you will find at fabscripts at `~/magma/lte/gateway/python/integ_tests/federated_tests` and `~/magma/lte/gateway/'

orc8r/tools/fab/dev_utils.py
uri200: You can just use `get_next_available_gateway_id("network_iid")` and it will take the defaults stated on the signature. 

If this were a magma service then that would be different. But this is a fab command/library.  The idea is to be clear on the signatures so when we use it we do not have to know the insights of it. Hiding defaults in the code defeats that purpose. Extensiveness is a trade off for ease and clarity on fab.

On the example  [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Ffabfile.py&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=243&start_col=17&end_row=243&end_col=17) shows how we ended up with a NONE becoming [1234]. That caused us lots of issues trying to find out why attaches were rejected on our local federated setup.  Should they have used the default on the signature, we would have seen that it immediately when building  the configuration [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Ffabfile.py&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=390&start_col=16&end_row=390&end_col=16)  (well in that case i would not have used any default at all :D )

uri200 (approved):
please if you plan to land this try to test adding and removing gateways with fab script at
```
~/magma/lte/gateway/python/integ_tests/federated_tests
~/magma/lte/gateway/

```

hcgatewood:
That feel when the UI moves "close pull request" under the cursor right before you click lol

hcgatewood:
@uri200 I put the "ensure tier exists" back into the network creation func, to preserve the existing functionality. Also tested `register_vm`, `register_federated_vm`, and `register_vm_remote` manually

I have reasonable confidence it won't break anything at this point, but definitely let me know if you hit any issues when the change lands and I'm happy to help debug 🙂

uri200:
Thanks for testing it!

PR #9993, opened 2021-10-31
feat: add settings to run precommit.py on save

themarwhal:
## Summary
Instead of configuring the python extension to run autopep8, configure a setting to just the python precommit.py script we have on save. This is better as we're guaranteed to get the same set of tools and its versions as CI.

TODO: this is not very useful unless the editor reloads the file from disk.

## Additional Information


themarwhal:
closing for now - would be nice to get this in the future though

PR #9999, opened 2021-11-01
feat(ci): Cancel old jobs when new commits lands on PR

quentinDERORY:

## Summary

Cancel old jobs when new commits lands on PR

## Test Plan

Test on PR

## Additional Information

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -12,6 +12,10 @@ on:  # yamllint disable-line rule:truthy
       - v1.*
     types: [opened, reopened, synchronize]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
 env:
   DEVCONTAINER_IMAGE: "ghcr.io/magma/devcontainer:sha-007ddb9"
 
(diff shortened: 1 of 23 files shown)


themarwhal (approved):

.github/workflows/agw-workflow.yml
```diff
@@ -12,6 +12,10 @@ on:  # yamllint disable-line rule:truthy
       - v1.*
     types: [opened, reopened, synchronize]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
```
themarwhal, line 16: what does the || mean in this context?

quentinDERORY: it's the regular `or` .
The concurrency group will be the pr number if it's defined or the sha of the commit.

themarwhal: ohhh i see - i guess pr number doesn't exist for master runs?

quentinDERORY: yes exact!

hcgatewood (commented):
This looks great! The one concern I see with this is we wouldn't get a green master build when there are successive merges to master -- so for especially long + flaky checks (lte-integ-test), this change might drastically reduce the number of passing commits

Is there any way to cancel-in-progress only if it's a pull request? cc @themarwhal

quentinDERORY:
> This looks great! The one concern I see with this is we wouldn't get a green master build when there are successive merges to master -- so for especially long + flaky checks (lte-integ-test), this change might drastically reduce the number of passing commits
> 
> Is there any way to cancel-in-progress only if it's a pull request? cc @themarwhal

This is only cancelling for pull request. There will be no cancelling on master or on release branches

PR #10003, opened 2021-11-01
refactor: move getindexerversion to more discoverable area

reginawang3495:
## Summary
Moved `versioner.GetIndexerVersion` to a more discoverable place

## Test Plan
Ran orc8r tests + CI

## Additional Information

### DIFF
--- a/orc8r/cloud/go/services/state/indexer/reindex/queue.go
+++ b/orc8r/cloud/go/services/state/indexer/reindex/queue.go
@@ -84,6 +84,10 @@ type Versioner interface {
 	// Intended for use when automatic reindexing is disabled.
 	GetIndexerVersions() ([]*indexer.Versions, error)
 
+	// GetIndexerVersion gets the tracked indexer versions for an indexer ID.
+	// Returns nil if not found.
+	GetIndexerVersion(indexerID string) (*indexer.Versions, error)
+
 	// SetIndexerActualVersion sets the actual version of an indexer, post-reindex.
 	// Intended for use when automatic reindexing is disabled.
 	SetIndexerActualVersion(indexerID string, actual indexer.Version) error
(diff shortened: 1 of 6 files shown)


hcgatewood (changes requested):
I love the intent behind this change, but I'm leaning toward leaving things as-is. Here's why:

We want MECE+minimal interfaces as much as possible -- "mutually exclusive, collectively exhaustive && minimal"

- Mutually exclusive: each method should strive to perform a unique functionality that's not replicated in part or full by other methods
- Collectively exhaustive: provides all the functionality needed
- Minimal: provides the minimum set of methods required

So if one method can be implemented by calling another method, then the interface is no longer minimal. Instead, we can move "default" / "implementation-agnostic" methods into package-level functions

It would be ideal if Golang would provide some affordance for providing interface-level methods (i.e. attaching these defaults to the interface itself, where they're implemented by calling the interface's methods). But, unfortunately Go doesn't support that

So the de facto solution is: package-level default functions that provide value-add wrappers on top of the interface

Lmk if you have questions about that, happy to chat about it during our 1-1!

reginawang3495:
Ok sounds good, I think this has to do with some nuances of Golang that I just assumed (like that `versioner` is and will always be the sole implementation of `Versioner` so I wouldn't have to think about minimal interfaces)

I am not super certain why `Versioner` isn't just a `struct` rather than a `interface` + `struct` (i think for testing reasons?), but that's ok

hcgatewood:
Right so it definitely could just be a struct. One principle reason to use an interface for single-impl interfaces is to be able to mock the interface -- so e.g. able to assert that certain things were called on the versioner object

This made more sense when this was part of the queue iface, since that's a complex iface we definitely want to mock. Here it may not be as strong of a win. But in either case, I'm always okay with leaning a little too far toward making things interfaces, because that's a much easier problem than the other direction 🙂

reginawang3495:
Sounds good, thanks for the info!

Issue #10027, opened 2021-11-02
[agw][orc8r] Log and exception handling audit in Python code


alexzurbonsen:
## Goal
 
Audit of logs and exception handling in order to (a) obtain cleaner and clearer logs and error handling and (b) enable Sentry with usable error rate and without performance issues.

## Path

_First step_: Assessment of effort levels for individual services. To that end statistics on the number of log and exception handling statements as well as the overall amount of code are needed.

_Next steps_: Look at error handling in different Python services and enable Sentry if possible.

alexzurbonsen:
**Statistics on logging and exception handling (for Python in AGW and orc8r)**

As a first step you find statistics on the number of log statements, exception handling statements and code size in the tables below. Details are explained in the following and underneath the tables.

General details on counting: We scanned python code in lte/gateway/python/magma and orc8r/gateway/python/magma for the logging and exception handling keywords, ignoring comments and tests directories. Results are reported in two separate tables, one for logging statements and one for exception handling statements.

**Logging**
| directory     | lines of code (Python) | critical | fatal | exception | error | warning | info | debug | total | total >= error | lines with log statements (in %) | fraction of log statements >= error (in %) |
|---------------|------------------------|----------|-------|-----------|-------|---------|------|-------|-------|----------------|----------------------------------|--------------------------------------------|
| pipelined     | 12871                  | 0        | 0     | 6         | 130   | 32      | 61   | 144   | 373   | 136            | 2.90%                            | 36.46%                                     |
| enodebd       | 7053                   | 2        | 0     | 3         | 21    | 32      | 13   | 29    | 100   | 26             | 1.42%                            | 26.00%                                     |
| magmad        | 3888                   | 0        | 0     | 4         | 45    | 12      | 44   | 17    | 122   | 49             | 3.14%                            | 40.16%                                     |
| subscriberdb  | 2938                   | 0        | 1     | 5         | 14    | 13      | 20   | 8     | 61    | 20             | 2.08%                            | 32.79%                                     |
| common        | 2340                   | 0        | 0     | 1         | 13    | 5       | 22   | 3     | 44    | 14             | 1.88%                            | 31.82%                                     |
| mobilityd     | 2220                   | 0        | 0     | 0         | 15    | 11      | 22   | 39    | 87    | 15             | 3.92%                            | 17.24%                                     |
[...]

PR #10029, opened 2021-11-02
chore(agw): Enable pylint test for kernsnoopd

sebathomas:
## Summary
Address pylint issues in kernsnoopd and enable pylint test. Also enable pylint for pylint test itself.

## Test plan
Executed pylint test

PR #10033, opened 2021-11-02
chore(smsd-relay): Don't log an error in happy path

sebathomas:
## Context
Going through the error handling in some of the smaller services for #10027, we noticed that the smsd relay logs an error for each message that it tries to send. It looks to me like this should be on info (or even debug) level instead.

## Summary
Log message on info level instead of error for each sms that the relay tries to send.

### DIFF
--- a/lte/gateway/python/magma/smsd/relay.py
+++ b/lte/gateway/python/magma/smsd/relay.py
@@ -72,7 +72,7 @@ class SmsRelay(Job):
             return
 
         for msg in smsd_resp.messages:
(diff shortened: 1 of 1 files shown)


themarwhal (approved):

themarwhal:
Hey @sebathomas ! you might want to rebase and resubmit, i think the GHA jobs might be stuck somehow.

sebathomas:
Rebased but not sure it's going help, same issue in #10029.

pshelar (approved):

PR #10036, opened 2021-11-02
fix(orc8r): fixed fab command to connect gateway

christinewang5:
## Summary
The API request `networks/{network_id}/gateways` now returns a paginated gateways rather than just a mapping of gw ID to full resource, so we need to unwrap the mapping. 

## Test Plan
Ran `fab -f dev_tools.py register_vm`

### DIFF
--- a/orc8r/tools/fab/dev_utils.py
+++ b/orc8r/tools/fab/dev_utils.py
@@ -118,11 +118,14 @@ def get_next_available_gateway_id(
     Returns:
         Next available gateway ID in the form gwN
     """
-    # gateways is a dict mapping gw ID to full resource
-    gateways = cloud_get(
+    # res is a dict mapping gw ID to full resource
+    res = cloud_get(
         f'networks/{network_id}/gateways',
         admin_cert=admin_cert,
     )
+
+    # res could also return paginated gateways, so need to unwrap the mapping
+    gateways = res.get('gateways', res)
 
     n = len(gateways) + 1
     candidate = f'gw{n}'


themarwhal (approved):

orc8r/tools/fab/dev_utils.py
```diff
@@ -118,12 +117,14 @@ def get_next_available_gateway_id(
...
-    # gateways is a dict mapping gw ID to full resource
+    # 'gateways' is a dict mapping gw ID to full resource
     gateways = cloud_get(
         f'networks/{network_id}/gateways',
         admin_cert=admin_cert,
     )
-
+    # 'gateways' could also paginated gateways, so need to unwrap the mapping
+    if 'gateways' in gateways:
+        gateways = gateways['gateways']
```
hcgatewood: I think this ci check is saying to use `gateways = gateways.get('gateways', gateways)` lol. Up to you

For clarity, maybe let's rename the outer object to `res`. So then it would be

```
res = cloud_get(...)
gateways = res['gateways'] or res

File: orc8r/tools/fab/dev_utils.py [hunk 2]
```diff
@@ -118,12 +117,14 @@ def get_next_available_gateway_id(
     Returns:
         Next available gateway ID in the form gwN
     """
-    # gateways is a dict mapping gw ID to full resource
+    # 'gateways' is a dict mapping gw ID to full resource
```
hcgatewood: Nit here and below: I'm generally not a fan of putting quotes around variable names, because then it's unclear whether you're referring to a variable or a string. To me it's clear to just leave it as it was, but up to you

hcgatewood (approved):
Lgtm, thanks for the fix! Two nits then msg me to merge 💯

PR #10043, opened 2021-11-02
fix(ci): Fix GCC warnings job not handling filenames with spaces

hcgatewood:
## Summary

Changed files are communicated, by default, as space-delimited. This doesn't work well when files have spaces in their names -- so the check intentionally fails.

Instead, we can communicate the output as JSON.

## Test Plan

PR CI checks

## Additional Information


tmdzk (approved):

electronjoe:
Thanks Hunter!

PR #10046, opened 2021-11-03
feat(feg): base accounting heartbeat feature

emakeev:

## Summary

Add idle AP accounting heartbeat to update the smart contract accounting service when AP doesn't serve any clients, but up and ready.

## Test Plan
unit tests; test on a partner setup later

## Additional Information

### DIFF
--- a/feg/gateway/services/aaa/base_acct/heartbeat_reset_go1_15.go
+++ b/feg/gateway/services/aaa/base_acct/heartbeat_reset_go1_15.go
@@ -1,0 +1,22 @@
+//
+// Copyright 2021 The Magma Authors.
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//go:build go1.15
+// +build go1.15
+
+// Package base_acct provides a client API for interacting with the
+// base_acct cloud service
+package base_acct
+
+func resetHeartbeat() {
+	heartbeatTicker.Reset(HeartbeatDuration)
+}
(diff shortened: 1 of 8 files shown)


feg/gateway/services/aaa/base_acct/heartbeat_reset_go1_15.go
```diff
@@ -0,0 +1,22 @@
...
+// LICENSE file in the root directory of this source tree.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//go:build go1.15
+// +build go1.15
```
uri200, line 14: does it mean that package `base_acct` is going to be built with go.1.15?

feg/gateway/services/aaa/base_acct/heartbeat_reset_prego1_15.go
```diff
@@ -0,0 +1,21 @@
...
+// LICENSE file in the root directory of this source tree.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//go:build !go1.15
+// +build !go1.15
```
uri200, line 14: what is this needed?  You create two binaries?

uri200 (approved):

feg/gateway/services/aaa/base_acct/heartbeat_reset_go1_15.go
emakeev: It just ticker reset capability, it didn't exist prior to Go 1.15, so the reset call for older compilers will be noop

feg/gateway/services/aaa/base_acct/heartbeat_reset_prego1_15.go
emakeev: one is "+build go1.15" -> Go 1.15 or newer & one is "+build !go1.15" -> older then Go 1.15

PR #9398, opened 2021-10-01
test(mme): Adding unit tests for S1AP MME handler functions

ardzoht:
Signed-off-by: Alex Rodriguez <ardzoht@gmail.com>

## Summary

- Adds first set of unit tests for S1AP tasks, using test mock ITTI framework introduced on #8487 
    - s1ap_handle_initial_ue_message
    - s1ap_handle_s1_setup_request
    - s1ap_handle_initial_context_setup_response
    - s1ap_handle_ue_context_release_request
    - s1ap_handle_ue_cap_indication
    - added unit tests for s1ap enb stale association
- Adds MME app mock class handler
- Refactoring some S1AP related code 
- Removing unused code and imports

## Test Plan

- make test_oai

## Additional Information

- [ ] This change is backwards-breaking

PR #10047, opened 2021-11-03
test(mme): Add further s1ap mme handlers tests

ardzoht:
## Summary

- Similar to #9398, this PR adds tests for uncovered S1AP handlers and procedures
- Refactors some of the repeated code into helper functions
- Adds missing declaration for some of the s1ap_mme_handlers functions

## Test Plan

- `make test_oai`

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/test/s1ap_task/test_s1ap_mme_handlers.cpp
+++ b/lte/gateway/c/core/oai/test/s1ap_task/test_s1ap_mme_handlers.cpp
@@ -72,6 +72,12 @@ class S1apMmeHandlersTest : public ::testing::Test {
     task_sctp.detach();
 
     s1ap_mme_init(&mme_config);
+
+    // Setup new association for testing
+    state     = S1apStateManager::getInstance().get_state(false);
+    assoc_id  = 1;
+    stream_id = 0;
+    setup_new_association(state, assoc_id);
     std::this_thread::sleep_for(std::chrono::milliseconds(500));
   }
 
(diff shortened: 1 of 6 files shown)


lte/gateway/c/core/oai/tasks/s1ap/s1ap_mme_nas_procedures.c
```diff
@@ -444,17 +445,6 @@ status_code_e s1ap_mme_handle_nas_non_delivery(
       S1ap_NASNonDeliveryIndication_IEs_t, ie, container,
       S1ap_ProtocolIE_ID_id_Cause, true);
 
-  /*
```
ardzoht: deleted repeated code

ulaskozat:
conflicting file.

lte/gateway/c/core/oai/test/s1ap_task/test_s1ap_mme_handlers.cpp
```diff
@@ -381,5 +372,338 @@ TEST_F(S1apMmeHandlersTest, GenerateUEContextReleaseCommand) {
...
+  // Freeing pdu and payload data
+  bdestroy_wrapper(&payload);
+  ASN_STRUCT_FREE_CONTENTS_ONLY(asn_DEF_S1ap_S1AP_PDU, &pdu);
+  ASN_STRUCT_FREE_CONTENTS_ONLY(asn_DEF_S1ap_S1AP_PDU, &pdu_s1);
+
+  // Sleep to ensure that messages are received and contexts are released
+  std::this_thread::sleep_for(std::chrono::milliseconds(500));
+}
+
+TEST_F(S1apMmeHandlersTest, HandleS1apErabRelCmd) {
```
ulaskozat, line 534: In general I think we should have expect calls for downstream and upstream tasks (e.g., expect to be called zero times). also should not we checking s1ap states at critical parts?

ardzoht: @ulaskozat Added state validation for tests that was missing it, as well as expect calls.

ulaskozat (approved):
LGTM

PR #10070, opened 2021-11-03
chore: mark service303 test as flaky

themarwhal:

## Summary
Noticed that this test was failing on a PR: https://github.com/magma/magma/runs/4093727669?check_suite_focus=true

By marking a test flaky, this test will get attempted 3 times before declaring the whole test run as a failure.

## Additional Information

### DIFF
--- a/orc8r/gateway/c/common/service303/test/BUILD.bazel
+++ b/orc8r/gateway/c/common/service303/test/BUILD.bazel
@@ -35,6 +35,7 @@ cc_test(
     name = "service303_test",
     size = "small",
     srcs = ["test_service303.cpp"],
+    flaky = True,
     deps = [
         "//orc8r/gateway/c/common/service303",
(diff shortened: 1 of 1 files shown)


ardzoht (approved):
lgtm, do we have any specific test under these that is failing?

themarwhal:
> lgtm, do we have any specific test under these that is failing?


Hey @ardzoht ! It looks like this is the error
```
service303_test: orc8r/gateway/c/common/service303/test/test_service303.cpp:112: static const io::prometheus::client::MetricFamily& magma::Service303Test::findFamily(const magma::orc8r::MetricsContainer&, const string&): Assertion `false' failed.
``` 
(You can search for it in the GHA job linked in the summary)

I think it's because this test brings up an actual magma test service for testing. This might be conflicting with other bad tests that do this because Bazel runs all tests in parallel unless specified otherwise.

PR #10073, opened 2021-11-03
feat(agw): Config Service for src/go

mstre123:
Add config service proto and gen.
Impl config service grpc.
Rewire service to use new config service.
Fix socket cleanup for upstream socket.

## Summary
Replacing mconfig with config service for src/go

Follow up work:
Right now we update the config but for some of these to take effect we restart the service.
Left the funcs in there that read configs from a file - we probably want to reuse that for starting the services.

## Test Plan

Unit testing
Running integ tests

## Additional Information

- [] This change is backwards-breaking

This breaks the mconfig impl in src/go - but that wasn't operational yet so we should be fine.


src/go/agwd/config/config.go
```diff
@@ -193,3 +197,9 @@ func LoadConfigFile(cm *ConfigManager, path string) error {
 	cm.Merge(loaded)
 	return nil
 }
+
+// UpdateConfig updates CongfigManager with a config file.
+func (cm *ConfigManager) UpdateConfig(config *config.AgwD) error {
```
markjen, line 203: should we just add `Merge` to the `Configer` interface instead? Or did you intend to add validity/error checks here? (if the latter, let's add a TODO comment)

markjen: also: add unit test pls

src/go/agwd/server/server.go
```diff
@@ -127,8 +137,32 @@ func startSctpdUplinkServer(
...
+func startConfigServer(
+	cfgr config.Configer, logger log.Logger,
+) {
+	target := config.ParseTarget(cfgr.Config().GetConfigServiceTarget())
+	if target.Scheme == "unix" {
+		cleanupUnixSocketOrDie(logger, target.Endpoint)
+	}
+
+	listener, err := net.Listen(target.Scheme, target.Endpoint)
+	if err != nil {
```
markjen, line 146: 1. Hmmm I wonder if this could be DRY'd up
2. Multiple services can be configured with the same target. we should update/refactor this code so that the startup code creates listeners for each unique target then maps services to those listeners. this can be in a followup

src/go/service/router.go
```diff
@@ -50,3 +56,7 @@ func (d *router) SctpdDownlinkClient() sctpd.SctpdDownlinkClient {
 func (d *router) SctpdUplinkClient() sctpd.SctpdUplinkClient {
 	return d.sctpdUplinkClient
 }
+
+func (d *router) ConfigClient() config.ConfigClient {
```
markjen: unit test?

src/go/agwd/server/server.go
mstre123: Agreed we can clean this up, ill add a follow on task for that to avoid adding to much to this pr. #10131

mstre123: Agreed we can clean this up, ill add a follow on task for that to avoid adding to much to this pr. #10131

src/go/agwd/config/config.go
mstre123: exactly - figured in the future we would we would break up the api a bit or limit what fields users can update etc. i'll add a todo.

src/go/agwd/server/server.go
```diff
@@ -18,6 +18,8 @@ import (
 	"runtime"
 	"time"
 
+	configpb "github.com/magma/magma/src/go/protos/magma/config"
+	configser "github.com/magma/magma/src/go/service/config"
```
markjen: nit: not sure what our convention would be, but probably more `service_config`?

src/go/service/config/config_server.go
```diff
@@ -0,0 +1,51 @@
...
+	return &pb.GetConfigResponse{Config: config}, nil
+}
+
+// UpdateConfig calls Config.UpdateConfig.
+func (c *ConfigServer) UpdateConfig(ctx context.Context, req *pb.UpdateConfigRequest) (*pb.UpdateConfigResponse, error) {
+	c.Logger.
+		With("config", req.GetConfig()).
+		Debug().Print("UpdateConfig")
+	if err := c.Configer.UpdateConfig(req.Config); err != nil {
+		return &pb.UpdateConfigResponse{Config: c.Configer.Config()}, err
```
markjen: do we want to return a response here? or should this be `return nil, err`

src/go/service/config/config_server_integ_test.go
```diff
@@ -0,0 +1,95 @@
+package config
```
markjen, line 12: copyright

src/go/service/router.go
```diff
@@ -23,23 +24,28 @@ type Router interface {
 	SctpdDownlinkClient() sctpd.SctpdDownlinkClient
 	// SctpdUplinkClient provides a sctpd.SctpdUplinkClient
 	SctpdUplinkClient() sctpd.SctpdUplinkClient
+	// ConfigClient provides a config.ConfigClient
+	ConfigClient() config.ConfigClient
```
markjen: Hmm... who would use this?

Extreme view: we could have all internal readers of Config use this; thus Config would become a service?

markjen (approved):

src/go/service/config/config_server.go
mstre123: yeah i debated this one a bit - so the config returned is actually fetched from the config object, so it would be returning the current state but i can see how thats confusing - will remove

src/go/service/router.go
mstre123: ill remove it was looking at it more as maybe the internal tests are running on multiple boxes and need to talk to each other but ill leave it out for now and bring it back if we need it later on. 
reading configs internally over grpc can lead to a ton of problems i think we should avoid.

PR #10080, opened 2021-11-03
fix: Fix continous deletion of default drop flow

ymasmoudi:

## Summary

Under high load, pipelined continuously sends FlowMod requests to ovs in order to delete the default drop rule.
This can reach as high as 3000 flow_mods requests per minute causing performance issues in control plane.

```
2021-11-03T22:18:49.038Z|02190|connmgr|INFO|cwag_br0<->tcp:127.0.0.1:6633: 2620 flow_mods in the 48 s starting 59 s ago (190 adds, 2430 deletes)
2021-11-03T22:20:46.771Z|02191|connmgr|INFO|cwag_br0<->tcp:127.0.0.1:6633: 2430 flow_mods in the 1 s starting 10 s ago (2430 deletes)
```

The reason behind this issue is that the rules store does not keep a rule version for the default drop flow,
Which means that curr_version will always be different from record.rule_version, triggering a delete request.
```
records {
  sid: "IMSI68248225537042"
  rule_id: "Facebook"
  rule_version: 1
}
records {
  sid: "IMSI68248225537042"
  rule_id: "Allow_All_Internet"
  rule_version: 1
}
records {
  sid: "IMSI68248225537042"
  rule_id: "internal_default_drop_flow_rule"
}
  ```

## Test Plan
Tested on a cwf setup

### DIFF
--- a/lte/gateway/python/magma/pipelined/rule_mappers.py
+++ b/lte/gateway/python/magma/pipelined/rule_mappers.py
@@ -146,7 +146,7 @@ class SessionRuleToVersionMapper:
         with self._lock:
             version = self._version_by_imsi_and_rule.get(key)
             if version is None:
-                """Set to zero to allow proper cleanup of old flows"""
+                # Set to zero to allow proper cleanup of old flows
                 version = 0
         return version
 
--- a/lte/gateway/python/magma/pipelined/rule_mappers.py
+++ b/lte/gateway/python/magma/pipelined/rule_mappers.py
@@ -146,7 +146,8 @@ class SessionRuleToVersionMapper:
         with self._lock:
             version = self._version_by_imsi_and_rule.get(key)
             if version is None:
-                version = -1
+                """Set to zero to allow proper cleanup of old flows"""
+                version = 0
         return version
 
     def remove(self, imsi: str, ip_addr: IPAddress, rule_id: str, version: int):
--- a/lte/gateway/python/magma/pipelined/tests/test_rule_mappers.py
+++ b/lte/gateway/python/magma/pipelined/tests/test_rule_mappers.py
@@ -75,13 +75,13 @@ class RuleMappersTest(unittest.TestCase):
             self._session_rule_version_mapper.get_version(
                 imsi, convert_ipv4_str_to_ip_proto(ip_addr), rule_ids[0],
             ),
-            -1,
+            0,
         )
         self.assertEqual(
             self._session_rule_version_mapper.get_version(
                 imsi, convert_ipv4_str_to_ip_proto(ip_addr), rule_ids[1],
             ),
-            -1,
+            0,
         )
 
     def test_session_rule_version_mapper_cwf(self):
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lte/gateway/python/magma/pipelined/app/enforcement_stats.py [hunk 1]
```diff
@@ -640,6 +640,9 @@ def _delete_old_flows(self, records):
             if current_ver == record.rule_version:
```
koolzz: I've initially added this for the drop flow rule case. The current_ver should be 0 and record.rule_version is also 0.

ymasmoudi: @koolzz I added logs just after this condition

+                self.logger.error('delete rule %s %s %s', record.rule_id, record.rule_version, current_ver)

ERROR:EnforcementStatsController:delete rule internal_default_drop_flow_rule 0 -1
ERROR:EnforcementStatsController:delete rule internal_default_drop_flow_rule 0 -1

The added condition does not catch up the default drop flow.

koolzz: ahh right I see what happened, basically initially it was 0, but when we were updating some rule cleanup the default return on rule not found changed to -1 on this pr https://github.com/magma/magma/commit/8dbf40f29cfe71084d1778f8d59d6429cf9c0f21.

Which accidentally broke this logic.
@pshelar  is -1 required, or should we just go back to returning 0 in `get_version`?

koolzz:
Just wanted to note that this is an awesome find, thanks for finding and fixing!

lte/gateway/python/magma/pipelined/app/enforcement_stats.py
pshelar: ok, gets revert to return zero.
Can you also add comment explaining why it needs to be zero.

pshelar (approved):

lte/gateway/python/magma/pipelined/rule_mappers.py
```diff
@@ -146,7 +146,8 @@ def get_version(self, imsi: str, ip_addr: IPAddress, rule_id: str) -> int:
         with self._lock:
             version = self._version_by_imsi_and_rule.get(key)
             if version is None:
-                version = -1
+                """Set to zero to allow proper cleanup of old flows"""
```
koolzz: throws ```W0105: 149, 16: String statement has no effect (pointless-string-statement)```
replace with ```#Set to zero to allow proper cleanup of old flows```

ymasmoudi: @koolzz fixed. actually, i followed the same style as the other comments in the file.

koolzz: linter works in mysterious ways, I think its just due to it being on one line. Thanks!

koolzz (approved):

koolzz:
Pretty sure this also resolves https://github.com/magma/magma/issues/10083, will run some more tests, but looks promising

PR #9981, opened 2021-10-29
feat(orc8r): K8s Include labels and annotations based on the image/chart being deployed

joary:
## Summary

1. Include a label in the orc8r and lte-orc8r charts including the current version being deployed
    1. The label value comes directly from the docker image tag, which is the closes information the k8s have to the magma version
2. Include an annotation in the orc8r and lte-orc8r charts including the cluster release name being deploied (production/staging/green/blue)
    1. Annotation values comes from the helm chart name being deployed (configurable in terraform values)
3. OBS: For bot annotation and labels some K8s objects does not include this label since they’re independent of the magma version being run. Non exaustive list: Secrets/ConfigMaps/Certs/Roles/RoleBindings

Also fixed some bugs to allow testing:

1. Avoid to running metrics storeconfig job if prometheus is not enabled
2. Avoid creating PodDisruptionBudgets if not enabled

## Test Plan

Expected result is that all the k8s objects being deployed have an additional annotation and label as shown bellow. (Existing labels still remains)

```
metadata:
  annotations:
    release-name: lte-staging
  labels:
    image-version: 1.5.0-8952929
  name: orc8r-ha
```

Using the helm template it is possible to list which services have the target label and annocation:

```
helm template \
    {staging|lte-staging}\
    {orc8r|lte-orc8r} \
    --repo <REPO> (https://raw.githubusercontent.com/joary/helm-chart-labels-n-annotations/master/)
    --username <USER> \
    --password <PASS> \
    --version 1.5.23 \
    --timeout 600s \
    --values values.yml \
    --debug —dry-run | egrep "Source|release-name|image-version"
```

OBS: values.yml needs to be configured in a proper way to deploy a standard Orc8r, no new value is needed.

**Result for `orc8r` helm chart:**

```
# Source: orc8r/templates/accessd.pdb.yaml
    release-name: orc8r
    image-version: 1.5.0-8952929
# Source: orc8r/templates/analytics.pdb.yaml
    release-name: orc8r
    image-version: 1.5.0-8952929
# Source: orc8r/templates/bootstrapper.pdb.yaml
    release-name: orc8r
    image-version: 1.5.0-8952929
# Source: orc8r/templates/certifier.pdb.yaml
    release-name: orc8r
    image-version: 1.5.0-8952929
# Source: orc8r/templates/configurator.pdb.yaml
    release-name: orc8r
    image-version: 1.5.0-8952929
# Source: orc8r/templates/ctraced.pdb.yaml
    release-name: orc8r
[...]

PR #10085, opened 2021-11-04
feat(orc8r): K8s Include labels and annotations based on the image/chart being deployed

### LINKED ISSUE #10103: Orc8r Pod Labels

joary:
## Summary

In essence the same as https://github.com/magma/magma/pull/9981, adding two points:

1. The labels also to pods/containers
2. The chart-version as a new label to both lte-orc8r and orc8r charts.

More details:

1. Include a label in the orc8r and lte-orc8r charts including the current version being deployed
    1. The label value comes directly from the docker image tag, which is the closes information the k8s have to the magma version
2. Include an annotation in the orc8r and lte-orc8r charts including the cluster release name being deploied (production/staging/green/blue)
    1. Annotation values comes from the helm chart name being deployed (configurable in terraform values)
3. OBS: For bot annotation and labels some K8s objects does not include this label since they’re independent of the magma version being run. Non exaustive list: Secrets/ConfigMaps/Certs/Roles/RoleBindings

Also fixed some bugs to allow testing:

1. Avoid to running metrics storeconfig job if prometheus is not enabled
2. Avoid creating PodDisruptionBudgets if not enabled

## Test Plan

A - Check the labels / annotations for each k8s object with helm template:

**Result for 'orc8r' helm chart:** https://gist.github.com/joary/e7156e6f6fce6834323aaa709925f6b8

**Result for 'lte-orc8r' helm chart:** 
https://gist.github.com/joary/aa3a8bf3974bc5ec9ca98ed35cd57325

B - Check the labels for each pod after deploying both orc8r and lte-orc8r charts

`kubectl get pods -n <namespace> --show-labels`:

https://gist.github.com/joary/c95906e247634f21f277d85772222c42

## Additional Information

### DIFF
--- a/orc8r/cloud/helm/orc8r/charts/nms/templates/nginx-proxy-deployment.yaml
+++ b/orc8r/cloud/helm/orc8r/charts/nms/templates/nginx-proxy-deployment.yaml
@@ -22,7 +22,12 @@ kind: Deployment
 metadata:
   name: nms-nginx-proxy
   labels:
+{{ include "default-labels" . | indent 4 }}
+{{ include "magmalte-image-version-label" . | indent 4 }}
 {{ tuple $envAll "nms" "nginx" | include "nms.labels" | indent 4 }}
+  annotations:
+{{ include "release-name-annotation" . | indent 4 }}
+
 spec:
   replicas: {{ .Values.nginx.replicas }}
   strategy:
--- a/orc8r/cloud/helm/orc8r/charts/nms/templates/nginx-proxy-deployment.yaml
+++ b/orc8r/cloud/helm/orc8r/charts/nms/templates/nginx-proxy-deployment.yaml
@@ -23,7 +23,7 @@ metadata:
   name: nms-nginx-proxy
   labels:
 {{ include "default-labels" . | indent 4 }}
-{{ include "magmalte-image-version-label" . | indent 4 }}
+{{ include "nginx-image-version-label" . | indent 4 }}
 {{ tuple $envAll "nms" "nginx" | include "nms.labels" | indent 4 }}
   annotations:
 {{ include "release-name-annotation" . | indent 4 }}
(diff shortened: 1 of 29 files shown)


hcgatewood (commented):
Looks like this is still stacked on the previous PR (?). Can you rebase it onto latest master then re-request a review?

orc8r/cloud/helm/orc8r/charts/nms/templates/nginx-proxy-deployment.yaml
```diff
@@ -35,10 +35,14 @@ spec:
   selector:
     matchLabels:
 {{ tuple $envAll "nms" "nginx" | include "nms.selector-labels" | indent 6 }}
+{{ include "magmalte-image-version-label" . | indent 6 }}
   template:
     metadata:
       labels:
 {{ tuple $envAll "nms" "nginx" | include "nms.selector-labels" | indent 8 }}
+{{ include "magmalte-image-version-label" . | indent 8 }}
```
arunuke: Should we use magmalte-image-version-label for nginx-proxy deployment as well? Looking at our [artifactory for helm charts](https://artifactory.magmacore.org/artifactory/helm/), we have one for orc8r and one for lte-orc8r which I presume handles it for the controller and magmalte images respectively. If so, is there a possibility that nginx-proxy and magmalte could have different chart versions?

arunuke (changes requested):
A general comment is to just check if all the pods we deploy are tagged with the image and chart versions as desired.

orc8r/cloud/helm/orc8r/charts/nms/templates/nginx-proxy-deployment.yaml
joary: > Should we use magmalte-image-version-label for nginx-proxy deployment as well

No, I think we should use the `nginx-image-version-label`. I corrected the PR to reflect that.

> If so, is there a possibility that nginx-proxy and magmalte could have different chart versions?

No, both nginx deployments will use the same [helm value configuration](https://github.com/magma/magma/blob/master/orc8r/cloud/helm/orc8r/values.yaml#L105).

Being more clear, there are two nginx deployments: 
1. [One handling the public access to NMS](https://github.com/magma/magma/blob/master/orc8r/cloud/helm/orc8r/charts/nms/templates/nginx-proxy-deployment.yaml)
2. [Another handling the public access to Orc8r services](https://github.com/magma/magma/blob/master/orc8r/cloud/helm/orc8r/templates/nginx.deployment.yaml)

We expect them to have the same image-label value, though the char-version will be different, #1 will receive the NMS sub-chart version, and the #2 will receive Orc8r version.

arunuke (approved):
Thanks for the changes. LGTM.

I remember we had a plan to upgrade the EKS version as well (due to future support for nginx ingress). Is it going in a different PR?

joary:
> I remember we had a plan to upgrade the EKS version as well (due to future support for nginx ingress). Is it going in a different PR?

Yes, I included in https://github.com/magma/magma/pull/10212, which is merged now. Rationale is that PR changes the k8s resources already, made more sense to be there.

hcgatewood (approved):
Skimmed, lgtm

PR #10088, opened 2021-11-04
chore(pipelined): ipv6-solicitation: remove dependency on tun-id map.

pshelar:
Egress tun-id can be retrieved from ovs flow. Use it before checking
the tun-id map.

## Additional Information


koolzz (approved):
lgtm, do unit tests work fine?

koolzz:
With this change should we deprecate the tunnel id store?
Also lets verify this with an ipv6 test to make sure it works

pshelar:
> With this change should we deprecate the tunnel id store? Also lets verify this with an ipv6 test to make sure it works

yes, we need to do some restructuring here. I wanted to validate it bit more before removing the fallback. ticket: #10118

PR #10097, opened 2021-11-04
fix(agw): Persistent IPv6 config

### LINKED ISSUE #10096: adding default route during provisioning fails
### Your Environment

- **Version:** master

### Describe the Issue

Running `vagrant up --provison magma` fails on all but the first time

**To Reproduce** (in `$MAGMA_ROOT/lte/gateway` folder)
1. `vagrant destroy magma`
2. `vagrant up magma` -> the first provisioning works
3. `vagrant halt magma && vagrant up --provision magma`  -> this fails with `fatal: [magma]: FAILED! => {"cmd": "ip -6 a add  2020::10/64 dev eth0 && ip -6 r add default via 2020::1", "stderr": "RTNETLINK answers: File exists"}`

**Reason for the current behaviour**
In the Vagrantfile we define a shell provisioning step that adds a default route. The Ansible provisioning also tries to add that route and fails as it already exists.

So why does it not fail on the very first provisioning? This is because the Ansible playbook will restart the `networking.service` which results in a call to `ifdown eth0` which also deletes the route created in the shell provisioning step. After the route is deleted the Ansible step that creates it again runs without errors. However, after the initial provisioning, the `networking.service` fails during boot, so a restart of `networking.service` doesn't stop the `networking.service` anymore. Thus `ifdown eth0` is not called during subsequent provisionings.

References:
- https://github.com/magma/magma/blob/master/lte/gateway/Vagrantfile#L43
- https://github.com/magma/magma/blob/master/lte/gateway/deploy/roles/magma/tasks/main.yml#L448

jheidbrink:
## Summary

fixes #10096 

Issue #10096 was caused by trying to add a route that was already added before. These changes persist the IPv6 address and default route in /etc/network/interfaces and thereby delegate their creation to the OS.

## Test Plan

```
cd $MAGMA_ROOT/lte/gateway
vagrant destroy magma
vagrant up magma
vagrant halt magma
vagrant up --provision magma
```

Assert that the second provision works. Assert that after the first provisioning, `eth0` has the IPv6 address `2020::10/64` and there is a default route via `2020::1` (by looking at the output of `ip -6 addr` and `ip -6 route`)

## Additional Information

### DIFF
--- a/lte/gateway/Vagrantfile
+++ b/lte/gateway/Vagrantfile
@@ -39,9 +39,6 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     # iperf3 trfserver routable IP.
     magma.vm.network "private_network", ip: "192.168.129.1", nic_type: "82540EM"
 
-    # IPv6 interface
-    config.vm.provision :shell, inline: "ip -6 r add  2020::10/64 dev eth0 && ip -6 r r default via 2020::1", run: 'always'
-
     magma.vm.provider "virtualbox" do |vb|
       vb.name = "magma-dev"
       vb.linked_clone = true
--- a/lte/gateway/deploy/roles/magma/tasks/main.yml
+++ b/lte/gateway/deploy/roles/magma/tasks/main.yml
@@ -293,6 +293,16 @@
     dest: /etc/network/interfaces.d/gtp
   when: full_provision
 
+- name: Configure IPv6 address on eth0 interface
+  blockinfile:
+    block: |
+        iface eth0 inet6 static
+              address 2020::10/64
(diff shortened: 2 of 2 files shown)


Neudrino (approved):

themarwhal (approved):

themarwhal:
cc @pshelar

themarwhal:
cc @pshelar

uri200 (approved):

pshelar (approved):
Thanks for working on this.

uri200:
Thanks @jheidbrink  this issue was blocking some of our local tests and the PR fixed it
🙏

PR #10098, opened 2021-11-04
refactor: multierror initialize to declare change

reginawang3495:
## Summary

Change all possible multierror initializes to declarations. https://github.com/magma/magma/blob/master/orc8r/cloud/go/obsidian/swagger/spec/combine.go#L68 could not be changed
## Test Plan

Ran all orch8r tests

## Additional Information

### DIFF
--- a/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
+++ b/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
@@ -527,7 +527,7 @@ func validateSubscriberProfiles(ctx context.Context, networkID string, profiles 
 	}
 
 	networkProfiles := networkConfig.(*ltemodels.NetworkCellularConfigs).Epc.SubProfiles
-	errs := &multierror.Error{}
+	var errs *multierror.Error
 	for _, p := range nonDefaultProfiles {
 		if _, ok := networkProfiles[p]; !ok {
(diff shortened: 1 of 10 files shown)


reginawang3495:
Not sure if this actually improves the codebase

I think it might not be especially it's not clear that `multierror` doesn't need to be initialized + I think `multierror` needs to be initialized if certain methods aren't directed used it. 

EG https://github.com/magma/magma/blob/master/orc8r/cloud/go/obsidian/swagger/spec/combine.go#L68 has to stay same (initialized, not declared)

themarwhal (approved):
Why is this considered better? (Not against the change, wondering why this is better practice :D )

reginawang3495:
@themarwhal I don't know either haha, https://github.com/magma/magma/pull/9936#discussion_r742326693 for context.

@hcgatewood What were you thinking for this?

hcgatewood:
@themarwhal @reginawang3495 	https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful. Basic idea is prefer zero values, and nil specifically, for types where they're useful

Though in this case, seems like their zero value isn't useful, so we can preserve how we are currently doing it, actually 🙂

reginawang3495:
Since  https://github.com/magma/magma/blob/master/orc8r/cloud/go/obsidian/swagger/spec/combine.go#L68 has to stay same (initialized, not declared), it would be better to keep everything initialized, so I'll just go ahead and close the PR!

PR #10101, opened 2021-11-04
chore(monitord): Enable selective opt-in logging to Sentry

sebathomas:
Merge after #9974

## Summary

Send errors to Sentry if mobilityd communication fails.  

## Context

In Sentry's opt-in mode only the most important errors should be sent to Sentry. Monitord actually only has a single error log, so opt-in mode would send errors from this log to Sentry while other error logs that might be logged in third-party dependencies will be suppressed.


themarwhal (approved):

pshelar (approved):

themarwhal:
@sebathomas please rebase then it is good to go! :D

alexzurbonsen:
please don't merge yet... Thanks!

PR #10146, opened 2021-11-05
feat(orc8r): reindexer singleton getBatches on new getJobs to reindex

### LINKED ISSUE #10135: Make reindexer a singleton service

reginawang3495:
## Summary

`reindexer_singleton` `getBatches` on new jobs to reindex
- previously, we only got batches on initialization (or `Run`) of the `reindexer_singleton` service
## Test Plan

Added a regression tests and ran all `orc8r` tests

## Additional Information

### DIFF
--- a/orc8r/cloud/go/services/state/indexer/reindex/reindexer_queue_test.go
+++ b/orc8r/cloud/go/services/state/indexer/reindex/reindexer_queue_test.go
@@ -69,10 +69,12 @@ const (
 	nid0 = "some_networkid_0"
 	nid1 = "some_networkid_1"
 	nid2 = "some_networkid_2"
+	nid3 = "some_networkid_3"
 
 	hwid0 = "some_hwid_0"
 	hwid1 = "some_hwid_1"
 	hwid2 = "some_hwid_2"
+	hwid3 = "some_hwid_3"
 
 	id0 = "some_indexerid_0"
 	id1 = "some_indexerid_1"
--- a/orc8r/cloud/go/services/state/indexer/reindex/reindexer_queue_test.go
+++ b/orc8r/cloud/go/services/state/indexer/reindex/reindexer_queue_test.go
@@ -69,16 +69,19 @@ const (
 	nid0 = "some_networkid_0"
 	nid1 = "some_networkid_1"
 	nid2 = "some_networkid_2"
+	nid3 = "some_networkid_3"
 
 	hwid0 = "some_hwid_0"
 	hwid1 = "some_hwid_1"
 	hwid2 = "some_hwid_2"
+	hwid3 = "some_hwid_3"
 
 	id0 = "some_indexerid_0"
 	id1 = "some_indexerid_1"
 	id2 = "some_indexerid_2"
 	id3 = "some_indexerid_3"
 	id4 = "some_indexerid_4"
+	id5 = "some_indexerid_5"
 
 	zero      indexer.Version = 0
 	version0  indexer.Version = 10
@@ -91,6 +94,8 @@ const (
 	version3a indexer.Version = 400
 	version4  indexer.Version = 50
 	version4a indexer.Version = 500
+	version5  indexer.Version = 60
+	version5a indexer.Version = 600
 )
 
 var (
--- a/orc8r/cloud/go/services/state/indexer/reindex/reindexer_queue_test.go
+++ b/orc8r/cloud/go/services/state/indexer/reindex/reindexer_queue_test.go
@@ -69,12 +69,10 @@ const (
 	nid0 = "some_networkid_0"
 	nid1 = "some_networkid_1"
 	nid2 = "some_networkid_2"
-	nid3 = "some_networkid_3"
 
 	hwid0 = "some_hwid_0"
 	hwid1 = "some_hwid_1"
 	hwid2 = "some_hwid_2"
-	hwid3 = "some_hwid_3"
 
 	id0 = "some_indexerid_0"
 	id1 = "some_indexerid_1"
--- a/orc8r/cloud/go/services/state/indexer/reindex/reindexer_queue_test.go
+++ b/orc8r/cloud/go/services/state/indexer/reindex/reindexer_queue_test.go
@@ -64,7 +64,10 @@ const (
 	nStatesToReindexPerCall    = 100 // copied from reindex.go
 	directoryRecordsPerNetwork = 2 * nStatesToReindexPerCall
 	nNetworks                  = 3
-	nBatches                   = 9
+	// 3 networks and 3 batches per network = 3 * 3 = 9
+	nBatches = 9
+	// 4 networks and 3 batches per network = 4 * 3 = 12
+	newNBatches = 12
 
 	nid0 = "some_networkid_0"
 	nid1 = "some_networkid_1"
@@ -95,7 +98,6 @@ const (
 	version4  indexer.Version = 50
 	version4a indexer.Version = 500
 	version5  indexer.Version = 60
-	version5a indexer.Version = 600
 )
 
 var (
@@ -304,6 +306,9 @@ func TestRunUnsafe(t *testing.T) {
 	assertVersions(t, q, id4, version4, version4)
 }
 
+// initReindexTest reports enough directory records to cause 3 batches per network
+// (with the +1 gateway status per network). It creates 3 networks,
+// so numBatches following this method will be 3 * 3 = 9
 func initReindexTest(t *testing.T, dbName string) (reindex.Reindexer, reindex.JobQueue) {
 	indexer.DeregisterAllForTest(t)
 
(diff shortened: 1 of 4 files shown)


orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -48,21 +48,33 @@ func (r *reindexerSingleton) Run(ctx context.Context) {
...
-	batches := r.getReindexBatches(ctx)
 	for {
 		if isCanceled(ctx) {
 			glog.Warning("State reindexing async job canceled")
 			return
 		}
 
-		r.reindexJobs(ctx, indexerID, batches, sendUpdate)
+		jobs, err := r.getJobs(indexerID)
+		if err == nil && jobs != nil {
```
hcgatewood: If there's an err, we want to log it

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -48,21 +48,33 @@ func (r *reindexerSingleton) Run(ctx context.Context) {
 	// TODO(reginawang3495) will support non-nil sendUpdate from Run after removing queue impl
 	var sendUpdate func(string) = nil
 
-	batches := r.getReindexBatches(ctx)
 	for {
```
hcgatewood, line 51: For infinite loops, a good pattern is: 1. an outer func that creates the infinite loop, then, 2. a single func called within the loop. Makes validation and conceptualization easier

So in that case, it would be: let's make inside the for loop just be 3 things

1. is canceled
2. reindex jobs
3. log any err from reindex jobs

This way, reindexJobs will handle creating its own jobs, rather than taking it as an arg

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
```diff
@@ -123,27 +124,47 @@ func TestSingletonRun(t *testing.T) {
 	fail3.AssertExpectations(t)
 	require.Equal(t, 2, reindexSuccessNum)
 	require.Equal(t, 5, reindexDoneNum)
+
+	// Test that a network/hardware pair that has been added after Run
+	// will have its states reindexed as well
```
hcgatewood: Add a newline under this comment, for clarity that it applies to the block rather than the line

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
```diff
@@ -123,27 +124,47 @@ func TestSingletonRun(t *testing.T) {
...
+	addNetworks(t, reindexer)
+
+	idx5 := getIndexerNoIndex(id5, zero, indexer.Version(version5), true)
+	idx5.On("GetTypes").Return(allTypes).Once()
+	idx5.On("Index", mock.Anything, mock.Anything).Return(nil, nil).Times(4 * 3)
```
hcgatewood: This number's a little too magic, even for tests -- why 4*3 ?

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
```diff
@@ -163,5 +184,39 @@ func initSingletonReindexTest(t *testing.T) reindex.Reindexer {
...
 	}
 
-	return reindexer
+	ctx, cancel := context.WithCancel(context.Background())
+	go reindexer.Run(ctx)
+
+	return cancel, reindexer
+}
+
+func addNetworks(t *testing.T, reindexer reindex.Reindexer) {
```
hcgatewood: "reportMoreState"

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
```diff
@@ -163,5 +184,39 @@ func initSingletonReindexTest(t *testing.T) reindex.Reindexer {
...
+	ctx, _ := context.WithCancel(context.Background())
+	go reindexer.Run(ctx)
```
hcgatewood: Hoist this outside this func -- running this indexer is unrelated to reporting more state

hcgatewood (changes requested):
Shape looks good! Couple small requests then I think we'll be ready to merge

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
reginawang3495: It is related since we run `Run` before reporting more state

reginawang3495: Ahhh I see what you mean

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
```diff
@@ -123,27 +124,53 @@ func TestSingletonRun(t *testing.T) {
 	fail3.AssertExpectations(t)
 	require.Equal(t, 2, reindexSuccessNum)
 	require.Equal(t, 5, reindexDoneNum)
+
+	// Test that a network/hardware pair that has been added after Run
+	// will have its states reindexed as well
+
+	ctx, _ := context.WithCancel(context.Background())
+	go reindexer.Run(ctx)
+
```
hcgatewood: Unless I'm misunderstanding: we shouldn't need to re-run the reindexer, because it's already running

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
```diff
@@ -115,6 +113,9 @@ func TestSingletonRun(t *testing.T) {
 	recvCh(t, ch)
 	recvCh(t, ch)
 	recvCh(t, ch)
+
+	// Unregister indexers
+	register(t)
```
hcgatewood: Instead of doing this here, let's register at the beginning of your changes (below)

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
```diff
@@ -163,5 +190,36 @@ func initSingletonReindexTest(t *testing.T) reindex.Reindexer {
...
+
+func reportMoreState(t *testing.T) {
+	configurator_test.RegisterNetwork(t, nid3, "Network 3 for reindex test")
+	configurator_test.RegisterGateway(t, nid3, hwid3, &models.GatewayDevice{HardwareID: hwid3})
+
+	ctxByNetwork := map[string]context.Context{
+		nid3: state_test.GetContextWithCertificate(t, hwid3),
+	}
+
+	// Report enough directory records to cause 3 batches per network (with the +1 gateway status per network)
```
hcgatewood: Move this comment to a func doc comment

hcgatewood (changes requested):
Couple small things, otherwise lgtm

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton_test.go
reginawang3495: I cancelled it in L119 because of the purposefully failing indexers we were running

hcgatewood: Ah I see. Okay so we want to find a way to test this without canceling+restarting the reindexer -- to ensure that it gracefully handles newly-reported state without requiring a restart

We can either remove the `cancel()`, or move these "more state" checks before checking the failed indexers

reginawang3495: Ok sounds good, I'll try moving before the failing indexers

orc8r/cloud/go/services/state/indexer/reindex/reindexer_queue_test.go
```diff
@@ -64,21 +64,27 @@ const (
 	nStatesToReindexPerCall    = 100 // copied from reindex.go
 	directoryRecordsPerNetwork = 2 * nStatesToReindexPerCall
 	nNetworks                  = 3
-	nBatches                   = 9
+	// 3 networks and 3 batches per network = 3 * 3 = 9
+	nBatches = 9
+	// 4 networks and 3 batches per network = 4 * 3 = 12
```
hcgatewood, line 69: 👍

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -48,21 +48,44 @@ func (r *reindexerSingleton) Run(ctx context.Context) {
...
+	if err != nil {
+		return wrap(err, ErrDefault, indexerID)
+	}
+	if jobs == nil {
+		return nil
+	}
+
 	batches := r.getReindexBatches(ctx)
-	return r.reindexJobs(ctx, indexerID, batches, sendUpdate)
+	return r.reindexJobs(ctx, jobs, batches, sendUpdate)
```
hcgatewood, line 76: This can probably just be "return find and reindex jobs". But is fine cause we're gonna remove this method anyway

hcgatewood (approved):
Looks great!

PR #10154, opened 2021-11-05
fix(pipelined): Add grpc error code for timeouts

koolzz:

## Summary

- Set proper return code for timeouts in pipelined rpc handling to handle errors on sessiond

## Test Plan

tested in spirient

## Additional Information


pshelar (approved):

ulaskozat (approved):
LGTM

PR #10158, opened 2021-11-05
feat(agw): add pipelined service to golang version of agwd

zbrock:
## Summary
* Build protos in Go for Pipelined and dependencies
* Add a gRPC server for pipelined
* Add support for gRPC services on tcp4/tcp6 ports

Signed-off-by: Zach Brock <zbrock@gmail.com>

## Test Plan
Unit and integration tests in Go

## Additional Information
None


src/go/.gitignore
```diff
@@ -0,0 +1 @@
+main.go
```
markjen: probably don't want to check this in

src/go/agwd/config/config.go
```diff
@@ -43,8 +43,8 @@ func LogLevel(l mconfig.AgwD_LogLevel) log.Level {
 }
 
 const (
-	ipv4Scheme = "ipv4"
-	ipv6Scheme = "ipv6"
+	ipv4Scheme = "tcp4"
```
markjen: the naming here says the target should be `ipv4` or `ipv6`

https://github.com/grpc/grpc/blob/master/doc/naming.md

so  i guess we should map it to `tcp4` or `tcp6` when we open the listener?

markjen (changes requested):

markjen:
@mstre123 FYI, @zbrock is also setting up a new grpc service and adding a config field.

mstre123:
> @mstre123 FYI, @zbrock is also setting up a new grpc service and adding a config field.

thanks for the heads up - so this is going to conflict with https://github.com/magma/magma/pull/10073
not a big deal but want to make clear that were moving away from mconfigs right now - that said the protos for agwd were copied over so shouldnt be that big a diff.

src/go/protos/magma/session_manager/session_manager.pb.go
```diff
@@ -0,0 +1,10603 @@
...
+	return nil
+}
+
+type isUPFNodeState_UpfNodeMessages interface {
+	isUPFNodeState_UpfNodeMessages()
+}
+
+type UPFNodeState_AssociatonState struct {
+	// Message for establishing association
+	AssociatonState *UPFAssociationState `protobuf:"bytes,2,opt,name=associaton_state,json=associatonState,proto3,oneof"`
```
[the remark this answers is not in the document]
zbrock: Hmm, is there a way to ignore this warning?

zbrock:
> > @mstre123 FYI, @zbrock is also setting up a new grpc service and adding a config field.
> 
> thanks for the heads up - so this is going to conflict with #10073 not a big deal but want to make clear that were moving away from mconfigs right now - that said the protos for agwd were copied over so shouldnt be that big a diff.

Just rebased on top of your change and cleaned up the conflicts, so I think we're in good shape.

src/go/agwd/config/config.go
```diff
@@ -46,22 +46,25 @@ func LogLevel(l config.AgwD_LogLevel) log.Level {
...
 // ParseTarget takes a target in string form and returns a resolved Target.
 // Extends functionality in grpc/internal/grpcutil.ParseTarget to support ipv4
-// and ipv6 schemes.
+// and ipv6 schemes. Return tcp4Scheme or tcp6Scheme so the Scheme can be passed
+// directly to net.Listen.
 func ParseTarget(target string) resolver.Target {
 	if strings.HasPrefix(target, ipv4Scheme+":") {
 		return resolver.Target{
-			Scheme:   ipv4Scheme,
+			Scheme:   tcp4Scheme,
```
mstre123, line 61: i think you need to add handlers for tcp4 and tcp6 schemes.

src/go/README.md
```diff
@@ -31,6 +34,11 @@ To generate BUILD.bazel files, run
 ```sh
 bazel run //:gazelle
 ```
+### Running tests
+To run all the tests, run
```
themarwhal, line 38: Thanks for adding this! :heart_eyes:

src/go/agwd/config/config.go
zbrock: Annoyingly, gRPC and Go's Net library proscribe different formats:
https://github.com/grpc/grpc/blob/master/doc/naming.md
https://pkg.go.dev/net#Dial

gRPC wants "ipv4"
Go wants  "ip4" or "tcp4"

So I stuck with the gRPC convention here and translate when parsing so downstream Go code doesn't have to worry about it.

markjen: I would recommend returning what gRPC wants and then having a separate helper `func NewTargetListner(target resolver.Target) (net.Listner, error)` that maps scheme to tcp{4,6}

markjen (approved):
also golangci caught a file that needs `gofmt`

src/go/protos/magma/session_manager/session_manager.pb.go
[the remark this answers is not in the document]
mstre123: i think nolint might work since golint underpins reviewdog
https://golangci-lint.run/usage/false-positives/

Issue #10166, opened 2021-11-08
GCC warning annotations seen on unrelated PR


themarwhal:
https://github.com/magma/magma/pull/10149/files <- seeing some unrelated GCC warning annotations. Might be good to address these.

In case the annotation goes away, I'll take screenshots of the warnings. 

<img width="1229" alt="Screen Shot 2021-11-08 at 10 50 08 AM" src="https://user-images.githubusercontent.com/37634144/140773634-4bd532c3-7927-4e4e-bc85-d0f07cdf9aec.png">

themarwhal:
Wasn't sure who to assign as an owner, so I'll just ping all the names I get from git for each line above.

* lte/gateway/c/core/oai/tasks/gtpv1-u/gtp_tunnel_upf.c#L36, lte/gateway/c/core/oai/tasks/gtpv1-u/gtp_tunnel_upf.c#L37 @panyogesh 
* lte/gateway/c/core/oai/tasks/nas/emm/Attach.c#L1076 @pruthvihebbani 
* lte/gateway/c/core/oai/tasks/mme_app/mme_app_statistics.c#L26 @ulaskozat 
* lte/gateway/c/core/oai/tasks/ngap/ngap_amf_handlers.c#L798 @ulaskozat (this looks like a cherrypick from something else though)
* lte/gateway/c/core/oai/tasks/ngap/ngap_amf_nas_procedures.c#L1229 @ganeshg87 
* lte/gateway/c/core/oai/tasks/amf/amf_config.c#L394 @ganeshg87

ulaskozat:
@sureshgorijavolu can you follow up on the 5G functions?

ulaskozat:
PR raised for lte/gateway/c/core/oai/tasks/mme_app/mme_app_statistics.c#L26 (cc @themarwhal)

themarwhal:
I'm also seeing GCC warnings on the OAI Jenkins job. https://jenkins-oai.eurecom.fr/job/MAGMA-MME-production/8981/artifact/build_results_magma_oai_mme.html
<img width="1248" alt="Screen Shot 2021-11-16 at 6 00 09 PM" src="https://user-images.githubusercontent.com/37634144/142079947-8dbe1a35-7fc1-49d1-b9d2-94cc036431ce.png">
Can we take care of these as well? @sureshgorijavolu

panyogesh:
[...]

PR #10167, opened 2021-11-08
fix(mme): add return type to the inline function

ulaskozat:

## Summary

Fix the signature of the inline function. See #10166 for reference.

## Additional Information


ardzoht (commented):
lgtm

ardzoht (approved):

PR #10172, opened 2021-11-08
feat: adding singleton reindexer service to docker

### LINKED ISSUE #10135: Make reindexer a singleton service

reginawang3495:
## Summary

Add service with `singleton_reindexer` to be run alongside `reindexer` servicer

## Test Plan
3 different manual tests

### Right Services are Spun Up
1. `./build.py --all`
2. `./run.py`
3. ` docker-compose exec controller bash`
4. `supervisorctl` to see whether services are running as expected

### Flags work properly
After Step 1 + 2 from above,
Using `docker-compose logs -f controller` to check that the right `reindexer_singleton` path is spun up or not spun up in both services as per `enable_singleton_reindex` flag

### Check that reindexing works
After Step 1 + 2 from above,
1. `docker exec -it orc8r_postgres_1 psql -U magma_dev`
2. `update indexer_versions set version_actual=0 where indexer_id='lte';`
3. (wait max a minute)
3. `select * from indexer_versions;` to see that it updated properly
```
  indexer_id  | version_actual | version_desired 
--------------+----------------+-----------------
 orchestrator |              1 |               1
 subscriberdb |              1 |               1
 lte          |              0 |               1
(3 rows)
```
4. `docker-compose logs -f controller`
    a. Found ` Attempt at state reindex job {id: lte, from: 0, to: 1} ...` as expected 
5. see that `select * from indexer_versions;` changed back as expected
```
  indexer_id  | version_actual | version_desired 
--------------+----------------+-----------------
 orchestrator |              1 |               1
 subscriberdb |              1 |               1
 lte          |              1 |               1
(3 rows)
```

## Additional Information


orc8r/cloud/go/services/singleton_state/singleton_state/main.go
```diff
@@ -0,0 +1,116 @@
...
+		store := blobstore.NewSQLStoreFactory(state.DBTableName, db, sqorc.GetSqlBuilder())
+		err = store.InitializeFactory()
+		if err != nil {
+			glog.Fatalf("Error initializing singleton_state database: %v", err)
+		}
+
+		stateServicer := newStateServicer(store)
+		protos.RegisterStateServiceServer(srv.GrpcServer, stateServicer)
+		indexerManagerServer := newSingletonIndexerManagerServicer(srv.Config, db, store)
+		indexer_protos.RegisterIndexerManagerServer(srv.GrpcServer, indexerManagerServer)
```
hcgatewood: This is really the only thing we need here -- the other servicers and goroutines are for the state service

hcgatewood (commented):
Shape looks great!

orc8r/cloud/go/services/singleton_state/singleton_state/main.go
reginawang3495: Are we going to have a state service and a reindexer service separately? As in are we completely replacing the `state/main.go`?

hcgatewood: Correct we will have 2 separate services unconditionally running in parallel

reginawang3495: To clarify my question, at the end of this milestone, are reindexer and state services separate?

hcgatewood: From offline: yes

hcgatewood: But note that we're calling it `orc8r_worker`, rather than `reindexer`, because `orc8r_worker` will be a generic singleton service for running any single-pod workflows -- which reindexer happens to be

reginawang3495: Ah that makes sense

orc8r/cloud/configs/orc8r_worker.yml
```diff
@@ -0,0 +1,21 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file should essentially be a duplicate of orc8r/cloud/configs/state.yml
```
themarwhal: If we update orc8r/cloud/configs/state.yml, do we update this file too? (In that case, it might be good to keep a note of that in the other file)

reginawang3495: I think that this is temporarily the same as the other file. By the end of the milestone, these two flags should be removed from the other file! Maybe I should add a TODO in the other file to make it more clear?

orc8r/cloud/go/services/orc8r_worker/doc.go
```diff
@@ -0,0 +1,18 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package orc8r_worker
+
+const (
+	ServiceName = "orc8r_worker"
```
hcgatewood: Nit: use a const.go for this instead of doc.go (ik we do it both ways)

orc8r/cloud/go/services/orc8r_worker/orc8r_worker/main.go
```diff
@@ -0,0 +1,93 @@
...
+		err = store.InitializeFactory()
+		if err != nil {
+			glog.Fatalf("Error initializing state database: %v", err)
+		}
+
+		indexerManagerServer := newSingletonIndexerManagerServicer(srv.Config, db, store)
+		indexer_protos.RegisterIndexerManagerServer(srv.GrpcServer, indexerManagerServer)
+	}
+
+	go metrics.PeriodicallyReportGatewayStatus(gatewayStatusReportInterval)
```
hcgatewood: Only applicable for state service

orc8r/cloud/go/services/orc8r_worker/orc8r_worker/main.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+		store := blobstore.NewSQLStoreFactory(state.DBTableName, db, sqorc.GetSqlBuilder())
+		err = store.InitializeFactory()
+		if err != nil {
+			glog.Fatalf("Error initializing state database: %v", err)
+		}
+
+		indexerManagerServer := newSingletonIndexerManagerServicer(srv.Config, db, store)
+		indexer_protos.RegisterIndexerManagerServer(srv.GrpcServer, indexerManagerServer)
+	}
```
hcgatewood: Since the `orc8r_worker` is generic (any number of singleton jobs), let's move this block into its own `startReindexer` func

hcgatewood (approved):
Looks great! Couple small comment, msg me when ready to merge 👍

orc8r/cloud/configs/orc8r_worker.yml
themarwhal: if it's a short term thing, then I think we're ok :)

orc8r/cloud/go/services/orc8r_worker/doc.go
reginawang3495: I can change the other ones into const.go -- just wondering, but is there a reason we prefer one over the other (besides for consistency reasons?)

orc8r/cloud/go/services/orc8r_worker/orc8r_worker/main.go
reginawang3495: it's the entire indexer if that's ok?

orc8r/cloud/go/services/orc8r_worker/doc.go
hcgatewood: Reason is doc.go is expected to only contain comments (documentation), is idiosyncratic/confusing when it contains real code

Also yes that would be great, but I already have it marked as a bootcamp task 🙂

orc8r/cloud/go/services/orc8r_worker/orc8r_worker/main.go
hcgatewood: Maybe I'm misreading -- I don't see where it's adding an indexer?

reginawang3495: https://github.com/magma/magma/pull/10172/files#diff-5cd2a79f7253b049992f66579be1a14a4cd5a9360c978afefdb10d92b745baccR81

(L81 if it doesn't highlight properly)

reginawang3495: I think it's adding an indexer? Not sure now

hcgatewood: Ah right. So that's an indexer *manager* servicer, which exposes the grpc servicer used by the (re)index CLI https://github.com/magma/magma/tree/13b485a9b6da5874761919c840e4c0a8be81641d/orc8r/cloud/go/tools/indexers

We do want that attached just to this singleton

reginawang3495: ahh makes sense thanks for the info!

orc8r/cloud/go/services/orc8r_worker/orc8r_worker/main.go
```diff
@@ -0,0 +1,91 @@
...
+	srv, err := service.NewOrchestratorService(orc8r.ModuleName, orc8r_worker.ServiceName)
+	if err != nil {
+		glog.Fatalf("Error creating orc8r_worker service %v", err)
+	}
+
+	// TODO(reginawang3495): rename function name when jobQueue reindexer is removed
+	singletonReindex := srv.Config.MustGetBool(state_config.EnableSingletonReindex)
+	if singletonReindex {
+		startSingletonReindexer(srv)
+	}
```
hcgatewood, line 47: I would even move this config check into the "start reindexer" func. But can do in next PR

hcgatewood (approved):
💯

hcgatewood:
Cloud tests are passing, force-merging to unblock @reginawang3495

PR #10182, opened 2021-11-09
chore: added checkin_cli test to federated integ test

uri200:

## Summary

Enhanced federated integ test to add checkin cli command for connectivity test

## Test Plan

fab test_all

## Additional Information

### DIFF
--- a/lte/gateway/dev_tools.py
+++ b/lte/gateway/dev_tools.py
@@ -16,10 +16,12 @@ import sys
 from typing import Any, List, Optional
 
 import urllib3
+from fabric.api import cd
 
 sys.path.append('../../orc8r')
 import tools.fab.dev_utils as dev_utils
 import tools.fab.types as types
+from tools.fab.hosts import vagrant_setup
 
 LTE_NETWORK_TYPE = 'lte'
 FEG_LTE_NETWORK_TYPE = 'feg_lte'
@@ -163,6 +165,28 @@ def deregister_feg_gw():
     subprocess.check_call(
         'fab deregister_feg_gw', shell=True, cwd=FEG_FAB_PATH,
     )
+
+
+def check_agw_cloud_connectivity(timeout=10):
+    """
+    Check connectivity of AGW with the cloud using checkin_cli.py
+    Args:
+        timeout: amount of time the command will retry
+    """
+    vagrant_setup("magma", destroy_vm=False, force_provision=False)
+    with cd("/home/vagrant/build/python/bin/"):
+        dev_utils.run_fab_command_with_repetition("./checkin_cli.py", timeout)
+
+
+def check_agw_feg_connectivity(timeout=10):
+    """
+    Check connectivity of AGW with FEG feg_hello_cli.py
+    Args:
+        timeout: amount of time the command will retry
+    """
+    vagrant_setup("magma", destroy_vm=False, force_provision=False)
+    with cd("/home/vagrant/build/python/bin/"):
+        dev_utils.run_fab_command_with_repetition("./feg_hello_cli.py m 0", timeout)
 
 
 def _register_network(network_type: str, payload: Any):
--- a/lte/gateway/python/integ_tests/federated_tests/README.md
+++ b/lte/gateway/python/integ_tests/federated_tests/README.md
@@ -6,74 +6,79 @@ hide_title: true
 # Federated Integration Tests (EXPERIMENTAL)
 
 The objective of Federated Integration Tests is to provide a test platform
-where to run Magma with all its components.
+to run Magma with all its components.
 
-As the diagram indicates, this test spins up AGW, FEG and Orc8r
-and uses S1aP and trf server to run tests.
+As the diagram indicates, this test spins up **AGW**, **FeG**, and **Orc8r**
+and uses `S1aP` and `magma_trfserver` as an eNb and SGi emulator to run tests.
 
 ```mermaid
 graph LR;
-    S1AP --> AGW --> orc8r --> FEG --> 3ggp_CORE;
-    3ggp_CORE --> FEG --> orc8r --> AGW --> S1AP;
-    AGW --> trf_server;
-    trf_server --> AGW;
+    A[[S1AP]] --- AGW ---|control| orc8r --- FEG --- HSS[(HSS)];
+    AGW ---|data| Z[[magma_trfserver]];
 ```
 
-The processes will run on the following machines:
-- AGW will run on **magma** vm.
-- FEG and Orc8r will run on our **local docker**.
-- Traffic server will run on **magma_trf** vm
-- Test will be started from **magma_test** vm
+The services will run either on Vagrant VM or on Docker:
+
+| Services          |  Vagrant VM       |  Docker   |
+|-------------------|:-----------------:|:----------:|
+| AGW               | magma             |           |
+| FeG               |                   | &check;   |
+| Orc8r             |                   | &check;   |
+| Traffic server    | magma_trfserver   |           |
+| S1AP tester       | magma_test        |           |
+
+
+## Build environment
+
+Next steps show you how to build the test environment automatically. If you
+want to build this step by step go to [Manual build](#Manual-build)
 
 Note commands for AGW will have to be run inside the vagrant VM. That is
 why all the actions include the `vagrant ssh magma` command first. To leave
-from vagrant just type `exit`. FEG and Orc8r will need to be run in the
+from vagrant just type `exit`. FeG and Orc8r will need to be run in the
 host itself (no vagrant involved)
 
-## Build environment
-### Automatic build
-On your host machine do
+On your host machine execute this command to build, start and conigure AGW,
+FEG and Orc8r and start them.
 ```bash
 cd magma/lte/gateway/python/integ_tests/federated_tests
-fab build_all
+fab build_all_and_configure
 ```
-That will build AGW, FEG and Orc8r and start them all. It will also try to
-register a federated AGW and a FEG gateway. After this is run you can check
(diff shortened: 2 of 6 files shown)


lte/gateway/python/integ_tests/federated_tests/README.md
```diff
@@ -6,87 +6,93 @@ hide_title: true
...
+
+| Services          |  Vagrant VM       |  Docker   |
+|-------------------|:-----------------:|:----------:|
+| AGW               | magma             |           |
+| FeG               |                   | &check;   |
+| Orc8r             |                   | &check;   |
+| Traffic server    | magma_trfserver   |           |
+| S1AP tester       | magma_test        |           |
+
+
```
uri200, line 30: image

lte/gateway/dev_tools.py
```diff
@@ -137,6 +139,28 @@ def deregister_feg_gw():
...
 
+def check_agw_cloud_connectivity(timeout=10):
+    """
+    Check connectivity of AGW with the cloud using checkin_cli.py
+    Args:
+        timeout: amount of time the command will retry
+    """
+    vagrant_setup("magma", False, force_provision=False)
+    with cd("/home/vagrant/build/python/bin/"):
+        dev_utils.run_fab_command_with_repetition("./checkin_cli.py", timeout)
```
hcgatewood, line 178: Nit here and below: can coalesce these into single line, by specifying the executable's full path

lte/gateway/python/integ_tests/federated_tests/README.md
```diff
@@ -6,87 +6,93 @@ hide_title: true
...
+and uses `S1aP` and `magma_trfserver` as an eNb and SGi emulator to run tests.
 
 ```mermaid
 graph LR;
-    S1AP --> AGW --> orc8r --> FEG --> 3ggp_CORE;
-    3ggp_CORE --> FEG --> orc8r --> AGW --> S1AP;
-    AGW --> trf_server;
-    trf_server --> AGW;
+    A[[S1AP]] --- AGW ---|control| orc8r --- FEG --- HSS[(HSS)];
+    AGW ---|data| Z[[magma_trfserver]];
```
hcgatewood, line 17: Wow that's so pretty / looks so easy to write. Do you have a pointer to the doc/tutorial you used to understand mermaid?

lte/gateway/python/integ_tests/federated_tests/README.md
```diff
@@ -6,87 +6,93 @@ hide_title: true
...
+| FeG               |                   | &check;   |
+| Orc8r             |                   | &check;   |
+| Traffic server    | magma_trfserver   |           |
+| S1AP tester       | magma_test        |           |
+
+
+## Build environment
+
+Next steps show you how to build the test environment automatically. If you
+want to build this step by step go to [Manual build](#Manual build)
```
hcgatewood: Does `#Manual build` work correctly? I think it might have to be `#manual_build` but not sure

orc8r/tools/fab/dev_utils.py
```diff
@@ -405,3 +407,43 @@ def cloud_delete(
...
+    """
+    run command on remote machine using fabric.api.run. Repeats on error
+    Args:
+        command: command to run
+        timeout: time to execute the command while it fails
+    """
+    _command_with_repetition(run, command, timeout)
+
+
+def _command_with_repetition(func, command, timeout=5):
```
hcgatewood, line 444: Nit: "timeout" gives the connotation it's time-based. Maybe something like "max retries"?

hcgatewood (approved):
Awesome! Couple small notes but lgtm

lte/gateway/python/integ_tests/federated_tests/README.md
uri200: https://mermaid-js.github.io/mermaid/#/./flowchart?id=flowcharts-basic-syntax

lte/gateway/dev_tools.py
uri200: Since fab it is a command line automation format, it is better to have each step separated so it is more readable and easier to debug.

It also provides more granularity in case of error (if that folder doesn't exist you will get an explicit error)

orc8r/tools/fab/dev_utils.py
uri200: In fact it is a time out.  We are not worried for the number of attempts (which happen every second)

The reason for that is on those integ test normally the system is not ready by the time the test starts. Adding this allow us to have some control on how long will the test retry.  This command does not pretend to be exact (if it takes more tjhan 5 seconds is fine, it will not take less in case the system is down)

In other words, this just mimics what a person would do while waiting for the service to start (similar to running a watch command)

If i am able to add this tests on github actions , I will call that command with 60 to make sure I give the system plenty of time to start and bootstrap

hcgatewood: Hm if I'm reading correctly, it looks like there's no timing functionality specifically in this function -- looks like it increments a count after every attempt, and after a max number of retires gives up

To me, timeout means "try for at most x seconds", while max tries means "try at most this number of times". Not a huge deal, and maybe I'm reading it wrong 🙂

uri200: check line 459

hcgatewood: Right right. I guess the consideration is still: consider if "func" takes e.g. 10 minutes to run -- then the total runtime would be `timeout*10m1s`, rather than `timeout*1s`. Which makes it feel like a max_retries rather than timeout

But again, just a nit, up to you!

themarwhal (approved):
some of the pep8 annotations seem like they are worth addressing but other than that lgtm!

orc8r/tools/fab/dev_utils.py
uri200: Yes I considered that, but since the commands used by the fab command here are normally limited in time I didn't see it necessary. But since the code is in a shared library it may be wise to add an upper bound. 

Will change some things so we can make the function more generic and make sure we don't exceed the time out

koolzz (approved):

lte/gateway/python/integ_tests/federated_tests/README.md
```diff
@@ -6,87 +6,93 @@ hide_title: true
...
-
-Once it is built, start the traffic and test vm
+## Run test
+Once you have built all the VMs, you can try to run a test from
+`magma_test` VM
 ```bash
 cd magma/lte/gateway
-vagrant up magma_test
-vagrant up magma_trfserver
+vagrannt ssh magma_test
```
ardzoht:
```suggestion
vagrant ssh magma_test
```

ardzoht (approved):
minor nit on docs typo

.github/workflows/federated-integ-test.yml
```diff
@@ -0,0 +1,46 @@
+---
+
+name: Federated integ test
+
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - master
+      - v1.*
+  pull_request:
```
quentinDERORY: we don't want heavy integration tests to be triggered on pull_request.
You should use the workflow_dispatch trigger to execute on demand on PR.

quentinDERORY (changes requested):
LGTM but need to change trigger on GHA job

.github/workflows/federated-integ-test.yml
uri200: yep, this is a test right now. I will probably remove the whole file once i get this running (if i get it)

Will create a separate pr for that

.github/workflows/federated-integ-test.yml
```diff
@@ -0,0 +1,74 @@
...
+        id: date
+      - uses: actions/checkout@v2
+      - name: setup pyenv
+        uses: "gabrielfalcao/pyenv-action@v8"
+        with:
+          default: 3.8.5
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.8.5'
+      - name: Install Docker
```
quentinDERORY: Use this module to install docker and docker-compose : https://github.com/marketplace/actions/setup-docker

.github/workflows/federated-integ-test.yml
```diff
@@ -0,0 +1,114 @@
...
+          docker save mariadb:latest  | gzip > mariadb.tar.gz
+          docker save feg_gateway_go:latest  | gzip > feg_gateway_go.tar.gz
+          docker save feg_gateway_python:latest  | gzip > feg_gateway_python.tar.gz
+      - uses: actions/upload-artifact@v2
+        with:
+          name: federated-integ-test-docker-images
+          path: images
+
+  federated-integ-test:
+    runs-on: macos-10.15
```
quentinDERORY: add 
`needs: docker-build-federated-integ-test`

.github/workflows/federated-integ-test.yml
uri200: Hey @quentinDERORY  I removed the github action file form this PR

lte/gateway/python/integ_tests/federated_tests/docker/docker-compose.override.yml
```diff
@@ -1,27 +1,6 @@
...
-      - gwcerts:/var/opt/magma/certs
-      - gwconfigs:/var/opt/magma/configs
-    command:
-      - /bin/bash
-      - -c
-      - |
-         mkdir -p ../../.cache/feg/
-         touch ../../.cache/feg/snowflake
-         tail -f /dev/null
-
```
uri200, line 24: test service is not needed for federated integ test

quentinDERORY (approved):

PR #10186, opened 2021-11-09
fix(mme): Bug fix in nw initiated bearer deactivation

ulaskozat:

## Summary

- Iteration needs to be stopped as soon as the LBI match is found for given IMSI. This was leading to fetching the wrong TEID and detachment of wrong IMSI.
- Bearer list to be deleted was not passed correctly.

## Test Plan

Spirent and integ tests.

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/grpc_service/SpgwServiceImpl.cpp
+++ b/lte/gateway/c/core/oai/tasks/grpc_service/SpgwServiceImpl.cpp
@@ -206,7 +206,7 @@ Status SpgwServiceImpl::DeleteBearer(
   itti_msg.no_of_bearers = 1;
   for (int i = 0; i < request->eps_bearer_ids_size() && i < BEARERS_PER_UE;
        i++) {
-    itti_msg.ebi[0] = request->eps_bearer_ids(i);
+    itti_msg.ebi[i] = request->eps_bearer_ids(i);
     send_deactivate_bearer_request_itti(&itti_msg);
   }
   return Status::OK;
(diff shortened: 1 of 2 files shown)


ardzoht (approved):
lgtm, we should target creating tests for these iterator functions, will create an issue

lte/gateway/c/core/oai/tasks/grpc_service/SpgwServiceImpl.cpp
```diff
@@ -206,7 +206,7 @@ Status SpgwServiceImpl::DeleteBearer(
   itti_msg.no_of_bearers = 1;
   for (int i = 0; i < request->eps_bearer_ids_size() && i < BEARERS_PER_UE;
        i++) {
-    itti_msg.ebi[0] = request->eps_bearer_ids(i);
+    itti_msg.ebi[i] = request->eps_bearer_ids(i);
```
ardzoht, line 209: @ulaskozat what is the fix here? on `spgw_handle_nw_initiated_bearer_deactv_req` the iteration over `itti_msg.ebi[]` will be bounded to `no_of_bearers` which will always be 1

ulaskozat: this does not fix the bug. but the size does not have to be 1 in general.

ulaskozat: we should change `itti_msg.no_of_bearers = 1;` as well.

PR #9900, opened 2021-10-26
Work in Progress: Bazel Python Build for mobilityd

electronjoe:
This will contain WIP for Bazel Python build. See the related GH Issue linked in status updates below.

Signed-off-by: GitHub <noreply@github.com>

Issue #9881, opened 2021-10-26
bazel: Bazelify Python mobilityd


electronjoe:

Neudrino:
With merge of the [POC for mobilityd](https://github.com/magma/magma/pull/10341) we have a big part of it done.
More additions for test execution support are made in https://github.com/magma/magma/pull/11216.

Neudrino:
:heavy_check_mark: Done, as #10341 and #11216 are merged.

PR #10188, opened 2021-11-09
bazel(python): Prototype for Bazel with Python

Neudrino:
## Summary

This PR is very much work in progress. It is intended as a basis for discussion regarding the **Python Bazel** setup.  It extends https://github.com/magma/magma/pull/9900 by @electronjoe with the aim to have the build and tests running.

Essetially the following changes were made:
* Some additions were made toward dependencies to have the build working.
* Unit test execution for mobilityd was added.

## Test Plan

Setup in Magma VM:
```
~/magma/lte/gateway$ make run
~/magma$ magtivate
~/magma$ sudo service magma@mobilityd stop
```

Executable without "packaging":
Build: `~/magma$ bazel build //lte/gateway/python/magma/mobilityd:main`
Run: `~/magma$ ./bazel-bin/lte/gateway/python/magma/mobilityd/main`

PAR executable:
Build: `~/magma$ bazel build //lte/gateway/python/magma/mobilityd:main_par.par`
Run: `~/magma$ ./bazel-bin/lte/gateway/python/magma/mobilityd/main_par.par`

Run all defined tests:
```
bazel test --test_tag_filters='unittest,mobilityd' //...
```

## Additional Information

* Relates to https://github.com/magma/magma/issues/9881.
* This POC requires a magma VM environment to run (see setup above). That setup is required for external dependency resolution, as they have been out of scope for the first prototype.
* Another idea which was discussed were separate bazel workspaces (/[repositories](https://docs.bazel.build/versions/4.2.1/build-ref.html)) to resolve possible issue with import path. Some findings were:
  * Bazel does not support [transitive dependencies](https://docs.bazel.build/versions/main/external.html#transitive-dependencies). I.e. "Dependecy chaining" or "transparent dependency resolution" from parent / other workspaces is not possible.
  Possible implications are
    * a lot code duplication for dependency resolution,
    * an elaborate dependency setup for any additional workspace, with the implication of a large effort,
    * might opose goal of code structural cleanup.
  * Personal impression currently: Multiple workspaces seem more effort than possible import path rewrite.
  * ❌ Opinion: Seems more effort than import path changes.
* 💡 Possible workaround instead of import path rewrite: `PYTHON_PATH` adoptions to allow for current import path.
(Currently probably included in makefiles).


.github/workflows/docker-push.yml
```diff
@@ -1,6 +1,12 @@
 ---
 name: "Dockerfile Push"
 on:  # yamllint disable-line rule:truthy
+  pull_request:
+    paths:
+      - .devcontainer/Dockerfile
+      - .github/workflows/docker-push.yml
+      - lte/gateway/docker/mme/Dockerfile.ubuntu20.04
+      - lte/gateway/docker/python-precommit/Dockerfile
```
Neudrino, line 9: This part was only inserted to trigger a new devcontainer image build, however, [it seems that only master branch builds are allowed to push new images to GHCR](https://github.com/magma/magma/runs/4153376728?check_suite_focus=true#step:6:23662).

.devcontainer/Dockerfile
```diff
@@ -69,7 +69,8 @@ RUN echo "Install general purpose packages" && \
         ninja-build \
         perl \
         pkg-config \
-        python2.7 \
+        python3 \
+        python-is-python3 \
```
Neudrino, line 73: There is a [dependency on Python 3](https://github.com/magma/magma/blob/master/orc8r/gateway/python/python.mk#L24) in the code. And using a focal fossa (=20.04) we here, an upgrade might be necessary.

It might be the case that this change is needed anyway and could be extracted into a small PR which could than actually merged. 🤔

.github/workflows/docker-push.yml
themarwhal: Yes you'll have to push the devcontainer changes and then update the hash.

themarwhal:
Can you also update the Dockerfile @ https://github.com/magma/magma/tree/master/experimental/bazel-base ?

I like to keep that dockerfile updated as it ONLY contains dependencies needed for Bazel builds.

This should pass
```
# needs docker
cd magma/experimental/bazel-base
docker-compose build magma-builder
docker-compose run magma-builder  bash -c  "bazel test ..."
```

Neudrino:
> Can you also update the Dockerfile @ https://github.com/magma/magma/tree/master/experimental/bazel-base ?

Python 3 will be added to the named images within https://github.com/magma/magma/pull/10245.
Afterwards we can exclude Docker image changes here.

Neudrino:
This PR was for start of an discussion on the topic and is now superseded by
* https://github.com/magma/magma/pull/10245
* https://github.com/magma/magma/pull/10341

Hence closing the PR.

PR #5661, opened 2021-03-22
[orc8r][docker]: Quick-start script: run.py error with '--metrics' parameter

pedroamfarias-cpqd:
## Summary
Ticket issue: https://github.com/magma/magma/issues/5600 

Looks like the run.py script is creating a comment "# DO NOT COMMIT THIS CHANGE" at same line of parameter "COMPOSE_FILE="
* file: magma/python/magma/orc8r/cloud/docker/.env
```
COMPOSE_FILE=docker-compose.yml:docker-compose.metrics.yml:docker-compose.override.yml  # DO NOT COMMIT THIS CHANGE

```

changing the run.py to add a "\n" in "'COMPOSE_FILE={}  {}\n'.format in this comment, the script run with success with --metrics
*file: magma/python/magma/orc8r/cloud/docker/run.py
```
        compose_line = 'COMPOSE_FILE={}  \n{}\n'.format(
            ':'.join(file_args),
            DO_NOT_COMMIT,
        )

```
and after that the .env file keep as it:
* file: magma/python/magma/orc8r/cloud/docker/.env
```
COMPOSE_FILE=docker-compose.yml:docker-compose.metrics.yml:docker-compose.override.yml 
# DO NOT COMMIT THIS CHANGE

```
After this change, the "run.py --metrics"  run with success

## Additional Information

- [ ] This change is backwards-breaking

PR #10203, opened 2021-11-09
fix(orc8r): Fix the file run.py to able to execute run.py -- metrics

talesmota:
Signed-off-by: Tales Mota Machado <tales@radtonics.com>

## Summary

There's a change was made on v1.4 that fixed this problem.
I reapply this change to make permanet.

Fix one issue:

- `run.py --metrics` throws error `ERROR: .FileNotFoundError: [Errno 2] No such file or directory: './docker-compose.override.yml  # DO NOT COMMIT THIS CHANGE'`

As related on issue #5661. 

## Test Plan
Test locally

### DIFF
--- a/orc8r/cloud/docker/run.py
+++ b/orc8r/cloud/docker/run.py
@@ -41,7 +41,7 @@ def main() -> None:
         if args.thanos:
             f.append('thanos')
         file_args = _make_file_args(f)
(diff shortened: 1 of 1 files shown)


hcgatewood (approved):
Thanks for the fix! 💯🙌

PR #10204, opened 2021-11-09
feat: build CWAG sessiond with Bazel

themarwhal:

## Summary
Blocked on https://github.com/magma/magma/pull/10394

Modified and cleaned up CWAG dockerfile to use bazel to build sessiond. I've also removed unused dependencies to clean up the image. Also add a disable sentry build option as I'm unable to build breakpad on the bionic base.

Some numbers on the difference
* fresh build time increases by 13% (goes from 6min 42 seconds on master to 7 min 38 seconds with this PR - taken on same machine) 
* build with only sessiond changes are about the same
* image size decreases by 58%(945MB on master to 393MB with this PR)

Possible improvements:
* Use docker_rules. My only concern here is that we have a lot of external dependencies needed for C/C++ that will require a lot of work to build with bazel. But maybe we can create out own custom base image with all dependencies and pull that down.

## Test Plan
Ran CWF integ test locally + CI passes

## Additional Information

### DIFF
--- a/cwf/gateway/docker/c/Dockerfile
+++ b/cwf/gateway/docker/c/Dockerfile
@@ -15,65 +15,43 @@
 # Builder image for C binaries and Magma proto files
 # -----------------------------------------------------------------------------
 
-# debian stretch image
-# ARG OS_DIST=debian
-# ARG OS_RELEASE=stretch
-# ARG EXTRA_REPO=https://facebookconnectivity.jfrog.io/artifactory/list/dev/
-# ARG CLANG_VERSION=3.8
-
 # ubuntu bionic image
-ARG OS_DIST=ubuntu
-ARG OS_RELEASE=bionic
-ARG EXTRA_REPO=https://facebookconnectivity.jfrog.io/artifactory/cwf-prod
-ARG CLANG_VERSION=3.9
+ARG OS_DIST=ubuntu OS_RELEASE=bionic EXTRA_REPO=https://facebookconnectivity.jfrog.io/artifactory/cwf-prod
 
 # Stretch is required for c build
 FROM $OS_DIST:$OS_RELEASE AS builder
-ARG OS_DIST
-ARG OS_RELEASE
-ARG EXTRA_REPO
-ARG CLANG_VERSION
+ARG OS_DIST OS_RELEASE EXTRA_REPO
+
+RUN apt-get update && \
+    # Setup necessary tools for adding the Magma repository
+    apt-get install -y apt-utils software-properties-common apt-transport-https gnupg wget && \
+    # Download Bazel
+    wget -P /usr/sbin --progress=dot:giga https://github.com/bazelbuild/bazelisk/releases/download/v1.10.0/bazelisk-linux-amd64 && \
+    chmod +x /usr/sbin/bazelisk-linux-amd64 && \
+    ln -s /usr/sbin/bazelisk-linux-amd64 /usr/sbin/bazel
 
 # Add the magma apt repo
-RUN apt-get update && \
-    apt-get install -y apt-utils software-properties-common apt-transport-https gnupg
 COPY orc8r/tools/ansible/roles/pkgrepo/files/jfrog.pub /tmp/jfrog.pub
-
 RUN apt-key add /tmp/jfrog.pub && \
     apt-add-repository "deb ${EXTRA_REPO} ${OS_RELEASE} main"
-
 
 # Install dependencies required for building
 RUN apt-get -y update && apt-get -y install \
   sudo \
   curl \
-  wget \
   unzip \
   cmake \
   git \
+  gcc \
+  g++ \
   build-essential \
-  autoconf \
-  libtool \
-  pkg-config \
-  libgflags-dev \
-  libgtest-dev \
-  clang-${CLANG_VERSION} \
-  libc++-dev \
-  protobuf-compiler \
-  grpc-dev \
-  ninja-build \
-  autogen \
-  ccache \
-  libprotoc-dev \
-  libxml2-dev \
-  libxslt-dev \
-  libyaml-cpp-dev \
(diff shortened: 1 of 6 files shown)


cwf/gateway/docker/c/Dockerfile
```diff
@@ -15,65 +15,48 @@
...
 
-# debian stretch image
-# ARG OS_DIST=debian
-# ARG OS_RELEASE=stretch
-# ARG EXTRA_REPO=https://facebookconnectivity.jfrog.io/artifactory/list/dev/
-# ARG CLANG_VERSION=3.8
-
 # ubuntu bionic image
 ARG OS_DIST=ubuntu
 ARG OS_RELEASE=bionic
```
arunuke: Should we standardize on focal (20.04) over bionic (18.04)? The containerized AGW uses focal.

cwf/gateway/docker/c/Dockerfile
```diff
@@ -15,65 +15,48 @@
...
     apt-add-repository "deb ${EXTRA_REPO} ${OS_RELEASE} main"
 
+# Install bazel
+WORKDIR /usr/sbin
+RUN wget --progress=dot:giga https://github.com/bazelbuild/bazelisk/releases/download/v1.10.0/bazelisk-linux-amd64 && \
+    chmod +x bazelisk-linux-amd64 && \
+    ln -s /usr/sbin/bazelisk-linux-amd64 /usr/sbin/bazel
 
 # Install dependencies required for building
 RUN apt-get -y update && apt-get -y install \
```
arunuke, line 39: If you are looking to optimize build time (the 1 minute difference seems okay to me, but if you are looking for areas to improvise), eliminating the additional apt-get update commands could be something to consider.

cwf/gateway/docker/c/Dockerfile
```diff
@@ -82,78 +65,66 @@ RUN apt-get -y update && apt-get -y install \
...
-ENV C_BUILD /build/c
-ENV OAI_BUILD $C_BUILD/oai
-
-ENV CCACHE_DIR $MAGMA_ROOT/.cache/gateway/ccache
-ENV MAGMA_DEV_MODE 1
-ENV XDG_CACHE_HOME $MAGMA_ROOT/.cache
+ENV XDG_CACHE_HOME ${MAGMA_ROOT}/.cache
+
+# Copy Bazel files at root and third_party
+COPY WORKSPACE.bazel ${MAGMA_ROOT}/
```
arunuke: COPY typically adds a single layer for each statement. It appears all these files in this block are from the same local source going to the same destination. If you are not keen on individual layers for each change, you could consider combining them to a single COPY statement that reduces the layers and potentially the build time as well.

cwf/gateway/docker/c/Dockerfile
arunuke: On second thoughts, considering docker uses a cache, future builds *may not* perform an update if a previous unchanged layer already had it. Let's leave the updates intact for every install, but see if they can be combined by any means.

cwf/gateway/docker/c/Dockerfile
```diff
@@ -82,78 +65,66 @@ RUN apt-get -y update && apt-get -y install \
...
+COPY --from=builder /usr/lib/x86_64-linux-gnu/libcrypto* /usr/lib/x86_64-linux-gnu/
+COPY --from=builder /usr/lib/x86_64-linux-gnu/libunwind* /usr/lib/x86_64-linux-gnu/
+
+COPY --from=builder /usr/local/lib/libfolly.so /usr/local/lib/
+
+RUN ldconfig 2> /dev/null
 
 # Copy the build artifacts.
-COPY --from=builder /build/c/session_manager/sessiond /usr/local/bin/sessiond
+COPY --from=builder /magma/bazel-bin/lte/gateway/c/session_manager/sessiond /usr/local/bin/sessiond
```
arunuke, line 116: Couldn't spot an apt-get clean (or similar) command at the end. It helps clean up any packages that are left in the local cache and are not needed anymore. It probably wouldn't help with build times, but could help in reducing image size.

arunuke (commented):
A general comment is to combine/minimize commands that create layers (COPY and RUN in this dockerfile). If there is a way to combine them without breaking any dependencies (and you don't need those specific layers), that might be a good option to consider.

cwf/gateway/docker/c/Dockerfile
themarwhal: Yeah I was trying to get a bit more context on why we chose bionic for this. I would prefer to use a single base so that we're consistent too.

cwf/gateway/docker/c/Dockerfile
themarwhal: Ah didn't realize you can copy multiple files in one step!

cwf/gateway/docker/c/Dockerfile
themarwhal: https://www.internalfb.com/diff/D21813984 for the original diff. Not sure if this is still relevant though

themarwhal: cc @arunuke ^

themarwhal: I will test some change to see if migrating to focal breaks anything in another PR.

uri200 (approved):
so using this new image now we can build with both cmake and bazel ?

themarwhal:
> so using this new image now we can build with both cmake and bazel ?

This is just the SessionD image used in CWAG so we won't be using this image to build anything. We build SessionD with Bazel (previously with cmake) inside the docker build step so that the container has an executable inside.

themarwhal:
Converting to draft to rebase on top of https://github.com/magma/magma/pull/10394

ardzoht (approved):
`orc8r/gateway/c` changes lgtm

PR #10208, opened 2021-11-09
refactor(pipelined): arp control simplification.

pshelar:
Use single pipeline for NAT and Non-NAT ARP flows.

## Test Plan
`make test` and S1ap integ test.

## Additional Information


koolzz (approved):
lgtm, why did we do it by ip block initially?

pshelar:
> lgtm, why did we do it by ip block initially?

not sure, I introduced the Non-NAT arp flow since agw does not know IPBLOCK in some case ahead of time. I think it is generic enough to handle both NAT and Non-NAT.

PR #10211, opened 2021-11-09
fix(sessiond): fix race condition on handle_activate_ue_flows_callback

uri200:

## Summary
This PR fixes a possible race condition that happens when we try to read redis content from inside pipelined client thread.

This PR just remove that readd_all_session from pipelined client and passes session_id as a parameter.

## Test Plan

make precommit_sm

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -529,18 +529,19 @@ void LocalEnforcer::execute_actions(
 // TODO look into whether we need to re-install all Gx rules on activation
 void LocalEnforcer::handle_activate_service_action(
     const std::unique_ptr<ServiceAction>& action_p) {
-  const std::string imsi      = action_p->get_imsi(),
-                    msisdn    = action_p->get_msisdn(),
-                    ip_addr   = action_p->get_ip_addr(),
-                    ipv6_addr = action_p->get_ipv6_addr();
-  const Teids teids           = action_p->get_teids();
-  const auto ambr             = action_p->get_ambr();
-  RulesToProcess to_process   = action_p->get_gx_rules_to_install();
+  const std::string imsi       = action_p->get_imsi(),
+                    msisdn     = action_p->get_msisdn(),
+                    ip_addr    = action_p->get_ip_addr(),
+                    ipv6_addr  = action_p->get_ipv6_addr(),
+                    session_id = action_p->get_session_id();
+  const Teids teids            = action_p->get_teids();
+  const auto ambr              = action_p->get_ambr();
+  RulesToProcess to_process    = action_p->get_gx_rules_to_install();
   pipelined_client_->activate_flows_for_rules(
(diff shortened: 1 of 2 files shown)


lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1944,29 +1953,18 @@ void LocalEnforcer::schedule_revalidation(
...
     return;
   }
 
-  MLOG(MERROR) << "Could not activate rules for " << imsi
+  MLOG(MERROR) << "Could not activate rules for " << session_id
                << ", rpc failed: " << status.error_message()
                << ", terminating session...";
 
-  SessionSearchCriteria criteria(imsi, IMSI_AND_UE_IPV4_OR_IPV6, ip_addr);
-  auto session_map = session_store_.read_sessions({imsi});
```
uri200, line 1960: race condition may happen here

themarwhal (approved):

ulaskozat (approved):
LGTM

lte/gateway/c/session_manager/LocalEnforcer.cpp
themarwhal: ah is this outside of the eventloop?

uri200:
backporting it to 1.6 here https://github.com/magma/magma/pull/10213

ulaskozat (changes requested):
seeing issues on spirent tests on 1.6.1. Blocking until cleared.

lte/gateway/c/session_manager/LocalEnforcer.cpp
uri200: I think so becasuse this is called by pieplined_client. Which i think it runs on a different thread (at least sessiond was showing different thread id on logs)

here is where it calls 

https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=lte%2Fgateway%2Fc%2Fsession_manager%2FLocalEnforcer.cpp&editor=JetBrains&version=v1.2.1&utm_product_name=CLion&utm_product_version=2021.1.2&start_row=367&start_col=10&end_row=367&end_col=10

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1671,7 +1676,8 @@ void LocalEnforcer::handle_set_session_rules(
     RuleSetBySubscriber rule_set_by_sub(rules_per_sub);
 
     for (const auto& session : session_it->second) {
-      auto& uc           = session_update[imsi][session->get_session_id()];
+      auto session_id    = session->get_session_id();
+      auto& uc           = session_update[imsi][session_id];
```
uri200, line 1680: fix this  on a second commit on this PR too

[second commit](https://github.com/magma/magma/pull/10211/commits/48ac91cc609a9cf8553ad0e0dddbe046c1f95193)

ulaskozat (approved):
LGTM

ulaskozat:
@uri200 looks like failing clang format?

PR #10212, opened 2021-11-09
feat(orc8r): increase workgroups size if blue-green-orc8r is enabled

joary:
Signed-off-by: Joary Paulo Wanzeler Fortuna <joarypl@fb.com>

## Summary

In terraform the [eks.worker_group](https://github.com/terraform-aws-modules/terraform-aws-eks#input_worker_groups) defines the size and amount of EC2 machines to be deployed in a given k8s cluster, orc8r by default uses 3 x t3.larget.
When deploying a blue-green orc8r setup a different set of machines is needed with more machines of smaller size 5 x t3.medium.

In order to maintain backwards compatibility it is proposed to have a flag indicating the a blue-green deployment is needed. Activation of this flag forces the usage of a different terraform variable pre-configured for blue-green deployment.

1. Include two new variables in orc8r-aws terraform configurations:
    1. Flag to enabled the blue-green deployment
    2. WorkerGroup configuration defaults for a blue-green deployment
1. Configure the EKS module to use blue-green WG values if `enable_orc8r_blue_green_deployment`is active

## Test Plan

With main.tf:

```
module "orc8r" {
   ...
   enable_orc8r_blue_green_deployment = true 
   ...
}
```

terraform plan shows:
```
  # module.orc8r.module.eks.aws_autoscaling_group.workers[0] will be created
  + resource "aws_autoscaling_group" "workers" {
      + arn                       = (known after apply)
      + availability_zones        = (known after apply)
      + default_cooldown          = (known after apply)
      + desired_capacity          = 5
      + force_delete              = false
      + force_delete_warm_pool    = false
      + health_check_grace_period = 300
      + health_check_type         = (known after apply)
      + id                        = (known after apply)
      + launch_configuration      = (known after apply)
      + max_instance_lifetime     = 0
      + max_size                  = 5
      + metrics_granularity       = "1Minute"
      + min_size                  = 1
      + name                      = (known after apply)
      + name_prefix               = "orc8r-wg-1"
      + protect_from_scale_in     = false
      + service_linked_role_arn   = (known after apply)
      + suspended_processes       = [
          + "AZRebalance",
        ]
      + termination_policies      = []
      + vpc_zone_identifier       = [
          + "subnet-0a59aa8e9d94571cd",
          + "subnet-0b488fb0d17e3b66f",
          + "subnet-0ea68e1c0df662ec2",
        ]
      + wait_for_capacity_timeout = "10m"

      + tag {
          + key                 = "Name"
          + propagate_at_launch = true
[...]

PR #10215, opened 2021-11-10
feat(orc8r): Include a configuration to deploy an additional namespce for staging orc8r

joary:
## Summary

1. Include a new variable to define the orc8r staging namespace (default empty)
2. Include a new resource creating the namespace when the variable is not empty

## Test Plan

If main.tf includes:

```
module "orc8r-app" {
  ...
  orc8r_staging_namespace = "orc8r-staging"
  ...
}
```

Terraform plan shows:

```
  # module.orc8r-app.kubernetes_namespace.staging-orc8r[0] will be created
  + resource "kubernetes_namespace" "staging-orc8r" {
      + id = (known after apply)

      + metadata {
          + generation       = (known after apply)
          + name             = "orc8r-staging"
          + resource_version = (known after apply)
          + self_link        = (known after apply)
          + uid              = (known after apply)
        }
    }
```

After terraform apply the namespace can be seen:

```
# kubectl get namespace
NAME              STATUS   AGE
default           Active   6d
kube-node-lease   Active   6d
kube-public       Active   6d
kube-system       Active   6d
monitoring        Active   44m
orc8r             Active   44m
orc8r-staging     Active   6s
```

**Attention**

OBS1: In case the namespace variable is removed after apply. Terraform will delete the existing namespace and all the content deploied on it will be lost (Need to be very carefull with this.)

OBS1: In case the namespace variable is changed after apply. Terraform will delete the existing namespace and create a new one with the updated name. All the content in the previous namespace will be lost.

## Additional Information

### DIFF
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
@@ -75,6 +75,12 @@ variable "monitoring_kubernetes_namespace" {
   default     = "monitoring"
 }
 
+variable "orc8r_staging_namespace" {
+  description = "K8s namespace to install staging Orchestrator components into."
+  type        = string
+  default     = ""
+}
+
 ##############################################################################
 # General Orchestrator configuration
 ##############################################################################
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/k8s.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/k8s.tf
@@ -17,6 +17,13 @@ resource "kubernetes_namespace" "orc8r" {
   }
 }
 
+resource "kubernetes_namespace" "staging-orc8r" {
+  count = var.orc8r_staging_namespace != "" ? 1 : 0
+  metadata {
+    name = var.orc8r_staging_namespace
+  }
+}
+
 resource "kubernetes_namespace" "monitoring" {
   metadata {
     name = var.monitoring_kubernetes_namespace


hcgatewood (commented):
Is this forcing us to set the `orc8r_staging_namespace` now? If yes, let's instead set some intelligent default (relative to the `orc8r` namespace)

arunuke:
> Is this forcing us to set the `orc8r_staging_namespace` now? If yes, let's instead set some intelligent default (relative to the `orc8r` namespace)

I think the current PR sets the default to an empty string and creates a secondary namespace only when a namespace is specified. Although this would work, my request for @joary would be to consider a global blue-green variable to be used as well. Creating an extra namespace doesn't cost anything, but it would be good to create it only if the customer is using blue-green upgrades (defaults to false until we GA and will then default to true).

orc8r/cloud/deploy/terraform/orc8r-helm-aws/variables.tf
```diff
@@ -75,6 +75,12 @@ variable "monitoring_kubernetes_namespace" {
   default     = "monitoring"
 }
 
+variable "orc8r_staging_namespace" {
```
arunuke, line 78: Should we consider adding a enable_blue_green variable and use that as a condition to create the namespace with the default value set to a non-empty string of our preference? This way, the secondary namespace will be created if customer using blue-green and will have a default value already. 

Might be a better alternative to consider over the check for empty vs valid string.

arunuke (changes requested):

arunuke:
I see the global variable for blue-green enabled in [another PR](https://github.com/magma/magma/pull/10212/files). Perhaps we should merge that first and use that variable here as a check for the namespace with a default value specified. Perhaps orc8r-staging?

joary:
I'm closing this PR since the namespace will be configured in a different way when deploying blue-green

PR #10219, opened 2021-11-10
fix(agw): Update default GW read timeout value on mobilityd

ardzoht:

## Summary

- Updates timeout value of read default GW to 5
- Increases frequency of reads for faster mobilityd startup

## Test Plan

- make integ_test
- make test

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/mobilityd/uplink_gw.py
+++ b/lte/gateway/python/magma/mobilityd/uplink_gw.py
@@ -57,7 +57,7 @@ class UplinkGatewayInfo:
         self._backing_map = gw_info_map
         self._read_default_gw_timer = None
         self._read_default_gw_timer6 = None
-        self._read_default_gw_interval_seconds = 20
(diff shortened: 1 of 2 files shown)


ulaskozat (approved):
LGTM

lte/gateway/python/magma/mobilityd/uplink_gw.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, gw_info_map: MutableMapping[str, GWInfo]):
         self._backing_map = gw_info_map
         self._read_default_gw_timer = None
         self._read_default_gw_timer6 = None
-        self._read_default_gw_interval_seconds = 20
+        self._read_default_gw_interval_seconds = 5
```
pshelar, line 61: In magma vagrant VM, the IPv4 and v6 gateway is configured, so this timer is never used, why does this frequency matter ?

PR #10226, opened 2021-11-10

spikey979:
…flow_parser.go

fix(feg): incorrect conversion between integer types

## Summary

- magma/feg/gateway/policydb/flow_parser.go -> parseProto() 
- magma/feg/gateway/policydb/flow_parser.go -> parseProto() and parseAddress()

String was parsed into an int using strconv.Atoi which can cause incorrect conversion  if int is converted into another integer type of a smaller size. Fixed to strconv.ParseInt function.

## Test Plan

- magma/feg/gateway/policydb/flow_parser_test.go
In tests TestFlowProto() and TestFlowAddresses() added tests for out of range int32 number 

Screenshot 2021-11-10 at 15 01 06

## Additional Information


nikolakn (approved):

uri200:
Thanks @spikey979  for this change and adding test cases

Can you go to `/magma/feg/gateway/docker` and run `./build -c`. That will trigger the formatting, linter, test etec. If there are any changes after that related to your code, just add them to the PR please


If that would not work you can do this manually 
```
docker-compose build test
docker-compose up -d 
docker-compose exec test bash
make precommit
```

lucasgonze:
> Can you go to `/magma/feg/gateway/docker` and run `./build -c`. That will trigger the formatting, linter, test etec.
> ```
> docker-compose build test
> ```

@uri200 What is the convention on trusting tests? If we want to do something potentially risky, should we trust the tests if they pass?

spikey979:
@uri200 do I have to run **./magma/feg/gateway/docker/build.py -c** every time I change the tests?

uri200:
> @uri200 do I have to run **./magma/feg/gateway/docker/build.py -c** every time I change the tests?

Hey @spikey979, that command should be run after any change of the code. That runs all the tests and also apply format so your code has proper indentation and so on. 

We call them precommit. And as the name indicates, they should be run before you commit to make sure Ci doesn't complain

Note the first tine you run it it may take longer because it has to build the image . On following runs it will be faster.

uri200:
> > Can you go to `/magma/feg/gateway/docker` and run `./build -c`. That will trigger the formatting, linter, test etec.
> > ```
> > docker-compose build test
> > ```
> 
> @uri200 What is the convention on trusting tests? If we want to do something potentially risky, should we trust the tests if they pass?

Hey @lucasgonze, in my case, when I change something I always check if there is a test case for that. Otherwise I add it for peace of mind. So what @spikey979 did in that case is proper. He bit just fixed the issue but also showed us it works as expected.

Running the rest of test may reduce the risk of breaking anything else.

For more complex stuff that implies protocol or end to end procedures, we can use cwf_integ_test which runs some test with a  mock PCRF and OCS. But that is not necessary for changes like this

electronjoe:
@spikey979  Looks like a rebase maybe pulled in unintended changes from others (perhaps the merge.. above) ? I see modifications that look to be well outside the intent of this PR description? E.g. Mconfig go changes that may be related to @mstre123 PR?

spikey979:
Yes, something went wrong in this PR, I will delete PR and make a new one

mstre123:
> @spikey979 Looks like a rebase maybe pulled in unintended changes from others (perhaps the merge.. above) ? I see modifications that look to be well outside the intent of this PR description? E.g. Mconfig go changes that may be related to @mstre123 PR?

yeah those were from https://github.com/magma/magma/pull/10073

Issue #8765, opened 2021-08-25
Make autolabel gha workflow  read from codeowner file


quentinDERORY:
Pull the component association with the paths from the CODEOWNER file directly.
The components are right now:
ci
feg
agw
orc8r
cwf
openwrt
showtech
xwf

PR #10227, opened 2021-11-10
fix(ci): Refactor PR label job

### LINKED ISSUE #9512: Save labels not assigned by autolabel PR before resetting them
In the Autolabel PR job, we are wiping every labels on every run.
This wipes ou the labels that people put and the sizing label too.

To address that, we should save all non component  labels, find the new component labels and then reassign the new component labels and the ones that were not assign by this job.

ranj8work:
Signed-off-by: Ranjini Iyer <ranj8work@gmail.com>

## Summary
Fixes for issue [9512](https://github.com/magma/magma/issues/9512) for autolabel PRs

## Test Plan
Testing is done using a test [PR](https://github.com/ranj8work/magma/pull/1) in my forked repo. The github action [log](https://github.com/ranj8work/magma/runs/4226354965?check_suite_focus=true) shows that the component labels getting applied correctly. It also shows that the non-component labels (ex: bigs, documentation) are retained as well with the component labels.

Tested update call working when the PR has no component labels. Here is the actions [log](https://github.com/ranj8work/magma/runs/4173717494?check_suite_focus=true).

Tested update call working when the PR has NO labels at all. Here is the actions [log](https://github.com/ranj8work/magma/runs/4173741944?check_suite_focus=true).

## Additional Information

- [NO] This change is backwards-breaking

### DIFF
--- a/.github/workflows/autolabel-pullrequests.yml
+++ b/.github/workflows/autolabel-pullrequests.yml
@@ -43,26 +43,6 @@ jobs:
 
             let newCompLbls = new Set(); // Set of new label strings
 
-            const curLblObjs = await github.issues.listLabelsOnIssue({
-              issue_number: context.issue.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-            });
-
-            // Delete current component labels from PR
-            // Will identify new component labels later based on the files modified in the PR
-            for (const l of curLblObjs['data']) {
-              //console.log("Current LabelName: " + l['name'] + " LabelDescription: " + l['description']);
-              if(l['name'].startsWith("component: "))  {
-                github.issues.removeLabel({
-                  issue_number: context.issue.number,
-                  owner: context.repo.owner,
-                  repo: context.repo.repo,
-                  name: l['name'],
-                });
-              } // end of if block
-            } // end of for loop
-
             // Fetch files modified in the PR
             const pulledFiles = await github.pulls.listFiles({
               owner: context.repo.owner,
@@ -131,6 +111,31 @@ jobs:
                 newCompLbls.add("component: docs");
                 break;
              } // end of switch case
+            } // end of for loop
+
+
+            const curLblObjs = await github.issues.listLabelsOnIssue({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+            });
+
+            // Delete current component labels from PR that are no more valid
+            for (const l of curLblObjs['data']) {
+              //console.log("Current LabelName: " + l['name'] + " LabelDescription: " + l['description']);
+              if(l['name'].startsWith("component: "))  {
+                if(newCompLbls.has(l['name'])  {
+                  newCompLbls.delete(l['name']);
+                }
+                else  {
+                  github.issues.removeLabel({
+                    issue_number: context.issue.number,
+                    owner: context.repo.owner,
+                    repo: context.repo.repo,
+                    name: l['name'],
+                  });
+                }
+              } // end of if block
             } // end of for loop
 
             if( newCompLbls.size == 0 )  {
--- a/.github/workflows/autolabel-pullrequests.yml
+++ b/.github/workflows/autolabel-pullrequests.yml
@@ -124,7 +124,7 @@ jobs:
             for (const l of curLblObjs['data']) {
               //console.log("Current LabelName: " + l['name'] + " LabelDescription: " + l['description']);
               if(l['name'].startsWith("component: "))  {
-                if(newCompLbls.has(l['name'])  {
+                if(newCompLbls.has(l['name']))  {
                   newCompLbls.delete(l['name']);
                 }
                 else  {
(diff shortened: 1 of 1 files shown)


quentinDERORY (changes requested):
I don't want the code to always use the set API:
- when there are no new labels, don't call any APIS.
- when there are one or more less component, just delete these component.
- when there are one or more added components, just add those.
- use the set API if you want when there are none.

All label changes are reflected on the UI of the PR, it will be way more understandable for users if they see only the modified labels by this bot.

quentinDERORY:
And this code does not address [#8765 ](https://github.com/magma/magma/issues/8765) , unlinking it from the PR

ranj8work:
> I don't want the code to always use the set API:
> 
> * when there are no new labels, don't call any APIS.
> * when there are one or more less component, just delete these component.
> * when there are one or more added components, just add those.
> * use the set API if you want when there are none.
> 
> All label changes are reflected on the UI of the PR, it will be way more understandable for users if they see only the modified labels by this bot.

@quentinDERORY - I rewrote the logic to address your comments. I tested and submitted the PR again. Please review.

.github/workflows/autolabel-pullrequests.yml
```diff
@@ -65,58 +79,85 @@ jobs:
...
             } // end of for loop
 
-            if( lbls.size == 0 )  {
-              console.log("No components changed in this PR: " + context.issue.number);
+            if( newLbls.size == 0 )  {
+              console.log("No new component files changed in this PR: " + context.issue.number);
             }
             else  {
-              let uniqLbls = Array.from(lbls);
+              for (const l of curLblObjs['data']) {
```
quentinDERORY: why don't you use curLbls here?

.github/workflows/autolabel-pullrequests.yml
```diff
@@ -65,58 +79,85 @@ jobs:
...
+                if(!(l['name'].startsWith("component: ")))  {
+                  console.log("Adding Label:" + l['name']);
+                  newLbls.add(l['name']);
+                }
+              }
 
+              let uniqLbls = Array.from(newLbls);
```
quentinDERORY: you are still using the set API, please only post the difference aka use put or delete API and only use set when the current labels are empty.

quentinDERORY (changes requested):

.github/workflows/autolabel-pullrequests.yml
ranj8work: @quentinDERORY Made changes to not sue set API no more.

.github/workflows/autolabel-pullrequests.yml
```diff
@@ -65,62 +82,74 @@ jobs:
...
+                    owner: context.repo.owner,
+                    repo: context.repo.repo,
+                    name: l,
+                  });
+                } // end of if block
+              } // end of for loop
 
-              github.issues.setLabels({
+              let uniqLbls = Array.from(newCompLbls);
+              github.issues.addLabels({
```
quentinDERORY, line 146: Are you sure your code will not duplicate labels that were already present before?

ranj8work: yes @quentinDERORY, because the `newCompLbls` is a Set too. Why you think it could duplicate ?

quentinDERORY: because you are using add api, and if there are labels already present on the PR, it might get duplicated? no?
I would recommend only adding the ones that are not already present on the PR.

ranj8work: you are right @quentinDERORY. I fixed it now to avoid duplicate labels due to the add. Please check now.

.github/workflows/autolabel-pullrequests.yml
```diff
@@ -41,15 +41,36 @@ jobs:
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+            });
+
+            // Delete current component labels from PR
+            // Will identify new component labels later based on the files modified in the PR
+            for (const l of curLblObjs['data']) {
+              //console.log("Current LabelName: " + l['name'] + " LabelDescription: " + l['description']);
+              if(l['name'].startsWith("component: "))  {
+                github.issues.removeLabel({
```
quentinDERORY: you should not remove all labels everytime. You should only apply the difference.

If there is one less remove one less, if there is one more just add one more.

Don't delete every components labels before applying them.

quentinDERORY (changes requested):
Still require changes

.github/workflows/autolabel-pullrequests.yml
ranj8work: @quentinDERORY - I understand and will do the needful. I should have it ready asap.

ranj8work: @quentinDERORY the changes are done and tested as per your last review comments. Please review and let me know if you have any questions. Thank you !

quentinDERORY (approved):

PR #10233, opened 2021-11-10
feat(orc8r): Support Minikube Orc8r with local container registry

hcgatewood:
## Summary

Previously, we required that Orc8r container images be uploaded to a non-local registry (generally DockerHub), then pulling those images back down to local when applying the Helm charts. This is painful and a terrible hit to validation cycle time.

Instead, Minikube supports using the local Docker daemon's registry as its source registry. This PR fixes a small blocker, and documents how to accomplish this.

## Test Plan

- Manual

## Additional Information


themarwhal (approved):

hcgatewood:
Force-merging to unblock @reginawang3495

PR #10241, opened 2021-11-11
chore(ci): disable cwf integ test notifications

quentinDERORY:
## Summary

Disable GHA cwf integ test notifications
Delete CWF integ tests from circle ci as they are running better in GHA now.

## Additional Information


themarwhal (approved):
yes let's revisit this if we can get the test to be less flaky. Can we get the cwf-integ-test results published in the dashabord page?

quentinDERORY:
This is the plan, we will have all integ test results in the dashboard ultimately.

PR #10243, opened 2021-11-11
feat: Add mme app latency  metric

wallyrb:

## Summary

Add mme app latency  metric. 
- For MME APP re-using the already calculated latency mme_app_last_msg_latency
- For S1AP re-using the already calculated latency s1ap_last_msg_latency

## Test Plan

- Tested in local environment

```
(python) vagrant@magma-dev-focal:~/magma/lte/gateway$ service303_cli.py metrics mme

....

name: "mme_app_last_msg_latency"
help: ""
type: GAUGE
metric {
  gauge {
    value: 2601.0
  }
}

name: "s1ap_last_msg_latency"
help: ""
type: GAUGE
metric {
  gauge {
    value: 65762.0
  }
}


```
image

### DIFF
--- a/lte/gateway/c/core/oai/tasks/mme_app/mme_app_main.c
+++ b/lte/gateway/c/core/oai/tasks/mme_app/mme_app_main.c
@@ -555,6 +555,8 @@ static int handle_stats_timer(zloop_t* loop, int id, void* arg) {
   stats_msg.nb_ue_connected        = mme_app_desc_p->nb_ue_connected;
   stats_msg.nb_default_eps_bearers = mme_app_desc_p->nb_default_eps_bearers;
   stats_msg.nb_s1u_bearers         = mme_app_desc_p->nb_s1u_bearers;
+  stats_msg.nb_mme_app_last_msg_latency = mme_app_last_msg_latency;
+
   return send_mme_app_stats_to_service303(
       &mme_app_task_zmq_ctx, TASK_MME_APP, &stats_msg);
 }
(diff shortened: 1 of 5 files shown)


ardzoht (changes requested):
we need to add `pre_mme_task_msg_latency` as well

ardzoht (approved):
lgtm

lte/gateway/c/core/oai/tasks/mme_app/mme_app_main.c
```diff
@@ -78,6 +78,9 @@ static int handle_message(zloop_t* loop, zsock_t* reader, void* arg) {
   mme_app_last_msg_latency =
       ITTI_MSG_LATENCY(received_message_p);  // microseconds
   pre_mme_task_msg_latency = ITTI_MSG_LASTHOP_LATENCY(received_message_p);
+  if (pre_mme_task_msg_latency < 0) {
```
wallyrb: @ardzoht I notice majority of times `pre_mme_task_msg_latency` is populated with large negative value `-4702111234474983746'. It was always the same negative value.   I think this is populated when there is not a value on pre_mme_task_msg_latency which is obtained from `[ittiMsgHeader.last_hop_latency](https://github.com/magma/magma/blob/master/lte/gateway/c/core/oai/lib/itti/intertask_interface.h#L64)` 

For this cases, I am setting the latency to zero for negative values

lte/gateway/c/core/oai/tasks/service303/service303_mme_stats.c
```diff
@@ -28,12 +28,20 @@ void service303_mme_app_statistics_read(
...
+      "pre_mme_task_msg_latency", stats_msg_p->nb_pre_mme_task_msg_latency,
+      label);
 }
 
 void service303_s1ap_statistics_read(
     application_s1ap_stats_msg_t* stats_msg_p) {
   size_t label = 0;
   set_gauge("enb_connected", stats_msg_p->nb_enb_connected, label);
+  set_gauge(
+      "s1ap_last_msg_latency", stats_msg_p->nb_s1ap_last_msg_latency, label);
```
wallyrb, line 41: Adding the metric 's1ap_last_msg_latency` since I think will be useful to measure latency of s1ap task

PR #10250, opened 2021-11-11
feat(orc8r): Added token-based REST API auth

christinewang5:
## Summary
* Added concepts **users** and **policies** and methods to manage them 
* Added middleware to check permissions with each request 

For more details: [https://fb.quip.com/q1XaAUVQ9MQt](https://fb.quip.com/q1XaAUVQ9MQt)

## Test Plan
* Tested w swagger UI
https://user-images.githubusercontent.com/25142344/143141163-5fbb8615-4f70-4f50-965e-9b525f5d0f15.mov
* Wrote unit tests in `rest_middleware_test` and `handlers`

## Next Steps 
- [x] Bootstrap admin token 
- [x] Scope endpoints by network id
- [x] Test more rigorously


orc8r/cloud/configs/service_registry.yml
```diff
@@ -75,10 +75,17 @@ services:
...
     labels:
       orc8r.io/analytics_collector: "true"
-
+      orc8r.io/obsidian_handlers: "true"
+      orc8r.io/swagger_spec: "true"
+    annotations:
+      orc8r.io/obsidian_handlers_path_prefixes: >
+        /magma/v1/user,
+        /magma/v1/user/:username,
+        /magma/v1//user/login,
```
hcgatewood: These are "prefixes", so just line 86 is sufficient

orc8r/cloud/go/models/swagger-common.yml
```diff
@@ -41,6 +41,15 @@ schemes:
   - https
   - http
 
+securityDefinitions:
+  tokenAuth:
+    type: apiKey
+    in: header
+    name: X-Magma-Client-Access-Token
```
hcgatewood: So I thought about it a bit, and I think we should go ahead and mirror github's style of doing username:token, and validating that the token is owned by the username

And in that case, we want to use the http basic auth security definition, rather than the apikey type

https://swagger.io/docs/specification/2-0/authentication/basic-authentication/

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -100,3 +104,104 @@ func makeErr(decorate logDecorator, status int, errFmt string, errArgs ...interf
 	glog.V(1).Infof("REST middleware (obsidian) rejected request: %s", decorate(errFmt, errArgs...))
 	return echo.NewHTTPError(status, fmt.Sprintf(errFmt, errArgs...))
 }
+
+// TokenMiddleware parses the <username>:<token> from the header, validates the token,
+// then checks if the request is within the specified permissions granted to the user
```
hcgatewood: Nit: sentence
```suggestion
// then checks if the request is within the specified permissions granted to the user.
```

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -100,3 +104,104 @@ func makeErr(decorate logDecorator, status int, errFmt string, errArgs ...interf
...
+				return echo.NewHTTPError(http.StatusInternalServerError, err)
+			}
+			username = RootUsername
+			token = adminToken
+			// return echo.NewHTTPError(http.StatusBadRequest, err)
+		}
+
+		if err := certifier.ValidateToken(token); err != nil {
+			return echo.NewHTTPError(http.StatusBadRequest, err)
+		}
```
hcgatewood: Performing this validation is useful, but let's move it to the certifier service -- that way, every request, regardless of who sends it, will receive the benefit of this validation

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -100,3 +104,104 @@ func makeErr(decorate logDecorator, status int, errFmt string, errArgs ...interf
...
+
+		// take tokenList, request type, resource and exchange for permission decision
+		requestType := getRequestAction(req, nil)
+		resource := req.RequestURI
+		getPDReq := certifierprotos.GetPolicyDecisionRequest{
+			TokenList:     tokensList,
+			RequestAction: requestType,
+			Resource:      resource,
+		}
+		pd, err := certifier.GetPolicyDecision(req.Context(), &getPDReq)
```
hcgatewood: I think we can simplify this to 1 rpc -- the GetPolicyDecisionRequest should

- Also take an (optional) username
- In this case, the token list will be exactly equal to the single token from the request header

This way, the policy decision rpc can be a single round-trip, with all the validation handled at the certifier service

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -100,3 +104,104 @@ func makeErr(decorate logDecorator, status int, errFmt string, errArgs ...interf
...
+		if err != nil || pd.Effect == certifierprotos.Effect_DENY {
+			return echo.NewHTTPError(http.StatusForbidden, err)
+		}
```
hcgatewood: Let's return different errors here -- if there's an internal err, we want to (a) log it with stack trace (`%+v`) and (b) return a server http err (5xx)

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -100,3 +104,104 @@ func makeErr(decorate logDecorator, status int, errFmt string, errArgs ...interf
...
+		Password: []byte("password"),
+	}
+	policy := certifierprotos.Policy{
+		Effect:    certifierprotos.Effect_ALLOW,
+		Action:    certifierprotos.Action_WRITE,
+		Resources: &certifierprotos.Policy_ResourceList{Resource: []string{"*"}},
+	}
+	req := certifierprotos.CreateUserRequest{
+		User:   &user,
+		Policy: &policy,
```
hcgatewood: General nit: define the protos as ptrs, then you won't have to take the addr here

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -100,3 +104,104 @@ func makeErr(decorate logDecorator, status int, errFmt string, errArgs ...interf
...
+	}
+	tokens, err := certifier.CreateUser(ctx, &req)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create bootstrap admin user")
+	}
+	return tokens.GetToken()[0], nil
```
hcgatewood: 1. don't use the "Get" helpers unless you explicitly expect that the return value might be empty, 2. if "token" is a list it should be "tokens"

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -43,15 +46,55 @@ func TestMiddlewareWithoutCertifier(t *testing.T) {
 	)
 	assert.NoError(t, err)
 	assert.Equal(t, 503, s)
+}
+
+func TestAuthMiddleware(t *testing.T) {
+	// set up auth middleware by creating root user, non-admin user bob, and their respective policies
```
hcgatewood: Nit: Sentence-case

```suggestion
	// Set up auth middleware by creating root user, non-admin user bob, and their respective policies
```

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -43,15 +46,55 @@ func TestMiddlewareWithoutCertifier(t *testing.T) {
...
+	bob := "bob"
+	root := "root"
+	bobUser, bobToken := test_utils.CreateTestUser(bob, "password")
+	rootUser, rootToken := test_utils.CreateTestUser(root, "password")
+	err := store.PutUser(bob, &bobUser)
+	err = store.PutUser(root, &rootUser)
+	bobPolicy := test_utils.CreateUserPolicy(t, bobToken)
+	rootPolicy := test_utils.CreateAdminPolicy(rootToken)
+	err = store.PutPolicy(bobToken, &bobPolicy)
+	err = store.PutPolicy(rootToken, &rootPolicy)
```
hcgatewood: For all of these err assignments, need to assert.NoError after each one, otherwise a non-nil err could be overwritten by a nil err

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -43,15 +46,55 @@ func TestMiddlewareWithoutCertifier(t *testing.T) {
 	)
 	assert.NoError(t, err)
 	assert.Equal(t, 503, s)
+}
+
+func TestAuthMiddleware(t *testing.T) {
+	// set up auth middleware by creating root user, non-admin user bob, and their respective policies
+	certifier_test_service.StartTestService(t)
+	store := test_utils.GetCertifierBlobstore(t)
+	bob := "bob"
```
hcgatewood: I would move this block into 2 helper fns -- 1. createTestUser 2. createTestAdmin. Would be easier to skim

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -43,15 +46,55 @@ func TestMiddlewareWithoutCertifier(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	e := startTestMiddlewareServer(t)
+	e.Use(access.TokenMiddleware)
+	listener := WaitForTestServer(t, e)
+	if listener == nil {
+		return
+	}
+	urlPrefix := "http://" + listener.Addr().String()
+	testNetworkURLRoot := urlPrefix + "/magma/v1/networks"
```
hcgatewood: 1. Let's do this with strformat rather than string append, just for uniformity
2. Is this the only way to get the listener addr? Does the echo object not expose it? Is fine either way but would prefer through echo obj if possible

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -43,15 +46,55 @@ func TestMiddlewareWithoutCertifier(t *testing.T) {
...
+
+	// Test READ endpoint
+	s, err := SendRequestWithToken("GET", testNetworkURLRoot+obsidian.UrlSep+TEST_NETWORK_ID, root, rootToken)
```
hcgatewood: Nit: coalesce these iterative requests into a table-driven test

https://medium.com/@matryer/5-simple-tips-and-tricks-for-writing-unit-tests-in-golang-619653f90742#:~:text=4.%20Write%20table%20driven%20tests

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -43,15 +46,55 @@ func TestMiddlewareWithoutCertifier(t *testing.T) {
...
+	e := startTestMiddlewareServer(t)
+	e.Use(access.TokenMiddleware)
+	listener := WaitForTestServer(t, e)
+	if listener == nil {
+		return
+	}
+	urlPrefix := "http://" + listener.Addr().String()
+	testNetworkURLRoot := urlPrefix + "/magma/v1/networks"
+
+	// Test READ endpoint
```
hcgatewood: Add a few more test cases -- e.g. cover a 403 for read endpoint

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -214,55 +257,39 @@ func TestMiddleware(t *testing.T) {
...
 
-	// Endpoint requiring Network Wildcard READ Access Permissions
-	e.GET(RegisterNetworkV1, func(c echo.Context) error {
+	dummyHandlerFunc := func(c echo.Context) error {
 		return c.String(http.StatusOK, "All good!")
-	})
+	}
+
+	// Endpoint requiring Network Wildcard READ Access Permissions
+	e.GET(RegisterNetworkV1, dummyHandlerFunc)
```
hcgatewood, line 161: Same here, make these table-driven

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,208 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/labstack/echo"
+	"golang.org/x/crypto/bcrypt"
+
+	"magma/orc8r/cloud/go/obsidian"
+	"magma/orc8r/cloud/go/services/certifier"
+	certProtos "magma/orc8r/cloud/go/services/certifier/protos"
```
hcgatewood: No need to alias

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,208 @@
...
+
+const (
+	UserParam  = ":username"
+	User       = "user"
+	ListUser   = obsidian.V1Root + User
+	ManageUser = ListUser + obsidian.UrlSep + UserParam
+	Login      = ListUser + obsidian.UrlSep + "login"
+)
+
+func GetHandlers(storage storage.CertifierStorage) []obsidian.Handler {
```
hcgatewood: So we probably don't want the handlers to share the underlying store with the certifier grpc servicer, since it breaks the DAO abstraction. Instead, have the handlers be a stateless component that forwards requests to the certifier grpc servicer

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,208 @@
...
+	User struct {
+		Username string `json:"username"`
+		Password string `json:"password"`
+	} `json:"user"`
+	Policy struct {
+		Effect    string   `json:"effect"`
+		Action    string   `json:"action"`
+		Resources []string `json:"resource"`
+	} `json:"policy"`
+}
```
hcgatewood: This will work great, but for uniformity let's define this via swagger object, and then use that generated swagger object in the handlers

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,208 @@
...
+
+func getCreateUserHandler(storage storage.CertifierStorage) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		var data CreateUserRequest
+		err := json.NewDecoder(c.Request().Body).Decode(&data)
+		if err != nil {
+			return echo.NewHTTPError(http.StatusInternalServerError, fmt.Errorf("error decoding request body for creating user: %v", err))
+		}
+
+		// parse user from response body
```
hcgatewood: Nit https://magma.github.io/magma/docs/next/contributing/contribute_conventions#:~:text=Avoid%20%22topic%20sentence%22%20comments

hcgatewood (commented):
Looking great

Got through to most of orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go -- couple items to tackle, I'll finish reviewing a little later

orc8r/cloud/go/services/certifier/tokens.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"crypto/rand"
+	"errors"
+	"hash/crc32"
+	"strings"
+
+	"github.com/jxskiss/base62"
+)
+
+type TokenPrefix string
```
hcgatewood: I would almost call this "token type", where the prefix indicates the type

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+		policy := &certProtos.Policy{
+			Token:  token,
+			Effect: effect,
+			Action: action,
+			Resources: &certProtos.Policy_ResourceList{
+				Resource: data.Policy.Resources,
+			},
+		}
+		if err = storage.PutPolicy(token, policy); err != nil {
```
hcgatewood: Store the policy first, before creating the user -- if something fails halfway through, we don't want to the user to half-exist

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,208 @@
...
+		var action certProtos.Action
+		switch data.Policy.Action {
+		case certProtos.Action_READ.String():
+			action = certProtos.Action_READ
+		case certProtos.Action_WRITE.String():
+			action = certProtos.Action_WRITE
+		default:
+			action = certProtos.Action_NONE
+		}
+		policy := &certProtos.Policy{
```
hcgatewood: Move this into a helper fn somewhere

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,208 @@
...
+			Token:  token,
+			Effect: effect,
+			Action: action,
+			Resources: &certProtos.Policy_ResourceList{
+				Resource: data.Policy.Resources,
```
hcgatewood: Yeah from offline: we want the a 1-1-1 ratio between effect-token-resource. I think that will provide a bit more flexibility, e.g. 1 policy can encode e.g. allow 1 resource but deny another resource. So a list of this triplet, rather than just a list of resources

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,208 @@
...
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package handlers
```
hcgatewood, line 14: Hm okay so having thought about it a bit -- I think we should go ahead and split out "users" and "users/:user/tokens" endpoints -- that will allow a lot more flexibility + separation of concerns

So then creating a user just creates the user -- no policies/tokens. Then, we use a separate "users/:user/tokens" endpoint to edit token(s)/policy(s) for that user

Thoughts?

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers_test.go
```diff
@@ -0,0 +1,118 @@
...
+type Policy struct {
+	Effect    string   `json:"effect"`
+	Action    string   `json:"action"`
+	Resources []string `json:"resource"`
+}
+
+type CreateUserRequest struct {
+	User   `json:"user"`
+	Policy `json:"policy"`
+}
```
hcgatewood: Yeah all of these should be swagger types

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
```diff
@@ -0,0 +1,153 @@
...
+  dependencies:
+    - 'orc8r/cloud/go/models/swagger-common.yml'
+  temp-gen-filename: certifier-swagger.yml
+  output-dir: orc8r/cloud/go/services/certifier/obsidian
+  types:
+    - go-struct-name: User
+      filename: user_swaggergen.go
+
+
+info:
```
hcgatewood, line 14: Seems like some of these fields don't exist in other swagger.v1.yml files, e.g. basePath. Maybe because they're in the common swagger yml. Thoughts?

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
```diff
@@ -0,0 +1,153 @@
...
+  description: Certifier REST APIs
+  version: 1.0.0
+
+basePath: /magma/v1
+
+paths:
+  /user:
+    get:
+      responses:
+        "200":
```
hcgatewood: Nit: use single quotes for uniformity

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
```diff
@@ -0,0 +1,153 @@
...
+          description: Success
+          schema:
```
hcgatewood, line 28: Let's return a list of user objects, rather than just a list of strings

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
```diff
@@ -0,0 +1,153 @@
...
+            $ref: '#/definitions/user_with_policy'
+      responses:
+        "201":
+          description: Success
+        default:
+          $ref: './orc8r-swagger-common.yml#/responses/UnexpectedError'
+      summary: Create a user
+      tags:
+        - User
+  /user/{username}:
```
hcgatewood, line 51: "get" for single user?

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
```diff
@@ -0,0 +1,153 @@
...
+    put:
+      parameters:
```
hcgatewood, line 67: Put == change entire user, not just their password

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
```diff
@@ -0,0 +1,153 @@
...
+          description: Success
+        default:
+          $ref: './orc8r-swagger-common.yml#/responses/UnexpectedError'
+      summary: Remove a user
+      tags:
+        - User
+
+  /user/login:
+    post:
+      security: [ ] # no security
```
hcgatewood: Add a why in the comment as well -- "why" no security

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
```diff
@@ -0,0 +1,153 @@
...
+      parameters:
+        - name: user
+          in: body
+          required: true
+          schema:
+            $ref: '#/definitions/user'
+      responses:
+        "200":
+          description: Success
+          schema:
```
hcgatewood: Rather than just the list of tokens, let's return a map of token->policy objects

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
```diff
@@ -0,0 +1,153 @@
...
+            type: array
+            items:
+              type: string
+        default:
+          $ref: './orc8r-swagger-common.yml#/responses/UnexpectedError'
+      summary: List all the tokens given a user
+      tags:
+        - User
+parameters:
```
hcgatewood, line 162: Add space above this section

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
```diff
@@ -0,0 +1,153 @@
...
+      - action
+      - resources
+    properties:
+      effect:
+        type: string
+        enum: [UNKNOWN, DENY, ALLOW]
+      action:
+        type: string
+        enum: [NONE, READ, WRITE]
+      resources:
```
hcgatewood, line 180: From offline: resources can be any of the following, at this point in time:

- a (relative) url path (and allow wildcards?)
- a network_id
- a tenant_id

And we want to encode in the type system the difference between these 3

orc8r/cloud/go/services/certifier/protos/certifier.proto
```diff
@@ -48,6 +48,58 @@ message GetCARequest {
   CertType cert_type = 1;
 }
 
+
+message User {
+  string username = 1;
+  bytes password = 2;
+  message TokenList {
```
hcgatewood: Nit: I'm not the biggest fan of nested message types like this, especially if it's conceivable the nested type could be used on its own. So I think it makes sense, here and below, to not nest these types

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -449,3 +453,143 @@ func (srv *CertifierServer) CollectGarbageImpl(ctx context.Context) (int, error)
...
 	return count, nil
 }
+
+// GetUserTokens fetches the tokens associated with a user
+func (srv *CertifierServer) GetUserTokens(ctx context.Context, getOpReq *certprotos.GetUserRequest) (*certprotos.User_TokenList, error) {
+	username := getOpReq.GetUsername()
+	user, err := srv.store.GetUser(username)
+	if err != nil {
+		return &certprotos.User_TokenList{}, status.Errorf(
+			codes.Internal, "failed to fetch user %s from database: %v", username, err)
```
hcgatewood: Nit: fix newlines to either be 1 line total or 1 arg per line

Also just return nil obj when also returning an err

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -449,3 +453,143 @@ func (srv *CertifierServer) CollectGarbageImpl(ctx context.Context) (int, error)
 	}
 	return count, nil
 }
+
+// GetUserTokens fetches the tokens associated with a user
+func (srv *CertifierServer) GetUserTokens(ctx context.Context, getOpReq *certprotos.GetUserRequest) (*certprotos.User_TokenList, error) {
```
hcgatewood: This fn seems to do 2 things in 1 -- list user tokens and check if a passed token is owned by a user. I think it would be cleaner to separate these into 2 separate fns

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -449,3 +453,143 @@ func (srv *CertifierServer) CollectGarbageImpl(ctx context.Context) (int, error)
...
+func (srv *CertifierServer) GetPolicyDecision(ctx context.Context, getPDReq *certprotos.GetPolicyDecisionRequest) (*certprotos.PolicyDecision, error) {
+	tokens := getPDReq.TokenList
+	resource := getPDReq.Resource
+	action := getPDReq.RequestAction
+	finalEffect := certprotos.Effect_DENY
+	for _, t := range tokens.GetToken() {
+		effect, _ := srv.getPolicyDecisionFromToken(t, resource, action)
+		switch effect {
+		case certprotos.Effect_DENY:
+			// returns early if there is a deny in any of the permissions
```
hcgatewood: Nit: "Return early ..."

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -449,3 +453,143 @@ func (srv *CertifierServer) CollectGarbageImpl(ctx context.Context) (int, error)
...
+		}
+	}
+	if !flag {
+		return &certprotos.User_TokenList{}, status.Errorf(codes.PermissionDenied, "token is not registered with user %s", user.Username)
+	}
+	return tokens, nil
+}
+
+// GetPolicyDecision makes a policy decision when a user attempts to access a resource
+func (srv *CertifierServer) GetPolicyDecision(ctx context.Context, getPDReq *certprotos.GetPolicyDecisionRequest) (*certprotos.PolicyDecision, error) {
```
hcgatewood: Flesh out this doc comment a bit with the details of how we make the decision. The description should make it clear what happens if e.g. 1 policy says "allow" and another says "deny"

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -449,3 +453,143 @@ func (srv *CertifierServer) CollectGarbageImpl(ctx context.Context) (int, error)
...
+	policy, err := srv.store.GetPolicy(token)
+	if err != nil {
+		status.Errorf(codes.Internal, "failed to get policy from db: %v", err)
+	}
+	effect, err := checkResource(resource, policy)
+	// return if the policy denies the user or is unknown for the resource
+	if effect == certprotos.Effect_DENY || effect == certprotos.Effect_UNKNOWN {
+		return effect, status.Errorf(codes.PermissionDenied, "not authorized to read/write resource")
+	} else if err != nil {
+		return effect, err
```
hcgatewood: Check err first. If there's an err, return "deny" and also log the err -- cause it's a weird situation we definitely want to communicate to the operator

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -449,3 +453,143 @@ func (srv *CertifierServer) CollectGarbageImpl(ctx context.Context) (int, error)
...
+	return effect, nil
+}
+
+// checkResource checks if the requested resource is authorized by the policy
+func checkResource(resource string, policy *certprotos.Policy) (certprotos.Effect, error) {
+	effect := policy.GetEffect()
+	// checks if any of policy's resource list allows/denies the requested resource
+	for _, pr := range policy.GetResources().GetResource() {
+		pattern := buildPatternHelper(pr, resource)
+		if ok, err := path.Match(pattern, resource); ok {
```
hcgatewood: Seems like we could use https://github.com/mattn/go-zglob/blob/master/zglob_test.go#L144 to support e.g. `lte/**`

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -449,3 +453,143 @@ func (srv *CertifierServer) CollectGarbageImpl(ctx context.Context) (int, error)
...
+			return effect, nil
+		} else {
+			glog.Errorf("failed to match resource path %v", err)
+		}
+	}
+	// defaults to unknown if resource is not explicitly allowed/denied by policy
+	return effect, nil
```
hcgatewood, line 739: return unknown (?)

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -449,3 +453,143 @@ func (srv *CertifierServer) CollectGarbageImpl(ctx context.Context) (int, error)
...
+}
+
+// checkAction checks if the requested action is authorized by the policy
+func checkAction(action certprotos.Action, policy *certprotos.Policy) certprotos.Effect {
+	if policy.GetAction() == certprotos.Action_WRITE {
+		return certprotos.Effect_ALLOW
+	} else if policy.GetAction() == certprotos.Action_READ && action == certprotos.Action_READ {
```
hcgatewood: General: when using early returns like this, don't use "else" or "else if" -- just take the next action at the same level

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -449,3 +453,143 @@ func (srv *CertifierServer) CollectGarbageImpl(ctx context.Context) (int, error)
...
+	resourceComponents := strings.Split(resource, "/")
+	lenDiff := len(resourceComponents) - len(patternComponents)
+	if lenDiff != 0 {
+		return pattern + strings.Repeat("/*", lenDiff)
+	}
+	return pattern
+}
+
+// CreateUser creates a new user with the specified password and policy
+func (srv *CertifierServer) CreateUser(ctx context.Context, createUserReq *certprotos.CreateUserRequest) (*certprotos.User_TokenList, error) {
```
hcgatewood: Nit: generally speaking, we wanna organize files as public funcs first, then private

orc8r/cloud/go/services/certifier/storage/storage.go
```diff
@@ -38,4 +38,22 @@ type CertifierStorage interface {
 	// DeleteCertInfo removes the serial number and its certificate info.
 	// Returns success even when nothing is deleted (i.e. serial number not found).
 	DeleteCertInfo(serialNumber string) error
+
+	// ListUser lists the usernames of all the current users
+	ListUser() ([]string, error)
```
hcgatewood: I think it might make sense to put these into a different storage iface -- 1. certinfo 2. users 3. policies

Or, maybe we keep them all in the same top-level storage iface, but then define it like this

```
type CertifierStorage interface {
	CertificateStorage
	UserStorage
	PolicyStorage
}

type CertificateStorage interface {
...
}


...

File: orc8r/cloud/go/services/certifier/storage/storage_blobstore.go [hunk 46]
```diff
@@ -168,3 +178,164 @@ func (c *certifierBlobstore) DeleteCertInfo(serialNumber string) error {
...
+	users, err := blobstore.ListKeys(store, placeholderNetworkID, UserType)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list keys")
+	}
+
+	return users, store.Commit()
+}
+
+func (c *certifierBlobstore) GetUser(username string) (*protos.User, error) {
+	store, err := c.factory.StartTransaction(nil)
```
hcgatewood: Can mark read-only txs as read-only in the tx type

orc8r/cloud/go/services/certifier/storage/storage_blobstore.go
```diff
@@ -168,3 +178,164 @@ func (c *certifierBlobstore) DeleteCertInfo(serialNumber string) error {
...
+	err = store.Delete(placeholderNetworkID, storage.TKs{tk})
+
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed to delete users: %s", err)
+	}
+
+	return store.Commit()
+}
+
+func userFromBlob(blob blobstore.Blob) (protos.User, error) {
```
hcgatewood: I would make these methods on the protos.User object, as a helpers file in the generated protos directory

orc8r/cloud/go/services/certifier/tokens.go
```diff
@@ -0,0 +1,92 @@
...
+		return "", err
+	}
+	// generate checksum bytes
+	checksum := crc32.ChecksumIEEE(bytes)
+	checksumBytes := i32tob(checksum)
+	// combine to form final token
+	finalBytes := append(bytes, checksumBytes...)
+	token := base62.StdEncoding.EncodeToString(finalBytes)
+
+	return string(prefix) + "_" + token, nil
```
hcgatewood: strformat this

orc8r/cloud/go/services/certifier/tokens.go
```diff
@@ -0,0 +1,92 @@
...
+	return nil
+}
+
+func stripTokenHeader(token string) (string, error) {
+	s := strings.Split(token, "_")
+	prefix, value := s[0], s[1]
+
+	if len(s) != 2 {
+		return "", errors.New("missing token prefix")
+	}
```
hcgatewood: Perform the len check first, for safety

orc8r/cloud/go/services/certifier/tokens.go
```diff
@@ -0,0 +1,92 @@
...
+	return "", errors.New("invalid token prefix")
+}
+
+func validateTokenChecksum(token string) error {
+	bytes, err := base62.DecodeString(token)
+	if err != nil {
+		return err
+	}
+	orignalChecksum := btoi32(bytes[len(bytes)-4:])
+	newChecksum := crc32.ChecksumIEEE(bytes[:len(bytes)-4])
```
hcgatewood: Nit: I would call this "claimedChecksum" and "calculatedChecksum"

orc8r/cloud/go/services/certifier/tokens.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	return "", errors.New("invalid token prefix")
+}
+
+func validateTokenChecksum(token string) error {
+	bytes, err := base62.DecodeString(token)
+	if err != nil {
+		return err
+	}
+	orignalChecksum := btoi32(bytes[len(bytes)-4:])
```
hcgatewood: 1. Check the bytes is actually long enough to perform this slice
2. "4" is a magic number -- use a const with a doc comment

orc8r/cloud/go/services/certifier/tokens.go
```diff
@@ -0,0 +1,92 @@
...
+	newChecksum := crc32.ChecksumIEEE(bytes[:len(bytes)-4])
+	if orignalChecksum != newChecksum {
+		return errors.New("invalid token checksum")
+	}
+	return nil
+}
+
+func i32tob(val uint32) []byte {
```
hcgatewood, line 91: For both of these, add the ref: https://gist.github.com/chiro-hiro/2674626cebbcb5a676355b7aaac4972d

hcgatewood (commented):
Looks fantastic! Let me know if you have any questions on the comments, we can DM or vc if desired

orc8r/cloud/go/obsidian/access/rest_middleware.go
christinewang5: Wait does it not already checks it for every request just by being in the middleware? 
The reason I put this here and not in the service is so that we don't have to make yet another rpc call to the service, we can return early if the token checksum isn't valid. 
But lmk if you mean sth else!

orc8r/cloud/go/services/certifier/servicers/certifier.go
christinewang5: Do you mean sth like this?
```
if policy.GetAction() == certprotos.Action_WRITE {
    return certprotos.Effect_ALLOW
}
if policy.GetAction() == certprotos.Action_READ && action == certprotos.Action_READ {
    return certprotos.Effect_ALLOW
}
```

orc8r/cloud/go/obsidian/access/rest_middleware.go
christinewang5: Does echo.NewHTTPError not already log things when it is processed through the middleware (from one of my first bootcamp tasks lol)? It doesn't log the stack trace though, should I just double log it?
https://github.com/christinewang5/magma/blob/auth_token/orc8r/cloud/go/service/service.go#L177-L192

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
christinewang5: I can't really find swagger.v1.yml files that don't have info and basePath... which ones were you referring to? 
Also, I tried removing basePath; it looks like it breaks the spec somewhere.
image

orc8r/cloud/go/services/certifier/servicers/certifier.go
christinewang5: Tried using this library, I'm not entirely sure why they treat ** the same as * ugh. Used another one though https://github.com/bmatcuk/doublestar

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
christinewang5: Hmm I think it might be better if this just handles the password update. The only other field in the user are the tokens. I feel like it would not make sense for the admin to put malformed tokens that are not associated with any policy here. The admin can manage the tokens from the separate token endpoints. Lmk what you think.

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
christinewang5: Agreed! Added these endpoints

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
christinewang5: I left the login endpoint to just return the tokens as strings. It seems more convenient for the user to use and copy.
I added the endpoint `GET /user/{username}/tokens` that returns the token->policy objects if they want to see the policy details of the token.

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
christinewang5: Couldn't really find how to expose through the echo object...

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -449,3 +320,444 @@ func (srv *CertifierServer) CollectGarbageImpl(ctx context.Context) (int, error)
 	}
 	return count, nil
 }
+
+// GetPolicyDecision makes a policy decision when a user attempts to access a resource.
+// For conflicting policy decisions from multiple tokens (e.g. one policy is ALLOW and the other DENY), the DENY effect
+// will take precedent.
+// For resources that do not have any policies addressing it, the policy decision defaults to DENY as well.
+func (srv *CertifierServer) GetPolicyDecision(ctx context.Context, getPDReq *certprotos.GetPolicyDecisionRequest) (*certprotos.PolicyDecision, error) {
```
christinewang5: Ahh I need to check my logic here, I don't think it's doing what this docstring is saying bc I changed stuff to support networkid/tenantid scoping...

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -100,3 +106,82 @@ func makeErr(decorate logDecorator, status int, errFmt string, errArgs ...interf
...
+	for _, p := range c.ParamNames() {
+		switch p {
+		case networkIDStr:
+			return certprotos.ResourceType_NETWORK_ID, c.Param(networkIDStr)
+		case tenantIDStr:
+			return certprotos.ResourceType_TENANT_ID, c.Param(tenantIDStr)
+		}
+	}
+	return certprotos.ResourceType_URI, c.Request().RequestURI
+}
```
christinewang5: Sooo to handle scoping by network/tenant ID, I think this is probably the best way to handle those endpoints (i.e. parsing the network id out at the middleware)? But this also means that when creating a policy, the operator needs to specify for each resource type.

orc8r/cloud/go/obsidian/access/rest_middleware.go
hcgatewood: Ah right. The idea I had is -- currently we perform a PD at the REST API middleware. But we could, and might eventually, pass this token around with each gRPC request. In which case we would have multiple callers needing a token->PD conversion

It's def not wrong to do it here, but I think it will be simpler to move it to the certifier service -- 1 owner

orc8r/cloud/go/obsidian/access/rest_middleware.go
hcgatewood: Ah right so -- the distinction here is, generally speaking, we want the "client-facing" part of an internal error to be security-considerate, meaning we don't give "all" the info away e.g. stack trace

We don't do a great job of being consistent about that. But the idea is you give a basic "client-facing" err, and log the details of the err e.g. stack trace

Is not end of the world though, up to you since we're not consistent about it

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
hcgatewood: It's possible I was completely wrong 🙂 Let me check it out during my next pass

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
hcgatewood: I think the value of splitting up /user/{username} and /user/{username}/tokens is that the "user" doesn't contain any of the tokens

I.e. a user is (currently) just a username and password. In which case we wouldn't have to worry about this

orc8r/cloud/go/services/certifier/servicers/certifier.go
hcgatewood: Yep!

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -100,3 +106,82 @@ func makeErr(decorate logDecorator, status int, errFmt string, errArgs ...interf
...
+
+// TokenMiddleware parses the <username>:<token> from the header, validates the token,
+// then checks if the request is within the specified permissions granted to the user.
+func TokenMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		req := c.Request()
+
+		// Skip middleware if request when there is no security requirement
+		// for an endpoint
+		noBasicAuthEndpoints := []string{"/magma/v1/user/login"}
```
hcgatewood: Make this a pkg-level private variable, and add comment for each line describing why we skip auth for it

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -100,3 +106,82 @@ func makeErr(decorate logDecorator, status int, errFmt string, errArgs ...interf
...
+		for _, endpoint := range noBasicAuthEndpoints {
+			if req.RequestURI == endpoint {
+				return next(c)
+			}
+		}
```
hcgatewood: Use a set (map) for this, rather than array

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -100,3 +106,82 @@ func makeErr(decorate logDecorator, status int, errFmt string, errArgs ...interf
...
+		resource := &certprotos.Resource{
+			Action:       getRequestAction(req, nil),
+			ResourceType: resourceType,
+			Resource:     resourceVal,
+		}
+		getPDReq := &certprotos.GetPolicyDecisionRequest{
+			Username: username,
+			Token:    token,
+			Resource: resource,
+		}
```
hcgatewood: I would just inline this into 1 big PDReq literal

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
christinewang5: Then how do users get any available set of tokens that they can use if not stored with the user object?!

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -100,3 +106,82 @@ func makeErr(decorate logDecorator, status int, errFmt string, errArgs ...interf
...
+// or READ+WRITE) corresponding to the request method.
+func getRequestAction(req *http.Request, decorate logDecorator) certprotos.Action {
+	switch req.Method {
+	case "GET", "HEAD":
+		return certprotos.Action_READ
+	case "PUT", "POST", "DELETE":
+		return certprotos.Action_WRITE
+	default:
+		glog.Info(decorate("Unclassified HTTP method '%s', defaulting to read+write requested permissions", req.Method))
+		return certprotos.Action_READ | certprotos.Action_WRITE
```
hcgatewood: Just "write" is sufficient. We're treating "write" as a superset of read

orc8r/cloud/go/obsidian/access/rest_middleware.go
hcgatewood: The *requested resource* should probably always just be a path. The *policy*, however, could encode

- Path
- Network ID(s)
- Tenant(s)

Then we always make a PD request with a requested resource, and the associated *policy* could be path, network ID, or tenant

So e.g. let's say

- Requested resource: /lte/foo
- Policy: NetworkID=foo
- Certifier: determine from resource that the requested network ID is "foo", and allow

Another example

- Requested resource: /lte/bar
- Policy: TenantID=some_tenant (where some_tenant has bar as one of its networks)
- Certifier: determine from resource that the requested network ID is "bar", and learn from the tenants service that "some_tenant" owns network "bar", so allow

Let me know if that makes sense. Can also vc about it tomorrow

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -43,226 +46,137 @@ func TestMiddlewareWithoutCertifier(t *testing.T) {
...
+		{"GET", urlPrefix + RegisterNetworkV1, test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"POST", urlPrefix + RegisterNetworkV1, test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"GET", urlPrefix + RegisterNetworkV1, test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"GET", urlPrefix + "/malformed/url", test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"PUT", urlPrefix + "/malformed/url", test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"GET", urlPrefix + tenantsh.TenantInfoURL, test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"POST", urlPrefix + tenantsh.TenantInfoURL, test_utils.TestRootUsername, rootToken, http.StatusOK},
+
+		// Test non-admin user who has read access to all URI endpoints and networks, read/write access to WRITE_TEST_NETWORK_ID, and no read/write access to specific tenants
+		{"GET", urlPrefix + RegisterNetworkV1 + "/" + TEST_NETWORK_ID, test_utils.TestUsername, userToken, http.StatusOK},
```
hcgatewood: Optional nit: I'm not a fan of manually appending strings -- it's less clear what's happening. Instead, fmt.Sprintf the string

orc8r/cloud/go/obsidian/reverse_proxy/reverse_proxy.go
```diff
@@ -74,6 +77,14 @@ func (r *ReverseProxyHandler) AddReverseProxyPaths(server *echo.Echo, pathPrefix
 				},
 			}
 			g := server.Group(prefix)
+			var serviceConfig certifier.Config
```
hcgatewood: This is reading the config in every time we call this func. Instead, read the config value in during the main.go file, and pass the bool in as config during creation of the ReverseProxyHandler object

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,212 @@
...
+	users, err := certifier.ListUsers(c.Request().Context())
+	if err != nil {
+		return obsidian.MakeHTTPError(err)
+	}
+	return c.JSON(http.StatusOK, users)
+}
+
+func createUserHandler(c echo.Context) error {
+	var data models.User
+	err := json.NewDecoder(c.Request().Body).Decode(&data)
```
hcgatewood: Use .Bind for these https://github.com/magma/magma/blob/fa03861d42609b332db8a0cd8d13309e771224a3/lte/cloud/go/services/lte/obsidian/handlers/handlers.go#L354-L360

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,212 @@
...
+	password := []byte(fmt.Sprintf("%v", *data.Password))
+	user := &protos.User{
+		Username: username,
+		Password: password,
+	}
+	err = certifier.Login(c.Request().Context(), user)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, fmt.Errorf("error logging in: %v", err))
+	}
+	return nil
```
hcgatewood: Shouldn't this return the same as "get user tokens" -- a map of token->policy ?

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func getUserTokensHandler(c echo.Context) error {
+	username := c.Param("username")
+	res, err := certifier.ListUserTokens(c.Request().Context(), &protos.User{Username: username})
```
hcgatewood, line 119: Ideally this would return a map of token->policy

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,212 @@
...
+	for i, resource := range resources {
+		effect := matchEffect(resource.Effect)
+		action := matchAction(resource.Action)
+		resourceType := matchResourceType(resource.ResourceType)
+		resourceProto := &protos.Resource{
+			Effect:       effect,
+			Action:       action,
+			ResourceType: resourceType,
+			Resource:     resource.Resource,
+		}
```
hcgatewood: Same here -- just construct this in 1 big literal

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
```diff
@@ -0,0 +1,241 @@
...
+          schema:
+            $ref: '#/definitions/resources'
+      responses:
+        '201':
+          description: Success
+        default:
+          $ref: './orc8r-swagger-common.yml#/responses/UnexpectedError'
+      tags:
+        - User
+  /user/{username}/{token}:
```
hcgatewood:
```suggestion
  /user/{username}/tokens/{token}:
```

orc8r/cloud/go/services/certifier/protos/certifier.proto
```diff
@@ -48,6 +48,79 @@ message GetCARequest {
...
+message ResourceList {
+  repeated Resource resources = 4;
+}
+
+message Policy {
+  string token = 1;
+  ResourceList resources = 4;
+}
+
+message GetPolicyDecisionRequest {
```
hcgatewood, line 104: Nit: separate out into two pseudo-sections: 1 for Req/Res, and 1 for "normal" types

orc8r/cloud/go/services/certifier/protos/certifier.proto
```diff
@@ -83,6 +156,36 @@ service Certifier {
...
   // cleanup expired certificates
-  //
   rpc CollectGarbage (Void) returns (Void) {}
+
+  // Returns a policy decision given a token, the request method's action
+  // (read/write), and the request resource
+  rpc GetPolicyDecision (GetPolicyDecisionRequest) returns (PolicyDecision) {}
+
+  // Create a new user with their username and password
+  rpc CreateUser (User) returns (Void) {}
```
hcgatewood: Nit: always use `rpc <Func> (<Func>Request) returns (<Func>Response) {}`

Even if the Req/Res is empty, is good practice for forward-compatibility

orc8r/cloud/go/services/certifier/protos/utils.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+const (
+	// UserType is the type of CertInfo used in blobstore type fields.
+	UserType = "user"
+	// PolicyType is the type of policy used in blobstore type fileds
+	PolicyType = "policy"
+)
+
+func UserFromBlob(blob blobstore.Blob) (User, error) {
```
hcgatewood: These 4 are really just "proto to/from blob" with passed type+key. I'm fine with keeping the definitions as-is, but at minimum let's use a shared impl

hcgatewood: P.S. 2 things

1. "utils.go" is generally bad practice for a file name. "Helpers" is also bad, but in this case let's use helpers.go since we're adding our own helpers to a generated directory
2. Consider making these methods on the underlying proto types, rather than standalone funcs

orc8r/cloud/go/services/certifier/storage/storage.go
```diff
@@ -17,8 +17,14 @@ import (
 	"magma/orc8r/cloud/go/services/certifier/protos"
 )
 
-// CertifierStorage provides storage functionality for mapping serial numbers to certificate information.
 type CertifierStorage interface {
```
hcgatewood, line 22: Still add a doc comment to the new CertifierStorage iface

orc8r/cloud/go/services/certifier/storage/storage_blobstore.go
```diff
@@ -30,6 +34,12 @@ const (
 	// CertInfoType is the type of CertInfo used in blobstore type fields.
 	CertInfoType = "certificate_info"
 
+	// UserType is the type of CertInfo used in blobstore type fields.
+	UserType = "user"
+
+	// PolicyType is the type of policy used in blobstore type fileds
+	PolicyType = "policy"
```
hcgatewood: These types were duplicated earlier in this PR -- let's have just 1 canonical home for them

orc8r/cloud/go/services/certifier/tokens.go
```diff
@@ -0,0 +1,102 @@
...
+	"hash/crc32"
+	"strings"
+
+	"github.com/jxskiss/base62"
+)
+
+type TokenType string
+
+const (
+	Personal TokenType = "op"
```
hcgatewood, line 17: What does "op" stand for? Leave a comment to make it clear, and/or choose something that has semantic meaning (if "op" doesn't)

orc8r/cloud/go/services/certifier/tokens.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+// Length of checksum of token's byte array
+const checksumLength = 4
```
hcgatewood: Nit: "checksumLen"

orc8r/cloud/go/services/certifier/tokens.go
```diff
@@ -0,0 +1,102 @@
...
+
+// GenerateToken generates a random 32-byte token with a checksum
+func GenerateToken(tokenType TokenType) (string, error) {
```
hcgatewood: Nit: any time you see a "stutter", it's a good place to be more concise

```suggestion
func GenerateToken(typ TokenType) (string, error) {
```

orc8r/cloud/go/services/certifier/tokens.go
```diff
@@ -0,0 +1,102 @@
...
+		return "", err
+	}
+
+	// Generate checksum bytes
+	checksum := crc32.ChecksumIEEE(bytes)
+	checksumBytes := i32tob(checksum)
+
+	// Combine to form final token
+	finalBytes := append(bytes, checksumBytes...)
+	token := base62.StdEncoding.EncodeToString(finalBytes)
```
hcgatewood, line 41: Why base62 instead of base64? Prefer base64 by default, but if there's a good reason for base62 then add comment to document why

orc8r/cloud/go/services/certifier/tokens.go
```diff
@@ -0,0 +1,102 @@
...
+func stripTokenHeader(token string) (string, error) {
+	s := strings.Split(token, "_")
+	if len(s) != 2 {
+		return "", errors.New("missing token type")
+	}
+	tokenType, value := s[0], s[1]
+
+	// Validate token tokenType
+	switch tokenType {
+	case string(Personal):
```
hcgatewood: Nit: move the type cast to the tokenType

orc8r/cloud/go/tools/accessc/handlers/add_admin_token.go
```diff
@@ -0,0 +1,78 @@
...
+	"magma/orc8r/cloud/go/services/certifier"
+	certprotos "magma/orc8r/cloud/go/services/certifier/protos"
+	"magma/orc8r/cloud/go/tools/commands"
+)
+
+func init() {
+	cmd := CommandRegistry.Add(
+		"add-admin-token",
+		"Add a new admin which has all permissions to read/write all entities and create a new admin token",
+		addAdminToken)
```
hcgatewood: Nit: move closing paren to own line

orc8r/cloud/go/tools/accessc/handlers/add_admin_token.go
```diff
@@ -0,0 +1,78 @@
...
+
+	f := cmd.Flags()
+	f.Usage = func() {
+		fmt.Fprintf(os.Stderr, // std Usage() & PrintDefaults() use Stderr
+			"\tUsage: %s %s [OPTIONS] <Admin Username> <Admin Password>\n", os.Args[0], cmd.Name())
```
hcgatewood: This is non-standard line-breaking. It's (almost) always:

- Open and close paren each on their own line, at same level of indentation
- Each arg on their own line

orc8r/cloud/helm/orc8r/values.yaml
```diff
@@ -216,8 +216,13 @@ certifier:
...
       orc8r.io/analytics_collector: "true"
-    annotations: {}
-
+      orc8r.io/obsidian_handlers: "true"
+      orc8r.io/swagger_spec: "true"
+    annotations:
+      orc8r.io/obsidian_handlers_path_prefixes: >
+        /magma/v1/user,
+        /magma/v1/user/:username,
+        /magma/v1/user/login,
```
hcgatewood: Just need the "user" prefix -- mirror service_registry.yml comments

hcgatewood (changes requested):
Looks great! The main thing is: the "type" is part of the policy, not the resource request

See the relevant comment. Getting this right is the last thing, then I think we'll be able to merge 🙂

orc8r/cloud/go/obsidian/access/rest_middleware.go
christinewang5: Sureee thing!

orc8r/cloud/go/obsidian/access/rest_middleware.go
christinewang5: Ig I am not sure how the certifier can determine that something in the path is a resource rather than just a normal path component. Would we not have to go through some registry of all endpoints to first check if an endpoint takes in network/tenant ID in its path param then do a bunch of parsing to actually get the id? Currently I'm getting this info from echo.Context, but gRPC it can only get context.Context, which I don't think retains the param info, but maybe I'm wrong, but yeah we can talk more tmrw.

orc8r/cloud/go/services/certifier/storage/storage_blobstore.go
christinewang5: Yeah this is kind of unfortunate. I used it in protos/utils and imported storage + storage imports protos, so duplicated it to break circular dependency, but then again it is also sketchy to put all these types into the protos pkg and use it in storage.

orc8r/cloud/go/services/certifier/tokens.go
christinewang5: Perhaps not really a good reason, but I just mimicked  the Github tokens and the say a bit more about it: https://github.blog/changelog/2021-03-31-authentication-token-format-updates-are-generally-available/. I'll just change to base64 though, seems more standard.

orc8r/cloud/go/services/certifier/tokens.go
christinewang5: Uhh I intended it to stand for orc8r personal token, though lmk if you have better ideas.

orc8r/cloud/go/services/certifier/tokens.go
christinewang5: Wdym here? I am unsure how to do type assertions for type aliases, i'll look into this later

orc8r/cloud/go/services/certifier/protos/utils.go
christinewang5: Cool, made the methods on the underlying proto types... 

How can we have a shared impl when they all have diff types and keys? this would be so nice if golang had generics, but idk how we would otherwise do this?

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
christinewang5: Oops forgot to return the response, though I'm just going to remove login handler, it is redundant to w `GetUserTokens`

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
christinewang5: Yup it does!!!

orc8r/cloud/go/services/certifier/tokens.go
christinewang5: Actually a good reason for it is that I use it is bc I split by "_" when checking the token header, and that is a base64 url safe character

orc8r/cloud/go/obsidian/reverse_proxy/reverse_proxy.go
christinewang5: Good point!

christinewang5:
Hey! Thanks @hcgatewood for all the thorough review and detailed comments! 🎉 I believe I have addressed majority of comments, and kept some comments unresolved.

As per our offline discussion, the fixes for network/tenant-scoped paths are here in a separate PR. https://github.com/magma/magma/pull/10757/

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -24,14 +24,29 @@ import (
...
+	endpoints map[string]bool
+}
+
+func InitNoBasicAuthEndpoints() NoBasicAuthEndpoints {
+	return NoBasicAuthEndpoints{endpoints: map[string]bool{
+		// Users should be able to access their own authentication information
+		"/magma/v1/user/login": true,
+	}}
+}
+
```
hcgatewood: This is weird lol. Just use a bare map

```go
unprotectedPaths := map[string]interface{}{}{
	...
}
```

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -43,226 +47,137 @@ func TestMiddlewareWithoutCertifier(t *testing.T) {
...
 
+	urlPrefix := fmt.Sprintf("http://%s", listener.Addr().String())
+	tests := []struct {
+		method   string
+		url      string
+		certSn   string
+		expected int
+	}{
+		// Test regular operator wildcard failures
+		{"GET", urlPrefix + RegisterNetworkV1 + "/" + TEST_NETWORK_ID, operCertSn, 200},
```
hcgatewood: Use fmt.sprintf for all these

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	if err := data.Validate(strfmt.Default); err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, err)
+	}
+	username := fmt.Sprintf("%v", *data.Username)
+	password := []byte(fmt.Sprintf("%v", *data.Password))
+	user := &protos.User{
+		Username: username,
+		Password: password,
+	}
```
hcgatewood, line 170: 1. Make the swagger username/password non-nil (since it's required), then you won't have to deal with ptrs
2. Make the proto password string, will be simpler (unless you had strong reason to make it bytes -- lmk if so)

orc8r/cloud/go/services/certifier/protos/certifier.proto
```diff
@@ -34,20 +38,141 @@ message CertificateInfoMap {
...
+  string resource = 4;
+}
+
+message ResourceList {
+  repeated Resource resources = 4;
+}
+
+message Policy {
+  string token = 1;
+  ResourceList resources = 4;
```
hcgatewood: Always number these sequentially + starting at 1, except with particular reason

orc8r/cloud/go/services/certifier/storage/storage_blobstore.go
hcgatewood: Can also potentially create a `const` pkg in certifier/const. I would almost try that. Up to you though

hcgatewood (approved):
Lgtm, merging! Added a couple comments to address in the next PR 🙂

hcgatewood:
Ah jk can't merge, looks like the cloud-test check is failing. Take a stab at fixing that, then msg me tmrw when you're ready for me to merge 👍👍

orc8r/cloud/go/obsidian/access/rest_middleware.go
christinewang5: Ok aha I was confused to how to make a package-level variable, apparently have to put `var` in front of it.

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
christinewang5: 1. It's already required... 🙃
2. Not a super strong reason but needed it to be bytes when using encrypting the password and either way I would have to convert things somewhere ig.

PR #10255, opened 2021-11-11
fix(ci/firebase) Always push to firebase. Even on failure

talkhasib:

## Summary

We need to push failed builds to Firebase database for visibility and tracking

## Test Plan

testing on ci_firebase_always branch

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -665,7 +665,7 @@ jobs:
           SLACK_FOOTER: ' '
   Publish_to_firebase:
     name: Publish to firebase
-    if: github.event_name == 'push'
+    if: always() && github.event_name == 'push'
     runs-on: ubuntu-latest
     needs: [agw-build]
     steps:


.github/workflows/build_all.yml
```diff
@@ -665,7 +666,7 @@ jobs:
           SLACK_FOOTER: ' '
   Publish_to_firebase:
     name: Publish to firebase
-    if: github.event_name == 'push'
+    if: always() && github.event_name == 'push'
     runs-on: ubuntu-latest
     needs: [agw-build]
```
quentinDERORY, line 670: remove the needs line here

.github/workflows/build_all.yml
```diff
@@ -665,7 +666,7 @@ jobs:
           SLACK_FOOTER: ' '
   Publish_to_firebase:
     name: Publish to firebase
-    if: github.event_name == 'push'
+    if: always() && github.event_name == 'push'
```
quentinDERORY, line 668: don't use always here

quentinDERORY (changes requested):
To push on failure you need to remove the needs link with agw-build that's all

.github/workflows/build_all.yml
talkhasib: I need it to wait for the AGW job. But always work regardless of exit state.

quentinDERORY (approved):
just remove the added branch trigger :)

PR #10259, opened 2021-11-11
feat: add possibility of singleton for singleton reindexer for prod

### LINKED ISSUE #10135: Make reindexer a singleton service

reginawang3495:
## Summary
add possibility of singleton for singleton reindexer for prod
tldr --> add `orc8r_worker` to prod

## Test Plan
Followed https://magma.github.io/magma/docs/next/orc8r/dev_minikube to deploy

## Additional Information


orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
```diff
@@ -60,6 +60,7 @@ func (r *reindexerSingleton) Run(ctx context.Context) {
 		}
 
 		clock.Sleep(reindexLoopInterval)
+		glog.Infof("Sleeping for %d seconds before looking for new jobs to reindex", reindexLoopInterval / time.Second)
```
hcgatewood: Nit: can just use the same time denomination as the var

hcgatewood (approved):
Lgtm! Can we add a doc.go in the services/orc8r_worker directory to overview the purpose of this service, as well as update the docs https://magma.github.io/magma/docs/next/orc8r/architecture_overview#orchestrator

hcgatewood:
Oh also let's make sure the relevant flag is *on* i.e. we want to be using singleton reindexing

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
reginawang3495: It's only 1 minute tho, so I feel like 60 seconds would be more descriptive. I'm down to change tho!

reginawang3495:
Ok, I was going to make another PR to flip it but i can put in this one

orc8r/cloud/go/services/orc8r_worker/doc.go
```diff
@@ -0,0 +1,20 @@
...
+limitations under the License.
+*/
+
+// Package orc8r_worker contains the orc8r_worker service.
+//
+// The orc8r_worker service is a singleton service that runs different classes.
+// It currently only runs the reindexer as a singleton.
+
+
+package orc8r_worker
```
hcgatewood: Attach the doc comments directly to the pkg
```suggestion
// Package orc8r_worker contains the orc8r_worker service.
//
// The orc8r_worker service is a singleton service that runs different classes.
// It currently only runs the reindexer as a singleton.
package orc8r_worker
```

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
hcgatewood: Is fine either way, my preference is simplicity of the codebase (it's lightly jarring to see the division -- makes me think "this is weird why are we dividing here" -- whereas just passing the var directly wouldn't set off any questions during a skim)

hcgatewood (commented):
1 request for the doc comment, then msg me to merge 👍

orc8r/cloud/go/services/state/indexer/reindex/reindexer_singleton.go
reginawang3495: Oh well in that case, I think this is better? Because there are 2 other options to not have division
1. Say that we sleeping for `60000000000` nanoseconds 
    - Unideal since that's hard to read
2. Hard code that we sleeping for a minute
    - Unideal since it's prone to human error --> someone might change the time, but forget to change the hard coding

hcgatewood: Either is fine. I prefer minutes since it's consistent with the same unit, but up to you

```
glog.Infof("...%v...", t.Seconds())
// or
glog.Infof("...%v...", t.Minutes())
```

reginawang3495: sure, i dont really mind either way

PR #10264, opened 2021-11-11
fix(agw): Updating import path for load test protos

ardzoht:

## Summary

- For AGW deployments on docker, magma folder path can be placed different from home path. This PR updates the import path of protos so it works for all deployments.

## Test Plan

- make load_test on VM
- running load tests individually on docker pods

## Additional Information


ulaskozat (approved):
LGTM

PR #9580, opened 2021-10-08
feat(agw): Added code to support deletion of dedicated bearer for roaming UE

rsarwad:
## Summary
feat(agw): Added code to support deletion of dedicated bearer for roaming UE
## Test Plan
Executed s1ap sanity test suite; one test case is failing, I will work on it

PR #10275, opened 2021-11-12
fix(mme): correct S8a proto generation for grpc_service

electronjoe:
Fix for #9580 `make build_oai` failure in devcontainer.

## Test Plan

Validated that `test build_oai` succeeds in GH Codespaces under this PR, while it fails without this PR on master).


ulaskozat (approved):
LGTM

Issue #10170, opened 2021-11-08
Shared Sentry mconfigs


hcgatewood:

Neudrino:
Is this issue still valid? Are there any open points here?

As far as one can see from https://github.com/magma/magma/issues/9508, this would be the last subtask for that bigger issue, right?

@christinewang5 @hcgatewood @themarwhal

themarwhal:
@Neudrino I think this can be closed now! Maybe a good follow up task is to migrate some of our lab setups to use the orc8r streamed config now.

PR #10276, opened 2021-11-12
chore(agw): Exclude message patterns in Sentry client

sebathomas:
## Summary

Currently, a lot of errors are excluded from monitoring in the
Sentry server, leading to 100,000s of filtered messages per day.
Instead, we add the possibility to configure such a filter in
the client. For a start, the client will use the same filters as
the server.

## Test Plan

The filter functionality was tested with a Sentry test account
using error logs, logged exceptions and explicit captured
messages.


lte/gateway/configs/control_proxy.yml
```diff
@@ -50,3 +50,17 @@ sentry_upload_mme_log: false
 # Rate at which we want to sample Python error events
 # Should be a number between 0 (0% of errors sent) and 1 (100% of errors sent)
 sentry_sample_rate: 1.0
+# Message patterns that are excluded from Sentry (regex substring match)
+sentry_excluded_errors:
```
themarwhal: @christinewang5 recently added the sentry config to the mconfig, which can be streamed down from the Orc8r. 
Let's add any new sentry configs there instead. https://github.com/magma/magma/blob/master/orc8r/protos/mconfig/mconfigs.proto#L25

if you have questions on the how the config gets streamed down from the orc8r, we can meet and talk through it. :) 
But the quick gist is:
1. The API model is defined here: https://github.com/magma/magma/blob/master/orc8r/cloud/go/services/orchestrator/obsidian/models/swagger.v1.yml#L1435
2. This gets generated into a go struct here: https://github.com/magma/magma/blob/master//orc8r/cloud/go/services/orchestrator/obsidian/models/network_sentry_config_swaggergen.go#L18
3. The values are passed onto the mconfig here: https://github.com/magma/magma/blob/master//orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go#L104
4. This struct then gets streamed down and is processed at https://github.com/magma/magma/tree/master/orc8r/gateway/python/magma/configuration
5. The file ends up in `/var/opt/magma/configs/gateway.mconfig` the default values are taken from `/etc/magma/gateway.mconfig`

^ This shared config to store config used by all services is a new concept, so you might have to add functions in the files in step4 to fetch this value.

sebathomas: I think I've done the first three steps but I'm unsure about 4 and 5. From what I can see, the Python Sentry configuration is still completely based on control_proxy.yml. Should this be changed to use mconfig here?

sebathomas: Another thing I've noticed is that the two URL fields are still called "URL" in the swagger interface but were renamed to "DSN" in the proto interface. Is that intentional?

themarwhal: Yes, we have to make the change to read from the mconfig. (So this will have to be a separate PR. We did something similar for C++ way back. We can take values from the Mconfig UNLESS it is overwritten by the control_proxy files.

sebathomas: Is #10170 the issue to do that?

themarwhal: https://github.com/magma/magma/issues/9508 is closer. We don't seem to have filed a task for modifying the python config part.

orc8r/protos/mconfig/mconfigs.proto
```diff
@@ -45,6 +45,8 @@ message SharedSentryConfig {
   // sample_rate should be a number between 0 (0% of errors sent) and 1 (100%
   // of errors sent)
   float sample_rate = 4;
+  // message patterns that are excluded from Sentry (regex substring match)
+  repeated string sentry_excluded_errors = 5;
```
themarwhal: We don't need the `sentry_` prefix as it's already under a message called sentry config. I think we can call it something like `exclusion_patterns`? :D

themarwhal:
If you rebase on top of https://github.com/magma/magma/pull/10321 after it merges, that should remove the NMS reviews

orc8r/protos/mconfig/mconfigs.proto
sebathomas: Changed it in both interfaces.

sebathomas:
> If you rebase on top of #10321 after it merges, that should remove the NMS reviews

Rebased.

lte/gateway/configs/control_proxy.yml
sebathomas: At the moment, we've added the new exclusion list to both mconfig and the control_proxy.yml. Since all the Sentry config is not yet used in the services, I would propose to keep it in both for this PR and then to have another PR where the Python services are changed to use the mconfig for Sentry.

Should we work on the shared Sentry config or is this already planned by someone else (e.g. @christinewang5 in #10170)?

themarwhal: It would be great if you could add another PR to add the code needed to read from shared mconfig. (Please let me know if you need guidance on this).

If the pattern exclusion feature needs to be done quickly, we can add it into control_proxy.yml. However, I am hesitant towards it as control_proxy.yml isn't supposed to be a place for sentry configurations. (I added it there as a quick solution when I initially built out the sentry configs, but probably should have contained it in a separate file.) So if we could avoid adding any more fields there, that would be good.

Neudrino:
Changed status back to draft as it is waiting for implementation of https://github.com/magma/magma/issues/9508 (and possibly https://github.com/magma/magma/issues/10170, too).

lte/gateway/configs/control_proxy.yml
alexzurbonsen: @themarwhal We created a draft (#10557) to get Sentry config from mconfig. Would be great to get some feedback! :)

sebathomas:
Rebased.

sebathomas:
Rebased on top of the mconfig change, and used the same logic for the exclusion patterns as for the other shared config options.

Still looking into the failed CodeQL CI result. Not sure if it is related to these changes.

lte/gateway/configs/gateway.mconfig
```diff
@@ -159,7 +159,21 @@
         "dsn_python": "",
         "dsn_native": "",
         "upload_mme_log": false,
-        "sample_rate": 1.0
+        "sample_rate": 1.0,
+        "exclusion_patterns": [
+          "GetServiceInfo",
```
themarwhal, line 143: This value should match the default value from the API northbound side. (I'm not sure if there is a way to specify a default value here, though).

If the on-device gateway mconfig value differs from the default cloud streamed version, that would lead to an unnecessary restart of all services.

lte/gateway/configs/control_proxy.yml
themarwhal: Should we just not have it configurable here? I think it would be good to fully migrate to the Mconfig for this.

sebathomas:
Rebased.

lte/gateway/configs/gateway.mconfig
sebathomas: I've added it as a default in swagger and confirmed that the default list is used if the PUT method is called without the exclusion list.

lte/gateway/configs/control_proxy.yml
sebathomas: Will remove it again.

sebathomas: Added a third commit to remove the patterns from control_proxy.yml.

orc8r/cloud/go/services/orchestrator/obsidian/models/swagger.v1.yml
```diff
@@ -1457,6 +1457,28 @@ definitions:
...
+        type: array
+        items:
+          type: string
+        minItems: 0
+        description: log message patterns that are excluded (regex substring match)
+    default:
+      upload_mme_log: false
+      sample_rate: 1
+      exclusion_patterns:
+        - 'GetServiceInfo'
```
themarwhal, line 1470: Changes look good! Last comment, I think. Is the intention here to eventually reduce these default values? In that case, I think it would be good to file an issue (and link it here in a comment)  for to migrate some of these issues to be handled directly.

themarwhal: And finally, please add this configuration to the sentry doc. (can be in a different PR)

sebathomas: Documentation: #11210

sebathomas: I agree, the goal should also be to reduce the list and handle the corresponding errors in other ways. Added a task to track that: #11211

themarwhal (approved):

themarwhal:
@sebathomas conflict with the other change i think :p

sebathomas:
> @sebathomas conflict with the other change i think :p

Rebased and regenerated mconfigs.pb.go... :)

PR #10278, opened 2021-11-12

spikey979:
fix(feg): fix incorrect conversion between integer types

## Summary
- feg/gateway/tools/gw_csfb_service_cli/main.go:164 -> main()
   String was parsed into an int which can cause incorrect conversion  if int is converted into another integer type of a smaller size. Added check if variable is in bounds of Unsigned32 type (0-4294967295)

- feg/gateway/tools/s6a_cli/main.go:324 -> getPlmnID()
   String was parsed into an int which can cause incorrect conversion  if int is converted into another integer type of a smaller size. Added check if variable is in bounds of byte type (0-255)

- feg/gateway/tools/swx_cli/main.go:240 -> sendMar()
   String was parsed into an int which can cause incorrect conversion  if int is converted into another integer type of a smaller size. Added check if variable is in bounds of Unsigned32 type (0-4294967295)

Fixes are for alerts:
https://github.com/magma/magma/security/code-scanning/16?query=ref%3Arefs%2Fheads%2Fmaster
https://github.com/magma/magma/security/code-scanning/17?query=ref%3Arefs%2Fheads%2Fmaster
https://github.com/magma/magma/security/code-scanning/18?query=ref%3Arefs%2Fheads%2Fmaster

### DIFF
--- a/feg/gateway/tools/swx_cli/main.go
+++ b/feg/gateway/tools/swx_cli/main.go
@@ -235,6 +235,10 @@ func sendSar(addr string, client swxClient) int {
 }
 
 func sendMar(addr string, client swxClient) int {
+	if numVectors < 0 || numVectors > 4294967295 {
+		fmt.Printf("numVectors %d is outside the bounds of unit32 type", numVectors)
+		return 2
+	}
 	req := &protos.AuthenticationRequest{
 		UserName:             imsi,
 		SipNumAuthVectors:    uint32(numVectors),
--- a/feg/gateway/tools/hello_cli/main.go
+++ b/feg/gateway/tools/hello_cli/main.go
@@ -45,6 +45,9 @@ func main() {
 	defer conn.Close()
 	cl := protos.NewHelloClient(conn)
 	fmt.Printf("Sending  Greeting: '%s', Code: %d\n", msg, code)
+	if code < 0 || code > 4294967295 {
+		log.Fatal(fmt.Errorf("Code %d is outside the bounds of unit32 type", code))
+	}
 	res, err := cl.SayHello(context.Background(), &protos.HelloRequest{Greeting: msg, GrpcErrCode: uint32(code)})
 	if err != nil {
 		log.Fatal(err)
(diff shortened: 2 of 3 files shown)


themarwhal (approved):

feg/gateway/tools/swx_cli/main.go
```diff
@@ -235,6 +235,10 @@ func sendSar(addr string, client swxClient) int {
 }
 
 func sendMar(addr string, client swxClient) int {
+	if numVectors < 0 || numVectors > 4294967295 {
```
electronjoe, line 238: So for some reason this PR did not go through the feg-workflow. Any ideas why not e.g. @quentinDERORY?

feg-workflow linting fails on this line because `numVectors` is a unsigned integer and cannot be less than zero.

Presently in Master, this blocks all PRs that execute feg-workflow (a required workflow when run).  See [feg-workflow failure of my PR](https://github.com/magma/magma/pull/10415).  I'll send out a PR now to modify this and make master healthy.  No fault of yours @spikey979 - I don't know why the feg-worfklow did not run (I would not expect contributors, internal or other, to know all the tests that need to run for a PR - that is what CI is for).

spikey979: I fixed this in PR: https://github.com/magma/magma/pull/10372

electronjoe: I've also posted a fix as of minutes ago - https://github.com/magma/magma/pull/10416 - happy to take whichever, though I have an additional minor nit fixup in my PR (using Golang builtin to detect out of range). Mind if we land mine? Thanks for the prompt fixup PR - we probably should have made this visible / fast tracked it since it broke master but I think nobody knew.

electronjoe: Still wondering why the `feg-worfklow` did not run on this PR - which let us get into a bad-master state. Hopefully DevOps folks have an idea there (e.g. @tmdzk or @quentinDERORY )

quentinDERORY: It did run, but the tests failed so early it did not outputed xml outputs. And that's why the comment job is showing no tests runs.
When merging jobs recently I introduced a bug where steps were marked as successful when they were not. (This is why master went into this bad state for feg workflow )
I just resolved it yesterday and now the status of the steps will be displayed correctly.
For the tests not run, we need to improve the comment job to display this specific corner case.

quentinDERORY: created issue for feg tests showing not being run: https://github.com/magma/magma/issues/10421

PR #10280, opened 2021-11-12
chore(magmad): reduce loglevel if any service is unreachable

alexzurbonsen:
## Summary

Context: This is one of the frequent sentry errors that are currently filtered on server-side. See #10027.

magmad periodically polls the status of other services via GetServiceInfo and keeps a counter of timeouts. It's not necessary to log at level error every time such a call fails.

### DIFF
--- a/orc8r/gateway/python/magma/magmad/service_poller.py
+++ b/orc8r/gateway/python/magma/magmad/service_poller.py
@@ -177,7 +177,7 @@ class ServicePoller(Job):
                 )
                 self._service_info[service].continuous_timeouts = 0
             except grpc.RpcError as err:
-                logging.error(
+                logging.warning(
                     "GetServiceInfo Error for %s! [%s] %s",
                     service,
                     err.code(),


alexzurbonsen:
There is a similar case for `GetOperationalStates` which currently is also filtered on server-side (in magma/magmad/state_reporter.py)

In `_get_operational_states` an error will be logged if any service is not reachable. However all exceptions are caught, not only grpc errors. @themarwhal Should I reduce the log level here as well?

orc8r/gateway/python/magma/magmad/service_poller.py
```diff
@@ -177,7 +177,7 @@ async def _get_service_info(self):
                 )
                 self._service_info[service].continuous_timeouts = 0
             except grpc.RpcError as err:
-                logging.error(
+                logging.warning(
```
themarwhal, line 180: Shall we keep the level at error but just not send it to sentry? cc @hcgatewood

hcgatewood: Correct. This is indeed an error (under the vanilla topology that assumes a constant Orc8r<>AGW connection), but it's not indicative that there's a bug. So let's keep at error but e.g. not send to sentry. Same goes for the other instances of inability to connect to orc8r

2 types of errors: application, platform. This is an *application* error, because something the operator's trying to do isn't working. If, on the other hand, the bootstrapper server crashed, violated an invariant, etc, then that would be a *platform* error

This whole distinction^ arises from the fact that we're providing Magma as a platform, that others then run applications on top of. So we principally care about errors that indicate a bug/breakage of the platform (for Sentry, at least)

alexzurbonsen: Thanks for explaining! I created a PR (#10349) to filter those errors via opt-in logging to Sentry. I will close this one.

PR #10281, opened 2021-11-12
test(spgw): Add injection of state loaded in SPGW state manager (issue #9606) (WIP)

thhung:
## Summary

Adding injection of state loaded in SPGW state manager by adding a method to "load" predefined state on the MME task - SPGW, mocking the read_ue_state_from_db() function. 

## Test Plan
This change should pass the tests in test_spgw_procedure

## Additional Information


lte/gateway/c/core/oai/test/spgw_task/mock_spgw_op.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+
+namespace magma {
+namespace lte {
+
+int mock_read_spgw_ue_state_db() {
+  int number_of_samples = 1;
+  for (int i = 0; i < number_of_samples; ++i) {
+    oai::SpgwUeContext ue_proto = oai::SpgwUeContext();
+    std::string name_of_sample = "/home/vagrant/magma/lte/gateway/c/core/oai/test/spgw_task/data/sample.bin.v6";
```
ardzoht: can this be refactored? instead of being hardcoded this can be a parameter for `mock_read_spgw_ue_state_db` function so it can be configured for different type of data.

lte/gateway/c/core/oai/test/spgw_task/mock_spgw_op.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "spgw_state.h"
+
+#include <google/protobuf/text_format.h>
+
+
+namespace magma {
+namespace lte {
+
+int mock_read_spgw_ue_state_db() {
+  int number_of_samples = 1;
```
ardzoht:
```suggestion
  int number_of_ue_samples = 1;
```

lte/gateway/c/core/oai/test/spgw_task/mock_spgw_op.cpp
```diff
@@ -0,0 +1,35 @@
...
+    
+  
+  spgw_ue_context_t* ue_context_p =
+      (spgw_ue_context_t*) calloc(1, sizeof(spgw_ue_context_t));
+  SpgwStateConverter::proto_to_ue(ue_proto, ue_context_p);
+  }
+  return 0;
+}
+
+int mock_spgw_app_init(){
```
ardzoht: I'd recommend to run the steps on https://docs.magmacore.org/docs/next/lte/dev_unit_testing#format-cc to take care of the formatting comments

ardzoht: check `spgw_app_init` definition on `lte/gateway/c/core/oai/tasks/sgw/sgw_task.c`, the error you were facing on other unit tests it's likely due to the missing calls of: `spgw_state_init` and `itti_create_task`.

my recommendation would be: to not mock `spgw_app_init` and call it as it's necessary to setup state and ITTI task mentioned above, but rather calling `mock_read_spgw_ue_state_db` as part of `SetUp()` on the tests

lte/gateway/c/core/oai/test/spgw_task/test_spgw_procedures_with_injected_state.cpp
```diff
@@ -0,0 +1,149 @@
...
+    init_task_context(
+        TASK_MAIN, task_id_list, 2, handle_message, &task_zmq_ctx_main_spgw);
+
+    std::thread task_mme_app(start_mock_mme_app_task, mme_app_handler);
+    task_mme_app.detach();
+
+    std::cout << "Running setup" << std::endl;
+    // initialize the SPGW task
+    // might need to mock this function
+    spgw_state_init(false, &spgw_config);
```
ardzoht: As mentioned above, let's call `spgw_app_init` instead of mocking it, and call `mock_read_spgw_ue_state_db` separately after initializing SPGW app.

ardzoht (changes requested):
Left some comments regarding formatting and mocking of SPGW app init

lte/gateway/c/core/oai/test/spgw_task/mock_spgw_op.cpp
thhung: Thanks, I think this is a good idea. After verifying this version, I will refactor it.

lte/gateway/c/core/oai/test/spgw_task/data/data_list.txt
```diff
@@ -0,0 +1,100 @@
+/home/vagrant/magma/lte/gateway/c/core/oai/test/spgw_task/data/IMSI001010000000066_SPGW
```
ardzoht: since `/home/vagrant/magma/lte/gateway/c/core/oai/test/spgw_task/data/` is constant, how about just define only the name of the files here? and then pass and append the folder path on the cpp file instead?

lte/gateway/c/core/oai/test/spgw_task/mock_spgw_op.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+int mock_read_spgw_ue_state_db(const std::vector<std::string>& ue_samples) {
+  for (auto name_of_sample : ue_samples) {
+    oai::SpgwUeContext ue_proto = oai::SpgwUeContext();
+    std::fstream input(name_of_sample.c_str(), std::ios::in | std::ios::binary);
+    if (!ue_proto.ParseFromIstream(&input)) {
+      std::cerr << "Failed to parse the sample: " << name_of_sample
+                << std::endl;
+      return -1;
```
ardzoht:
```suggestion
      return RETURNerror;
```

lte/gateway/c/core/oai/test/spgw_task/mock_spgw_op.cpp
```diff
@@ -0,0 +1,48 @@
...
+    }
+  } else {
+    std::cerr << "couldn't open file: " << file_name << std::endl;
+  }
+
+  file_content.close();
+  return vector_of_lines;
+}
+
+int mock_read_spgw_ue_state_db(const std::vector<std::string>& ue_samples) {
```
ardzoht:
```suggestion
status_code_e mock_read_spgw_ue_state_db(const std::vector<std::string>& ue_samples) {
```

lte/gateway/c/core/oai/test/spgw_task/mock_spgw_op.cpp
```diff
@@ -0,0 +1,48 @@
...
+      std::cerr << "Failed to parse the sample: " << name_of_sample
+                << std::endl;
+      return -1;
+    }
+
+    spgw_ue_context_t* ue_context_p =
+        (spgw_ue_context_t*) calloc(1, sizeof(spgw_ue_context_t));
+    SpgwStateConverter::proto_to_ue(ue_proto, ue_context_p);
+  }
+  return 0;
```
ardzoht:
```suggestion
  return RETURNok;
```

lte/gateway/python/scripts/state_cli.py
```diff
@@ -194,6 +194,14 @@ def _parse_state_proto(self, key_type, value):
             deserializer = get_proto_deserializer(proto)
             version_deserializer = get_proto_version_deserializer()
             print(deserializer(value))
+            v = deserializer(value)
+            print(v)
+            # print(type(v))
+            # Write the new address book back to disk.
+            print(">>>>>>> save proto file")
```
ardzoht: are these changes needed?

lte/gateway/python/integ_tests/s1aptests/test_attach_detach.py
```diff
@@ -54,6 +56,7 @@ def test_attach_detach(self):
                 "************************* Running UE detach for UE id ",
                 req.ue_id,
             )
+            time.sleep(300)
```
ardzoht: remove this?

lte/gateway/c/core/oai/test/spgw_task/spgw_test_util.cpp
```diff
@@ -42,8 +43,12 @@ bool is_num_sessions_valid(
 
   // If one UE context exists, check that teids also exist in hashtable
   spgw_ue_context_t* ue_context_p = spgw_get_ue_context(imsi64);
-  int num_teids                   = 0;
-  sgw_s11_teid_t* s11_teid_p      = nullptr;
+  if (expected_num_teids == 0) {
```
ardzoht: can you add a comment to explain the if case here?

ardzoht (changes requested):
left some minor comments, thanks for the changes, this is looking ready to be merged!

lte/gateway/python/integ_tests/s1aptests/test_attach_detach.py
thhung: Yes, it should be removed.

lte/gateway/python/scripts/state_cli.py
thhung: No I don't think so. I will remove it.

lte/gateway/c/core/oai/test/spgw_task/data/data_list.txt
thhung: This is a good idea! My intention at that time was to allow more dynamic way to specify the path of samples. But if it is not necessary, we can do as your suggestions.

lte/gateway/c/core/oai/test/spgw_task/spgw_test_util.cpp
ssanadhya: @thhung ,  we need two different validation functions here:
-- `is_ue_teids_valid(unit64_t imsi64, int expected_num_ue_teids)` to check if UE context is non-null and tunnel count is valid for a single UE
-- `is_num_sessions_valid(int expected_num_ue_contexts)` to check conditions for multiple UEs 

Note that on latest master, I have removed the `spgw_state` parameter from this function as it is unused

PR #10298, opened 2021-11-13
feat(feg): N7 Proxy SM Policy Create translations

GANESH-WAVELABS:
## Summary
- Converts to convert from CreateSessionRequest proto to SmPolicyContextData to be sent to N7 interface
- Converts to convert from N7 SmPolicyDecision to CreateSessionResponse proto to be sent to SessionD

### DIFF
--- a/feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
+++ b/feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
@@ -1,0 +1,585 @@
+/*
+Copyright 2021 The Magma Authors.
+
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package n7
+
+import (
+	b64 "encoding/base64"
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/golang/protobuf/ptypes/timestamp"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"magma/feg/gateway/policydb"
+	common5g "magma/feg/gateway/sbi/specs/TS29122CommonData"
+	n7_sbi "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
+	sbi "magma/feg/gateway/sbi/specs/TS29571CommonData"
+	"magma/lte/cloud/go/protos"
+)
+
+//
+// From Proto to SBI types
+//
+
+func GetSmPolicyContextDataN7(
+	request *protos.CreateSessionRequest,
+	notifyApiRoot string,
+) *n7_sbi.PostSmPoliciesJSONRequestBody {
+	imsi := removeIMSIPrefix(request.GetCommonContext().GetSid().Id)
+	common := request.GetCommonContext()
+	ratType := common.GetRatType()
+
+	reqBody := &n7_sbi.PostSmPoliciesJSONRequestBody{
+		Supi:              sbi.Supi(imsi),
+		Ipv4Address:       getSbiIpv4(common.GetUeIpv4()),
+		Ipv6AddressPrefix: getSbiIpv6(common.GetUeIpv6()),
+		Dnn:               sbi.Dnn(common.GetApn()),
+		Gpsi:              getSbiGpsi(string(common.GetMsisdn())),
+		RatType:           getSbiRatType(ratType),
+		AccessType:        getSbiAccessType(ratType),
+		UeTimeZone:        GetSbiTimeZone(request.GetAccessTimezone()),
+		NotificationUri:   GenNotifyUrl(notifyApiRoot, request.SessionId),
+	}
+	if request.RatSpecificContext != nil {
+		ratSpecific := request.GetRatSpecificContext().GetContext()
+		switch context := ratSpecific.(type) {
+		case *protos.RatSpecificContext_M5GsmSessionContext:
+			m5gCtx := context.M5GsmSessionContext
+			reqBody.PduSessionId = sbi.PduSessionId(m5gCtx.GetPduSessionId())
+			reqBody.Gpsi = getSbiGpsi(m5gCtx.GetGpsi())
+			reqBody.PduSessionType = getSbiPduSesionType(m5gCtx.GetPduSessionType())
+		}
+	}
+
+	return reqBody
+}
+
+func getSbiRatType(ratType protos.RATType) *sbi.RatType {
+	var sbiRatType sbi.RatType
+	switch ratType {
+	case protos.RATType_TGPP_LTE:
+		sbiRatType = sbi.RatTypeEUTRA
+	case protos.RATType_TGPP_NR:
+		sbiRatType = sbi.RatTypeNR
+	case protos.RATType_TGPP_WLAN:
+		sbiRatType = sbi.RatTypeWLAN
+	default:
+		return nil
+	}
+	return &sbiRatType
+}
+
+func getSbiAccessType(ratType protos.RATType) *sbi.AccessType {
+	var accessType sbi.AccessType
+	switch ratType {
+	case protos.RATType_TGPP_LTE:
+		accessType = sbi.AccessTypeN3GPPACCESS
+	case protos.RATType_TGPP_NR:
+		accessType = sbi.AccessTypeN3GPPACCESS
+	case protos.RATType_TGPP_WLAN:
+		accessType = sbi.AccessTypeNON3GPPACCESS
+	default:
+		return nil
+	}
+	return &accessType
+}
+
+func GetSbiTimeZone(tzProto *protos.Timezone) *sbi.TimeZone {
+	if tzProto == nil {
+		return nil
+	}
+	offsetMin := tzProto.GetOffsetMinutes()
+	isNegative := false
+	if offsetMin < 0 {
+		isNegative = true
+		offsetMin = -offsetMin
+	}
+	tz := fmt.Sprintf("%02d:%02d", offsetMin/60, offsetMin%60)
+	if isNegative {
+		tz = "-" + tz
+	} else {
+		tz = "+" + tz
+	}
+	tzStr := sbi.TimeZone(tz)
+	return &tzStr
+}
+
+func getSbiPduSesionType(pduSessionType protos.PduSessionType) sbi.PduSessionType {
+	switch pduSessionType {
+	case protos.PduSessionType_IPV4:
+		return sbi.PduSessionTypeIPV4
+	case protos.PduSessionType_IPV4IPV6:
+		return sbi.PduSessionTypeIPV4V6
+	case protos.PduSessionType_IPV6:
+		return sbi.PduSessionTypeIPV6
+	case protos.PduSessionType_UNSTRUCTURED:
+		return sbi.PduSessionTypeUNSTRUCTURED
+	default:
+		return sbi.PduSessionTypeIPV4
+	}
+}
+
+// From SBI types to proto
+
+func GetCreateSessionResponseProto(
+	request *protos.CreateSessionRequest,
+	smPolicyDecision *n7_sbi.SmPolicyDecision,
+	smPolicyUrl string,
+) *protos.CreateSessionResponse {
+	staticRules, dynamicRules, _ := getPccRules(smPolicyDecision)
+	eventTriggers, revalidationTime := getEventTriggersInfoProto(
+		smPolicyDecision.PolicyCtrlReqTriggers, smPolicyDecision.RevalidationTime)
+
+	tgppCtx := &protos.TgppContext{GxDestHost: smPolicyUrl}
+	monitors := getUsageMonitorsProto(request, tgppCtx, eventTriggers, revalidationTime, smPolicyDecision)
+
+	rpcResp := &protos.CreateSessionResponse{
+		StaticRules:      staticRules,
+		DynamicRules:     dynamicRules,
+		TgppCtx:          tgppCtx,
+		SessionId:        request.SessionId,
+		EventTriggers:    eventTriggers,
+		RevalidationTime: revalidationTime,
+		UsageMonitors:    monitors,
+	}
+	if smPolicyDecision.Online != nil {
+		rpcResp.Online = *smPolicyDecision.Online
+	}
+	if smPolicyDecision.Offline != nil {
+		rpcResp.Offline = *smPolicyDecision.Offline
+	}
+	return rpcResp
+}
+
+func getPccRules(
+	policyDecision *n7_sbi.SmPolicyDecision,
+) ([]*protos.StaticRuleInstall, []*protos.DynamicRuleInstall, []string) {
+	var (
+		staticRules   []*protos.StaticRuleInstall
+		dynamicRules  []*protos.DynamicRuleInstall
+		rulesToRemove []string
+	)
+
+	if policyDecision.PccRules == nil {
+		return staticRules, dynamicRules, rulesToRemove
+	}
+
+	for ruleId, pccRule := range policyDecision.PccRules.AdditionalProperties {
+		if pccRule.PccRuleId == "" {
+			// This rule to be removed
+			rulesToRemove = append(rulesToRemove, ruleId)
+			continue
+		}
+		activationTime, deactivationTime := getActivateDeactivationTime(policyDecision, &pccRule)
+		policyRule := getPolicyRuleProto(policyDecision, &pccRule)
+		if policyRule == nil {
+			// This is a static rule. Static rule only has pccRuleId and RefCondData
+			staticRule := &protos.StaticRuleInstall{
+				RuleId:           ruleId,
+				ActivationTime:   activationTime,
+				DeactivationTime: deactivationTime,
+			}
+			staticRules = append(staticRules, staticRule)
+			continue
+		}
+
+		dynamicRule := &protos.DynamicRuleInstall{
+			PolicyRule:       policyRule,
+			ActivationTime:   activationTime,
+			DeactivationTime: deactivationTime,
+		}
+		dynamicRules = append(dynamicRules, dynamicRule)
+	}
+
+	return staticRules, dynamicRules, rulesToRemove
+}
+
+func getActivateDeactivationTime(
+	policyDecision *n7_sbi.SmPolicyDecision,
+	pccRule *n7_sbi.PccRule,
+) (activationTime, deactivationTime *timestamp.Timestamp) {
+	if pccRule.RefCondData == nil || *pccRule.RefCondData == "" {
+		return
+	}
+	condData, found := policyDecision.Conds.Get(*pccRule.RefCondData)
+	if !found {
+		return
+	}
+	activationTime = ConvertToProtoTimeStamp(condData.ActivationTime)
+	deactivationTime = ConvertToProtoTimeStamp(condData.DeactivationTime)
+	return
+}
+
+func getPolicyRuleProto(
+	policyDecision *n7_sbi.SmPolicyDecision,
+	pccRule *n7_sbi.PccRule,
+) *protos.PolicyRule {
+	// For all Ref*Data Maximum one entry is expected.
+	// These are declared as arrays for future compatibility
+	policyRule := &protos.PolicyRule{
+		Id:            pccRule.PccRuleId,
+		Priority:      getSbiUinteger(pccRule.Precedence),
+		MonitoringKey: getMonKeyFromPccRule(policyDecision, pccRule),
+		Redirect:      getTcDataFromPccRule(policyDecision, pccRule),
+		Qos:           getQosDataFromPccRule(policyDecision, pccRule),
+		FlowList:      getFlowListFromPccRule(pccRule),
+	}
+	insertChargingDataToPolicyRule(policyDecision, pccRule, policyRule)
+	policyRule.TrackingType = getTrackingTypeProto(len(policyRule.MonitoringKey) != 0, policyRule.RatingGroup != 0)
+
+	if len(policyRule.MonitoringKey) == 0 && policyRule.Redirect == nil &&
+		policyRule.Qos == nil && len(policyRule.FlowList) == 0 && policyRule.RatingGroup == 0 {
+		// must be a static rule if these are not present
+		return nil
+	}
+	return policyRule
+}
+
+func insertChargingDataToPolicyRule(
+	policyDecision *n7_sbi.SmPolicyDecision,
+	pccRule *n7_sbi.PccRule,
+	policyRule *protos.PolicyRule,
+) {
+	if pccRule.RefChgData == nil || len(*pccRule.RefChgData) == 0 || policyDecision.ChgDecs == nil {
+		return
+	}
+	chgData, found := policyDecision.ChgDecs.Get((*pccRule.RefChgData)[0])
+	if !found {
+		return
+	}
+	policyRule.RatingGroup = uint32(*chgData.RatingGroup)
+	policyRule.ServiceIdentifier = &protos.ServiceIdentifier{
+		Value: uint32(*chgData.ServiceId),
+	}
+	policyRule.Offline = getSbiBool(chgData.Offline)
+	policyRule.Online = getSbiBool(chgData.Online)
+}
+
+func getMonKeyFromPccRule(
+	policyDecision *n7_sbi.SmPolicyDecision,
+	pccRule *n7_sbi.PccRule,
+) []byte {
+	if pccRule.RefUmData != nil && len(*pccRule.RefUmData) > 0 && policyDecision.UmDecs != nil {
+		umData, found := policyDecision.UmDecs.Get((*pccRule.RefUmData)[0])
+		if found {
+			return []byte(umData.UmId)
+		}
+	}
+	return []byte{}
+}
+
+// getTcDataFromPccRule extracts Traffic Control data (Redirect Information) from the SmPolicyDecision
+// and returns a proto Redirect Information
+func getTcDataFromPccRule(
+	policyDecision *n7_sbi.SmPolicyDecision,
+	pccRule *n7_sbi.PccRule,
+) *protos.RedirectInformation {
+	if pccRule.RefTcData != nil && len(*pccRule.RefTcData) > 0 && policyDecision.TraffContDecs != nil {
+		tcData, found := policyDecision.TraffContDecs.Get((*pccRule.RefTcData)[0])
+		if found {
+			return getRedirectInfoProto(tcData.RedirectInfo)
+		}
+	}
+	return nil
+}
+
+func getQosDataFromPccRule(
+	policyDecision *n7_sbi.SmPolicyDecision,
+	pccRule *n7_sbi.PccRule,
+) *protos.FlowQos {
+	if pccRule.RefQosData != nil && len(*pccRule.RefQosData) > 0 && policyDecision.QosDecs != nil {
+		qosData, found := policyDecision.QosDecs.Get((*pccRule.RefQosData)[0])
+		if found {
+			return getFlowQosProto(&qosData)
+		}
+	}
+	return nil
+}
+
+func getFlowListFromPccRule(pccRule *n7_sbi.PccRule) []*protos.FlowDescription {
+	var flowList []*protos.FlowDescription
+	if pccRule.FlowInfos == nil {
+		return flowList
+	}
+	for _, flowInfo := range *pccRule.FlowInfos {
+		if flowInfo.FlowDescription == nil {
+			continue
+		}
+		flow, err := policydb.GetFlowDescriptionFromFlowString(string(*flowInfo.FlowDescription))
+		if err != nil {
+			glog.Errorf("Could not get flow for description %s : %s", *flowInfo.FlowDescription, err)
+			continue
+		}
+		flowList = append(flowList, flow)
+	}
+	return flowList
+}
+
+func getTrackingTypeProto(monKeyPresent bool, ratingGrpPresent bool) protos.PolicyRule_TrackingType {
+	if monKeyPresent && ratingGrpPresent {
+		return protos.PolicyRule_OCS_AND_PCRF
+	} else if monKeyPresent && !ratingGrpPresent {
+		return protos.PolicyRule_ONLY_PCRF
+	} else if !monKeyPresent && ratingGrpPresent {
+		return protos.PolicyRule_ONLY_OCS
+	} else {
+		return protos.PolicyRule_NO_TRACKING
+	}
+}
+
+func getRedirectInfoProto(redirInfo *n7_sbi.RedirectInformation) *protos.RedirectInformation {
+	redirProto := &protos.RedirectInformation{}
+	if redirInfo == nil {
+		return redirProto
+	}
+	if redirInfo.RedirectEnabled != nil {
+		redirProto.Support = protos.RedirectInformation_DISABLED
+		if *redirInfo.RedirectEnabled {
+			redirProto.Support = protos.RedirectInformation_ENABLED
+		}
+	}
+	if redirInfo.RedirectAddressType != nil {
+		redirProto.AddressType = getRedirectAddrTypeProto(redirInfo.RedirectAddressType)
+	}
+	if redirInfo.RedirectServerAddress != nil {
+		redirProto.ServerAddress = *redirInfo.RedirectServerAddress
+	}
+
+	return redirProto
+}
+
+func getRedirectAddrTypeProto(
+	redirAddrType *n7_sbi.RedirectAddressType,
+) protos.RedirectInformation_AddressType {
+	if redirAddrType == nil {
+		return protos.RedirectInformation_IPv4
+	}
+	switch *redirAddrType {
+	case n7_sbi.RedirectAddressTypeIPV4ADDR:
+		return protos.RedirectInformation_IPv4
+	case n7_sbi.RedirectAddressTypeIPV6ADDR:
+		return protos.RedirectInformation_IPv6
+	case n7_sbi.RedirectAddressTypeSIPURI:
+		return protos.RedirectInformation_SIP_URI
+	case n7_sbi.RedirectAddressTypeURL:
+		return protos.RedirectInformation_URL
+	default:
+		glog.Error("Unknown Redirect Address Type, defaulting to IPv4(default value)")
+		return protos.RedirectInformation_IPv4
+	}
+}
+
+func getFlowQosProto(qosData *n7_sbi.QosData) *protos.FlowQos {
+	if qosData == nil {
+		return nil
+	}
+	return &protos.FlowQos{
+		MaxReqBwDl: bitRateStringToUint32(qosData.MaxbrDl),
+		MaxReqBwUl: bitRateStringToUint32(qosData.MaxbrUl),
+		GbrUl:      bitRateStringToUint32(qosData.GbrUl),
+		GbrDl:      bitRateStringToUint32(qosData.GbrDl),
+		Qci:        n5qiToQci(qosData.N5qi),
+	}
+}
+
+func getEventTriggersInfoProto(
+	triggers *[]n7_sbi.PolicyControlRequestTrigger,
+	revalTimeout *time.Time,
+) ([]protos.EventTrigger, *timestamp.Timestamp) {
+	var (
+		protoRevalidationTime *timestamp.Timestamp
+		protoTriggers         []protos.EventTrigger
+	)
+	if triggers == nil {
+		return protoTriggers, protoRevalidationTime
+	}
+	for _, trigger := range *triggers {
+		switch trigger {
+		case n7_sbi.PolicyControlRequestTriggerRETIMEOUT:
+			protoTriggers = append(protoTriggers, protos.EventTrigger_REVALIDATION_TIMEOUT)
+			protoRevalidationTime = ConvertToProtoTimeStamp(revalTimeout)
+		}
+	}
+	return protoTriggers, protoRevalidationTime
+}
+
+func getUsageMonitorsProto(
+	request *protos.CreateSessionRequest,
+	tgppCtx *protos.TgppContext,
+	eventTriggers []protos.EventTrigger,
+	revalidationTime *timestamp.Timestamp,
+	smPolicyDecision *n7_sbi.SmPolicyDecision,
+) []*protos.UsageMonitoringUpdateResponse {
+	var monitors []*protos.UsageMonitoringUpdateResponse
+	// Usage monitorting at pcc rule level
+	if smPolicyDecision.PccRules != nil {
+		for _, pccRule := range smPolicyDecision.PccRules.AdditionalProperties {
+			if pccRule.RefUmData == nil || len((*pccRule.RefUmData)[0]) == 0 {
+				continue
+			}
+			umEntry, found := smPolicyDecision.UmDecs.Get((*pccRule.RefUmData)[0])
+			if !found {
+				continue
+			}
+			umResp := newUsageMonitoringUpdateResponse(request, tgppCtx, eventTriggers, revalidationTime)
+			umResp.Credit = getUsageMonitoringCreditsProto((*pccRule.RefUmData)[0], &umEntry, protos.MonitoringLevel_PCC_RULE_LEVEL)
+			monitors = append(monitors, umResp)
+		}
+	}
+	// Usage monitoring at session level
+	if smPolicyDecision.SessRules != nil {
+		for _, sessRule := range smPolicyDecision.SessRules.AdditionalProperties {
+			if sessRule.RefUmData == nil || len(*sessRule.RefUmData) == 0 {
+				continue
+			}
+			umEntry, found := smPolicyDecision.UmDecs.Get(*sessRule.RefUmData)
+			if !found {
+				continue
+			}
+			umResp := newUsageMonitoringUpdateResponse(request, tgppCtx, eventTriggers, revalidationTime)
+			umResp.Credit = getUsageMonitoringCreditsProto(*sessRule.RefUmData, &umEntry, protos.MonitoringLevel_SESSION_LEVEL)
+			monitors = append(monitors, umResp)
+		}
+	}
+
+	return monitors
+}
+
+func newUsageMonitoringUpdateResponse(
+	request *protos.CreateSessionRequest,
+	tgppCtx *protos.TgppContext,
+	eventTriggers []protos.EventTrigger,
+	revalidationTime *timestamp.Timestamp,
+) *protos.UsageMonitoringUpdateResponse {
+	return &protos.UsageMonitoringUpdateResponse{
+		SessionId:        request.SessionId,
+		TgppCtx:          tgppCtx,
+		Sid:              request.GetCommonContext().GetSid().Id,
+		Success:          true,
+		EventTriggers:    eventTriggers,
+		RevalidationTime: revalidationTime,
+	}
+}
+
+func getUsageMonitoringCreditsProto(
+	umId string,
+	umData *n7_sbi.UsageMonitoringData,
+	monLevel protos.MonitoringLevel,
+) *protos.UsageMonitoringCredit {
+	monAction := protos.UsageMonitoringCredit_CONTINUE
+	if len(umData.UmId) == 0 {
+		// UmData not present for ID. It is supposed to be removed
+		monAction = protos.UsageMonitoringCredit_DISABLE
+	}
+	return &protos.UsageMonitoringCredit{
+		MonitoringKey: []byte(umId),
+		Action:        monAction,
+		Level:         monLevel,
+		GrantedUnits:  getGrantedUnitsProto(umData),
+	}
+}
+
+func getGrantedUnitsProto(umData *n7_sbi.UsageMonitoringData) *protos.GrantedUnits {
+	if umData == nil {
+		return &protos.GrantedUnits{
+			Total: &protos.CreditUnit{IsValid: false},
+			Tx:    &protos.CreditUnit{IsValid: false},
+			Rx:    &protos.CreditUnit{IsValid: false},
+		}
+	}
+	gsuProto := &protos.GrantedUnits{
+		Total: getCreditUnitsProto(umData.VolumeThreshold),
+		Tx:    getCreditUnitsProto(umData.VolumeThresholdUplink),   // Input == Tx == Uplink
+		Rx:    getCreditUnitsProto(umData.VolumeThresholdDownlink), // Output == Rx == Downlink
+	}
+
+	return gsuProto
+}
+
+func getCreditUnitsProto(volume *common5g.VolumeRm) *protos.CreditUnit {
+	if volume == nil {
+		return &protos.CreditUnit{IsValid: false}
+	}
+	return &protos.CreditUnit{IsValid: true, Volume: uint64(*volume)}
+}
+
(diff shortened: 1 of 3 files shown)


feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
 	)
 	return client, err
 }
+
```
uri200, line 43: see how this is done at other services [here](http://github.com/magma/magma/blob/master/feg/gateway/services/session_proxy/credit_control/gx/model_conversion.go#L32-L32)

model_conversion files have different kind of functions that converts from proto to diameter message and reverse
- [This](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fsession_proxy%2Fcredit_control%2Fgx%2Fmodel_conversion.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=84&start_col=1&end_row=84&end_col=1) converts from diameter structure to proto
- [This](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fsession_proxy%2Fcredit_control%2Fgx%2Fmodel_conversion.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=66&start_col=31&end_row=66&end_col=31) converts from proto to diameter struct

Note how in this case we do use pointers receivers to extend `RuleDefinition` or `QosRequestInfo`.  In cases like [this](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fsession_proxy%2Fcredit_control%2Fgx%2Fmodel_conversion.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=100&start_col=11&end_row=100&end_col=11), (for strings types) we just indicate the action we do on the function (that function convert from diameter type to proto type)

uri200: you will need to. put this on a different file

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+}
+
+func AddIMSIPrefix(imsi string) string {
+	return "IMSI" + imsi
+}
+
+func GetSmPolicyContextDataN7(
+	request *protos.CreateSessionRequest,
+	notifyApiRoot string,
+) *n7_sbi.PostSmPoliciesJSONRequestBody {
```
uri200:
```
func (request *protos.CreateSessionRequest,) ToSbi(
	notifyApiRoot string,
) *n7_sbi.PostSmPoliciesJSONRequestBody {
```

To use this 

```
request.ToSbi(notifyApiRoot)
```

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+
+func GenNotifyUrl(apiRoot string, sessionId string) sbi.Uri {
+	return sbi.Uri(fmt.Sprintf("%s/%s", apiRoot, b64.URLEncoding.EncodeToString([]byte(sessionId))))
+}
+
+func GetCreateSessionResponseProto(
+	request *protos.CreateSessionRequest,
+	smPolicyDecision *n7_sbi.SmPolicyDecision,
+	smPolicyUrl string,
+) *protos.CreateSessionResponse {
```
uri200:
```
func (r *n7_sbi.SmPolicyDecision) ToProto(
	request *protos.CreateSessionRequest,
	smPolicyUrl string,
) *protos.CreateSessionResponse {
```

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+	staticRules, dynamicRules, _, pccUmIds := GetPccRules(smPolicyDecision)
+	eventTriggers, revalidationTime := GetEventTriggersInfoProto(
+		smPolicyDecision.PolicyCtrlReqTriggers, smPolicyDecision.RevalidationTime)
+
+	tgppCtx := &protos.TgppContext{GxDestHost: smPolicyUrl}
```
uri200: why are you using Gx field here.  What is the relationship between GxDestHost and smPolicyUrl.

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+}
+
+func getSbiBool(val *bool) bool {
+	if val == nil {
+		return false
+	}
+	return *val
+}
+
+func GetPccRules(
```
uri200: i think this should also be a pointer receiver of *n7_sbi.SmPolicyDecision

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+	policyDecision *n7_sbi.SmPolicyDecision,
+) ([]*protos.StaticRuleInstall, []*protos.DynamicRuleInstall, []string, []string) {
+	staticRules := []*protos.StaticRuleInstall{}
```
uri200: use this construction to initiate slices
```
var staticRules []*protos.StaticRuleInstall
```

https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+	}
+
+	for ruleId, pccRule := range policyDecision.PccRules.AdditionalProperties {
+		if pccRule.PccRuleId == "" {
+			// This rule to be removed
+			rulesToRemove = append(rulesToRemove, ruleId)
+			continue
+		}
+		var activationTime, deactivationTime *timestamp.Timestamp
+		if pccRule.RefCondData != nil && *pccRule.RefCondData != "" {
```
uri200: this code here is calling for you to create a function like this:

```
func (pr *pccRule) GetActivationDeactivationTimes(activationTime *timestamp.Timestamp , deactivationTime *timestamp.Timestamp ) (){
	if pccRule.RefCondData != nil && *pccRule.RefCondData != "" {
		condData, found := policyDecision.Conds.Get(*pccRule.RefCondData)
		if found {
			activationTime = ConvertToProtoTimeStamp(condData.ActivationTime)
			deactivationTime = ConvertToProtoTimeStamp(condData.DeactivationTime)
		}
	}
	return 
}
```
Note in go if you specify the name of the var on the return part, you dont need to define them

Note go is not cpp. functions are good.

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+		}
+	}
+
+	return staticRules, dynamicRules, rulesToRemove, umIds
+}
+
+func GetPolicyRuleProto(
+	policyDecision *n7_sbi.SmPolicyDecision,
+	pccRule *n7_sbi.PccRule,
+) *protos.PolicyRule {
```
uri200: same as before  
```
func (pd *n7_sbi.SmPolicyDecision) ToProto .....
	
```

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+	policyRule := &protos.PolicyRule{
+		Id: pccRule.PccRuleId,
+	}
+	if pccRule.Precedence != nil {
+		policyRule.Priority = uint32(*pccRule.Precedence)
+	}
+	// For all Ref*Data Maximum one entry is expected.
+	// These are declared as arrays for future compatibility
```
uri200: create functions to encapsulate all this logic as much as you can.
For example `func (pcc *n7_sbi.PccRule)getRefUmData....`

You should not have long functions if this can be encapsulated on some kind of function with meaning

If you want you can do it reverse and just modify policy rule with the values extrcted from pcc

```
func (pr *policyRule)getRefUmData(pcc **n7_sbi.PccRule)error{
    .. logic to extract values....
    pr.MonitoringKey = []byte(umData.UmId)
    return nil
}
```

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+	monKeyPresent := false
+	ratingGrpPresent := false
+	dynamicRule := false
+	if pccRule.RefUmData != nil && len(*pccRule.RefUmData) > 0 && policyDecision.UmDecs != nil {
+		umData, found := policyDecision.UmDecs.Get((*pccRule.RefUmData)[0])
```
uri200: hardcoded index means trouble normally. Can you justify the need of a hardcoded 0 here?
Can pccRule.RefUmData be more than 1?

And if it is more than 1 why are not caring?

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+				flow, err := policydb.GetFlowDescriptionFromFlowString(string(*flowInfo.FlowDescription))
+				if err != nil {
+					glog.Errorf("Could not get flow for description %s : %s", *flowInfo.FlowDescription, err)
+				} else {
+					policyRule.FlowList = append(policyRule.FlowList, flow)
+				}
+			}
+		}
+	}
+	if !dynamicRule {
```
uri200: what is this?  why don't just return right after finding dynamicRule == false?

better to have 4 returns right after where it is supposed to be return that carrying the whole thing up to here just to return nil

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+			policyRule.Qos = GetFlowQosProto(&qosData)
+			dynamicRule = true
+		}
+	}
+	if pccRule.FlowInfos != nil {
+		for _, flowInfo := range *pccRule.FlowInfos {
+			if flowInfo.FlowDescription != nil {
+				flow, err := policydb.GetFlowDescriptionFromFlowString(string(*flowInfo.FlowDescription))
+				if err != nil {
+					glog.Errorf("Could not get flow for description %s : %s", *flowInfo.FlowDescription, err)
```
uri200: why this error is just handled here with a simple log?

What will happen to the rule in sessiond if it is sent without any flow description? Will this rule be valid?

uri200 (changes requested):

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+					policyRule.FlowList = append(policyRule.FlowList, flow)
+				}
+			}
+		}
+	}
+	if !dynamicRule {
+		// not a dynamic rule
+		return nil
+	}
+	return policyRule
```
uri200: this whole function needs you to rethink it. Do not use booleans to carry results, just execute the functuion `policyRule.TrackingType = GetTrackingTypeProto(monKeyPresent, ratingGrpPresent)` inside each switch case its needed 

Each switch case should look like a receipt of how to create the rule

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+	} else if monKeyPresent && !ratingGrpPresent {
+		return protos.PolicyRule_ONLY_PCRF
+	} else if !monKeyPresent && ratingGrpPresent {
+		return protos.PolicyRule_ONLY_OCS
+	} else {
+		return protos.PolicyRule_NO_TRACKING
+	}
+}
+
+func GetRedirectInfoProto(redirInfo *n7_sbi.RedirectInformation) *protos.RedirectInformation {
```
uri200: apply here and in the rest of the functions pointers receiver either to tproto or sbi interface

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+	return protoTriggers, protoRevalidationTime
+}
+
+func GetUsageMonitorsProto(
+	request *protos.CreateSessionRequest,
+	tgppCtx *protos.TgppContext,
+	eventTriggers []protos.EventTrigger,
+	revalidationTime *timestamp.Timestamp,
+	smPolicyDecision *n7_sbi.SmPolicyDecision,
+	pccUmIds []string,
```
uri200: those are lots of arguments. I havent been able to review them , but please make sure that things like `pccUmIds` are not included on other objects here

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +50,509 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
+			protoTriggers = append(protoTriggers, protos.EventTrigger_REVALIDATION_TIMEOUT)
+			protoRevalidationTime = ConvertToProtoTimeStamp(revalTimeout)
+		}
+	}
+	return protoTriggers, protoRevalidationTime
+}
+
+func GetUsageMonitorsProto(
+	request *protos.CreateSessionRequest,
+	tgppCtx *protos.TgppContext,
```
uri200: what is this for? I think i remember that was something related with gx/gy but not sure if this will be needed in here. 
Can you investigate in the code to understand what this means please?

feg/gateway/services/n7_n40_proxy/n7/n7_create_test.go
```diff
@@ -0,0 +1,233 @@
...
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
```
uri200: this should be on the conversion_test,go file

feg/gateway/services/n7_n40_proxy/n7/n7_create_test.go
```diff
@@ -0,0 +1,233 @@
...
+	Gpsi1              = sbi.Gpsi(GPSI1)
+	UeIpv4             = sbi.Ipv4Addr(UE_IPV4)
+	PduSessionTypeIpv4 = sbi.PduSessionType("IPV4")
+	RatTypeNR          = sbi.RatType("NR")
+	UeTzIST            = sbi.TimeZone("+05:30")
+	SmPolicyUrl        = "https://example.com//npcf-smpolicycontrol/v1/sm-policies/12345"
+	ActTime            = time.Unix(1634906551, 0)
+	DeactTime          = time.Unix(1634913751, 0)
+)
+
```
uri200: good. The reason to have those variables here is that it is easy to change them in case you need them. And if test fails you can easily see values here

feg/gateway/services/n7_n40_proxy/n7/n7_create_test.go
```diff
@@ -0,0 +1,233 @@
...
+	csrProto := &protos.CreateSessionRequest{
+		CommonContext: &protos.CommonSessionContext{
+			Sid: &protos.SubscriberID{Id: IMSI1},
+		},
+		SessionId: genSessionID(IMSI1),
+	}
+	csResp := n7.GetCreateSessionResponseProto(
+		csrProto, &policyDecision, SmPolicyUrl)
+
+	expected := &protos.CreateSessionResponse{
```
uri200: `GetCreateSessionResponseProto` should be executed after expected

In other words, when executing `GetCreateSessionResponseProto` your test should be fully set up

feg/gateway/services/n7_n40_proxy/n7/n7_create_test.go
```diff
@@ -0,0 +1,233 @@
...
+	// Unmarshal json to openapi struct
+	var policyDecision n7_sbi.SmPolicyDecision
+	err := json.Unmarshal([]byte(policyDecisionStr), &policyDecision)
+	require.NoError(t, err)
+
+	csrProto := &protos.CreateSessionRequest{
+		CommonContext: &protos.CommonSessionContext{
+			Sid: &protos.SubscriberID{Id: IMSI1},
+		},
+		SessionId: genSessionID(IMSI1),
```
uri200: do you really need to use this function?  Can't not you use a constant ?
Better to have tests with deterministic values (even that you may have find otherwise in session proxy unfortunately)

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: Don't think we can use pointer receivers for structs defined in another package. Will get this error - "invalid receiver *protos.CreateSessionRequest (type not defined in this package)"

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: As mentioned earlier can't define receivers for structs defined in another package. I can probably rename it as CreateSessionRequestToSbi

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: See other comments. Can rename it as SmPolicyDecisionToProto

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: The TGPP Context is an opaque data sent to Sessiond for storing session state related to Feg. This TGPP context is send by SessionD for every subsequent request. This is getting reused to get the SmPolicyUrl or UpdateSession and TerminateSession. SmPolicyUrl uniquely identifies a policy session at the PCF. PCF does not know about SessionId.

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: See my other responses related to this.

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: 3GPP specs says there can be only one element for refUmData. In the openapi minItems and maxItems are set as 1. It is declared as an array for future usage.

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: It will be determined as dynamic rule only when all the previous condition are not satisfied. Hence the return at the end.

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: See other responses related to this.

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: Usage monitors are set by the PCF/PCRF for reporting Usage when the set volume thresholds are reached. SessionD sends the UsageMonitoringReport in the UpdateSession and also in the TerminateSession.

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: This happens on a reponse and we can't return an error to PCF. This rule will not have the flow information, might have other info line Qos, Charging data, user monitoring etc..

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: The tracking type can be determined only based on the presence of Usage Monitoring data and Charging data. Can you please elaborate on the usage of switch case? In the above function, switch case cannot be used as all the conditions can be satisfied.

feg/gateway/services/n7_n40_proxy/n7/n7.go
uri200: You are right. In case Gx.Gy the base structs are part of diameter but messages are part of definition.go that belongs to gx.

Then you can ignore all the comments related to that. But please, find a consistent way to label all them. You will have functions that convert from proto to sbi and from sbi to proto

feg/gateway/services/n7_n40_proxy/n7/n7.go
uri200: Thanks. Yes I remember when we added it for 4g. I just wanted to make sure it is also used for 5g

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+	return rpcResp
+}
+
+func getSbiIpv4(ipv4 string) *sbi.Ipv4Addr {
+	if len(ipv4) == 0 {
+		return nil
+	}
+	ret := sbi.Ipv4Addr(ipv4)
+	return &ret
+}
```
uri200: Just curious if we could save those functions. Will `PostSmPoliciesJSONRequestBody` work the same with a nil value than with an empty string `""`?

If you are not able to find that on the code, you can just remove one of those  `if len(ipv4) == 0 {`  and  run the unit test to see if it works. If you still want confirmation you can use the debugger to see the values.

Just curious to understand how robust is the sbi api to see if we can save some code

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+		Supi:              sbi.Supi(imsi),
+		Ipv4Address:       getSbiIpv4(common.GetUeIpv4()),
+		Ipv6AddressPrefix: getSbiIpv6(common.GetUeIpv6()),
+		Dnn:               sbi.Dnn(common.GetApn()),
+		Gpsi:              getSbiGpsi(string(common.GetMsisdn())),
+		RatType:           getSbiRatType(ratType),
+		AccessType:        getSbiAccessType(ratType),
+		UeTimeZone:        getSbiTimeZone(request.GetAccessTimezone()),
+		NotificationUri:   GenNotifyUrl(notifyApiRoot, request.SessionId),
+	}
```
uri200, line 55: Can you confirm that when you set a field to nil, this is not going to break the API?

Meaning does the sbi API tell you to do it so?

Otherwise we may need to do ifs 

```
sbiIpv4 := getSbiIpv4(common.GetUeIpv4())
if sbiIpv4 != nil {
	reqBody.Ipv4Address = sbiIpb4
}
```
(lets hope we don't have to do it that way, but since this is a new api, lets make sure it works with nils as expected)

You can create a simple test case for that, or you can add checks to current test case to confirm empty parameters are not icluded

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+	tz := fmt.Sprintf("%02d:%02d", offsetMin/60, offsetMin%60)
+	if isNegative {
+		tz = "-" + tz
+	} else {
+		tz = "+" + tz
+	}
+	tzStr := sbi.TimeZone(tz)
+	return &tzStr
+}
```
uri200, line 118: this function may deserve a test case by itself to make sure positive and negative and 30 min time zones are covered

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+	}
+}
+
+func getSbiBool(val *bool) bool {
+	if val == nil {
+		return false
+	}
+	return *val
+}
+
```
uri200, line 136: yes, we have to make sure if there is a better way to do all this tiny functions. Check if the API already provides support for that. 

If not, lets put all those tiny functions ate the end so they don't distract the attention of the reader

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+) (activationTime, deactivationTime *timestamp.Timestamp) {
+	if pccRule.RefCondData != nil && *pccRule.RefCondData != "" {
+		condData, found := policyDecision.Conds.Get(*pccRule.RefCondData)
+		if found {
+			activationTime = ConvertToProtoTimeStamp(condData.ActivationTime)
+			deactivationTime = ConvertToProtoTimeStamp(condData.DeactivationTime)
+		}
+	}
+	return
+}
```
uri200, line 224: If possible, and it makes sense, try to have as less indentation levels as possible. It makes the code easier le read.

See how in this code you don't have to jump back and forth to understand, It is lineal

```
func GetActivateDeactivationTime(policyDecision *n7_sbi.SmPolicyDecision, pccRule *n7_sbi.PccRule,
) (activationTime, deactivationTime *timestamp.Timestamp) {
	if pccRule.RefCondData != nil && *pccRule.RefCondData == "" {
		return
	}
	condData, found := policyDecision.Conds.Get(*pccRule.RefCondData)
	if found == false {
		return
	} 
	activationTime = ConvertToProtoTimeStamp(condData.ActivationTime)
	deactivationTime = ConvertToProtoTimeStamp(condData.DeactivationTime)
}
```

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+		chgData, found := policyDecision.ChgDecs.Get((*pccRule.RefChgData)[0])
+		if found {
+			policyRule.RatingGroup = uint32(*chgData.RatingGroup)
+			policyRule.ServiceIdentifier = &protos.ServiceIdentifier{
+				Value: uint32(*chgData.ServiceId),
+			}
+			policyRule.Offline = getSbiBool(chgData.Offline)
+			policyRule.Online = getSbiBool(chgData.Online)
+		}
+	}
```
uri200: let's make this a function

```
func insertChargingAndMonitoringKeyIfNecessary(pccRule *n7_sbi.PccRule , policyRule *protos.PolicyRule){
      if pccRule.RefChgData != nil && len(*pccRule.RefChgData) > 0 && policyDecision.ChgDecs != nil {
		chgData, found := policyDecision.ChgDecs.Get((*pccRule.RefChgData)[0])
		if found {
			policyRule.RatingGroup = uint32(*chgData.RatingGroup)
			policyRule.ServiceIdentifier = &protos.ServiceIdentifier{
				Value: uint32(*chgData.ServiceId),
			}
			policyRule.Offline = getSbiBool(chgData.Offline)
			policyRule.Online = getSbiBool(chgData.Online)
		}
	}
```
see that `policyRule` object is a piointer so you will be able to modify it

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+	if monKeyPresent && ratingGrpPresent {
+		return protos.PolicyRule_OCS_AND_PCRF
+	} else if monKeyPresent && !ratingGrpPresent {
+		return protos.PolicyRule_ONLY_PCRF
+	} else if !monKeyPresent && ratingGrpPresent {
+		return protos.PolicyRule_ONLY_OCS
+	} else {
+		return protos.PolicyRule_NO_TRACKING
+	}
+}
```
uri200, line 341: did you find this in our codebase? Can you point me where

I remember we had some trouble with that. We should not infering parameters from others. That cause unexpected results if we ever change one of the functons without knowing it impacts others

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+	pccRule *n7_sbi.PccRule,
+) []byte {
+	if pccRule.RefUmData != nil && len(*pccRule.RefUmData) > 0 && policyDecision.UmDecs != nil {
+		umData, found := policyDecision.UmDecs.Get((*pccRule.RefUmData)[0])
+		if found {
+			return []byte(umData.UmId)
+		}
+	}
+	return []byte{}
+}
```
uri200, line 282: shound't you be returning `nil`  instad of `[]byte{}`

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+
+func getFlowListFromPccRule(pccRule *n7_sbi.PccRule) []*protos.FlowDescription {
+	var flowList []*protos.FlowDescription
+	if pccRule.FlowInfos != nil {
+		for _, flowInfo := range *pccRule.FlowInfos {
+			if flowInfo.FlowDescription != nil {
+				flow, err := policydb.GetFlowDescriptionFromFlowString(string(*flowInfo.FlowDescription))
+				if err != nil {
+					glog.Errorf("Could not get flow for description %s : %s", *flowInfo.FlowDescription, err)
```
uri200: this error is still not handled. 
What will happen to the rule in sessiond if it is sent without any flow description? Will this rule be valid?

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+	}
+
+	return redirProto
+}
+
+func GetRedirectAddrTypeProto(
+	redirAddrType *n7_sbi.RedirectAddressType,
+) protos.RedirectInformation_AddressType {
+	if redirAddrType == nil {
+		return protos.RedirectInformation_IPv4
```
uri200, line 368: do you really need to povide redirect information when it is nil?

Note redirect will be empty almost always. Now you are converting this to a specific value. Is that correct?

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+	case n7_sbi.RedirectAddressTypeIPV4ADDR:
+		return protos.RedirectInformation_IPv4
+	case n7_sbi.RedirectAddressTypeIPV6ADDR:
+		return protos.RedirectInformation_IPv6
+	case n7_sbi.RedirectAddressTypeSIPURI:
+		return protos.RedirectInformation_SIP_URI
+	case n7_sbi.RedirectAddressTypeURL:
+		return protos.RedirectInformation_URL
+	default:
+		return protos.RedirectInformation_IPv4
```
uri200, line 381: if this is not supposed to happen you need to trigger an error. At least log it and leave it to 0

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+				Sid:              request.GetCommonContext().GetSid().Id,
+				Success:          true,
+				EventTriggers:    eventTriggers,
+				RevalidationTime: revalidationTime,
+			})
+		}
+	}
+
+	return monitors
+}
```
uri200, line 459: please remember about avoiding code repetition. If you find yourself with repeated code, then you should refactor that. 

It looks like you could create a function to take of extracting monitors out of  `AdditionalProperties`

```
appendToMonitors(AdditionalProperties type(?), monitoringLevel protos.Monitoring_Level, []*protos.UsageMonitoringUpdateResponse)  
```

You can pass the slice and append in the function itself

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,567 @@
...
+	}
+	return &protos.CreditUnit{IsValid: true, Volume: uint64(*volume)}
+}
+
+func ConvertToProtoTimeStamp(srcTime *time.Time) *timestamp.Timestamp {
+	if srcTime == nil {
+		return nil
+	}
+	return timestamppb.New(*srcTime)
+}
```
uri200, line 528: please can you go over the file and group all the Proto fucntions and SBI functions toghetehr.  You can add a comment per each of section so it is visible


Another thing I personally consider good is to put all the small functions at the end (for example `BitRateStringToUint32` should be at the end of the proto block). The reason is those functions are kind of obvious, and they are distracting when reading the code.

uri200 (changes requested):

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: An empty string is treated as an empty string. The generated json has ipv4address field and the value as "". As mentioned before 'nil' means not present. Any other value, even if it is default, means the field is present.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: OpenAPIs has mandatory (required) fields. For these mandatory fields, the generated code does not generate pointer types. For all other non-mandatory fields pointer type is generated. If a field is set to nil, it means that the field is not present and a json-value will not be generated for it.

Added a test case for this

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: This value is returned to protobuf generated code which is of type []byte and not *[]byte.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Since this is received in the answer response, will not be able to inform PCF about the error. This behavior was taken from [here](https://github.com/magma/magma/blob/f9840abd66b4616695b7ec6405ae87186c65c516/feg/gateway/services/session_proxy/credit_control/gx/model_conversion.go#L155).

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Found it [here](https://github.com/magma/magma/blob/f9840abd66b4616695b7ec6405ae87186c65c516/feg/gateway/services/session_proxy/credit_control/gx/model_conversion.go#L122).

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: See comment below

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: RedirectInformation_IPv4 is the default value (0). Again here it is getting converted to protobuf and the type is not a pointer. Hence setting to default value when not set in the sbi type.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: SessRules.AdditionalProperties and PccRules.AdditionalProperties are of different types. But common code can be moved to a new function

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Added a test for this

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Tiny functions moved to the end.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
uri200: thanks @GANESH-WAVELABS  for clarifying

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
uri200: Ok that is fine, I think it would work with either. 

But note  slices are interfaces, meaning they can be nil. See [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fcloud%2Fgo%2Fprotos%2Fs8_proxy.pb.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=303&start_col=55&end_row=303&end_col=55)

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
uri200: I see... not good at all. But lets leave it as it is then

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,593 @@
...
+				if err != nil {
+					glog.Errorf("Could not get flow for description %s : %s", *flowInfo.FlowDescription, err)
+				} else {
+					flowList = append(flowList, flow)
+				}
+			}
+		}
+	}
+	return flowList
+}
```
uri200, line 329: can you un nest this function please?
```
func getFlowListFromPccRule(pccRule *n7_sbi.PccRule) []*protos.FlowDescription {
	var flowList []*protos.FlowDescription
	if pccRule.FlowInfos == nil {
		continue
	}
	for _, flowInfo := range *pccRule.FlowInfos {
		if flowInfo.FlowDescription == nil {
			continue
		}
		flow, err := policydb.GetFlowDescriptionFromFlowString(string(*flowInfo.FlowDescription))

```

Continues are very easy to see at first sight and indicated that case is ignored. Otherwise you have to look for the condition and you end up with lots of indentation levels imposible to follow

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
uri200: on protos, values are converted into default: 0 if int "" if string. 

But arrays are pointers and they are converted into nil when sending the proto

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
uri200: Hey @GANESH-WAVELABS , per what I see both `AdditionalProperties` are the same object  (I can be wrong though)
```
AdditionalProperties map[string]externalRef1.PresenceInfo 
```

[here](https://github.com/magma/magma/blob/36455a93824e3c93c08d06351c73762c1c4fa02e/feg/gateway/sbi/specs/TS29514NpcfPolicyAuthorization/TS29514NpcfPolicyAuthorization.gen.go#L314:2) 

If you use intellij (free version) you ll be able to follow the definition to the ts file automatically

GANESH-WAVELABS: [This](https://github.com/magma/magma/blob/5584180910f1847b56c3e6c410b443640c194f97/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl/TS29512NpcfSMPolicyControl.gen.go#L813) is type of AdditionalProperties in SessRules and [this](https://github.com/magma/magma/blob/5584180910f1847b56c3e6c410b443640c194f97/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl/TS29512NpcfSMPolicyControl.gen.go#L793) is the type of AdditionalProperties in PccRules.

In OpenAPI/Swagger AdditionalProperties represents a map/dict like object. The key will be string and the value type can be defined in the openapi file.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: That is correct, but here the proto type is a enum and the go type is an int and hence returning a default value.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

uri200: lets do this

```
	...
		if err != nil {
			glog.Errorf("Could not get flow for description %s : %s", *flowInfo.FlowDescription, err)
			continue
		}
		flowList = append(flowList, flow)
	}
```

The reason is you ever have to add mor stuff to that function you will have to add it inside the else. If for any reason the person that adds that new code puts it outside the else, the logic will be wrong.

Plus having a continue is very easy to tell that after that glog there is nothing else to do for that path.

Once you have change this, take a look at the function and you will see this patter of having all the edge cases at the top, and the main logic at the bottom. This way reading a function becomes very easy.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,595 @@
...
+		accessType = sbi.AccessTypeN3GPPACCESS
+	case protos.RATType_TGPP_WLAN:
+		accessType = sbi.AccessTypeNON3GPPACCESS
+	default:
+		return nil
+	}
+	return &accessType
+}
+
+func GetSbiTimeZone(tzProto *protos.Timezone) *sbi.TimeZone {
```
uri200, line 100: btw, I haven't check the scope of all those functions, but note that if you are not going to use those functions outside `package n7` you do not need to capitalize any (meaning they can be private)

Note normally public methods normally require a definition/doc string (not necessary here). So the more private the less description go will require.

In this case, just make sure you really need to export those functions or not.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,595 @@
...
+	if pccRule.RefUmData != nil && len(*pccRule.RefUmData) > 0 && policyDecision.UmDecs != nil {
+		umData, found := policyDecision.UmDecs.Get((*pccRule.RefUmData)[0])
+		if found {
+			return []byte(umData.UmId)
+		}
+	}
+	return []byte{}
+}
+
+func getTcDataFromPccRule(
```
uri200, line 286: `getTcDataFromPccRule` this name is not very descriptive . Can you add a comment on what TC is 
Something like:
```
// getTcDataFromPccRule extracts Redirect Information from the SmPolicyDecision and returns a proto SmPolicyDecision

```

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
uri200: I think this is hiding a possible error. Can you at least trigger an log message saying that the field was defaulted?

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
uri200: you are right, they are not the same. I got fooled by the name.
Thanks for adding the links

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -0,0 +1,595 @@
...
+func N5qiToQci(n5Qi *sbi.N5Qi) protos.FlowQos_Qci {
+	if n5Qi == nil {
+		return protos.FlowQos_QCI_0
+	}
+	return protos.FlowQos_Qci(*n5Qi)
+}
+
+func GenNotifyUrl(apiRoot string, sessionId string) sbi.Uri {
+	return sbi.Uri(fmt.Sprintf("%s/%s", apiRoot, b64.URLEncoding.EncodeToString([]byte(sessionId))))
+}
```
uri200: I think none of those functions need to be capitalized.

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -39,3 +40,11 @@ func NewN7Client(cfg *PCFConfig) (*n7_sbi.ClientWithResponses, error) {
...
 	return client, err
 }
+
+func RemoveIMSIPrefix(imsi string) string {
+	return strings.TrimPrefix(imsi, "IMSI")
+}
+
+func AddIMSIPrefix(imsi string) string {
+	return "IMSI" + imsi
+}
```
uri200: same here. i think no need of capitalization

uri200 (changes requested):

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7/n7.go
GANESH-WAVELABS: Done

uri200 (approved):

PR #10301, opened 2021-11-14
feat(feg): N7 Proxy SM Policy Update request implementation

GANESH-WAVELABS:
## Summary
- Parse the multiple user monitoring updates received on the CentralSessionController UpdateSession request and conver them to SmPolicyUpdateContextData. This produces multiple SmPolicyUpdate requests
- Send all the update requests to PCF in paralle and wait for the responses
- Accumulate all the sent responses and populate UsageMonitoringReportResponse to be sent as a response to CentralSessionController UpdateSession

### DIFF
--- a/feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
+++ b/feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
@@ -61,6 +61,16 @@ var FailureCodeProtoToN7Map = []n7_sbi.FailureCode{
 	"UNKNOWN",                // 26
 }
 
+type SmPolicyUpdateReqCtx struct {
+	SmPolicyId    string
+	SessionId     string
+	IMSI          string
+	MonitoringKey []byte
+	Level         protos.MonitoringLevel
+	TgppCtx       *protos.TgppContext
+	ReqBody       *n7_sbi.PostSmPoliciesSmPolicyIdUpdateJSONRequestBody
+}
+
 //
 // From Proto to SBI types
 //
@@ -134,6 +144,49 @@ func GetSmPolicyDeleteReqBody(
 	return &n7_sbi.PostSmPoliciesSmPolicyIdDeleteJSONRequestBody{
 		AccuUsageReports: getUmReportN7(request.MonitorUsages),
 	}
+}
+
+func GetSmPolicyUpdateRequestsN7(
+	updates []*protos.UsageMonitoringUpdateRequest,
+) []*SmPolicyUpdateReqCtx {
+	updatesPerSession := make(map[string][]*protos.UsageMonitoringUpdateRequest)
+	// Sort updates per session
+	for _, update := range updates {
+		updatesPerSession[update.SessionId] = append(updatesPerSession[update.SessionId], update)
+	}
+
+	// Merge the updates for the same session into a single request
+	reqCtxs := []*SmPolicyUpdateReqCtx{}
+	for _, updatesList := range updatesPerSession {
+		firstUpdate := updatesList[0]
+		smPolicyId, err := GetSmPolicyId(firstUpdate.TgppCtx)
+		if err != nil {
+			// No 3gpp context. Cannot get the PolicyId
+			glog.Errorf("SmPolicyUpdate: Session %s does not have SmPolicyId: %s", firstUpdate.SessionId, err)
+			continue
+		}
+		accUsageReport := []n7_sbi.AccuUsageReport{}
+		for _, umUpdate := range updatesList {
+			if umUpdate.EventTrigger == protos.EventTrigger_USAGE_REPORT {
+				accUsageReport = append(accUsageReport, getAccUsageReportN7(umUpdate.Update))
+			}
+		}
+		reqCtxs = append(reqCtxs, &SmPolicyUpdateReqCtx{
+			SmPolicyId:    smPolicyId,
+			SessionId:     firstUpdate.SessionId,
+			IMSI:          firstUpdate.Sid,
+			MonitoringKey: firstUpdate.Update.MonitoringKey,
+			Level:         firstUpdate.Update.Level,
+			TgppCtx:       firstUpdate.TgppCtx,
+			ReqBody: &n7_sbi.PostSmPoliciesSmPolicyIdUpdateJSONRequestBody{
+				Ipv4Address:      getSbiIpv4(firstUpdate.UeIpv4),
+				RatType:          getSbiRatType(firstUpdate.RatType),
+				AccessType:       getSbiAccessType(firstUpdate.RatType),
+				AccuUsageReports: &accUsageReport,
+			},
+		})
+	}
+	return reqCtxs
 }
 
 func getSbiRatType(ratType protos.RATType) *sbi.RatType {
@@ -284,6 +337,48 @@ func GetPolicyReauthRequestProto(
 		EventTriggers:          eventTriggers,
 		RevalidationTime:       revalidationTime,
 		UsageMonitoringCredits: umCredits,
+	}
+}
+
+// GetUsageMonitoringResponsesProto converts smPolicyDecision received from PCF to UsageMonitoringUpdatreResponse proto.
+// For the conversion it uses the updateCtx that was built before sending the request to PCF.
+func GetUsageMonitoringResponsesProto(
+	updateCtx *SmPolicyUpdateReqCtx,
+	smPolicyDecision *n7_sbi.SmPolicyDecision,
+) []*protos.UsageMonitoringUpdateResponse {
+	staticRules, dynamicRules, rulesToRemove := getPccRules(smPolicyDecision)
+	responses := getUsageMonitorsProtoForUpdate(updateCtx, smPolicyDecision)
+	if len(responses) == 0 {
+		responses = append(responses, GetUsageMonitoringUpdateResponseProto(updateCtx, true))
+	}
+
+	eventTriggers, revalidationTime := getEventTriggersInfoProto(
+		smPolicyDecision.PolicyCtrlReqTriggers, smPolicyDecision.RevalidationTime)
+	firstResponse := responses[0]
+	firstResponse.StaticRulesToInstall = staticRules
+	firstResponse.DynamicRulesToInstall = dynamicRules
+	firstResponse.RulesToRemove = rulesToRemove
+	firstResponse.EventTriggers = eventTriggers
+	firstResponse.RevalidationTime = revalidationTime
+
+	return responses
+}
+
+// GetUsageMonitoringUpdateResponseProto builds a default UsageMonitoringUpdateResponse proto from
+// updateCtx that was built before sending SmPolicyUpdate request to PCF.
+func GetUsageMonitoringUpdateResponseProto(
+	updateCtx *SmPolicyUpdateReqCtx,
+	success bool,
+) *protos.UsageMonitoringUpdateResponse {
+	return &protos.UsageMonitoringUpdateResponse{
+		Success:   success,
+		SessionId: updateCtx.SessionId,
+		Sid:       updateCtx.IMSI,
+		TgppCtx:   updateCtx.TgppCtx,
+		Credit: &protos.UsageMonitoringCredit{
+			MonitoringKey: updateCtx.MonitoringKey,
+			Level:         updateCtx.Level,
+		},
 	}
 }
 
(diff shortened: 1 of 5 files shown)


feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -136,6 +146,52 @@ func GetSmPolicyDeleteReqBody(
...
+	}
+
+	// Merge the updates for the same session into a single request
+	reqCtxs := []*SmPolicyUpdateReqCtx{}
+	for _, updatesList := range updatesPerSession {
+		firstUpdate := updatesList[0]
+		smPolicyId, err := GetSmPolicyId(firstUpdate.TgppCtx)
+		if err != nil {
+			// No 3gpp context. Cannot get the PolicyId
+			glog.Errorf("SmPolicyUpdate: Session %s does not have SmPolicyId", firstUpdate.SessionId)
```
uri200: don't hide error returned by`GetSmPolicyId`, add it to the string preceding it by `:`

```
glog.Errorf("SmPolicyUpdate: Session %s does not have SmPolicyId: %s", firstUpdate.SessionId, err)
```

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -136,6 +146,52 @@ func GetSmPolicyDeleteReqBody(
...
+			continue
+		}
+
+		reqBody := &n7_sbi.PostSmPoliciesSmPolicyIdUpdateJSONRequestBody{
+			Ipv4Address: getSbiIpv4(firstUpdate.UeIpv4),
+			RatType:     getSbiRatType(firstUpdate.RatType),
+			AccessType:  getSbiAccessType(firstUpdate.RatType),
+		}
```
uri200: move this block right before `reqCtxs = append(reqCtxs, &SmPolicyUpdateReqCtx{` and build the  `PostSmPoliciesSmPolicyIdUpdateJSONRequestBody` there. It is easier to read if you place how you build an object on the same place, if possible

feg/gateway/services/n7_n40_proxy/servicers/update_session.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (srv *CentralSessionController) sendSingleSmPolicyUpdate(
+	ctx context.Context,
+	updateCtx *n7.SmPolicyUpdateReqCtx,
+) []*protos.UsageMonitoringUpdateResponse {
+	resp, err := srv.policyClient.PostSmPoliciesSmPolicyIdUpdateWithResponse(
+		ctx, updateCtx.SmPolicyId, *updateCtx.ReqBody)
+	if err == nil && resp.StatusCode() != http.StatusOK {
+		err = fmt.Errorf("http error status-code=%d", resp.StatusCode())
+	}
```
uri200, line 72: why is this if outside the `PostSmPoliciesSmPolicyIdUpdateWithResponse`

Or if this doesnt belong to `PostSmPoliciesSmPolicyIdUpdateWithResponse`, why do we need this here?

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -287,6 +343,44 @@ func GetPolicyReauthRequestProto(
 	}
 }
 
+func GetUsageMonitoringResponsesProto(
```
uri200, line 345: can you add a docstring so we remember `updateCtx` is comming from the request made by the client and `smPolicyDecision` is the received response?

It is weird to me to have a conversion function that depends on the request (more comments below)

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -597,6 +691,61 @@ func newUsageMonitoringUpdateResponse(
...
+		return nil
+	}
+	return &protos.UsageMonitoringUpdateResponse{
+		Credit:    getUsageMonitoringCreditsProto(umId, &umEntry, monLevel),
+		SessionId: updateCtx.SessionId,
+		TgppCtx:   updateCtx.TgppCtx,
+		Sid:       updateCtx.IMSI,
+		Success:   true,
+	}
+}
```
uri200, line 752: all these new functions have as argument the request and the response, which, in my opinion makes it a little bit weird to read (it make sense, but weird). One solution would be just pass `sessionID`, `tgppCtx` and `Sid` separateley. But that may create a big signature. We could create a new type to pass those arguments (which would be similar object than `updateCtx`). But none of those solutions seems better than what you have.

So lets add a doc string to all those function reminding that `SmPolicyUpdateReqCtx` is used just to fill some of the fields the response from PCF does not provide

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -287,6 +343,44 @@ func GetPolicyReauthRequestProto(
...
 
+func GetUsageMonitoringResponsesProto(
+	updateCtx *SmPolicyUpdateReqCtx,
+	smPolicyDecision *n7_sbi.SmPolicyDecision,
+) []*protos.UsageMonitoringUpdateResponse {
+	staticRules, dynamicRules, rulesToRemove := getPccRules(smPolicyDecision)
+	responses := getUsageMonitorsProtoForUpdate(updateCtx, smPolicyDecision)
+	if len(responses) == 0 {
+		responses = append(responses, NewUsageMonitoringUpdateResponseProto(updateCtx, true))
+	}
```
uri200, line 353: why do you need this `if` here?

And if you need it, wouldint it be better to have it inside `getUsageMonitorsProtoForUpdate`? Or is this creating a fake `response` in case there is no response? Is it possible we discuss something similar for omnipresent rules?

uri200: the reason  I don't like this is because we will forget in the future. Maybe we need to do this somewhere else (for example at the omnirules, this way we just have to do it once

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -287,6 +343,44 @@ func GetPolicyReauthRequestProto(
...
+
+	eventTriggers, revalidationTime := getEventTriggersInfoProto(
+		smPolicyDecision.PolicyCtrlReqTriggers, smPolicyDecision.RevalidationTime)
+	firstResponse := responses[0]
```
uri200, line 357: i think I already asked about this, but why is it 0? Why can't it have more responses?

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -287,6 +343,44 @@ func GetPolicyReauthRequestProto(
...
+	firstResponse.StaticRulesToInstall = staticRules
+	firstResponse.DynamicRulesToInstall = dynamicRules
+	firstResponse.RulesToRemove = rulesToRemove
+	firstResponse.EventTriggers = eventTriggers
+	firstResponse.RevalidationTime = revalidationTime
+
+	return responses
+}
+
+func NewUsageMonitoringUpdateResponseProto(
```
uri200: wouldn't `getUsageMonitoringUpdateResponseProto` look better here?

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -287,6 +343,44 @@ func GetPolicyReauthRequestProto(
...
+		SessionId: updateCtx.SessionId,
+		Sid:       updateCtx.IMSI,
+		TgppCtx:   updateCtx.TgppCtx,
+		Credit: &protos.UsageMonitoringCredit{
+			MonitoringKey: updateCtx.MonitoringKey,
+			Level:         updateCtx.Level,
+		},
+	}
+}
+
```
uri200, line 384: do you need all those functions capitalized?

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -597,6 +691,61 @@ func newUsageMonitoringUpdateResponse(
...
+			if umResp == nil {
+				continue
+			}
+			monitors = append(monitors, umResp)
+		}
+	}
+	return monitors
+}
+
+func newUsageMonitoringUpdateResponseFromUpdateCtx(
```
uri200: same here, i think it is better to use get

uri200 (changes requested):

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: This is similar to what was done for [Gx](https://github.com/magma/magma/blob/ae3224e0f5bb71b03aade6f85e89d1112d42d511/feg/gateway/services/session_proxy/servicers/policy.go#L359). 

This is because of the way UpdateSession RPC behaves. There can be multiple UpdateSessionRequests from sessiond and each can be for different SessionID. N7 proxy groups the UpdateSessionRequests based on SessionID and sends one request to per PolicySession. 

When the response for one PolicySession is received, it is converted to  multiple UsageMonitoringUpdateResponse that can contain mutiple UsageMonitoringUpdates. The first UsageMonitoringUpdate is updated with the rules received from PCF. Other UsageMonitoringUpdates will contain only usage monitoring data.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: The update from PCF can have policy rule updates and also usage monitoring updates. If usage monitoring updates are not received then the responses will be 0. But we can still have rule updates which needs to be sent as a part of UsageMonitoringUpdateResponse proto. Hence adding a default response. Similar logic is [here](https://github.com/magma/magma/blob/ae3224e0f5bb71b03aade6f85e89d1112d42d511/feg/gateway/services/session_proxy/servicers/policy.go#L336) for Gx.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: This is used outside this module too for propagating the error received from PCF. See update_session.go `sendSingleSmPolicyUpdate()`.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/servicers/update_session.go
GANESH-WAVELABS: PostSmPoliciesSmPolicyIdUpdateWithResponse is N7 interface generated code. The if is for adding the HTTP error status to error.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
uri200: thanks. I now remember that. The problem in gx gy is I found our code was only taking into consideration one single CCA-U. So if PCRF sends more we are ignoring it. So to maintain it consistent with what we had on sessiond but have the way to send more than one credit on one single request, I had to do that. I guess n7 will suffer of the same issue due to the way sessiond reports the updates .

uri200 (approved):

uri200:
is this ready to be merged?

GANESH-WAVELABS:
> is this ready to be merged?

Yes is is ready to merge.

PR #10299, opened 2021-11-13
feat(feg): N7 Proxy SM Policy Delete implementation

GANESH-WAVELABS:
Signed-off-by: GANESH IRRINKI <ganesh.irrinki@wavelabs.ai>

## Summary
- Convert SessionTerminateRequest proto message received on the CentralSessionController to N7 SmPolicyDeleteData message.
- Using the N7 Openapi generated code send the SmPolicyDelete to PCF
- Check the N7 delete response and then construct SessionTerminateResponse proto and send it as a grpc response.

PR #10308, opened 2021-11-15
feat(feg): N7 Proxy handle PCF initiated updated and terminate

GANESH-WAVELABS:
## Summary
Implements the following
- Notification HTTP endpoints to receive N7 notifications from PCF
- Handle N7 SMPolicyUpdateNotify from PCF, fetch the session-id and send it to feg-relay via PolicyReAuth grpc.
- Handle the PolicyReAuth answer received from feg-relay and convert it to PartialSuccessReport and sent it to PCF as response
- Handle N7 SMPolicyTerminateNotify from PCF. etch the session-id and send it to feg-relay via AbortSession grpc.
- Handle AbortSession response and send the error code if any to PCF


uri200 (commented):
is notify the same as Update in 4g?

GANESH-WAVELABS:
> is notify the same as Update in 4g?

This is similar to RAR from PCRF.

feg/gateway/services/n7_n40_proxy/n7/n7_notify.go
```diff
@@ -0,0 +1,151 @@
...
+// 		return "", "", ctx.NoContent(http.StatusNotFound)
+// 	}
+// 	sessionId := obj.(string)
+// 	imsi, err := protos.GetIMSIwithPrefixFromSessionId(sessionId)
+// 	if err != nil {
+// 		glog.Errorf("Error retrieving IMSI from session ID %s: %s", sessionId, err)
+// 		return "", "", ctx.NoContent(http.StatusNotFound)
+// 	}
+// 	return sessionId, imsi, nil
+// }
```
uri200: I know this is a draft, but once you are done, remember to remove the commented stuff

feg/gateway/services/n7_n40_proxy/n7/n7_notify.go
```diff
@@ -0,0 +1,151 @@
...
+	"UNKNOWN",                // 18
+	"MISS_REDI_SER_ADDR",     // 19
+	"CM_END_USER_SER_DENIED", // 20
+	"CM_CREDIT_CON_NOT_APP",  // 21
+	"CM_AUTH_REJ",            // 22
+	"CM_USER_UNK",            // 23
+	"CM_RAT_FAILED",          // 224
+	"UNKNOWN",                // 25
+	"UNKNOWN",                // 26
+}
```
uri200: if those values are used by other files, maybe we can find a better place to put it

uri200: By the way, arent those values already defined  [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fsbi%2Fspecs%2FTS29512NpcfSMPolicyControlServer%2FTS29512NpcfSMPolicyControlServer.gen.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=37&start_col=3&end_row=37&end_col=3). Why do we need to convert them into ints?

Also, can you confirm those values come from the standard? ot is this an arbitrary value?

feg/gateway/services/n7_n40_proxy/n7/n7_notify.go
```diff
@@ -0,0 +1,151 @@
...
+
+func getFailureCodeN7(failureCodeProto protos.PolicyReAuthAnswer_FailureCode) *n7_sbi.FailureCode {
+	failureCodeNum := int(failureCodeProto)
+	if failureCodeNum > len(FailureCodeProtoToN7Map) {
+		return nil
+	}
+	retFailCode := FailureCodeProtoToN7Map[failureCodeNum]
+	return &retFailCode
+}
```
uri200: would it make sense to put those functions into conversions file?

feg/gateway/services/n7_n40_proxy/n7_n40_proxy/main.go
```diff
@@ -37,21 +39,37 @@ func main() {
...
+		glog.Fatalf("Creating N7 Client failed: %s", err)
+	}
+	sessController, err := servicers.NewCentralSessionController(n7config, dbClient, policyClient)
 	if err != nil {
 		glog.Fatalf("Error creating session controller in N7_N40 Proxy: %s", err)
 	}
+	notifHandler := notify.NewNotificationHandler(&n7config.Client, cloudReg)
+	if notifHandler == nil {
+		glog.Fatalf("Error creating noitification handler")
+	}
```
uri200: I think handlers shouldn't be added in main 

Check other systems how this works: see how clean is the s8_proxy main [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fs8_proxy%2Fs8_proxy%2Fmain.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=30&start_col=5&end_row=30&end_col=5). Note session_proxy main is a little mess. Lets try no to repeat the same 

See how the hanlders are added on s8_proxy [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fs8_proxy%2Fservicers%2Fgtp_handlers.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=25&start_col=10&end_row=25&end_col=10)

feg/gateway/services/n7_n40_proxy/n7_n40_proxy/main.go
```diff
@@ -37,21 +39,37 @@ func main() {
...
 
 	// Add GRPC handlers to the service
 	lteprotos.RegisterCentralSessionControllerServer(srv.GrpcServer, sessController)
 	protos.RegisterServiceHealthServer(srv.GrpcServer, sessController)
 
+	// Run the Notification Handler HTTP service
+	_, err = notifHandler.Start()
+	if err != nil {
+		glog.Fatalf("Failed to start Notification Handler")
+	}
```
uri200: same here. I think this should be part of the n7 proxy.

Or is will this server be shared with the n40?

[here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fsession_proxy%2Fcredit_control%2Fgx%2Fgx_client.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=86&start_col=23&end_row=86&end_col=23) is how one of the hanlders is added on session proxy. As you can see it is a little bit of a mess

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,210 @@
...
+				return addr.String(), nil
+			}
+		case err := <-errChan:
+			if err == http.ErrServerClosed {
+				return "", nil
+			}
+			return "", err
+		}
+	}
+}
```
uri200, line 184: I have seen this function before, maybe it would be good to make it part of the sbi client itself

feg/gateway/services/n7_n40_proxy/n7/n7_notify.go
GANESH-WAVELABS: These are from the standard and this slice is used to map proto FailureCode an integer to N7 failure code which is a string.

feg/gateway/services/n7_n40_proxy/n7_n40_proxy/main.go
GANESH-WAVELABS: N40 will also use the same notification handler.

NotificationHandler can be moved to session_controller.go. However in that case for create, delete or update tests (example create_session_test.go) the NotificationHandler will be started and has to be Closed after the test, even though it is not used. Will this be okay?

feg/gateway/services/n7_n40_proxy/n7/n7_notify.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7/n7_notify.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: Done

feg/gateway/sbi/gen.go
```diff
@@ -0,0 +1,63 @@
...
+//go:generate oapi-codegen -generate types,skip-prune -o specs/TS29122ReportingNetworkStatus/TS29122ReportingNetworkStatus.gen.go --import-mapping TS29122_CommonData.yaml:magma/feg/gateway/sbi/specs/TS29122CommonData,TS29571_CommonData.yaml:magma/feg/gateway/sbi/specs/TS29571CommonData defs/TS29122_ReportingNetworkStatus.yaml
+//go:generate oapi-codegen -generate types,skip-prune -o specs/TS29122GMDviaMBMSbyxMB/TS29122GMDviaMBMSbyxMB.gen.go --import-mapping TS29572_Nlmf_Location.yaml:magma/feg/gateway/sbi/specs/TS29572NlmfLocation,TS29122_CommonData.yaml:magma/feg/gateway/sbi/specs/TS29122CommonData,TS29571_CommonData.yaml:magma/feg/gateway/sbi/specs/TS29571CommonData defs/TS29122_GMDviaMBMSbyxMB.yaml
+//go:generate oapi-codegen -generate types,skip-prune -o specs/TS29510NnrfNFDiscovery/TS29510NnrfNFDiscovery.gen.go --import-mapping TS29510_Nnrf_NFManagement.yaml:magma/feg/gateway/sbi/specs/TS29510NnrfNFManagement,TS29503_Nudm_SDM.yaml:magma/feg/gateway/sbi/specs/TS29503NudmSDM,TS29571_CommonData.yaml:magma/feg/gateway/sbi/specs/TS29571CommonData defs/TS29510_Nnrf_NFDiscovery.yaml
+
+package sbi
```
uri200, line 63: Thanks. I think this way looks better

feg/gateway/sbi/sbi_test.go
```diff
@@ -15,17 +15,16 @@ import (
...
 
 	"github.com/labstack/echo/v4"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"magma/feg/gateway/sbi"
 	n7_client "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
 	n7 "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControlServer"
-	sbi "magma/feg/gateway/sbi/specs/TS29571CommonData"
+	sbi_common "magma/feg/gateway/sbi/specs/TS29571CommonData"
```
uri200, line 28: I am kind of lost with the specs numbers. Is `TS29571` n7 common data or sbi general common data

feg/gateway/sbi/sbi_test.go
```diff
@@ -125,34 +124,7 @@ func startServer(t *testing.T) *MockPcf {
 	}()
 
 	var err error
-	mockPcf.ListenAddr, err = waitForTestServerAndGetAddr(eserver, errChan)
+	mockPcf.ListenAddr, err = sbi.WaitForSbiServerAndGetAddr(eserver, errChan)
```
uri200: Sorry I know i commented about this function, but where is the PR with this change ?

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
```diff
@@ -55,6 +69,7 @@ func NewCentralSessionController(
 			RequestTimeout: DefaultN7Timeout,
 		},
 		healthTracker: metrics.NewSessionHealthTracker(),
+		notifHandler:  notifHandler,
```
uri200, line 67: why do you need to store `notifHandler` ?

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+		cloudRegistry: cloudReg,
+	}
+	urlDef, err := url.ParseRequestURI(config.NotifyApiRoot)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing notify api root - %s", err)
+	}
+	basePath := fmt.Sprintf("%s/:%s", urlDef.Path, EncodedSessionId)
+	// Register the handlers
+	handler.NotifyServer.POST(basePath+"/update", handler.PostSmPolicyUpdateNotification)
+	handler.NotifyServer.POST(basePath+"/terminate", handler.PostSmPolicyTerminateNotification)
```
uri200: Registering the handlers may be the most important thing this struct has. Maybe it would be good to set a different function for that. Or maybe it would be good to place it somewhere more visible.

Check in our codebase how other services does this (i think I shred some examples but i lost the comments)

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	"magma/feg/gateway/services/session_proxy/relay"
+	"magma/gateway/service_registry"
+	"magma/lte/cloud/go/protos"
+)
+
+const (
+	NOTIFY_PATH      = "/"
+	UnableToDeliver  = 3002
+	EncodedSessionId = "encodedSessionId"
+	ShutdownTimeout  = 10 * time.Second
```
uri200: what is this ShutdownTimeout?

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/labstack/echo/v4"
+
+	"magma/feg/gateway/sbi"
+	n7_client "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
+	"magma/feg/gateway/services/n7_n40_proxy/n7"
+	"magma/feg/gateway/services/session_proxy/relay"
+	"magma/gateway/service_registry"
+	"magma/lte/cloud/go/protos"
+)
+
```
uri200, line 34: I see this file defines the notify server and also defines the handlers. If hte handlers were bigger in size It would make sense to put htem in different files. But since they are small, maybe it is fine to leave them at the same file

However, since this file defines important stuff that sometimes is not easy to finde, maybe this file deserves a small sumary so the reader can understand what is going on here, saying that notify_handler creates a service to handle requests initiated from PCF, and which requests does it handle.


Like [this](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fswx_proxy%2Fservicers%2Fmultiple_swx_proxy_test.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=29&start_col=6&end_row=29&end_col=6)

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	UnableToDeliver  = 3002
+	EncodedSessionId = "encodedSessionId"
+	ShutdownTimeout  = 10 * time.Second
+)
+
+type NotificationHandler struct {
+	Config        *n7.N7ClientConfig
+	NotifyServer  *echo.Echo
+	cloudRegistry service_registry.GatewayRegistry
+}
```
uri200, line 53: make those properties lower case if you don't need to access them (i was not able to find a case to access them directly

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	}()
+
+	return sbi.WaitForSbiServerAndGetAddr(handler.NotifyServer, errChan)
+}
+
+func (handler *NotificationHandler) Shutdown() {
+	ctx, cancel := context.WithTimeout(context.Background(), ShutdownTimeout)
+	defer cancel()
+	handler.NotifyServer.Shutdown(ctx)
+}
```
uri200, line 99: why do you need Start/Shutdown?  What is the use case?

uri200: Maybe it would make sense to start the handler direclty on `NewNotificationHandler` or maybe create `NewStartedNotificationHandler` to encapsulate that

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+
+func (handler *NotificationHandler) PostSmPolicyUpdateNotification(ctx echo.Context) error {
```
uri200: make sure you don't export unnecessary functions. I think this should be lowercase

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	client, err := relay.GetSessionProxyResponderClient(handler.cloudRegistry)
+	if err != nil {
+		glog.Error(err)
```
uri200: since the error comes from feg_relay which is a grpc service running in a totally different gateway, when you print it in the log you will not have any reference where this was triggered in your code. So better add some context to the error 
```
glog.Error("PostSmPolicyUpdateNotification was not able to get relay client: %s", err)
```

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+		return ctx.NoContent(http.StatusBadRequest)
+	}
+	smPolicyDecision := smUpdateNotify.SmPolicyDecision
+	sessionId, imsi, err := getSessionIdAndIMSI(ctx.Param(EncodedSessionId))
+	if err != nil {
+		glog.Errorf("Unable to fetch session-id for UpdateNotify- %s", err)
+		return err
+	}
+
+	policyReauthProto := n7.GetPolicyReauthRequestProto(sessionId, imsi, smPolicyDecision)
```
uri200: use `smUpdateNotify.SmPolicyDecision`

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	defer client.Close()
+
+	var smUpdateNotify n7_client.SmPolicyNotification
+	err = ctx.Bind(&smUpdateNotify)
+	if err != nil {
+		glog.Errorf("Invalid SmPolicyNotification received: %s", err)
+		return ctx.NoContent(http.StatusBadRequest)
+	}
+	smPolicyDecision := smUpdateNotify.SmPolicyDecision
```
uri200: remove this

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	"magma/feg/gateway/services/n7_n40_proxy/n7"
+	"magma/feg/gateway/services/session_proxy/relay"
+	"magma/gateway/service_registry"
+	"magma/lte/cloud/go/protos"
+)
+
+const (
+	NOTIFY_PATH      = "/"
+	UnableToDeliver  = 3002
+	EncodedSessionId = "encodedSessionId"
```
uri200, line 44: what is this field? Does it come from some kind of sbi spec?

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	"magma/feg/gateway/sbi"
+	n7_client "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
+	"magma/feg/gateway/services/n7_n40_proxy/n7"
+	"magma/feg/gateway/services/session_proxy/relay"
+	"magma/gateway/service_registry"
+	"magma/lte/cloud/go/protos"
+)
+
+const (
+	NOTIFY_PATH      = "/"
```
uri200: i may be missing it but where is this used?

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	return err
+}
+
+func getSessionIdAndIMSI(encSessionId string) (string, string, error) {
+	if len(encSessionId) == 0 {
+		return "", "", echo.NewHTTPError(http.StatusBadRequest, "Invalid path parameter encodedSessionId Param")
+	}
+	sessionIdBytes, err := base64.URLEncoding.DecodeString(encSessionId)
+	if err != nil {
+		return "", "", echo.NewHTTPError(http.StatusBadRequest, "Invalid path parameter unable to decode session-id")
```
uri200: use `sessionID`

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+		err = ctx.NoContent(http.StatusNoContent)
+	default:
+		err = ctx.NoContent(http.StatusNoContent)
+	}
+	return err
+}
+
+func getSessionIdAndIMSI(encSessionId string) (string, string, error) {
+	if len(encSessionId) == 0 {
+		return "", "", echo.NewHTTPError(http.StatusBadRequest, "Invalid path parameter encodedSessionId Param")
```
uri200: Shouldn't it this be something like
```
Invalid path parameter: encodedSessionId
```

Is `Param` part of the name?

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return "", "", echo.NewHTTPError(http.StatusBadRequest, "Invalid path parameter unable to decode session-id")
+	}
+	sessionId := string(sessionIdBytes)
+	imsi, err := protos.GetIMSIwithPrefixFromSessionId(sessionId)
+	if err != nil {
+		return "", "", echo.NewHTTPError(http.StatusBadRequest, "Invalid session-id unable to decode imsi")
+	}
+	return sessionId, imsi, nil
+}
```
uri200, line 203: you are hiding errors: you catch the error and return a total different error. Either you print it or you embed the error on your message (this second option should be chosen here). Handling the error this way you will never know why the previous call failed

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: Why is this error mentioning `path parameter`

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+		err = ctx.NoContent(http.StatusInternalServerError)
+	case protos.AbortSessionResult_SESSION_REMOVED:
+		err = ctx.NoContent(http.StatusNoContent)
+	default:
+		err = ctx.NoContent(http.StatusNoContent)
+	}
+	return err
+}
+
+func getSessionIdAndIMSI(encSessionId string) (string, string, error) {
```
uri200: this is not very nice `(string, string, error)`

add a doc string 
or 
```
func getSessionIdAndIMSI(encSessionId string) (sessionId, imsi string, error)
```

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: why does this function needs to return an  `echo.NewHTTPError`?  Since all of them are `StatusBadRequest` can't not you just do it at caller level?

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		glog.Errorf("Invalid SmPolicyNotification received: %s", err)
+		return ctx.NoContent(http.StatusBadRequest)
+	}
+
+	// smPolicyUrl := string(smTermNotify.ResourceUri)
+	sessionId, imsi, err := getSessionIdAndIMSI(ctx.Param(EncodedSessionId))
+	if err != nil {
+		glog.Errorf("Unable to fetch session-id for TerminateNotify - %s", err)
+		return err
```
uri200: here is where you should convert your error into http code (unles the reason is different for each error this function returns)

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+		glog.Error(err)
+		return err
+	}
+	defer client.Close()
+
+	var smTermNotify n7_client.TerminationNotification
+	err = ctx.Bind(&smTermNotify)
+	if err != nil {
+		glog.Errorf("Invalid SmPolicyNotification received: %s", err)
+		return ctx.NoContent(http.StatusBadRequest)
```
uri200: why dont you add the `err` into the return?

uri200 (changes requested):

feg/gateway/services/n7_n40_proxy/n7_n40_proxy/main.go
```diff
@@ -51,15 +51,13 @@ func main() {
 	if err != nil {
 		glog.Fatalf("Creating N7 Client failed: %s", err)
 	}
-	sessController, err := servicers.NewCentralSessionController(n7config, dbClient, policyClient)
+	sessController, err := servicers.NewCentralSessionController(n7config, dbClient, policyClient, cloudReg)
```
uri200, line 54: why do you need to pass cloudReg here?

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: add a docstring describing what handlers handle. Is this handleing the update response? Or this update something like Rar

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	policyReauthProto := n7.GetPolicyReauthRequestProto(sessionId, imsi, smPolicyDecision)
+	ans, err := client.PolicyReAuth(context.Background(), policyReauthProto)
+	if err != nil {
+		glog.Errorf("Error relaying N7 policy reauth request to gateway: %s", err)
+		return err
+	}
+	partialSuccessReport := n7.GetPartialSuccessReportN7(ans)
+	if partialSuccessReport == nil {
+		return ctx.NoContent(http.StatusOK)
+	}
```
uri200, line 134: this is too convoluted.  `GetPartialSuccessReportN7` does to many thing. The return is also doing to many things, not only provides the value but serves as an error. 

Why don't you create a function in this file to process the response from `PolicyReAuth` which will return eitner `ctx.JSON(http.StatusOK, partialSuccessReport)` or `ctx.NoContent(http.StatusOK)`.  You can put part of the logic of `GetPartialSuccessReportN7` in this function maybe.

Note in go it is better to return error inside the iff, and return the proper case outside the iff (here is done reversed)

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: It is SBI Common data. 29.512 is for N7.

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: #9524

feg/gateway/services/n7_n40_proxy/n7_n40_proxy/main.go
GANESH-WAVELABS: cloudReg is required for creating NotificationHandler. NotificationHandler uses it for getting SessionProxyReponderClient.

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: The HTTP Echo server can be gracefully terminated using Shutdown() method. Sometimes this might take more time and may not complete. Taken from this [recipe](https://echo.labstack.com/cookbook/graceful-shutdown/).

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: This is the part of the notificationUrl sent to PCF during the Sm Policy Create request. PCF does not know about Session-ID. For each policy session it maintains a PolicyId. The N7_proxy extracts Session-ID from this url path element. This is not part of SBI spec, we are using this to get back Session-ID for the session.  

The notification url is of the form {notifyRoot}/{encodedSessionId}/{operation}
  Example: http://magma-feg.magma.com/sm-policy-control/v1/MTIzNDU2Nzg5MDsxMjM0NQo=/update
  Here notifyRoot = http://magma-feg.magma.com/sm-policy-control/v1
        encodedSessionId = MTIzNDU2Nzg5MDsxMjM0NQo= (Session-Id is urlencoded)
        operation = update

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: Shutdown is used for graceful termination of the HTTP server. This is specifically useful in case of unit-tests where if Shutdown is not used, the old server will still be running and may cause bind errors.

The flow is Create server, register handlers and then start. If register handlers are made a separate method then Start also needs to a separate method.

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: This is handling something like RAR

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: done

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: done

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: 'GetPartialSuccessReportN7' is a conversion function. Hence it is in model_conversion. From N7_proxy point of view partialSuccessReport is not an error. It merely translates what was received from sessiond.

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
GANESH-WAVELABS: This is used for graceful Shutdown of the http server.

feg/gateway/sbi/sbi_test.go
uri200: ohh i think i saw it somewhere else and I commented you should place this function on  `sbi.go`. So this way other modules can use it.

Note it would be good to create a wrapper over sbi so we can add stuff, like log printing (defined here https://github.com/magma/magma/issues/10598). So basically we would need some kind of struct called `sbi_client` which would have functions like `WaitForSbiServerAndGetAddr` of `print_http_request/response`

uri200: btw, on my master i do not find that `sbi.WaitForSbiServerAndGetAddr`. Can you point me to the definition

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -30,6 +30,35 @@ import (
...
+	"UNKNOWN",                // 18
+	"MISS_REDI_SER_ADDR",     // 19
+	"CM_END_USER_SER_DENIED", // 20
+	"CM_CREDIT_CON_NOT_APP",  // 21
+	"CM_AUTH_REJ",            // 22
+	"CM_USER_UNK",            // 23
+	"CM_RAT_FAILED",          // 24
+	"UNKNOWN",                // 25
+	"UNKNOWN",                // 26
+}
```
uri200, line 62: i remember i commented this in another PR

Why do you need that list here? Why can't you just forward the string?  Will sessiond need to know what those number mean and transform it back ? Are those values from the standard?

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -67,6 +96,30 @@ func GetSmPolicyContextDataN7(
...
+			FailureCode: getFailureCodeN7(failureCode),
+		}
+		ruleReports = append(ruleReports, ruleReport)
+	}
+
+	return &n7_sbi.PartialSuccessReport{
+		FailureCause: n7_sbi.FailureCausePCCRULEEVENT,
+		RuleReports:  &ruleReports,
+	}
+}
```
uri200, line 129: this function is doing too many things: it is finding if errors and then translating them into something.

Not only that, but the result value is used for 2 different purposes: if ni then we are good (any pattern to avoid), if not nil, then there is an error in fact.

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: IN CONVERSIONS.GO
```
// BuildPartialSuccessReportN7 define this function builds a partial success repoirt. 
// Highlight that this function will generate a report no matter if the answer was really partial
func BuildPartialSuccessReportN7(failedRules *protos.Rules (???)) *n7_sbi.PartialSuccessReport {
	failedRuleReports := []n7_sbi.RuleReport{}
	for ruleId, failureCode := range failedRules {
		failedRuleReport := n7_sbi.RuleReport{
			PccRuleIds:  []string{ruleId},
			RuleStatus:  n7_sbi.RuleStatusINACTIVE,
			FailureCode: getFailureCodeN7(failureCode),
		}
		failedRuleReports = append(ruleReports, failedRuleReport)
	}

	return &n7_sbi.PartialSuccessReport{
		FailureCause: n7_sbi.FailureCausePCCRULEEVENT,
		RuleReports:  &failedRuleReports,
	}
}
```

IN NOTIFY_HANDLER 
```
    	if isItPartialSuccess(ans) {
		ctx.JSON(http.StatusOK,, n7.GetPartialSuccessReportN7(ans))
	}

	return ctx.NoContent(http.StatusOK)
}

//(maybe that could also go in CONVERSIONS.GO)
// isItPartialSuccess define this clearly
func isItPartialSuccess(raa *protos.PolicyReAuthAnswer) bool; {
	if raa.Result == protos.ReAuthResult_UPDATE_INITIATED ||
		raa.Result == protos.ReAuthResult_UPDATE_NOT_NEEDED ||
		len(raa.FailedRules) == 0 {
		// no errors to send
		return true
	}
	return false 
}


```



Maybe we should start thinking on a `conversion.go` file under sbi 
See how we extend swagger auto generated objects [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fcloud%2Fgo%2Fservices%2Ffeg%2Fobsidian%2Fmodels%2Fconversion.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=38&start_col=0&end_row=38&end_col=0). Sme we do with some autogenerated protos [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fcloud%2Fgo%2Fprotos%2Fmock_core_conversions.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=13&start_col=10&end_row=13&end_col=10)

uri200: Lets do the change as it is but I think adding `isItPartialSuccess` to a  `conversions.go` may be the best . Do you know if this is partial logic is used somewhere else

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,206 @@
...
+		return fmt.Errorf("internal server error")
+	}
+	partialSuccessReport := n7.GetPartialSuccessReportN7(ans)
+	if partialSuccessReport != nil {
+		ctx.JSON(http.StatusOK, partialSuccessReport)
+	}
+	return ctx.NoContent(http.StatusOK)
+}
+
+func (handler *NotificationHandler) postSmPolicyTerminateNotification(ctx echo.Context) error {
```
uri200, line 139: as mentioned, add description in all the handlers

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,206 @@
...
+	defer cancel()
+	handler.notifyServer.Shutdown(ctx)
+}
+
+// postSmPolicyUpdateNotification handles SM policy update notification request from PCF.
+func (handler *NotificationHandler) postSmPolicyUpdateNotification(ctx echo.Context) error {
+	client, err := relay.GetSessionProxyResponderClient(handler.cloudRegistry)
+	if err != nil {
+		glog.Errorf("postSmPolicyUpdateNotification failed to get SessionProxyResponderClient: %s", err)
+		return fmt.Errorf("internal server error")
```
uri200, line 106: what is the logic on not returning the error string here?
Will the PCF complain?

Note i mean all the returns in here

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,206 @@
...
+	err = ctx.Bind(&smUpdateNotify)
+	if err != nil {
+		err = fmt.Errorf("invalid SmPolicyNotification received: %s", err)
+		glog.Errorf("postSmPolicyUpdateNotification: %s", err)
+		return echo.NewHTTPError(http.StatusBadRequest, err)
+	}
+	sessionId, imsi, err := getSessionIdAndIMSI(ctx.Param(EncodedSessionId))
+	if err != nil {
+		glog.Errorf("postSmPolicyUpdateNotification unable to fetch session-id for UpdateNotify - %s", err)
+		return echo.NewHTTPError(http.StatusBadRequest, err)
```
uri200, line 120: why some returns are `echo.NewHTTPErro` and some others are not?

Check all of them

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: i dont understand the logic: when there is a partial success we modify the contex? but still return ok?

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,206 @@
...
+	ans, err := client.PolicyReAuth(context.Background(), policyReauthProto)
+	if err != nil {
+		glog.Errorf("Error relaying N7 policy reauth request to gateway: %s", err)
+		return fmt.Errorf("internal server error")
+	}
+	partialSuccessReport := n7.GetPartialSuccessReportN7(ans)
+	if partialSuccessReport != nil {
+		ctx.JSON(http.StatusOK, partialSuccessReport)
+	}
+	return ctx.NoContent(http.StatusOK)
```
uri200, line 135: This function seems to need more consistency on the returns:  either we return http errors, or fplain errors or ctx errors. 

why are you responding with this code?  Isn't nil enough

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,206 @@
...
+	default:
+		err = ctx.NoContent(http.StatusNoContent)
+	}
+	return err
+}
+
+func getSessionIdAndIMSI(encSessionId string) (sessionId string, imsi string, err error) {
+	if len(encSessionId) == 0 {
+		return "", "", fmt.Errorf("encodedSessionId path parameter empty")
+	}
```
uri200, line 190: strings default value is ""
err default value is nil

```
func getSessionIdAndIMSI(encSessionId string) (sessionId string, imsi string, err error) {
	if len(encSessionId) == 0 {
                err =  fmt.Errorf("encodedSessionId path parameter empty")
		return 
	}
.....
```

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,206 @@
...
+		err = ctx.NoContent(http.StatusInternalServerError)
+	case protos.AbortSessionResult_RADIUS_SERVER_ERROR:
+		err = ctx.NoContent(http.StatusInternalServerError)
+	case protos.AbortSessionResult_SESSION_REMOVED:
+		err = ctx.NoContent(http.StatusNoContent)
+	default:
+		err = ctx.NoContent(http.StatusNoContent)
+	}
+	return err
+}
```
uri200, line 184: can we be consistent on how we build the errors we return

Shouldn't those errors contain the string of the error itself too? (i dont remember if the string can be embedded on http error)

If you have a logic why you are adding the errors this way, please share it

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
```diff
@@ -46,7 +50,18 @@ func NewCentralSessionController(
...
+	notifHandler := notify.NewNotificationHandler(&n7config.Client, cloudReg)
+	err := notifHandler.RegisterHandlers()
+	if err != nil {
+		return nil, err
+	}
+	// Start the notification handler HTTP service
+	_, err = notifHandler.Start()
+	if err != nil {
+		return nil, fmt.Errorf("notification handler failed to start - %s", err)
+	}
```
uri200: As mentioned This logic here should be placed on the NewNotificationHandler.
 SessionController should receive an already working object with all the hanlders mounted

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: It makes sense but , as mentioned, try to encapsulate it

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,206 @@
...
+func NewNotificationHandler(
+	config *n7.N7ClientConfig,
+	cloudReg service_registry.GatewayRegistry,
+) *NotificationHandler {
+	return &NotificationHandler{
+		config:        config,
+		notifyServer:  echo.New(),
+		cloudRegistry: cloudReg,
+	}
+}
```
uri200, line 73: I still think that NewNotificationHandler could encapsulate all that (if you want you can create a new function called NewStartedNotificationHandlerWithHanlders which will run NewNotificationHandler and add the handlers and start

Will you ever need a non registered, non started handler?

A hint for that is that you are repeating the same code in the test than in the controller. Remember, if you see repeated code, you probably are missing an opportunity to encapsulate an object

feg/gateway/services/n7_n40_proxy/notify/notify_handler_test.go
```diff
@@ -0,0 +1,158 @@
...
+	ActTime       = time.Unix(1634906551, 0)
+	DeactTime     = time.Unix(1634913751, 0)
+	UnkRuleId     = n7_sbi.FailureCodeUNKRULEID
+	IncorrectFlow = n7_sbi.FailureCodeINCORFLOWINFO
+)
+
+func TestUpdateNotify(t *testing.T) {
+	sm, cloudRegistry := relay_mocks.StartMockSessionProxyResponder(t)
+	notifyHandler := notify.NewNotificationHandler(getClientConfig(), cloudRegistry)
+	err := notifyHandler.RegisterHandlers()
```
uri200: here you have the repeated code

uri200 (changes requested):

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: In the master the function was part of [feg/gateway/sbi/sbi_test.go ](https://github.com/magma/magma/blob/fee5c727d358c31b935ca8d5dc3fad13fcebb73d/feg/gateway/sbi/sbi_test.go#L134). This function is moved to sbi as part of this PR.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: This was the reply I had given 

> These are from the standard and this slice is used to map proto FailureCode an integer to N7 failure code which is a string

The proto FailureCode comes from SessionD and it is getting converted to N7 FailureCode which is a string. The string values defined in the proto and N7 does not match. Hence this table.

The proto number to string has the following [values](https://github.com/magma/magma/blob/fee5c727d358c31b935ca8d5dc3fad13fcebb73d/lte/cloud/go/protos/session_manager.pb.go#L896) which are not the same as N7 Failure Code strings.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: The PartialSuccessReport actually reports a Partial Error Report.

This method gets the failed rules sent by SessionD in proto RAA and then sends them as a report to PCF. It should be seen as only a translation function.

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: Not used anywhere, removing it.

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: The PCF does not complain, we will be providing details which the PCF would not understand and can't take action based on the error string.. We can get the cause of the error from our logs.

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: This is the success response. It returns no content and a 200 OK. NewHTTPError is used for returning failure response. When there is a content to be sent and success response we are responding with ctx.JSON.

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: > i dont understand the logic: when there is a partial success we modify the contex? but still return ok?

The PartialSuccessReport actually contains a partial error report. So when it is partial success, it means that there were errors installing the rule. If there are no FailedRules then it means that it is completely successful.

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/notify/notify_handler_test.go
GANESH-WAVELABS: Changed

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: For a successful response the Terminate does not have any content. For a error response, we need to send appropriate status code. However we can add error string in the response body.

feg/gateway/sbi/sbi_test.go
uri200: there is something wrong here. The function you point is in lower case and it is part of test file

This function you are using here is part of sbi package and it is exported. So the question is, where `WaitForSbiServerAndGetAddr` (capitalized, is defined). Is this on some other PR?



image

feg/gateway/sbi/sbi_test.go
```diff
@@ -125,34 +124,7 @@ func startServer(t *testing.T) *MockPcf {
...
 	require.NoError(t, err)
 	return mockPcf
 }
-
```
uri200, line 132: so where did this function go?

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
uri200: can you point me to the answer? I was looking that PR. That is the one that also has my comments on `waitForTestServerAndGetAddr`

uri200: Why do you need to define them here instead of defining this on proto file?
Why `PolicyReAuthAnswer_FailureCode_value` is defined in protos and not this one?

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
```diff
@@ -132,6 +191,15 @@ func getSbiPduSesionType(pduSessionType protos.PduSessionType) sbi.PduSessionTyp
 	}
 }
 
+func getFailureCodeN7(failureCodeProto protos.PolicyReAuthAnswer_FailureCode) *n7_sbi.FailureCode {
+	failureCodeNum := int(failureCodeProto)
+	if failureCodeNum > len(FailureCodeProtoToN7Map) {
+		return nil
```
uri200, line 207: doing things like that we are hiding errors. 
If for some reason we receive wrong formed value we just ignore, and don't say anything about it. 

I think some functions on `model_conversion` will need to be refactored to trigger error and stop process. It doesnt make any sense to send to AGW a proto that we know it is not formed properly.

So either PCF sends a proper message, or we error

uri200: Please, let's create another PR to refactor moder_conversion to add error in function that deserve to error. For example this one.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
uri200: That looks better. Thanks

feg/gateway/services/n7_n40_proxy/n7_n40_proxy/main.go
uri200: can't not NotificationHandler just own it ? Why do we need to pass it

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: now i see your comment below

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,222 @@
...
+	return handler, err
+}
+
+// RegisterHandlers registers the UpdateNotification and TerminateNotification handlers.
+// The notification url is of the form {notifyRoot}/{encodedSessionId}/{operation}
+//   Example: http://magma-feg.magma.com/sm-policy-control/v1/MTIzNDU2Nzg5MDsxMjM0NQo=/update
+//   Here notifyRoot = http://magma-feg.magma.com/sm-policy-control/v1
+//        encodedSessionId = MTIzNDU2Nzg5MDsxMjM0NQo= (Session-Id is urlencoded)
+//        operation = update
+// This notification url is send to PCF in the SmPolicyCreate request
```
uri200, line 84: Thanks for the description.
Lets put this 
```
// RegisterHandlers registers the UpdateNotification and TerminateNotification handlers.
// The notification url is of the form 
//	{notifyRoot}/{encodedSessionId}/{operation}
// Example: 
//   	http://magma-feg.magma.com/sm-policy-control/v1/MTIzNDU2Nzg5MDsxMjM0NQo=/update
// where 
//	notifyRoot = http://magma-feg.magma.com/sm-policy-control/v1
//      encodedSessionId = MTIzNDU2Nzg5MDsxMjM0NQo= (urlencoded Session-Id)
//      operation = update
// This notification url is send to PCF in the SmPolicyCreate request
```

uri200: what does this mean
```
// This notification url is send to PCF in the SmPolicyCreate request
```

Is't the notification sent by the PCF to n7_proxy?

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
```diff
@@ -0,0 +1,222 @@
...
+func (handler *NotificationHandler) registerHandlers() error {
+	urlDef, err := url.ParseRequestURI(handler.config.NotifyApiRoot)
+	if err != nil {
+		return fmt.Errorf("error parsing notify api root - %s", err)
+	}
+	basePath := fmt.Sprintf("%s/:%s", urlDef.Path, EncodedSessionId)
```
uri200: maybe you can use `"path"`  module here too

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: now I understand start and and stop.

I think this would look nicer if we create a sbi client. This wway we will be able to reuse it on other interfaces.
Functions to include there would be start, stop,  WaitForSbiServerAndGetAddr

We did that for [GTP](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fgtp%2Fgtp_client.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=51&start_col=11&end_row=51&end_col=11) client

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: ok now it make more sense. Before it was confusing what it was being returned in case of partial success

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: can we just return in the switch?
```
	switch ans.Code {
	case protos.AbortSessionResult_SESSION_NOT_FOUND:
		return echo.NewHTTPError(http.StatusNotFound, "Session not found")
	case protos.AbortSessionResult_USER_NOT_FOUND:
...

```
Simpler to read

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: This function moved to sbi.go as a part of a review [comment](https://github.com/magma/magma/pull/10308#discussion_r758964409) given in this PR.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: > can you point me to the answer? I was looking that PR. That is the one that also has my comments on

[Here](https://github.com/magma/magma/pull/10308#discussion_r759484769) is that outdated comment.

> Why do you need to define them here instead of defining this on proto file?

This is a response that is sent by SessionD and needs to get translated to N7. The proto definitions sent from SessionD are named differently, hence this translation from proto to N7.

feg/gateway/services/n7_n40_proxy/n7/model_conversion.go
GANESH-WAVELABS: > So either PCF sends a proper message, or we error

In this particular case the message is sent by SessionD. Again this comes in a response message from SessionD and we can send back an error to SessionD. We can probably log the error though.

feg/gateway/services/n7_n40_proxy/n7_n40_proxy/main.go
GANESH-WAVELABS: Also used to create mock cloud registry in case of unit tests and pass it as an argument. cloudReg is also used when creating NewRedisPolicyDBClient() in main.go.

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: the notification is sent by the PCF. But the notification url string is sent by n7_proxy to PCF as a part of initial Policy Session create request. This url will then by used by the PCF to send the notifications related to the policy session to N7 proxy.

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/n7_n40_proxy/main.go
uri200: The problem i have here is how the test is making us modify the api, which is not good

For that reason in other services we created mock ernd points (for example HSS for s6a or PGW for s8_client)

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
```diff
@@ -97,3 +106,7 @@ func (srv *CentralSessionController) TerminateSession(
 
 	return (&protos.UnimplementedCentralSessionControllerServer{}).TerminateSession(ctx, request)
 }
+
+func (srv *CentralSessionController) Close() {
+	srv.notifHandler.Shutdown()
+}
```
uri200, line 139: this is good. This method may be missing in some of our services.

The reason we do not care much about shutting down is that those services run on docker, so when the service is shut down, the whole container is shut down, so no routines are left behind

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
```diff
@@ -97,3 +106,7 @@ func (srv *CentralSessionController) TerminateSession(
 
 	return (&protos.UnimplementedCentralSessionControllerServer{}).TerminateSession(ctx, request)
 }
+
+func (srv *CentralSessionController) Close() {
```
uri200, line 137: add a comment on this function mentioning this is to gracefully close CentralSessionController. If we ever have to add to add more tasks to close we will know where to add them

Also note that it would be good to have a sbi client that would take care of that instead of placing this on the n7 level

feg/gateway/services/n7_n40_proxy/n7_n40_proxy/main.go
uri200: Adding a defer on the test is the right way better than modifying the design fo the test
If you are concerned about routines left behind you can check this

`"go.uber.org/goleak"`

https://medium.com/analytics-vidhya/goroutines-unit-testing-handling-unexpected-goroutines-dd0003be193c

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: Can you update the comment using the format provided please. It will make it more readable

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
uri200: Not sure how this is implemented, but pre my understanding we have http error code that PCF should be using to understand the type of error. It is true that PCF doesn't need to know about our implementation. But why do we care on the specific message?

uri200:
Just modify that [comment](https://github.com/magma/magma/pull/10308#discussion_r764855150) and we can land it

Is there any order those pr must be merged?

feg/gateway/services/n7_n40_proxy/notify/notify_handler.go
GANESH-WAVELABS: Done

GANESH-WAVELABS:
> Is there any order those pr must be merged?

#10299 should be merged before this.

uri200:
you may need to fix the merge conflicts, #10299 was merged first

uri200 (approved):

uri200 (approved):

PR #10232, opened 2021-11-10
chore: prototype the usage of composite github actions

maxhbr:
**This is WIP / just an experiment**

## Summary

This tries to replace https://github.com/magma/magma/pull/9787 and https://github.com/magma/magma/pull/9723 with the use of composite actions.

## Additional Information

- [ ] This change is backwards-breaking

PR #10312, opened 2021-11-15
feat: create new Github action for bazel docker image build

Neudrino:
## Summary

* A new Github action is added to build the [Bazel base docker image](https://github.com/magma/magma/blob/master/experimental/bazel-base/Dockerfile).

## Test Plan

* The Github Action defined here runs fine and builds a docker image. See [this run](https://github.com/magma/magma/runs/4222135094?check_suite_focus=true).
* ℹ️ There were temporary builds on PR added for testing of the CI. However, usually the build is only intended for `master` as PR builds are expected to fail on GHCR push due to inability to access credentials.

## Additional Information

* This PR is composite actions ready, i.e. it can easily refactored to introduce and use a composite step for building docker images. Also see https://github.com/magma/magma/pull/10232 for discussion on composite actions. I.e. More base image builds could utilize the same workflow with only minimal CI configuration.

### DIFF
--- a/.github/workflows/docker-bazel-base-builder.yml
+++ b/.github/workflows/docker-bazel-base-builder.yml
@@ -1,0 +1,51 @@
+---
+name: "Build Bazel base Docker image"
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - master
+    paths:
+      - experimental/bazel-base/Dockerfile
+      - .github/workflows/docker-bazel-base-builder.yml
+  schedule:
+    - cron: '13 3 * * *'
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_STREAM: ${{ github.repository }}/bazel-base
+  IMAGE_TAGS: type=sha
+  DOCKERFILE: experimental/bazel-base/Dockerfile
+
+jobs:
+  build_dockerfile:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Out Repo
+        uses: actions/checkout@v2
+      - name: Set up Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          # defines the image stream the image is pushed to
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_STREAM }}
+          # defines the image tags added to the image
+          tags: ${{ env.IMAGE_TAGS }}
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to GHCR
+        uses: docker/login-action@v1
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push Docker image
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ${{ env.DOCKERFILE }}
+          # without docker/metadata-action will only be accessible by hash
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/experimental/bazel-base/Dockerfile
+++ b/experimental/bazel-base/Dockerfile
@@ -26,7 +26,7 @@ RUN apt-get update && \
         unzip \
         vim \
         wget \
-        zip \
+        zip
 
 # Install bazel
 WORKDIR /usr/sbin


.github/workflows/docker-bazel-base-builder.yml
```diff
@@ -0,0 +1,51 @@
...
+          username: ${{ github.repository_owner }}
+          password:  ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push Docker image
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ${{ env.DOCKERFILE }}
+          # without docker/metadata-action will only be accessible by hash
+          push: true # will lead to authentication error when startet from PR
```
maxhbr: I think this rule will not run for PRs due to:
```
  push:
    branches:
      - master
  ```
  So It will only run for pushes to master. For PRs one could use something like https://github.com/magma/magma/blob/b737e22959a49b8585ae4f25b2c871ae9a3ff8e2/.github/workflows/agw-workflow.yml#L9-L13

maxhbr (commented):
Please fix linter warnings

.github/workflows/docker-bazel-base-builder.yml
Neudrino: To check the build validity, the build was intermediatly be run on PR in https://github.com/magma/magma/commit/69520081f036ebf096d97b80f6a598c29cc7b651.

electronjoe:
> * This PR is composite actions ready

Awesome!

It will be good to have a minimum viable docker container that can bazel build our targets and which we can use in CI. Thanks for the PR.

electronjoe (approved):

themarwhal (approved):
<3 Amazing! Thanks for adding this

themarwhal:
Any thoughts about consolidating this workflow with the existing Docker push workflow? (@Neudrino)

themarwhal:
> Any thoughts about consolidating this workflow with the existing Docker push workflow? (@Neudrino)

https://github.com/magma/magma/blob/master/.github/workflows/docker-push.yml 

^Might be good to refactor this workflow so that we can easily add new docker images by just specifying dockerfiles.

electronjoe:
> ^Might be good to refactor this workflow so that we can easily add new docker images by just specifying dockerfiles.

Yeah I thought about that, but wasn't sure how cleanly the refactoring could re-use the workflow. Maybe an experiment to see how that looks would be a good step before we merge this? Edit: doing things twice doesn't necessarily require re-use and refactoring. But by the third time it's fix-up time =D

Neudrino:
> Any thoughts about consolidating this workflow with the existing Docker push workflow? (@Neudrino)

Yes, this was exactly my idea, and that was what I meant with "composite actions ready". There is only a small amount of refactoring needed to use composite actions. However, as they were still discussed in https://github.com/magma/magma/pull/10232, I was inclined not to introduce them here through the back door, if they were still out for debate.

The idea would be to have a reusable action doing the docker build (and push) and reusing it for the actions in the docker-push workflow mentioned above. In fact, this PR/workflow was designed with that in mind.

In case Github composite actions are accepted, I would follow the step by step approach and still merge this PR, and in addition one could do a follow up with a new one, doing the refactoring separate from providing the new functionallity. 🙂

themarwhal:
> > Any thoughts about consolidating this workflow with the existing Docker push workflow? (@Neudrino)
> 
> Yes, this was exactly my idea, and that was what I meant with "composite actions ready". There is only a small amount of refactoring needed to use composite actions. However, as they were still discussed in #10232, I was inclined not to introduce them here through the back door, if they were still out for debate.
> 
> The idea would be to have a reusable action doing the docker build (and push) and reusing it for the actions in the docker-push workflow mentioned above. In fact, this PR/workflow was designed with that in mind.
> 
> In case Github composite actions are accepted, I would follow the step by step approach and still merge this PR, and in addition one could do a follow up with a new one, doing the refactoring separate from providing the new functionallity. 🙂

Awesome

> > Any thoughts about consolidating this workflow with the existing Docker push workflow? (@Neudrino)
> 
> Yes, this was exactly my idea, and that was what I meant with "composite actions ready". There is only a small amount of refactoring needed to use composite actions. However, as they were still discussed in #10232, I was inclined not to introduce them here through the back door, if they were still out for debate.
> 
> The idea would be to have a reusable action doing the docker build (and push) and reusing it for the actions in the docker-push workflow mentioned above. In fact, this PR/workflow was designed with that in mind.
> 
> In case Github composite actions are accepted, I would follow the step by step approach and still merge this PR, and in addition one could do a follow up with a new one, doing the refactoring separate from providing the new functionallity. 🙂

Ok! If that is being considered. I am ok with landing this as it is. What I would like to have is to have  a single workflow that builds the containers for build health on pull request targets (similar to what we do with the `dockerfile-health.yml` workflow) and pushes to ghcr on push to master. We can simplify and dedup most of the workflow by using github composite actions it seems! (which is very exciting)

rdefosse (approved):

Neudrino:
> Ok! If that is being considered. I am ok with landing this as it is. What I would like to have is to have a single workflow that builds the containers for build health on pull request targets (similar to what we do with the `dockerfile-health.yml` workflow) and pushes to ghcr on push to master. We can simplify and dedup most of the workflow by using github composite actions it seems! (which is very exciting)

Thanks for landing this PR. It is good to see, that the docker build and push works as a first step ([docker image here](https://github.com/magma/magma/pkgs/container/magma%2Fbazel-base)).
A refactoring towards Github composite actions, which also includes the `docker-health.yml` is proposed in https://github.com/magma/magma/pull/10378.

PR #10317, opened 2021-11-15
fix(cwg): guard cwf-integ-test assertions from out of index

themarwhal:

## Summary
Noticed that in some failure cases we get to out of index errors: https://github.com/magma/magma/runs/4197587303?check_suite_focus=true

This is bad because this will make the test run abort without trying the failed tests again.

## Additional Information


uri200 (approved):
good catch!

`assert.require` would have made the trick, but I don't think it makes sense here due to the nature of the test

PR #10326, opened 2021-11-15
chore(ci): use devcontainer image for mme clang tidy / warning

themarwhal:

## Summary
1. Migrate all jobs to use the devcontainer image. This will allow us to just use a pre-built image and not build the docker container every time. (We utilize cache to speed this up, but since our cache is limited I'd rather use it for other things :) )
2. Remove Dockerfile.ubuntu20.04 as it is no longer used by CI. 
3. Split mme clang-tidy/warning collection from cpp-codecov as they are both long (and we're no longer constrained by gha nodes)

## Test Plan
Job went from 2hrs to 1hr 15min. (20% shorter)

Before

Screen Shot 2021-11-15 at 6 28 09 PM
After

Screen Shot 2021-11-15 at 6 28 43 PM

## Additional Information


tmdzk (approved):

rdefosse (approved):
@themarwhal 

OK to remove `lte/gateway/docker/mme/Dockerfile.ubuntu20.04` file.

Not used by OAI CI pipelines.

PR #10353, opened 2021-11-16
feat(orc8r): add control_proxy.yml endpoints

reginawang3495:
## Summary
Add 2 endpoints:
image

## Test Plan
- Added tests
- Ran orc8r tests
- Ran through curl requests manually in Swagger UI

## Additional Information


hcgatewood (commented):
Shape looks great 👍

orc8r/cloud/go/services/tenants/servicers/servicer.go
```diff
@@ -98,3 +98,38 @@ func (s *tenantsServicer) DeleteTenant(c context.Context, request *protos.GetTen
...
+}
+
+func (s *tenantsServicer) CreateOrUpdateControlProxy(c context.Context, request *protos.IDAndControlProxy) (*protos.Void, error) {
+	_, err := s.store.GetTenant(request.Id)
+	switch {
+	case err == errors.ErrNotFound:
+		return nil, status.Errorf(codes.NotFound, "Tenant %d not found", request.Id)
+	case err != nil:
+		return nil, status.Errorf(codes.Internal, "Error getting tenant %d: %v", request.Id, err)
+	}
```
reginawang3495: Duplicated a couple times -- can be extracted if desired. 

**Reasoning for not doing so:** I like as is because it's easy to read and abstracting too much makes the code hard to follow sometimes

orc8r/protos/tenants.proto
```diff
@@ -35,10 +35,17 @@ message IDAndTenant {
     Tenant tenant = 2;
```
reginawang3495, line 35:
```
message GetTenantRequest {
    int64 id = 1;
}
```
Could be renamed to something more `controlProxy` inclusive if desired

**Reasoning for not doing so:** I don't think this is necessary since we are living in the `tenants` service anyhow

reginawang3495 (commented):
2 things I don't think are necessary, but improvement ideas to point out!

orc8r/cloud/go/services/tenants/client_api.go
```diff
@@ -126,3 +126,39 @@ func DeleteTenant(ctx context.Context, tenantID int64) error {
...
+	}
+
+	controlProxy, err := oc.GetControlProxy(ctx, &protos.GetTenantRequest{Id: tenantID})
+	if err != nil {
+		switch {
+		case status.Convert(err).Code() == codes.NotFound:
+			return controlProxy, merrors.ErrNotFound
+		default:
+			return controlProxy, err
+		}
```
hcgatewood: Let's pull this switch into a separate helper fn

orc8r/cloud/go/services/tenants/obsidian/handlers/handlers.go
```diff
@@ -148,3 +159,55 @@ func DeleteTenantHandler(c echo.Context) error {
...
+		return terr
+	}
+
+	_, err := tenants.GetTenant(c.Request().Context(), tenantID)
+	switch {
+	case err == errors.ErrNotFound:
+		return echo.NewHTTPError(http.StatusNotFound, fmt.Errorf("Tenant %d does not exist", tenantID))
+	case err != nil:
+		return echo.NewHTTPError(http.StatusInternalServerError, err)
+	}
```
hcgatewood: Same here -- let's pull switch into a helper fn

orc8r/cloud/go/services/tenants/obsidian/models/swagger.v1.yml
```diff
@@ -94,6 +94,38 @@ paths:
         default:
           $ref: './orc8r-swagger-common.yml#/responses/UnexpectedError'
 
+  /tenants/{tenant_id}/control_proxy:
+    get:
+      summary: Retrieve ControlProxy Info By Tenant ID
```
hcgatewood: Here and below: sentence-case

orc8r/cloud/go/services/tenants/obsidian/models/swagger.v1.yml
```diff
@@ -94,6 +94,38 @@ paths:
...
+        default:
+          $ref: './orc8r-swagger-common.yml#/responses/UnexpectedError'
+    put:
+      summary: Create or Update ControlProxy
+      tags:
+        - Tenants
+      parameters:
+        - $ref: './orc8r-swagger-common.yml#/parameters/tenant_id'
+        - in: body
+          name: controlProxy
```
hcgatewood: Here and elsewhere: snake_case

orc8r/cloud/go/services/tenants/obsidian/models/swagger.v1.yml
```diff
@@ -114,3 +146,17 @@ definitions:
         type: array
         items:
           type: string
+  control_proxy:
+    type: object
+    required:
+      - id
+      - control_proxy
+    properties:
+      id:
```
hcgatewood: Unclear that we need ID in this object -- since we're only PUT-ing, we can grab the ID from the PUT endpoint

orc8r/protos/tenants.proto
```diff
@@ -35,10 +35,17 @@ message IDAndTenant {
...
+}
+
 service TenantsService {
     rpc GetAllTenants (Void) returns (TenantList) {}
     rpc GetTenant (GetTenantRequest) returns (Tenant) {}
     rpc CreateTenant (IDAndTenant) returns (Void) {}
     rpc SetTenant (IDAndTenant) returns (Void) {}
     rpc DeleteTenant (GetTenantRequest) returns (Void) {}
+    rpc GetControlProxy (GetTenantRequest) returns (IDAndControlProxy) {}
+    rpc CreateOrUpdateControlProxy (IDAndControlProxy) returns (Void) {}
```
hcgatewood: We always want the req and res to be `<rpc name>Request` and `<rpc name>Response` -- I know we don't follow that pattern correctly everywhere 🙂

hcgatewood (changes requested):
Couple small things, overall looks great though 👍

hcgatewood:
cc @sudhikan

orc8r/cloud/go/services/tenants/obsidian/models/swagger.v1.yml
reginawang3495: We could, but I wanted to be same as `tenant` (see L130-148) 

Thoughts?

hcgatewood: I generally prefer conciseness (e.g. don't repeat a value in both the key and value), but I also get that there's benefit to a value being able to self-identify. I'll let you make the decision then

reginawang3495: Maybe I can make an issue for removing `id` for both?

orc8r/protos/tenants.proto
reginawang3495: Will do, but also just wondering why? Since the message is the same thing duplicated

hcgatewood: For sure. So the reason is backward compatibility -- *currently* they are the same message (same fields), but that might change over time

Specifically, to preserve proto/grpc backward compatibility, there are a specific set of rules, but basically, if you only append new fields, you're in the clear

Potential issues of sharing, especially when a large number of rpcs share 1 underlying msg type

1. RPCs could diverge in purpose over time, which would cause the shared message type to effectively multiplex multiple logical message types, which causes confusion and larger surface for bugs
2. We could need to deprecate a field, but just for 1 of the services -- no clear solution in that case, so it just wouldn't get deprecated

More info: https://github.com/uber/prototool/blob/dev/style/README.md#servicesrpcs

hcgatewood: Obv this isn't really an issue for specifically our orc8r-internal grpc/proto defns, but it's good practice to keep this consistent across all gprc/proto defns

reginawang3495: Thanks for all the info!

orc8r/cloud/go/services/tenants/obsidian/handlers/handlers.go
reginawang3495: I did it, but I'm not sure it's an improvement since it's provides less clarity on exactly how the error handling is + not much less code

reginawang3495: I could also theoretically pass in a map of `StatusInternalServerError` to `error`, but I don't think that really decreases the code since that's the same thing as a `switch case`

orc8r/cloud/go/services/tenants/servicers/storage/storage.go
```diff
@@ -147,6 +150,47 @@ func (b *blobstoreStore) DeleteTenant(tenantID int64) error {
...
+		return err
+	}
+	defer store.Rollback()
+
+	controlProxyBlob := blobstore.Blob{
+		Type:  tenants.ControlProxyInfoType,
+		Key:   strconv.FormatInt(tenantID, 10),
+		Value: []byte(controlProxy),
+	}
+	glog.Infof("controlProxyBlob: %v type %v ket %v tid %v", controlProxyBlob, controlProxyBlob.Type, controlProxyBlob.Key, tenantID)
```
reginawang3495, line 184: here

orc8r/cloud/go/services/tenants/servicers/servicer.go
```diff
@@ -98,3 +95,43 @@ func (s *tenantsServicer) DeleteTenant(c context.Context, request *protos.GetTen
...
+	err = errorHandlingForGet(err, "controlProxy", request.Id)
+	if err != nil {
+		return nil, err
+	}
+	return &protos.GetControlProxyResponse{Id: request.Id, ControlProxy: controlProxy}, nil
+}
+
+func (s *tenantsServicer) CreateOrUpdateControlProxy(c context.Context, request *protos.CreateOrUpdateControlProxyRequest) (*protos.Void, error) {
+	_, err := s.store.GetTenant(request.Id)
+	glog.Infof("requestid %v", request.Id)
```
reginawang3495: here

orc8r/cloud/go/services/tenants/obsidian/handlers/handlers.go
hcgatewood: Mapping errors like this from a func is preferable imo. It 1. encapsulates the repeated logic in a single location, 2. yes it hides the impl, but you can jump to defn once and then you understand what it's doing, and 3. it strongly improves the skim-ability of the code

orc8r/cloud/go/services/tenants/client_api.go
```diff
@@ -72,17 +72,8 @@ func GetTenant(ctx context.Context, tenantID int64) (*protos.Tenant, error) {
...
-	if err != nil {
-		switch {
-		case status.Convert(err).Code() == codes.NotFound:
-			return tenant, merrors.ErrNotFound
-		default:
-			return tenant, err
-		}
-	}
-	return tenant, nil
+	return tenant, errorHandling(err)
```
hcgatewood: General: we still want to have the "if err != nil" block in this code. It's just the mapping between error types that we want to encapsulate

Also, when there's an err, we want to return zero-y values for all the non-err return vars. That's a good general default, because we don't want callers to become reliant on this implicit/unofficial part of the function contract

So it would be e.g.

```
if err != nil {
	return nil, mapErr(err)
}

File: orc8r/cloud/go/services/tenants/obsidian/handlers/handlers.go [hunk 12]
```diff
@@ -103,11 +114,9 @@ func GetTenantHandler(c echo.Context) error {
 		return terr
 	}
 	tenantInfo, err := tenants.GetTenant(c.Request().Context(), tenantID)
-	switch {
-	case err == errors.ErrNotFound:
-		return echo.NewHTTPError(http.StatusNotFound, fmt.Errorf("Tenant %d does not exist", tenantID))
-	case err != nil:
-		return echo.NewHTTPError(http.StatusInternalServerError, err)
+	err = errorHandlingForGet(err, fmt.Errorf("Tenant %d does not exist", tenantID), err)
```
hcgatewood: Same idea here -- call this `mapErr` (or `mapGetErr` if necessary), and only perform the wrap as part of the return, so e.g. `return mapErr(err)`

orc8r/cloud/go/services/tenants/obsidian/handlers/handlers.go
```diff
@@ -140,11 +147,65 @@ func DeleteTenantHandler(c echo.Context) error {
...
+	tenantID, terr := obsidian.GetTenantID(c)
+	if terr != nil {
+		return terr
+	}
+
+	_, err := tenants.GetTenant(c.Request().Context(), tenantID)
+	err = errorHandlingForGet(err, fmt.Errorf("Tenant %d does not exist", tenantID), err)
+	if err != nil {
+		return err
+	}
```
hcgatewood: Nit: here and below, we could refactor this shared functionality out to `validateTenantExists` function

orc8r/cloud/go/services/tenants/obsidian/models/swagger.v1.yml
```diff
@@ -114,3 +146,11 @@ definitions:
         type: array
         items:
           type: string
+  control_proxy:
+    type: object
+    required:
+      - control_proxy
+    properties:
+      control_proxy:
+        description: Control proxy of the tenant
```
hcgatewood:
```suggestion
        description: Contents of a gateway control_proxy.yml
```

orc8r/cloud/go/services/tenants/servicers/servicer.go
```diff
@@ -63,23 +63,19 @@ func (s *tenantsServicer) CreateTenant(c context.Context, request *protos.IDAndT
...
 func (s *tenantsServicer) GetTenant(c context.Context, request *protos.GetTenantRequest) (*protos.Tenant, error) {
 	tenant, err := s.store.GetTenant(request.Id)
-	switch {
-	case err == errors.ErrNotFound:
-		return nil, status.Errorf(codes.NotFound, "Tenant %d not found", request.Id)
-	case err != nil:
-		return nil, status.Errorf(codes.Internal, "Error getting tenant %d: %v", request.Id, err)
+	err = errorHandlingForGet(err, "tenant", request.Id)
+	if err != nil {
+		return nil, err
```
hcgatewood: Same idea here and below

orc8r/protos/tenants.proto
```diff
@@ -35,10 +35,22 @@ message IDAndTenant {
...
+}
+
 service TenantsService {
     rpc GetAllTenants (Void) returns (TenantList) {}
     rpc GetTenant (GetTenantRequest) returns (Tenant) {}
     rpc CreateTenant (IDAndTenant) returns (Void) {}
     rpc SetTenant (IDAndTenant) returns (Void) {}
     rpc DeleteTenant (GetTenantRequest) returns (Void) {}
+    rpc GetControlProxy (GetTenantRequest) returns (GetControlProxyResponse) {}
+    rpc CreateOrUpdateControlProxy (CreateOrUpdateControlProxyRequest) returns (Void) {}
```
hcgatewood: These should both be `rpc <name> (<name>Request) returns (<name>Response) {}`

hcgatewood (approved):
Looks great! Couple small requests on the error handling then msg me to merge 👍

orc8r/cloud/go/services/tenants/obsidian/handlers/handlers.go
reginawang3495: this is only for get though, (otherwise would need more paramters), can I call it `mapErrForGet`?

orc8r/protos/tenants.proto
reginawang3495: Even the void?

orc8r/cloud/go/services/tenants/servicers/storage/storage.go
```diff
@@ -20,6 +20,8 @@ type Store interface {
 	GetAllTenants() (*protos.TenantList, error)
 	SetTenant(tenantID int64, tenant protos.Tenant) error
```
reginawang3495, line 21: implementation of `SetTenant` and `CreateTenant` are exactly the same. Prefer code reuse

orc8r/protos/tenants.proto
hcgatewood: Correct -- cause e.g. if we suddenly needed to return something in that return proto, we'd have to add the field to `Void` to remain backwards-compatible

Again, not a big issue for our purposes, but is a good general pattern to follow

orc8r/cloud/go/services/tenants/obsidian/handlers/handlers.go
hcgatewood: I like `mapGetErr` -- conciseness/skim-ability is a huge pro for Go code. But `mapErrForGet` is also fine, up to you 👍

reginawang3495: But the method functionality could be confused for that "the method gets err"?

hcgatewood: Up to you 🙂

orc8r/cloud/go/services/tenants/servicers/storage/storage_test.go
```diff
@@ -161,3 +165,35 @@ func TestBlobstoreStore_DeleteTenant(t *testing.T) {
 	err = s.DeleteTenant(0)
 	assert.EqualError(t, err, "error")
 }
+
+// TODO(reginawang3495): Do or make ticket for making other tests in this file use less mocking
```
reginawang3495: xx

hcgatewood: This amount of testing seems sufficient 👍

orc8r/cloud/go/services/tenants/obsidian/handlers/handlers.go
```diff
@@ -139,12 +147,75 @@ func DeleteTenantHandler(c echo.Context) error {
...
+
+	err := validateTenantExists(c, tenantID)
+	if err != nil {
+		return err
+	}
+
+	controlProxy, err := tenants.GetControlProxy(c.Request().Context(), tenantID)
+	if err != nil {
+		return mapErr(err, fmt.Errorf("control proxy for tenantID %d does not exist", tenantID), err)
+		return mapErr(err, fmt.Errorf("control proxy for tenantID %d does not exist", tenantID), err)
```
hcgatewood: Duplicate

hcgatewood (approved):
Lgtm! One small fix then msg to merge

PR #10223, opened 2021-11-10
feat(feg): N7 Proxy service

GANESH-WAVELABS:
Signed-off-by: GANESH IRRINKI <ganesh.irrinki@wavelabs.ai>

## Summary
- Added new N7 proxy service. Docker compose changes and new service code addition
- Added MConfig fields required for N7 Proxy service
- CentralSessionController RPC skeleton implentation
- Metrics and health check

The file feg/cloud/go/protos/mconfig/mconfigs.pb.go is a generated one from feg/protos/mconfig/mconfigs.proto.

This PR is now split into two - #10360 and #10362

## Additional Information

- This PR is dependent on #9524. To be merged after 9524 is merged. The feg build will fail as the files changed as part of 9524 is not present in this PR.

PR #10360, opened 2021-11-16
feat(feg): N7 Proxy Service

GANESH-WAVELABS:

## Summary
- Added new N7 proxy service. Docker compose changes and new service code addition
- Added MConfig fields required for N7 Proxy service
- CentralSessionController RPC skeleton implentation

## Additional Information
This is dependant on #9524. Shall be merged only after 9524 PR is merged.

This is a split from #10223. 10223 will be closed once all split PRs are raised.


uri200:
This PR comes from https://github.com/magma/magma/pull/10223

uri200:
Awesome! size of this PR is now manageable. Thanks @GANESH-WAVELABS

uri200:
Now that I understand the PR better I have some concerns (maybe they are already been taken into consideration):

4g centralSessionController manages both Gx and Gy. So a message from sessiond can trigger either a message to PCRF and/or a message. Does n7 interfaces handles Policies and Charging? Or do we need another client for OCS?

GANESH-WAVELABS:
@uri200 
> 4g centralSessionController manages both Gx and Gy. So a message from sessiond can trigger either a message to PCRF and/or a message. Does n7 interfaces handles Policies and Charging? Or do we need another client for OCS?

N7 is an interface to connect SMF with Policy Control Function(PCF). For charging N40 interface is used between SMF and CHF (Charging Function). It will require another client (HTTP) to connect with CHF. The scope of the work was only for Policy.

uri200:
> @uri200 
> > 4g centralSessionController manages both Gx and Gy. So a message from sessiond can trigger either a message to PCRF and/or a message. Does n7 interfaces handles Policies and Charging? Or do we need another client for OCS?
> 
> N7 is an interface to connect SMF with Policy Control Function(PCF). For charging N40 interface is used between SMF and CHF (Charging Function). It will require another client (HTTP) to connect with CHF. The scope of the work was only for Policy.
> 

Then I think we need to consider Thibkibg on a similar client that session proxy that will send messages to either PCF or CHF. The reason is what I mentioned: the day we need to include chf support we will need too do it through this service.

uri200:
I think the http client can be the same for all the interfaces. So we can have one service talking to both PCF and CHF. 

In session proxy we have different diamenter clients because they have to maintain connection. But since 5g is http I understand we are just sending tcp messages, so this is just a matter of sending the proper message to the right interface. 

I think we may need to change the name of this service to something like `session_proxy_5g` or `N7_N40` (check if the name is right). The reason is session proxy will send PCF and CHF messages on the same message. So we need one single service to receive it

We can have a folder strucutre like 
```
\n7_n40_proxy
	\n7
	\n40
```
We don't need to implement now n40 of course

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -0,0 +1,49 @@
...
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package n7
```
uri200, line 14: https://github.com/magma/magma/pull/10223#discussion_r748646612

feg/gateway/services/n7_n40_proxy/n7/config.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+type N7ConsumerConfig struct {
+	LocalAddr     string // ip:port or :port
+	NotifyApiRoot string
+}
+
+type N7Config struct {
+	DisableN7 bool
+	Server    *PCFConfig
```
uri200: PCFConfig doesnt need to be a pointer. It is just a struct of strings

feg/gateway/services/n7_n40_proxy/n7/config.go
```diff
@@ -0,0 +1,136 @@
...
+
+	DefaultPcfApiRoot        = "https://localhost"
+	DefaultPcfTokenUrl       = "https://localhost/token"
+	DefaultClientId          = "magma_client_id"
+	DefaultClientSecret      = "magma_client_secret"
+	DefaultN7ConsumerAddr    = "localhost:0"
+	DefaultN7ConsumerApiRoot = "https://localhost/npcf-smpolicycontrol/v1"
+)
+
+type PCFConfig struct {
```
uri200, line 46: why don't you use simplified names
```
N7CLientConfig
N7ServerConfig
```

or 

```
N7ProducerConfig
N7ConsumerConfig
```
but this one can be difficult to understand

feg/gateway/services/n7_n40_proxy/n7/config.go
```diff
@@ -0,0 +1,136 @@
...
+	NotifyApiRoot string
+}
+
+type N7Config struct {
+	DisableN7 bool
+	Server    *PCFConfig
+	Consumer  N7ConsumerConfig
+}
+
+func (cfg *PCFConfig) Validate() error {
```
uri200: To reduce the clutter, and since those structs are simple, can you do a single function that validates the whole config. You can just do a simple function (no need of a pointer receiver)
```
func Validate(config PCFConfig)
```

You can check it at the end of creation of the object.  This will make the file clearer and if we ever add more check we will only have to check in one single place

feg/gateway/services/n7_n40_proxy/n7/config.go
```diff
@@ -0,0 +1,136 @@
...
+	DefaultPcfTokenUrl       = "https://localhost/token"
+	DefaultClientId          = "magma_client_id"
+	DefaultClientSecret      = "magma_client_secret"
+	DefaultN7ConsumerAddr    = "localhost:0"
+	DefaultN7ConsumerApiRoot = "https://localhost/npcf-smpolicycontrol/v1"
+)
+
+type PCFConfig struct {
+	ApiRoot      string
+	TokenUrl     string
```
uri200, line 48: I am not sure how you are using those URL, but would it make sense to use an URL object here. You are parsing the URL on the config. So if you need to parse it later, maybe it would be good to just store it as URL type

If you are consuming it as string, then it is fine

feg/gateway/services/n7_n40_proxy/n7/config.go
uri200: I see the mconfig  proto is also using a weird mixture of words. Can you decide either or . I think mixing both makes more difficult to understand

feg/gateway/services/n7_n40_proxy/n7/config.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+type N7Config struct {
+	DisableN7 bool
+	Server    *PCFConfig
+	Consumer  N7ConsumerConfig
+}
+
+func (cfg *PCFConfig) Validate() error {
+	_, err := url.ParseRequestURI(cfg.ApiRoot)
```
uri200: why in this case you are not checking the size of ApIRoot

feg/gateway/services/n7_n40_proxy/n7/config.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return fmt.Errorf("invalid ApiRoot")
+	}
+	if len(cfg.TokenUrl) == 0 {
```
uri200: is this correct?  why are you parsing a token with len 0

feg/gateway/services/n7_n40_proxy/n7/config.go
```diff
@@ -0,0 +1,136 @@
...
+			LocalAddr:     utils.GetValueOrEnv("", N7ConsumerLocalAddr, n7configPtr.Consumer.LocalAddr),
+			NotifyApiRoot: utils.GetValueOrEnv("", N7ConsumerNotifyApiRoot, n7configPtr.Consumer.NotifyApiRoot),
+		}
+		err = conf.Consumer.Validate()
+		if err != nil {
+			glog.Errorf("Managed Config Load Error, N7 Consumer config failed: %v", err)
+			return nil
+		}
+	}
+	return conf
```
uri200: use the new validate function before `return conf`. Remove your other validate functions

Just make sure your Validate function returns nice errors informing the wrong parameter

feg/gateway/services/n7_n40_proxy/n7/config.go
```diff
@@ -0,0 +1,136 @@
...
+	PcfClientId             = "PCF_CLIENT_ID"
+	PcfClientSecret         = "PCF_CLIENT_SECRET"
+	N7ConsumerLocalAddr     = "N7_CONSUMER_LOCAL_ADDR"
+	N7ConsumerNotifyApiRoot = "N&_CONSUMER_NOTIFY_API_ROOT"
+
+	DefaultPcfApiRoot        = "https://localhost"
+	DefaultPcfTokenUrl       = "https://localhost/token"
+	DefaultClientId          = "magma_client_id"
+	DefaultClientSecret      = "magma_client_secret"
+	DefaultN7ConsumerAddr    = "localhost:0"
```
uri200: This line is fine, but keep in mind that your client configuration will require a specific port. So you will always have to define a port. The reason is, the server PCF needs to be configured with an address and port. If you use 0, your port will be random and will change in every restart. So your PCF will never have the right port configured in their end


For testing we use 0 because we don;'t want to collide with any port used by the machine you run the test.

feg/gateway/services/n7_n40_proxy/n7/config.go
```diff
@@ -0,0 +1,136 @@
...
+			ApiRoot:      utils.GetValueOrEnv("", PcfApiRoot, DefaultPcfApiRoot),
+			TokenUrl:     utils.GetValueOrEnv("", PcfTokenUrl, DefaultPcfTokenUrl),
+			ClientId:     utils.GetValueOrEnv("", PcfClientId, DefaultClientId),
+			ClientSecret: utils.GetValueOrEnv("", PcfClientSecret, DefaultClientSecret),
+		}
+		conf.DisableN7 = false
+		conf.Consumer = N7ConsumerConfig{
+			LocalAddr:     utils.GetValueOrEnv("", N7ConsumerLocalAddr, DefaultN7ConsumerAddr),
+			NotifyApiRoot: utils.GetValueOrEnv("", N7ConsumerNotifyApiRoot, DefaultN7ConsumerApiRoot),
+		}
```
uri200, line 85: this configuration is not validated. Lets do it at the end

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -0,0 +1,49 @@
...
+	"net/http"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+
+	n7_client "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
+)
+
+// createN7Client creates a N7 oapi client and sets the OAuth2 client credentiatls for authorizing requests
+func CreateN7Client(cfg *PCFConfig) (*n7_client.ClientWithResponses, error) {
```
uri200: always use New prefix when creating a new struct
```
NewN7Client
```

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -0,0 +1,49 @@
...
+package n7
+
+import (
+	"context"
+	"net/http"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+
+	n7_client "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
```
uri200: since we will have lots of clients and servers, what about `n7_sbi_client`

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// createN7Client creates a N7 oapi client and sets the OAuth2 client credentiatls for authorizing requests
```
uri200: You are missing `C`

Note the IDE should highlight the word differently if you did it right

image

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -0,0 +1,49 @@
...
+	serverStr, err := GetServerStringFromUrl(cfg.ApiRoot)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create new N7 client object and assosiate it with oAuth2 HTTP client
+	client, err := n7_client.NewClientWithResponses(
+		serverStr,
+		n7_client.WithHTTPClient(doer),
+	)
```
uri200, line 39: Create the variables closer where you are going to use them

Also try to reduce the usage of variables that are only used once. Unles it makes the code way more understandeabe

```
	tokenCtxt := context.WithValue(context.Background(), oauth2.HTTPClient, httpClient)
	client, err := n7_client.NewClientWithResponses(
		serverStr,
		n7_client.WithHTTPClient(tokenConfig.Client(tokenCtxt)),
	)

```

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -0,0 +1,49 @@
...
+
+// createN7Client creates a N7 oapi client and sets the OAuth2 client credentiatls for authorizing requests
+func CreateN7Client(cfg *PCFConfig) (*n7_client.ClientWithResponses, error) {
+	httpClient := &http.Client{}
+	tokenConfig := clientcredentials.Config{
+		ClientID:     cfg.ClientId,
+		ClientSecret: cfg.ClientSecret,
+		TokenURL:     cfg.TokenUrl,
+	}
+	tokenCtxt := context.WithValue(context.Background(), oauth2.HTTPClient, httpClient)
```
uri200: maybe you can do that 

```
tokenCtxt := context.WithValue(context.Background(), oauth2.HTTPClient, &http.Client{})
```

uri200 (changes requested):

feg/gateway/services/n7_n40_proxy/n7/util.go
```diff
@@ -0,0 +1,28 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package n7
+
+import (
```
uri200: https://github.com/magma/magma/pull/10223#discussion_r748646612

feg/gateway/services/n7_n40_proxy/n7/n7.go
```diff
@@ -0,0 +1,49 @@
...
+	doer := tokenConfig.Client(tokenCtxt)
+
+	serverStr, err := GetServerStringFromUrl(cfg.ApiRoot)
```
uri200: I think if you parse the URL string  into URL type on config.go you can safely extract the server addres easily

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		glog.Errorf("Creating N7 Client failed: %s", err)
+		return nil, err
+	}
+	return &CentralSessionController{
+		policyClient: policyClient,
+		dbClient:     dbClient,
+		cfg: &SessionControllerConfig{
+			N7Config:       n7config,
+			RequestTimeout: 3 * time.Second,
```
uri200: Never ever hardcode a value in middle of the code. It gets lost and it is impossible to find

Create a const like here 

https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fgtp%2Fgtp_client.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=31&start_col=12&end_row=31&end_col=12

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
```diff
@@ -0,0 +1,88 @@
...
+		},
+	}, nil
+}
+
+// CreateSession begins a UE session by requesting rules from PCF and returning them.
+func (srv *CentralSessionController) CreateSession(
+	ctx context.Context,
+	request *protos.CreateSessionRequest,
+) (*protos.CreateSessionResponse, error) {
+
```
uri200, line 70: What you wrote is nice, but note you can also do this. 

```
	return protos.UnimplementedCentralSessionControllerServer.CreateSession(ctx, request)
```

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
```diff
@@ -0,0 +1,88 @@
...
+type SessionControllerConfig struct {
+	N7Config       *n7.N7Config
+	RequestTimeout time.Duration
+}
+
+func NewCentralSessionController(
+	dbClient policydb.PolicyDBClient,
+) (*CentralSessionController, error) {
+	n7config := n7.GetN7Config()
+	if n7config == nil {
```
uri200: no needed. your `GetN7Config` will always return an object

feg/protos/mconfig/mconfigs.proto
```diff
@@ -244,3 +244,41 @@ message S8Config {
     string pgw_address = 3;
     string apn_operator_suffix = 4;
 }
+
+message HttpServerConfig {
```
uri200: shouldn' this be `SbiServerConfig`?

Or does this object really reprepresents a generic HTTP server?

feg/protos/mconfig/mconfigs.proto
```diff
@@ -244,3 +244,41 @@ message S8Config {
...
+    string notify_api_root = 2;
+}
+
+message N7Config {
+    // Disables N7 interface
+    bool disable_n7 = 2;
+    // PCF configuration
+    HttpServerConfig server = 3;
+    // N7 consumer config for handling notifications
+    N7ConsumerConfig consumer = 6;
```
uri200: ohh it is here!  

What about changing the name to `client`?

feg/gateway/services/n7_n40_proxy/n7/config.go
GANESH-WAVELABS: TokenUrl is used as is but ApiRoot is parsed to extract host-port with http scheme.

feg/gateway/services/n7_n40_proxy/n7/config.go
GANESH-WAVELABS: Done. Using the terms Client and Server

feg/gateway/services/n7_n40_proxy/n7/config.go
GANESH-WAVELABS: Done.

feg/gateway/services/n7_n40_proxy/n7/config.go
```diff
@@ -0,0 +1,132 @@
...
+		conf.Client = N7ClientConfig{
+			LocalAddr:     utils.GetValueOrEnv("", N7ClientLocalAddr, n7configPtr.Client.LocalAddr),
+			NotifyApiRoot: utils.GetValueOrEnv("", N7ClientNotifyApiRoot, n7configPtr.Client.NotifyApiRoot),
+		}
+	}
+	err = validateN7Config(conf)
+	if err != nil {
+		return nil, err
+	}
+	return conf, nil
```
uri200, line 108: you don't need to change it but in those cases sometimes we do

```
return validateN7Config(conf)
```

where validateN7Config is
```
func validateN7Config(config *N7Config) (*N7Config, error) {
```

feg/gateway/services/n7_n40_proxy/n7/config.go
```diff
@@ -0,0 +1,132 @@
...
+	_, err = url.ParseRequestURI(config.Client.NotifyApiRoot)
+	if err != nil {
+		return fmt.Errorf("invalid Client NotifyApiRoot - %s", err)
+	}
+	_, err = net.ResolveTCPAddr("tcp", config.Client.LocalAddr)
+	if err != nil {
+		return fmt.Errorf("invalid Client LocalAddr - %s", err)
+	}
+	return nil
+}
```
uri200, line 132: note how clean this file looks like now.
That is one of the beauties of go

uri200 (approved):

PR #10364, opened 2021-11-16
chore(agw): clean up grpc vs net.listener network scheme handling

zbrock:
* clean up grpc vs net.listener network scheme handling
* clean up repeated grpc server setup code
* Fix missing gofmt run on script file

## Summary

Per PR comments on 10158

## Test Plan
Ensured that gRPC service still comes up on local box.
Unit tests should be sufficient.

## Additional Information

N/A

### DIFF
--- a/src/go/agwd/config/config.go
+++ b/src/go/agwd/config/config.go
@@ -53,22 +53,33 @@ const (
 
 // ParseTarget takes a target in string form and returns a resolved Target.
 // Extends functionality in grpc/internal/grpcutil.ParseTarget to support ipv4
-// and ipv6 schemes. Return tcp4Scheme or tcp6Scheme so the Scheme can be passed
-// directly to net.Listen.
+// and ipv6 schemes.
 func ParseTarget(target string) resolver.Target {
 	if strings.HasPrefix(target, ipv4Scheme+":") {
 		return resolver.Target{
-			Scheme:   tcp4Scheme,
+			Scheme:   ipv4Scheme,
 			Endpoint: target[len(ipv4Scheme)+1:],
 		}
 	}
 	if strings.HasPrefix(target, ipv6Scheme+":") {
 		return resolver.Target{
-			Scheme:   tcp6Scheme,
+			Scheme:   ipv6Scheme,
 			Endpoint: target[len(ipv6Scheme)+1:],
 		}
 	}
 	return grpcutil.ParseTarget(target, false)
+}
+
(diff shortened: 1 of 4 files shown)


src/go/agwd/config/config_test.go
```diff
@@ -132,6 +132,53 @@ func TestParseTarget(t *testing.T) {
...
+			},
+		},
+		{
+			input: resolver.Target{
+				Scheme: "ipv6",
+			},
+			want: resolver.Target{
+				Scheme: "tcp6",
+			},
+		},
```
mstre123, line 166: probably good to add one more test that scheme with tcp4 outputs scheme with tcp4

src/go/agwd/config/config.go
```diff
@@ -53,24 +53,35 @@ const (
...
 		}
 	}
 	return grpcutil.ParseTarget(target, false)
 }
 
+// TargetSchemeAdapter takes a gRPC formatted Target such as {ipv4, ipv6}
+// and translates to net.Listen's required network type such as {tcp4, tcp6}.
+func TargetSchemeAdapter(target resolver.Target) resolver.Target {
+	if target.Scheme == ipv4Scheme {
+		target.Scheme = tcp4Scheme
```
mstre123, line 77: should we do immediate return in the conditional statements like we do in ParseTarget - makes it clearer that the target.Scheme isn't being altered by anything further down in the function.

mstre123 (approved):
lgtm

src/go/agwd/config/config.go
```diff
@@ -53,24 +53,35 @@ const (
...
+			Scheme:   ipv6Scheme,
 			Endpoint: target[len(ipv6Scheme)+1:],
 		}
 	}
 	return grpcutil.ParseTarget(target, false)
 }
 
+// TargetSchemeAdapter takes a gRPC formatted Target such as {ipv4, ipv6}
+// and translates to net.Listen's required network type such as {tcp4, tcp6}.
+func TargetSchemeAdapter(target resolver.Target) resolver.Target {
```
markjen, line 75: interesting. why not `func NewTargetListner(target resolver.Target) (net.Listner, error)`

zbrock: I started adding an injectable Factory interface for net.Listeners so I could write tests for it, but the code ended up being mostly abstractions to make it testable with almost no actual business logic. This was simpler to get under test coverage.

PR #10372, opened 2021-11-17

spikey979:
fix(feg): fixed unnecessary check if the number is lower than 0

## Summary

- feg/gateway/tools/swx_cli/main.go:238 -> sendMar()
   There was unnecessary checking if the number is lower than 0, because unit64 type number can't be < 0

Screenshot 2021-11-17 at 09 51 23

## Test Plan
I have run unit tests /magma/feg/gateway/docker ./build.py -c

Screenshot 2021-11-17 at 10 08 36

### DIFF
--- a/feg/gateway/tools/swx_cli/main.go
+++ b/feg/gateway/tools/swx_cli/main.go
@@ -235,7 +235,7 @@ func sendSar(addr string, client swxClient) int {
 }
 
 func sendMar(addr string, client swxClient) int {
-	if numVectors < 0 || numVectors > 4294967295 {
+	if numVectors > 4294967295 {
 		fmt.Printf("numVectors %d is outside the bounds of unit32 type", numVectors)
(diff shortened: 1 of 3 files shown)


feg/gateway/tools/swx_cli/main.go
```diff
@@ -235,7 +235,7 @@ func sendSar(addr string, client swxClient) int {
 }
 
 func sendMar(addr string, client swxClient) int {
-	if numVectors < 0 || numVectors > 4294967295 {
+	if numVectors > 4294967295 {
```
themarwhal, line 238: can we use a constant here?

uri200:
fixed here http://github.com/magma/magma/blob/929f524eb2c2aed3ed544091fde224d7f727f215/feg/gateway/tools/swx_cli/main.go#L239-L239

PR #10378, opened 2021-11-17
chore(ci): Github composite action docker image builds

Neudrino:
## Summary

* POC for Github composite actions.
* Refactoring of docker image builds to make use of Github composite actions.
  * ⚠️ As this only a refactoring, a change to the execution logic was not made, e.g. what is executed when. I would see this outside the scope of this PR.
* Also include the docker health check for builds on PR, as differences are minimal.
* I think it is cleaner and more flexible to have one docker build per file, so I refactored it that way.
  * Docker image builds will only get triggered for images actually changed. I.e. this safes build resources, by factor of configured overall docker builds, e.g. currently 4. In my experience, Docker builds are often quite resource hungry.
  * Changes and possible differences in the CI for the different docker files are easier to implement. ~~E.g. deprecation and deletion of `Dockerfile.ubuntu20.04` would be simple.~~ E.g. it could be desireable to have different execution scheduler frequencies or times. Different execution times would act somewhat as a manual load balancing.
  * With respect to https://github.com/magma/magma/pull/10326, `Dockerfile.ubuntu20.04` was not migrated / removed.
* With this refactoring all docker builds are consistently defined, and all will have a health check docker build as well.

## Test Plan

* CI is running and building images as for the non composite Github actions
  *  See [this run](https://github.com/magma/magma/runs/4238714752?check_suite_focus=true) for a slightly adoped pipeline, which omits the GHCR login and also triggers on PR. Testing changes were https://github.com/magma/magma/pull/10378/commits/c452cf05a986e11b9f44c4e4139cef0804e19f01 and https://github.com/magma/magma/pull/10378/commits/44b6ae00c3df5578e4283da34a4fc4772c3db226.
* ⚠️ Docker push action could not be tested due to limited access to credentials from PRs.

## Additional Information

* This PR is a followup on https://github.com/magma/magma/pull/10312.
* Personally, I would tend to split this PR in two, to have an incremental addition of the features here. First one to refactor the `docker-push` action with the composite action, and the second to include the `docker-health` into it as well. Somehow this feels safer to me. However, this is matter of taste, I guess. Nevertheless I prepared one big PR, to give the full overview of where I think this could go.


electronjoe (approved):
At the top level - I appreciate your PR's re-use improvements, and see value in your rationale for splitting container builds into separate workflows that re-use the common build / push workflow.

.github/workflows/docker-builder-python-precommit.yml
```diff
@@ -0,0 +1,40 @@
+---
+name: "Build Docker image for Python Precommit"
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - master
+    paths:
+      - lte/gateway/docker/python-precommit/Dockerfile
+      - .github/workflows/docker-bazel-base-builder.yml
```
themarwhal: bad copy paste I'm wondering?

.github/workflows/composite/docker-builder/action.yml
```diff
@@ -0,0 +1,51 @@
...
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: docker/setup-buildx-action@v1
+    - name: Login to GHCR
+      uses: docker/login-action@v1
+      with:
+        registry: ${{ inputs.REGISTRY }}
+        username: ${{ github.repository_owner }}
+        password: ${{ inputs.GITHUB_TOKEN }}
+      if: ${{ github.event_name == 'push' && github.ref_name == 'master' }}
```
maxhbr: Maybe something like:
```suggestion
      if: ${{ github.event_name == 'push' && github.ref_name == 'master' && inputs.GITHUB_TOKEN != null }}
```

themarwhal (approved):
This looks good to me!

quentinDERORY (approved):
LGTM

.github/workflows/composite/docker-builder/action.yml
Neudrino: ✔️ Added.

.github/workflows/docker-builder-python-precommit.yml
Neudrino: Fixed and force pushed. ✔️

themarwhal:
@Neudrino please rebase on top of master, then it should be ready to go!

Issue #9828, opened 2021-10-21
Move all doc.go code to a corresponding const.go


hcgatewood:
doc.go files are meant to be exclusively for documentation. However, we use them in a number of cases to store constants. Instead, for each occasion, we should move those consts from the doc.go file to a corresponding const.go file

Example: https://github.com/magma/magma/blob/509e5b4bcdb68b0021044fb1dc8cf47554446b8f/orc8r/cloud/go/services/streamer/doc.go#L31-L33

> 🙌 Bulk-uploaded by https://github.com/hcgatewood/ghissue

PR #10377, opened 2021-11-17
fix(orc8r): lowercase orc8r service names

spikey979:
fix(orc8r): lowercase orc8r service names

## Summary
In this files:
- fbinternal/cloud/go/services/download/download.go
- fbinternal/cloud/go/services/fbinternal/doc.go
- fbinternal/cloud/go/services/testcontroller/defs.go
- fbinternal/cloud/go/services/vpnservice/client_api.go
- feg/cloud/go/services/basic_acct/definitions.go
- feg/cloud/go/services/feg_relay/feg_relay.go
- feg/gateway/services/aaa/base_acct/client_api.go
- feg/gateway/services/radiusd/doc.go
- lte/cloud/go/services/lte/doc.go
- lte/cloud/go/services/policydb/client_api.go
- lte/cloud/go/services/smsd/const.go
- orc8r/cloud/go/services/accessd/client_api.go
- orc8r/cloud/go/services/analytics/doc.go
- orc8r/cloud/go/services/bootstrapper/const.go
- orc8r/cloud/go/services/certifier/client_api.go
- orc8r/cloud/go/services/configurator/doc.go
- orc8r/cloud/go/services/ctraced/doc.go
- orc8r/cloud/go/services/device/doc.go
- orc8r/cloud/go/services/directoryd/client_api.go
- orc8r/cloud/go/services/dispatcher/client_api.go
- orc8r/cloud/go/services/eventd/doc.go
- orc8r/cloud/go/services/metricsd/doc.go
- orc8r/cloud/go/services/state/doc.go
- orc8r/cloud/go/services/streamer/doc.go
- orc8r/gateway/go/directoryd/client_api.go

service names in Go code are standardized to all-lowercase.

This is for issue:
https://github.com/magma/magma/issues/9829

## Test Plan
I ran unit tests in: /magma/orc8r/cloud/docker/  ./build.py -c

Issue #9829, opened 2021-10-21
Lowercase all Orc8r service names


hcgatewood:
For legacy reasons, our cloud service names are a mix of all-lowercase and all-uppercase.

Our current service registry code always coerces service names to all-lowercase, so we should standardize all service names in our code to all-lowercase.

To find a superset of all the service names: search for `servicename =` in all our Go code. Example: https://github.com/magma/magma/blob/b52a366b6ff1493bc04b912b06b7ac5e5c406cfc/lte/cloud/go/services/smsd/const.go#L16

> 🙌 Bulk-uploaded by https://github.com/hcgatewood/ghissue

PR #10380, opened 2021-11-17
fix(orc8r): move constants from doc.go files

spikey979:
fix(orc8r): move constants from doc.go files

## Summary

Moved constants from doc.go files to a corresponding const.go files

For issue:
https://github.com/magma/magma/issues/9828


themarwhal (approved):

orc8r/cloud/go/services/streamer/const.go
```diff
@@ -16,5 +16,6 @@ package streamer
 import "syscall"
 
 const (
-	EAGAIN = syscall.EAGAIN
+	EAGAIN      = syscall.EAGAIN
+	ServiceName = "streamer"
```
themarwhal, line 20: Why is the casing different? it looks like it used to be STREAMER?

spikey979: I solve this issue as well: https://github.com/magma/magma/issues/9829

spikey979: Here is my PR for lowercasing: https://github.com/magma/magma/pull/10377

themarwhal: ohhhh I see! Looks good then :)

hcgatewood (approved):
Thanks for the fix! I'd say wait till tomorrow, rebase onto latest master, then re-push to see if we can get the flaky checks to pass

hcgatewood (approved):

lte/gateway/c/core/oai/test/s6a_task/s6a_test.cpp
```diff
@@ -21,4 +21,4 @@ int main(int argc, char** argv) {
   ::testing::InitGoogleTest(&argc, argv);
   OAILOG_INIT("S6A", OAILOG_LEVEL_DEBUG, MAX_LOG_PROTOS);
   return RUN_ALL_TESTS();
-}
```
themarwhal: we can remove this change :)

spikey979: I removed that change... sorry about that

PR #10382, opened 2021-11-17
chore(orc8r): Fix master not insync

hcgatewood:
## Additional Information


themarwhal (approved):
how did this happen? :o

hcgatewood:
> how did this happen? :o

My assumption is some formatter config or version got bumped, but because it's not in a covered path, that didn't trigger the check to run

cc @quentinDERORY for just making insync-checkin run on every PR, with no path filtering etc

Issue #9861, opened 2021-10-25
Migrate c/c++ includes to full rep-relative path


electronjoe:
Today Magma's C / C++ code base includes reflect a mentality of multi-repo behavior, within a single monorepo code base.

This leads to dangers (of file name collision and misbehaivor), lack of reviewer clarity (of where the mentioned file actually lives in the repo) - and prevents best practices by some modern build systems e.g. `Bazel` which we are interested in migrating to (see #8338).

electronjoe:
This was accomplished but new regressions exist, tracking them seperately.

PR #10383, opened 2021-11-17
fix(mme): correct CMakeLists.txt race condition

electronjoe:
## Problem discovery and root cause

This PR addresses a race condition observed in both `devcontainer` and `magma VM` enviornments by myself - and in `devcontainer` environment by @themarwhal.

Execution of `cd lte/gateway; make test_oai;` fails with error signature:

```
In file included from /home/vagrant/magma/lte/gateway/c/core/oai/test/mock_tasks/grpc_mock.cpp:18:
/home/vagrant/magma/lte/gateway/c/core/oai/tasks/grpc_service/S8ServiceImpl.h:19:10: fatal error: feg/protos/s8_proxy.grpc.pb.h: No such file or directory
   19 | #include "feg/protos/s8_proxy.grpc.pb.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Exploring the `build/c/...` generated directory and looking for protobuf / grpc dynamically generated outputs, I saw none yet present within `.../build/c/core/oai/lte/gateway/c/core/oai/tasks/grpc_service/`. This to me indicated a race condition between generation of grpc_service's protos / grpc outputs, and the build of `.../test/mock_tasks/grpc_mock.cpp`.

Looking at `test/mock_tasks/CmakeLists.txt` we actually se no explicit reference to the dependency of `TASK_GRPC_SERVICE`.  Includes are being pulled in by "include_directory" clauses - which is [not best practice](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#forget-the-commands-add_compile_options-include_directories-link_directories-link_libraries) and hides issues such as we are seeing here.

## Correction

The correction in this PR is to explicitly call out the target's dependency upon `TASK_GRPC_SERVICE` module - by adding it to the `target_link_libraries` clause.

- This notifies CMake that there is a sequencing dependency here - even if linking is not necessary
- Further it makes the exported library (including headers) available to the local build target

Applying this change highlighted a secondary fixup - `test/mock_tasks/grpc_mock.cpp` has an include clause of `-#include "tasks/grpc_service/S8ServiceImpl.h"` - which relies upon the now-removed `include_directories` specification.  Instead this interface should be exported by grpc_service's library definition in `CMakeLists.txt` - and should have no relative path.  So we make these two changes here.

### DIFF
--- a/lte/gateway/c/core/oai/test/mock_tasks/CMakeLists.txt
+++ b/lte/gateway/c/core/oai/test/mock_tasks/CMakeLists.txt
@@ -12,11 +12,9 @@ cmake_minimum_required(VERSION 3.7.2)
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
-
+ 
 include_directories("/usr/src/googletest/googlemock/include/")
 link_directories(/usr/src/googletest/googlemock/lib/)
-include_directories("${MAGMA_ROOT}/lte/gateway/c/core/oai/")
-include_directories("${PROJECT_BINARY_DIR}/tasks/grpc_service/")
 
 add_library(MOCK_TASKS
     ha_mock.cpp
@@ -35,4 +33,5 @@ add_library(MOCK_TASKS
 
 target_link_libraries(MOCK_TASKS
     LIB_ITTI gmock_main gtest gtest_main gmock
+    TASK_GRPC_SERVICE
 )
--- a/lte/gateway/c/core/oai/test/mock_tasks/grpc_mock.cpp
+++ b/lte/gateway/c/core/oai/test/mock_tasks/grpc_mock.cpp
@@ -15,7 +15,7 @@
 #include <grpcpp/security/server_credentials.h>
 #include "mock_tasks.h"
 #include "grpc_service.h"
-#include "tasks/grpc_service/S8ServiceImpl.h"
+#include "lte/gateway/c/core/oai/tasks/grpc_service/S8ServiceImpl.h"
 
 task_zmq_ctx_t task_zmq_ctx_grpc;
 grpc_service_data_t grpc_service_config = {0};


lte/gateway/c/core/oai/tasks/grpc_service/CMakeLists.txt
```diff
@@ -110,6 +110,7 @@ add_library(TASK_GRPC_SERVICE
     grpc_service.cpp
     grpc_service_task.c
     S8ServiceImpl.cpp
+    S8ServiceImpl.h
```
electronjoe: Explanation: This header is used by consumers of grpc_service but was missing from the add_library clause.

lte/gateway/c/core/oai/test/mock_tasks/grpc_mock.cpp
```diff
@@ -15,7 +15,7 @@
 #include <grpcpp/security/server_credentials.h>
 #include "mock_tasks.h"
 #include "grpc_service.h"
-#include "tasks/grpc_service/S8ServiceImpl.h"
+#include "S8ServiceImpl.h"
```
electronjoe: Explanation: As grpc_service library now exports `S8ServiceImpl.h` (as it should) - it is now offered without any path prefix to this build target.

lte/gateway/c/core/oai/test/mock_tasks/CMakeLists.txt
```diff
@@ -35,4 +33,5 @@ add_library(MOCK_TASKS
 
 target_link_libraries(MOCK_TASKS
     LIB_ITTI gmock_main gtest gtest_main gmock
+    TASK_GRPC_SERVICE
```
electronjoe, line 36: Explanation: Though we do not need to link with `TAKS_GRPC_SERVICE` - we do need / want it's PUBLIC headers, and we want to guarantee it is built before this target (to ensure proto / grpc gen has occurred).

themarwhal (approved):

lte/gateway/c/core/oai/test/mock_tasks/grpc_mock.cpp
ssanadhya: @electronjoe , in keeping with https://github.com/magma/magma/issues/9861, should we expand all these include paths to be repo relative?

lte/gateway/c/core/oai/tasks/grpc_service/CMakeLists.txt
ulaskozat: Isn't inclusion of `S8ServiceImpl.cpp` done for that purpose?

lte/gateway/c/core/oai/test/mock_tasks/grpc_mock.cpp
electronjoe: Good point! Done!

lte/gateway/c/core/oai/tasks/grpc_service/CMakeLists.txt
electronjoe: Per @ssanadhya 's advice, I've fixed up the include of `S8ServiceImpl.h` to use the full repo-relative path.  As such, this particular change is not necessary and in fact may be not advised as we want to force people to use the full repo-relative path moving forward.  I just removed this line and re-pushed the PR.  But as things were in this PR prior to that change, you have to include `S8ServiceImp.h` in add_library if you want to make it accessible by default by those linking against the target. Happy to quick video chat if that's unclear (tough to write out). Also, as with most things CMake, there are other paths to achieving these goals...

ssanadhya (approved):

PR #10406, opened 2021-11-18
fix(orc8r): Fix code scanning alert - Disabled TLS certificate check …

### LINKED ISSUE #10405: Fix code scanning alert - Disabled TLS certificate check
Tracking issue for:
- [ ] https://github.com/magma/magma/security/code-scanning/10

nikolakn:
…#10405

Signed-off-by: Nikola Knezevic <nkcodeplus@gmail.com>

fix(orc8r): Fix code scanning alert - Disabled TLS certificate check

## Summary

adding constant in config instead of using it directly in code

### DIFF
--- a/orc8r/gateway/go/config/control_proxy_config.go
+++ b/orc8r/gateway/go/config/control_proxy_config.go
@@ -35,6 +35,7 @@ const (
 	DefaultBootstrapPort         = 443
 	DefaultLocalPort             = 8443
 	DefaultProxyCloudConnection  = true
+	DefaultSkipVerify            = true
 )
 
 // ControlProxyCfg represents control_proxy.yml configuration struct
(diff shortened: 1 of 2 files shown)


hcgatewood:
cc @emakeev for review on how this affects his golang agw

hcgatewood (approved):
Lgtm, but please wait to merge till we hear from @emakeev

themarwhal:
Removing the ready2merge label until we hear from @emakeev ? (see Hunter's comment above)

orc8r/gateway/go/config/control_proxy_config.go
```diff
@@ -46,27 +47,29 @@ type ControlProxyCfg struct {
...
+		RootCaFile:                DefaultRootCaFile,
+		GwCertFile:                DefaultGwCertFile,
+		GwCertKeyFile:             DefaultGwCertKeyFile,
+		LocalPort:                 DefaultLocalPort,
+		CloudAddr:                 "",
+		CloudPort:                 DefaultCloudPort,
+		BootstrapAddr:             "",
+		BootstrapPort:             DefaultBootstrapPort,
+		ProxyCloudConnection:      DefaultProxyCloudConnection,
+		CloudConnectionSkipVerify: DefaultSkipVerify,
```
emakeev, line 72: CloudBootstrapperSkipVerify: ...

orc8r/gateway/go/services/bootstrapper/service/rpc.go
```diff
@@ -101,7 +102,7 @@ func (b *Bootstrapper) getGrpcOpts(useProxy bool, cfg *config.ControlProxyCfg) [
 			}
 		} else {
 			tlsCfg = &tls.Config{
-				InsecureSkipVerify: true,
+				InsecureSkipVerify: cfg.CloudConnectionSkipVerify,
```
emakeev, line 105: func (b *Bootstrapper) getGrpcOpts(useProxy bool, cfg *config.ControlProxyCfg) ([]grpc.DialOption, error) {
...
...
...
} else if cfg.CloudBootstrapperSkipVerify {
    tlsCfg = &tls.Config{
				InsecureSkipVerify: true ,
    }
} else {
    return opts, fmt.Errorf("no root CA to verify server authenticity")
}
...

emakeev (commented):
The rpc.go is only using insecure cloud connection if there are neither Root CA nor OS cert chain available. And it only uses this connection for bootstrapping.

I'd call the new flag/option 'CloudConnectionSkipVerify' that you added something like: 'AllowInsecureBootstrap' or 'CloudBootstrapperSkipVerify' to avoid confusion and the logic should probably change to erroring out immediately if there are no root CA/OS CA chain and insecure cloud connection is not allowed (please see inline comments) it would prevent an attempt to connect that is going to fail anyway.

At a minimum, I'd rename the option, erroring out immediately is an optimization and is mostly nice to have.

PR #10470, opened 2021-11-20
feat(feg): N7 interface SM Policy creation flow

GANESH-WAVELABS:
## Summary
- Session controller receives proto CreateSessionRequest, converts to SMPolicyContextData
- Uses the openapi generated code to send the SMPolicyContextData to N7 PCF and received SmPolicyDecision
- Convert SmPolicyDecision to CreateSessionResponse proto

Continuation of #10298.

## Additional Information
This PR is dependant on #10298. Can be merged only after that is merged.


feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,107 @@
...
+	request *protos.CreateSessionRequest,
+) (*n7_sbi.SmPolicyDecision, string, error) {
+	if !srv.cfg.N7Config.DisableN7 {
+		reqBody := n7.GetSmPolicyContextDataN7(request, srv.cfg.N7Config.Client.NotifyApiRoot)
+		return srv.postSmPolicies(reqBody)
+	}
+	return &n7_sbi.SmPolicyDecision{
+		PccRules: &n7_sbi.SmPolicyDecision_PccRules{},
+	}, "", nil
+}
```
uri200: can you reverse this?
Normally it is easier to read when the if contains the exception , that it

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,107 @@
...
+	n7_sbi "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
+	"magma/feg/gateway/services/n7_n40_proxy/metrics"
+	"magma/feg/gateway/services/n7_n40_proxy/n7"
+	"magma/lte/cloud/go/protos"
+)
+
+// postSmPoliciesWrapper sends the request to PCF and get the response when N7 is enabled,
+// otherwise returns an emmpty SmpolicyDescision is returned so that omni-present rules
+// can be added to it
+func (srv *CentralSessionController) postSmPoliciesWrapper(
```
uri200: this function is doing more than post. The name is confusing

Also normally when we add a condition like this we add that to the name of the function
```
getSmPolicyRulesIfEnabled
```

The idea is the name of the function gives information of what will the result be, this way you dont even need to go to the function definition to understand it

Words like `Wrapper` normally never add  any value

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,107 @@
...
+		return srv.postSmPolicies(reqBody)
+	}
+	return &n7_sbi.SmPolicyDecision{
+		PccRules: &n7_sbi.SmPolicyDecision_PccRules{},
+	}, "", nil
+}
+
+func (srv *CentralSessionController) postSmPolicies(
+	reqBody *n7_sbi.PostSmPoliciesJSONRequestBody,
+) (*n7_sbi.SmPolicyDecision, string, error) {
```
uri200: i think you should remove `postSmPoliciesWrapper`  function, exit this function if `srv.disabled` and change the name of this function to `getSmPolicyRulesIfEnabled`

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,107 @@
...
+	if !srv.cfg.N7Config.DisableN7 {
+		reqBody := n7.GetSmPolicyContextDataN7(request, srv.cfg.N7Config.Client.NotifyApiRoot)
+		return srv.postSmPolicies(reqBody)
+	}
+	return &n7_sbi.SmPolicyDecision{
+		PccRules: &n7_sbi.SmPolicyDecision_PccRules{},
+	}, "", nil
+}
+
+func (srv *CentralSessionController) postSmPolicies(
```
uri200: add a description of what function does. Do that specially when the return parameters are not obvious (like string, which i do not know what it is unless i read the function)

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,107 @@
...
+	if !found {
+		err = fmt.Errorf("SmPolicyCreate request failure: Location header not found")
+		metrics.ReportCreateSmPolicy(err)
+		glog.Errorf("%s", err)
+		return nil, "", err
+	}
+	metrics.ReportCreateSmPolicy(nil)
+	glog.V(2).Infof("SmPolicy %s created", smPolicyUrl[0])
+
+	return resp.JSON201, smPolicyUrl[0], nil
```
uri200: This looks ugly. Why can't you return `resp` and leave the caller to pick the parameters they want to use.

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,107 @@
...
+		err = fmt.Errorf("SmPolicyCreate request failure: status-code=%d", resp.StatusCode())
+		metrics.ReportCreateSmPolicy(err)
+		glog.Errorf("%s", err)
+		return nil, "", err
+	}
+	smPolicyUrl, found := resp.HTTPResponse.Header["Location"]
+	if !found {
+		err = fmt.Errorf("SmPolicyCreate request failure: Location header not found")
+		metrics.ReportCreateSmPolicy(err)
+		glog.Errorf("%s", err)
```
uri200: I think it would be better to leave the caller to print the error. If you had more places where you are returning errors on this function you would need to print each one

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, "", err
+	}
+	metrics.ReportCreateSmPolicy(nil)
+	glog.V(2).Infof("SmPolicy %s created", smPolicyUrl[0])
```
uri200: same here, i think this should not be here. Maybe the caller at `session_controller` should print it

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return resp.JSON201, smPolicyUrl[0], nil
+}
+
+func (srv *CentralSessionController) injectOmnipresentRules(
+	smPolicyDecision *n7_sbi.SmPolicyDecision,
+) error {
+	// No Base-Names in N7
```
uri200: if no basenames, why are you adding them in the code and on the logs?
This will be confusiong for the user

What you can do is trigger a waringin if `omnipresentBaseNames` are included saying those are not going to take aney effect

Then  
```
glog.V(2).Infof("Adding omnipresent rules %v ", omnipresentRuleIDs)
```

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,107 @@
...
+		for _, ruleId := range omnipresentRuleIDs {
+			smPolicyDecision.PccRules.Set(ruleId, n7_sbi.PccRule{
+				PccRuleId: ruleId,
+			})
+		}
+	}
+	return nil
+}
+
+func checkCreateSessionRequest(req *protos.CreateSessionRequest) error {
```
uri200: `validate` instead of `check`

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
```diff
@@ -70,8 +63,21 @@ func (srv *CentralSessionController) CreateSession(
 	ctx context.Context,
 	request *protos.CreateSessionRequest,
 ) (*protos.CreateSessionResponse, error) {
+	glog.V(2).Info("Trying to create session")
```
uri200: Remove this
Use logs wisely

Orcr8r framework services have a flag where you can see GRPC messages coming to you, so it is easy to see when a create session starts.

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
```diff
@@ -70,8 +63,21 @@ func (srv *CentralSessionController) CreateSession(
...
+		return nil, err
+	}
+
+	smPolicyDecision, smPolicyId, err := srv.postSmPoliciesWrapper(request)
+	if err != nil {
+		return nil, err
+	}
+	err = srv.injectOmnipresentRules(smPolicyDecision)
+	if err != nil {
+		glog.Errorf("Failed to inject omnipresent rules")
```
uri200: Doing this you are hiding the error returned by `injectOmnipresentRules` fooling you when debugging. 

Either 
```
glog.Errorf("Failed to inject omnipresent rules: %s", err)
```
or in this case you don't need to add anything else
glog.Error(err)

uri200 (changes requested):

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
```diff
@@ -0,0 +1,336 @@
...
+		HTTPResponse: &http.Response{
+			StatusCode: 201,
+			Header: map[string][]string{
+				"Location": {SmPolicyUrl},
+			},
+		},
+	}
+}
+
+func newCreateSessionResponse(imsi string) *protos.CreateSessionResponse {
```
uri200: in this case you can use `generateDefaultCreateSessionResponse`

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
```diff
@@ -0,0 +1,336 @@
...
+func TestCreateSession(t *testing.T) {
+	srv, mockDb, mockN7 := newCentralSessionControllerForTest(t, false)
+
+	mockN7.On("PostSmPoliciesWithResponse", mock.Anything,
+		mock.MatchedBy(matchSmPolicyContextData(IMSI1_NOPREFIX)),
+	).Return(newSmPolicyResponse(t), nil).Once()
+
+	// no omnipresent rules
+	mockDb.On("GetOmnipresentRules").Return([]string{}, []string{}).Once()
+	// mockDb.On("GetRuleIDsForBaseNames", []string{"base_10"}).Return([]string{"base_rule_1", "base_rule_2"})
```
uri200: To save test cases, you can test all those rules here. You can add omni rules, static and dynamic and see if the create session looks right

Or you can create a separate test case

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
```diff
@@ -0,0 +1,336 @@
...
+	UeIpv4             = sbi.Ipv4Addr(UE_IPV4)
+	PduSessionTypeIpv4 = sbi.PduSessionType("IPV4")
+	RatTypeNR          = sbi.RatType("NR")
+	UeTzIST            = sbi.TimeZone("+05:30")
+	SmPolicyUrl        = "https://example.com//npcf-smpolicycontrol/v1/sm-policies/12345"
+	ActTime            = time.Unix(1634906551, 0)
+	DeactTime          = time.Unix(1634913751, 0)
+)
+
+func TestCreateSession(t *testing.T) {
```
uri200, line 73: this test should be big, including almost all parameters. 

See [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fs8_proxy%2Fservicers%2Fs8_proxy_test.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=59&start_col=28&end_row=59&end_col=28)  an example. This is not the best and proper way to do it, but it checks many cases at one time without having to repeat the same case forever.

Once you have tested the all parameters then you can add tests for more specific cases if needed (see on the same example file)

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
```diff
@@ -0,0 +1,336 @@
...
+
+	assert.Equal(t, genSessionID(IMSI1), response.SessionId)
+	allRuleIds := []string{}
+	for _, staticRule := range response.StaticRules {
+		allRuleIds = append(allRuleIds, staticRule.RuleId)
+	}
+	assert.ElementsMatch(t, []string{"omni_rule_1", "omni_rule_2", "base_rule_1", "base_rule_2"}, allRuleIds)
+}
+
+func newCentralSessionControllerForTest(t *testing.T, disableN7 bool) (*servicers.CentralSessionController, *mockPolicyDB.PolicyDBClient, *mockN7.ClientWithResponsesInterface) {
```
uri200: new is normally used to define something similar to a constructor.  And constructors are kind of definitions of different ways to build an object,

In this case you are just creating a default object 

```
createTestCentralSessionController
```

uri200: change all the new ones that are not constructors of the object

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
```diff
@@ -0,0 +1,336 @@
...
+			},
+		},
+		SessionId:      genSessionID(IMSI1),
+		AccessTimezone: &protos.Timezone{OffsetMinutes: 330},
+	}
+}
+
+func genSessionID(imsi string) string {
+	return fmt.Sprintf("%s-1234", imsi)
+}
```
uri200: i don't think you need this. Use a static value

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
```diff
@@ -0,0 +1,336 @@
...
+			"cond_data1": {
+				"condId": "cond_data1",
+				"activationTime": "2021-10-22T12:42:31Z",
+				"deactivationTime": "2021-10-22T14:42:31Z"
+			}
+		},
+		"policyCtrlReqTriggers": ["RE_TIMEOUT"],
+		"revalidationTime": "2021-10-22T14:42:31Z",
+		"online": true
+	}`
```
uri200, line 289: see you can use placeholders in case you need to change some parameters

for example 

```
policyDecisionStr := "pccRules": {
			"rule1": {
				"pccRuleId": "%s",
				....
				}

policyDecisionStr = fmt.Sprintf(policyDecisionStr, "ppc_rule_1")
```
You don't need to add in each field, but just in case you need to reuse `policyDecisionStr` in other places where the response may be different

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
```diff
@@ -0,0 +1,336 @@
...
+func genSessionID(imsi string) string {
+	return fmt.Sprintf("%s-1234", imsi)
+}
+
+func matchSmPolicyContextData(imsi string) interface{} {
+	return func(reqBody n7_sbi.PostSmPoliciesJSONRequestBody) bool {
+		return reqBody.Supi == sbi.Supi(imsi) &&
+			reqBody.NotificationUri == n7.GenNotifyUrl(NOTIFY_API_ROOT, genSessionID("IMSI"+imsi))
+	}
+}
```
uri200, line 221: Maybe this is checked somewhere else, but should't you have a matcher to see if all the parameters you requested did make it to the request you are sending?

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: The idea is to encapsulate the http responses and error handling within this method. Response is received in both the request body and the as a HTTP header. Hence multiple return values.

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: The comment was incorrect. Intention is to have the same behavior in case of omni-present rules for both Gx and N7.

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
GANESH-WAVELABS: The test does take care of all possible parameters.

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
GANESH-WAVELABS: This is required in some places where imsi is passed as an argument.

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,100 @@
...
+	n7_sbi "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
+	"magma/feg/gateway/services/n7_n40_proxy/metrics"
+	"magma/feg/gateway/services/n7_n40_proxy/n7"
+	"magma/lte/cloud/go/protos"
+)
+
+// getSmPolicyRulesIfEnabled sends the request to PCF and get the response when N7 is enabled,
+// otherwise returns an empty SmPolicyDecision is returned so that omni-present rules
+// can be added to it.
+// Returns SmPolicyDecision rules and smPolicyUrl that uniquely identifies the policy session.
```
uri200, line 30: the description of this function shouldn't add  logic from omnirules that is somewhere else. 

```
// getSmPolicyRulesIfEnabled if N7 is enabled gets policies from PCF otherwise returns an empty SmPolicyDecision 
// Returns SmPolicyDecision rules and smPolicyUrl that uniquely identifies the policy session.
```

uri200: I don't remember omni rules much, but why do they depend on this to empty

Per my understanding omni rules are omni rules, no matter if PCF has or not a response

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
uri200: I still see it weird, but that may be just me

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,100 @@
...
+	// Convert and send the request to PCF
+	reqBody := n7.GetSmPolicyContextDataN7(request, srv.cfg.N7Config.Client.NotifyApiRoot)
+	reqCtx, cancel := context.WithTimeout(context.Background(), srv.cfg.RequestTimeout)
+	defer cancel()
+	resp, err := srv.policyClient.PostSmPoliciesWithResponse(reqCtx, *reqBody)
+	if err != nil {
+		metrics.ReportCreateSmPolicy(err)
+		if errors.Is(err, context.DeadlineExceeded) {
+			metrics.N7Timeouts.Inc()
+		}
```
uri200: shouldn't you add this condition `if errors.Is` to the `ReportCreateSmPolicy` itself?

uri200: I still think metrics should be handled separately outside this function. 
The reason is: this function should do one thing, which is get something from PCF.  

Then another function should deal to understand the errors coming out from `getSmPolicyRulesIfEnabled` and determine the metrics

See the comment at the end of this pr

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
```diff
@@ -70,8 +63,19 @@ func (srv *CentralSessionController) CreateSession(
...
 ) (*protos.CreateSessionResponse, error) {
+	if err := validateCreateSessionRequest(request); err != nil {
+		return nil, err
+	}
 
-	return (&protos.UnimplementedCentralSessionControllerServer{}).CreateSession(ctx, request)
+	smPolicyDecision, smPolicyId, err := srv.getSmPolicyRulesIfEnabled(request)
+	if err != nil {
+		return nil, err
+	}
```
uri200, line 74:
```
	smPolicyDecision, smPolicyId, err := srv.getSmPolicyRulesIfEnabled(request)
	metrics.ReportCreateSmPolicy(err)
	if err != nil {
		return nil, err
	}
```

ReportCreateSessionPolicy should include any case to handle errors like 
```
if errors.Is....
```

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (srv *CentralSessionController) injectOmnipresentRules(
+	smPolicyDecision *n7_sbi.SmPolicyDecision,
+) error {
+	// No Base-Names returned in N7 response, however since these are statically configured,
+	// fetch all the omnipresent rules including the ones referred using the base-names. Will
+	// achieve the same behavior for both Gx and N7.
+	omnipresentRuleIDs, omnipresentBaseNames := srv.dbClient.GetOmnipresentRules()
+	if len(omnipresentRuleIDs) > 0 {
```
uri200: unnest this 

```
if len(omnipresentRuleIDs) == 0 {
	return nil
}

```

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,100 @@
...
+		glog.V(2).Infof("Adding omnipresent rules %v omnipresent basenames %v", omnipresentRuleIDs, omnipresentBaseNames)
```
uri200: please review my comments. Once you push force they may be outdated and not display here anymore, so you need to back trace them.

Per my understanding you dont have basenames, isnt it?

uri200 (changes requested):

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: If we return a nil reponse, omni rules cannot be added. Anyways changing the comment description as it can be confusing.

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: Base names are not present in the N7 response. However, for omni present rules, which we get it from policy db we should fetch both the static omni present rules and the omni present basename rules. The basename rules again translate to static rules. Basenames provide a way to group static rules.

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
GANESH-WAVELABS: This is done.

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
GANESH-WAVELABS: Added some placeholders

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
uri200: Why don't you add 

```
		if errors.Is(err, context.DeadlineExceeded) {
			metrics.N7Timeouts.Inc()
		}
```

To the metrics function? Shouldn't you trigger a timeout everytime there is a timeout?

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,95 @@
...
+	smPolicyDecision *n7_sbi.SmPolicyDecision,
+) error {
+	// No Base-Names returned in N7 response, however since these are statically configured,
+	// fetch all the omnipresent rules including the ones referred using the base-names. Will
+	// achieve the same behavior for both Gx and N7.
+	omnipresentRuleIDs, omnipresentBaseNames := srv.dbClient.GetOmnipresentRules()
+	if len(omnipresentRuleIDs) == 0 {
+		return nil
+	}
+	glog.V(2).Infof("Adding omnipresent rules %v omnipresent basenames %v", omnipresentRuleIDs, omnipresentBaseNames)
```
uri200, line 71: please check the comment i made about this earlier

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
```diff
@@ -70,8 +63,20 @@ func (srv *CentralSessionController) CreateSession(
 	ctx context.Context,
 	request *protos.CreateSessionRequest,
 ) (*protos.CreateSessionResponse, error) {
+	if err := validateCreateSessionRequest(request); err != nil {
+		return nil, err
+	}
 
-	return (&protos.UnimplementedCentralSessionControllerServer{}).CreateSession(ctx, request)
+	smPolicyDecision, smPolicyId, err := srv.getSmPolicyRulesIfEnabled(request)
```
uri200: `smPolicyDecision` is a weird name that makes me think there is more than a policy there. But checking the definition [here](smPolicyDecision) I can see it is just a policy. 

What about if we use easier names like 
```
policy, policyId, err := srv.getSmPolicyRulesIfEnabled(request)
```

Adding the SM everywhere may be overkill

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
uri200: Then there is a problem here. Omni Rules are mainly used when there is no PCF. So we should be able to add omnirules even if no response from PCF

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
uri200: Just define a variable on the top of the file with that value

```
sessionID1 :=  fmt.Sprintf("%s-1234", IMSI1)
sessionID1 :=  fmt.Sprintf("%s-1234", IMSI2)
```

The reason for that is when debugging a test it is easier to see how the values were generated.

Also in test is better to be verbose some times. Meaning, doing this on a regular file would be bad, but on a test may be helpful because is easier to understand for debugging

uri200 (changes requested):

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: Omni rules are set if there is a PCF and adds additional rules to the ones pushed from PCF. In case the N7 is disabled we only provide the Omni rules. The comment is now changed.

What I meant in the previous comment was the return value of nil rather than an empty Policy decision. An empty policy decision or the one returned by PCF allows the addition of omni rules. However a nil return value will not be able to do that.

Here is the logic behind adding Omni basename rules
- When N7 is disabled fetch the omni static rules and fetch the other omni static rules grouped together with a basename.
- Even when PCF is present add  omni static rules and fetch the other omni static rules grouped together with a basename, in addition to the rules returned by PCF. Though PCF does not send basename rules, to retain the same behavior of Omni rules as in 4G, N7_N40_proxy is adding them to the response.

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: Moves this to metrics.

feg/gateway/services/n7_n40_proxy/servicers/session_controller.go
GANESH-WAVELABS: The logic behind adding sm prefix is to make it clear the type is a N7 type and to differentiate it from a proto type. However will change the name as suggested.

feg/gateway/services/n7_n40_proxy/servicers/create_session_test.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: I'm sorry I'm not able to identify the previous comment. Can you please point out which one it is?

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,90 @@
...
+	if !found {
+		err = fmt.Errorf("SmPolicyCreate request failure: Location header not found")
+		return nil, "", err
+	}
+	return resp.JSON201, policyUrl[0], nil
+}
+
+func (srv *CentralSessionController) injectOmnipresentRules(
+	policy *n7_sbi.SmPolicyDecision,
+) error {
```
uri200: try to use one line function if it fits.
Note when you do build.py -c it will break the line if necessary

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
```diff
@@ -0,0 +1,90 @@
...
+	// achieve the same behavior for both Gx and N7.
+	omnipresentRuleIDs, omnipresentBaseNames := srv.dbClient.GetOmnipresentRules()
+	if len(omnipresentRuleIDs) == 0 {
+		return nil
+	}
+	glog.V(2).Infof("Adding omnipresent rules %v omnipresent basenames %v", omnipresentRuleIDs, omnipresentBaseNames)
+	baseNameRuleIds := srv.dbClient.GetRuleIDsForBaseNames(omnipresentBaseNames)
+	omnipresentRuleIDs = append(omnipresentRuleIDs, baseNameRuleIds...)
+	for _, ruleId := range omnipresentRuleIDs {
+		policy.PccRules.Set(ruleId, n7_sbi.PccRule{
```
uri200, line 75: just in case  add a check to make sure neither policy nor PccRules are nil

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
uri200: Why is tis function returning error?  I don't see any being returned

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
uri200: Ok understood. Then maybe we go back to the original name `getSmPolicyRules`. The reason now is that either if it is enabled or not, we will get a policy rule answer, not matter if it is enabled or not.

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
uri200: Per what you mentioned BaseNames do not exist on 5g. Why do we need them then?

And if you are injecting the definition of all the rules from the base name, can you point me where?

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: Done

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: Done.

feg/gateway/services/n7_n40_proxy/servicers/create_session.go
GANESH-WAVELABS: Rules Ids of all BaseName groups are obtained in line no 71 (next line).

All static rules including the ones obtained from base name grouping is set in line no 74. 
`policy.PccRules.Set(ruleId, n7_sbi.PccRule{` 

The static rules that are set in the PccRules are converted to Static Rules proto in `model_conversion.go:getPccRules()` (#10298). `getpccRules()` is invoked when `GetCreateSessionResponseProto()` is invoked after `injectOmnipresentRules()`.

uri200: Got it, i was missing `omnipresentBaseNames` are set on orc8r.

uri200 (approved):

uri200:
rebase it and I think it will merge by itself

PR #10474, opened 2021-11-21
fix(amf): ambr values of profile not being sent to sessiond.

### LINKED ISSUE #10218: [Qos]: Configure AMBR values are not getting used while establishing PDU-Session.
- Version: magma/focal-ci, now 1.7.0-1636428543-ef82f726 
- Affected Component : Access Gateway
- Affected Subcomponent : AGW service
- Deployment Environment : Vagrant (AGW)

Describe the Issue

AMF is taking default ambr value , even APN ambr's are configured. 

**To Reproduce**
1. Configure an APN with ambr values other than default values.
2. Try to configure a Qos-Profile with ambr values.
3. Bind the Qos-profile and APN on the subscriber. 
4. Trigger the PDU session from UE with AON name configure.
5. Check that PDU-Session is coming up with configured ambr's by validating the packets.
6. Noticed that configured ambr's values are not getting reflected.

**Expected behavior**
PDU-Session should come up with configured values. 

**Screenshots**

Configuration Screenshots:

**Additional context**
Please find the PCAP's and logs from attachment.[https://app.zenhub.com/files/170803235/438de22b-5401-4b0c-8e48-f3f504328e28/download](https://app.zenhub.com/files/170803235/438de22b-5401-4b0c-8e48-f3f504328e28/download)

### LINKED ISSUE #10220: [Qos]: Encoding is not proper in the downlink-PDUEstablishmentAccept message.
### Your Environment

Version: magma/focal-ci, now 1.7.0-1636428543-ef82f726
Affected Component : Access Gateway
Affected Subcomponent : AGW service
Deployment Environment : Vagrant (AGW)

### Describe the Issue

To Reproduce
1. Configure an APN with ambr values other than default values.
2. Try to configure a Qos-Profile with ambr values.
3. Bind the Qos-profile and APN on the subscriber.
4. Trigger the PDU session from UE with AON name configure.
5. Check that PDU-Session is coming up with configured ambr's by validating the packets.
6. Noticed that configured ambr's values are not getting decoded properly.

**Expected behavior**
PDU-Session should come up with configured values.

**Screenshots**
[https://app.zenhub.com/files/170803235/3f0fb030-b4cf-45ca-b6eb-942209f48d89/download](https://app.zenhub.com/files/170803235/3f0fb030-b4cf-45ca-b6eb-942209f48d89/download)

**Additional context**

Lkishor123:
Signed-off-by: LKishor123 <laawanya.kishor@wavelabs.ai>

## Summary
1. Fixed: [Qos]: Configure AMBR values are not getting used while establishing PDU-Session.

subscriber ambr profile in subscriberDB.
apn_5_subscriberDB

grpc call log between sessiond and amf
syslog_apn_5_ambr

2. Fixed: [Qos]: Encoding is not proper in the downlink-PDUEstablishmentAccept message.
ambr_encode

3. Code Refactoring: Removed ambr parameter from mobilityd calls.

## Test Plan

1. UERANSIM

2. Unit Test
unit_test

## Additional Information


lte/gateway/c/core/oai/common/common_defs.h
```diff
@@ -120,6 +120,35 @@ typedef enum {
   S8_SUBSCRIBER    = 2,  // will use federated HSS + s8_task
 } mode_map;
 
+typedef enum {
+  value_not_used    = 000000000,
+  Multiples_1kbps   = 0b00000001,
```
ardzoht: these enum values should be on uppercase

lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
```diff
@@ -769,24 +786,8 @@ int amf_smf_notification_send(
...
-    return rc;
-  }
-
-  smf_ctx = amf_get_smf_context_by_pdu_session_id(ue_context, pdu_session_id);
-  if (NULL == smf_ctx) {
-    return rc;
-  }
-
+    const std::shared_ptr<smf_context_t>& smf_ctx, paa_t* address_info) {
+  OAILOG_INFO(LOG_AMF_APP, "smf context pdu address updated\n");
```
ardzoht:
```suggestion
  OAILOG_INFO(LOG_AMF_APP, "SMF context PDU address updated\n");
```

ardzoht (changes requested):

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
```diff
@@ -660,6 +661,55 @@ int amf_app_handle_pdu_session_response(
...
+ **  one defined in create_pdu_session_response to one defined in          **
+ **  pdu_session_estab_accept message                                      **
+ **                                                                        **
+ ** Inputs:  pdu_ambr_response_unit, pdu_ambr_response_value               **
+ **          ambr_unit, ambr_value                                         **
+ **                                                                        **
+ **      Return:   void                                                    **
+ **                                                                        **
+ ***************************************************************************/
+void convert_ambr(
```
ulaskozat, line 677: this conversion does not make too much sense to me. If say the rate is 1.9Mbps, you would be encoding it as 1 Mbps? You should instead define thresholds for the unit changes (e.g., in kbps or in bps) and encode accordingly (e.g., if rate is above X kbps and below Y kbps, use unit Z).

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
```diff
@@ -660,6 +661,55 @@ int amf_app_handle_pdu_session_response(
...
+    uint32_t* pdu_ambr_response_unit, uint32_t* pdu_ambr_response_value,
+    uint8_t* ambr_unit, uint16_t* ambr_value) {
+  int count = 0;
+  if (*pdu_ambr_response_unit == KBPS) {
+    *pdu_ambr_response_value *= 1024;
```
ulaskozat: 1 kbps = 1000 bps not 1024 bps. (unlike memory storage units).

ulaskozat: also if multiplying/dividing by powers of 2, use bit shift operations.

lte/gateway/c/core/oai/common/common_defs.h
ulaskozat: +1

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
ulaskozat: add unit test.

lte/gateway/c/core/oai/common/common_defs.h
Lkishor123: Sure, I'll update it.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: @ulaskozat , still we won't be able to cover all decimal as finally the values need to be in `uint16`.  `You should instead define thresholds for the unit changes` : Do you mean rounding off to nearest integer?

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: Sure, I'll update it.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: should 1.9 Mbps be converted to 2Mbps?

Lkishor123: Do you mean if it's 1.9 Mbps, I should use 1900 kbps?

ulaskozat: Yes, you should encode it such that there is minimum loss of rate information when you convert the value. You should find the absolute thresholds with this objective. It is clear that with 16bits, we have 2^16 * units * (bps/unit). A simple approximation would go to the next unit if the rate to be converted is not covered with smaller unit.

lte/gateway/c/core/oai/common/common_defs.h
Lkishor123: Comment addressed.

lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
Lkishor123: comment addressed.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: @ulaskozat , please review the new function. I've made 65535 as threshold value.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: Comment addressed.

lte/gateway/c/core/oai/test/amf/test_amf_encode_decode.cpp
```diff
@@ -1465,4 +1465,56 @@ TEST(test_optional_pdu, test_pdu_session_accept_optional) {
   sm_free_protocol_configuration_options(&msg_accept_pco);
 }
 
+TEST(test_ambr, test_ambr_covert) {
```
ulaskozat: covert -> convert

Also I don't think `test_amf_encode_decode.cpp` is the right place to test common utility functions. cc @panyogesh

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
```diff
@@ -660,6 +661,62 @@ int amf_app_handle_pdu_session_response(
...
+ **                                                                        **
+ **      Return:   void                                                    **
+ **                                                                        **
+ ***************************************************************************/
+void convert_ambr(
+    uint32_t* pdu_ambr_response_unit, uint32_t* pdu_ambr_response_value,
+    uint8_t* ambr_unit, uint16_t* ambr_value) {
+  int count = 1;
+  if (*pdu_ambr_response_unit == KBPS) {
+    *pdu_ambr_response_value *= 1000;
```
ulaskozat: this would not work if `pdu_ambr_response_calue` is larger than MAX/1000. Also what is the contract of this function? Are you supposed to modify `pdu_ambr_response_value`? Use consts in decleration and definitions if they are not supposed to be modified.

ulaskozat: if unit is KBPS, you can simply start count from 2 otherwise it is at 1.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
```diff
@@ -660,6 +661,62 @@ int amf_app_handle_pdu_session_response(
...
+  }
+  uint32_t temp_pdu_ambr_response_value = *pdu_ambr_response_value;
+  if (temp_pdu_ambr_response_value / 1000 == 0) {
+    count = 1;
+  } else {
```
ulaskozat: You do not need if/else as you can initialize the count based on the `pdu_ambr_response_unit`.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
```diff
@@ -660,6 +661,62 @@ int amf_app_handle_pdu_session_response(
...
+      *ambr_unit = static_cast<uint8_t>(
+          magma5g::M5GSessionAmbrUnit::MULTIPLES_1GBPS);  // Gbps
+      break;
+    default:
+      *ambr_unit = static_cast<uint8_t>(
+          magma5g::M5GSessionAmbrUnit::MULTIPLES_1GBPS);  // Gbps
+      break;
+  }
+  int k       = pow(1000, (count - 1));
+  *ambr_value = static_cast<uint16_t>(*pdu_ambr_response_value / k);
```
ulaskozat: you should be able to get the value from `temp_pdu_ambr_response_value`.

ulaskozat (changes requested):
More comments left.

lte/gateway/c/core/oai/test/amf/test_amf_encode_decode.cpp
Lkishor123: I changed the location for that unit test function.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: Comment addressed.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: Comment addressed.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: Comment addressed.

Lkishor123: Comment addressed.

Lkishor123:
> More comments left.

Comments addressed.

lte/gateway/c/core/oai/test/amf/test_amf_encode_decode.cpp
Lkishor123: I changed the location for it.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: Comment addressed.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: Comment addressed.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: Comment addressed.

Lkishor123:
> More comments left.

Comments addressed.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
```diff
@@ -660,6 +660,60 @@ int amf_app_handle_pdu_session_response(
...
+void convert_ambr(
+    const uint32_t* pdu_ambr_response_unit,
+    const uint32_t* pdu_ambr_response_value, uint8_t* ambr_unit,
+    uint16_t* ambr_value) {
+  int count = 1;
+  if (*pdu_ambr_response_unit == KBPS) {
+    count = 2;
+  }
+  uint32_t temp_pdu_ambr_response_value = *pdu_ambr_response_value;
+  if (temp_pdu_ambr_response_value / 1000 != 0) {
```
ulaskozat: why do you need this if statement? should not this block simply be:
``` 
while (temp_pdu_ambr_response_value >= AMBR_UNIT_CONVERT_THRESHOLD) {
     temp_pdu_ambr_response_value = (temp_pdu_ambr_response_value / 1000);
     count++;
}
```

Lkishor123: Initially I implemented a do while loop, hence I wrote that if statement there. I'll remove it.

Lkishor123: @ulaskozat  `if (temp_pdu_ambr_response_value / 1000 != 0)` is a necessary check. The main reason we raised this PR is to fix the bug in zenhub task : #10220. In this bug is there because we are not converting the sessiond itti response ambr format. We are sending absolute value and ambr_unit as 0. In pcap we can see that this is treated as illegal value. This means that anything less than kbps is illegal. But if I remove the if condition it will start making all kbps values upto 65535 as illegal value.

Lkishor123: @ulaskozat `if (temp_pdu_ambr_response_value / 1000 != 0)` is a necessary check. The main reason we raised this PR is to fix the bug in zenhub task : #10220. This bug is there because we are not converting the sessiond itti response ambr format. We are sending absolute value and ambr_unit as 0. In pcap we can see that this is treated as illegal value. This means that anything less than kbps is illegal. But if I remove the if condition it will start making all kbps values upto 65535 as (bps) that is illegal value.

ulaskozat: So `case 1` on line 694 should never occur? This is absolutely not clearly documented anywhere in this function implementation. If KBPS is the minimum unit, than the logic should check `if (*pdu_ambr_response_unit == BPS)` on line 682 instead and if so divide it by 1000. The count then should be initialized as 1 and original case 1 can be KBPS, case 2 can be MBPS, and so on. Also default case looks like an error (this should not happen at all given input is uint32_t) and should be handled accordingly.

ulaskozat: if `pdu_ambr_response_value` is less that 1kbps and with units bps, you need to round it up to 1Kbps. so you need to cover that as a special case.

Lkishor123: @ulaskozat yes, default case will not occur. But other than that the logic works fine. Because if we send anything less than 1000 BPS eg: `987 BPS`, it will be converted as `987` and uint will be `Value not used`. Hence it will be forwarded to gnb as illegal value [Suggesting that operator/user should check subscription data], which is the case. As ambr value was subscribed like that (In subscriberDB). But if we increase the value by ourself we are over ruling the subscription data as well as policy information. As this conversion takes place after validation by sessiond (Validation of subscription data vs policy).

panyogesh: @Lkishor123 : Lets fix it for 1Kbps.

ulaskozat: @Lkishor123 as long as the documentation clearly states that AMBR is set minimum at 1Kbps regardless of policy, it is not an issue. Note that we currently use TC and it cannot really enforce anything smaller than 10Kbps.

Lkishor123: @ulaskozat understood, I'll update it.

Lkishor123:
## Illegal ambr value (< 1KBPS) will be converted to 1 KBPS.

apn_1

apn_1_pcap

subscriberDB_ambr

Lkishor123:
# Illegal ambr value (< 1KBPS) will be converted to 1 KBPS.

## Subscribed_ambr:
apn_1


## Ambr received by gnb:
ambr

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
Lkishor123: @ulaskozat, I've updated the code, please review.

Lkishor123: @ulaskozat, I've updated the code and rebased with latest master, please review.

lte/gateway/c/core/oai/tasks/amf/amf_app_handler.cpp
```diff
@@ -660,6 +660,61 @@ int amf_app_handle_pdu_session_response(
...
+  if (*pdu_ambr_response_unit == BPS and
+      temp_pdu_ambr_response_value / 1000 == 0) {
+    // Values less than 1Kbps are defaulted to 1Kbps
+    *ambr_value = static_cast<uint16_t>(1);
+    *ambr_unit  = static_cast<uint8_t>(
+        magma5g::M5GSessionAmbrUnit::MULTIPLES_1KBPS);  // Kbps
+    return;
+  }
+
+  if (*pdu_ambr_response_unit == BPS) {
```
ulaskozat, line 694: add a comment that minimum rate unit is KBPS

Lkishor123: Comment addressed.

ulaskozat (approved):
LGTM

ulaskozat:
@ardzoht this is waiting for your approval.

ardzoht (approved):
lgtm

PR #10483, opened 2021-11-22
fix(ci): Only trigger integ tests on push and not on PR

quentinDERORY:

## Summary
Currently LTE/CWF integ tests are launched on PR which should not happen

## Additional Information


tmdzk (approved):

talkhasib (approved):
LGTM

PR #10490, opened 2021-11-22
fix: Update BASE_REFERENCE in AMIS workflow

jwagantall:
Set BASE_REFERENCE to "master" for schedule triggered
runs since github.base_ref is not available.
Use github.base_ref for workflow triggers.

## Additional Information

### DIFF
--- a/.github/workflows/amis-workflow.yml
+++ b/.github/workflows/amis-workflow.yml
@@ -35,8 +35,10 @@ jobs:
         run: |
             if ${{ github.event_name == 'workflow_dispatch' }}; then
               echo AWS_ACCOUNT="LF" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/amis-workflow.yml
```diff
@@ -35,8 +35,10 @@ jobs:
         run: |
             if ${{ github.event_name == 'workflow_dispatch' }}; then
               echo AWS_ACCOUNT="LF" >> $GITHUB_ENV
+              echo BASE_REFERENCE=${{ github.base_ref }} >> $GITHUB_ENV
```
quentinDERORY, line 38: can we use it on line 54 too?

quentinDERORY (approved):

PR #10495, opened 2021-11-23
feat(orc8r): setup for registration servicers

reginawang3495:
## Summary
- setup protos for use in making `cloudRegistration` servicer and `registration servicer`
- setup storage
- setup constants needed

## Test Plan
- added tests
- run all orc8r tests

## Additional Information

Part of some Stacked PRs
1. [Current PR on a fork](https://github.com/reginawang3495/magma/pull/6)
   - has some review comments
2. [CloudRegistration PR](https://github.com/reginawang3495/magma/pull/4)
   - on a fork
3. [Add registration client_api.go PR](https://github.com/reginawang3495/magma/pull/12)
   - on a fork
4. [Registration PR](https://github.com/reginawang3495/magma/pull/11)
   - on a fork
5. [Add Servicers to Bootstrapper Service PR](https://github.com/reginawang3495/magma/pull/7)
   - on a fork


orc8r/cloud/go/services/bootstrapper/const.go
```diff
@@ -13,4 +13,15 @@ limitations under the License.
...
 package bootstrapper
 
-const ServiceName = "BOOTSTRAPPER"
+const (
+	ServiceName        = "BOOTSTRAPPER"
+	BlobstoreTableName = "bootstrapper"
+	TokenPrepend       = "reg_"
+)
+
+type DBBlobType string
```
hcgatewood: The fact that this backs to a DB is an impl detail

```suggestion
type BlobType string
```

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/pkg/errors"
+
+	"magma/orc8r/cloud/go/blobstore"
+	"magma/orc8r/cloud/go/services/bootstrapper"
+	"magma/orc8r/cloud/go/storage"
+	merrors "magma/orc8r/lib/go/errors"
+	"magma/orc8r/lib/go/protos"
+)
+
+const networkWildcard = "*"
```
hcgatewood: https://github.com/magma/magma/blob/2104a1c17c4a773dca84d5397665e2db89d565a2/orc8r/cloud/go/services/accessd/storage/storage_blobstore.go#L34-L36

And as a bonus, could move that placeholder to a const in a shared orc8r pkg

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
```diff
@@ -0,0 +1,163 @@
...
+	GetTokenInfoFromLogicalID(networkID string, logicalID string) (*protos.TokenInfo, error)
+	GetTokenInfoFromNonce(nonce string) (*protos.TokenInfo, error)
+}
+
+type blobstoreStore struct {
+	factory blobstore.StoreFactory
+}
+
+func NewBlobstoreStore(factory blobstore.StoreFactory) Store {
+	return &blobstoreStore{factory}
```
hcgatewood: Nit: always construct struct literals with the "field: value" pattern, rather than passing in values anonymously

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (b *blobstoreStore) SetTokenInfo(tokenInfo protos.TokenInfo) error {
+	isUnique, err := b.isNonceUnique(tokenInfo.Nonce)
```
hcgatewood: We want this check to occur within the same tx -- so after the "start transaction"

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	nonceBlob, err := tokenInfoToBlob(bootstrapper.NonceTokenToInfoMap, tokenInfo.Nonce, tokenInfo)
+	if err != nil {
+		return err
+	}
+	err = store.Write(networkWildcard, blobstore.Blobs{nonceBlob})
+	if err != nil {
+		return err
+	}
```
hcgatewood: Add a comment explaining why we're inserting two objects

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	tokenInfo, err := tokenInfoFromBlob(logicalIDBlob)
+	if err != nil {
+		return nil, err
+	}
+
+	return &tokenInfo, store.Commit()
```
hcgatewood: Here and throughout: rather than returning addr here, return a ptr from tokenInfoFromBlob

hcgatewood (changes requested):
Looks good! Couple small things, mainly we want to perform the isunique check within same tx

orc8r/protos/bootstrapper.proto
```diff
@@ -71,3 +73,83 @@ service Bootstrapper {
   // Returns signed certificate.
   rpc RequestSign (Response) returns (Certificate) {}
 }
+
+// --------------------------------------------------------------------------
+// Protos for CloudRegistration and Registration servicer
+// --------------------------------------------------------------------------
```
andreilee, line 79: This file is organized a little strangely with the additions. A typical pattern we've followed for `.proto` files is to to group the messages under the corresponding service definition.

orc8r/protos/bootstrapper.proto
```diff
@@ -71,3 +73,83 @@ service Bootstrapper {
...
+service CloudRegistration {
+  // GetToken saves and sends out a token related to input:networkID and input:logicalID
+  // It will refresh the token if input:refresh is true or token is expired
+  rpc GetToken(GetTokenRequest) returns (GetTokenResponse) {}
```
andreilee, line 85: Adding comments for the service definition can be helpful in explaining how this is intended to be used at a high level. As a new developer, I would wonder how the token/nonce is used, but it's not clear where to look after seeing this service definition.

orc8r/protos/bootstrapper.proto
```diff
@@ -71,3 +73,83 @@ service Bootstrapper {
...
+  string token = 1;
+  AccessGatewayID hwid = 2;
+  // challenge_key is gateway's long-term public key
+  ChallengeKey challenge_key = 3;
+}
+
+message RegisterResponse {
+  oneof response {
+    // control_proxy is the OK that the gateway has registered properly
+    string control_proxy = 1;
```
andreilee: What sort of values would `control_proxy` taken on if the gateway has registered correctly?

andreilee (changes requested):

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
reginawang3495: ok created tix for me later: https://github.com/magma/magma/issues/10533

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
reginawang3495: `IsNonceUnique` calls `GetTokenInfoFromNonce` which has its own transaction. Would you prefer for me to pull out the part inside of the transaction from `GetTokenInfoFromNonce`? Or leave this as is

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
reginawang3495: When return ptr vs value?

orc8r/protos/bootstrapper.proto
reginawang3495: They are pretty connected though, and I think it might be easier to understand if they next to each other?

orc8r/protos/bootstrapper.proto
reginawang3495: added example in protos

orc8r/protos/bootstrapper.proto
reginawang3495: Added a bit of a written diagram -- bloats the code a bit

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
reginawang3495: Why?

hcgatewood: Easier to skim -- when there are lots of struct fields, it's unclear which args are which fields. And then we want uniformity even when there is only 1 field

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
hcgatewood: If we're using txs, the check needs to be in the same tx as the action. So whichever impl you think works best for that

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
hcgatewood: Consistency is more important for our purposes. I'm fine with either, but let's not be converting back and forth internally. Ptr is probably a little easier since protos generally flow better as ptrs, but up to you

orc8r/cloud/go/services/bootstrapper/const.go
```diff
@@ -13,4 +13,15 @@ limitations under the License.
 
 package bootstrapper
 
-const ServiceName = "BOOTSTRAPPER"
+const (
+	ServiceName        = "BOOTSTRAPPER"
+	BlobstoreTableName = "bootstrapper"
+	TokenPrepend       = "reg_"
```
hcgatewood:
```suggestion
	TokenPrefix       = "reg_"
```

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
hcgatewood: Let's go ahead and make it "placeholder_network" in this PR, since changing things like this later generally requires a data migration

orc8r/protos/bootstrapper.proto
```diff
@@ -71,3 +73,114 @@ service Bootstrapper {
   // Returns signed certificate.
   rpc RequestSign (Response) returns (Certificate) {}
 }
+
+// --------------------------------------------------------------------------
+// Protos for CloudRegistration and Registration servicer
+// --------------------------------------------------------------------------
+/**
```
hcgatewood: I find this block comment to be a little under-useful -- I'd rather this info be in a sequence diagram in the docusaurus, as I think it would be easier to read there

I'm not opposed to putting this here, but I just think it won't get read as-is

orc8r/protos/bootstrapper.proto
```diff
@@ -71,3 +73,114 @@ service Bootstrapper {
...
+  Registration<->CloudRegistration:
+    GetGatewayRegistrationInfo(nonce): network_id, logical_id
+
+  Registration->Device:
+    RegisterDevices(network_id, hwid)
+
+Registration->AGW:control_proxy
+ */
+
+service CloudRegistration {
```
hcgatewood, line 82: To Andrei's point: let's add doc comments directly to the `service` definitions here, same as we do for the endpoints within the service

orc8r/protos/bootstrapper.proto
```diff
@@ -71,3 +73,114 @@ service Bootstrapper {
...
+  rpc GetGatewayDeviceInfo(GetGatewayDeviceInfoRequest) returns (GetGatewayDeviceInfoResponse) {}
+}
+
+service Registration {
+  // Register should be called by the gateway device
+  rpc Register(RegisterRequest) returns (RegisterResponse) {}
+}
+
+message GetTokenRequest {
+  GatewayDeviceInfo gateway_device_info = 1;
```
hcgatewood, line 104: Each field should have a golang-style doc comment

orc8r/protos/bootstrapper.proto
```diff
@@ -71,3 +73,114 @@ service Bootstrapper {
...
+}
+
+message GetTokenResponse {
+  // token is a nonce prepended by bootstrapper.tokenPrepend
+  string token = 1;
+  google.protobuf.Timestamp timeout = 2;
+}
+
+message GetGatewayRegistrationInfoRequest {
+}
```
hcgatewood: Nit: can just be 1 line

orc8r/protos/bootstrapper.proto
```diff
@@ -71,3 +73,114 @@ service Bootstrapper {
...
+message RegisterResponse {
+  oneof response {
+    // control_proxy is the OK that the gateway has registered properly
+    // EG magma/lte/gateway/configs/control_proxy.yml
+    string control_proxy = 1;
+    string error = 2;
+  }
+}
+
+message GatewayDeviceInfo {
```
hcgatewood, line 169: Add doc comment for all message types that aren't Request/Responses

hcgatewood (changes requested):
Looking good! Couple small things, main request is the txs and the shared network name

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
reginawang3495: Hm I copied this one: https://github.com/magma/magma/blob/e8c7ef943cf80fc8c2883b448a0cc1e09ce5e5a3/orc8r/cloud/go/services/tenants/servicers/storage/storage.go#L15

Would you like me to call it "placeholder_network" in this PR?

orc8r/protos/bootstrapper.proto
reginawang3495: Ok, made a tix for documenting later :)) #10576

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
reginawang3495: placeholder_network

andreilee (commented):
lgtm. Deferring approval to @hcgatewood

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
```diff
@@ -0,0 +1,170 @@
...
+		Key:  nonce,
+	}
+	nonceBlob, err := store.Get(placeholderNetworkID, nonceTK)
+	if err != nil {
+		if err == merrors.ErrNotFound {
+			return true, nil
+		}
+		return false, err
+	}
+	return nonceBlob.Value == nil, nil
```
hcgatewood: If the blob isn't found, ErrNotFound is returned. So this edge case isn't worth checking -- what does it mean for the blob to exist but have a nil value?

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
```diff
@@ -0,0 +1,170 @@
...
+		Type: string(bootstrapper.NonceTokenToInfoMap),
+		Key:  nonce,
+	}
+	nonceBlob, err := store.Get(placeholderNetworkID, nonceTK)
+	if err != nil {
+		if err == merrors.ErrNotFound {
+			return true, nil
+		}
+		return false, err
+	}
```
hcgatewood: Nit for uniformity and reducing cyclomatic complexity, don't nest the err checking. Eg https://github.com/magma/magma/blob/4a29306de8048c40749a6f59135a018ef1d7a2ce/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go#L294-L299

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
```diff
@@ -0,0 +1,170 @@
...
+	merrors "magma/orc8r/lib/go/errors"
+	"magma/orc8r/lib/go/protos"
+)
+
+const placeholderNetworkID = "placeholder_network"
+
+type Store interface {
+	SetTokenInfo(tokenInfo protos.TokenInfo) error
+	GetTokenInfoFromLogicalID(networkID string, logicalID string) (*protos.TokenInfo, error)
+	GetTokenInfoFromNonce(nonce string) (*protos.TokenInfo, error)
```
hcgatewood: We're mixing whether we treat protos.TokenInfo as a value or a ptr. Instead let's be consistent about it. Preference for using ptr, but up to you

orc8r/protos/bootstrapper.proto
```diff
@@ -71,3 +73,88 @@ service Bootstrapper {
...
+message RegisterResponse {
+  oneof response {
+    // control_proxy is the OK that the gateway has registered properly
+    // EG magma/lte/gateway/configs/control_proxy.yml
+    string control_proxy = 1;
+    string error = 2;
+  }
+}
+
+// Basic info that an AGW needs to register
```
hcgatewood: Nit: we want these comments to still be golang-style, so e.g. 
```suggestion
// GatewayDeviceInfo contains basic info that a gateway needs to register
```

orc8r/protos/bootstrapper.proto
```diff
@@ -71,3 +73,88 @@ service Bootstrapper {
...
+message GatewayDeviceInfo {
+  string network_id = 1;
+  string logical_id = 2;
+}
+
+// Info saved that is keyed by token. This is what is saved on CloudRegistration's initial register call (GetToken)
+message TokenInfo {
+  GatewayDeviceInfo gateway_device_info = 1;
+  string nonce = 2;
```
hcgatewood, line 185: We want to provide doc comments for every proto field

hcgatewood (approved):
Looks great! Fix couple small requests, and looks like insync-checkin is failing (and cloud lint), then msg me and I'll merge 👍

orc8r/cloud/go/services/bootstrapper/servicers/registration/storage.go
reginawang3495: nothing -- wasnt sure the exact impl of blobstore of whether nil could be a ret val

orc8r/protos/bootstrapper.proto
reginawang3495: In general?  Most seem pretty self explanatory and redundant

hcgatewood: Yep. Makes things more approachable, and since the comments get replicated into generated code, its a double-win

hcgatewood:
Force-merging to unblock

PR #10007, opened 2021-11-01
fix(mme): Disabling S1 handover integ test from sanity list

ardzoht:
## Summary

- S1 Handover sanity S1ap test is failing locally and on CI, this PR disables the `s1_handover` test for the moment while a resolution is provided.

## Test Plan
- `make integ_test` on local VM

## Additional Information

- [ ] This change is backwards-breaking

PR #10506, opened 2021-11-23
fix(agw): Re-enabled S1 handover test case as part of sanity

VinashakAnkitAman:
## Title
Re-enabled S1 handover test case as part of sanity

## Summary
The S1 Handover test cases were added as part of PR #9523 which was reverted in PR #10007 because the S1 Handover test cases were failing in CI as part of the issue #10123. The S1APTester PR #66 fixes the issue and the test cases are ready to be reenabled under this PR.

## Test plan
Verified with s1 handover test cases individually and as part of sanity/non-sanity


ardzoht (approved):
lgtm

PR #10509, opened 2021-11-23
fix(agw): Updated non sanity configuration

VinashakAnkitAman:
## Title
Updated non sanity configuration

## Summary
The non-sanity configuration updation script is configuring ipv6 solicitation which is now by default set in latest master codebase. Running non-sanity was disabling this configuration after execution and causing some test cases to fail. This PR removes the configuration and fixes the issue

## Test plan
Verified with non-sanity and individual test case execution after that.


ssanadhya (approved):
LGTM, thanks for the clean up!

PR #10518, opened 2021-11-24
feat(agw): Add ghz docker files and playbooks

ardzoht:

## Summary

- Adds Dockerfile and docker-compose files for building gHZ based containers based on c and python docker AGW services
- Adds  `ghz-provision.yaml` playbook to automate setup of containers for load testing
- Adds roles to `ghz-provision.yaml` playbook to apply necessary configuration changes to run load testing

## Test Plan

- tested on local and cloudstrapper GW

```
PLAY [localhost] *************************************************************************************************************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] *******************************************************************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [ghz-platform : install docker-compose pydep] ***************************************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [ghz-platform : copy redis conf template] *******************************************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [ghz-platform : apply changes to pipelined config] **********************************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [ghz-platform : bring up docker agw services] ***************************************************************************************************************************************************************************************************************************************************************************
changed: [localhost]

PLAY RECAP *******************************************************************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=5    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0

```

## Additional Information

### DIFF
--- a/lte/gateway/docker/ghz/Dockerfile
+++ b/lte/gateway/docker/ghz/Dockerfile
@@ -1,0 +1,52 @@
+# AGW test docker private repository location
+ARG REPO_LOCATION=agw-test.artifactory.magmacore.org
+ARG MAGMA_ROOT=/magma
+ARG BASE_VERSION=latest
+
+FROM ${REPO_LOCATION}/agw_gateway_c AS c_builder
+
+ARG GHZ_REPO=https://github.com/bojand/ghz
+
+RUN apt-get update && apt-get install -y \
+  git \
+  golang \
+  build-essential
+
+WORKDIR ${MAGMA_ROOT}
+
+RUN git clone --depth 1 ${GHZ_REPO}
+
+WORKDIR ${MAGMA_ROOT}/ghz
+
+RUN make build
+
+FROM ${REPO_LOCATION}/agw_gateway_python as python_builder
+
+ARG GHZ_REPO=https://github.com/bojand/ghz
+
+RUN apt-get update && apt-get install -y \
+  git \
+  golang \
+  build-essential
+
+ENV MAGMA_ROOT /magma
+
+WORKDIR ${MAGMA_ROOT}
+
+RUN git clone --depth 1 ${GHZ_REPO}
+
+WORKDIR ${MAGMA_ROOT}/ghz
+
+RUN make build
+
+FROM ${REPO_LOCATION}/agw_gateway_c as agw_c_ghz
+
+COPY --from=c_builder /magma/ghz/dist/ghz /usr/local/bin/
+
+WORKDIR ${MAGMA_ROOT}/lte/gateway/python/load_tests
+
+FROM ${REPO_LOCATION}/agw_gateway_python as agw_python_ghz
+
+COPY --from=python_builder /magma/ghz/dist/ghz /usr/local/bin/
+
+WORKDIR ${MAGMA_ROOT}/lte/gateway/python/load_tests
--- a/lte/gateway/docker/ghz/docker-compose.yaml
+++ b/lte/gateway/docker/ghz/docker-compose.yaml
@@ -1,0 +1,147 @@
+---
+version: "3.7"
+
+# Standard logging for each service
+x-logging: &logging_anchor
+  driver: ${LOG_DRIVER}
+
+# Standard volumes mounted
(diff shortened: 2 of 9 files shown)


ulaskozat (approved):
LGTM

lte/gateway/docker/ghz/docker-compose.yaml
```diff
@@ -0,0 +1,147 @@
+---
```
arunuke, line 1: If this file is largely similar to the [base docker-compose](https://github.com/magma/magma/blob/master/lte/gateway/docker/docker-compose.yaml) file to startup all the services, would it help if we use the same docker-compose file as above, but only startup selective set of containers (perhaps using the -d option with the container name?)

This would ensure that we have a single source file to maintain and capture changes for the common dev-ops, scale-test and other use-cases.

arunuke (changes requested):

lte/gateway/docker/ghz/docker-compose.yaml
ardzoht: @arunuke This is indeed really similar to our docker AGW `docker-compose` file, the only important change is the definition of the `image:` on both services (`image: ${DOCKER_REGISTRY}/ghz_gateway_python:${IMAGE_VERSION}`) 

is there any way we could use the base file and just change this value? maybe defining a variable on the `.env` file?

arunuke: How about using the [ansible module for docker-compose](https://docs.ansible.com/ansible/latest/collections/community/docker/docker_compose_module.html) which supports the 'services' option which allows the user to specify only a subset of services to start? The link includes an example where you use the docker-compose module within and specify a set of services and any custom changes.

ardzoht: Looks like that ansible module does not support changing the `image` definition (which is different from the base `docker-compose.yaml` file)

lte/gateway/docker/ghz/Dockerfile
```diff
@@ -0,0 +1,52 @@
+# AGW test docker private repository location
```
themarwhal, line 1: Do we need a copyright header here?

lte/gateway/docker/ghz/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+ARG REPO_LOCATION=agw-test.artifactory.magmacore.org/
```
themarwhal: It might look nicer to leave the trailing slash out so that using it looks like `${REPO_LOCATION}/`.  Just a style preference though, feel free to ignore :)

lte/gateway/docker/ghz/Dockerfile
ardzoht: None of our other Dockerfiles have copyright header added, maybe this can be added on a separate PR but that's why I left it like that

quentinDERORY (approved):
LGTM
Note on the publish step: I think it's not worth putting it in ansible as all publishing scripts are put directly in GHA workflows.
THe other cloudstrapper script usually have the option to build the artifacts on the host (in this case docker images) from scratch when not specifying any repo. (But I don't recall any publishing stuff)
When I will setup the agw container build workflow in gha I will simply call the docker-compose and publish and not use ansible.
I plan to use ansible only for the automation of the load test

ardzoht:
> LGTM Note on the publish step: I think it's not worth putting it in ansible as all publishing scripts are put directly in GHA workflows. THe other cloudstrapper script usually have the option to build the artifacts on the host (in this case docker images) from scratch when not specifying any repo. (But I don't recall any publishing stuff) When I will setup the agw container build workflow in gha I will simply call the docker-compose and publish and not use ansible. I plan to use ansible only for the automation of the load test

Sounds good, I removed the `build-ghz` role since the build and publishing steps can be called separately and are not needed in ansible

arunuke (approved):
LGTM.

PR #10522, opened 2021-11-24
feat(agw): Added ipv6 paging support

pruthvihebbani:
Signed-off-by: Pruthvi Hebbani <pruthvi.hebbani@radisys.com>

feat(agw): Added ipv6 paging support
## Summary

1. Added ipv6 paging rules verification
2. Updated ipv6 address to ip_map

## Test Plan

1. Verified s1ap tester sanity
2. Verified the ipv6 TC by manually sending data using the script. Script and TC files attached below:
[ipv6PagingTest.zip](https://github.com/magma/magma/files/7594329/ipv6PagingTest.zip)


lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -589,6 +597,12 @@ def generate_imsi(self, prefix=None):
         print("Using subscriber IMSI %s" % imsi)
         return imsi
 
+    def update_ipv6_address(self, ue_id, ipv6_addr):
```
ulaskozat, line 600: where is this used?

pruthvihebbani: This function will be used in the test script to update ipv6 address to _ue_ip_map like how it is done for ipv4 address during attach. This will later be used to send traffic request to trfserver for data path.

ulaskozat (approved):
LGTM

PR #10538, opened 2021-11-24
fix(mme): Fix the semantics and bug in timer management

ulaskozat:

## Summary

- Fix the bug in timer management where expired timers were not removed from the map. Since expired timer ids can be circulated by ZMQ for new timer events, this would lead to silent insertion failures and fetching timer arguments for an expired timer.
- To clarify the semantics, replaced all gets by pops.
- Added a unit test (which originally helped to catch this bug) for t3485 and t3495 timers.

To Do:
- having separate timer management for AMF is simply wrong. Move mme and amf timers to a common library.

## Test Plan

Unit tests and s1ap integ tests.

## Additional Information


ardzoht (approved):
lgtm

ssanadhya (approved):
LGTM

PR #10544, opened 2021-11-24
feat: pull down Bazel build cache to run agw-build and lte-integ-test…

themarwhal:
… jobs

## Summary
Soon, I will be flipping the default build system for sessiond/sctpd/liagentd/connectiond to Bazel from CMake. As the Bazel build system will build more dependencies from scratch, the build time will increase. To manage the impact in CI, we will pull down a prebuilt cache that is refreshed daily.

Tested with this PR: https://github.com/magma/magma/pull/10513

## Additional Information

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -12,6 +12,11 @@ on:  # yamllint disable-line rule:truthy
       - v1.*
     types: [opened, reopened, synchronize]
 
+env:
+  S3_BUCKET: "magma-cache"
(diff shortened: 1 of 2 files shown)


quentinDERORY:
just wondering don't we want to wait to have the changes ready to use this cache before always pulling it in these jobs?

themarwhal:
> just wondering don't we want to wait to have the changes ready to use this cache before always pulling it in these jobs?

Yes not committing this until I have the PR to flip it. Just trying to get all the pieces out.

quentinDERORY (approved):

PR #10551, opened 2021-11-26
feat(orc8r): Disable logging and metrics in orc8r helm chart if staging is enabled

joary:

## Summary

Disable specific resources for a orc8r-app module deployment marked as staging (meaning the same resources already exists from main orc8r).

Changelog:
- Added variable to flag orc8r-app module is a staging one.
- On Terraform use the flag to disable resources that are already deployied in the main Orc8r:
    - EFS Provisioner
    - Tiller & its rolebindings
    - Namespace for monitoring (wich already exists)
    - External DNS helm deployment
    - Fluentd for orc8r-app logging
    - Elasticsearch-curator for orc8r-app metrics
    - Disabled logging and metrics from orc8r helm chart
- On orc8r helm chart, make logging and metrics sub-charts to be configurable instead of always deployied

## Test Plan

1. Deploy the main orc8r without any special configuration, including both modules orc8r & orc8r-helm-aws in the main.tf.

2. Create a new orc8r-helm-aws module marked as staging with variable `orc8r_is_staging_deployment=true` and with `orc8r_kubernetes_namespace = "staging"`

Expected: 
- Terraform should run without problems (not being blocked by duplicated resources)
- New namespace should be created and the orc8r & lte-orc8r pods deployed.
- Neither Logging/Monitoring/Metrics resources will be created.

## Additional Information

### DIFF
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/efs.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/efs.tf
@@ -13,6 +13,8 @@
 
 # k8s requires provisioner to treat efs as a persistent volume
 resource "helm_release" "efs_provisioner" {
+  count = var.orc8r_is_staging_deployment == true ? 0 : 1
+
   name       = var.efs_provisioner_name
   repository = local.stable_helm_repo
   chart      = "efs-provisioner"
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/helm.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/helm.tf
@@ -13,7 +13,7 @@
 
 # helm tiller service account
 resource "kubernetes_service_account" "tiller" {
-  count = var.existing_tiller_service_account_name == null ? 1 : 0
+  count = var.install_tiller == true ? 1 : 0
 
   metadata {
     name      = "tiller"
@@ -25,7 +25,7 @@ resource "kubernetes_service_account" "tiller" {
 
 # helm tiller cluster role
 resource "kubernetes_cluster_role_binding" "tiller" {
-  count = var.existing_tiller_service_account_name == null ? 1 : 0
+  count = var.install_tiller == true ? 1 : 0
 
   metadata {
     name = "tiller"
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/k8s.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/k8s.tf
@@ -18,6 +18,8 @@ resource "kubernetes_namespace" "orc8r" {
 }
 
 resource "kubernetes_namespace" "monitoring" {
+  count = var.orc8r_is_staging_deployment == true ? 0 : 1
+
   metadata {
     name = var.monitoring_kubernetes_namespace
   }
@@ -25,6 +27,8 @@ resource "kubernetes_namespace" "monitoring" {
 
 # external dns maps route53 to ingress resources
 resource "helm_release" "external_dns" {
+  count = var.orc8r_is_staging_deployment == true ? 0 : 1
+
   name       = var.external_dns_deployment_name
   repository = local.stable_helm_repo
   chart      = "external-dns"
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/main.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/main.tf
@@ -212,5 +212,9 @@ data "template_file" "orc8r_values" {
     thanos_store_selector   = var.thanos_store_node_selector != "" ? format("compute-type: %s", var.thanos_store_node_selector) : "{}"
 
     region = var.region
+
+    # Staging deployments does not deploy neither logging nor metrics
+    enable_logging = !var.orc8r_is_staging_deployment
+    enable_metrics = !var.orc8r_is_staging_deployment
   }
 }
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/yace.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/yace.tf
@@ -54,7 +54,7 @@ resource "helm_release" "yace_exporter" {
   count = var.cloudwatch_exporter_enabled ? 1 : 0
 
   name       = "yace-exporter"
-  namespace  = kubernetes_namespace.monitoring.metadata[0].name
+  namespace  = kubernetes_namespace.monitoring[0].metadata[0].name
   repository = "https://mogaal.github.io/helm-charts/"
   chart      = "prometheus-yace-exporter"
   timeout    = 600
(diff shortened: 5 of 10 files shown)


orc8r/cloud/deploy/terraform/orc8r-helm-aws/main.tf
```diff
@@ -212,5 +212,9 @@ data "template_file" "orc8r_values" {
     thanos_store_selector   = var.thanos_store_node_selector != "" ? format("compute-type: %s", var.thanos_store_node_selector) : "{}"
 
     region = var.region
+
+    # Staging deployments will does not deploy neighter logging nor metrics
```
hcgatewood: Unclear wording

orc8r/cloud/deploy/terraform/orc8r-helm-aws/helm.tf
```diff
@@ -13,7 +13,8 @@
 
 # helm tiller service account
 resource "kubernetes_service_account" "tiller" {
-  count = var.existing_tiller_service_account_name == null ? 1 : 0
+  # Deploy tiller only if it does not exist yet and we're in a non-staging environment
+  count = var.existing_tiller_service_account_name == null && !var.orc8r_is_staging_deployment ? 1 : 0
```
hcgatewood: I think it would be valuable to create an additional layer of indirection, similar to how you're doing it in the helm charts -- `enableLogging` is the variable these modules actually care about. And "is staging deployment" is just 1 way we could decide to disable logging

hcgatewood (approved):
Lgtm, one request

orc8r/cloud/deploy/terraform/orc8r-helm-aws/helm.tf
joary: Cool, we had an existing variable to enable tiller, I just used it here.

Though the cost we have to pay is to include both `orc8r_is_staging_deployment` and `install_tiller` in main.tf instead of just the first one.

orc8r/cloud/deploy/terraform/orc8r-helm-aws/helm.tf
```diff
@@ -13,7 +13,7 @@
 
 # helm tiller service account
 resource "kubernetes_service_account" "tiller" {
-  count = var.existing_tiller_service_account_name == null ? 1 : 0
+  count = var.install_tiller == true ? 1 : 0
```
arunuke, line 16: Helm v3 has completely eliminated tiller and our eventual helm3 upgrade should eliminate any tiller references as well. Not a change required for this PR, but something to capture as a future issue.

orc8r/cloud/deploy/terraform/orc8r-helm-aws/efs.tf
```diff
@@ -13,6 +13,8 @@
 
 # k8s requires provisioner to treat efs as a persistent volume
 resource "helm_release" "efs_provisioner" {
+  count = var.orc8r_is_staging_deployment == true ? 0 : 1
```
arunuke, line 16: To state my understanding, we don't create a new efs provisioner since we intend to use the same backend storage for the new services as well. Is that a fair understanding?

orc8r/cloud/deploy/terraform/orc8r-helm-aws/helm.tf
```diff
@@ -25,7 +25,7 @@ resource "kubernetes_service_account" "tiller" {
 
 # helm tiller cluster role
 resource "kubernetes_cluster_role_binding" "tiller" {
-  count = var.existing_tiller_service_account_name == null ? 1 : 0
+  count = var.install_tiller == true ? 1 : 0
```
arunuke, line 28: Same as before w.r.t tiller deprecation in the future. I understand why we don't need a new tiller object. but not quite clear on the variable name changing here.

orc8r/cloud/deploy/terraform/orc8r-helm-aws/k8s.tf
```diff
@@ -18,13 +18,17 @@ resource "kubernetes_namespace" "orc8r" {
...
+  count = var.orc8r_is_staging_deployment == true ? 0 : 1
+
   metadata {
     name = var.monitoring_kubernetes_namespace
   }
 }
 
 # external dns maps route53 to ingress resources
 resource "helm_release" "external_dns" {
+  count = var.orc8r_is_staging_deployment == true ? 0 : 1
```
arunuke, line 30: To clarify, we need to add external DNS entries for the "Green" cluster. I will address this as part of the PR that I will push shortly.

orc8r/cloud/deploy/terraform/orc8r-helm-aws/yace.tf
```diff
@@ -54,7 +54,7 @@ resource "helm_release" "yace_exporter" {
   count = var.cloudwatch_exporter_enabled ? 1 : 0
 
   name       = "yace-exporter"
-  namespace  = kubernetes_namespace.monitoring.metadata[0].name
+  namespace  = kubernetes_namespace.monitoring[0].metadata[0].name
   repository = "https://mogaal.github.io/helm-charts/"
```
arunuke, line 58: There would only be one namespace for monitoring (since we will not create one if orc8r_is_staging_deployment is true). Given that, why would be need a [0] for monitoring snamespace?

arunuke (changes requested):

orc8r/cloud/deploy/terraform/orc8r-helm-aws/efs.tf
joary: Yes, to add more color this would be used to hold the prometheus configurations for metrics, since we do not have a prometheus in the new namespace we disabled this EFS as well.

If we enable it, the creation will fail since the efs_file_system_id would still be configured to the main orc8r value.

orc8r/cloud/deploy/terraform/orc8r-helm-aws/helm.tf
joary: Previously we were using both existing_tiller_service_account_name & orc8r_is_staging_deployment to disable tiller.

Hunter proposed to use just the one above and configure it all in main.tf

orc8r/cloud/deploy/terraform/orc8r-helm-aws/yace.tf
joary: The problem I found was: when I add count [terraform sees the resource as a list](https://www.terraform.io/docs/language/expressions/references.html#resources) not an object anymore. Then access with dot-notation fails and I had to include the bracket notation.

arunuke: Thanks for clarifying @joary

arunuke (approved):
Looks good to me. @joary , can you please create an issue for helm v3 upgrade and label it as "controller"?

joary:
Followup issue to track helm v3 and tiller: https://github.com/magma/magma/issues/10863

PR #10552, opened 2021-11-26
gRPC message for SMF-AMF/AMF-SMF for Session Modification

prabinakpattnaik:

## Summary

The Network Initiated Session Modification Support for SMF-AMF grpc communication and vice-versa.

1) Introduced new RPC method for SMF_AMF modification Request 
     rpc SetSmfSessionModContext(SetSMSessionModContextAccess) returns (SmContextVoid);
2) Introduced new SMSessionFSMState as a MODIFICATION_5 for AMF to SMF modification response.

Reauth_5g


uri200 (commented):
Note sessiond does not currently support modification of rules. We just add and remove rules. 

Is modification of rules really necessary? If so, I would first review sessiond to understand how modification of rules would have to be implemented.

PR #10553, opened 2021-11-27
chore(orc8r): add feature to remove verbose protos from orc8r service…

uri200:
…s log

## Summary

Visualizing protos in go services is key feature for troubleshooting. However, since we are printing any kind of proto, sometimes we get spammed by protos that are too verbose and do not provide any insight (like metrics periodic update)

This PR adds a new option to reduce the amount of protons to be displayed. 

The list of protos to be hidden is right now hardcoded. If necessary in the future we can move that list to the config file and add some regex style syntax to make this feature more powerful if. For now we will test this way.

## Test Plan

`./build.py -c`
## Additional Information

### DIFF
--- a/orc8r/lib/go/service/logcodec.go
+++ b/orc8r/lib/go/service/logcodec.go
@@ -16,6 +16,9 @@ package service
 import (
 	"bytes"
 	"fmt"
+	"os"
+	"strconv"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/jsonpb"
@@ -24,15 +27,62 @@ import (
 	grpc_proto "google.golang.org/grpc/encoding/proto"
 )
 
+// logModes
+type grpcLogVerbosityLevel int
+
+const (
+	GRPCLOG_DISABLED    = grpcLogVerbosityLevel(iota) // Default value
+	GRPCLOG_FULL                                      // Prints all protos
+	GRPCLOG_HIDEVERBOSE                               // Prints all except verboseProtos
+	grpclog_end
+)
+
+var (
+	// Default list of verbose protos
+	verboseProtos = map[string]grpcLogVerbosityLevel{
+		"protos.Void":                  GRPCLOG_HIDEVERBOSE,
+		"protos.ServiceInfo":           GRPCLOG_HIDEVERBOSE,
+		"protos.HealthStatus":          GRPCLOG_HIDEVERBOSE,
+		"protos.MetricsContainer":      GRPCLOG_HIDEVERBOSE,
+		"protos.SubmitMetricsRequest":  GRPCLOG_HIDEVERBOSE,
+		"protos.SubmitMetricsResponse": GRPCLOG_HIDEVERBOSE,
+		"grpc.MetricFamilies":          GRPCLOG_HIDEVERBOSE,
+		"grpc.Void":                    GRPCLOG_HIDEVERBOSE,
+	}
+)
+
 // logCodec is a debugging Codec implementation for protobuf.
 // It'll be used if debug GRPC printout is enabled
 type logCodec struct {
-	protoCodec encoding.Codec
+	protoCodec     encoding.Codec
+	verbosityLevel grpcLogVerbosityLevel
 }
 
-func printMessage(prefix string, v interface{}) {
+// Marshal of GRPC Codec interface
+func (lc logCodec) Marshal(v interface{}) ([]byte, error) {
+	printMessage("Sending: ", v, lc.verbosityLevel)
+	return lc.protoCodec.Marshal(v)
+}
+
+// Unmarshal of GRPC Codec interface
+func (lc logCodec) Unmarshal(data []byte, v interface{}) error {
+	err := lc.protoCodec.Unmarshal(data, v)
+	printMessage("Received: ", v, lc.verbosityLevel)
+	return err
+}
+
+// Name of GRPC Codec interface
+func (logCodec) Name() string {
+	return grpc_proto.Name
+}
+
+func printMessage(prefix string, v interface{}, verboseLevel grpcLogVerbosityLevel) {
 	var payload string
 	if pm, ok := v.(proto.Message); ok {
+		if verboseLevel == GRPCLOG_HIDEVERBOSE && isProtoHidden(pm) {
+			// do not print verbose proto
+			return
+		}
 		var buf bytes.Buffer
 		err := (&jsonpb.Marshaler{EmitDefaults: true, Indent: "\t", OrigName: true}).Marshal(&buf, pm)
 		if err == nil {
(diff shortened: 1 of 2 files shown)


orc8r/lib/go/service/logcodec.go
```diff
@@ -24,15 +25,41 @@ import (
...
+		"protos.Void":                  true,
+		"protos.ServiceInfo":           true,
+		"protos.HealthStatus":          true,
+		"protos.MetricsContainer":      true,
+		"protos.SubmitMetricsRequest":  true,
+		"protos.SubmitMetricsResponse": true,
+		"grpc.MetricFamilies":          true,
+		"grpc.Void":                    true,
+	}
+)
```
emakeev: Why don't you make DISABLED/VERBOSE/SHORT a type and make this type a value in your proto logging map:

```
type ProtosLogVerbocity int

const (
	DISABLED = ProtosLogVerbocity(iota)
	VERBOSE
	SHORT
)

var (
	// Default list of verbose protos
	logVerbosity = map[string]ProtosLogVerbocity{
		"protos.Void":                  SHORT,
		"protos.ServiceInfo":           SHORT,
		"protos.HealthStatus":          SHORT,
		"protos.MetricsContainer":      SHORT,
		"protos.SubmitMetricsRequest":  SHORT,
		"protos.SubmitMetricsResponse": SHORT,
		"grpc.MetricFamilies":          SHORT,
		"grpc.Void":                    SHORT,
	}
)

...

type logCodec struct {
	protoCodec encoding.Codec
	logMode    ProtosLogVerbocity
}

```
etc.

Then make VERBOSE a default (not in the map) and convert the env variable to int/ProtosLogVerbocity

uri200: The reason to have strings instead of ints is for backwards compatibility (due to this [line](https://github.com/magma/magma/pull/10553/files#diff-5130dd3e34b85d89a4f73c2661cb115d910f4a242f620ff9425c8af3e0f00f34L108)). Basically envar `PrintGrpcPayloadEnv` could be `1`, `true`, `True` or anything else. To avoid issues what i do is, if if `PrintGrpcPayloadEnv` contains any value, then i nable the full log. If that value matches with string `"2"` then I enable the HIDE feature. 

If i add the ints then, `PrintGrpcPayloadEnv: True` will not b valid anymore (or will requiere special case)

andreilee: I'm strongly in favor of Evgeniy's suggestion here. `logMode` field is too permissive as a `string` type. Another issue with the current implementation here is that it decouples the solution (logMode being string) far away from the issue (PrintGrpcPayloadEnv being multiple different values we support). 

I think a good way to deal with `PrintGrpcPayloadEnv` having different potential values is:
1. `service.go` should by default set `logMode` field to `FULL`, or to `HIDEVERBOSE` if `PrintGrpcPayloadEnv` == "2"
2. We really should throw errors if the input values for an env variable is invalid, or `PrintGrpcPayloadFlag` has an invalid value. If we're too permissive, then we don't really have a well documented public interface.

orc8r/lib/go/service/service.go
```diff
@@ -106,7 +109,10 @@ func NewServiceWithOptionsImpl(moduleName string, serviceName string, serverOpti
 
 	// Check if service was started with print-grpc-payload flag or MAGMA_PRINT_GRPC_PAYLOAD env is set
 	if printGrpcPayload || util.IsTruthyEnv(PrintGrpcPayloadEnv) {
-		ls := logCodec{encoding.GetCodec(grpc_proto.Name)}
+		ls := logCodec{
+			protoCodec: encoding.GetCodec(grpc_proto.Name),
+			logMode:    strings.TrimSpace(fmt.Sprintf(os.Getenv(PrintGrpcPayloadEnv))),
```
andreilee: logMode should ideally be an enum field type or equivalent, and be given a default, or set based on `PrintGrpcPayloadEnv`

andreilee (changes requested):

orc8r/lib/go/service/service.go
uri200: Note that the idea for this was to maintain backwards compatibility 

Current code enables grpc printing as long as long as you have any value different than 0 or False. 

I can change that and make it an enum (so only 0, 1, 2... will be good for that). But then this may stop working for anybody that is using it on a different way

orc8r/lib/go/service/logcodec.go
uri200: For concern 1, default value is 0 (disabled). And this is current behaviour. We should not enable grpc logging by default due to its verbosity and its possible impact on performance. That is why we keep it separate from the glog level. 

For concern 2, as mentioned in my other comment, this will change the current behaviour. But in case we change that, then it is a good idea to trigger an error in case the value is not right (and default it to 0). 

Our documentation specifies to use 1 to enable grpc logging. But our code says otherwise. So no gap on documentation.

orc8r/lib/go/service/logcodec.go
```diff
@@ -24,15 +27,62 @@ import (
...
+// Unmarshal of GRPC Codec interface
+func (lc logCodec) Unmarshal(data []byte, v interface{}) error {
+	err := lc.protoCodec.Unmarshal(data, v)
+	printMessage("Received: ", v, lc.verbosityLevel)
+	return err
+}
+
+// Name of GRPC Codec interface
+func (logCodec) Name() string {
+	return grpc_proto.Name
```
uri200, line 76: those functions have just been moved here. No change except for `printMessage` arguments

orc8r/lib/go/service/service.go
```diff
@@ -59,7 +56,8 @@ var defaultKeepaliveParams = keepalive.ServerParameters{
 }
 
 func init() {
-	flag.BoolVar(&printGrpcPayload, PrintGrpcPayloadFlag, false, "Enable GRPC Payload Printout")
+	flag.IntVar(&printGrpcPayload, PrintGrpcPayloadFlag, int(GRPCLOG_DISABLED),
+		"Enable GRPC Payload Printout (0: disabledl 1: enabled 2: hide verbose")
```
andreilee: nit: typo
```suggestion
		"Enable GRPC Payload Printout (0: disabled 1: enabled 2: hide verbose")
```

orc8r/lib/go/service/logcodec.go
```diff
@@ -24,15 +27,62 @@ import (
 	grpc_proto "google.golang.org/grpc/encoding/proto"
 )
 
+// logModes
+type grpcLogVerbosityLevel int
+
+const (
+	GRPCLOG_DISABLED    = grpcLogVerbosityLevel(iota) // Default value
+	GRPCLOG_FULL                                      // Prints all protps
```
andreilee: nit: typo
```suggestionx
	GRPCLOG_FULL                                      // Prints all protos
```

orc8r/lib/go/service/logcodec.go
```diff
@@ -24,15 +27,62 @@ import (
...
+	verboseProtos = map[string]grpcLogVerbosityLevel{
+		"protos.Void":                  GRPCLOG_HIDEVERBOSE,
+		"protos.ServiceInfo":           GRPCLOG_HIDEVERBOSE,
+		"protos.HealthStatus":          GRPCLOG_HIDEVERBOSE,
+		"protos.MetricsContainer":      GRPCLOG_HIDEVERBOSE,
+		"protos.SubmitMetricsRequest":  GRPCLOG_HIDEVERBOSE,
+		"protos.SubmitMetricsResponse": GRPCLOG_HIDEVERBOSE,
+		"grpc.MetricFamilies":          GRPCLOG_HIDEVERBOSE,
+		"grpc.Void":                    GRPCLOG_HIDEVERBOSE,
+	}
```
andreilee, line 51: Not sure if I have a recommendation here, but I do want to point out that with the way things are currently structured, we might need to include `/lte` protobufs, or `/feg` protobufs, and those dependencies should typically go the other way, at least based on how `xjtian` originally intended the plugin system to be used.

cc: @hcgatewood for visibility

andreilee (approved):

orc8r/lib/go/service/logcodec.go
uri200: Since this is a debug tool, I wanted to pollute as less as possible service with extra modules, that is why I used hardcoded strings instead of structs itself.

But you guys are ok with those dependencies under lib  I can add those there.


I will land this for now. We can change this later otherwise

PR #10557, opened 2021-11-29
chore(agw): implement shared mconfig and use it for Sentry config

alexzurbonsen:
## Summary

As part of #9508 all the Sentry config for Python services should be migrated to shared mconfig.
- implement loading of shared mconfig (follow up to #9964)
- use it to initialize Sentry SDK
- remove Sentry config from control_proxy.yml

## Test Plan

Locally tested that config can be read.
Included shared_mconfig in respective unit test.

### DIFF
--- a/orc8r/gateway/python/magma/common/sentry.py
+++ b/orc8r/gateway/python/magma/common/sentry.py
@@ -14,11 +14,12 @@ import logging
 import os
 from enum import Enum
 from functools import wraps
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Tuple
 
 import sentry_sdk
 import snowflake
 from magma.configuration.service_configs import get_service_config_value
+from orc8r.protos.mconfig import mconfigs_pb2
 
 CONTROL_PROXY = 'control_proxy'
 SENTRY_CONFIG = 'sentry'
@@ -26,6 +27,7 @@ SENTRY_URL = 'sentry_url_python'
 SENTRY_SAMPLE_RATE = 'sentry_sample_rate'
 CLOUD_ADDRESS = 'cloud_address'
 ORC8R_CLOUD_ADDRESS = 'orc8r_cloud_address'
+DEFAULT_SAMPLE_RATE = 1.0
 COMMIT_HASH = 'COMMIT_HASH'
 HWID = 'hwid'
 SERVICE_NAME = 'service_name'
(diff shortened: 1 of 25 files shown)


lte/gateway/configs/control_proxy.yml
```diff
@@ -39,14 +39,3 @@ proxy_cloud_connections: True
 
 # Allows http_proxy usage if the environment variable is present
 allow_http_proxy: False
-
```
themarwhal: Let's keep these in place because
1. We have people who use it still
2. We still want to be able to override the values on a per-agw basis. (This is probably useful while we're testing out all the Sentry enablement changes. Once we can confirm we've gotten rid of noise and any performance risks, we can just have the configuration be all or nothing per-network.)

So I would modify the PR to always take the control_proxy value as a prescendence if it's not empty during this migration phase.

alexzurbonsen: Good point. Will take care of it.

alexzurbonsen: @themarwhal The PR now reads the config from control_proxy.yml first (if sentry_url is present) :)

orc8r/gateway/python/magma/common/sentry.py
```diff
@@ -83,30 +85,56 @@ def get_sentry_status(service_name: str) -> SentryStatus:
...
         SENTRY_URL,
         default='',
     )
-    if not sentry_url:
+
+    if not dsn_python:
+        dsn_python = sentry_mconfig.dsn_python
+        sample_rate = sentry_mconfig.sample_rate
+        return dsn_python, sample_rate
+    else:
```
themarwhal: do we need this else if we're returning in the if condition above?

orc8r/gateway/python/magma/common/sentry.py
```diff
@@ -83,30 +85,56 @@ def get_sentry_status(service_name: str) -> SentryStatus:
...
+    """Initialize connection and start piping errors to sentry.io."""
+
+    sentry_status = get_sentry_status(service_name)
+    if sentry_status == SentryStatus.DISABLED:
+        return
+
+    dsn_python, sample_rate = get_sentry_dsn_and_sample_rate(sentry_mconfig)
+
+    if not dsn_python:
+        logging.info('Sentry disabled because of missing dsn_python.')
```
themarwhal: Can we add some more detail on where this value can be placed? (So we can say what's described in the function doc above somewhere. control_proxy config for per-agw override and mconfig.shared_config.sentry_config for network wide configs)

themarwhal: Hmm actually, maybe it is too noisy if we print that for every service

orc8r/gateway/python/magma/configuration/mconfig_managers.py
```diff
@@ -172,6 +192,21 @@ def load_service_mconfig(
         service_mconfig = mconfig.configs_by_key[service_name]
         return unpack_mconfig_any(service_mconfig, mconfig_struct)
 
+    def load_shared_mconfig(
```
themarwhal, line 195: Can we suppress this pep8 warning and say the docs are in the abstract method

orc8r/gateway/python/magma/common/sentry.py
alexzurbonsen: True, it might be... Do you think printing the single line, as we currently do, might be too noisy as well? (On the other hand it may be helpful for users, who wonder why sentry isn't working, even though they enabled it in the service config.)

orc8r/gateway/python/magma/configuration/mconfig_managers.py
alexzurbonsen: Disabled the missing-docstrings warning for the entire class since all methods are already documented in base class. Would that be a reasonable alternative?

orc8r/gateway/python/magma/common/sentry.py
themarwhal: Yeah since we already have the single line message and that doesn't seem to bother us, let's just make it a bit more descriptive then :) You make a good point about there being an additional config about enabling / disabling sentry so it is confusing.

themarwhal (approved):

orc8r/gateway/python/magma/common/sentry.py
```diff
@@ -83,30 +85,55 @@ def get_sentry_status(service_name: str) -> SentryStatus:
...
+
+    sentry_status = get_sentry_status(service_name)
+    if sentry_status == SentryStatus.DISABLED:
+        return
+
+    dsn_python, sample_rate = get_sentry_dsn_and_sample_rate(sentry_mconfig)
+
+    if not dsn_python:
+        logging.info('Sentry disabled because of missing dsn_python. '
+                     'See documentation (Configure > AGW) on how to configure '
```
themarwhal: 👍

alexzurbonsen: Pushed one more time to make format check happy (and also markdown lint check seems to be broken on master. ran make precommit_fix in docs and included changes here - only `contribute_id_mappings.md` is affected).

ardzoht (commented):
`lte/gateway` changes lgtm

hcgatewood:
cc @christinewang5

hcgatewood (commented):
Looking good! I just skimmed, but didn't see anywhere that we were handling restarting a service based on its shared mconfig changing

I'm assuming we want to handle that somewhere -- or maybe I missed it? cc @themarwhal

themarwhal:
> Looking good! I just skimmed, but didn't see anywhere that we were handling restarting a service based on its shared mconfig changing
> 
> I'm assuming we want to handle that somewhere -- or maybe I missed it? cc @themarwhal

Yes no service restart on shared config update for now. It's not that crucial for Sentry config at this current stage of testing, but we will get it in eventually :) I think I filed an issue for this, will dig that up or create a new one and link it here for future context.

EDIT: Also wanted to separate out the restart logic from just reading the config as the restart logic might require a bit more thought.

cc @hcgatewood 

cc @alexzurbonsen Do you mind adding a TODO item inside the mconfig handling logic to make it clear that it's pending work?

alexzurbonsen:
> cc @alexzurbonsen Do you mind adding a TODO item inside the mconfig handling logic to make it clear that it's pending work?

@themarwhal Of course, added a TODO in process_update of config_manager.py

themarwhal:
I think this PR only needs a review from either @pshelar or @koolzz

koolzz (approved):

themarwhal (approved):

ardzoht (approved):

themarwhal:
Hey @alexzurbonsen ! This PR is good to go now, let me know if you want me to merge it. (Or label `ready2merge`)

alexzurbonsen:
> Hey @alexzurbonsen ! This PR is good to go now, let me know if you want me to merge it. (Or label `ready2merge`)

Great. Yes, @themarwhal feel free to merge it. Thanks everyone :)

PR #10558, opened 2021-11-29
feat(mme): Use control plane TEID coming from feg_relay

### LINKED ISSUE #9624: [MME] modify mme to work with centralized TEID on inbound roaming
Currently for inbound roaming, MME is the originator of TEIDs. MME is the one in charge to generate teids and keep them unique within AGW instance.

Since we are moving to a centralized TEID model, MME will not do this task anymore. Instead the Create Session Request and Create Bearer Request will not be sending the TEID value (neither for control or data plane). Instead, FEG relay will inject that value (coming from orc8r) during the request.

For that reason we need to modify current MME and make sure
- No teid for inbound roaming is generated
- Since sessions are stored using control plane TEID, we will need to make sure MME is able to us the received tied to store the session.\

We will first migrate Control Plane. So MME will not know control plane teid until it receives `CreateSessionResponse`. The field where this value will be is `c_agw_teid`

rsarwad:
fix (agw): Rsarwad fix for 9624
## Summary

Modified code such that sgw_s8 module, on reception of Create Session Request, sgw_s8 shall allocate new temporary session id and create hashlist of pdn sessions with key as temporary_session_id, FefGW shall send the control plane teid on s8 interface. On reception of Create Session Response message, sgw_s8 shall fetch the pdn session and update the teid hash list and remove the pdn session from temporary session id hash list.

## Test Plan
Executed s1ap sanity test suite and verifed the teid updating in Tera VM system


lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
```diff
@@ -15,59 +15,47 @@
...
       RETURNok);
-  EXPECT_EQ(
-      hashtable_ts_get(
-          sgw_state->imsi_ue_context_htbl, (const hash_key_t) imsi64,
-          reinterpret_cast<void**>(&ue_context_p)),
-      HASH_TABLE_OK);
+  sgw_state_exit();
+  std::this_thread::sleep_for(
+      std::chrono::milliseconds(END_OF_TESTCASE_SLEEP_MS));
+}
```
uri200, line 39: sleeps are never good. Why do you need it here?

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
```diff
@@ -137,53 +127,180 @@ TEST_F(SgwS8Config, update_pdn_session_on_cs_rsp) {
...
+      &ds_req, csresp.context_teid + 1, csresp.eps_bearer_id);
+
+  EXPECT_CALL(*mme_app_handler, mme_app_handle_delete_sess_rsp()).Times(1);
+  EXPECT_EQ(
+      sgw_s8_handle_s11_delete_session_request(sgw_state, &ds_req, imsi64),
+      RETURNerror);
+
+  sgw_state_exit();
+  std::this_thread::sleep_for(
+      std::chrono::milliseconds(END_OF_TESTCASE_SLEEP_MS));
```
uri200: nice extra cases

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
```diff
@@ -137,53 +127,180 @@ TEST_F(SgwS8Config, update_pdn_session_on_cs_rsp) {
...
+  fill_delete_session_response(&ds_rsp, csresp.context_teid, REQUEST_ACCEPTED);
+  EXPECT_CALL(*mme_app_handler, mme_app_handle_delete_sess_rsp()).Times(1);
+  EXPECT_EQ(
+      sgw_s8_handle_delete_session_response(sgw_state, &ds_rsp, imsi64),
+      RETURNok);
+  EXPECT_TRUE((sgw_get_sgw_eps_bearer_context(csresp.context_teid)) == nullptr);
+  sgw_state_exit();
+  std::this_thread::sleep_for(
+      std::chrono::milliseconds(END_OF_TESTCASE_SLEEP_MS));
+}
```
uri200, line 295: can you add some comments in different blocks of the tests to easily understand what is going on.

In the future, when you will have to modify the test and you have forgotten about it,  having those comments will be very very useful.

lte/gateway/c/core/oai/include/s8_messages_types.h
```diff
@@ -42,6 +42,13 @@ typedef struct s8_create_session_response_s {
   fteid_t pgw_s8_cp_teid;
   uint32_t cause;
   protocol_configuration_options_t pco;
+  uint32_t temporary_session_id;  // temporary_session_id is created per pdn
+                                  // session while sending create session
+                                  // request and is temporarily used to fetch
+                                  // session, once sgw_s8_teid is allocated by
+                                  // orc8r, temporary session id is removed and
+                                  // shall update imsi_ue_context_htbl based on
+                                  // received sgw_s8_teid
```
uri200: just curious, but is this field related with a session? Meaning, since we replaced teid. wouldn't it be better to use something like `temporary_teid` or `temporary_create_session_request_identifier`

Just saying because session_id reminds me to the id provided by sessiond

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
rsarwad: Adding some time to clear the contexts before exiting from test case;  Currently it's 500 ms,

uri200: is this needed in other test?
is there a better way to do it, explicitly clear the context? or wait for context cleared? 
These things may break on CI where the runners are slow

rsarwad: During execution of "make test_oai" rarely I saw some memory leaks inconsistently in spite of clearing the memory. Then I looked at the other modules; where sleep was introduced at the end of each test case, So I thought of adding sleep for all test cases. 
Reduced the duration to 250 ms and it's working fine

lte/gateway/c/core/oai/lib/s8_proxy/s8_client_api.cpp
```diff
@@ -172,31 +172,37 @@ static void recv_s8_create_session_response(
...
+  OAILOG_DEBUG_UE(
+      LOG_SGW_S8, imsi64, "Sending create session response to sgw_s8 task");
   if ((send_msg_to_task(&grpc_service_task_zmq_ctx, TASK_SGW_S8, message_p)) !=
       RETURNok) {
     OAILOG_ERROR_UE(
         LOG_SGW_S8, imsi64,
         "Failed to send S8 CREATE SESSION RESPONSE message to sgw_s8 task "
-        "for context_teid " TEID_FMT "\n",
-        context_teid);
+        "for temporary_create_session_procedure_id:%u sgw_s8_cp_teid " TEID_FMT
```
ulaskozat: remove ":" for consistency. no need for "\n".

lte/gateway/c/core/oai/lib/s8_proxy/s8_client_api.cpp
```diff
@@ -172,31 +172,37 @@ static void recv_s8_create_session_response(
...
   } else {
-    OAILOG_ERROR(
-        LOG_SGW_S8,
+    OAILOG_ERROR_UE(
+        LOG_SGW_S8, imsi64,
         "Received gRPC error for create session response for "
-        "context_teid " TEID_FMT "\n",
-        context_teid);
+        "temporary_create_session_procedure_id:%u sgw_s8_cp_teid " TEID_FMT
+        "\n",
```
ulaskozat: no need for "\n"

lte/gateway/c/core/oai/lib/s8_proxy/s8_client_api.cpp
```diff
@@ -172,31 +172,37 @@ static void recv_s8_create_session_response(
     OAILOG_ERROR_UE(
         LOG_SGW_S8, imsi64,
         "Failed to allocate memory for S8_CREATE_SESSION_RSP for "
-        "context_teid" TEID_FMT "\n",
-        context_teid);
+        "temporary_create_session_procedure_id:%u sgw_s8_cp_teid " TEID_FMT
+        "\n",
```
ulaskozat: no need for "\n" as the macro itself takes care of it.

ulaskozat:
what is the expected behavior for stateless upon service restarts?

ulaskozat:
which unit test does check for successful deletion of the temporary state?

rsarwad:
> what is the expected behavior for stateless upon service restarts?

After processing create session request message if service restarts, sgw_s8 will not be able to fetch the contexts based on temporary session_id, because sgw_s8 module rely on gRPC's response callback for matching the create session request and create session response messages. 
Since the session is in transition state, sgw_s8 will not recover the intermediate state. If service restarts, we rely on UE to re-transmit the message, pdn connectivity request message. In order to handle the re-transmitted message MME code needs to be changed. I shall take up this activity in separate PR. Currently MME code drops the re-transmitted message

lte/gateway/c/core/oai/lib/s8_proxy/s8_client_api.cpp
rsarwad: Thank you for letting me know. Earlier we explicitly needed to have "\n".

rsarwad:
> which unit test does check for successful deletion of the temporary state?

I added this check for test case, https://github.com/magma/magma/pull/10558/files#diff-416c5a669e5c52593224fe3654fbe3efd6485bcde5f9655b3fbb854bfc919ab1R139

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_csr.cpp
ulaskozat: These sleeps are removed in `test_mme_procedures.cpp`. What exactly are you seeing as the failure?

rsarwad: After test case is completed, it showed up memory leak. UE context and sgw_eps_bearer context are not released.

ulaskozat: check out the recently merged PR https://github.com/magma/magma/pull/10721/files. It probably is due to TearDown implementation.

ulaskozat: Let me know if you do the same changes whether the issue still persists or not.

ulaskozat (approved):
LGTM

PR #10562, opened 2021-11-29
chore(agw): Send orc8r cloud address as tag to Sentry

sebathomas:
## Summary

Obtain cloud address from control_proxy.yml and include it as a
tag in Sentry issues. This is helpful to find out which systems
are affected by an issue.

## Test Plan

Used Sentry test instance to send an error.
sentry-tag

### DIFF
--- a/orc8r/gateway/python/magma/common/sentry.py
+++ b/orc8r/gateway/python/magma/common/sentry.py
@@ -24,6 +24,8 @@ CONTROL_PROXY = 'control_proxy'
 SENTRY_CONFIG = 'sentry'
 SENTRY_URL = 'sentry_url_python'
 SENTRY_SAMPLE_RATE = 'sentry_sample_rate'
+CLOUD_ADDRESS = 'cloud_address'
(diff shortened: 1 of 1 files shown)


orc8r/gateway/python/magma/common/sentry.py
```diff
@@ -107,5 +109,12 @@ def sentry_init(service_name: str):
         traces_sample_rate=sentry_sample_rate,
         before_send=_ignore_if_not_marked if sentry_status == SentryStatus.SEND_SELECTED_ERRORS else None,
     )
+
+    cloud_address = get_service_config_value(
```
themarwhal, line 113: Can we also add this here? https://github.com/magma/magma/blob/master/orc8r/gateway/c/common/sentry/SentryWrapper.cpp

themarwhal (approved):
looks great! :)

hcgatewood (approved):
💯💯

orc8r/gateway/python/magma/common/sentry.py
sebathomas: Done in #10580.

PR #10579, opened 2021-11-30
fix(orc8r): Incorrect conversion between integer types

spikey979:
fix(orc8): fix incorrect conversion between integer types

## Summary

- magma/orc8r/cloud/go/services/state/indexer/indexer.go:71

Check for lower and upper bounds for uint32 type added.

Fix for: https://github.com/magma/magma/security/code-scanning/19?query=ref%3Arefs%2Fheads%2Fmaster

## Test Plan

I ran test:  
/magma/orc8r/cloud/docker/  ./build.py -c

### DIFF
--- a/orc8r/cloud/go/services/state/indexer/indexer.go
+++ b/orc8r/cloud/go/services/state/indexer/indexer.go
@@ -69,7 +69,7 @@ type Version uint32
 // the required integer size.
 func NewIndexerVersion(version int64) (Version, error) {
 	if version < 0 || version > 4294967295 {
-		return 0, fmt.Errorf("version %d is outside the boundaries of unit32 type", version)
+		return 0, fmt.Errorf("version value %d is outside the boundaries of unit32 type", version)
 	}
 	capped := Version(version)
 	if int64(capped) != version {
--- a/orc8r/cloud/go/services/state/indexer/indexer.go
+++ b/orc8r/cloud/go/services/state/indexer/indexer.go
@@ -68,8 +68,8 @@ type Version uint32
 // NewIndexerVersion returns a new indexer version, ensuring it fits into
 // the required integer size.
 func NewIndexerVersion(version int64) (Version, error) {
-	if version < 0 || version > 4294967295 {
-		return 0, fmt.Errorf("version value %d is outside the boundaries of unit32 type", version)
+	if version < 0 || version > math.MaxUint32 {
+		return 0, fmt.Errorf("version value %d is outside the boundaries of uint32 type", version)
 	}
 	capped := Version(version)
 	if int64(capped) != version {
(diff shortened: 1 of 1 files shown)


orc8r/cloud/go/services/state/indexer/indexer.go
```diff
@@ -68,6 +68,9 @@ type Version uint32
 // NewIndexerVersion returns a new indexer version, ensuring it fits into
 // the required integer size.
 func NewIndexerVersion(version int64) (Version, error) {
+	if version < 0 || version > 4294967295 {
+		return 0, fmt.Errorf("version value %d is outside the boundaries of unit32 type", version)
+	}
```
sebathomas: Might be good to use the constant for this type of check.
```suggestion
	if version < 0 || version > math.MaxUint32 {
		return 0, fmt.Errorf("version value %d is outside the boundaries of uint32 type", version)
	}
```

spikey979:
@sebathomas thanks for advice....

andreilee (approved):

orc8r/cloud/go/services/state/indexer/indexer.go
```diff
@@ -68,6 +69,9 @@ type Version uint32
 // NewIndexerVersion returns a new indexer version, ensuring it fits into
 // the required integer size.
 func NewIndexerVersion(version int64) (Version, error) {
+	if version < 0 || version > math.MaxUint32 {
+		return 0, fmt.Errorf("version value %d is outside the boundaries of uint32 type", version)
+	}
 	capped := Version(version)
 	if int64(capped) != version {
 		return 0, fmt.Errorf("indexer version %v too large for %T", version, Version(0))
```
hcgatewood: This check is doing the same thing as the check below it. I'm fine with the "new" way if it silences an error, but then let's remove the old way

hcgatewood (changes requested):

orc8r/cloud/go/services/state/indexer/indexer.go
spikey979: I removed the old way...

orc8r/cloud/go/services/state/indexer/indexer.go
```diff
@@ -68,10 +69,11 @@ type Version uint32
 // NewIndexerVersion returns a new indexer version, ensuring it fits into
 // the required integer size.
 func NewIndexerVersion(version int64) (Version, error) {
-	capped := Version(version)
-	if int64(capped) != version {
-		return 0, fmt.Errorf("indexer version %v too large for %T", version, Version(0))
+	if version < 0 || version > math.MaxUint32 {
+		return 0, fmt.Errorf("version value %d is outside the boundaries of uint32 type", version)
 	}
+	capped := Version(version)
```
hcgatewood, line 75: Looks like the code scan is still flagging this -- any thoughts on how to appease the scanner?


Screen Shot 2021-12-13 at 4 44 45 PM

PR #10583, opened 2021-11-30
feat: Add Bazel configurations to build with LSAN

themarwhal:

## Summary
Blocked on https://github.com/magma/magma/pull/10587

This PR adds Bazel configurations needed to enable lsan. This is not enabled by default for now, as implicit configurations in bazelrc can be hard to catch. I will wrap the config in a make command in a later diff.
I also moved around some of the bazelrc configs to clean it up.

## Test Plan
Run bazel test //... --config=lsan with https://github.com/magma/magma/pull/10587 cherry-picked 

## Additional Information


.bazelrc
```diff
@@ -1,47 +1,57 @@
...
-build:asan --strip=never
+# ASAN
 build:asan --copt=-fsanitize=address
 build:asan --copt=-O0
 build:asan --copt=-fno-omit-frame-pointer
 build:asan --linkopt=-fsanitize=address
 build:asan --action_env=ASAN_OPTIONS=detect_leaks=1:color=always
+build:asan --test_tag_filters=-no_asan
+
+# LSAN (Enabled by default)
```
electronjoe: I think it's worth discussing whether this is enabled by default or not.

In my prior life, we separated the various sanitizers and required they be specified with `--config` clauses. I did not at the time stop to ask why, but here are some ideas (not all of them strong):

- Most sanitizers cannot be run simultaneously (except LSAN, ASAN?), so having one default to one is a little misleading IMO
- I wonder (but don't know) if test runtime behavior is sometimes modified by `LSAN` presence (at a minimum timings? memory use...)
- Failure of a CI job that clearly specifies it is an `ASAN unit test run` provides improved signal at-a-glance as to root cause (e.g. you note that non-ASAN is fine, ASAN failing -> better signal on problem at-a-glance)
- Allows test authors to begin with functional tests and get them passing in their PR, before raising the bar to also passing ASAN/etc (at least gives that option)

Upshot: I would advocate (but welcome discussion) that we do not default to ASAN-on / LSAN on.

Downsides to not defaulting on ASAN (that I can think of):

- Longer build / test cycles (have to build multiple unit test variants and run them)
- Contributor has to know there are `--config=` options that need to be tested aside from vanilla `bazel test //to/target` (but they will learn thanks to CI...)

themarwhal: yeah that is a good point. I was mostly defaulting to the current behavior, which is ASAN for debug builds, LSAN for production builds. I definitely think we don't want to enable ASAN by default, but rather have a CI check that runs all tests with ASAN.

electronjoe: Whoops - I misread the PR (actually, perhaps minimally read the actual diffs) - sorry about that. I saw preserve existing behavior and knew that debug builds in CMake use ASAN and thought: would prefer not!

themarwhal: For LSAN, I do not know too much about the side effects yet to see if keeping them enabled by default is a good / bad idea.

electronjoe: Aside but we also tended to run all unit tests in release mode as well, as part of CI. So you end up with debug + release mode, and in DEBUG modes you cross product with {ASAN, MSAN, TSAN, UBSAN, and on}. LOTS of tests =D About 6 test builds and executions per unit test, with differing compiler / linking behavior on each.

themarwhal: Yeah... and changing these configs trigger a re-build in all the external dependencies as well. Which can be slow :/ I'll do a quick look around in sourcegraph to see how people test these configs. :D

electronjoe: > For LSAN, I do not know too much about the side effects yet to see if keeping them enabled by default is a good / bad idea.

Me neither, but I generally hesitate as to whether this is best practice for production builds (or again, whether bundling with default build behavior is the right call).  But if we don't people may forget the `--config=lsan` and produce production binaries not what they expect (though a human should not be running these commands...)

themarwhal: yeah, definitely agreed about limiting implicit default configs. Maybe I will keep lsan not enabled by default, but wrap it inside like a `make build_prod` Make command just to keep them explicit.

themarwhal: modified PR to not enable it by default.

themarwhal: I am starting to wonder if having a project wide lsan/asan configuration is not the way to go as it affects external dependency builds? Maybe we can create a custom rule that handles ASAN/LSAN configurations for Magma C/C++ targets...

envoyproxy does something like this, but they are configuring a lot more than what I'm proposing.https://sourcegraph.com/github.com/envoyproxy/envoy/-/blob/bazel/envoy_library.bzl?L84

electronjoe: > I am starting to wonder if having a project wide lsan/asan configuration is not the way to go as it affects external dependency builds?

Interesting. Does seem like we are less likely to need lsan/asan coverage for third party build targets being statically linked... I'm personally unaware if there are any gotchas in not having total lsan/asan coverage for all statically linked assets in a binary? Would want to know that.  E.g. not havinbg lsan/asan on for Folly - does that in any way hinder lsan/asan behavior at the use of folly lib (can it potentially mask issues in our code).

themarwhal: Yeah I was wondering if this would mask issues as well. Will google around. I was also wondering if we could use the custom rule concept to have different levels of compiler warnings for external libraries / internal. (it's just a little noisy for some libraries right now) But not clear on whether that is good practice or not.

electronjoe (approved):

PR #10585, opened 2021-11-30
fix(magmad): Skip upgrade if no sensible target version is set

### LINKED ISSUE #10418: Frequent Sentry Error: git checkout in magmad upgrade loop fails for tag 0.0.0-0
Sentry Issue: [LAB-AGWS-PYTHON-29](https://sentry.io/organizations/lf-9c/issues/2771068653/?referrer=github_integration)

```
Upgrade loop failed! Error: Exception('Command failed: git -C /tmp/magma_upgrade/magma checkout 0.0.0-0')
```

sebathomas:
The DockerUpgrader will fail for the default version 0.0.0-0 (see #10418) while the MagmaUpgrader will already skip the update because 0.0.0-0 is the smallest possible version.

Fixes #10418

### DIFF
--- a/orc8r/gateway/python/magma/magmad/upgrade/upgrader.py
+++ b/orc8r/gateway/python/magma/magmad/upgrade/upgrader.py
@@ -13,6 +13,8 @@ limitations under the License.
 import asyncio
 import logging
 
+DEFAULT_VERSION = '0.0.0-0'
+
 
 class Upgrader(object):
     """
@@ -81,26 +83,26 @@ def start_upgrade_loop(magmad_service, upgrader):
 
     while True:
         logging.info('Checking for upgrade...')
-        try:
-            target_ver = _get_target_version(magmad_service.mconfig)
-            upgrader.perform_upgrade_if_necessary(target_ver)
-        except Exception:  # pylint: disable=broad-except
-            logging.exception(
-                'Error encountered while upgrading, will try again after delay',
-            )
+        target_ver = _get_target_version(magmad_service.mconfig)
         poll_interval = max(  # No faster than 1/minute
             60,
             magmad_service.mconfig.autoupgrade_poll_interval,
         )
+        if target_ver == DEFAULT_VERSION:
+            logging.warning(
+                'magmad package_version config missing or set to '
+                'default %s, skipping upgrade', DEFAULT_VERSION,
+            )
+        else:
+            try:
+                upgrader.perform_upgrade_if_necessary(target_ver)
+            except Exception:  # pylint: disable=broad-except
+                logging.exception(
+                    'Error encountered while upgrading, '
+                    'will try again after %s seconds', poll_interval,
+                )
         yield from asyncio.sleep(poll_interval)
 
 
 def _get_target_version(magmad_mconfig):
-    if magmad_mconfig.package_version is None:
-        logging.warning(
-            'magmad package_version config not found, '
-            'returning 0.0.0-0 as target package version.',
-        )
-        return '0.0.0-0'
-
-    return magmad_mconfig.package_version
+    return magmad_mconfig.package_version or DEFAULT_VERSION


orc8r/gateway/python/magma/magmad/upgrade/upgrader.py
```diff
@@ -81,12 +83,19 @@ def start_upgrade_loop(magmad_service, upgrader):
...
-                'Error encountered while upgrading, will try again after delay',
+        target_ver = _get_target_version(magmad_service.mconfig)
+        if target_ver != DEFAULT_VERSION:
+            try:
+                upgrader.perform_upgrade_if_necessary(target_ver)
+            except Exception:  # pylint: disable=broad-except
+                logging.exception(
+                    'Error encountered while upgrading, '
+                    'will try again after delay',
+                )
```
Neudrino: * If there is a retry here somewhere, would it be sensible to have a lower log level?
* Could the delay be specified more precisely here?
``` python
SLEEP_PERIOD = 120
...
yield from asyncio.sleep(SLEEP_PERIOD)
...
'will try again after %s', SLEEP_PERIOD
```
* Any reason for splitting the message in multiple lines?

orc8r/gateway/python/magma/magmad/upgrade/upgrader.py
```diff
@@ -13,6 +13,8 @@
 import asyncio
 import logging
 
+DEFAULT_VERSION = '0.0.0-0'
```
Neudrino, line 16: Any reason, we set a dummy version and not `None`?

themarwhal: Good Q, can we figure out where this value is coming from?

sebathomas: It's the default in the mconfig and I thought it should be treated the same way as a missing value.

(The MagmaUpgrader has a version check that will always skip if that value is given, I guess that's why it was introduced.)

orc8r/gateway/python/magma/magmad/upgrade/upgrader.py
sebathomas: * It catches any exception so I assume it's an unexpected error and should have a high log level even if the retry happens.
* The sleep period is not a constant but we can put it into the log message. :+1:
* It's split to avoid having a very long line.

orc8r/gateway/python/magma/magmad/upgrade/upgrader.py
themarwhal: Yeah that makes sense to me! Is the 0 value harcoded somewhere on the Orc8r side? Or is it a default value that protobuf assigns when nothing is specified?

orc8r/gateway/python/magma/magmad/upgrade/upgrader.py
```diff
@@ -81,26 +83,26 @@ def start_upgrade_loop(magmad_service, upgrader):
...
-        except Exception:  # pylint: disable=broad-except
-            logging.exception(
-                'Error encountered while upgrading, will try again after delay',
-            )
+        target_ver = _get_target_version(magmad_service.mconfig)
         poll_interval = max(  # No faster than 1/minute
             60,
             magmad_service.mconfig.autoupgrade_poll_interval,
         )
+        if target_ver != DEFAULT_VERSION:
```
themarwhal: This is mostly my preference but I think it might be clearly if it's like
```python
if target_ver == DEFAUL...:
  # log warning
else: 
  # do work
```

themarwhal (approved):

orc8r/gateway/python/magma/magmad/upgrade/upgrader.py
sebathomas: I just see the 0 value in the default gateway.mconfig files. I don't think it's hardcoded anywhere else.

The skip in MagmaUpgrader works by parsing the version and only upgrading if it's larger than the current version.

themarwhal:
Also, unrelated but we should file an issue to rename the upgrader2 file to something a bit more descriptive :p

sebathomas:
> Also, unrelated but we should file an issue to rename the upgrader2 file to something a bit more descriptive :p

Yes, that's not a very intuitive name. ;)
DockerUpgrader is the only subclass of the abstract Upgrader2. Maybe this could be refactored to get rid of Upgrader2 entirely?

PR #10613, opened 2021-12-01
chore(agw): enable IPv6 for UE and S1 networking

pshelar:

## Additional Information

### DIFF
--- a/lte/gateway/configs/mme.yml
+++ b/lte/gateway/configs/mme.yml
@@ -35,4 +35,4 @@ mme_app_zmq_smc_th_us: 1000000  # delay threshold used for dropping SMC Complete
 accept_combined_attach_tau_wo_csfb: true
 
 # Enable IPv6 support for S1AP SCTP endpoint
-s1ap_ipv6_enabled: false
+s1ap_ipv6_enabled: true
--- a/lte/gateway/configs/spgw.yml
+++ b/lte/gateway/configs/spgw.yml
@@ -57,7 +57,7 @@ agw_l3_tunnel: false
 ovs_gtpu_checksum: false
 
 # Enable IPv6 support for S1AP SCTP endpoint
-s1_ipv6_enabled: false
+s1_ipv6_enabled: true
 
 # Enabled eBPF map operations
 ebpf_enabled: false
--- a/lte/gateway/python/scripts/generate_oai_config.py
+++ b/lte/gateway/python/scripts/generate_oai_config.py
@@ -451,16 +451,23 @@ def _get_context():
         iface_name = get_service_config_value(
             'spgw', 'sgw_s5s8_up_iface_name_non_nat', '',
         )
+    s1ap_ipv6_addr = _get_iface_ipv6("mme", "s1ap_iface_name")
+    if s1ap_ipv6_addr and not s1ap_ipv6_addr.startswith('fe80'):
+        s1ap_ipv6_enabled = get_service_config_value(
+            "mme", "s1ap_ipv6_enabled", default=False,
+        )
+    else:
+        s1ap_ipv6_addr = '::'
+        s1ap_ipv6_enabled = False
+
     context = {
         "mme_s11_ip": _get_iface_ip("mme", "s11_iface_name"),
         "sgw_s11_ip": _get_iface_ip("spgw", "s11_iface_name"),
         'sgw_s5s8_up_iface_name': iface_name,
         "remote_sgw_ip": get_service_config_value("mme", "remote_sgw_ip", ""),
         "s1ap_ip": _get_iface_ip("mme", "s1ap_iface_name"),
-        "s1ap_ipv6": _get_iface_ipv6("mme", "s1ap_iface_name"),
-        "s1ap_ipv6_enabled": get_service_config_value(
-            "mme", "s1ap_ipv6_enabled", default=False,
-        ),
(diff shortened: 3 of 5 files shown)


lte/gateway/configs/spgw.yml
```diff
@@ -57,4 +57,4 @@ agw_l3_tunnel: false
 ovs_gtpu_checksum: false
 
 # Enable IPv6 support for S1AP SCTP endpoint
-s1_ipv6_enabled: false
+s1_ipv6_enabled: true
```
ulaskozat, line 60: this looks like not used anywhere (s1ap_ipv6_enabled) in mme.yml looks like the universal config. You can remove this completely I think.

lte/gateway/configs/mme.yml
```diff
@@ -35,4 +35,4 @@ mme_app_zmq_smc_th_us: 1000000  # delay threshold used for dropping SMC Complete
 accept_combined_attach_tau_wo_csfb: true
 
 # Enable IPv6 support for S1AP SCTP endpoint
-s1ap_ipv6_enabled: false
+s1ap_ipv6_enabled: true
```
ulaskozat, line 38: can we instead use `cp_ipv6_enabled` for control plane and be RAT agnostic?

ulaskozat (changes requested):
see my comments.

ulaskozat:
@pshelar, @koolzz, @pruthvihebbani, I am a bit puzzled here. Is this flag used for enabling IPv6 in data path for GTP tunnels or used for addressing SCTP end point?

lte/gateway/configs/spgw.yml
ulaskozat: or is the intent to enable/disable CP and UP IPv6 paths separately?

pshelar:
> @pshelar, @koolzz, @pruthvihebbani, I am a bit puzzled here. Is this flag used for enabling IPv6 in data path for GTP tunnels or used for addressing SCTP end point?

I want to enable both, what is downside of this?

lte/gateway/configs/spgw.yml
pshelar: right. this allows separate control over CP and UP over IPv6.

lte/gateway/configs/mme.yml
pshelar: I am fine with the rename, but that would be different change.

ulaskozat:
> 

The comments and the names for the flags do not reflect that. As long as the intent is clear, it is fine.

lte/gateway/configs/spgw.yml
ulaskozat: but the current code generation does not use this at all. check the template files.

pshelar: it is there. checkout MME_S1_IPV6_ENABLED and SGW_S1_IPV6_ENABLED.

ulaskozat: please check what is used in spgw.template.conf and mme.template.conf and spgw.template.conf:
`MME_S1_IPV6_ENABLED                   = "{{ s1ap_ipv6_enabled }}";`
`SGW_S1_IPV6_ENABLED = "{{ s1ap_ipv6_enabled }}";`

ulaskozat: they both use the same flag.

ulaskozat: Note that I am not questioning why we enable both at the same time (as I think we should have a single flag). I am simply suggesting that we remove s1_ipv6_enabled as
1. It is not really consumed in conf generation
2. We do not have a use case where CP does not support IPv6 while UP is supporting it (it is at the end of the day the same interface).

pshelar: Internally these are two different features, so for me it makes sense to have separate flag on AGW to protect each feature.

At user API level we can have single switch to enable IPv6, But I am fine with unifying it even at AGW level.

I think it would it would be tricky to unify these two configurations in single yml file due to existing config management in spgw and mme. @koolzz can you confirm?

pshelar: > 

typo in template for SGW_S1_IPV6_ENABLED needs to be fixed.

ekowtaylor:
@pshelar what's pending with this PR?

lte/gateway/python/scripts/generate_oai_config.py
```diff
@@ -421,16 +421,23 @@ def _get_context():
         iface_name = get_service_config_value(
             'spgw', 'sgw_s5s8_up_iface_name_non_nat', '',
         )
+    s1ap_ipv6_addr = _get_iface_ipv6("mme", "s1ap_iface_name")
+    if s1ap_ipv6_addr and s1ap_ipv6_addr.startswith('fe80') is False:
```
sebathomas: Nitpicking
```suggestion
    if s1ap_ipv6_addr and not s1ap_ipv6_addr.startswith('fe80'):
```

sebathomas (approved):
The Python code looks good.

lte/gateway/configs/spgw.yml
pshelar: fixed it

pruthvihebbani:
LGTM

pruthvihebbani (approved):

ulaskozat (approved):
LGTM

PR #10617, opened 2021-12-01
fix(ci): update firebase used sha

quentinDERORY:

## Additional Information

### DIFF
--- a/.github/workflows/cwf-integ-test.yml
+++ b/.github/workflows/cwf-integ-test.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          ref: ${{ env.sha }}
+          ref: ${{ env.SHA }}
       - name: Run docker compose
         run: |
           cd cwf/gateway/docker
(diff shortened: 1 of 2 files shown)


.github/workflows/cwf-integ-test.yml
```diff
@@ -12,14 +12,16 @@ on:  # yamllint disable-line rule:truthy
...
+  SHA: ${{ github.event.workflow_run.head_commit.id || github.sha }}
 
 jobs:
   docker-build:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
         with:
-          ref: ${{ github.event.workflow_run.head_commit.id || github.sha }}
+          ref: ${{ env.sha }}
```
talkhasib: is env context case sensitive? why not use env.SHA just like we do everywhere else?

.github/workflows/cwf-integ-test.yml
```diff
@@ -140,4 +142,4 @@ jobs:
           SLACK_AVATAR: ":boom:"
         uses: Ilshidur/action-slack@2.1.0
         with:
-          args: 'CWF integration test: tests failed on [${{github.sha}}](${{github.event.repository.owner.html_url}}/magma/commits/${{github.sha}}): ${{ steps.commit.outputs.title}}'
+          args: 'CWF integration test: tests failed on [${{ env.sha }}](${{github.event.repository.owner.html_url}}/magma/commits/${{ env.sha }}: ${{ steps.commit.outputs.title}}'
```
talkhasib: you lost a ")" here

talkhasib (changes requested):

.github/workflows/cwf-integ-test.yml
quentinDERORY: It's not case sensitive but better use CAPS everywhere

talkhasib (approved):

PR #10621, opened 2021-12-02
chore: add utility script to dev_tools/

themarwhal:

## Summary
1. create a dev_tools directory (renamed from tools/)
2. add a minidump-2-core script that is useful for analyzing minidumps uploaded to Sentry.io. @ssanadhya and I discussed whether to bake it into our dev VM, but it is more useful if the script is available for all dev env.

## How-to-use
First, download the minidump file uploaded to Sentry and the corresponding artifacts.

The debug artifacts (essentially the service binary but stripped) are available here
Screen Shot 2021-12-02 at 10 32 02 AM

The minidump should be available as an attachment
Screen Shot 2021-12-02 at 10 33 06 AM

Then use the `minidump-2-core` tool to convert the dump into a core file.
```bash
# see the script for all options
./dev_tools/minidump-2-core 7db7a189-6faf-4b1a-77e05fb0-f142bfb5.dmp -o core
gdb -c core 97c782eb-283f-5f11-75e1-8602ae9927ae.debug
```

Note that it seems to be missing some information but not all. 

## Test Plan
no functional change

## Additional Information


ardzoht (approved):
lgtm, maybe worth to include some usage notes in `docs/readmes/lte/dev_notes.md`?

ssanadhya (approved):
❤️

PR #10629, opened 2021-12-02
fix(subscriberdb): handle 'check subscribers' errors with known solution separately

alexzurbonsen:
## Summary

Translates discussion in #10387 into a proposal: differentiate grpc error with known solution from other grpc errors.

## Test Plan

Tested the introduced logging function `_log_grpc_error` locally.

For StatusCode INTERNAL the following message is logged:

`ERROR:root:Internal Error in subscriberdb-Orc8r communication. If AGW and Orc8r are deployed at a commit hash, upgrade AGW and Orc8r to latest commit hash.`

For StatusCode UNAVAILABLE (or all other StatusCodes != INTERNAL) the following message would be logged:

`ERROR:root:_sync_subscribers request error! [StatusCode.UNAVAILABLE] Received http2 header with status: 502
`

(The first part of the message depends on the caller. An error message starting with `_check_subscribers_in_sync request error! ...` is also possible.)

Without the proposed changes the error message always is `Check subscribers in sync request error! ...` or `Sync subscribers request error! ...`

### DIFF
--- a/lte/gateway/python/magma/subscriberdb/client.py
+++ b/lte/gateway/python/magma/subscriberdb/client.py
@@ -12,6 +12,7 @@ limitations under the License.
 """
 import asyncio
 import datetime
+import inspect
 import logging
 from typing import List, NamedTuple, Optional
 
@@ -136,10 +137,7 @@ class SubscriberDBCloudClient(SDWatchdogTask):
                 self._loop,
             )
         except grpc.RpcError as err:
-            logging.error(
-                "Check subscribers in sync request error! [%s] %s", err.code(),
-                err.details(),
-            )
+            _log_grpc_error(err)
             return False
         return res.in_sync
 
@@ -164,10 +162,7 @@ class SubscriberDBCloudClient(SDWatchdogTask):
                 self._loop,
             )
         except grpc.RpcError as err:
-            logging.error(
-                "Sync subscribers request error! [%s] %s", err.code(),
-                err.details(),
-            )
+            _log_grpc_error(err)
             return True
 
         if not res.resync:
@@ -374,3 +369,18 @@ class SubscriberDBCloudClient(SDWatchdogTask):
                 "Detach Deleted Subscribers Error! [%s] %s",
                 err.code(), err.details(),
             )
+
+
+def _log_grpc_error(err: grpc.RpcError) -> None:
+    if err.code() == grpc.StatusCode.INTERNAL or err.code() == grpc.StatusCode.UNIMPLEMENTED:
+        logging.error(
+            'Internal Error in subscriberdb-Orc8r communication. '
+            'If AGW and Orc8r are deployed at a commit hash, '
+            'upgrade AGW and Orc8r to latest commit hash.',
+        )
+    else:
+        logging.error(
+            # inspect.stack() indices: [1] -> frame record of penultimate caller; [3] -> function name
+            "%s request error! [%s] %s", inspect.stack(context=0)[1][3], err.code(),
+            err.details(),
+        )


lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -164,10 +171,17 @@ async def _sync_subscribers(self) -> bool:
...
+                logging.info(
+                    'Internal Error in subscriberdb-Orc8r communication. '
+                    'If AGW and Orc8r are deployed at a commit hash, '
+                    'upgrade AGW and Orc8r to latest commit hash.'
+                )
+            else:
+                logging.error(
+                    "Sync subscribers request error! [%s] %s", err.code(),
+                    err.details(),
+                )
```
sebathomas: Since it's the same logic in both cases (and almost the same messages), I think you can extract a function to handle the logging.

sebathomas (approved):

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -164,10 +161,7 @@ async def _sync_subscribers(self) -> bool:
                 self._loop,
             )
         except grpc.RpcError as err:
-            logging.error(
-                "Sync subscribers request error! [%s] %s", err.code(),
-                err.details(),
-            )
+            _log_grpc_error(err, "Sync subscribers")
```
themarwhal: If we want to pass the function name, we should use the exact same naming pattern: `_sync_subscribers` in this case. Is there a way to get this name programmatically? something like this? https://stackoverflow.com/questions/251464/how-to-get-a-function-name-as-a-string

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -374,3 +368,17 @@ def _detach_deleted_subscribers_done(self, delete_future):
                 "Detach Deleted Subscribers Error! [%s] %s",
                 err.code(), err.details(),
             )
+
+
+def _log_grpc_error(err: grpc.RpcError, function_name: str) -> None:
+    if err.code() == grpc.StatusCode.INTERNAL:
+        logging.error(
+            'Internal Error in subscriberdb-Orc8r communication. '
```
themarwhal, line 377: @hcgatewood not sure what would be an appropriate message here, any ideas?

lte/gateway/python/magma/subscriberdb/client.py
alexzurbonsen: Good idea. Changed the arguments accordingly.

themarwhal: Actually do you think we could make it so that the caller doesn't even have to pass in its name manually? I can imagine we might start seeing some bad copy pastes leading to confusion :D

alexzurbonsen: Sure :D Amended another suggestion. What do you think?

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -374,3 +369,17 @@ def _detach_deleted_subscribers_done(self, delete_future):
...
+def _log_grpc_error(err: grpc.RpcError) -> None:
+    if err.code() == grpc.StatusCode.INTERNAL:
+        logging.error(
+            'Internal Error in subscriberdb-Orc8r communication. '
+            'If AGW and Orc8r are deployed at a commit hash, '
+            'upgrade AGW and Orc8r to latest commit hash.',
+        )
+    else:
+        logging.error(
+            "%s request error! [%s] %s", inspect.stack(0)[1][3], err.code(),
```
themarwhal: inspect.stack(0)[1][3] can we annotate what the indecies mean here? (I'm guessing it get's the name out?)

lte/gateway/python/magma/subscriberdb/client.py
themarwhal: Looking good! added another comment

lte/gateway/python/magma/subscriberdb/client.py
alexzurbonsen: Yes, it is. I added a comment to explain.

hcgatewood (changes requested):
Separating out grpc-internal errs is fine, but we're also throwing away the existing error log msg -- let's keep those (or, at least, add an example err log from these changes to the PR summary)

alexzurbonsen:
> Separating out grpc-internal errs is fine, but we're also throwing away the existing error log msg -- let's keep those (or, at least, add an example err log from these changes to the PR summary)

@hcgatewood I added explanations to the Test Plan section of the PR. What do you think?

hcgatewood (approved):
Awesome thanks for the clarification, lgtm!

themarwhal (approved):

alexzurbonsen:
Amended the PR to include StatusCode UNIMPLEMENTED as an error with the same known solution as INTERNAL. This error code surfaced recently, after the scale testing agw in the lab was connected to Sentry. See #10387.

PR #10633, opened 2021-12-02
test(sessiond): add logs to should_be_unsuspended function

uri200:

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/ChargingGrant.cpp
+++ b/lte/gateway/c/session_manager/ChargingGrant.cpp
@@ -94,7 +94,8 @@ CreditValidity ChargingGrant::is_valid_credit_response(
       MLOG(MWARNING)
           << "GSU for RG: " << key << " " << session_id
           << " is invalid, but accepting it as it has a final unit action or "
-             "suspended credit ";
+             "suspended credit. Credit_validity is "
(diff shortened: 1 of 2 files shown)


lte/gateway/c/session_manager/ChargingGrant.cpp
```diff
@@ -267,10 +268,12 @@ ServiceActionType ChargingGrant::get_action(
 bool ChargingGrant::should_be_unsuspended() const {
   // transitioning out of FUA-redirect/restrict
   if (service_state == SERVICE_NEEDS_ACTIVATION) {
+      MLOG(MDEBUG) << "Credit should be unsuspended SERVICE_NEEDS_ACTIVATION";
     return true;
   }
   // transitioning out of credit suspension
   if (suspended && !credit.is_quota_exhausted(1)) {
+      MLOG(MDEBUG) << "Credit should be unsuspended because we still have quota";
     return true;
```
uri200, line 277: i think this condition is wrong and may be causing the session to unsuspend right after being suspended

crbertoldo: @uri200 please: do you suggest keeping it true or changing it to false?

PR #10634, opened 2021-12-02
feat(feg_relay): replace c_agw_teid for feg_relay generated teid

uri200:

## Summary

This PR just finalizes the change on feg_relay to inject teid. 
Note we were using a temporary field on protos which is now removed. (safe to remove because it was not being used, just used for display while logging protos)

## Test Plan
./builc.py -c at orc8r

## Additional Information


feg/protos/s8_proxy.proto
```diff
@@ -64,8 +64,6 @@ message CreateSessionRequestPgw {
     bytes indication_flag = 16;
     string charging_characteristics = 17;
     TimeZone time_zone = 18;
-
-    uint32 feg_relay_generated_c_agw_teid = 99;
```
uri200, line 68: safe to remove. We just used temporarily to make sure teid was generated properly. Now we use CAgwTeid to inject the value

themarwhal (approved):

PR #10635, opened 2021-12-02
fix(amf): PDU Session faiulre due to incorrect duplication PDU count

ganeshg87:
test cases:
   - reproduced and verified fix with UERANSIM
       triggered 5 times PDU Session create & Release
   - oai test executed

Signed-off-by: ganeshg87 <ganesh.gedela@wavelabs.ai>

## Summary
   PDU Session setup is failing after the 5th time successful PDUSession Setup and Release. 

**Fix:**
 if PDU Session is already ACTIVE state then only amf will start duplicate session count of specific PDU session-id and reject after MAX duplication Session count.

## Test Plan
Repro:
image

pcap & logs after the fix:
image

[mme_reg_02_dec_21.log](https://github.com/magma/magma/files/7641866/mme_reg_02_dec_21.log)

Test oai Results:
image

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
+++ b/lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
@@ -189,25 +189,15 @@ void set_amf_smf_context(
 **                                                                        **
 **                                                                        **
 ***************************************************************************/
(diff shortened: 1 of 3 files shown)


ulaskozat (approved):
LGTM

PR #10655, opened 2021-12-02
test(mme): Add SPGW unit tests for force push pco and nas signalling

ssanadhya:

## Summary

This change adds SPGW unit tests to check the logic for processing PCO request with:
- Different configuration protocols
- IP allocation through NAS signaling option present
- Force push enabled/disabled for PCO

It also adds helper functions to check the PCO id list returned and the PCOs present in the PCO response

## Test Plan

`make test_oai`


ulaskozat (approved):
LGTM

PR #10665, opened 2021-12-03
test(mme): Add unit test for TAU procedure

pruthvihebbani:
test(mme): Add unit test for TAU procedure

## Summary

Added unit testcases for periodic and normal TAU without EPS bearer status IE

## Test Plan

- Verified unit testcases are working consistently


ulaskozat (approved):
LGTM

PR #10669, opened 2021-12-03
feat(ci): Add build of new agw container

quentinDERORY:

## Summary

Publish new agw containers to artifactory

## Test Plan

On PR

## Additional Information

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -277,6 +277,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       MAGMA_ROOT: "${{ github.workspace }}"
+      DOCKER_BUILDKIT: 1
     steps:
       - uses: actions/checkout@v2
       - name: Run agw docker compose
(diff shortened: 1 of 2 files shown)


ardzoht (approved):
lgtm

.github/workflows/build_all.yml
```diff
@@ -272,6 +272,103 @@ jobs:
           SLACK_ICON_EMOJI: ":heavy_check_mark:"
           SLACK_COLOR: "#00FF00"
           SLACK_FOOTER: ' '
+  agw-container-build:
```
arunuke, line 275: Is there a way to add multiple labels to the deployed images? If not labels, any other form of annotation? My hope is to be able to see images built over a number of days (ex: last 7 days) that we can then use for nightly builds and tests.

arunuke (approved):

.github/workflows/build_all.yml
quentinDERORY: we could add labels to the docker images. What labels exactly would you want?

arunuke: Can we add commit ids? My hope is to identify which build this is based on.

quentinDERORY: It's already tagging with short sha and latest

PR #10671, opened 2021-12-03
fix(sessiond): prevent to un-suspend if we receive several transient errors

### LINKED ISSUE #10609: [Sessiond] Credits gets suspended and immediately un-suspended on CCA-U with FUA

### Your Environment

- **Version:** 1.5
- **Affected Component:** AGW
- **Affected Subcomponent:** Sessiond
- **Deployment Environment:** Federeted FWA with Gx/Gy federation

### Describe the Issue

When the subscriber quota ends the Gy will answer any CCR-U coming from magma with a CCA-U containing FUI and action configured to TERMINATE.

The behaviour seen is:

1. On the first CCA-U the rules are dropped as expected.
2. The next CCA-U the rules are dropped but are immediately reactivated
3. The subsequent CCA-U the rule is dropped again as expected.
4. The next CCA-U behavior on 2 reproduces

Follows the sessiond log for the suspention and imediatelly unsuspension seen in #2 and #4. (All IMSI values are the same)

image

**Expected behavior**

Magma should always drop the rules that are linked to the RG contained in the CCA-U with FUA TERMINATE

uri200:

## Summary

When we receive two a transient error, we set session into suspended stage. But if we receive another transient error, the session is un-suspended. This PR prevents this situation to happen

## Test Plan

extended test case `test_update_with_transient_error_suspendand_unsuspend` to capture suspension, send of a transient error while suspended and unsuspend using valid quota

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/LocalEnforcer.cpp
+++ b/lte/gateway/c/session_manager/LocalEnforcer.cpp
@@ -1474,8 +1474,13 @@ void LocalEnforcer::update_charging_credits(
         session->receive_charging_credit(credit_update_resp, &uc);
     session->set_tgpp_context(credit_update_resp.tgpp_ctx(), &uc);
 
-    bool should_activate = session->is_credit_ready_to_be_activated(credit_key);
-    if (!should_activate) {
+    // Here we will decide if credit should be activated/un-suspended
+    bool credit_ready_to_be_activated =
+        session->is_credit_ready_to_be_activated(credit_key);
+    auto credit_validity_type =
+        ChargingGrant::get_credit_response_validity_type(credit_update_resp);
+    if (credit_validity_type == TRANSIENT_ERROR ||
+        !credit_ready_to_be_activated) {
       continue;
     }
     // This credit is now out of quota and need to be acted on
--- a/lte/gateway/c/session_manager/ChargingGrant.cpp
+++ b/lte/gateway/c/session_manager/ChargingGrant.cpp
@@ -56,7 +56,7 @@ StoredChargingGrant ChargingGrant::marshal() {
   return marshaled;
 }
 
-CreditValidity ChargingGrant::is_valid_credit_response(
+CreditValidity ChargingGrant::get_credit_response_validity_type(
     const CreditUpdateResponse& update) {
   const uint32_t key             = update.charging_key();
   const std::string session_id   = update.session_id();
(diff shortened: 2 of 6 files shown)


themarwhal (approved):

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1474,8 +1474,11 @@ void LocalEnforcer::update_charging_credits(
         session->receive_charging_credit(credit_update_resp, &uc);
     session->set_tgpp_context(credit_update_resp.tgpp_ctx(), &uc);
 
+    // Here we will decide if a traffic should be activated/un-suspended
     bool should_activate = session->is_credit_ready_to_be_activated(credit_key);
-    if (!should_activate) {
+    auto typeOfCredityValidity =
```
themarwhal:
```suggestion
    auto type_of_credit_validity =
```

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1474,8 +1474,11 @@ void LocalEnforcer::update_charging_credits(
...
+        ChargingGrant::is_valid_credit_response(credit_update_resp);
```
themarwhal: Should we just rename it to `get_credit_response_type`?

lte/gateway/c/session_manager/LocalEnforcer.cpp
```diff
@@ -1474,8 +1474,11 @@ void LocalEnforcer::update_charging_credits(
...
+    if (!should_activate || typeOfCredityValidity == TRANSIENT_ERROR) {
```
themarwhal: Should we just put the TRANSIENT ERROR check inside of `is_credit_ready_to_be_activated` ?

uri200: I was thinking on that but then we have to pass the response to the `is_credit_ready_to_be_activated`

My reasoning is, `is_credit_ready_to_be_activated ` is calculated out of a  current state of the session. `credit_validity_type == TRANSIENT_ERROR` is not a session state. 

So `credit_validity_type == TRANSIENT_ERROR` is just a condition that invalidates any state derived from the session 

That is why I didn't wanted to mix it and i wanted to make it explicit in the condition

PR #10680, opened 2021-12-04
feat(agw): dp_probe_cli Enhancements

saurabhm3hra:
dp_probe_cli Enhancements
1. Check if UE is connected
2. Check if the UE flows are programmed on table 0
3. Check if there is an ACL dropping the packet, if the packet is getting dropped, print the table number
4. Check if there is a UE enforcement flow in table 13
image


hcgatewood (approved):
Doc changes lgtm

pshelar (changes requested):
Can you add examples for all enhancements to this doc for reference.

ekowtaylor:
@saurabhm3hra, we are cleaning up the backlog and saw this PR has been pending some action from you. Do you plan to still work on this or no?

pshelar (approved):
Lets get the enhancement in, we can improve do later.

PR #10682, opened 2021-12-05
feat(amf): Minor clean up in ambr calculation

panyogesh:
Fix:
  1. Storing the AMBR calculation in smf_ctxt when received
     from AMF
  2. Using the same for the PDUSessionAggregateMaximumBitRate
  3. Correcting the DL bandwidth update from SMF

Test:
  1. Basic Unit Testing
  2. PDU Session Establishment with Simulator

## Additional Information


lte/gateway/c/session_manager/SessionStateEnforcer.cpp
```diff
@@ -651,7 +651,7 @@ void SessionStateEnforcer::prepare_response_to_access(
   rsp->mutable_session_ambr()->set_max_bandwidth_dl(
       config.rat_specific_context.m5gsm_session_context()
           .default_ambr()
-          .max_bandwidth_ul());
+          .max_bandwidth_dl());
```
themarwhal, line 654: Was this a bug? Maybe we should have a unit test to cover these conversions?

themarwhal (approved):
sessiond changes look good to me! (see small note about potentially adding unit test coverage for the modified function)

lte/gateway/c/session_manager/SessionStateEnforcer.cpp
panyogesh: Sure.. Will add the testcase

uri200:
Hey @panyogesh  is this fixed somewhere else? should we close this PR?

panyogesh:
Fixed as part of #10821

PR #10687, opened 2021-12-06
chore(agw): Benchmark state serialization protobuf versus flatbuffer

### LINKED ISSUE #8085: Benchmark state serialization with minimal test object between protobuf and flatbuffers
Quantify and obtain measurements of the memory and time spent on serialization using protobuf vs flatbuffers library, using a minimal test object (can be retrieved running any of the attach detach S1AP tests and obtaining it by running `state_cli.py`: https://github.com/magma/magma/blob/master/docs/readmes/lte/dev_notes.md#checking-redis-entries-for-stateless-services) 

Benchmark performance data such as memory, CPU spent, time during serialization and deserialization using protobuf against flatbuffers.

Plan to follow:
1. Write 2 test scripts where we create a protobuf message and a flatbuffer msg that contains the same values / structure (the message can be retrieved from state_cli.py while running different s1ap tests and having UEs attached)
2. Include both protobuf and flatbuffer serialization methods (Message to string)
3. Quantify and measure memory and time spent on each script

Get results for:
1. 50 UEs attached test state
2. 200 UEs attached test state

- - - - - - - - - - - - - - - - - - - - - - -

Results : 

Below are summered the bench-marking of protobuf versus FlatBuffers serialization libraries.

All units are in nano seconds.
"FB REDIS X UEs" means with FlatBuffers libs, the measurement is the time to send X UE MM context data to Redis DB.
"FB REDIS X UEs" means with Protobuf libs, the measurement is the time to send X UE MM context data to Redis DB.
"FB SER X UEs" means with FlatBuffers libs, the measurement is the time to hash the X flatbuffers corresponding to X UE MM contexts before sending to Redis DB, there is zero serialization cost in the way we used Flatbuffers.
"PB SER X UEs" means with Protobuf libs, the measurement is the time to serialize X UE MM contexts in buffers, hash the buffers before sending to Redis DB.
 
| Measurements         |  Min Avg  | Max Avg  | **Mean Avg** | Std Avg |
| -----------------------     |-------------:|-------------:|-------------:|-------------:|
| FB REDIS 50 UEs   | 311408    | 2426675  |   **744692** |   400495 |
| PB REDIS 50 UEs   | 140896    | 1905692  |   **427056** |   322449 |
| FB REDIS 200 UEs | 272628    | 3345360  |  **623899**  | 345354   |
| PB REDIS 200 UEs | 115315    | 2221363  |  **331389**  | 237935   |
| FB SER   50 UEs    |   17954    | 307389    |  **51136**     | 49664     |
| PB SER   50 UEs    |   48540    | 494194    |  **146876**     | 93116     |
| FB SER  200 UEs   |    16232   | 403065    | **48225**      | 46835     |
| PB SER  200 UEs   |    43140   | 550563    | **126045**      | 78286     |

Here are the details of the series of measurements :

figure_2

Facts : 
The serialization size of the UE context has been reduced to 35 KB approximately, it cannot be greater than 64 KB due to the FlatBuffers API, the real initial size of the buffer to transmit should have been 65 KB.
The Flatbuffers schema do not support bit-fields, it would have been a nice to have.
In this benchmark, the Flatbuffer serialization cost is zero. The API has been used exclusively for this purpose, having a zero copy API.

Observations:
Redis transfer takes roughly twice for FlatBuffers compared to Protobuf.
Protobuf serialization is greater by a magnitude of 3 compared to Flatbuffers.
Protobuf serialization benchmarks are negligible compared to the Redis transfer overhead for Flatbuffers despite that we cheat by a factor of 2 for the size of FlatBuffers buffer.

Magma source code for benchmarking flatbuffers can be found [here](https://github.com/lionelgo/magma/tree/bench-flatbuffers2)

lionelgo:
## Summary

The purpose of the benchmark is to compare the serialization of an AGW LTE  UE MM context into redis DB with protobuf versus FlatBuffers


lionelgo:
Update:

## Summary

The purpose of the benchmark is to compare the serialization of an AGW LTE  UE MM context into redis DB with protobuf versus [FlatBuffers](http://google.github.io/flatbuffers/index.html#flatbuffers_overview)

## Additional Information

### FlatBuffers

The goal of using  FlatBuffers is to take advantage of zero-copy, meaning no serialization cost (CPU) from the handled contexts in tasks MME_APP, S1AP, etc. 

So this first trial is done with mutables FlatBuffers. In order to have no serialization phase, we work with pre-allocated arrays of structs and pre-allocated arrays of bytes for strings, meaning we do not use object based API (flatc --gen-object-api).
The consequence of this choice would be that the cost of serialization would be mainly the cost of the Redis client call to send the context to Redis server. and could outperform the protobuf implementation.

We encountered an issue after trying to handle most of the attributes of the UE MM context in the [IDL schemas](http://google.github.io/flatbuffers/flatbuffers_guide_writing_schema.html) files (lte/gateway/idl/oai/mme_ue_state.fbs) . This issue is that we already got a root UE_MM_CONTEXT object size slightly greater than allowed (65536 bytes). We had to reduce the number of allowed PDN contexts (5 instead of 10 in current implementation), and reduce the biggest size of ESM message we could receive.

### Benchmarking

For time benchmarking both PB and FB, std::chrono is used and measurements are flushed to stdout.
For memory benchmarking, [rusage](https://linux.die.net/man/2/getrusage)  is used.


## How to benchmark

 Inside magma VM, build mme: 
$ cd $MAGMA_ROOT/lte/gateway
$ make build_oai
# Prepare forder for results
$ cd $MAGMA_ROOT/lte/gateway/c/scripts; mkdir RUN1; cd RUN1
# Do the benchmarking for Protobuf and Flatbuffers
$ $MAGMA_ROOT/lte/gateway/c/scripts/mme_ue_context_serialization_benchmarking
# You should see several log files in RUN1 folder

To display the timing measurements, on your host, go in the same RUN1 folder:
 $MAGMA_ROOT/lte/gateway/c/scripts/mme_ue_context_serialization_plotting

lte/gateway/c/core/oai/CMakeLists.txt
```diff
@@ -41,9 +41,9 @@ add_definitions(-DCMAKER)
 # Generate *.gcno rather than *.c.gcno files for gcov
 set(CMAKE_C_OUTPUT_EXTENSION_REPLACE 1)
 
-set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS} -fstack-protector-all -g -DMALLOC_CHECK_=3 -DDEBUG_IS_ON=1 -DTRACE_IS_ON=1 -O0 -fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer")
```
ssanadhya: Why do we have to disable ASAN/LSAN for this benchmarking?

ardzoht: +1, same question regarding ASAN

lte/gateway/c/core/oai/tasks/mme_app/mme_app_test_serialization.cpp
```diff
@@ -0,0 +1,297 @@
+/**
```
ardzoht: I would suggest, moving this file along with the `..._generated.h` files from flatbuffer compiler to a folder called `lte/gateway/c/core/oai/experimental` and then modify CMakeLists to include this as a library, adding this to the `tasks` and `include` folders does not seem like the right place, thoughts @ssanadhya ?

ardzoht (changes requested):

lte/gateway/c/core/oai/include/mme_config.h
```diff
@@ -388,6 +392,9 @@ typedef struct mme_config_s {
   uint8_t daylight_saving_time;
 
   run_mode_t run_mode;
+  // Integer value for testing serialization
+  test_type_t test_type;
+  uint32_t test;
```
ardzoht: this can have a more meaningful name, how about `test_param`?

lte/gateway/c/core/oai/CMakeLists.txt
```diff
@@ -56,14 +57,19 @@ set(CMAKE_EXE_LINKER_FLAGS_DEBUG "${CMAKE_EXE_LINKER_FLAGS_DEBUG} -fprofile-arcs
...
 
 ################################################################
 # Set proto directory for state proto definitions
 ################################################################
 set(STATE_PROTO_DIR $ENV{MAGMA_ROOT}/lte/protos/oai)
 
+################################################################
+# Set flatbuf directory for state Flatbuffers definitions
+################################################################
+set(STATE_FLAT_DIR $ENV{MAGMA_ROOT}/lte/idl/oai/experimental)
```
ardzoht, line 71: 👍

lte/gateway/c/core/oai/include/state_manager.h
```diff
@@ -165,21 +168,77 @@ class StateManager {
...
 
+    auto start = std::chrono::high_resolution_clock::now();
     std::string proto_str;
     ProtoUe ue_proto = ProtoUe();
     StateConverter::ue_to_proto(ue_context, &ue_proto);
     redis_client->serialize(ue_proto, proto_str);
     std::size_t new_hash = std::hash<std::string>{}(proto_str);
-
+    auto stop            = std::chrono::high_resolution_clock::now();
+    std::cout << "TIME PROTOBUF CONVERSION : "
```
ardzoht, line 185: I would suggest enclosing all benchmarking code that is added on a CMake boolean flag, similar to the one we're using for MME unit testing: https://sourcegraph.com/github.com/magma/magma/-/blob/lte/gateway/c/core/oai/tasks/mme_app/mme_app_context.c?L733:6 

This will help to keep current implementation code (without benchmark code) separate, what do you think?

lte/idl/oai/experimental/mme_ue_state.fbs
```diff
@@ -0,0 +1,123 @@
...
+  ts10_retransmission_count:uint;
+  ts13_timer:NasTimer;
+  ts13_retransmission_count:uint;
+  // MessageDef* message_p;
+  sgsap_msg:SgsApMsg;
+  ongoing_procedure:OngoingProcedure;
+  tau_active_flag:bool;
+  service_indicator:ubyte;
+  //csfb_service_type_t csfb_service_type;
+  call_cancelled:bool;
```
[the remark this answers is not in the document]
lionelgo: done for this fbs file

lte/idl/oai/experimental/mme_nas_state.fbs
```diff
@@ -0,0 +1,404 @@
...
+}
+
+enum EmmScType:ubyte {
+  SECURITY_CTX_TYPE_NOT_AVAILABLE = 0,
+  SECURITY_CTX_TYPE_PARTIAL_NATIVE,
+  SECURITY_CTX_TYPE_FULL_NATIVE,
+  SECURITY_CTX_TYPE_MAPPED
+}
+
+// Be carefull, was bitfield
```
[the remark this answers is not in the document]
lionelgo: Done

lte/idl/oai/experimental/mme_nas_state.fbs
```diff
@@ -0,0 +1,404 @@
...
+struct Count {
+  spare:ubyte;
+  overflow:uint16;
+  seq_num:ubyte;
+}
+
+// Be carefull, was bitfield with quartets
```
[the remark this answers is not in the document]
lionelgo: Done

lte/gateway/c/core/oai/CMakeLists.txt
lionelgo: Forgot to re-enable it after valgrind trials

lte/gateway/c/core/oai/include/mme_config.h
lionelgo: yes, better.

lte/gateway/c/core/oai/include/state_manager.h
lionelgo: Yes, I propose EXPERIMENTAL_TEST or BENCHMARK_TEST ? Please suggest.

ardzoht:
@lionelgo Looks like latest commit included unrelated changes which massively expanded the codeowners and also the changed files, please resolve by either reverting the commit or do `git reflog` to fix.

lionelgo:
Replaced by https://github.com/magma/magma/pull/11185

PR #10691, opened 2021-12-06
chore: Skip build on CI while integ testing

tmdzk:
## Summary

chore: Skip build on CI while integ testing

## Test Plan

Install specific version and skip build part on CI 

## Additional Information

### DIFF
--- a/lte/gateway/fabfile.py
+++ b/lte/gateway/fabfile.py
@@ -219,7 +219,7 @@ def s1ap_setup_cloud():
 
 def integ_test(
     gateway_host=None, test_host=None, trf_host=None,
-    destroy_vm='True', provision_vm='True',
+    destroy_vm='True', build='True', provision_vm='True',
 ):
     """
     Run the integration tests. This defaults to running on local vagrant
@@ -255,7 +255,10 @@ def integ_test(
         gateway_ip = gateway_host.split('@')[1].split(':')[0]
 
     execute(_dist_upgrade)
-    execute(_build_magma)
+    if build:
+        execute(_build_magma)
+    else:
+        execute(_from_registry)
     execute(_run_sudo_python_unit_tests)
     execute(_start_gateway)
 
@@ -535,6 +538,14 @@ def _dist_upgrade():
     run('sudo apt-get update')
     run('sudo DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade')
 
+def _from_registry():
+    """
+    Install magma from dev registry
+    """
+    hash = pkg.get_commit_hash('git')
+    with cd(AGW_ROOT):
+        run('cd deploy')
+        run('sudo bash agw_install_specific.sh %s' % hash)
 
 def _build_magma():
     """
--- a/lte/gateway/fabfile.py
+++ b/lte/gateway/fabfile.py
@@ -219,7 +219,8 @@ def s1ap_setup_cloud():
 
 def integ_test(
     gateway_host=None, test_host=None, trf_host=None,
-    destroy_vm='True', build='True', provision_vm='True',
+    destroy_vm='True', build=None, hash=None,
+    provision_vm='True',
 ):
     """
     Run the integration tests. This defaults to running on local vagrant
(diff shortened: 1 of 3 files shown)


lte/gateway/fabfile.py
```diff
@@ -535,6 +541,13 @@ def _dist_upgrade():
     run('sudo apt-get update')
     run('sudo DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade')
 
+def _from_registry():
```
uri200, line 544: shouldn't you be passing the hash here?

lte/gateway/fabfile.py
```diff
@@ -219,7 +219,8 @@ def s1ap_setup_cloud():
 
 def integ_test(
     gateway_host=None, test_host=None, trf_host=None,
-    destroy_vm='True', provision_vm='True',
+    destroy_vm='True', build=None, hash=None,
```
uri200, line 222: maybe it would be better to use `github_hash` or `magmacore_hash`. Python has a hash module

uri200: maybe instead of `build` you can use something like `useBuildFromRepo` or `use_repo` or something different that `build` 

The reason for that is for cwf integ test we already have build flag. That flag means that the test is run with whatever is installed on the vm. That is used for testing when you are building specific services on agw locally . That allow us to run the test without building the whole thing.   Maybe in the future we should add that `build` behaviour to  `lte_integ_test` too. So it would be good to be consistent

uri200: In fact, maybe you can combine both flags. IF `magmacore_hash` is not null, that means you want to build from repo (?)

lte/gateway/fabfile.py
```diff
@@ -253,11 +254,16 @@ def integ_test(
...
+    print("bbuild is set to: %s" % build)
 
-    execute(_dist_upgrade)
-    execute(_build_magma)
+    if not build:
+        execute(_dist_upgrade)
+        execute(_build_magma)
+        execute(_start_gateway)
+    else:
+        execute(_from_registry)
```
uri200, line 264: what about reversing the conditions. It is always nicer to read positive conditions

```
    if  useBuildFromRepo:
        execute(_from_registry)
   else
        execute(_dist_upgrade)
        execute(_build_magma)
        execute(_start_gateway)
```

lte/gateway/fabfile.py
```diff
@@ -535,6 +541,13 @@ def _dist_upgrade():
     run('sudo apt-get update')
     run('sudo DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade')
 
+def _from_registry():
+    """
+    Install magma from dev registry
+    """
+    hash = pkg.get_commit_hash('git')
```
uri200, line 548: shouldn't this take the hash as an argument?

lte/gateway/fabfile.py
quentinDERORY, line 222: you added hash here but you're getting from elsewhere on line 548?

lte/gateway/fabfile.py
tmdzk: so right now, I'm dynamically fetching it, but yeah it's WIP

lte/gateway/fabfile.py
tmdzk: yes that's the plan my dear Oriol. It will be changed when I implement the install_magma_from_registry function, right now it's just calling the _install_from_registry that dynamically grab the hash and install the package 

In the future I want dev to be able to do `fab install_magma:hash=<MY_HASH>` and have their built version tested locally

lte/gateway/fabfile.py
tmdzk: Nah I'm not a positive guy sorry

lte/gateway/fabfile.py
tmdzk: soon soon with the new public function

lte/gateway/fabfile.py
tmdzk: yeah I was about to write the feature that @uri200 requested but wanted to ship that quickly, will get back to it in a different PR

quentinDERORY (approved):

PR #10701, opened 2021-12-07
chore: In integ_test don't run unit test

tmdzk:

chore: In integ_test don't run unit test

## Summary

We're already running them on the agw workflow, no need to run them while we do the integ_test
https://github.com/magma/magma/blob/master/.github/workflows/agw-workflow.yml#L108

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -76,10 +76,10 @@ jobs:
       - name: Run apt-focal-install-aioeventlet
         run: |
             wget https://artifactory.magmacore.org:443/artifactory/api/gpg/key/public -O /tmp/public
-            apt-key add /tmp/public
+            sudo apt-key add /tmp/public
             echo "deb https://artifactory.magmacore.org/artifactory/debian ${{ env.OS_VERSION }}-${{ env.DISTRIBUTION}} main" >> /etc/apt/sources.list.d/magma.list
             apt update
-            apt install magma-libfluid libopenvswitch  openvswitch-switch  openvswitch-common  openvswitch-datapath-dkms 
+            apt install magma-libfluid libopenvswitch  openvswitch-switch  openvswitch-common  openvswitch-datapath-dkms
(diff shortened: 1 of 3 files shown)


tmdzk:
Tests are passing: https://github.com/magma/magma/runs/4438216314?check_suite_focus=true even with the flag

quentinDERORY (changes requested):
Looks like you are skipping them in both places?

quentinDERORY (approved):

lte/gateway/Makefile
```diff
@@ -202,7 +202,7 @@ scan_oai: ## Scan OAI
 test_python: stop ## Run all Python-specific tests
 	make -C $(MAGMA_ROOT)/lte/gateway/python test_all
 
-test_sudo_python: stop ## Run Python tests that require sudo (datapath, etc.)
+test_sudo_python: stop ## Run Python tests that require sudo (datapath, etc.) hello
```
ardzoht: ??

ardzoht (approved):
lgtm

.github/workflows/agw-workflow.yml
```diff
@@ -106,6 +106,9 @@ jobs:
       - name: Execute test_all
         run: |
             make -C ${{ env.MAGMA_ROOT }}/lte/gateway/python test_all
+      - name: Execute test_all with skip flag
+        run: |
+            make -C ${{ env.MAGMA_ROOT }}/lte/gateway/python test_all SKIP_NON_SUDO_TESTS=1
```
themarwhal: Did you try removing the env var above? `SKIP_SUDO_TESTS: 1`

lte/gateway/fabfile.py
```diff
@@ -256,7 +256,6 @@ def integ_test(
 
     execute(_dist_upgrade)
     execute(_build_magma)
-    execute(_run_sudo_python_unit_tests)
```
themarwhal, line 259: If we're committing this let's just remove the function from the file too

PR #10703, opened 2021-12-07
fix(mme): Add unit test for paging

ulaskozat:

## Summary

- Add unit test for paging using UE IP address. 
- Make paging timeout value passed from config to reduce the timer value in unit tests.

## Test Plan

`make test_oai` in magma vm.

## Additional Information


ardzoht (approved):
lgtm

PR #10710, opened 2021-12-07
docs: develop devcontainer locally

Neudrino:

## Summary

If DevContainer is to become more popular, we should be able to develop the devcontainer itself locally. Changes to the container would otherwise be rather hard to implement and test.
For testing we want to be able to rebuild the devcontainer locally, however for normal usage, it should still be pulled from GHCR.
Thus this was implemented as an extension of the documentation around devcontainer.

## Test Plan

I was able to build a new devcontainer locally on a Linux host system applying the changes suggested in the documentation. Essentially replace
```json
    "image": "ghcr.io/magma/magma/devcontainer:latest",
```
with
```json
    "build": {
        "dockerfile": "Dockerfile",
        "context": "..",
    },
```
and use the option F1 - "Remote-Containers: Rebuild Container" within VScode.

You can also rebuild the devcontainer in Github codespaces by opening a devcontainer, changing the above lines, and than activating the rebuild, which is suggested by the UI.

## Additional Information

With the restructuring of the documentation in progress, I found it safest to collect the documentation of the devcontainer as close to the code as possible and hence placed it inside the readme accompanying the devcontainer setup.


.devcontainer/README.md
```diff
@@ -2,6 +2,40 @@
...
+```json
+	"build": {
+        "dockerfile": "Dockerfile",
+        "context": "..",
+    },
+```
+which requires a connection to the internet to download the respective dependencies. It is important to note that the default context for the `build:` block in the `devcontainer.json` is the `.devcontainer` folder below the repository root. Hence, you must set the context next to the dockerfile to be able to resolve all dependency imports properly.
+If you would want to build the dockerfile manually from the console, you would start the command `docker build -f .devcontainer/Dockerfile .` in the repository root.
+
 ## Setup
```
themarwhal: I think some of the doc below might be outdated, do you mind scanning through those as well?

.devcontainer/README.md
```diff
@@ -2,6 +2,40 @@
...
+## Prerequisites and Restrictions
+
+DevContainer is a technology developed by Microsoft. Thus, it is only available for [_VS Code_](https://code.visualstudio.com/docs/remote/containers), but **not** for _IntelliJ_. However, as Github is also currently owned by Microsoft, Github provides a DevContainer integration named ["Codespaces"](https://docs.github.com/en/codespaces).
+In order to use the Github codespaces, you must be member of the Magma Github organisation.
+In any case you will need a working Docker installation, as DevContainer is a docker based technology.
+
+Prerequisites in Summary:
+* Local usage:
+  * Docker installed
+  * VS code installed
```
themarwhal: Might want to mention here the container extension we need to utilize this

.devcontainer/README.md
```diff
@@ -2,6 +2,40 @@
 
 For a comprehensive developer environnment supporting AGW c / c++ workflows (extensible with ~effort to other Magma uses) - see the instructions below.
 
+## Prerequisites and Restrictions
+
+DevContainer is a technology developed by Microsoft. Thus, it is only available for [_VS Code_](https://code.visualstudio.com/docs/remote/containers), but **not** for _IntelliJ_. However, as Github is also currently owned by Microsoft, Github provides a DevContainer integration named ["Codespaces"](https://docs.github.com/en/codespaces).
+In order to use the Github codespaces, you must be member of the Magma Github organisation.
```
themarwhal: For Codespaces, we have a doc here so we should just link to it. https://docs.magmacore.org/docs/next/contributing/contribute_vscode

.devcontainer/README.md
Neudrino: Yes, I can have a look there as well. I might not touch the Mac specific stuff though, because I am on a Linux system. :innocent:

.devcontainer/README.md
Neudrino: Very good idea!

As a next step one could even have a look, whether a predefined workspace equivalent to the ones defined in https://github.com/magma/magma/tree/master/vscode-workspaces would be an option.
But I would prefer to have this as a separate step and PR, as this would be another topic (for me at least).

.devcontainer/README.md
Neudrino: Yes, I was aware of this section, however, with the ongoing documentation restructuring I was afraid to intervene there. :innocent: But you are right, adding a link will probably be a good idea nevertheless.

electronjoe:
Thank you for the documentation update - I myself copy and paste this snippet change in order to test locally my changes in the Dockerfile. So clearly it deserves documentation (and I didn't think of it).

.devcontainer/README.md
electronjoe: There are a couple `misspell` GH Action findings in this PR, would you mind addressing? I don't think this is a colour vs color type of thing but I could be wrong =D

Neudrino: Yes, it seems that `misspell` favours the American English over British English, but at my school we were rather trained in BE flavour. :innocent: I will go over it again, if the `misspell` complains.

Neudrino:
* I reworked it a bit more and move some parts from the ReadMe to the official DocuSaurus. I hope, this is to your liking. :slightly_smiling_face: 
* I added some whitespace changes on unrelated code to make the markdown precommit linter happy.
* I did _not_ change anything on the C/C++ specific part, as I am not so much into it. :innocent:

.devcontainer/README.md
```diff
@@ -125,22 +133,3 @@ root@ecee08edef4b:/build/c/oai# cat clang.findings | egrep "android|bugprone|cer
...
-- Docker running out of disk on Mac VM
-  - Docker just leaks resources like mad
-  - Eventually you'll get a "out of space" error
-  - Or an error about apt-get failure
-  - TODO: find good docker command to wipe everything
-
-## TODOs
-
-- Get the Docker image up to DockerHub?
-  - Want easy way to build from Dockerfile locally though..
```
Neudrino, line 146: ### Sharp edges
- One can also a local copy of the repo and only mount it. This is imho much safer and less resource hungry. It is documented now. So the first point should be mitigated.
- Docker image cleanup is also covered in the official documentation now in the troubleshooting section, including a cleanup command.

### TODOs
- My assumption was that having GHCR, DockerHub us superfluous.
- Build locally. :heavy_check_mark:

electronjoe: Thanks for the additional cleanup. LGTM.

electronjoe (approved):

PR #10711, opened 2021-12-07
feat(ci): more tags on images makes destinction easier

Neudrino:

## Summary

When developing the devcontainer image, it makes like much easier to have more speaking tags on the images to select from. So for one change, the default tagging was not replaced by the `sha`, but the `sha` was added in addition to [the default tagging](https://github.com/marketplace/actions/docker-metadata-action#tags-input) (`type=ref` and `type=schedule`).
On the other hand it is seen value in using the latest tag as well, to shorten feedback cycles and get to know issues with any new version of devcontainer much faster. I personally would also see this as a prerequisite for more steps towards a continuous delivery workflow for the devcontainer.

## Test Plan

As this are CI changes, I think the CI is the way to test it.

## Additional Information

As far as I am concerned, additional tags on images never hurt, as they provide additional useful information at nearly no cost. However, this is a more general view, which would exceed the scope of this specific use case.

### DIFF
--- a/.github/workflows/composite/docker-builder/action.yml
+++ b/.github/workflows/composite/docker-builder/action.yml
@@ -10,7 +10,12 @@ inputs:
     required: true
   IMAGE_TAGS:
     required: true
-    default: type=sha
+    default: |
+      type=sha
+      type=schedule
+      type=ref,event=branch
+      type=ref,event=tag
+      type=ref,event=pr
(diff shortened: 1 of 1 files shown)


quentinDERORY (approved):

electronjoe (approved):
Thanks! Seems well motivated.

PR #10721, opened 2021-12-07
test(mme): Cut the execution time for test_mme_app by half

ulaskozat:

## Summary

I removed unnecessary wait at the end of each test case that was used as a workaround for reducing test flakes before message based syncs. The average run time for `test_mme_app` is reduced from ~50 seconds to ~25 seconds.

## Test Plan

Run `make test_oai` 100 times with no failures on `test_mme_app`.

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/test/mme_app_task/test_mme_procedures.cpp
+++ b/lte/gateway/c/core/oai/test/mme_app_task/test_mme_procedures.cpp
@@ -116,10 +116,10 @@ class MmeAppProcedureTest : public ::testing::Test {
   virtual void TearDown() {
     bdestroy_wrapper(&nas_msg);
     send_terminate_message_fatal(&task_zmq_ctx_main);
+    // Sleep to ensure that messages are received and contexts are released
+    std::this_thread::sleep_for(std::chrono::milliseconds(500));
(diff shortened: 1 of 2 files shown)


ulaskozat:
cc @ardzoht @ssanadhya @uri200 @rsarwad @pruthvihebbani

ardzoht (approved):
lgtm

ssanadhya (approved):

lte/gateway/c/core/oai/test/mme_app_task/test_mme_procedures.cpp
```diff
@@ -116,10 +116,10 @@ class MmeAppProcedureTest : public ::testing::Test {
   virtual void TearDown() {
     bdestroy_wrapper(&nas_msg);
     send_terminate_message_fatal(&task_zmq_ctx_main);
+    // Sleep to ensure that messages are received and contexts are released
+    std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
rsarwad, line 120: We can define macro for 500

ulaskozat: Since it is only used here now, I did not do that.

PR #10736, opened 2021-12-08
chore(ci): update devcontainer image in CI

alexzurbonsen:

## Summary

Update the devcontainer image in CI and devcontainer.json

The image stream change is due to parametrisation and the use of Github variables within the Github composite actions.
A list of available imagestreams can be found under [packages](https://github.com/orgs/magma/packages?repo_name=magma).

## Additional Information

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -19,7 +19,7 @@
 		"augustocdias.tasks-shell-input",
 		"ryuta46.multi-command",
 	],
-	"image": "ghcr.io/magma/devcontainer:sha-3c02c72",
+	"image": "ghcr.io/magma/magma/devcontainer:sha-2f4586b",
 	"settings": {
 		"terminal.integrated.shell.linux": "/bin/bash",
 		"files.watcherExclude": {
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -17,7 +17,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  DEVCONTAINER_IMAGE: "ghcr.io/magma/devcontainer:sha-3c02c72"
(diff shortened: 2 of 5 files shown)


Neudrino (approved):

Neudrino:
Btw, the image stream change is due to parametrisation and the use of Github variables within the Github composite actions. It should not be any issue though.

quentinDERORY (approved):

electronjoe:
> Btw, the image stream change is due to parametrisation and the use of Github variables within the Github composite actions. It should not be any issue though.

Thanks for the explanation!

electronjoe (approved):
Looks like all workflows pass! CodeCov is in progress but I expect that to be GTG (plus it's required so LGTM here is safe).

Neudrino (changes requested):
This PR depends on https://github.com/magma/magma/pull/10755. We will update the hash again, when changes from https://github.com/magma/magma/pull/10755 are available in GHCR.

To not change image twice, I will mark this as "Request changes" to avoid accidental merging.

Neudrino (approved):

Neudrino:
Docker image hash was updated after https://github.com/magma/magma/pull/10755 was merged.

Neudrino:
This PR is required by https://github.com/magma/magma/pull/10341.

themarwhal:
Hey @Neudrino, does this mean we should update other usages like this one? https://github.com/magma/magma/blob/master/lte/gateway/python/precommit.py#L27

Neudrino:
> Hey @Neudrino, does this mean we should update other usages like this one? https://github.com/magma/magma/blob/master/lte/gateway/python/precommit.py#L27

Hi @themarwhal , that depends.
Of course the old image is still available, and if there are no changes to the Dockerfile it will be the same as the new one. So there is no immediate necessity to do so. However, in long term, an update of the image stream tag is sensible also for all other images used. I created an issue for it https://github.com/magma/magma/issues/10828 and assigned it to myself.

PR #10742, opened 2021-12-08
test(mme): Include packet filter count in unit test NW initiated bearer activation

ssanadhya:
## Summary

The sample UL and DL traffic flow template in dedicated bearer activation/deactivation unit test did not specify the number of packet filters, causing it to skip add_tunnel_helper function call. This change fixes it.

This change also leans up an unused function on MME context

## Test Plan

make test_oai


ardzoht (approved):
lgtm

PR #10750, opened 2021-12-09
fix(agw): Fix websocket-client pydep version mismatch

ardzoht:

## Summary

- Fixes `websocket-client` python package mismatch version between magma and magma_test VMs due to difference on installed python versions.

## Test Plan

- `make` on magma VM
- `make` on magma_test VM

## Additional Information


pshelar (approved):
Thanks for fixing this.

PR #10757, opened 2021-12-09
feat(orc8r): Added network- and tenant-scoped endpoint security

christinewang5:
## Summary
* Added network and tenant ID scoping functionality 

## Test Plan
* Unit tests in `rest_middleware_test`


hcgatewood:
Will check this out once the dependent PR merges 🙂

orc8r/cloud/go/services/certifier/protos/certifier.proto
```diff
@@ -76,7 +76,13 @@ message Resource {
   Effect effect = 1;
   Action action = 2;
   ResourceType resourceType = 3;
+  // Resource is the path of the resource
   string resource = 4;
+  // Resource ID is an optional field for tenant and network resources
+  oneof resource_id {
+    string tenant_id = 5;
+    string network_id = 6;
+  }
```
christinewang5, line 91: Although these seem a bit redundant, I did this because network/tenant resources should match for both their network/tenant ID resource policies and their URL types so here it just passes both through. I figured if we could parse the resource out at the rest middleware stage, we wouldn’t have to do the thing where we had to semantically understand each path to get the resource id.

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -768,6 +801,33 @@ func checkAction(reqResource *certprotos.Resource, policyResource *certprotos.Re
...
+	resourceID := ""
+	if reqResource.ResourceType == certprotos.ResourceType_NETWORK_ID {
+		resourceID = reqResource.GetNetworkId()
+	}
+	if reqResource.ResourceType == certprotos.ResourceType_TENANT_ID {
+		resourceID = reqResource.GetTenantId()
+	}
+	if finalEffect != certprotos.Effect_DENY && resourceID == policyResource.Resource {
+		finalEffect = policyResource.Effect
+	}
```
christinewang5: Hmmm should I create two separate types for reqResource and policyResource?
Right now what's happening is:
reqResource.resource is always a path & reqResource.resourceID is the network/tenantID
policyResource.resource could be path, networkID, and tenantID

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -125,15 +125,24 @@ func TokenMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
...
+		getPDReq := certprotos.GetPolicyDecisionRequest{
 			Username: username,
 			Token:    token,
 			Resource: &certprotos.Resource{
-				Action:   getRequestAction(req, nil),
-				Resource: req.RequestURI,
+				Action:       getRequestAction(req, nil),
+				ResourceType: resourceType,
+				Resource:     req.RequestURI,
 			},
```
hcgatewood: I think there's still a slight disconnect on what we wanna send here. Specifically, we always want to send *both* the request URI and and network ID (if any)

So e.g.

```go
		getPDReq := certprotos.GetPolicyDecisionRequest{
			Username: username,
			Token:    token,
			Resource: &certprotos.Resource{
				Action:       getRequestAction(req, nil),
				Resource:   req.RequestURI,
				NetworkID: /* network_id, if any */,
			},
```

That way, whether the token contains a path-based vs. networkID-based vs. tenantID-based token, it has all the info it needs to make a decision

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -83,18 +93,18 @@ func TestAuthMiddleware(t *testing.T) {
 		{"GET", fmt.Sprintf("%s%s", urlPrefix, RegisterNetworkV1), test_utils.TestRootUsername, rootToken, http.StatusOK},
 		{"GET", fmt.Sprintf("%s%s", urlPrefix, "/malformed/url"), test_utils.TestRootUsername, rootToken, http.StatusOK},
 		{"PUT", fmt.Sprintf("%s%s", urlPrefix, "/malformed/url"), test_utils.TestRootUsername, rootToken, http.StatusOK},
-		{"GET", fmt.Sprintf("%s%s", urlPrefix, tenantsh.TenantInfoURL), test_utils.TestRootUsername, rootToken, http.StatusOK},
-		{"POST", fmt.Sprintf("%s%s", urlPrefix, tenantsh.TenantInfoURL), test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"GET", fmt.Sprintf("%s%s%s%s", urlPrefix, TenantRootPathV1, obsidian.UrlSep, strconv.Itoa(test_utils.TestTenantId)), test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"POST", fmt.Sprintf("%s%s%s%s", urlPrefix, TenantRootPathV1, obsidian.UrlSep, strconv.Itoa(test_utils.TestTenantId)), test_utils.TestRootUsername, rootToken, http.StatusOK},
```
hcgatewood: Only admins should be able to make edits to the tenants path

The "tenant" policy type is just a dynamic alias for a list of network IDs

hcgatewood (changes requested):
Skimmed first half -- lmk if have questions 🙂

orc8r/cloud/go/obsidian/access/rest_middleware.go
christinewang5: I set it in the next few lines, bc using one of in proto makes it so difficult to just put any field.

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
christinewang5: Ok so should I just put in a const like adminOnlyPaths?

hcgatewood: Uh nah it's okay if someone creates e.g. a path-specific policy giving someone access to the tenants endpoints. But it's more in reference to my other comment about the purpose of the tenant-scoped policies. Hopefully my msg with the proposed proto changes will help align us

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -76,25 +86,26 @@ func TestAuthMiddleware(t *testing.T) {
...
 		{"GET", fmt.Sprintf("%s%s", urlPrefix, RegisterNetworkV1), test_utils.TestRootUsername, rootToken, http.StatusOK},
 		{"POST", fmt.Sprintf("%s%s", urlPrefix, RegisterNetworkV1), test_utils.TestRootUsername, rootToken, http.StatusOK},
-		{"GET", fmt.Sprintf("%s%s", urlPrefix, RegisterNetworkV1), test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"GET", fmt.Sprintf("%s%s%s%s", urlPrefix, RegisterNetworkV1, obsidian.UrlSep, TEST_NETWORK_ID), test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"PUT", fmt.Sprintf("%s%s%s%s", urlPrefix, RegisterNetworkV1, obsidian.UrlSep, TEST_NETWORK_ID), test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"GET", fmt.Sprintf("%s%s%s%s", urlPrefix, RegisterNetworkV1, obsidian.UrlSep, WRITE_TEST_NETWORK_ID), test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"PUT", fmt.Sprintf("%s%s%s%s", urlPrefix, RegisterNetworkV1, obsidian.UrlSep, WRITE_TEST_NETWORK_ID), test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"GET", fmt.Sprintf("%s%s%s%s", urlPrefix, TenantRootPathV1, obsidian.UrlSep, strconv.Itoa(test_utils.TestTenantId)), test_utils.TestRootUsername, rootToken, http.StatusOK},
```
hcgatewood: Use `%d` instead of strconv

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -76,25 +86,26 @@ func TestAuthMiddleware(t *testing.T) {
...
+		{"POST", fmt.Sprintf("%s%s%s%s", urlPrefix, TenantRootPathV1, obsidian.UrlSep, strconv.Itoa(test_utils.TestTenantId)), test_utils.TestRootUsername, rootToken, http.StatusOK},
 		{"GET", fmt.Sprintf("%s%s", urlPrefix, "/malformed/url"), test_utils.TestRootUsername, rootToken, http.StatusOK},
 		{"PUT", fmt.Sprintf("%s%s", urlPrefix, "/malformed/url"), test_utils.TestRootUsername, rootToken, http.StatusOK},
-		{"GET", fmt.Sprintf("%s%s", urlPrefix, tenantsh.TenantInfoURL), test_utils.TestRootUsername, rootToken, http.StatusOK},
-		{"POST", fmt.Sprintf("%s%s", urlPrefix, tenantsh.TenantInfoURL), test_utils.TestRootUsername, rootToken, http.StatusOK},
 
 		// Test non-admin user who has read access to all URI endpoints and networks, read/write access to WRITE_TEST_NETWORK_ID, and no read/write access to specific tenants
```
hcgatewood: Can you also add a test for a user who has *only* tenant-based access, and make sure that he can (a) access networks under that tenant, and (b) not access networks outside that tenant

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -176,28 +177,35 @@ func loginHandler(c echo.Context) error {
...
-		resourceProto := &protos.Resource{
-			Effect:       matchEffect(resource.Effect),
-			Action:       matchAction(resource.Action),
-			ResourceType: matchResourceType(resource.ResourceType),
-			Resource:     resource.Resource,
+		resourceProto := &protos.PolicyResource{
+			Effect: matchEffect(resource.Effect),
+			Action: matchAction(resource.Action),
+		}
+		switch resource.ResourceType {
```
hcgatewood: Can you also do a "matchResource" function here, instead of the switch block?

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -176,28 +177,35 @@ func loginHandler(c echo.Context) error {
...
-	case protos.ResourceType_TENANT_ID.String():
-		return protos.ResourceType_TENANT_ID
-	default:
-		return protos.ResourceType_URI
+func convertTenantResourceIDs(ids []string) []int64 {
+	var intIDs []int64
+	for _, i := range ids {
+		j, err := strconv.ParseInt(i, 10, 64)
+		if err != nil {
+			panic(err)
```
hcgatewood: No need to panic here -- just bubble up an error

orc8r/cloud/go/services/certifier/obsidian/models/swagger.v1.yml
```diff
@@ -206,31 +210,43 @@ definitions:
...
+    description: |
+      The resource to which the user is either denied or allowed access to read/write.
+      If the resource is of type URI, the path field should be filled in. If the resource is of
+      type NETWORK_ID or TENANT_ID, the resourceIDs field should be filled in.
     type: object
     properties:
       effect:
         type: string
-        enum: [DENY, ALLOW]
+        enum: [ DENY, ALLOW ]
```
hcgatewood: Fix yamllints

orc8r/cloud/go/services/certifier/protos/certifier.proto
```diff
@@ -65,22 +65,46 @@ enum Action {
...
+  // resource_id is the network/tenant to which the request is scoped, if applicable
+  oneof resource_id {
+    string network_id = 3;
+    int64 tenant_id = 4;
+  }
 }
 
-message Resource {
+// PolicyResource is the resource that is used in the policy
+message PolicyResource {
```
hcgatewood: I personally like leaving this as `Resource` rather than `PolicyResource` -- I was initially confused what the "Policy" part meant

orc8r/cloud/go/services/certifier/protos/certifier.proto
```diff
@@ -65,22 +65,46 @@ enum Action {
...
+}
+
+message TenantResource {
+  repeated int64 tenants = 1;
 }
 
 message ResourceList {
-  repeated Resource resources = 4;
+  repeated PolicyResource resources = 1;
 }
```
hcgatewood: Hm actually let's just get rid of this message definition altogether, and use `repeated Resource resources = 2;` below. That way we can avoid the 1-2 weird things that result in the Go code

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -712,51 +714,95 @@ func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.To
...
 		return certprotos.Effect_UNKNOWN, status.Errorf(codes.Internal, "failed to get policy from db, checking next policy: %v", err)
 	}
 	effect := certprotos.Effect_UNKNOWN
+
+	// Networks are registered with tenants, hence any tenant scoped policies
+	// have additional policies for their networks.
+	tenantNetworkResource, err := getTenantPolicyNetworkResources(ctx, policy)
+	policy.Resources.Resources = append(policy.Resources.Resources, tenantNetworkResource...)
+
 	for _, policyResource := range policy.Resources.Resources {
```
hcgatewood: Generally "policyResource" -> "resource" if you're okay with it, cause the former is a bit too verbose / we end up putting "policy" everywhere

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -712,51 +714,95 @@ func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.To
 	return &certprotos.GetPolicyDecisionResponse{Effect: finalEffect}, nil
 }
 
-func (srv *CertifierServer) getPolicyDecisionFromToken(token string, reqResource *certprotos.Resource) (certprotos.Effect, error) {
+func (srv *CertifierServer) getPolicyDecisionFromToken(ctx context.Context, token string, reqResource *certprotos.RequestResource) (certprotos.Effect, error) {
 	policy, err := srv.store.GetPolicy(token)
 	if err != nil {
 		return certprotos.Effect_UNKNOWN, status.Errorf(codes.Internal, "failed to get policy from db, checking next policy: %v", err)
```
hcgatewood: Generally: thought it over, and I think on any unexpected errs we should default to "deny" -- cause e.g. we don't want a transient DB outage to allow unexpected access to a resource. That is, availability breakage is preferred over security breakage

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -712,51 +714,95 @@ func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.To
...
 
 	return effect, nil
 }
 
-// checkResource checks if the requested resource is authorized by the policy
-func checkResource(reqResource, policyResource *certprotos.Resource) certprotos.Effect {
-	// The policy does not handle this case, so return UNKNOWN.
-	if reqResource.ResourceType != policyResource.ResourceType {
-		return certprotos.Effect_UNKNOWN
+func getTenantPolicyNetworkResources(ctx context.Context, policy *certprotos.Policy) ([]*certprotos.PolicyResource, error) {
```
hcgatewood: Rather than returning a list of resources, just return 1 resource with the full (de-duplicated) list of networks

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -712,51 +714,95 @@ func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.To
...
+		policyResourceType = constants.TenantID
+	}
+	return policyResourceType
+}
+
+// checkResourceType checks if the policy type applies to the requested type
+func checkResourceType(reqResource *certprotos.RequestResource, policyResource *certprotos.PolicyResource) bool {
+	reqResourceType := getRequestResourceType(reqResource)
+	policyResourceType := getPolicyResourceType(policyResource)
+	return reqResource.GetResourceId() != nil && policyResourceType != constants.Path && reqResourceType != policyResourceType
```
hcgatewood: Why `!= constants.Path` ?

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -712,51 +714,95 @@ func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.To
...
+	case *certprotos.PolicyResource_Network:
+		policyResourceType = constants.NetworkID
+	case *certprotos.PolicyResource_Tenant:
+		policyResourceType = constants.TenantID
+	}
+	return policyResourceType
+}
+
+// checkResourceType checks if the policy type applies to the requested type
+func checkResourceType(reqResource *certprotos.RequestResource, policyResource *certprotos.PolicyResource) bool {
```
hcgatewood: For all these "check x", it's unclear what the true/false mean. Consider renaming them to be something like `checkResourceTypesMatch`, `doResourceTypesMatch`, or even `validateResourceType`

For all cases, the docstring should have something to the effect of "returns true if x, else false"

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -768,6 +814,48 @@ func checkAction(reqResource *certprotos.Resource, policyResource *certprotos.Re
...
+			if reqID == policyID {
+				matched = true
+			}
+		}
+	}
+
+	if finalEffect != certprotos.Effect_DENY && matched {
+		finalEffect = policyResource.Effect
+	}
+	return finalEffect
```
hcgatewood: Why the "matched" var etc? Is a little confusing / unclear why not just do an early return like for the path-based resources

hcgatewood (changes requested):
Looking real good. Couple small comments, couple discussion points. Maybe easiest to talk about the discussion points in Friday's 1-1, up to you!

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
christinewang5: The reason I just put in the switch statement is bc I don't think go can handle returning a generic type of `PolicyResource` in a function, so I just set the Resource field directly in the switch statement, though correct me if I'm wrong.

orc8r/cloud/go/services/certifier/protos/certifier.proto
christinewang5: Hmm just thought that adding Policy in front can differentiate between the two Resources: one is the resource that is requested and the other is how the resource is defined in the policy.
I think it is slightly confusing to use the term resource everywhere too.
But let me know if you feel strongly, definitely not a big deal to change. :)

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
christinewang5: Nvm I created `setResource` helper function to set it directly

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -692,12 +694,12 @@ func (srv *CertifierServer) getPolicyFromToken(token string) (*certprotos.Policy
 	return policy, nil
 }
 
-func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.TokenList, getPDReq *certprotos.GetPolicyDecisionRequest) (*certprotos.GetPolicyDecisionResponse, error) {
+func (srv *CertifierServer) getPolicyDecisionFromTokenMany(ctx context.Context, tokens *certprotos.TokenList, getPDReq *certprotos.GetPolicyDecisionRequest) (*certprotos.GetPolicyDecisionResponse, error) {
```
christinewang5, line 697: Do we still need this? I think initially when I build it out, I was under the impression that a single user will have multiple tokens, though so far it's just one

orc8r/cloud/go/services/certifier/servicers/certifier.go
christinewang5: My understanding was that given a policy where the user has read access to everything and denied write access to a specific network, if the user were to make a request with a random network `test`, s/he could still read `test` based on the first resource of the policy, i.e., the requested resource type here would be `network_id_type` but because there is no policy resource matching for this type and this specific resource id, we would have to match based on the first part of the policy. So the resource type matching for a given policyResource and requestedResource would only matter when the resource is of type network or tenant, thus the `!= constants.Path`
```
{
	Effect:   certprotos.Effect_ALLOW,
	Action:   certprotos.Action_READ,
	Resource: &certprotos.PolicyResource_Path{Path: &certprotos.PathResource{Path: "**"}},
},
{
	Effect:   certprotos.Effect_DENY,
	Action:   certprotos.Action_WRITE,
	Resource: &certprotos.PolicyResource_Network{Network: &certprotos.NetworkResource{Networks: []string{WriteTestNetworkId}}},
},
```
Does it make sense for policies to be handled this way? I also thought of the alternatives, like if we didn't do it this way, to achieve the same effect as the above policy, we have to explicitly grant read access to all networks by listing them out

orc8r/cloud/go/services/certifier/servicers/certifier.go
christinewang5: I also thought about this... if had something like
```
{
	Effect:   certprotos.Effect_ALLOW,
	Action:   certprotos.Action_WRITE,
	Resource: &certprotos.PolicyResource_Tenant{Tenant: &certprotos.TenantResource{Tenant: "0"}},
},
{
	Effect:   certprotos.Effect_DENY,
	Action:   certprotos.Action_WRITE,
	Resource: &certprotos.PolicyResource_Tenant{Tenant: &certprotos.TenantResource{Tenant: "1"}},
},
```
Then we can't just lump them in one list of networks because we want to apply different effects on them. All of tenant 0's networks need to be granted write access and all of tenant 1's networks need to be denied write access. This is looping through all tenant policies and getting the networks for each

I'll separate the inner for-loop for clarity though, maybe putting it all together is confusing

orc8r/cloud/go/services/certifier/servicers/certifier.go
christinewang5: Yup I agree, removing it 👍

orc8r/cloud/go/services/certifier/servicers/certifier.go
christinewang5: Yeah this makes sense!

orc8r/cloud/go/obsidian/access/rest_middleware.go
```diff
@@ -125,15 +127,27 @@ func TokenMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
...
 		}
-		pd, err := certifier.GetPolicyDecision(req.Context(), getPDReq)
+		switch resourceType {
+		case constants.NetworkID:
+			getPDReq.Request.ResourceId = &certprotos.Request_NetworkId{NetworkId: resourceVal}
+		case constants.TenantID:
+			id, err := strconv.ParseInt(resourceVal, 10, 64)
+			if err != nil {
+				obsidian.MakeHTTPError(err, http.StatusInternalServerError)
+			}
```
hcgatewood: This is probably a 400 [0] since the client is passing a bad argument

[0] https://devdocs.io/http/status/400

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -76,25 +77,44 @@ func TestAuthMiddleware(t *testing.T) {
 		expected int
 	}{
 		// Test admin user
-		{"GET", fmt.Sprintf("%s%s%s%s", urlPrefix, RegisterNetworkV1, obsidian.UrlSep, WRITE_TEST_NETWORK_ID), test_utils.TestRootUsername, rootToken, http.StatusOK},
-		{"PUT", fmt.Sprintf("%s%s%s%s", urlPrefix, RegisterNetworkV1, obsidian.UrlSep, TEST_NETWORK_ID), test_utils.TestRootUsername, rootToken, http.StatusOK},
 		{"GET", fmt.Sprintf("%s%s", urlPrefix, RegisterNetworkV1), test_utils.TestRootUsername, rootToken, http.StatusOK},
 		{"POST", fmt.Sprintf("%s%s", urlPrefix, RegisterNetworkV1), test_utils.TestRootUsername, rootToken, http.StatusOK},
-		{"GET", fmt.Sprintf("%s%s", urlPrefix, RegisterNetworkV1), test_utils.TestRootUsername, rootToken, http.StatusOK},
+		{"GET", fmt.Sprintf("%s%s%s%s", urlPrefix, RegisterNetworkV1, obsidian.UrlSep, TEST_NETWORK_ID), test_utils.TestRootUsername, rootToken, http.StatusOK},
```
hcgatewood: Small nit: for the test inputs, just use a bare `/` instead of the UrlSep constant, that way we can validate that `/` works as expected. Also slightly easier to ready

orc8r/cloud/go/obsidian/access/tests/rest_middleware_test.go
```diff
@@ -76,25 +77,44 @@ func TestAuthMiddleware(t *testing.T) {
...
 		{"PUT", fmt.Sprintf("%s%s%s%s", urlPrefix, RegisterNetworkV1, obsidian.UrlSep, WRITE_TEST_NETWORK_ID), test_utils.TestUsername, userToken, http.StatusForbidden},
-		{"GET", fmt.Sprintf("%s%s", urlPrefix, RegisterNetworkV1), test_utils.TestUsername, userToken, http.StatusOK},
-		{"POST", fmt.Sprintf("%s%s", urlPrefix, RegisterNetworkV1), test_utils.TestUsername, userToken, http.StatusForbidden},
-		{"GET", fmt.Sprintf("%s%s", urlPrefix, tenantsh.TenantInfoURL), test_utils.TestUsername, userToken, http.StatusOK},
-		{"POST", fmt.Sprintf("%s%s", urlPrefix, tenantsh.TenantInfoURL), test_utils.TestUsername, userToken, http.StatusForbidden},
+		// User does not have write access to any tenants
+		{"GET", fmt.Sprintf("%s%s%s%d", urlPrefix, TenantRootPathV1, obsidian.UrlSep, test_utils.TestTenantId), test_utils.TestUsername, userToken, http.StatusOK},
+		{"POST", fmt.Sprintf("%s%s%s%d", urlPrefix, TenantRootPathV1, obsidian.UrlSep, test_utils.TestTenantId), test_utils.TestUsername, userToken, http.StatusForbidden},
+
+		// Test non-admin user who only has tenant-based access
```
hcgatewood, line 103: 💯💯

orc8r/cloud/go/services/certifier/constants/constants.go
```diff
@@ -10,3 +10,11 @@ const (
...
 	PolicyType = "policy"
 )
+
+type ResourceType int
+
+const (
+	Path ResourceType = iota
+	NetworkID
+	TenantID
+)
```
hcgatewood: I would almost make these strings, rather than iota, for easier debugging

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -172,32 +177,34 @@ func loginHandler(c echo.Context) error {
...
+			Effect: matchEffect(resource.Effect),
+			Action: matchAction(resource.Action),
+		}
+		if err := setResource(resource, resourceProto); err != nil {
+			return nil, err
 		}
 		resourceList[i] = resourceProto
 	}
-	return &protos.ResourceList{Resources: resourceList}
+	return resourceList, nil
```
hcgatewood: We're using "resource" and "resourceList" but returning "policy" and "policies". Think should all be the latter (?)

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -221,3 +228,19 @@ func matchAction(rawAction string) protos.Action {
 		return protos.Action_NONE
 	}
 }
+
+func setResource(policyModel *models.Policy, policyProto *protos.Policy) error {
+	tenantIDs, err := convertTenantResourceIDs(policyModel.ResourceIDs)
+	if err != nil {
+		return err
+	}
```
hcgatewood: Why do this early? Maybe only do it under the corresponding case

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -221,3 +228,19 @@ func matchAction(rawAction string) protos.Action {
 		return protos.Action_NONE
 	}
 }
+
+func setResource(policyModel *models.Policy, policyProto *protos.Policy) error {
```
hcgatewood, line 234: Can you add a docstring for this, is unclear immediately what it's doing

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -221,3 +228,19 @@ func matchAction(rawAction string) protos.Action {
...
+	tenantIDs, err := convertTenantResourceIDs(policyModel.ResourceIDs)
+	if err != nil {
+		return err
+	}
+	switch policyModel.ResourceType {
+	case models.PolicyResourceTypeNETWORKID:
+		(*policyProto).Resource = &protos.Policy_Network{Network: &protos.NetworkResource{Networks: policyModel.ResourceIDs}}
```
hcgatewood: Does this work?

```suggestion
		policyProto.Resource = &protos.Policy_Network{Network: &protos.NetworkResource{Networks: policyModel.ResourceIDs}}
```

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -712,60 +718,166 @@ func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.To
...
+
+	for _, policy := range policyList.Policies {
+		actionEffect := isActionAuthorized(req, policy)
+		resourceEffect := isResourceAuthorized(req, policy)
 		// The effects from both action and resource should match
-		// for the effect of the policy to apply to the request.
+		// for the effect of the policyList to apply to the request.
 		if actionEffect != certprotos.Effect_UNKNOWN &&
 			resourceEffect != certprotos.Effect_UNKNOWN &&
 			actionEffect == resourceEffect {
```
hcgatewood: Split this boolean into two named sub-booleans. That way the comment probably won't be necessary. Eg

```golang
notBothUnknown := ...
equal := ...
if equal && notBothUnknown { ...
```

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -712,60 +718,166 @@ func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.To
...
+	// have additional policies for their networks.
+	tenantNetworkResource, err := getTenantPolicyNetworkResourceMany(ctx, policyList)
+	if err != nil {
+		return effect, nil
+	}
+	policyList.Policies = append(policyList.Policies, tenantNetworkResource...)
+
+	for _, policy := range policyList.Policies {
+		actionEffect := isActionAuthorized(req, policy)
+		resourceEffect := isResourceAuthorized(req, policy)
```
hcgatewood: `is` generally implies a boolean. Instead, let's call these `getActionAuthorization` and `getResourceAuthorization`

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -712,60 +718,166 @@ func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.To
...
-		// for the effect of the policy to apply to the request.
+		// for the effect of the policyList to apply to the request.
 		if actionEffect != certprotos.Effect_UNKNOWN &&
 			resourceEffect != certprotos.Effect_UNKNOWN &&
 			actionEffect == resourceEffect {
 			effect = actionEffect
 		}
+
+		if actionEffect == certprotos.Effect_DENY &&
+			resourceEffect == certprotos.Effect_DENY {
```
hcgatewood: 1. Keep this on a single line for better skim-ability
2. I think this needs a comment. It's not immediately obvious why we would need these to *both* be "deny"

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -712,60 +718,166 @@ func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.To
...
 	// The policy does not handle this case, so return UNKNOWN.
-	if reqResource.ResourceType != policyResource.ResourceType {
+	if !doResourceTypesMatch(req, policy) {
 		return certprotos.Effect_UNKNOWN
 	}
-	if policyResource.Action == certprotos.Action_WRITE {
-		return policyResource.Effect
+
+	finalEffect := certprotos.Effect_UNKNOWN
+	if path := policy.GetPath(); path != nil {
```
hcgatewood, line 856: "GetPath" will return an empty string if not found, I believe. Resolved offline

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -712,60 +718,166 @@ func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.To
...
-	if policyResource.Action == certprotos.Action_READ && reqResource.Action == certprotos.Action_READ {
-		return policyResource.Effect
+
+	reqType := getRequestType(req)
+	policyType := getPolicyType(policy)
+	if reqType == constants.NetworkID && policyType == constants.NetworkID {
+		reqID := req.GetNetworkId()
+		policyIDs := policy.GetNetwork().Networks
+		for _, policyID := range policyIDs {
+			if finalEffect != certprotos.Effect_DENY && reqID == policyID {
```
hcgatewood, line 868: Rather than checking if == deny on every loop, just return early when you hit a deny

orc8r/cloud/go/services/certifier/servicers/certifier.go
```diff
@@ -712,60 +718,166 @@ func (srv *CertifierServer) getPolicyDecisionFromTokenMany(tokens *certprotos.To
...
 	return certprotos.Effect_UNKNOWN
 }
 
-// checkAction checks if the requested action is authorized by the policy
-func checkAction(reqResource *certprotos.Resource, policyResource *certprotos.Resource) certprotos.Effect {
+// isResourceAuthorized checks if the requested resource (either a path, network, or tenant) is authorized by the policy
+// Returns the effect if the policy's path matches the requested path and if the policy's tenant/network ID matches that
+// of the requested when applicable
+// Otherwise, returns an unknown effect, which implies that the policy does not apply to this requested resource
+func isResourceAuthorized(req *certprotos.Request, policy *certprotos.Policy) certprotos.Effect {
```
hcgatewood: Unless I'm mistaken, only 1 of these 3 blocks (path auth, network auth, tenant auth) will occur for any call to this fn. If that's an accurate description, then let's make that clear in the function body by using a switch statement or early returns

hcgatewood (approved):
Lgtm! Couple style/readability comments -- msg me when ready to merge

orc8r/cloud/go/services/certifier/servicers/certifier.go
christinewang5: Was looking at the `GetPath` function in `certifier.pb.go`, I think it does return nil. Also because we were using `oneof` operator in the proto, I believe it is the best way to get the path. The only other way might be `policy.Resource.Path` but because `policy.Resource` resolves to the strange interface function, not sure how to get the path from that.

hcgatewood: You're totally right. Thanks for looking into this and getting to a thorough understanding for both of us 🙂

PR #10778, opened 2021-12-10
fix(orc8r): Upgraded EKS version to 1.21

ShubhamTatvamasi:

## Summary

We already have the same version set on `variables.tf`:
https://github.com/magma/magma/blob/master/orc8r/cloud/deploy/terraform/orc8r-aws/variables.tf#L82

## Additional Information


hcgatewood (approved):
Lgtm but cc @arunuke for input

arunuke (commented):
Although this would work too, we addressed this in another PR (https://github.com/magma/magma/pull/10212) where the default version is set to 1.21. The cloudstrapper master deployment would automatically pick 1.21

hcgatewood (approved):

Issue #10319, opened 2021-11-15
Rename C++ headers from .h to .hpp extension


electronjoe:
Magma repository currently uses `.h` extension for all c **AND** c++ headers.

Because our code base contains mixed c and c++ - mingled by directory and in compilation units - this lack of naming convention differentiation causes several fundamental problems.

A: Static analysis tooling cannot know if a header is intended to be a CPP header or not, C headers held to CPP linting rules generate many linting findings that cannot be filtered easily.
B: Because we `#include` C headers from CPP, and even CPP headers from C -> and this requires e.g. `extern "C"` and `ifdef __cplusplus` on C++ side to handle the complexities - it is very hard to do this correctly as a human author of code without at least clear differentiation between header file types.  I have found numerous bugs in Magma where we compile by luck (e.g. ifdef and cycles in includes saving us) CPP headers that include C headers without `extern "C"` for example. With proper file naming, we could even turn up CI checks that all C `#includes` are guarded by `extern "C"`, for example. Ditto for `ifdef _cplusplus`.

PR #10786, opened 2021-12-10
fix(agw): migrate all orc8r/gateway/c/ cpp headers to hpp extension

electronjoe:
# Summary

This is PR 1 of ~3 in which we address Issue #10319 and migrate ~all Magma cpp headers to `.hpp` extension.

This PR was achieved in three parts.

## Part One: Automated renaming in `/orc8r/gateway/c/`

This PR was achieved by execution of [this Golang script](https://gist.github.com/electronjoe/e88fa09b27d5c104aed5483681e7c410) as follows, from within a GH Codespace / Devcontainer:

```
go run hppify.go -root /workspaces/magma/ -target /workspaces/magma/orc8r/gateway/c/
```

## Part Two: manual fixup of `session_manager`

Because Session Manager has not been fixed up with full include paths relative to repo root (an outstanding task) - `hppify.go` is unable to precisely correct the renamed `.h` -> `.hpp` `#include` directives. So I fixed them by hand.

## Part Three: correct CMake build include copies of `*.h`

Various CMake build steps in Magma unfortunately rely on copying include directories to various build directories.  These often copy all `*.h` headers, and now need a broader copy directive.

# Implications to outstanding PRs

I'm unaware of any major PRs to the `orc8r/gateway/c/` sub-directories, though this PR can cause isolated disruption via header `#include` diffs. It is anticipated that the move git semantics will enable clean rebasing for any outstanding PRs that touched the now-renamed `.hpp` files.


themarwhal (approved):
changes lgtm, are we committing hippify.go?

electronjoe:
> changes lgtm, are we committing hippify.go?

For these various automation scripts I have been electing not to commit them - because to get them into commit / review worthy shape would easily double the effort involved and they are one-time scripts (at least one-time for the basis of this migration).

I understand that lacking review and test, there is higher chance that they are missing some pieces of the migration puzzle, but given the re-use of this script is so low (N=1) I really think it's pragmatically better not to get them into commit-worthy shape.

Up for discussion if we want to reconsider however.

themarwhal:
> > changes lgtm, are we committing hippify.go?
> 
> For these various automation scripts I have been electing not to commit them - because to get them into commit / review worthy shape would easily double the effort involved and they are one-time scripts (at least one-time for the basis of this migration).
> 
> I understand that lacking review and test, there is higher chance that they are missing some pieces of the migration puzzle, but given the re-use of this script is so low (N=1) I really think it's pragmatically better not to get them into commit-worthy shape.
> 
> Up for discussion if we want to reconsider however.

Ok thought that was the case. Just asking because it is included in the PR :)

electronjoe:
> > > changes lgtm, are we committing hippify.go?
> > 
> > 
> > For these various automation scripts I have been electing not to commit them - because to get them into commit / review worthy shape would easily double the effort involved and they are one-time scripts (at least one-time for the basis of this migration).
> > I understand that lacking review and test, there is higher chance that they are missing some pieces of the migration puzzle, but given the re-use of this script is so low (N=1) I really think it's pragmatically better not to get them into commit-worthy shape.
> > Up for discussion if we want to reconsider however.
> 
> Ok thought that was the case. Just asking because it is included in the PR :)

Whoops!

electronjoe:
My concern with this PR is that I have come to realize there are collections of header files backed by multiple and dissimilarly named cpp files - which would not be detected by my go script.  So my .h->.hpp will miss a subset of known C++ header files... Not sure if a partial solution is advisable. The complexity to detect CPP headers by symbol contents might not be awful, I would have to build a ~symbol table myself (but it would not need to be comprehensively correct).  Thoughts @themarwhal ?

electronjoe:
Closing for now as Bazel effort has been achieved without this - we want to revisit this goal after Bazel MME work is ~in place.

PR #10803, opened 2021-12-10
feat(feg): SBI print HTTP request and response logs

### LINKED ISSUE #10598: [sbi] Print request and response in logs
Currently our FEG services print the messages that are sent thorugh the 3gpp interface. For example below you can see an example of s6a where we can sdee protos being printed and AVP (AIR and AIA)

We should do the same for N7 and other sbi based interfaces, where the N7 client should print the request is making and the response it receives when we enable log level V=2 

Here are two examples of how this is done at s6a
[here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fs6a_proxy%2Fservicers%2Ful.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=73&start_col=20&end_row=73&end_col=20)
[here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Fs6a_proxy%2Fservicers%2Ful.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=124&start_col=31&end_row=124&end_col=31)

```
s6a_proxy        | I1201 00:43:36.444585       1 logcodec.go:46] Received: *protos.AuthenticationInformationRequest: {
s6a_proxy        |      "user_name": "999991234567810",
s6a_proxy        |      "visited_plmn": "APEQ",
s6a_proxy        |      "num_requested_eutran_vectors": 1,
s6a_proxy        |      "immediate_response_preferred": false,
s6a_proxy        |      "resync_info": null,
s6a_proxy        |      "num_requested_utran_geran_vectors": 0,
s6a_proxy        |      "utran_geran_resync_info": null,
s6a_proxy        |      "feature_list_id_2": null
s6a_proxy        | }
s6a_proxy        | I1201 00:43:36.444618       1 ai.go:68] Sending S6a AIR message
s6a_proxy        | Authentication-Information-Request (AIR)
s6a_proxy        | {Code:318,Flags:0xc0,Version:0x1,Length:208,ApplicationId:16777251,HopByHopId:0xce8e944c,EndToEndId:0xd6da4f7e}
s6a_proxy        |      Session-Id {Code:263,Flags:0x40,Length:60,VendorId:0,Value:UTF8String{string-s6a;3130621136473049474;3465872240;98295ba0},Padding:2}
s6a_proxy        |      Origin-Host {Code:264,Flags:0x40,Length:16,VendorId:0,Value:DiameterIdentity{string},Padding:2}
s6a_proxy        |      Origin-Realm {Code:296,Flags:0x40,Length:16,VendorId:0,Value:DiameterIdentity{string},Padding:2}
s6a_proxy        |      User-Name {Code:1,Flags:0x40,Length:24,VendorId:0,Value:UTF8String{999991234567810},Padding:1}
s6a_proxy        |      Auth-Session-State {Code:277,Flags:0x40,Length:12,VendorId:0,Value:Enumerated{1}}
s6a_proxy        |      Visited-PLMN-Id {Code:1407,Flags:0xc0,Length:16,VendorId:10415,Value:OctetString{0x00f110},Padding:1}
s6a_proxy        |      Requested-EUTRAN-Authentication-Info {Code:1408,Flags:0xc0,Length:44,VendorId:10415,Value:Grouped{
s6a_proxy        |              Number-Of-Requested-Vectors {Code:1410,Flags:0xc0,Length:16,VendorId:10415,Value:Unsigned32{1}},
s6a_proxy        |              Immediate-Response-Preferred {Code:1412,Flags:0xc0,Length:16,VendorId:10415,Value:Unsigned32{0}},
s6a_proxy        |      }}
s6a_proxy        | I1201 00:43:36.445032       1 ai.go:99] Received S6a AIA message:
s6a_proxy        | Authentication-Information-Answer (AIA)
s6a_proxy        | {Code:318,Flags:0x40,Version:0x1,Length:344,ApplicationId:16777251,HopByHopId:0xce8e944c,EndToEndId:0xd6da4f7e}
s6a_proxy        |      Session-Id {Code:263,Flags:0x40,Length:60,VendorId:0,Value:UTF8String{string-s6a;3130621136473049474;3465872240;98295ba0},Padding:2}
s6a_proxy        |      Vendor-Specific-Application-Id {Code:260,Flags:0x40,Length:32,VendorId:0,Value:Grouped{
s6a_proxy        |              Vendor-Id {Code:266,Flags:0x40,Length:12,VendorId:0,Value:Unsigned32{10415}},
s6a_proxy        |              Auth-Application-Id {Code:258,Flags:0x40,Length:12,VendorId:0,Value:Unsigned32{16777251}},
s6a_proxy        |      }}
s6a_proxy        |      Result-Code {Code:268,Flags:0x40,Length:12,VendorId:0,Value:Unsigned32{2001}}
s6a_proxy        |      Auth-Session-State {Code:277,Flags:0x40,Length:12,VendorId:0,Value:Enumerated{1}}
s6a_proxy        |      Origin-Host {Code:264,Flags:0x40,Length:28,VendorId:0,Value:DiameterIdentity{tvm-vhss.magma.com},Padding:2}
s6a_proxy        |      Origin-Realm {Code:296,Flags:0x40,Length:20,VendorId:0,Value:DiameterIdentity{magma.com},Padding:3}
s6a_proxy        |      Authentication-Info {Code:1413,Flags:0xc0,Length:160,VendorId:10415,Value:Grouped{
s6a_proxy        |              E-UTRAN-Vector {Code:1414,Flags:0xc0,Length:148,VendorId:10415,Value:Grouped{
s6a_proxy        |                      Item-Number {Code:1419,Flags:0xc0,Length:16,VendorId:10415,Value:Unsigned32{0}},
s6a_proxy        |                      RAND {Code:1447,Flags:0xc0,Length:28,VendorId:10415,Value:OctetString{0xfedcba9876543210fedcba9876543210},Padding:0},
s6a_proxy        |                      XRES {Code:1448,Flags:0xc0,Length:20,VendorId:10415,Value:OctetString{0xfe85aa11ae14e4ec},Padding:0},
s6a_proxy        |                      AUTN {Code:1449,Flags:0xc0,Length:28,VendorId:10415,Value:OctetString{0x80fb9d559765b9b9beff592850bb4dcd},Padding:0},
s6a_proxy        |                      KASME {Code:1450,Flags:0xc0,Length:44,VendorId:10415,Value:OctetString{0xb7224763f6c1d51d7559cead6f7e7d944370dc6836c23fc45a06141788fa0067},Padding:0},
s6a_proxy        |              }}
s6a_proxy        |      }}
s6a_proxy        | I1201 00:43:36.445328       1 logcodec.go:46] Sending: *protos.AuthenticationInformationAnswer: {
s6a_proxy        |      "error_code": "UNDEFINED",
s6a_proxy        |      "eutran_vectors": [
s6a_proxy        |              {
s6a_proxy        |                      "rand": "/ty6mHZUMhD+3LqYdlQyEA==",
s6a_proxy        |                      "xres": "/oWqEa4U5Ow=",
s6a_proxy        |                      "autn": "gPudVZdlubm+/1koULtNzQ==",
s6a_proxy        |                      "kasme": "tyJHY/bB1R11Wc6tb359lENw3Gg2wj/EWgYUF4j6AGc="
s6a_proxy        |              }
s6a_proxy        |      ],
s6a_proxy        |      "utran_vectors": [
s6a_proxy        |      ],
s6a_proxy        |      "geran_vectors": [
s6a_proxy        |      ]
s6a_proxy        | }

```

GANESH-WAVELABS:

## Summary
Generic logging mechanism for all SBI based requests, responses, notifications and notifcation responses.

- Client Requests and Responses are logged by using a HTTP logging client
- Notification requests and responses are logged by setting a logging middleware to the echo server

This is a PR for https://github.com/magma/magma/issues/10598

### DIFF
--- a/feg/gateway/sbi/sbi.go
+++ b/feg/gateway/sbi/sbi.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/golang/glog"
 	"github.com/labstack/echo/v4"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/clientcredentials"
@@ -66,7 +67,11 @@ type BaseClientWithNotifier struct {
 }
 
 func NewEchoServer() *EchoServer {
-	return &EchoServer{echo.New()}
+	e := echo.New()
+	if glog.V(2) {
+		e.Use(ServerLoggingMiddleware())
+	}
+	return &EchoServer{e}
 }
 
 func NewBaseClientWithNotifyServer(notifierConfig NotifierConfig, remoteConfig RemoteConfig) *BaseClientWithNotifier {
@@ -83,7 +88,7 @@ func NewNotifierServer(notifierConfig NotifierConfig) *NotifierServer {
 	}
 }
 
-func (s *EchoServer) StartWithhWait(addrs string) error {
+func (s *EchoServer) StartWithWait(addrs string) error {
 	addr, err := net.ResolveTCPAddr("tcp", addrs)
 	if err != nil {
 		return err
@@ -155,12 +160,18 @@ func (rc RemoteConfig) BuildHttpClient() *http.Client {
 		ClientSecret: rc.ClientSecret,
 		TokenURL:     rc.TokenUrl,
 	}
-	tokenCtxt := context.WithValue(context.Background(), oauth2.HTTPClient, &http.Client{})
+	var httpClient *http.Client
+	if glog.V(2) {
+		httpClient = NewLoggingHttpClient()
+	} else {
+		httpClient = &http.Client{}
+	}
+	tokenCtxt := context.WithValue(context.Background(), oauth2.HTTPClient, httpClient)
 	return tokenConfig.Client(tokenCtxt)
 }
 
 func (s *NotifierServer) Start() error {
-	err := s.Server.StartWithhWait(s.NotifierCfg.LocalAddr)
+	err := s.Server.StartWithWait(s.NotifierCfg.LocalAddr)
 	if err != nil {
 		return fmt.Errorf("NotifierServer could not Start: %s", err)
 	}
--- a/feg/gateway/sbi/sbi_log_middleware.go
+++ b/feg/gateway/sbi/sbi_log_middleware.go
@@ -1,0 +1,141 @@
+/*
+Copyright 2021 The Magma Authors.
+
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sbi
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/labstack/echo/v4"
+)
+
+// LoggingTransport implements the http.RoundTripper interface and logs the
+// HTTP request and response information
+type LoggingTransport struct {
+	Transport http.RoundTripper
+	Logger    SbiLogger
+}
+
+type responseGrabber struct {
+	io.Writer
+	http.ResponseWriter
+}
+
+func NewLoggingTransport() LoggingTransport {
+	return LoggingTransport{
+		Transport: http.DefaultTransport,
+		Logger:    SbiLogger{},
+	}
+}
+
+func NewLoggingHttpClient() *http.Client {
+	return &http.Client{
+		Transport: NewLoggingTransport(),
+	}
+}
+
+// RoundTrip overrides the http.DefaultTransport RoundTrip behavior and logs the HTTP request and response.
+func (lt LoggingTransport) RoundTrip(req *http.Request) (res *http.Response, err error) {
+	// Before sending
+	var reqBody []byte
+
+	if req.Body != nil {
+		reqBody, err = ioutil.ReadAll(req.Body)
+		if err != nil {
+			return
+		}
+		// reset the req.Body back to original
+		req.Body = ioutil.NopCloser(bytes.NewBuffer(reqBody))
+	}
+	lt.Logger.LogRequest(req.Method, req.URL, reqBody, req.Header)
+
+	// Make the actual request
+	start := time.Now()
+	res, err = lt.Transport.RoundTrip(req)
+	took := time.Since(start)
+	if err != nil {
+		// error responses shall be logged by caller
+		return
+	}
+
+	// After sending
+	var resBody []byte
+	if res.Body != nil {
+		resBody, err = ioutil.ReadAll(res.Body)
+		if err != nil {
+			return
+		}
+		// reset the res.Body back to original
+		res.Body = ioutil.NopCloser(bytes.NewBuffer(resBody))
+	}
+	lt.Logger.LogResponse(req.URL, res.Status, resBody, res.Header, took)
+	return
+}
+
+func (w *responseGrabber) WriteHeader(code int) {
+	w.ResponseWriter.WriteHeader(code)
+}
+
+func (w *responseGrabber) Write(b []byte) (int, error) {
+	return w.Writer.Write(b)
+}
+
+func (w *responseGrabber) Flush() {
+	w.ResponseWriter.(http.Flusher).Flush()
+}
+
+func (w *responseGrabber) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	return w.ResponseWriter.(http.Hijacker).Hijack()
+}
+
+// ServerLoggingMiddleware logs HTTP request and response whenever a request is received.
+// When logger is nil, the messages are logged to default glog verbose level 2 logger.
+func ServerLoggingMiddleware() echo.MiddlewareFunc {
+	logger := SbiLogger{}
+
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) (err error) {
+			// Request
+			var reqBody []byte
+			if c.Request().Body != nil { // Read
+				reqBody, _ = ioutil.ReadAll(c.Request().Body)
+				c.Request().Body = ioutil.NopCloser(bytes.NewBuffer(reqBody)) // Reset
+			}
+			logger.LogRequest(c.Request().Method, c.Request().URL, reqBody, c.Request().Header)
+
+			// Setting up for grabbing response
+			resBody := new(bytes.Buffer)
+			mw := io.MultiWriter(c.Response().Writer, resBody)
+			writer := &responseGrabber{Writer: mw, ResponseWriter: c.Response().Writer}
+			c.Response().Writer = writer
+
+			start := time.Now()
+			// invoke request handler
+			if err = next(c); err != nil {
+				c.Error(err)
+			}
+			latency := time.Since(start)
+			status := fmt.Sprintf("%d %s", c.Response().Status, http.StatusText(c.Response().Status))
+			logger.LogResponse(c.Request().URL, status, resBody.Bytes(), c.Response().Header(), latency)
+
+			return
+		}
+	}
+}
--- a/feg/gateway/sbi/sbi_test.go
+++ b/feg/gateway/sbi/sbi_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"magma/feg/gateway/sbi"
 	"magma/feg/gateway/sbi/mocks"
 	sbi_NpcfSMPolicyControl "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
 	sbi_CommonData "magma/feg/gateway/sbi/specs/TS29571CommonData"
@@ -44,7 +45,8 @@ func TestCreateSMPolicy(t *testing.T) {
 
 	// Create new N7 client object
 	client, err := sbi_NpcfSMPolicyControl.NewClientWithResponses(
-		fmt.Sprintf("http://%s", pcfAddr.String()))
+		fmt.Sprintf("http://%s", pcfAddr.String()),
+		sbi_NpcfSMPolicyControl.WithHTTPClient(sbi.NewLoggingHttpClient()))
 
 	require.NoError(t, err, "BaseClientWithNotifier creation failed")
 
--- a/feg/gateway/sbi/mocks/mock_pcf.go
+++ b/feg/gateway/sbi/mocks/mock_pcf.go
@@ -24,7 +24,7 @@ func NewMockPcf(localAddr string) (*MockPcf, error) {
 		policies:   make(map[string]sbi_NpcfSMPolicyControlServer.SmPolicyControl),
 	}
 	sbi_NpcfSMPolicyControlServer.RegisterHandlers(mockPcf, mockPcf)
-	err := mockPcf.StartWithhWait(localAddr)
+	err := mockPcf.StartWithWait(localAddr)
 	if err != nil {
 		return nil, err
 	}
--- a/feg/gateway/sbi/sbi_logger.go
+++ b/feg/gateway/sbi/sbi_logger.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2021 The Magma Authors.
+
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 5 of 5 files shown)


uri200:
note you can link the pr to the issue using this at right side of the PR

image

feg/gateway/sbi/sbi_test.go
```diff
@@ -90,14 +91,25 @@ func (pcf *MockPcf) PostSmPoliciesSmPolicyIdUpdate(ctx echo.Context, smPolicyId
 	return ctx.NoContent(http.StatusOK)
 }
 
+func init() {
+	// Force the logLevel to 2, for testing the logs
+	flag.Set("alsologtostderr", fmt.Sprintf("%t", true))
+	var logLevel string
+	flag.StringVar(&logLevel, "logLevel", "2", "test")
+	flag.Lookup("v").Value.Set(logLevel)
+}
```
uri200: i think this was left forgotten

uri200 (commented):
I am curious, but would it be easier to implement using [logger](https://echo.labstack.com/middleware/logger/) to print requests and [body dump](https://echo.labstack.com/middleware/body-dump/) to print responses (example of body dump [here](https://titanwolf.org/Network/Articles/Article?AID=c462169f-a60f-4015-b4f2-2eb179ba77d9))

uri200:
PEr comment in slack, lets see if we can build an object to encapsulate a sbi instance which includes client and server with loggers, stop, start, shutdown, etc. This way other future interfaces (kike n40) can use it

GANESH-WAVELABS:
@uri200 
> I am curious, but would it be easier to implement using logger to print requests and body dump to print responses (example of body dump here)

Logger middleware does not print the request body or the response body (json data). And the BodyDump middleware does not print the request url or the status code (only prints the message body). Both the middleware logs the message only after the response request handler returns, which might not provide a chronological order of logs.

uri200:
> @uri200 
> > I am curious, but would it be easier to implement using logger to print requests and body dump to print responses (example of body dump here)
> 
> Logger middleware does not print the request body or the response body (json data). And the BodyDump middleware does not print the request url or the status code (only prints the message body). Both the middleware logs the message only after the response request handler returns, which might not provide a chronological order of logs.
> 
> 

Sounds good. But now we have the logger doing the actual call. I think we need to separate them on different files, functions. The function that runs the call to the server should call a logger to log the request and then call a logger function that prints the response.

feg/gateway/sbi/sbi_test.go
GANESH-WAVELABS: This was added to force print the logs by setting the log level to 2.

GANESH-WAVELABS:
> Sounds good. But now we have the logger doing the actual call. I think we need to separate them on different files, functions. The function that runs the call to the server should call a logger to log the request and then call a logger function that prints the response.

Separated the middlware from the logger.

feg/gateway/sbi/sbi_log_middleware.go
```diff
@@ -0,0 +1,164 @@
...
+	return w.Writer.Write(b)
+}
+
+func (w *responseGrabber) Flush() {
+	w.ResponseWriter.(http.Flusher).Flush()
+}
+
+func (w *responseGrabber) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	return w.ResponseWriter.(http.Hijacker).Hijack()
+}
```
uri200, line 106: Not sure if all those functions are just wrappers or implementation of ResponseWriter methods, but if they do not add any extra logic, you don't need them.  

Since `responseGrabber` embeds those two structs, `responseGrabber` already has those methods for free. You can try to access them if you remove those functions.

```
type responseGrabber struct {
	io.Writer
	http.ResponseWriter
}
```

https://yourbasic.org/golang/inheritance-object-oriented/

feg/gateway/sbi/sbi_log_middleware.go
```diff
@@ -0,0 +1,164 @@
...
+		Transport: NewLoggingTransport(nil),
+	}
+}
+
+// ServerLoggingMiddleware logs HTTP request and response whenever a request is received.
+// When logger is nil, the messages are logged to default glog verbose level 2 logger.
+func ServerLoggingMiddleware(logger Logger) echo.MiddlewareFunc {
+	if logger == nil {
+		logger = &DefaultGlogLogger{}
+	}
```
uri200: this is weird. Let me write a comment about the logging and maybe you can refactor some of the code out of it

uri200 (changes requested):
The idea is to print those messages only when logger is set to v=2 (or more). 
So for example when docker-compose runs this 
```
envdir /var/opt/magma/envdir /var/opt/magma/bin/session_proxy -logtostderr=true -v=2
```
the logger should print the message. If v=0 it should not print the messages

Note you can easily check the logger level with `if glog.V(2)`. For example [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fdiameter%2Fconnection.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=203&start_col=0&end_row=204&end_col=0)

uri200:
Hey @GANESH-WAVELABS  do you have any update on this PR?

feg/gateway/sbi/sbi_log_middleware.go
GANESH-WAVELABS: This code is based on [BodyDumpMiddleware](https://github.com/labstack/echo/blob/4fffee2ec8a4efe5cca66cdfe17e6eeec59df60a/middleware/body_dump.go#L93).

Removing those code results in a compilation error, since bothe io.Writer and http.ResponseWriter are interfaces and not structs.

feg/gateway/sbi/sbi_log_middleware.go
GANESH-WAVELABS: Can you please elaborate on what is not correct here?

GANESH-WAVELABS:
@uri200 
> The idea is to print those messages only when logger is set to v=2 (or more). 

This is understood and the Skipper method in the Logging middleware (both client and server) will not proceed with logging related code when it returns true. For DefaultGlogLogger the Skipper method will return true when verbose level is less than 2.
```
// Skipper returns true only when the glog verbose level is set to 2 or above
func (logger *DefaultGlogLogger) Skipper() bool {
	return !bool(glog.V(2))
}
```

feg/gateway/sbi/sbi_test.go
uri200: i still don't understand this. Why do you need to print the logs? You don't have any test checking if they are working or not, so this is kind of not useful.

If you want to test it, you can create a test file for the functions itself. But lets worry about that later

feg/gateway/sbi/sbi_log_middleware.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (w *responseGrabber) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	return w.ResponseWriter.(http.Hijacker).Hijack()
+}
+
+func NewLoggingHttpClient() *http.Client {
+	return &http.Client{
+		Transport: NewLoggingTransport(nil),
+	}
+}
```
uri200: place this new function at the top next to the other new function

feg/gateway/sbi/sbi_log_middleware.go
uri200: sorry. I meant, having a nil argument to create a default logger doesn't look right. It is too convoluted? 

Why is this default logger needed. Per what i see this ServerLoggingMiddleware is only used once and its argument is nil.

feg/gateway/sbi/sbi.go
```diff
@@ -29,8 +29,10 @@ type SbiServer struct {
 }
 
 func NewSbiServer(listenAddr string) *SbiServer {
+	e := echo.New()
+	e.Use(ServerLoggingMiddleware(nil))
 	return &SbiServer{
-		Server:     echo.New(),
+		Server:     e,
```
uri200: i think all would be simpler if you were to do

```
func NewSbiServer(listenAddr string) *SbiServer {
	e := echo.New()
	if glog..v == 2 {
		e.Use(ServerLoggingMiddleware(nil))
	}
	return &SbiServer{
		Server:     e,
		ListenAddr: listenAddr,
	}
}
```

Using that you can forget about the logger. Note the log is not dynamic. Either is enabled from start ot it is not

feg/gateway/sbi/sbi_log_middleware.go
```diff
@@ -0,0 +1,164 @@
...
+)
+
+type Logger interface {
+	// Skipper returns true to skip the logging
+	Skipper() bool
+	// LogRequest logs http request relateed info before making the client request
+	LogRequest(method string, url *url.URL, reqBody []byte, header http.Header)
+	// LogResponse logs http response related info after receiving the response from the server
+	LogResponse(url *url.URL, status string, resBody []byte, header http.Header, latency time.Duration)
+}
```
uri200: why is this interface needed at all?

With the changes mentioned at the beginning, you can remove the skipper method. Then `LogRequest` and `LogResponse` can be just functions on the same `sbi_log_middlewear`

uri200 (changes requested):
I am sorry but i am having hard time understanding. I think this is overcomplicated for a simple task

I think hte problem is you should have an object caled sbi_client (same as sbi_server ) at sbi.go.  That client should have the logger on it depending on the log value or not

feg/gateway/sbi/sbi.go
```diff
@@ -83,7 +88,7 @@ func NewNotifierServer(notifierConfig NotifierConfig) *NotifierServer {
 	}
 }
 
-func (s *EchoServer) StartWithhWait(addrs string) error {
+func (s *EchoServer) StartWithWait(addrs string) error {
```
uri200, line 91: 🙏

feg/gateway/sbi/sbi.go
```diff
@@ -66,7 +67,11 @@ type BaseClientWithNotifier struct {
 }
 
 func NewEchoServer() *EchoServer {
-	return &EchoServer{echo.New()}
+	e := echo.New()
+	if glog.V(2) {
+		e.Use(ServerLoggingMiddleware())
+	}
+	return &EchoServer{e}
```
uri200, line 74: awesome! 
This is where we had to get: just one place to add the logger and forget about it

feg/gateway/sbi/sbi_log_middleware.go
```diff
@@ -0,0 +1,141 @@
...
+		resBody, err = ioutil.ReadAll(res.Body)
+		if err != nil {
+			return
+		}
+		// reset the res.Body back to original
+		res.Body = ioutil.NopCloser(bytes.NewBuffer(resBody))
+	}
+	lt.Logger.LogResponse(req.URL, res.Status, resBody, res.Header, took)
+	return
+}
```
uri200, line 90: I am concerned about this because this code is only used when logging. That means that if for any reason there is a bug there, we will only be able to see it when v==2. 

Instead of checking `glog.V(2)` outside, wouldn't it make sense to check it on the function? 
This way you just need to always add this custom logger on sbi. The logger will take care of printing or not printing.  Plus we will always be using this `RoundTrip` function and `NewLoggingHttpClient` object

Let me know if that doesn't make sense to you. Maybe I am missing something


```
func (lt LoggingTransport) RoundTrip(req *http.Request) (res *http.Response, err error) {
	// Before sending
	var reqBody []byte


	if glog.V(2) {
		  if req.Body != nil {
			  reqBody, err = ioutil.ReadAll(req.Body)
			  if err != nil {
				  return
			  }
			  // reset the req.Body back to original
			  req.Body = ioutil.NopCloser(bytes.NewBuffer(reqBody))
		  }
		  lt.Logger.LogRequest(req.Method, req.URL, reqBody, req.Header)
	}
	// Make the actual request
	start := time.Now()
	res, err = lt.Transport.RoundTrip(req)
	took := time.Since(start)
	if err != nil {
		// error responses shall be logged by caller
		return
	}

	// After sending
	var resBody []byte
	if glog.V(2) { 
		  if res.Body != nil {
			  resBody, err = ioutil.ReadAll(res.Body)
			  if err != nil {
				  return
			  }
			  // reset the res.Body back to original
			  res.Body = ioutil.NopCloser(bytes.NewBuffer(resBody))
		  }
		  lt.Logger.LogResponse(req.URL, res.Status, resBody, res.Header, took)
	}
	return
}
```

uri200 (commented):
.

feg/gateway/sbi/sbi_log_middleware.go
GANESH-WAVELABS: This is a hook over a normal Client HTTP Transport and will incur some overhead. This hook will get installed only when  the log  verbose level is set to 2.

> That means that if for any reason there is a bug there, we will only be able to see it when v==2

Even in the code proposed above, only a small portion if the code is outside the check. A bug inside the log check will also be not identified at a lower log level.

uri200 (approved):

PR #10804, opened 2021-12-10
test(mme): add unit test for paging failure for dedicated bearer

ulaskozat:

## Summary

- Add unit test for paging failure for dedicated bearer
- Fix grammer in comment
- Reduce wait_for timer to 2 seconds that looks like a good balance between no-flakiness and run time. With a lot of test cases, occasionally itti latency goes over 1 second but it is not observed to be as high as 2 seconds.

## Test Plan

Run `make test_oai OAI_TESTS='.*mme_app.*'` in magma VM 100 times.

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/test/mme_app_task/test_mme_procedures.cpp
+++ b/lte/gateway/c/core/oai/test/mme_app_task/test_mme_procedures.cpp
@@ -198,7 +198,7 @@ TEST_F(MmeAppProcedureTest, TestInitialUeMessageFaultyNasMsg) {
 
   MME_APP_EXPECT_CALLS(1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1);
 
-  // Construction and sending Initial Attach Request to mme_app mimicing S1AP
+  // Constructing and sending Initial Attach Request to mme_app mimicing S1AP
   // The following buffer just includes an attach request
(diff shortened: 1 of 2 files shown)


lte/gateway/c/core/oai/test/mme_app_task/test_mme_procedures.cpp
```diff
@@ -2886,5 +2887,140 @@ TEST_F(MmeAppProcedureTest, TestTauRejDueToInvalidTac) {
...
+  EXPECT_EQ(mme_state_p->nb_ue_connected, 0);
+  EXPECT_EQ(mme_state_p->nb_default_eps_bearers, 1);
+  EXPECT_EQ(mme_state_p->nb_ue_idle, 1);
+  EXPECT_EQ(mme_state_p->nb_s1u_bearers, 0);
+
+  // Trigger paging via bearer request in control plane
+  // Send activate dedicated bearer request mimicing SPGW
+  send_s11_create_bearer_req();
+  EXPECT_CALL(*s1ap_handler, s1ap_handle_paging_request())
+      .Times(2)
```
ssanadhya:
```suggestion
      .Times(MAX_PAGING_RETRY_COUNT + 1)
```

ssanadhya (approved):
LGTM, with nit comment

lte/gateway/c/core/oai/test/mme_app_task/test_mme_procedures.cpp
ulaskozat: updated.

PR #10819, opened 2021-12-11
feat(nms): Update the device_class to config the Enodeb Nova227.

hakeen0618:
Signed-off-by:Hanhufeng <hanhufeng@baicells.com>

## Summary

Add the new enodeb type nova227 for magma product.
## Test Plan

I already add the RTS unittest in other PR 10472.
## Additional Information


andreilee (changes requested):
There are some generated files that are missing

See https://github.com/magma/magma/blob/master/nms/README.md?plain=1#L166-L169

hakeen0618:
> There are some generated files that are missing
> 
> See https://github.com/magma/magma/blob/master/nms/README.md?plain=1#L166-L169

I already update and add the temp file for the nms and orc8r.
magma0107

PR #10822, opened 2021-12-12
fix(amf): set default slice service type if ue doen't send slice service type.

Lkishor123:
Signed-off-by: LKishor123 <laawanya.kishor@wavelabs.ai>

## Summary
Right now amf returns empty snssai field, if ue doesn't send a snssai. This fix will set a default slice service type if ue doen't send slice service type.

## Test Plan
1. Unit Test
sit_2

2. Spirent

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
+++ b/lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
@@ -453,7 +453,12 @@ int amf_smf_process_pdu_session_packet(
         return rc;
       }
 
-      smf_ctx->sst = msg->nssai.sst;
+      // NSSAI
+      if (msg->nssai.sst) {
+        smf_ctx->sst = msg->nssai.sst;
+      } else {
+        amf_get_default_sst_config(&(smf_ctx->sst));
+      }
       if (msg->nssai.sd[0]) {
         memcpy(smf_ctx->sd, msg->nssai.sd, SD_LENGTH);
       }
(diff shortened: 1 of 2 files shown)


lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
```diff
@@ -453,7 +453,12 @@ int amf_smf_process_pdu_session_packet(
         return rc;
       }
 
-      smf_ctx->sst = msg->nssai.sst;
+      // NSSAI
+      if (msg->nssai.sst) {
+        smf_ctx->sst = msg->nssai.sst;
+      } else {
+        smf_ctx->sst = static_cast<uint8_t>(eMBB);
+      }
```
ulaskozat: you should be getting these values from AMF configurations. It does not make sense to respond sst value with whatever the UE is asking as we do not support URLLC, MIoT, etc.

Lkishor123: @ulaskozat, I fixed the second part: when the ue doesn't send nssai IE, I am using user configured sst value. But for the first part: when UE sends nssai IE, there is ongoing work for slicing feature support which will cover validation of nssai and assignment.

Lkishor123:
UERANSIM testing: when ue doesn't send the snssai IE:

sit_snssai

ulaskozat:
Please do not keep squashing all commits into one. I want to see the changes in each commit separately to review faster.

ulaskozat:
There are a few things that needs to be clarified. What you are storing in the SMF context should contain UE requested slice type and AMF assigned one as separate entries no? Otherwise we lose context about what is requested vs. what is assigned. If UE is sending a slice type, based on this PR, are we simply accepting it and returning back as the assigned one? We should not be doing this and we do not need to wait for slicing support for it. cc @panyogesh

panyogesh:
> There are a few things that needs to be clarified. What you are storing in the SMF context should contain UE requested slice type and AMF assigned one as separate entries no? Otherwise we lose context about what is requested vs. what is assigned. If UE is sending a slice type, based on this PR, are we simply accepting it and returning back as the assigned one? We should not be doing this and we do not need to wait for slicing support for it. cc @panyogesh

Agreed on comments. There are some clean up planned for SMF context(#10853). Will ask Ganesh G to  incorporate the change in that PR.

Issue #10827, opened 2021-12-13
Introduce semantic versions in Sentry to allow error resolution


sebathomas:
# Introduce semantic versions in Sentry to allow error resolution

## Problem

Sentry is an error monitoring solution that collects crashes and error logs from running AGW services. After these errors have been fixed, it is possible to resolve them within the Sentry UI for a given version. Sentry will always consider the version if new reports of the same error come in, and only show the error again as a regression if it occurs for a higher version than the one that resolved it.

For the resolve / regression flow to work, Sentry needs to know the correct order of the versions. Unfortunately, it cannot parse the current versioning scheme of Magma so the versions are ordered by the timestamp when they were first seen. This means all resolved errors that were not backported will be reported as a regression as soon as a new release from an older branch (currently 1.6.1) starts talking to Sentry.

## Solution

Sentry supports semantic release versions if they are given in the [correct format](https://docs.sentry.io/platforms/python/configuration/releases/#bind-the-version). For magma, we propose changing the release version given in the Sentry SDK to the following format:
```bash
VERSION="1.7.0"; echo "magma@${VERSION}.$(git rev-list --count HEAD)-$(git log --pretty=format:'%h' -n 1)"
# magma@version.commit_count-commit_hash
# Example: magma@1.7.0.9829-ae3224e0f
# Sentry versions: major: 1, minor: 7, bugfix: 0, revision: 9829
```
Setting this version instead of the current version (`version-commit_hash`) allows Sentry to understand the order of releases correctly. Using the commit count as revision version guarantees that a later commit in the main branch will be treated as a later release independent of when it was deployed.

Versions using the scheme are shown in Sentry like this:
![releases](https://user-images.githubusercontent.com/14236667/145837891-3427a9e7-8ef1-4566-b9b3-efbb5c7dab48.png)

## Open point

Currently the CI build process sends release dumps to Sentry that follow a third version scheme `version-timestamp-commit_hash` that is incompatible with the versions used in the SDK. It might be useful to use the same version scheme everywhere.

ssanadhya:
@sebathomas, is this proposal still active? Can you present it in the next TSC meeting? If yes, I can add it to the [Mar 28 agenda](https://wiki.magmacore.org/pages/resumedraft.action?draftId=2366682&draftShareId=6e21b3b0-dcac-4944-80cd-f077eb7e9a09&) .

sebathomas:
@ssanadhya This was already implemented in #10831.

PR #10831, opened 2021-12-13
chore(agw): use version scheme that is understood by sentry

alexzurbonsen:
## Summary

Sentry cannot understand the current version format that it is given. Thus, the format is changed to "magma@major.minor.bugfix.revision-commithash" where commit count (number of commits on master) is used as revision.

Note, that only the version reported to Sentry is affected here. No other version format is changed.

The intention is to back port this change to 1.6. as well, to allow for Sentry's resolve mechanism to work properly. 

Relates to #10827.

## Test Plan

Built package with build-magma.sh (via `fab release package`) and checked that the version can be found in the respective `COMMIT_HASH_FILE`.

Tested locally that Sentry version is understood correctly and enables usage of Sentry's resolve-mechanism for issues.  

## Additional Information

### DIFF
--- a/lte/gateway/release/build-magma.sh
+++ b/lte/gateway/release/build-magma.sh
@@ -27,7 +27,8 @@ SCTPD_MIN_VERSION=1.6.0 # earliest version of sctpd with which this version is c
 BUILD_TYPE=RelWithDebInfo
 
 # Cmdline options that overwrite the version configs above
-COMMIT_HASH=""  # hash of top magma commit (hg log $MAGMA_PATH)
+COMMIT_HASH=""  # hash of top magma commit
+COMMIT_COUNT="" # count of commits on git main
 CERT_FILE="$MAGMA_ROOT/.cache/test_certs/rootCA.pem"
 CONTROL_PROXY_FILE="$MAGMA_ROOT/lte/gateway/configs/control_proxy.yml"
 OS="ubuntu"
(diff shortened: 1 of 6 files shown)


themarwhal:
Looks good! in order to avoid confusion let's mention in the summary that this only affects versions reported to sentry and nothing else.

themarwhal (approved):

lte/gateway/release/build-magma.sh
```diff
@@ -260,7 +260,7 @@ if [ -d ${PY_TMP_BUILD} ]; then
 fi
 
 FULL_VERSION=${VERSION}-$(date +%s)-${COMMIT_HASH}
-COMMIT_HASH_WITH_VERSION=${VERSION}-${COMMIT_HASH}
+COMMIT_HASH_WITH_VERSION="magma@${VERSION}.$(git rev-list --count HEAD)-$(git log --pretty=format:'%h' -n 1)"
```
sebathomas: I just noticed that COMMIT_HASH is also given by a CLI parameter of this script. Are we sure that this is executed inside the repository so that the `git` commands will work?

themarwhal (approved):
thanks for cleaning up the unused vcs param!

lte/gateway/release/build-magma.sh
alexzurbonsen: Yes, works as expected. The path with which the script is called, shows that it is called from inside the repository. https://github.com/magma/magma/blob/32a8a31ed4eb40cc8e0e0b0fc91ce8c0856ce05d/lte/gateway/fabfile.py#L125

quentinDERORY (approved):

alexzurbonsen:
Thanks for merging! :)

PR #10693, opened 2021-12-06
feat(ci): migrate sentry to gha

quentinDERORY:
## Summary

Migrating sentry to gha and will be kicked as part of every agw build done on master and new releases.

## Test Plan

Tested on my forked master

## Additional Information

For releases we still need to find a way to dynamically adjust the version (currently being 1.7.0)

PR #10832, opened 2021-12-13
chore(agw): consolidate sentry-cli and sentry_sdk releases

jheidbrink:
## Summary

Currently the schema of the release names created with `sentry-cli` is
`<semver>-<timestamp>-<commit_hash>` while the schema of the release names created ad-hoc by the `sentry_sdk` is `<semver>-<commit_hash>` (the latter is about to change in #10831) This PR reads the release name used by `sentry-cli` from the `/usr/local/share/magma/commit_hash` file of the Magma Debian package, which is also what is passed to `sentry_sdk.init`.

## Test Plan

Tested the new command in a local bash shell

## Additional Information

### DIFF
--- a/orc8r/tools/fab/pkg.py
+++ b/orc8r/tools/fab/pkg.py
@@ -141,11 +141,10 @@ def upload_pkgs_to_aws():
 
 def get_magma_version():
     return run(
-        'ls ~/magma-packages'
-        ' | grep "^magma_[0-9].*"'
-        ' | xargs -I "%" dpkg -I ~/magma-packages/%'
-        ' | grep "Version"'
-        ' | awk \'{print $2}\'',
+        'directory=$(mktemp -d) &&'
+        'dpkg-deb --extract ~/magma-packages/magma_[0-9]*.deb $directory &&'
+        'source $directory/usr/local/share/magma/commit_hash &&'
+        'echo $COMMIT_HASH',
     )
 
 


themarwhal:
I'm not seeing a connection as to how this get version function is used by function, can you elaborate?

jheidbrink:
The current situation with two different naming schemes can be seen in the releases page of the lab-agws-python project:

sentry-two_release_name_schemes

quentinDERORY:
I don't think we are still uploading to aws, and this function is only use by upload to aws...
Maybe we should remove this part altogether? @themarwhal

themarwhal:
> I don't think we are still uploading to aws, and this function is only use by upload to aws...
> Maybe we should remove this part altogether? @themarwhal

Yeah if it's not used let's just delete it :D

jheidbrink:
> I don't think we are still uploading to aws, and this function is only use by upload to aws... Maybe we should remove this part altogether? @themarwhal

Currently it's used by the `copy_packages` function in `pkg.py`: https://github.com/magma/magma/blob/04208a3c89002d2adf488c424f61a1295aed8492/orc8r/tools/fab/pkg.py#L199-L200

The `copy_packages` function is called from Circle CI via Fabric (in the `_run_remote_lte_package` function): https://github.com/magma/magma/blob/04208a3c89002d2adf488c424f61a1295aed8492/circleci/fabfile.py#L365

The `_run_remote_lte_package` function is called from the `integ_test` function.


In the `_deploy_lte_packages` function (which is also called from the `integ_test` function), we copy the previously generated `/tmp/magma_version`: https://github.com/magma/magma/blob/04208a3c89002d2adf488c424f61a1295aed8492/circleci/fabfile.py#L374

As far as I understand it, that `magma_version` file is then read in the the `sentry-release` CircleCI command: https://github.com/magma/magma/blob/b9aa78dbbc483f64beefe9e1f8dbb03f32cad1c0/.circleci/config.yml#L342

So I think what happens is roughly:
`integ_test`:
 * run `get_magma_version()` and write the result to `/tmp/magma_version` in a VM
 * copy that file to `circleci/magma_version` in a Circle CI runner

`sentry-release` CircleCI command:
 * relies on a previously run `integ_test` that has the side effect of writing `magma_version`
 * read that `magma_version` and use it to create a Sentry release


As the Sentry release creation is currently moving to Github Actions in #10693, I'm not sure if it's worth to change it now in the CircleCI job. But this PR can already serve as a discussion for consolidating the Sentry release versions, and I think this command would work equally well in the new Github Actions workflow.

themarwhal:
If you build a package with `fab release pkg` do we still get the same version as now with this change? I just want to make sure we're not affecting anything other than the sentry version.

jheidbrink:
The result of `fab release package` is not influenced by this. I tried to find all places affected by this change and found:

* the release created with `sentry_cli`
* The "Artifact Version" reported in Slack for the "LGE AWG" artifact in the #ci-artifacts slack channel, like [here](https://magmacore.slack.com/archives/C01RGBRKGRJ/p1637355774164000)

themarwhal (approved):

quentinDERORY (approved):

PR #10833, opened 2021-12-13
fix(agw): Update dependency issue python3-websocket-client

jaredmullane:

## Summary

Previously (https://github.com/magma/magma/pull/10723) the python3-websocket-client was updated to version 1.2.2, however, the error below shows 1.2.3 (latest version) to be installed. 

image


ardzoht (changes requested):
this will break magma test VM dependency, as that VM has python3.5 installed this version is not found, @tmdzk is working on the pydep issue that is impeding to build the package if there's no version specified

jaredmullane:
@tmdzk has a new fix landing.

PR #10838, opened 2021-12-13
feat(feg_hello): counters for agw-orc8r-feg latency

uri200:

## Summary

Added some counters to see if we can catch latency between elements.

Note the only true value is `total time` which is taken at the client sending request and client receiving request. The other counters are best effort and may depend on how well synchronized are AGW, orc8r and Feg with time servers. 

## Test Plan

Tested using federated local setup on mac

```
(python) vagrant@magma-dev-focal:~$ feg_hello_cli.py m 0
- Request:
 greeting: "m"

- Response: m

- Stats:
  * Total time:     61 ms
  * Aproximate path (ms):
    ┌─────┐─>     15 ─>┌───────┐->     16 ─>┌─────┐
    │ AGW │            │ Orc8r │            │ Feg │
    └─────┘<─ 10     <─└───────┘<─ 20     <─└─────┘
```

## Additional Information


feg/protos/hello.proto
```diff
@@ -40,13 +42,24 @@ message HelloRequest {
...
+  google.protobuf.Timestamp agw_to_feg_relay_timestamp = 3;
 }
 
 message HelloReply {
   string greeting = 1;
+  ResponseTimestamps timestamps = 2;
+}
+
+message ResponseTimestamps {
+  // timestamp oming from HelloRequest
```
emakeev: coming ...

emakeev (commented):
Here is another hello client, you may want to update: https://github.com/magma/magma/blob/master/feg/gateway/tools/hello_cli/main.go

uri200:
> Here is another hello client, you may want to update: https://github.com/magma/magma/blob/master/feg/gateway/tools/hello_cli/main.go

Oh that's true. Will update it too.

But if this one is fine i would like to land before tomorrow (Tuesday) so we can have, at least, the feg abd orc8r  part in place so we can test  it in real deployments like teravm

themarwhal (approved):

lte/gateway/python/scripts/feg_hello_cli.py
```diff
@@ -22,10 +23,35 @@
...
 @cloud_grpc_wrapper
 def echo(client, args):
+    start_time = datetime.datetime.now()
     req = HelloRequest(greeting=args.msg, grpc_err_code=args.err_code)
-    print("request:\n", req)
+    print("- Request:\n", req)
     resp = client.SayHello(req)
-    print("resp: ", resp)
+    end_time = datetime.datetime.now()
+    print(f'- Response: {resp.greeting}')
```
ardzoht, line 31: should this be inside the `if resp` condition?

lte/gateway/python/scripts/feg_hello_cli.py
```diff
@@ -45,6 +71,24 @@ def create_parser():
     return parser
 
 
+def to_ms_string(seconds, left=False):
+    """
+    Convert seconds into a sting of milliseconds. If the string is smaller
```
ardzoht:
```suggestion
    Convert seconds into a string of milliseconds. If the string is smaller
```

lte/gateway/python/scripts/feg_hello_cli.py
uri200: good catch,  I think i forgot to remove that

ardzoht (approved):
🚀

emakeev (approved):

emakeev:
> But if this one is fine i would like to land before tomorrow (Tuesday) so we can have, at least, the feg abd orc8r part in place so we can test it in real deployments like teravm

Sure, it definitely can go into a separate PR

themarwhal (approved):

PR #10842, opened 2021-12-13
chore(mme): Move proto to GUTI and ECGI converters to base class

ssanadhya:

## Summary

The GUTI/ECGI serialization/deserialization functions were split across base state converter class and NAS state converter class. This change brings them all under State converter base class.

## Test Plan

- `make build_oai`
- Run following stateless tests:
test_attach_detach_with_mme_restart.py
test_idle_mode_with_mme_restart.py


ardzoht (approved):
lgtm

PR #10848, opened 2021-12-14
test(mme): Add S1 Handover MME unit tests

ardzoht:

## Summary

- Adding unit test for MME procedures related with S1 Handover 
- Adding them in single unit test to ensure state correctness and order of function calls

## Test Plan

- make test_oai

## Additional Information


ulaskozat (approved):
LGTM

PR #10866, opened 2021-12-14
feat(agw): Mock out the gateway.mconfig dependency while running update test for ConfigManager

### LINKED ISSUE #10841: Mock out the gateway.mconfig dependency while running update test for ConfigManager
### Describe the problem

**Description**
Right now we have to copy the gateway.mconfig file to the CI, because there is a call to load_service_mconfig, which in turn leads to a read call to gateway config file.

test file:- https://github.com/magma/magma/blob/f05b82dbc2cd8095d2ab69719a019dee8aa9ed1c/orc8r/gateway/python/magma/magmad/tests/config_manager_tests.py

### Describe your idea (optional)
Instead of physically copying the config file, it would be helpful if we can mock out the dependency on this load_service_mconfig function.

https://github.com/magma/magma/blob/f05b82dbc2cd8095d2ab69719a019dee8aa9ed1c/orc8r/gateway/python/magma/configuration/mconfig_managers.py#L50

used here:-
https://github.com/magma/magma/blob/f05b82dbc2cd8095d2ab69719a019dee8aa9ed1c/orc8r/gateway/python/magma/magmad/config_manager.py#L116

ankitMeta1:
Signed-off-by: Ankit Bhadoria <abhadoria@fb.com>

feat(agw): Mock out the gateway.mconfig dependency while running update test for ConfigManager

## Summary

mocked the dependency on gateway.mconfig file and changed the build pipeline to not copy the gateway file

## Test Plan

ran unit tests

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -105,9 +105,6 @@ jobs:
             wget https://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/2.2.3/swagger-codegen-cli-2.2.3.jar -O ${{ env.SWAGGER_CODEGEN_JAR }}
       - name: Execute test_all
         run: |
-            sudo mkdir /etc/magma
-            sudo -E chmod a+rx /etc/magma
-            sudo cp ${{ env.MAGMA_ROOT }}/lte/gateway/configs/gateway.mconfig /etc/magma/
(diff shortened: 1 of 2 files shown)


themarwhal (approved):
Thanks for cleaning this up :)

hcgatewood (approved):

talkhasib (approved):

PR #10867, opened 2021-12-14
test(mme): Expand unit test for Nas state converter

### LINKED ISSUE #9605: Add MME_APP/NAS state conversion unit tests
MME APP state converter (https://github.com/magma/magma/blob/master/lte/gateway/c/core/oai/tasks/mme_app/mme_app_state_converter.cpp) needs code coverage by adding unit testing. These are well-contained functions that don't have any dependency to mock/include.

For reference and an example of a state converter unit test, please refer to: https://github.com/magma/magma/blob/master/lte/gateway/c/core/oai/test/spgw_task/test_spgw_state_converter.cpp

ssanadhya:
## Summary

This change adds EMM procedures in the sample EMM context for Nas state converter test.

It also fixes:
- Cpp-lint warnings in state converter base class
- Fixes implementation of delete EMM procedures function
- Add checks in bstring initialization function

## Test Plan

`make test_oai OAI_TESTS="test_nas.*"`

### DIFF
--- a/lte/gateway/c/core/oai/include/state_converter.h
+++ b/lte/gateway/c/core/oai/include/state_converter.h
@@ -186,7 +186,7 @@ class StateConverter {
 
  private:
   static void plmn_to_chars(const plmn_t& state_plmn, char* plmn_array);
-  static void chars_to_plmn(const char* plmn_array, plmn_t& state_plmn);
+  static void chars_to_plmn(const char* plmn_array, plmn_t* state_plmn);
 };
 
 }  // namespace lte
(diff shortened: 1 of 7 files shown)


lte/gateway/c/core/oai/include/state_converter.h
```diff
@@ -168,7 +168,7 @@ class StateConverter {
 
  private:
   static void plmn_to_chars(const plmn_t& state_plmn, char* plmn_array);
-  static void chars_to_plmn(const char* plmn_array, plmn_t& state_plmn);
+  static void chars_to_plmn(const char* plmn_array, plmn_t* state_plmn);
```
ulaskozat, line 189: what is wrong with pass by reference?

ulaskozat (approved):
LGTM with a nit. I personally prefer using pass by reference unless there is a specific need for passing the pointer.

lte/gateway/c/core/oai/include/state_converter.h
ssanadhya: Cpp lint complains if you use reference for non-const references.

ulaskozat: strange. okay let's go with linter then...

PR #10868, opened 2021-12-14
fix(orc8r): Duplicate region name for terraform

ShubhamTatvamasi:

## Additional Information


hcgatewood (commented):
There's some now-lost tribal knowledge for why we didn't do this. Lmk if you're able to validate on a fresh install and then I'll be all for it!

ShubhamTatvamasi:
> There's some now-lost tribal knowledge for why we didn't do this. Lmk if you're able to validate on a fresh install and then I'll be all for it!

So every module does require a region variable and we can't import it from another module. Will try to solve this in future.

PR #10881, opened 2021-12-15
fix(mobilityd): Less strict APN parsing in GetSubscriberIPTable

### LINKED ISSUE #10880: Cannot parse APN in mobilityd GetSubscriberIPTable
Sentry Issue: [LAB-AGWS-PYTHON-43](https://sentry.io/organizations/lf-9c/issues/2861309167/?referrer=github_integration)

```
ValueError: not enough values to unpack (expected 2, got 1)
  File "grpc/_server.py", line 435, in _call_behavior
    response_or_iterator = behavior(argument, context)
  File "/usr/local/lib/python3.8/dist-packages/magma/mobilityd/rpc_servicer.py", line 383, in GetSubscriberIPTable
    apn, _ = apn_part.split(',')

Exception calling application: not enough values to unpack (expected 2, got 1)
```

sebathomas:
## Summary

Sentry monitoring shows that parsing of the composite SID into
an APN can fail because there is no comma and ipvx appendix in
the composite string.

This change makes the parser ignore this error and simply return
the part after the period as APN if there is no comma.

Fixes #10880.

### DIFF
--- a/lte/gateway/python/magma/mobilityd/rpc_servicer.py
+++ b/lte/gateway/python/magma/mobilityd/rpc_servicer.py
@@ -380,7 +380,7 @@ class MobilityServiceRpcServicer(MobilityServiceServicer):
         for composite_sid, ip in csid_ip_pairs:
             # handle composite sid to sid and apn mapping
             sid, _, apn_part = composite_sid.partition('.')
-            apn, _ = apn_part.split(',')
+            apn, *_ = apn_part.split(',')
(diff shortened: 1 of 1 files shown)


sebathomas:
This would avoid raising an error if such an entry is in the Subscriber IP Table.

I am not sure if this is the right way to handle this or if the entry is actually invalid and should raise an error.

pshelar (approved):
LGTM

PR #10895, opened 2021-12-15
chore: add update multiple fegs to TeraVM fabfile

uri200:

## Summary

Added support to update multiple fegs.

Note fab configuration file needs to be updated to use an array `[ ` on feg description. 

Remember that in order to have passwordless ssh access to all your fegs, you need to execute this command on your driver
```
ssh-copy-id magma@your_fegs_ip
```

Also, note you will need to add magma user to sudoers so it can do sudo commands without password

## Additional Information

### DIFF
--- a/ci-scripts/teravm/fabfile.py
+++ b/ci-scripts/teravm/fabfile.py
@@ -240,33 +240,34 @@ def upgrade_teravm_feg(setup, hash, key_filename=DEFAULT_KEY_FILENAME):
     the remote host. IIf empty file name is passed, password-based ssh will
     work instead. This can be used if the script is run manually.
     """
-    fastprint("\nUpgrade teraVM FEG to %s\n" % hash)
-    _setup_env("magma", VM_IP_MAP[setup]["feg"], key_filename)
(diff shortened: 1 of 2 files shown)


ci-scripts/teravm/fabfile.py
```diff
@@ -240,33 +240,34 @@ def upgrade_teravm_feg(setup, hash, key_filename=DEFAULT_KEY_FILENAME):
...
-                "doesn't exist. Reverting to the original "
-                "config\n".format(hash)
-            )
-            fastprint(err)
-            with hide("running", "stdout"):
-                sudo("mv docker-compose.yml.backup docker-compose.yml")
-                sudo("mv .env.backup .env")
+    for feg_ip in VM_IP_MAP[setup]["feg"]:
+        fastprint("\nUpgrade teraVM FEG (%s) to %s\n" % (feg_ip, hash))
+        _setup_env("magma", feg_ip, key_filename)
```
uri200, line 245: i just added this for (and use `feg_ip`). The rest is unchanged

tmdzk (approved):

PR #9926, opened 2021-10-27
chore(agw): fully specify orc8r/gw/c/common included paths

electronjoe:
## Summary

To be clear, this PR strictly maps include paths and has no intended functional impact.

This PR is one of many intended to migrate our C/C++ codebase to full repo-relative paths. See #9861.

## Process

This PR was generated by using [this go binary](https://gist.github.com/electronjoe/dce83fcb6b9171dd3668f76a961b9cd1) executed as follows:

```bash
go run pathify.go -root /workspaces/magma/ -target /workspaces/magma/orc8r/gateway/c/common/ -dry_run=false -target_filter ".h,.c,.cpp"
```

Additionally some CMake project changes were required. Namely I added `$MAGMA_ROOT` to the include path for the `session_manager/CMakeLists.txt` top level file

## Test Plan

All builds and tests pass - compilation success is an overwhelming indicator of correct include modifications.

Signed-off-by: Scott Moeller <electronjoe@gmail.com>

PR #10904, opened 2021-12-16
feat(sctpd): Enable mconfig and Sentry monitoring

### LINKED ISSUE #9509: Add Sentry to SCTPD
1. Remove sigmask of SEGV: https://github.com/magma/magma/blob/master/lte/gateway/c/sctpd/src/sctpd.cpp#L52 . Blocking SEGV will interfere with Sentry's ability to catch segfaults.
2. Copy how sentry is initialized in [sessiond_main.cpp](https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/sessiond_main.cpp#L210). You can ignore the bit about mconfig until https://github.com/magma/magma/issues/9508 is done. 
3. Modify CI to upload SCTPD debug artifact. See how this is done for SessionD in [.circleci/config.yml](https://github.com/magma/magma/blob/master/.circleci/config.yml#L354) (see also [this](https://github.com/magma/magma/blob/master/.circleci/config.yml#L358))

### LINKED ISSUE #10877: Enable support for mconfig in SCTPd
At the moment, SCTPd is not configurable from the cloud. The goal of this issue is to enable mconfig in SCTPd in order to configure a Sentry monitoring URL and sample rate as well as the log level of the service.

sebathomas:
## Summary

Introduce mconfig for SCTPd, and use it to set the log level and Sentry configuration. Enable Sentry if a URL is given.

This uses the default log level INFO, which is also used for the other services, instead of DEBUG as hardcoded before.

Closes #10877.

## Test Plan

Sent a message to Sentry test account via local SCTPd after configuring the Sentry URL in Swagger.

## Additional Information


lte/gateway/c/sctpd/src/BUILD.bazel
```diff
@@ -24,6 +24,9 @@ cc_binary(
         ":config",
         ":sctpd_downlink_impl",
         ":sctpd_event_handler",
+        "//lte/protos:mconfigs_cpp_proto",
+        "//orc8r/gateway/c/common/config:mconfig_loader",
+        "//orc8r/gateway/c/common/sentry:sentry_wrapper",
```
sebathomas, line 29: This is so much more intuitive than the cmake file. :smile:

lte/gateway/c/sctpd/src/sctpd.cpp
```diff
@@ -75,11 +80,27 @@ int signalHandler(int* end, std::unique_ptr<Server>& server,
...
+    mconfig.set_log_level(magma::orc8r::LogLevel::INFO);
+  }
+  return mconfig;
+}
+
 int main() {
   signalMask();
 
-  magma::init_logging("sctpd");
-  magma::set_verbosity(MDEBUG);
```
sebathomas, line 82: We used INFO because all other services seem to use INFO as default in the mconfig. Was there a reason to set DEBUG for this service?

lte/gateway/c/sctpd/src/sctpd.cpp
```diff
@@ -75,11 +80,27 @@ int signalHandler(int* end, std::unique_ptr<Server>& server,
...
+  auto mconfig = load_mconfig();
+
+  magma::init_logging(SCTPD_SERVICE);
+  magma::set_verbosity(mconfig.log_level());
```
sebathomas: It seems like the log levels between mconfig and verbosity are inverted so INFO is interpreted as ERROR (not just here but in other services too). Is this a known issue or am I misunderstanding something?

sebathomas:
Looking into the CI errors...

lte/gateway/c/sctpd/src/sctpd.cpp
themarwhal: That's interesting... How do other services handle it? I remember we have some logic to remap log values to glog log levels.

themarwhal:
@sebathomas I think we may have to add "-ldl" to `bazel/external/sentry_native.BUILD`

sebathomas:
> @sebathomas I think we may have to add "-ldl" to `bazel/external/sentry_native.BUILD`

That seems to fix it locally, thank you! Strange it wasn't needed for sessiond.

lte/gateway/c/sctpd/src/sctpd.cpp
```diff
@@ -75,11 +80,27 @@ int signalHandler(int* end, std::unique_ptr<Server>& server,
...
+
+  sentry_config_t sentry_config;
+  sentry_config.sample_rate = mconfig.sentry_config().sample_rate();
+  strncpy(sentry_config.url_native,
+          mconfig.sentry_config().dsn_native().c_str(), MAX_URL_LENGTH);
```
themarwhal: seeing this in CI
```
lte/gateway/c/sctpd/src/sctpd.cpp: In function 'int main()':
lte/gateway/c/sctpd/src/sctpd.cpp:101:10: warning: 'char* strncpy(char*, const char*, size_t)' specified bound 255 equals destination size [-Wstringop-truncation]
  101 |   strncpy(sentry_config.url_native,
      |   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
  102 |           mconfig.sentry_config().dsn_native().c_str(), MAX_URL_LENGTH);
```

lte/protos/mconfig/mconfigs.proto
```diff
@@ -530,3 +530,12 @@ message AgwD {
...
   string sentry_dsn = 6;
 }
+
+//------------------------------------------------------------------------------
+// SctpD configs
+//------------------------------------------------------------------------------
+message SctpD {
+  orc8r.LogLevel log_level = 1;
+  // sentry_config stores the Sentry.io configuration for this service
+  SentryConfig sentry_config = 2;
```
themarwhal: Instead of taking this value here, let's take it from the Shared mconfig

themarwhal:
> > @sebathomas I think we may have to add "-ldl" to `bazel/external/sentry_native.BUILD`
> 
> That seems to fix it locally, thank you! Strange it wasn't needed for sessiond.

yeah I'm not sure why exactly :/ maybe it was already linked in for something else

lte/gateway/c/sctpd/src/sctpd.cpp
themarwhal: ah looks like same warning exists for sessiond? (would be great if you could address both 🙏 )

lte/protos/mconfig/mconfigs.proto
themarwhal: I think it's still useful to have a Magma SCTPD config section though. (We probably have other configurations that should be changeable from the cloud)  But let's pull sentry config from shared mconfig.

lte/gateway/c/sctpd/src/sctpd.cpp
sebathomas: I think there's a mismatch between the values in `magma::orc8r::LogLevel` and the `magma_logging.h` constants. Putting this into the main
```cpp
magma::set_verbosity(magma::orc8r::LogLevel::DEBUG);
magma::set_verbosity(magma::orc8r::LogLevel::INFO);
magma::set_verbosity(magma::orc8r::LogLevel::WARNING);
magma::set_verbosity(magma::orc8r::LogLevel::ERROR);
magma::set_verbosity(magma::orc8r::LogLevel::FATAL);
magma::set_verbosity(MDEBUG);
magma::set_verbosity(MINFO);
magma::set_verbosity(MWARNING);
magma::set_verbosity(MERROR);
magma::set_verbosity(MFATAL);
```
results in:
```
I1216 15:31:06.167461 203384 magma_logging_init.h:28] Setting verbosity to 0
I1216 15:31:06.167547 203384 magma_logging_init.h:28] Setting verbosity to 1
I1216 15:31:06.167569 203384 magma_logging_init.h:28] Setting verbosity to 2
I1216 15:31:06.167587 203384 magma_logging_init.h:28] Setting verbosity to 3
I1216 15:31:06.167603 203384 magma_logging_init.h:28] Setting verbosity to 4
I1216 15:31:06.167619 203384 magma_logging_init.h:28] Setting verbosity to 4
I1216 15:31:06.167635 203384 magma_logging_init.h:28] Setting verbosity to 3
I1216 15:31:06.167675 203384 magma_logging_init.h:28] Setting verbosity to 2
I1216 15:31:06.167696 203384 magma_logging_init.h:28] Setting verbosity to 1
I1216 15:31:06.167714 203384 magma_logging_init.h:28] Setting verbosity to 0
```
So there's still a conversion missing. Sessiond, liagent and connection tracker have a function  `get_log_verbosity` that will also return the mconfig value directly if it's not overwritten by the config.

themarwhal: yikes - we should probably fix that :D Can you file an issue?

sebathomas: #10911

lte/gateway/c/sctpd/src/BUILD.bazel
electronjoe: Yeah.. CMake exposes all sorts of the guts of building and linking as part of the configuration files - when that is absolutely not necessary and a better abstraction talks about e.g. Dependencies for the most part (which is the approach of Bazel / Buck).  This is precisely one of our arguments on Bazel - it is *hard* to contribute to our C/C++ today given the CMake situation (which admittedly is worse than it needs to be, if we were to completely fix all the CMake non-best-practices).  Bazel offers a minimum API and a "blessed path" primarily - so it's easy to replicate and do things the right way and you need not get into the guts of compilation.  /rant

lte/gateway/c/sctpd/src/sctpd.cpp
```diff
@@ -22,6 +23,10 @@
 #include "lte/gateway/c/sctpd/src/sctpd_uplink_client.h"
 #include "lte/gateway/c/sctpd/src/util.h"
 #include "orc8r/gateway/c/common/logging/magma_logging_init.h"
+#include "includes/SentryWrapper.h"
+#include "includes/MConfigLoader.h"
```
themarwhal:
```suggestion
#include "orc8r/gateway/c/common/sentry/includes/SentryWrapper.h"
#include "orc8r/gateway/c/common/config/includes/MConfigLoader.h"
```
@electronjoe recently made it so that we fully specify include paths (https://github.com/magma/magma/pull/9926) We still have not made the switch for this in SessionD as it breaks CLion <-> CMake integration, apparently.

lte/gateway/c/sctpd/src/sctpd.cpp
sebathomas: I added a manual null termination but I am not sure if there is a default way how to do it in Magma? I saw that many people recommend `strlcpy` but it doesn't seem to be used in the project.

orc8r/protos/BUILD.bazel
```diff
@@ -69,10 +69,15 @@ proto_library(
...
+cpp_proto_library(
+    name = "mconfigs_cpp_proto",
+    protos = [":mconfigs_proto"],
+    deps = [":common_cpp_proto"],
+)
+
 proto_library(
     name = "mconfigs_proto",
     srcs = ["mconfig/mconfigs.proto"],
-    strip_import_prefix = "mconfig/",
```
sebathomas, line 75: I had to remove this to get the `#include` to work. It's now analogous to the mconfig settings in lte.

lte/protos/mconfig/mconfigs.proto
sebathomas: Done.

themarwhal (approved):

sebathomas:
(Rebased and tried to fix sctpd.cpp include order warning.)

orc8r/protos/BUILD.bazel
themarwhal: Yeah I wonder why I added this in the first place :o If the rest of C/C++ services still build fine, this is OK :)

lte/gateway/c/sctpd/src/sctpd.cpp
```diff
@@ -75,11 +83,41 @@ int signalHandler(int* end, std::unique_ptr<Server>& server,
...
+  return mconfig;
+}
+
+static magma::mconfig::SharedMconfig load_shared_mconfig() {
+  magma::mconfig::SharedMconfig mconfig;
+  magma::load_service_mconfig_from_file(SHARED_MCONFIG, &mconfig);
+  return mconfig;
+}
+
+static uint32_t get_log_verbosity(magma::mconfig::SctpD mconfig) {
```
themarwhal: Might be good to place this function in the common mconfig library so that it can be used by other services as well?

themarwhal:
@sebathomas While we wait for reviews from @magma/approvers-agw-mme , want to add the debug artifact upload logic to https://github.com/magma/magma/blob/master/.github/workflows/build_all.yml#L222 ?

lte/gateway/c/sctpd/src/sctpd.cpp
sebathomas: Done in #10969, will rebase and do it for sctpd when one of the commits is merged.

quentinDERORY (approved):

Neudrino (approved):

ulaskozat:
@ssanadhya which line of code do we need to touch to make sure that sctpd is also upgraded?

ssanadhya:
> @ssanadhya which line of code do we need to touch to make sure that sctpd is also upgraded?

@ulaskozat, @sebathomas, you can upgrade the Sctpd dependency by upgrading SCTPD_MIN_VERSION here https://github.com/magma/magma/blob/master/lte/gateway/release/build-magma.sh#L24

ulaskozat:
> > @ssanadhya which line of code do we need to touch to make sure that sctpd is also upgraded?
> 
> @ulaskozat, @sebathomas, you can upgrade the Sctpd dependency by upgrading SCTPD_MIN_VERSION here https://github.com/magma/magma/blob/master/lte/gateway/release/build-magma.sh#L24

this PR does not break any backward compatibility as far as I see. The question is whether we should simply advise in 1.7 release to update sctpd service (as part of release notes) or make it mandatory with this line of code change.

lte/gateway/c/sctpd/src/sctpd.cpp
ulaskozat: not really. INFO level is good.

lte/gateway/c/sctpd/src/sctpd.cpp
```diff
@@ -75,11 +83,35 @@ int signalHandler(int* end, std::unique_ptr<Server>& server,
   return 0;
 }
 
+static magma::mconfig::SctpD load_sctpd_mconfig() {
+  magma::mconfig::SctpD mconfig;
+  if (!magma::load_service_mconfig_from_file(SCTPD_SERVICE, &mconfig)) {
+    mconfig.set_log_level(magma::orc8r::LogLevel::INFO);
```
ulaskozat, line 90: can we instead try to get it first from sctpd.yml file before hardwiring the default here?

ulaskozat: in terms of priority yml file takes precedence over streamed down config which takes precedence over hardwired log level as the convention in AGW.

sebathomas: There is no sctpd.yml config file yet. Should we introduce one for the log level?

sebathomas: Added a second commit with a yaml config.

lte/gateway/c/sctpd/src/sctpd.cpp
```diff
@@ -75,11 +84,58 @@ int signalHandler(int* end, std::unique_ptr<Server>& server,
...
+  } else if (log_level == "WARNING") {
+    return MWARNING;
+  } else if (log_level == "ERROR") {
+    return MERROR;
+  } else if (log_level == "FATAL") {
+    return MFATAL;
+  } else {
+    MLOG(MINFO) << "Invalid log level in config: "
+                << config["log_level"].as<std::string>();
+    return MINFO;
```
sebathomas, line 120: This conversion is used in a lot of services and could be extracted somewhere (maybe as a function in ServiceConfigLoader?).

lte/gateway/configs/sctpd.yml
```diff
@@ -0,0 +1,14 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+log_level: INFO
```
ulaskozat: since this will take precedence over orc8r streamed configs, let's comment it out and place a comment on top stating that either it can be commented out here or an override config can be specified under `/var/opt/magma/configs/sctpd.yml`.

ulaskozat (approved):
LGTM except for the comment I provided about commenting out the line in yml file and placing a comment before it that explains that it overrides cloud config if commented out or yml file is specified in the override directory.

ssanadhya:
> > > @ssanadhya which line of code do we need to touch to make sure that sctpd is also upgraded?
> > 
> > 
> > @ulaskozat, @sebathomas, you can upgrade the Sctpd dependency by upgrading SCTPD_MIN_VERSION here https://github.com/magma/magma/blob/master/lte/gateway/release/build-magma.sh#L24
> 
> this PR does not break any backward compatibility as far as I see. The question is whether we should simply advise in 1.7 release to update sctpd service (as part of release notes) or make it mandatory with this line of code change.

It is best to minimize upgrades to Sctpd (for hitless Magma package upgrades). We can just recommend upgrade to Sctpd in 1.7 release notes, without making it a mandatory dependency.

sebathomas:
> LGTM except for the comment I provided about commenting out the line in yml file and placing a comment before it that explains that it overrides cloud config if commented out or yml file is specified in the override directory.

Good point, done. :+1:

themarwhal:
@sebathomas Any clue why the Magma OAI Jenkins test is failing? (cc @rdefosse)

rdefosse:
> @sebathomas Any clue why the Magma OAI Jenkins test is failing? (cc @rdefosse)

All 3 CI runs over the past 24 hours failed.

`https://jenkins-oai.eurecom.fr/job/MAGMA-MME-production/11084/artifact/test_results_magma_epc_rhel8.html`

Looks like MAGMA-MME SCTPD does not respond to eNB S1AP request.

Then `https://jenkins-oai.eurecom.fr/job/MAGMA-MME-production/11084/artifact/test_results_magma_oai_epc.html`

when it does, UE attachments are not easy..

I would suspect `sctpd` is incorrect.

sebathomas:
> @sebathomas Any clue why the Magma OAI Jenkins test is failing? (cc @rdefosse)

Oops, I missed that, will take a look...

sebathomas:
It seems that sctpd fails to start if no file gateway.mconfig can be found.

rdefosse:
> It seems that sctpd fails to start if no file gateway.mconfig can be found.

Is there a template?

I guess if a template exists, it has to be added in the following dockerfiles:
`ci-scripts/docker/Dockerfile.mme.ci.rhel8  ci-scripts/docker/Dockerfile.mme.ci.ubuntu18  lte/gateway/docker/mme/Dockerfile.rhel8  lte/gateway/docker/mme/Dockerfile.ubuntu18.04`

sebathomas:
I think I was wrong. The problem seems to be sctpd.yml not existing. If the mconfig file is missing there is an error message but the service doesn't crash.

sebathomas:
Not sure what to do about it. It seems the other services also cannot handle it if the service.yml file is missing.

Is there something I need to do for the file to exist in the CI?

rdefosse:
> 

Because in OAI deployment use case, we only activate MME and SCTPD services

rdefosse:
Sorry I clicked on the wrong button

sebathomas:
> > 
> 
> Because in OAI deployment use case, we only activate MME and SCTPD services

With the changes here, it would be necessary to create a file `/etc/magma/sctpd.yml` even for that use case.

rdefosse:
@sebathomas 

OK. It's late here in France and I am in the midst of the quarterly SW release.

Let me have a look tomorrow morning. 

I will provide you with a patch to apply to your branch

rdefosse:
@sebathomas 

Here is the patch I promised.

I've tested manually and the `sctpd` executable is now starting properly.

Apply the patch `git apply sctpd-mconfig-sentry.patch` and push again on your branch. CI will trigger again.

[sctpd-mconfig-sentry.zip](https://github.com/magma/magma/files/7895993/sctpd-mconfig-sentry.zip)

Regards
Raphael

sebathomas:
@rdefosse Awesome, thank you. Just pushed again with the patch.

rdefosse (approved):

sebathomas:
Rebasing on master.

PR #10922, opened 2021-12-16
feat(directoryd): add new SgwUteid to directoryd and indexer

uri200:

## Summary

Added Sgw User Plane TEID management on orc8r (same we did for Control Plane TEID)

## Test Plan

Teravm (to be done soon)

## Additional Information


orc8r/cloud/go/services/directoryd/servicers/servicer.go
```diff
@@ -160,3 +162,54 @@ func (d *directoryLookupServicer) GetNewSgwCTeid(ctx context.Context, req *proto
 	}
 	return &protos.GetNewSgwCTeidResponse{Teid: fmt.Sprint(sgwCTeid)}, nil
 }
+
```
uri200, line 165: Note that if directoryd were more object oriented friendly, we could reduce the repetition of code. Maybe once we have time we can refactor all this and other functions which are just a duplication with different parameters

orc8r/lib/go/protos/directoryd_helper.go
```diff
@@ -70,6 +70,16 @@ func (m *GetHWIDForSgwCTeidRequest) Validate() error {
 	return nil
 }
 
+func (m *GetHWIDForSgwUTeidRequest) Validate() error {
+	if m == nil {
+		return errors.New("request cannot be nil")
+	}
+	if m.Teid == "" {
+		return errors.New("request teid cannot be empty")
```
emakeev, line 78: nit: It may be easier to debug if you change 'request" to "GetHWIDForSgwUTeidRequest"

orc8r/lib/go/protos/directoryd_helper.go
```diff
@@ -83,6 +93,19 @@ func (m *MapSgwCTeidToHWIDRequest) Validate() error {
 	return nil
 }
 
+func (m *MapSgwUTeidToHWIDRequest) Validate() error {
+	if m == nil {
+		return errors.New("request cannot be nil")
```
emakeev, line 98: same as above (nit, up to you)

orc8r/protos/directoryd.proto
```diff
@@ -84,9 +85,38 @@ message GetNewSgwCTeidResponse{
...
+}
+
+message MapSgwUTeidToHWIDRequest {
+  string networkID = 1;
+  map<string, string>  teidToHwid = 2;
+}
+
+message UnmapSgwUTeidToHWIDRequest {
+  string networkID = 1;
+  repeated string teids= 2;
```
emakeev, line 105: formatting ...

emakeev (approved):
LGTM

orc8r/lib/go/protos/directoryd_helper.go
uri200: in this case all the functions on htis file do not include the name, just the error. The caller does include the name of the functions, so for debugging purposes it is more useful to know the caller of this function than the specific Validate function

PR #10926, opened 2021-12-17
feat(agw): Added unit test case to validate the handling of modify bearer request

rsarwad:
## Summary
 Added unit test case to validate the handling of modify bearer request for inbound user

## Test Plan

### DIFF
--- a/lte/gateway/c/core/oai/tasks/sgw_s8/sgw_s8_handlers.c
+++ b/lte/gateway/c/core/oai/tasks/sgw_s8/sgw_s8_handlers.c
@@ -32,9 +32,12 @@ limitations under the License.
 
 extern task_zmq_ctx_t sgw_s8_task_zmq_ctx;
 extern struct gtp_tunnel_ops* gtp_tunnel_ops;
+
+#if !MME_UNIT_TEST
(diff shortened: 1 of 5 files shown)


ulaskozat (approved):
LGTM

lte/gateway/c/core/oai/tasks/sgw_s8/sgw_s8_handlers.c
```diff
@@ -662,6 +665,7 @@ void sgw_s8_handle_modify_bearer_request(
         pgw_ipv6 = &bearer_ctx_p->p_gw_address_in_use_up.address.ipv6_address;
       }
 
+#if !MME_UNIT_TEST
```
ulaskozat, line 668: any reason why this is not moved to just before `gtp_tunnel_ops->send_end_marker(enb, modify_bearer_pP->teid);`

rsarwad: I put it here because after validating the condition, ue_ip is fetched and  ovs related functions, 
send_end_marker() and gtpv1u_del_s8_tunnel() are called. 
For executing unit test case, we can avoid validating the check and calling ovs related functions()

PR #10934, opened 2021-12-17

spikey979:

## Summary

function works as follows:
it remove line breaks,
it remove extra whitespace

## Test Plan

I ran tests /magma/orc8r/cloud/docker/  ./build.py -c

This is for alert:
https://github.com/magma/magma/security/code-scanning/40?query=ref%3Arefs%2Fheads%2Fmaster

### DIFF
--- a/orc8r/cloud/go/obsidian/server/metrics.go
+++ b/orc8r/cloud/go/obsidian/server/metrics.go
@@ -50,12 +50,19 @@ func CollectStats(next echo.HandlerFunc) echo.HandlerFunc {
 		requestCount.Inc()
 		status := strconv.Itoa(c.Response().Status)
 		respStatuses.WithLabelValues(status, c.Request().Method).Inc()
+		url := sanitizeString(c.Request().URL.String())
 		glog.V(2).Infof(
 			"REST API code: %v, method: %v, url: %v\n",
 			status,
 			c.Request().Method,
-			c.Request().URL,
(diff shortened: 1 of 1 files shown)


spikey979:
Can anyone tell me why i get the error: Changes not staged for commit?

orc8r/cloud/go/obsidian/server/metrics.go
```diff
@@ -50,12 +50,21 @@ func CollectStats(next echo.HandlerFunc) echo.HandlerFunc {
...
 	}
 }
+
+func sanitizeString(strString string) string {
+	strSanitizedString := html.EscapeString(strString) //escape special charatchters in HTML text
+	strSanitizedString = strings.Replace(strSanitizedString, "\r", "", -1) //remove line breaks
+	strSanitizedString = strings.Replace(strSanitizedString, "\n", "", -1)
+	strSanitizedString = strings.Join(strings.Fields(strSanitizedString), " ") //remove extra whitespace
+	return strSanitizedString
+}
```
uri200, line 68: when did you see an url containing `\r` or `\n`?  If that is the case maybe there are another issue here

spikey979: I threw out  strings.Replace for \r and \n...

lucasgonze:

Thoughts?

PR #10939, opened 2021-12-20

spikey979:

## Summary

function works as follows:
it remove line breaks,
it remove extra whitespace

## Test Plan
ran /magma/orc8r/cloud/docker/  ./build.py -c

## Additional Information
This is for alert: https://github.com/magma/magma/security/code-scanning/39?query=ref%3Arefs%2Fheads%2Fmaster

### DIFF
--- a/orc8r/cloud/go/obsidian/handler.go
+++ b/orc8r/cloud/go/obsidian/handler.go
@@ -16,8 +16,10 @@ package obsidian
 import (
 	"crypto/x509"
 	"fmt"
+	"html"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/labstack/echo"
@@ -227,7 +229,16 @@ func GetNetworkId(c echo.Context) (string, *echo.HTTPError) {
 	if nid == "" {
 		return nid, NetworkIdHttpErr()
 	}
-	return nid, CheckNetworkAccess(c, nid)
+	sanNid := sanitizeString(nid)
+	return sanNid, CheckNetworkAccess(c, sanNid)
+}
+
+func sanitizeString(strString string) string {
+	strSanitizedString := html.EscapeString(strString)                    //escape special charatchters in HTML text
+	strSanitizedString = strings.ReplaceAll(strSanitizedString, "\r", "") //remove line breaks
+	strSanitizedString = strings.ReplaceAll(strSanitizedString, "\n", "")
+	strSanitizedString = strings.Join(strings.Fields(strSanitizedString), " ") //remove extra whitespace
+	return strSanitizedString
 }
 
 func GetTenantID(c echo.Context) (int64, *echo.HTTPError) {


orc8r/cloud/go/obsidian/handler.go
```diff
@@ -227,7 +229,16 @@ func GetNetworkId(c echo.Context) (string, *echo.HTTPError) {
...
+	sanNid := sanitizeString(nid)
+	return sanNid, CheckNetworkAccess(c, sanNid)
+}
+
+func sanitizeString(strString string) string {
+	strSanitizedString := html.EscapeString(strString)                     //escape special charatchters in HTML text
+	strSanitizedString = strings.Replace(strSanitizedString, "\r", "", -1) //remove line breaks
+	strSanitizedString = strings.Replace(strSanitizedString, "\n", "", -1)
+	strSanitizedString = strings.Join(strings.Fields(strSanitizedString), " ") //remove extra whitespace
+	return strSanitizedString
```
uri200, line 241: what about adding to orc8r utils [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=orc8r%2Flib%2Fgo%2Futil%2Futil.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=163&start_col=0&end_row=163&end_col=0). You can add an extra [test case](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=orc8r%2Flib%2Fgo%2Futil%2Futil_test.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=28&start_col=0&end_row=28&end_col=0) to make sure the format is what you expected and it doesn't break anything

uri200:
I see the comment [here](https://github.com/magma/magma/security/code-scanning/39). Can you please elaborate this? I currently don't see the security issue here. Users adding networks to NMS are admins. So they have all the control.

orc8r/cloud/go/obsidian/handler.go
```diff
@@ -227,7 +229,16 @@ func GetNetworkId(c echo.Context) (string, *echo.HTTPError) {
...
 		return nid, NetworkIdHttpErr()
 	}
-	return nid, CheckNetworkAccess(c, nid)
+	sanNid := sanitizeString(nid)
+	return sanNid, CheckNetworkAccess(c, sanNid)
+}
+
+func sanitizeString(strString string) string {
+	strSanitizedString := html.EscapeString(strString)                     //escape special charatchters in HTML text
+	strSanitizedString = strings.Replace(strSanitizedString, "\r", "", -1) //remove line breaks
```
uri200: why not to use `ReplaceAll`?

orc8r/cloud/go/obsidian/handler.go
```diff
@@ -227,7 +229,16 @@ func GetNetworkId(c echo.Context) (string, *echo.HTTPError) {
 	if nid == "" {
 		return nid, NetworkIdHttpErr()
 	}
-	return nid, CheckNetworkAccess(c, nid)
+	sanNid := sanitizeString(nid)
+	return sanNid, CheckNetworkAccess(c, sanNid)
```
uri200, line 233: you can just do 
```
return sanNid, CheckNetworkAccess(c, sanitizeString(nid))
```
No need to create a temp variable since the name of the function is self explanatory

uri200, line 233: why does nid needs to be sanitized here? why is `nid` coming with `\n` or `\r`?

Note this function is used in other places than logs

spikey979:
Hi @uri200 ... 
I think your first, deleted suggestion, to put this function in orc8r utils, is ok. So it could be used in other places as well. 
Your comment that you don't see the security issue there…. You may be right. I don't really know what kind of string comes to input… Since there is a code scanning alert, I thought the input must be sanitized… In this particular case, how can I check what exactly is coming to my input?

uri200:
Hey @spikey979, sorry I deleted the comments once I better understood the changes. I don't think we have to move the function to a lib now since it is just impacting the log.

My question about the input was on the assumption you have found this issue somewhere. So I was asking where did you find it to try to reproduce it.  Note tat that is just a log and the code pretends to show raw value of its content, that is why it is using `%v` to print it.

To reproduce it you can modify/create a test case for this file [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=orc8r%2Fcloud%2Fgo%2Fobsidian%2Fhandler_test.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=1&start_col=5&end_row=1&end_col=5). If you have an IDE like vscode or intellij you can even run and set breakpoints to see how it works. Otherwise you can run `go test ./..` on that directory

spikey979:
I changed strings.Replace() -> strings.ReplaceAll()... but I’m still not sure that input needs to be sanitized...

mstre123:
> Hey @spikey979, sorry I deleted the comments once I better understood the changes. I don't think we have to move the function to a lib now since it is just impacting the log.
> 
> My question about the input was on the assumption you have found this issue somewhere. So I was asking where did you find it to try to reproduce it. Note tat that is just a log and the code pretends to show raw value of its content, that is why it is using `%v` to print it.
> 
> To reproduce it you can modify/create a test case for this file [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=orc8r%2Fcloud%2Fgo%2Fobsidian%2Fhandler_test.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=1&start_col=5&end_row=1&end_col=5). If you have an IDE like vscode or intellij you can even run and set breakpoints to see how it works. Otherwise you can run `go test ./..` on that directory

i think theres some confusion here - the need for this pr comes from a sec scan issue https://github.com/magma/magma/security/code-scanning/39 
that said I really think any fix needs to be centralized as it comes up a few times #10934

lucasgonze:

Thoughts?

spikey979:
I solve all “Log enteries created from user input” code scanning allerts in one new PR: https://github.com/magma/magma/pull/11738

PR #10940, opened 2021-12-20
build(ci): Generalize c build image.

nstng:

## Summary

**Problem**: Bazel fetches all workspace dependencies regardless of the build target, i.e., if c is build, also all global python dependencies are pulled. For python also dependencies of the requirements.txt are pulled - this currently breaks CI (agw and cwag build image) for https://github.com/magma/magma/pull/10341. This is because only content of the repository for the CI build is copied into the image - especially not the requirements.txt.

**Approach**: Copy the complete repository into the c build image.
* This allows CI builds if python Bazel build is integrated.
* This is consistent to other build images (e.g., lte/gateway/docker/services/python/Dockerfile).

**Note**: It might be a good idea to think about a/one generalized build image for all artifact images.

## Test Plan

* vs CI
* see also checks for https://github.com/magma/magma/pull/10341 were the change here was cherry picked

## Additional Information

### DIFF
--- a/lte/gateway/docker/services/c/Dockerfile
+++ b/lte/gateway/docker/services/c/Dockerfile
@@ -104,9 +104,8 @@ RUN apt-get update && apt-get install -y \
 ENV MAGMA_ROOT /magma
 WORKDIR /magma
 
-# Copy Bazel files at root and third_party
-COPY WORKSPACE.bazel BUILD.bazel .bazelignore .bazelrc .bazelversion ${MAGMA_ROOT}/
-COPY bazel/ ${MAGMA_ROOT}/bazel
+COPY . ${MAGMA_ROOT}
+COPY lte/gateway/docker/mme/configs/ ${MAGMA_ROOT}/lte/gateway/docker/configs/
 
 # Build external dependencies first. This will help not rebuilt all dependencies triggered by Magma changes.
 RUN bazel build \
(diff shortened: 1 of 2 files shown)


lte/gateway/docker/services/c/Dockerfile
```diff
@@ -104,9 +104,8 @@ RUN apt-get update && apt-get install -y \
 ENV MAGMA_ROOT /magma
 WORKDIR /magma
 
-# Copy Bazel files at root and third_party
-COPY WORKSPACE.bazel BUILD.bazel .bazelignore .bazelrc .bazelversion ${MAGMA_ROOT}/
-COPY bazel/ ${MAGMA_ROOT}/bazel
+COPY . ${MAGMA_ROOT}
+COPY lte/gateway/docker/mme/configs/ ${MAGMA_ROOT}/lte/gateway/docker/configs/
```
nstng, line 108: This configs are moved manually - is this important for this ci flow or should they be moved in the repository?

Neudrino (approved):

Neudrino (approved):

themarwhal:
I'm worried that this will result in a full rebuild every time any file is changed in the magma repository? Can you document what the average dev build time will look like with this change? (cc @arunuke )

nstng:
> I'm worried that this will result in a full rebuild every time any file is changed in the magma repository? Can you document what the average dev build time will look like with this change? (cc @arunuke )

You are right @themarwhal, thank you for the input.
For the lte build image:
```
time docker-compose --file docker-compose.yaml --file docker-compose.override.yaml build --parallel
1. fresh build
	real	22m18,203s
2. build with change
	real	18m23,523s
3. build without change
	real	0m24,309s
```
The times are reproducible with minor variance. And a build with change happens now also if an unrelated, e.g., python file is changed.

New approach: the CI currently fails for the python PoC (https://github.com/magma/magma/pull/10341) because of a missing requirements.txt. As this file is only needed for python builds/tests we will move it to the bazel folder, i.e., it will be copied with the Docker logic before this PR. If there is no other opinion, then I would close this PR as obsolete.

themarwhal:
> > I'm worried that this will result in a full rebuild every time any file is changed in the magma repository? Can you document what the average dev build time will look like with this change? (cc @arunuke )
> 
> You are right @themarwhal, thank you for the input. For the lte build image:
> 
> ```
> time docker-compose --file docker-compose.yaml --file docker-compose.override.yaml build --parallel
> 1. fresh build
> 	real	22m18,203s
> 2. build with change
> 	real	18m23,523s
> 3. build without change
> 	real	0m24,309s
> ```
> 
> The times are reproducible with minor variance. And a build with change happens now also if an unrelated, e.g., python file is changed.
> 
> New approach: the CI currently fails for the python PoC (#10341) because of a missing requirements.txt. As this file is only needed for python builds/tests we will move it to the bazel folder, i.e., it will be copied with the Docker logic before this PR. If there is no other opinion, then I would close this PR as obsolete.

Thanks for looking into the build time change @nstng ! Yes, I think keeping the necessary file inside the bazel directory is probably the best option for now. We can reconsider this if we start having files that cannot be contained in that directory. :)

PR #10951, opened 2021-12-21
fix(agw): use pipe instead of redirect

alexzurbonsen:

## Summary

In `else` path: write `COMMIT_HASH` to /etc/environment with pipe instead of redirect (why? sudo is not passed to redirect operator)

## Test Plan

Tested locally.

## Additional Information

### DIFF
--- a/lte/gateway/release/magma-postinst
+++ b/lte/gateway/release/magma-postinst
@@ -17,11 +17,11 @@ sed -i "s/.*OVS_CTL_OPTS.*/OVS_CTL_OPTS='--delete-bridges'/" /etc/default/openvs
 mkdir -p /var/core
 
 value=`cat /usr/local/share/magma/commit_hash`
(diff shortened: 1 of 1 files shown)


alexzurbonsen:
@themarwhal This is not a fix for the Sentry version but something that we realized on the way.
Regarding the broken commit count: Currently we think that `build-magma.sh` might be called directly/manually somewhere? The reasoning behind that: we couldn't reproduce the commit count of 1 locally and `build-magma.sh`' is usually called through `fab ... package` commands in the repo.

themarwhal (approved):

themarwhal:
@alexzurbonsen is this ready2merge?

Neudrino:
Yes, thanks @themarwhal ! :slightly_smiling_face:

Neudrino (approved):

PR #10957, opened 2021-12-21
fix: s6a_proxy fix bug on VisitedPLMNID type

uri200:

## Summary

While trying to run `federated integ test` we were unable to authenticate same way we do it on lte_integ test. After some debugging we found out that PLMNID our mock HSS was reading was always 0. 

The cause for this to be 0 is that there is a bug on go-diameter where they use type as `Unsigned32` instead of `OctetString` for VisitedPLMNID. Since the AVP is in fact an OctetString,  this causes go-diameter marshall function to fail  (in fact it does not fail, it just return 0, without error).

This PR fixes that  on our s6a_proxy. That change also trigger lots of changes on mock HSS test.

Note we also raised a PR for go-diameter
https://github.com/fiorix/go-diameter/pull/167
## Test Plan

This PR adds a check to make sure received PLMN is proper on an existing test case

./build.py -c at feg
And  New federated integ test

## Additional Information

### DIFF
--- a/feg/gateway/services/s6a_proxy/servicers/test/test_s6a_server.go
+++ b/feg/gateway/services/s6a_proxy/servicers/test/test_s6a_server.go
@@ -58,7 +58,7 @@ func StartTestS6aServer(network, addr string, useStaticResp bool) error {
 	} else {
 		mux.HandleIdx(
 			diam.CommandIndex{AppID: diam.TGPP_S6A_APP_ID, Code: diam.AuthenticationInformation, Request: true},
-			testHandleAIR(settings))
+			testHandleAIR(settings, datatype.OctetString(TEST_PLMN_ID)))
 	}
 
 	// expected ULRFlags = 290 (100100010) where the fifth 1 is DualRegistration_5GIndicator : true
@@ -100,7 +100,7 @@ func testHandleALL(results chan error) diam.HandlerFunc {
 }
 
 // S6a AI
-func testHandleAIR(settings *sm.Settings) diam.HandlerFunc {
+func testHandleAIR(settings *sm.Settings, expectedVisitedPLMNID datatype.OctetString) diam.HandlerFunc {
 	return func(c diam.Conn, m *diam.Message) {
 		var req servicers.AIR
 		var code uint32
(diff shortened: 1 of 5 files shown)


feg/gateway/services/testcore/hss/servicers/ai.go
```diff
@@ -72,8 +72,7 @@ func NewAIA(srv *HomeSubscriberServer, msg *diam.Message) (*diam.Message, error)
 		return ConvertAuthErrorToFailureMessage(err, msg, air.SessionID, srv.Config.Server), err
 	}
 
-	const plmnOffsetBytes = 1
-	plmn := air.VisitedPLMNID.Serialize()[plmnOffsetBytes:]
```
uri200, line 76: this changes needs to be reviewed. 

I am not sure why previous code was shifting the plmn by one byle. 

Not removing that code causes PLMN to be just 2 bytes and milenage complains that field should be 3

feg/gateway/services/s6a_proxy/servicers/test/test_s6a_server.go
```diff
@@ -109,6 +109,10 @@ func testHandleAIR(settings *sm.Settings) diam.HandlerFunc {
 		if err != nil {
 			fmt.Printf("AIR Unmarshal for message: %s failed: %s", m, err)
 			code = diam.UnableToComply
+		} else if expectedVisitedPLMNID != req.VisitedPLMNID {
+			// Flags needs to exist for this test
+			fmt.Printf("error: VisitedPLMNID (%v) doesnt match with the expected VisitedPLMNID (%v)\n", req.VisitedPLMNID, expectedVisitedPLMNID)
+			code = diam.UnableToComply
```
uri200, line 115: added this to the test server to make sure received PLMN is what we sent.
Before we were just sending 0 and not checking

themarwhal (approved):

feg/gateway/services/s6a_proxy/servicers/test/test_s6a_server.go
```diff
@@ -109,6 +109,10 @@ func testHandleAIR(settings *sm.Settings) diam.HandlerFunc {
 		if err != nil {
 			fmt.Printf("AIR Unmarshal for message: %s failed: %s", m, err)
 			code = diam.UnableToComply
+		} else if expectedVisitedPLMNID != req.VisitedPLMNID {
+			// Flags needs to exist for this test
+			fmt.Printf("error: VisitedPLMNID (%v) doesnt match with the expected VisitedPLMNID (%v)\n", req.VisitedPLMNID, expectedVisitedPLMNID)
```
emakeev: ... doesn't match expected VisitedPLMNID ...

feg/gateway/services/testcore/hss/servicers/ai.go
emakeev: Screen Shot 2022-01-04 at 5 48 35 PM

I think, there is a byte padding which this was removing

emakeev (approved):
LGTM

feg/gateway/services/testcore/hss/servicers/ai.go
uri200: What I understood after checking go-diameter is it takes care of that and provides the value as octet string. but I may be wrong

Issue #10116, opened 2021-11-04
Development with a Windows OS


Neudrino:
### Your Environment

- Windows 10
- Virtualbox 6.1

#### Additional context

The Magma documentation describes prerequisits and steps needed for developer onboarding. [1, 2]
However, this documentation is primarily focused on Mac OS systems.

[1] https://docs.magmacore.org/docs/basics/prerequisites
[2] https://docs.magmacore.org/docs/basics/quick_start_guide

#### Description

No comment is made about setting up an development environment on a Windows OS.
(Neighter is a comment made on Unix OS, however, this is out of scope of this issue.)

### Describe your idea

This issue is written partly as documentation of the experience, when trying to use a Windows OS as developer.
It aims to avoid any unnecessary time spent and frustration caused.

##### Path 1: Windows native + git-bash

In path 1 an attempt was made to use Windows native solutions. This includes [git-bash](https://gitforwindows.org/) as command line terminal.

❌ Within path 1 a working development environment could _not_ be created successfully.

Problems were:
* Ansible is not supported for Windows. See [Matt Davis' blog](http://blog.rolpdog.com/2020/03/why-no-ansible-controller-for-windows.html) and [following issue](https://github.com/ansible/ansible/issues/31419)
* Directory paths of pip installed packages are too long. [Workarounds](https://stackoverflow.com/questions/48694085/how-to-overcome-pip-install-ansible-on-windows-failing-with-filename-or-extens) are not garantueed to work out flawlessly and are troublesome.
* Other modules not being available for Windows, e.g. [`fcntl`](https://stackoverflow.com/questions/45228395/error-no-module-named-fcntl). Possibly more.
* Vagrant machines not configured for Windows hosts, e.g. interfaces configured to write to Unix devices such as `/dev/null`.

##### Path 2: Virtual machine nested virtualization with Virtualbox

In path 2 an attempt was made with nested virtualisation in Virtualbox.

❌ Within path 2 a working development environment could _not_ be created successfully.

The setup of the virtual machines inside a virtual machine worked out ok. However, everything depends on the function of nested virtualisation. This can be tricky and very hardware dependend.

Problems were:
* If nested hardware virtualisation is not given, machine startup times are so slow, that the machine times out. If it would start it is likely unusably slow.
* At least for my hardware nested virtualisation did not work out, i.e. could not be properly activated as desired.
[...]

PR #10966, opened 2021-12-22
docs: commenting on untested OS will lead to frustration

Neudrino:
... especially when seen with respect to https://github.com/magma/magma/issues/10116.

Of course this can be different for the cloud deployment tooling in contrast to the developer tooling, however, it should be tested, made clear that the scope is different, and given the restrictions or requirements for it to work.

### DIFF
--- a/docs/readmes/basics/prerequisites.md
+++ b/docs/readmes/basics/prerequisites.md
@@ -166,14 +166,9 @@ git checkout v1.6
 git tag -l
 ```
 
-## Build/Deploy Tooling
+## Deployment Tooling
 
-We support building the AGW and Orchestrator on macOS and Linux host operating
-systems. Doing so on a Windows environment should be possible but has not been
-tested. You may prefer to use a Linux virtual machine if you are on a Windows
-host.
-
-First, follow the previous section on developer tools. Then, install some
+First, follow the previous section on [developer tools](#development-tools). Then, install some
 additional prerequisite tools.
 
 ### macOS


electronjoe:
Did this review slip through the cracks? What's the status.

docs/readmes/basics/prerequisites.md
```diff
@@ -168,10 +168,7 @@ git tag -l
 
 ## Build/Deploy Tooling
 
-We support building the AGW and Orchestrator on macOS and Linux host operating
-systems. Doing so on a Windows environment should be possible but has not been
-tested. You may prefer to use a Linux virtual machine if you are on a Windows
-host.
+We support building the AGW and Orchestrator on macOS and Linux host operating systems.
```
themarwhal: Should we explicitly say that Windows is not supported/tested?

Neudrino: I am not sure here. The heading is ambivalent, as it talks about "Build **and** Deploy" tooling. However, the below instructions are only about deployment and not about build, afaiks.

So if we are only talking about _deployments only_, the toolchain of the five tools given below, is likely to be available on Windows. But if not properly tested by the respective developers, I would not add untested speculations to the official documentation.
However, I doubt that _building_ is possible on Windows.

So we might even want to change the title of this section to `Deployment tools`.

Thoughts?

electronjoe: @themarwhal ping

themarwhal: o darn I think i forgot to actually submit the comment for this. :'(

I think it's reasonable to rename this section as Deployment tools for clarity. Other than that, lgtm!

electronjoe: I might lean towards explicitly noting that Windows use is not supported?

electronjoe (approved):

docs/readmes/basics/prerequisites.md
themarwhal: Add-on: I can't see Scott's original comment but I agree with him that we should explicitly say what is not supported. (building on Windows).

Neudrino: On second thought, the complete section about supported system is actually duplicated and should be removed, as it is clearly stated at the beginning of the document, what is supported and what not.
So I removed the unnecessary duplication such that we have a "single source of truth" [at the beginning of the document](https://docs.magmacore.org/docs/next/basics/prerequisites#operating-system).

@electronjoe @themarwhal If you agree to this notion, feel free to merge. :slightly_smiling_face:

themarwhal (approved):
lgtm! Thanks!

Neudrino:
Affected by #11350.

Issue #10434, opened 2021-11-18
[Qos]Configured Qos parameters are not getting updated in the respective IE's in PDUSessionResourceSetupRequest and PDU session establishment accept messages 


NeelimaKalpana:
Version:  1.7.0-1637044298-4402e275
Affected Component: Access Gateway
Affected Subcomponent: AGW service
Deployment Environment: Vagrant (AGW)

Describe the Issue
Configured Qos parameters(QFI, Priority(ARP), Preemption parameters) are not getting updated in the respective IE's in PDUSessionResourceSetupRequest and PDU session establishment accept messages 

To Reproduce
1. Configure the APN parameters with values other than default values from NMS-Dash board.
2. Bind the APN to subscriber. 
3. Check that Subscriber info is getting updated with the configured values
4. Trigger the PDU-Session and check that PDU is coming up successfully  5. Configured Qos parameters are not getting reflected in Qos related IE. 

**Expected behavior**
Configured Qos-Profile values are getting reflected in the Qos-IE in the PDUSessionResourceSetupRequest and PDU session establishment accept messages

**Screenshots**
APN info from NMS
![image.png](https://images.zenhubusercontent.com/6139f1931bcb9da2af1bf651/855507a8-c288-412f-9901-c9bc1f880a8a)

Subscriberinfo
![image.png](https://images.zenhubusercontent.com/6139f1931bcb9da2af1bf651/0c2ec51e-f9b4-4870-9b3e-05ce7c7e6d30)

Syslog 
![image.png](https://images.zenhubusercontent.com/6139f1931bcb9da2af1bf651/5f61a83f-285d-45d4-89a2-ff39e7f38b24)

PCAP:
![image.png](https://images.zenhubusercontent.com/6139f1931bcb9da2af1bf651/27ac3825-47d5-4c62-9ff9-9253ca3baf4a)

[https://app.zenhub.com/files/170803235/2aed76e9-e57e-4e5a-b732-4abd4e9a59bf/download](https://app.zenhub.com/files/170803235/2aed76e9-e57e-4e5a-b732-4abd4e9a59bf/download)

moinuddin1980:
In API SessionStateEnforcer::prepare_response_to_access()
We see that with the current implementation, the default value of Qos are getting set, if static Qos rule list are not configured.
From the pcap attached it is confirmed that all the Qos values are of default in nature.
So, the issue need fix from sessiond side also.

GANESH-WAVELABS:
[...]

PR #10972, opened 2021-12-22
fix(amf): Correct SubscriberID set as "IMSI +number" in amf

Rajeshs23:
Signed-off-by: Rajeshs23 <rajesh.sure@wavelabs.ai>
Fix:
Set all requests as a same pattern for IMSI value

## Summary
 Issue:
For 5G, requests are coming from AMF as a “number” without IMSI keyword and from UPF, coming as a “IMSI + number” So in SMF searching the session info from SessionStore getting failed.
fix:
Set all requests as a same pattern for IMSI value and in line with 4G.
## Test Plan
1.Unit Testing
2.Verified with Ueransim 
## Additional Information
Associated zenhub ticket is #10434 
ueransim_capture
unittest_results


ulaskozat (approved):
LGTM

PR #10974, opened 2021-12-22
fix(feg): refactor sbi client and config

uri200:

## Summary

In order to prepare for more 5g  services on FEG, we need to build a base Sbi client that can be reused by other services.
That is why we need to consider Notify Handler as part of the same service, and not a separate one.

This PR creates a sbi Client which includes a Notify server. This will allow us to use the same object for the client and for the handler of notifications. Shutdown and Start methods will be inherited from the Server struct so all the clients can use them.

## Test Plan

`./build.py -c`

## Additional Information


themarwhal (approved):

feg/gateway/sbi/mocks/mock_pcrf.go
```diff
@@ -0,0 +1,72 @@
...
+func (pcf *MockPcf) PostSmPoliciesSmPolicyIdDelete(ctx echo.Context, smPolicyId string) error {
+
+	return ctx.NoContent(http.StatusOK)
+}
+
+// PostSmPoliciesSmPolicyIdUpdate handles POST /sm-policies/{smPolicyId}/update
+func (pcf *MockPcf) PostSmPoliciesSmPolicyIdUpdate(ctx echo.Context, smPolicyId string) error {
+
+	return ctx.NoContent(http.StatusOK)
+}
```
uri200: we may be able to replace this with your mock pcrf

feg/gateway/sbi/sbi.go
```diff
@@ -15,51 +15,91 @@ package sbi
...
 
 	"github.com/labstack/echo/v4"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
 )
 
-type SbiServer struct {
-	Server     *echo.Echo
-	ListenAddr string
+type NotifierServer struct {
```
GANESH-WAVELABS, line 42: This can be also used as a N7 server for example in the Mock PCF. It is better to name it differently.

feg/gateway/services/n7_n40_proxy/n7_n40_proxy/main.go
```diff
@@ -47,11 +47,11 @@ func main() {
 	if err != nil {
 		glog.Fatalf("Error connecting to redis store from N7_N40 Proxy: %s", err)
 	}
-	policyClient, err := n7.NewN7Client(&n7config.Server)
+	policyClient, err := n7.NewN7ClientWithHandlers(n7config, cloudReg)
 	if err != nil {
-		glog.Fatalf("Creating N7 Client failed: %s", err)
+		glog.Fatalf("Creating N7 BaseClientWithNotifier failed: %s", err)
```
GANESH-WAVELABS, line 52: Should the log be `BaseClientWithNotifier`? Here N7 Client us getting created.

feg/gateway/sbi/sbi_test.go
```diff
@@ -18,93 +18,40 @@ import (
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"magma/feg/gateway/sbi"
-	n7_client "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
-	n7 "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControlServer"
-	sbi_common "magma/feg/gateway/sbi/specs/TS29571CommonData"
+	"magma/feg/gateway/sbi/mocks"
+	sbi_NpcfSMPolicyControl "magma/feg/gateway/sbi/specs/TS29512NpcfSMPolicyControl"
+	sbi_CommonData "magma/feg/gateway/sbi/specs/TS29571CommonData"
```
GANESH-WAVELABS, line 26: In some places `sbi_CommonData` is used and in some places `sbiCommonData`. Better to have a uniform convention.

feg/gateway/sbi/sbi.go
uri200: My reasoning here is that N7 server is nothing else than a NotifierServer. The difference is that the api already implements the methods.

In fact, I was tempted to use `BaseClientWithNotifier` for the mock PCF server. Same reason as before: I think those http endpoints are nothing else than a client+server under the hood

feg/gateway/sbi/sbi_test.go
uri200: you are right! let me fix that.

I think to avoid confusion and highlight the name lets use the _ one.
This we will know that when we have an `sbi_` that means this is  sbi autogenerated code

uri200: in fact we are just replacing TSnumber for sbi

feg/gateway/sbi/sbi.go
GANESH-WAVELABS: A Notifier Server is essentially a client, that gets notified by a server. But in case of a N7 Server for example Mock PCF, it will be acting as a pure HTTP server and not as a notification endpoint. It would be better to name this struct accordingly where it can be used by both SBI Client and also a SBI server.

uri200: But notifications are nothing else that handlers. The difference I see is N7 server has the handlers already prebuilt  by the sbi interface, but it is nothing else than an echo server with handlers

to me the action done here (client)
```
	c.NotifyServer.Server.POST(path.Join(basePath, "update"), c.postSmPolicyUpdateNotification)
	c.NotifyServer.Server.POST(path.Join(basePath, "terminate"), c.postSmPolicyTerminateNotification)
	return nil
```

Is the same as the action done here  (mock PCF)
```
	sbi_NpcfSMPolicyControlServer.RegisterHandlers(mockPcf.Server, mockPcf)
```

The difference is on the server you use the autogenerated API and for the client you do it manually. But booth are getting `notified`

(see [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fsbi%2Fspecs%2FTS29512NpcfSMPolicyControlServer%2FTS29512NpcfSMPolicyControlServer.gen.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=1650&start_col=4&end_row=1650&end_col=4) how the server code looks like the same as the client)

GANESH-WAVELABS: REST APIs (also SBI specs) generally makes distinction between REST API requests served and notifications (callbacks) served, though both are HTTP Request handles.  REST API requests handlers are REST Servers and the Notification or Callback handlers are part of REST Client. This piece of code can be used as both REST Server and a REST client (for the notification handler). Naming it as NotifiyServer is confusing (atleast for me). We should probably go a layer below and name it as HTTPServer or HTTPHandler or RESTHandler.

uri200: Yes yesterday I was considering one layer below thing too. But then things become more complicated, with start/stop functions not working anymore as they are and having to pass parameters

Let me try it again

PR #10976, opened 2021-12-22

spikey979:

## Summary

function works as follows:
it remove line breaks,
it remove extra whitespace

## Test Plan
ran /magma/orc8r/cloud/docker/  ./build.py -c

For code scanning alerts:
https://github.com/magma/magma/security/code-scanning/35?query=ref%3Arefs%2Fheads%2Fmaster
https://github.com/magma/magma/security/code-scanning/36?query=ref%3Arefs%2Fheads%2Fmaster

### DIFF
--- a/fbinternal/cloud/go/services/download/servicers/server_lib.go
+++ b/fbinternal/cloud/go/services/download/servicers/server_lib.go
@@ -2,6 +2,7 @@ package servicers
 
 import (
 	"fmt"
+	"html"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


uri200:
I think we can ignore this warning for now
I will consult with the team but I think this is something the linter has introduced recently. So we will have to look into it.


image

andreilee:
@uri200 Was there a consensus? To weigh in, I don't see any reason for us not to sanitize the user input here.

uri200:
Note this seem to have come from a linter error.

Not sure how legit this linter error is. And not sure how many more instances like that we will have.

The log is just printing the content of the variable so we can see the key value. We should t modify that content since then it may mean we are displaying a different key.

spikey979:
On the recommendation of the reviewer, I will close PR as a false positive.

PR #10981, opened 2021-12-23
feat(feg): N7 interface PCF stub

GANESH-WAVELABS:

## Summary
Mock PCF server implementation to handle N7 requests from PCF and initiate N7 notifications towards feg.


feg/gateway/services/testcore/pcf/servicers/mock_driven_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+func createAccuUsageReport() string {
+	usageReport := `{
+		"refUmIds": "%s",
+		"volUsageUplink": %d,
+		"volUsageDownlink": %d,
+		"volUsage": %d
+	}`
+	return fmt.Sprintf(usageReport, UM_DATA1, UsageTx1, UsageRx1, UsageTotal1)
+}
```
uri200, line 287: Note the mock driven tests was thought as a quick way to start running cwag integ test. 
I think it would be more valuable if we could build a working PCF, were we can load subscribers/profiles/rules though grpc calls instead of just a mock one.

Note I am not saying all features, but at least we should be able to load PCF with rules for a specific subscriber and send them upon create session. Update of monitor data could also be done since we already have the code in mock PCRF and OCS. Here in this case, you can copy the code if you want since they are test services that may differ in the future

GANESH-WAVELABS: GRPC based subscriber/rule additions similar to mock PCRF are supported. Will have to add a unit test for that.

uri200 (approved):

PR #10987, opened 2021-12-23

### LINKED ISSUE #10986: Fix code scanning alert - Log entries created from user input
Tracking issue for:
- [ ] https://github.com/magma/magma/security/code-scanning/37

nikolakn:
## Summary

issue: https://github.com/magma/magma/issues/10986
GitHub Code scanning alerts are created because of using URL  in log directly: "If unsanitized user input is written to a log entry, a malicious user may be able to forge new log entries."
If the log is displayed as HTML, then arbitrary HTML may be included to spoof log entries.

## Test Plan

go build and go test in  orc8r/cloud/go/obsidian

### DIFF
--- a/lte/gateway/deploy/roles/dev_common/tasks/main.yml
+++ b/lte/gateway/deploy/roles/dev_common/tasks/main.yml
@@ -222,13 +222,14 @@
     state: link
     force: yes
 
-- name: Set up environment variables for VSCode integration
-  lineinfile:
-    # We set this variable in .profile so that VSCode can determine which build config to use 
-    path: /home/{{ ansible_user }}/.profile
-    line: "export VSCODE_BAZEL_BUILD_CONFIG=vm"
+- name: Install lld
+  retries: 5
+  # TODO: Make this preburn
+  apt:
     state: present
-  when: full_provision
+    update_cache: no
+    pkg:
+      - lld
 
 ########################################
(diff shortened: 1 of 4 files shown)


lte/gateway/deploy/roles/dev_common/tasks/main.yml
```diff
@@ -222,13 +222,14 @@
     state: link
     force: yes
 
-- name: Set up environment variables for VSCode integration
-  lineinfile:
-    # We set this variable in .profile so that VSCode can determine which build config to use 
-    path: /home/{{ ansible_user }}/.profile
-    line: "export VSCODE_BAZEL_BUILD_CONFIG=vm"
+- name: Install lld
```
andreilee, line 225: Why did this need to be added?

hcgatewood (changes requested):
1. Don't think the lld install is part of this PR
2. GH code scanning still picks this up as an issue... ideally we'd deal with that warning
3. Idk if this is the best way to sanitize the logged string... in my mind, I think we can just (a) remove all newlines and (b) cap the len to something long but reasonable. Correct me if I'm wrong but I think this would provide the necessary sanitation, without making the url look weird

In any case, would love to see a filled out test plan, as it's not clear how this escaping changes things, or e.g. what types of malicious inputs we're now protected from

PR #11010, opened 2021-12-28
feat(agw): EXPERIMENTAL: introduce eBPF datapath

pshelar:
Following PR adds infrastructure to build eBPF based datapath for AGW.

1. Basic uplink traffic handler using eBPF.
2. Pipelined integration is done by ebpf_managaer
3. Packaging is updated to add eBPF lib dependency
4. bcc-tools packaging script is added to build latest bcc-tools
   debian package for ubuntu
5. unit test infra along with unit test is added to validate UL traffic
   Handler.

At the moment this feature is intended for developer only.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
+++ b/lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
@@ -1,0 +1,219 @@
+"""
+Copyright 2021 The Magma Authors.
+
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import ctypes
+import logging
+import socket
+import struct
+import subprocess
+from builtins import input
+from socket import AF_INET, htons
+from subprocess import call
+from sys import argv
+from threading import Thread
+
+import netifaces
+from bcc import BPF
+from magma.pipelined.mobilityd_client import get_mobilityd_gw_info
+from pyroute2 import IPDB, IPRoute, NetlinkError, NetNS, NSPopen
+from scapy.layers.l2 import getmacbyip
+
+LOG = logging.getLogger("pipelined.ebpf")
+# LOG.setLevel(logging.DEBUG)
+
+BASE_MAP_FS = "/sys/fs/bpf/"
+BPF_UL_FILE = "/var/opt/magma/ebpf/ebpf_ul_handler.c"
+UL_MAP_NAME = "ul_map"
+
+"""
+    Use pipelineD configuration to initialize eBPF manager for AGW.
+"""
+
+
+def get_ebpf_manager(config):
+
+    if 'ebpf' in config:
+        enabled = config['ebpf']['enabled']
+    else:
+        enabled = False
+    gw_info = get_mobilityd_gw_info()
+    for gw in gw_info:
+        if gw.vlan == "":
+            bpf_man = ebpf_manager(config['nat_iface'], config['enodeb_iface'], gw.ip, enabled)
+            if enabled:
+                # TODO: For Development purpose dettch and attach latest eBPF code.
+                # Remove this for production deployment
+                bpf_man.detach_ul_ebpf()
+                bpf_man.attach_ul_ebpf()
+                LOG.info("eBPF manager: initilized: enabled: %s", enabled)
+            return bpf_man
+
+    LOG.info("eBPF manager: Not initilized")
+    return None
+
+
+"""
+    eBPF manager for AGW.
+    Initialize eBPF based datapath for AGW as per the pipelineD config.
+
+    Returns:
+        eBPF manager object.
+"""
+
+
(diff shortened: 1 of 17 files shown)


lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
```diff
@@ -0,0 +1,222 @@
...
+
+    """Attach eBPF Uplink traffic handler
+    """
+
+    def attach_ul_ebpf(self):
+        s1_if_index = self._get_ifindex(self.s1_if_name)
+
+        ipr = IPRoute()
+        try:
+            ipr.tc("add", "ingress", s1_if_index, "ffff:")
```
koolzz: we can also explore this type of filter
```
    ipr.tc("add", "clsact", ifindex)
    action = {"kind": "bpf", "fd": fn.fd, "name": fn.name, "action": "ok"}
    ipr.tc(
        "add-filter", "u32", ifindex, ":1", parent="ffff:fff2", action=[action],
        protocol=protocols.ETH_P_ALL, classid=1, target=0x10002, keys=['0x0/0x0+0'],
    )
```
Also adding action that way allow to control the packet path directly, we would be able to do `return bpf_redirect(if_idx,0);` instead of doing bpf clone redirect

lte/gateway/python/magma/pipelined/ebpf/ebpf_ul_handler.c
```diff
@@ -0,0 +1,110 @@
...
+};
+
+struct ul_map_info {
+  u32 mark;
+  u32 e_if_index;
+  u8  mac_src[ETH_ALEN];
+  u8  mac_dst[ETH_ALEN];
+};
+
+BPF_TABLE_PINNED("hash", struct ul_map_key, struct ul_map_info, ul_map, 1024*512, "/sys/fs/bpf/ul_map");
```
koolzz: Also I don't think we need to pin the map here, pinning is just when we need to share the map in other ebpf programs or other services that need to modify it. Here we only use pipelined to start/modify it afaik?

lte/gateway/python/magma/pipelined/ebpf/ebpf_ul_handler.c
```diff
@@ -0,0 +1,110 @@
...
+int gtpu_get_tunnel(struct __sk_buff *skb)
```
koolzz: maybe a more generic name like handle_ingress_pkt/handle_egress_pkt ?

lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
pshelar: switched to clsact.

lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
```diff
@@ -0,0 +1,224 @@
...
+
+bpf_man = None
+
+"""
+    Use pipelineD configuration to initialize eBPF manager for AGW.
+"""
+
+
+def get_ebpf_manager(config):
+    global bpf_man
```
koolzz: why is this global, seems like we can only init it through policy mixin?

lte/gateway/python/magma/pipelined/policy_converters.py
```diff
@@ -62,6 +62,20 @@ def get_ipv6_match_value(ip_addr) -> str:
     return get_ipv6_prefix(str(ip_addr.address.decode('utf-8'))) + "/" + IPV6_PREFIX_LEN
 
 
+def is_flow_match_set(match):
```
koolzz, line 65: nice

lte/gateway/python/magma/pipelined/ebpf/ebpf_ul_handler.c
pshelar: I do expect other services or debugging tools examining the datapath state, so it needs to be pinned.

lte/gateway/python/magma/pipelined/ebpf/ebpf_ul_handler.c
pshelar: I would rename it to gtpu_ingress_handler()

lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
pshelar: policy mixin init is called multiple times, this get_ebpf_manager() return cached bpf_man ref in case of already initialized.

koolzz: Ah right, should we init it when we create the global objects like rulemapper?

koolzz: And pass it around through kwargs

lte/gateway/python/magma/pipelined/ebpf/ebpf_ul_handler.c
koolzz: Good point

koolzz (approved):
Lgtm some things could probably be cleaned up but as it's experimental let's merge it quicker so we can both work on it

lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
pshelar: done

PR #11011, opened 2021-12-28
chore(agw): Backport consolidation of sentry-cli and sentry_sdk releases

alexzurbonsen:
## Summary

Backport PR #10832.

And #9319, which is a prerequisite for #10832. 

## Test Plan

Tested the new shell commands on VM bash locally.

## Additional Information

### DIFF
--- a/orc8r/tools/fab/pkg.py
+++ b/orc8r/tools/fab/pkg.py
@@ -131,11 +131,11 @@ def upload_pkgs_to_aws():
 
 
 def get_magma_version():
-    return run('ls ~/magma-packages'
-               ' | grep "^magma_[0-9].*"'
-               ' | xargs -I "%" dpkg -I ~/magma-packages/%'
-               ' | grep "Version"'
-               ' | awk \'{print $2}\'')
+    return run('directory=$(mktemp -d) &&'
+               'dpkg-deb --extract ~/magma-packages/magma_[0-9]*.deb $directory &&'  # noqa: E501
+               'source $directory/usr/local/share/magma/commit_hash &&'
+               'echo $COMMIT_HASH',
+               )
 
 
 def copy_packages():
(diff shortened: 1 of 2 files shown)


jheidbrink (approved):

orc8r/tools/fab/pkg.py
```diff
@@ -131,11 +131,11 @@ def upload_pkgs_to_aws():
 
 
 def get_magma_version():
-    return run('ls ~/magma-packages'
-               ' | grep "^magma_[0-9].*"'
-               ' | xargs -I "%" dpkg -I ~/magma-packages/%'
-               ' | grep "Version"'
-               ' | awk \'{print $2}\'')
+    return run('directory=$(mktemp -d) &&'
+               'dpkg-deb --extract ~/magma-packages/magma_[0-9]*.deb $directory &&'
```
Neudrino: Maybe we can ignore this one. https://wemake-python-stylegui.de/en/0.16.0/pages/usage/configuration.html#configuration

Neudrino (commented):
While linter warnings should be fixed in general, maybe we can add a tag to ignore this specific one.

orc8r/tools/fab/pkg.py
alexzurbonsen: Sure. Inserted an inline-comment.

Neudrino (approved):

themarwhal (approved):

themarwhal:
Just making sure, this change is needed to achieve semantic versioning for 1.6 changes?

alexzurbonsen:
> Just making sure, this change is needed to achieve semantic versioning for 1.6 changes?

Strictly speaking: no. But it unifies the versions that appear in sentry. Otherwise the releases created by sentry-cli have a different format than the ones created by sentr_sdk.

So this PR is mainly tidying up the releases that are send to sentry, to avoid having a confusing mismatch in the sentry GUI between both release types.

quentinDERORY (approved):

Neudrino:
Could we get a review here, please? @emakeev 
Can we merge this than? @themarwhal @electronjoe

PR #11039, opened 2022-01-01
fix(orc8r): Setup IAM Role for cert-manager

ShubhamTatvamasi:

## Summary

When we first integrated cert-manager with Terraform we were generating secret keys for adding TXT records in Route53 which was not a secure method.
Now we are moving to IAM Role for accessing Route53 which better way of doing this.

## Test Plan

First we will have to deploy Orc8r using the code from this PR:
```bash
terraform apply -target=module.orc8r
terraform apply
```

Then we need to check if we are able to generate `orc8r-nms-tls` secrets from let's encrypt.

## Additional Information

We need to be sure the NS records are updated before we apply orc8r-app terraform module.


andreilee:
Could you add a summary and test plan?

ShubhamTatvamasi:
> Could you add a summary and test plan?

Just updated.

orc8r/cloud/deploy/terraform/orc8r-aws/cert-manager.tf
```diff
@@ -0,0 +1,75 @@
+resource "random_id" "cert_manager_route53_random_id" {
```
hcgatewood, line 1: Can you add an explanatory comment for why including the random ID -- is it to ensure uniqueness within an AWS account? If yes, then a 1/256 chance of a "bug" with 2 Orc8rs feels a bit high (?)

In a future PR perhaps we could tie this to the Orc8r UUID... did we end up getting that into the Terraform @arunuke ?

hcgatewood (approved):
💯💯

orc8r/cloud/deploy/terraform/orc8r-aws/cert-manager.tf
arunuke: Yes, Magma UUIDs are now a thing that can be set when deploying via terraform (default value is....'default') and passed on to be set as additional annotations across various services deployed.

ShubhamTatvamasi: IAM Roles is a Global AWS service. If we have another Orc8r in a different region it can conflict with the same role name.
I am looking into the Magma UUID code and trying to update it here.

ShubhamTatvamasi: So if we use Magma UUID the user will have to know about this feature and update it manually on the second orc8r deployment. But if we use Random ID then there is no need for it because terraform will take care of this.
What are your suggestions for this @hcgatewood and @arunuke?
https://github.com/magma/magma/blob/master/orc8r/cloud/deploy/terraform/orc8r-aws/variables.tf#L47

hcgatewood: I suppose the middle ground is we automatically create a uuid for the orc8r if the user doesn't provide one? That way we get the benefit of this random solution, but remove the probability of conflict. Thoughts?

PR #11046, opened 2022-01-04
feat(orc8r): FEG_RELAY update to redirect session request to N7_N40 Proxy

shivesh-wavelabs:
Signed-off-by: shivesh-wavelabs <shivesh.ojha@wavelabs.ai>

## Summary
FEG_RELAY code changes to redirect sessiond request to FegN7N40PROXY based on the RAT-TYPE

## Test Plan
All UTs passed.

### DIFF
--- a/feg/cloud/configs/service_registry.yml
+++ b/feg/cloud/configs/service_registry.yml
@@ -36,6 +36,8 @@ services:
         port: 9103
       session_proxy:
         port: 9079
+      n7_n40_proxy:
+        port: 9103
       swx_proxy:
         port: 9103
       csfb:
--- a/feg/cloud/go/services/feg_relay/gw_to_feg_relay/tests/relay_test.go
+++ b/feg/cloud/go/services/feg_relay/gw_to_feg_relay/tests/relay_test.go
@@ -31,6 +31,7 @@ import (
 	"magma/feg/cloud/go/services/feg_relay/gw_to_feg_relay"
 	servicers "magma/feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/southbound"
 	healthTestUtils "magma/feg/cloud/go/services/health/test_utils"
+	lte_protos "magma/lte/cloud/go/protos"
 	"magma/orc8r/cloud/go/orc8r"
 	"magma/orc8r/cloud/go/services/configurator"
 	"magma/orc8r/cloud/go/services/device"
@@ -91,6 +92,147 @@ func (tp *testHelloServer) SayHello(c context.Context, req *feg_protos.HelloRequ
 	}, nil
 }
 
+type testCentralSessionController struct {
+	lte_protos.UnimplementedCentralSessionControllerServer
+	resultChan chan string // Calling FeG ID string on success
+}
+
+func (tp *testCentralSessionController) UpdateSession(
+	ctx context.Context,
+	req *lte_protos.UpdateSessionRequest) (*lte_protos.UpdateSessionResponse, error) {
+
+	var res *lte_protos.UpdateSessionResponse
+	var ratType lte_protos.RATType
+
+	if tp == nil {
+		return nil, fmt.Errorf("nil test Central Session Controller")
+	}
+	if tp.resultChan == nil {
+		return nil, fmt.Errorf("nil test Central Session Controller channel")
+	}
+	var targetFegId = "<MISSING METADATA>"
+	ctxMetadata, ok := metadata.FromIncomingContext(ctx)
+	if ok && ctxMetadata != nil {
+		targetFegId = "<MISSING GW ID>"
+		values, ok := ctxMetadata[gateway_registry.GatewayIdHeaderKey]
+		if !ok {
+			values, ok = ctxMetadata[strings.ToLower(gateway_registry.GatewayIdHeaderKey)]
+		}
+		if ok && len(values) > 0 {
+			targetFegId = values[0]
+		}
+	}
+	tp.resultChan <- targetFegId
+
+	if len(req.Updates) > 0 {
+		ratType = req.Updates[0].GetCommonContext().GetRatType()
+	} else if len(req.UsageMonitors) > 0 {
+		ratType = req.UsageMonitors[0].GetRatType()
+	} else {
+		return nil, fmt.Errorf("empty data in UpdateSessionRequest")
+	}
+
+	if ratType == lte_protos.RATType_TGPP_NR {
+		res = &lte_protos.UpdateSessionResponse{
+			Responses: []*lte_protos.CreditUpdateResponse{
+				createCreditUsageResponse(nhImsi2, 2),
+			},
+			UsageMonitorResponses: []*lte_protos.UsageMonitoringUpdateResponse{
+				createUsageMonitoringResponse(nhImsi2, "mkey2"),
+				createUsageMonitoringResponse(nhImsi2, "mkey3"),
+			},
+		}
+	} else {
+		res = &lte_protos.UpdateSessionResponse{
+			Responses: []*lte_protos.CreditUpdateResponse{
+				createCreditUsageResponse(nhImsi, 1),
+			},
+			UsageMonitorResponses: []*lte_protos.UsageMonitoringUpdateResponse{
+				createUsageMonitoringResponse(nhImsi, "mkey"),
+			},
+		}
+	}
+
+	return res, nil
+}
+
+func createCreditUsageRequest(
+	imsi string,
+	chargingKey uint32,
+	requestNumber uint32,
+	requestType lte_protos.CreditUsage_UpdateType,
+	ratType lte_protos.RATType,
+) *lte_protos.CreditUsageUpdate {
+	return &lte_protos.CreditUsageUpdate{
+		Usage: &lte_protos.CreditUsage{
+			BytesTx:     1024,
+			BytesRx:     2048,
+			ChargingKey: chargingKey,
+			Type:        requestType,
+		},
+		SessionId:     imsi,
+		RequestNumber: requestNumber,
+		CommonContext: &lte_protos.CommonSessionContext{
+			Sid: &lte_protos.SubscriberID{
+				Id: imsi,
+			},
+			RatType: ratType,
+		},
+	}
+}
+
+func createCreditUsageResponse(
+	imsi string,
+	chargingKey uint32,
+) *lte_protos.CreditUpdateResponse {
+	return &lte_protos.CreditUpdateResponse{
+		Success:     true,
+		SessionId:   genSessionID(imsi),
+		Sid:         imsi,
+		ChargingKey: chargingKey,
+	}
+}
+
+func createUsageMonitoringRequest(
+	imsi string,
+	monitoringKey string,
+	requestNumber uint32,
+	monitoringLevel lte_protos.MonitoringLevel,
+	ratType lte_protos.RATType,
+) *lte_protos.UsageMonitoringUpdateRequest {
+	return &lte_protos.UsageMonitoringUpdateRequest{
+		Update: &lte_protos.UsageMonitorUpdate{
+			BytesTx:       1024,
+			BytesRx:       2048,
+			MonitoringKey: []byte(monitoringKey),
+			Level:         monitoringLevel,
+		},
+		SessionId:     genSessionID(imsi),
+		RequestNumber: requestNumber,
+		Sid:           imsi,
+		EventTrigger:  lte_protos.EventTrigger_USAGE_REPORT,
+		RatType:       ratType,
+	}
+}
+
+func createUsageMonitoringResponse(
+	imsi string,
+	monitoringKey string,
+) *lte_protos.UsageMonitoringUpdateResponse {
+	return &lte_protos.UsageMonitoringUpdateResponse{
+		Success:   true,
+		SessionId: genSessionID(imsi),
+		Sid:       imsi,
+		Credit: &lte_protos.UsageMonitoringCredit{
+			MonitoringKey: []byte(monitoringKey),
+		},
+	}
+}
+
+func genSessionID(imsi string) string {
+	return fmt.Sprintf("%s-1234", imsi)
+}
+
 func TestNHRouting(t *testing.T) {
 	testHealthServiser := setupNeutralHostNetworks(t)
 
@@ -113,6 +255,9 @@ func TestNHRouting(t *testing.T) {
 	// Register FeG's test Hello Server
 	feg_protos.RegisterHelloServer(srv.GrpcServer, &testHelloServer{})
 
+	// Register FeG's Central Session Controller Server
+	sessionProxy := &testCentralSessionController{resultChan: make(chan string, 2)}
+	lte_protos.RegisterCentralSessionControllerServer(srv.GrpcServer, sessionProxy)
 	go srv.RunTest(lis)
 
 	// Add Serving FeG Host to directoryd
@@ -128,6 +273,7 @@ func TestNHRouting(t *testing.T) {
 	relayRouter := servicers.NewRelayRouter()
 	feg_protos.RegisterS6AProxyServer(relaySrv.GrpcServer, relayRouter)
 	feg_protos.RegisterHelloServer(relaySrv.GrpcServer, relayRouter)
+	lte_protos.RegisterCentralSessionControllerServer(relaySrv.GrpcServer, relayRouter)
 	go relaySrv.RunTest(relayLis)
 
 	ctx := service_test_utils.GetContextWithCertificate(t, agwHwId)
@@ -149,6 +295,37 @@ func TestNHRouting(t *testing.T) {
 	case <-time.After(3 * time.Second):
 		t.Fatal("Neutral Host Routed S6a Proxy Call timed out")
 	}
+
+	// Test UpdateSession routing
+	sessProxyClient := lte_protos.NewCentralSessionControllerClient(conn)
+	toutctx, cancel = context.WithTimeout(ctx, 5*time.Second)
+	updateSessReq := &lte_protos.UpdateSessionRequest{
+		Updates: []*lte_protos.CreditUsageUpdate{
+			createCreditUsageRequest(nhImsi, 1, 1, lte_protos.CreditUsage_QUOTA_EXHAUSTED, lte_protos.RATType_TGPP_LTE),
+			createCreditUsageRequest(nhImsi2, 2, 2, lte_protos.CreditUsage_TERMINATED, lte_protos.RATType_TGPP_NR),
+		},
+		UsageMonitors: []*lte_protos.UsageMonitoringUpdateRequest{
+			createUsageMonitoringRequest(nhImsi, "mkey", 1, lte_protos.MonitoringLevel_SESSION_LEVEL, lte_protos.RATType_TGPP_LTE),
+			createUsageMonitoringRequest(nhImsi2, "mkey2", 2, lte_protos.MonitoringLevel_PCC_RULE_LEVEL, lte_protos.RATType_TGPP_NR),
+			createUsageMonitoringRequest(nhImsi2, "mkey3", 3, lte_protos.MonitoringLevel_SESSION_LEVEL, lte_protos.RATType_TGPP_NR),
+		},
+	}
+
+	res, err := sessProxyClient.UpdateSession(toutctx, updateSessReq)
+	cancel()
+	assert.NoError(t, err)
+
+	// Based on RATType, feg_relay UpdateSession splits and relays the request to session_proxy and n7_n40_proxy
+	// in serving feg. As testCentralSessionController UpdateSession is executed twice,
+	// verifying the servingFegHwId twice
+	for i := len(sessionProxy.resultChan); i > 0; i-- {
+		servingFegHwId := <-sessionProxy.resultChan
+		assert.Equal(t, fegHwId, servingFegHwId)
+	}
+
+	// Verifying the collective response from session_proxy and n7_n40_proxy of serving Feg
+	assert.Equal(t, len(updateSessReq.Updates), len(res.Responses))
+	assert.Equal(t, len(updateSessReq.UsageMonitors), len(res.UsageMonitorResponses))
 
 	// Test SayHello routing & NH argumentation regex
 	helloClient := feg_protos.NewHelloClient(conn)
--- a/feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/southbound/relay_router.go
+++ b/feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/southbound/relay_router.go
@@ -29,6 +29,7 @@ const (
 	FegSessionProxy gateway_registry.GwServiceType = "session_proxy"
 	FegHello        gateway_registry.GwServiceType = "feg_hello"
 	FegSwxProxy     gateway_registry.GwServiceType = "swx_proxy"
+	FegN7N40Proxy   gateway_registry.GwServiceType = "n7_n40_proxy"
 )
 
 // RelayRouter implements generic routing logic and currently just embeds gw_to_feg_relay.Router functionality
--- a/feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/southbound/session_proxy_relay.go
+++ b/feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/southbound/session_proxy_relay.go
@@ -19,17 +19,18 @@ import (
 	"github.com/golang/glog"
 
 	"magma/lte/cloud/go/protos"
+	"magma/orc8r/cloud/go/services/dispatcher/gateway_registry"
 )
 
 // SessionProxyServer implementation
 //
-// Notify OCS/PCRF of new session and return rules associated with subscriber
+// CreateSession Notifies OCS/PCRF of new session and return rules associated with subscriber
 // along with credits for each rule
 func (s *RelayRouter) CreateSession(
 	ctx context.Context, r *protos.CreateSessionRequest) (*protos.CreateSessionResponse, error) {
 
 	// CreateSession's SID is just IMSI with "IMSI" prefix which findServingNHFeg() should remove
-	client, ctx, cancel, err := s.getSessionProxyClient(ctx, r.GetCommonContext().GetSid().GetId())
+	client, ctx, cancel, err := s.getSessionControllerProxyClient(ctx, getSessionControllerProxyType(r.GetCommonContext().GetRatType()), r.GetCommonContext().GetSid().GetId())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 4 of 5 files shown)


feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -71,33 +73,45 @@ func (s *RelayRouter) UpdateSession(
...
+		}
 	}
+
 	// Combine received responses
 	resp := &protos.UpdateSessionResponse{
 		Responses:             []*protos.CreditUpdateResponse{},
 		UsageMonitorResponses: []*protos.UsageMonitoringUpdateResponse{},
 	}
-	for i := len(reqMap); i > 0; i-- {
+	for i := len(reqMap) * 2; i > 0; i-- {
```
anusha-konda, line 114: 2 - can be replaced with len(proxyMap)

shivesh-wavelabs: replaced

uri200:
I think we will need to merge this first and test it 
Right now we are not using feg_relay code for session proxy, that is why it uses a different port. 
Switching session_proxy ti 9103 will force the code to use the end points [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fcloud%2Fgo%2Fservices%2Ffeg_relay%2Fgw_to_feg_relay%2Fservicers%2Fsession_proxy_relay.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=27&start_col=25&end_row=27&end_col=25)

https://github.com/magma/magma/pull/11080

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -111,18 +135,17 @@ func (s *RelayRouter) TerminateSession(
...
 	defer cancel()
 	return client.TerminateSession(ctx, r)
 }
 
-func (s *RelayRouter) getSessionProxyClient(
-	c context.Context, imsi string) (protos.CentralSessionControllerClient, context.Context, context.CancelFunc, error) {
-
-	conn, ctx, cancel, err := s.GetFegServiceConnection(c, imsi, FegSessionProxy)
+// Get Proxy Client based on RatType
+func (s *RelayRouter) getProxyClient(c context.Context, r protos.RATType, imsi string) (protos.CentralSessionControllerClient, context.Context, context.CancelFunc, error) {
```
emakeev: Please use more specific naming for RelayRouter methods, relay router is shared between multiple FeG Proxies, they are all proxies - swx, s6a, session_proxy... getProxyClient may imply any of them and potentially conflict with any of them

feg/cloud/configs/service_registry.yml
```diff
@@ -36,6 +36,8 @@ services:
         port: 9103
       session_proxy:
         port: 9079
+      n7_n40_proxy:
```
emakeev, line 39: you don't have n7 n40 proxy router itself, I don't think - this mapping does anything/needed

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -71,33 +72,56 @@ func (s *RelayRouter) UpdateSession(
 			req.UsageMonitors = append(req.UsageMonitors, m)
 		}
 	}
-	resultChan := make(chan *protos.UpdateSessionResponse, len(reqMap))
+
+	/*Each request in reqMap is segregated into two requests and relays each request to respective proxies (session_proxy and n7n40_proxy).
+	Hence length of resultChan is len(reqMap) * 2
+	*/
```
emakeev: please format your comments similar to the rest of codebase

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -71,33 +72,56 @@ func (s *RelayRouter) UpdateSession(
...
-				glog.Errorf("failed Session Proxy Update for PLMNID '%s': %v", plmnid, err)
-				resultChan <- genUpdateErrorResp(req)
-				return
+		for i := 0; i < len(req.Updates); i++ {
+			if req.Updates[i].CommonContext.RatType == protos.RATType_TGPP_NR {
+				n7n40ProxyReq.Updates = append(n7n40ProxyReq.Updates, req.Updates[i])
+				n7n40ProxyReq.UsageMonitors = append(n7n40ProxyReq.UsageMonitors, req.UsageMonitors[i])
+			} else {
+				sessionProxyReq.Updates = append(sessionProxyReq.Updates, req.Updates[i])
+				sessionProxyReq.UsageMonitors = append(sessionProxyReq.UsageMonitors, req.UsageMonitors[i])
```
emakeev: How is it going to work???

sessionProxyReq AND n7n40ProxyReq are both nil at this point. You are going to crash this service as soon as you reference a field of any of them.

How was this PR tested?

emakeev (changes requested):
This PR has several issues. There is a nil pointer reference which would crash the relay service (please see inline comments).
Also, session_proxy (as of this PR) requests do not get routed via NH capable relay, so the handlers will not get invoked.
PR https://github.com/magma/magma/pull/11080 should enable it, but it hasn't been merged yet.

And the last thing, relay routing is probably the most complicated part of FeG & Magma, session_proxy relay routing was not enabled until now and wasn't tested in production.
I'd suggest to enable session_proxy NH routing first (PR 11080), deploy & test it as is first and only after that start making changes to support n7 n40.

anusha-konda:
> This PR has several issues. There is a nil pointer reference which would crash the relay service (please see inline comments). Also, session_proxy (as of this PR) requests do not get routed via NH capable relay, so the handlers will not get invoked. PR #11080 should enable it, but it hasn't been merged yet.
> 
> And the last thing, relay routing is probably the most complicated part of FeG & Magma, session_proxy relay routing was not enabled until now and wasn't tested in production. I'd suggest to enable session_proxy NH routing first (PR 11080), deploy & test it as is first and only after that start making changes to support n7 n40.

Thanks for reviewing the PR. 
@emakeev , There is certain gap and confusion here. As you suggested will verify first with PR #11080 . And later enhance the current PR code to support n7 n40.

feg/cloud/configs/service_registry.yml
uri200: Per what we discussed with @emakeev  we may need the router since we will have to check request parametrs (same router check IMSI on a request)

uri200:
Hey @anusha-konda, note right now feg_relay is just forwarding the request. The code here [UpdateSession](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fcloud%2Fgo%2Fservices%2Ffeg_relay%2Fgw_to_feg_relay%2Fservicers%2Fsession_proxy_relay.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=40&start_col=22&end_row=40&end_col=22), is never used until #11080  is landed. Since 5g will use that code, could you try to see if you can add a test to check that function ?  

You can see a similar test done on other services (feg_hello and s6a) [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fcloud%2Fgo%2Fservices%2Ffeg_relay%2Fgw_to_feg_relay%2Ftests%2Frelay_test.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=93&start_col=12&end_row=93&end_col=12)

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
shivesh-wavelabs: Corrected. Verified with UT

anusha-konda:
> Hey @anusha-konda, note right now feg_relay is just forwarding the request. The code here [UpdateSession](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fcloud%2Fgo%2Fservices%2Ffeg_relay%2Fgw_to_feg_relay%2Fservicers%2Fsession_proxy_relay.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=40&start_col=22&end_row=40&end_col=22), is never used until #11080 is landed. Since 5g will use that code, could you try to see if you can add a test to check that function ?
> 
> You can see a similar test done on other services (feg_hello and s6a) [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fcloud%2Fgo%2Fservices%2Ffeg_relay%2Fgw_to_feg_relay%2Ftests%2Frelay_test.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=93&start_col=12&end_row=93&end_col=12)

@uri200 , Added UT for UpdateSession. In the UT, verified servingFegHwId from gRPC incoming context and verified the collective response of UpdateSession responses.
Please review it.

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
shivesh-wavelabs: Done.

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
shivesh-wavelabs: Done.

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -71,33 +72,95 @@ func (s *RelayRouter) UpdateSession(
...
+			if pu.CommonContext.RatType == protos.RATType_TGPP_NR {
+				n7n40ProxyReq, ok := proxyMap[FegN7N40Proxy]
+				if !ok || n7n40ProxyReq == nil {
+					n7n40ProxyReq = &protos.UpdateSessionRequest{
+						Updates:       []*protos.CreditUsageUpdate{},
+						UsageMonitors: []*protos.UsageMonitoringUpdateRequest{}}
+					proxyMap[FegN7N40Proxy] = n7n40ProxyReq
+				}
+				n7n40ProxyReq.Updates = append(n7n40ProxyReq.Updates, pu)
+			} else {
```
uri200: in the previous loop you already constructed this object below

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -71,33 +72,95 @@ func (s *RelayRouter) UpdateSession(
...
+				sessionProxyReq, ok := proxyMap[FegSessionProxy]
+				if !ok || sessionProxyReq == nil {
+					sessionProxyReq = &protos.UpdateSessionRequest{
+						Updates:       []*protos.CreditUsageUpdate{},
+						UsageMonitors: []*protos.UsageMonitoringUpdateRequest{}}
+					proxyMap[FegSessionProxy] = sessionProxyReq
+				}
+				sessionProxyReq.UsageMonitors = append(sessionProxyReq.UsageMonitors, pm)
+			}
+		}
```
uri200: try to use a bi dimensional map

I haven't checked but somethin like this

Remember to encapsulate in functions if you can
```
reqMap := map[string]map[string]*protos.UpdateSessionRequest{}
```

Then

```
for plmnid, req := range reqMap {
	for _, u := range r.GetUpdates() {
		if u == nil {
			continue
		}
		plmnid := getPlmnId6(u.GetCommonContext().GetSid().GetId())
		req, ok := reqMap[plmnid]
		if !ok || req == nil {
			addReqMapEntry(req)
		} 
		switch {
			req.CommonContext.RatType == protos.4g:
				reqMap[plmnid][FegSessionProxy].Updates = append(reqMap[plmnid][FegSessionProxy].Updates, u)
			req.CommonContext.RatType == protos.RATType_TGPP_NR:
				reqMap[plmnid][FegN7N40Proxy].Updates = append(reqMap[plmnid][FegN7N40Proxy].Updates , u)
			default:
				return error

	
		}
	}
	

func addReqMapEntry(reqMap map[string]map[string]*protos.UpdateSessionRequest{}){
	reqMap[plmnid][FegSessionProxy]= &protos.UpdateSessionRequest{
						Updates:       []*protos.CreditUsageUpdate{},
						UsageMonitors: []*protos.UsageMonitoringUpdateRequest{} 		 				

	reqMap[plmnid][FegN7N40Proxy]=  &protos.UpdateSessionRequest{
						Updates:       []*protos.CreditUsageUpdate{},
						UsageMonitors: []*protos.UsageMonitoringUpdateRequest{}}
}





```

uri200 (changes requested):

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
anusha-konda: @uri200 , updated. please check

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -44,60 +45,85 @@ func (s *RelayRouter) UpdateSession(
 	// Group requests by PLMNID
 	// Use the longest possible PLMN ID (6), in the worst case we'll fragment requests more then needed, but the routing
 	// will still be logically correct
-	reqMap := map[string]*protos.UpdateSessionRequest{}
+	reqMap := map[string]map[gateway_registry.GwServiceType]*protos.UpdateSessionRequest{}
```
uri200: lets add comments so we don't forget what we are doing here.

You can use something like
```
	// Group requests by [PLMNID][GwServiceType]
	// - PLMNID: Use the longest possible PLMN ID (6), in the worst case we'll fragment requests more then needed, but the routing will still be logically correct
	// - GwServiceType: each request will be sent to either Session_Proxy or for N7_N40_proxy
```

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -44,60 +45,85 @@ func (s *RelayRouter) UpdateSession(
...
 	for _, u := range r.GetUpdates() {
 		if u == nil {
 			continue
 		}
 		plmnid := getPlmnId6(u.GetCommonContext().GetSid().GetId())
-		req, ok := reqMap[plmnid]
+		pMap, ok := reqMap[plmnid]
```
uri200: Add a comment here
```
 // Group by PLMNID
```
(before line 53)

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -44,60 +45,85 @@ func (s *RelayRouter) UpdateSession(
...
+		if !ok || pMap == nil {
```
uri200: Add a comment here 
```
// Group by GwServiceType
```

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -44,60 +45,85 @@ func (s *RelayRouter) UpdateSession(
...
+			pMap = map[gateway_registry.GwServiceType]*protos.UpdateSessionRequest{}
+			reqMap[plmnid] = pMap
+		}
```
uri200: you need to encapsulate that. You are using the same code in two places

Check the code i added previously on my comments

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -152,3 +177,20 @@ func genUpdateErrorResp(req *protos.UpdateSessionRequest) *protos.UpdateSessionR
...
+func getSessionControllerProxyType(r protos.RATType) gateway_registry.GwServiceType {
+	switch r {
+	case protos.RATType_TGPP_NR:
+		return FegN7N40Proxy
+	default:
+		return FegSessionProxy
+	}
+}
+
+func addUpdateSessionReqEntry() *protos.UpdateSessionRequest {
```
uri200: this function could do much more if you pass the map

Note maps are passes as a reference so anything you modify there will be modified in the map. Check the code I added

uri200 (changes requested):

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
anusha-konda: Done

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
anusha-konda: Done

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
anusha-konda: Done

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
anusha-konda: @uri200 , Got it. I thought to initialize the inner map dynamically than statically, if the request contains specific RATType. So it was not encapsulated in the common function. Now I placed the common function to initialize the inner map for GwServiceType to avoid code redundancy. Please check

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
anusha-konda: Done. please check.

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -152,3 +183,28 @@ func genUpdateErrorResp(req *protos.UpdateSessionRequest) *protos.UpdateSessionR
...
+func getSessionControllerProxyType(r protos.RATType) gateway_registry.GwServiceType {
+	switch r {
+	case protos.RATType_TGPP_NR:
+		return FegN7N40Proxy
+	default:
+		return FegSessionProxy
+	}
+}
+
+// Defines the map for Session_Proxy and N7_N40_Proxy per PLMN
```
uri200: in go comments need to start with name of the function. You will see that highlighted on the IDE
```
//addReqMapEntry Defines the map for Session_Proxy and N7_N40_Proxy per PLMN
```
Add it in all function descriptions

image

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -41,63 +42,94 @@ func (s *RelayRouter) CreateSession(
...
-	for plmnid, req := range reqMap {
-		go func(plmnid string, req *protos.UpdateSessionRequest) {
-			client, ctx, cancel, err := s.getSessionProxyClient(ctx, plmnid)
-			if err != nil {
-				glog.Errorf("failed connect to Session Proxy for PLMNID '%s': %v", plmnid, err)
-				resultChan <- genUpdateErrorResp(req)
-				return
+	for plmnid, proxyMap := range reqMap {
+		for _, req := range proxyMap {
+			if req == nil || len(req.Updates) == 0 && len(req.UsageMonitors) == 0 {
```
uri200: can you ad parenthesis for readability. Sometimes they can save you of some unexpected results

feg/cloud/go/services/feg_relay/gw_to_feg_relay/tests/relay_test.go
```diff
@@ -150,6 +286,37 @@ func TestNHRouting(t *testing.T) {
...
+	toutctx, cancel = context.WithTimeout(ctx, 5*time.Second)
+	updateSessReq := &lte_protos.UpdateSessionRequest{
+		Updates: []*lte_protos.CreditUsageUpdate{
+			createCreditUsageRequest(nhImsi, 1, 1, lte_protos.CreditUsage_QUOTA_EXHAUSTED, lte_protos.RATType_TGPP_LTE),
+			createCreditUsageRequest(nhImsi2, 2, 2, lte_protos.CreditUsage_TERMINATED, lte_protos.RATType_TGPP_NR),
+		},
+		UsageMonitors: []*lte_protos.UsageMonitoringUpdateRequest{
+			createUsageMonitoringRequest(nhImsi, "mkey", 1, lte_protos.MonitoringLevel_SESSION_LEVEL, lte_protos.RATType_TGPP_LTE),
+			createUsageMonitoringRequest(nhImsi2, "mkey2", 2, lte_protos.MonitoringLevel_PCC_RULE_LEVEL, lte_protos.RATType_TGPP_NR),
+		},
```
uri200, line 311: can you add one monitoring request more and. This way the test will cover more cases with 2 updates and 3 monitor request

feg/cloud/go/services/feg_relay/gw_to_feg_relay/tests/relay_test.go
```diff
@@ -113,6 +245,9 @@ func TestNHRouting(t *testing.T) {
 	// Register FeG's test Hello Server
 	feg_protos.RegisterHelloServer(srv.GrpcServer, &testHelloServer{})
 
+	// Register FeG's Central Session Controller Server
+	sessionProxy := &testCentralSessionController{resultChan: make(chan string, 2)}
```
uri200, line 259: shouldnt you have to have two `testCentralSessionController`

uri200, line 259: shouldnt you have to have two `testCentralSessionController` assure they are sent to the right one?

anusha-konda: @uri200 , might not need two testCentralSessionController. I could see [here](https://github.com/magma/magma/blob/master/feg/cloud/go/services/feg_relay/gw_to_feg_relay/tests/relay_test.go#L107) only one serving FeG Server is simulated and different gRPC Servicers are registered to it. [Here](https://github.com/magma/magma/blob/master/feg/cloud/go/services/feg_relay/gw_to_feg_relay/tests/relay_test.go#L121), trying to bypass the request to simulated Feg Server. So added one testCentralSessionController, and in UpdateSession rpc definition, checking for RATType and  the response object is filled accordingly. This ensured that UpdateSession is triggered twice and the collective responses are verified

anusha-konda: not added as explained below

feg/cloud/go/services/feg_relay/gw_to_feg_relay/tests/relay_test.go
anusha-konda: @uri200 , Done. Please check

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
anusha-konda: Done

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
anusha-konda: Done

uri200 (approved):

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -19,85 +19,117 @@ import (
...
 
-	// Group requests by PLMNID
-	// Use the longest possible PLMN ID (6), in the worst case we'll fragment requests more then needed, but the routing
-	// will still be logically correct
-	reqMap := map[string]*protos.UpdateSessionRequest{}
+	// Group requests by [PLMNID][GwServiceType]
+	// - PLMNID: Use the longest possible PLMN ID (6), in the worst case we'll fragment requests more then needed,
+	//   but the routing will still be logically correct
+	// - GwServiceType: Each request will be sent to either Session_Proxy or for N7_N40_proxy
+	reqMap := map[string]map[gateway_registry.GwServiceType]*protos.UpdateSessionRequest{}
```
emakeev: why do you need map of maps here (the map of proxy types)? It seems that you are not using the proxy type map, just initializing it and then using GetRatType() instead anyway. Maps are relatively expensive and it complicates your code. I may have missed the instances of proxy type based map lookups, please LMK if this is the case.

uri200: Hey @emakeev  I am trying to remember why but I think the idea is to group the request. 

So for example, sessiond will send all the updates in one message
```
{
Update A:
	PLMN 1
	RAT 4G
   Update B:  
	PLMN 1
	RAT 4G
   Update C: 
	PLMN 2
	RAT 5G
}
```

The result of the `map[PLMN][RAT]` should be 
```
[1][4g] = [A, B]
[2][4g] = [C]
```

Note that [A,B] should be sent  from feg relay to session_proxy in one single message which contains both updates


Norte it is true that RAT is not being used later anymore

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
```diff
@@ -19,85 +19,117 @@ import (
...
 	// send a separate Update request for each unique PLMN ID
-	for plmnid, req := range reqMap {
-		go func(plmnid string, req *protos.UpdateSessionRequest) {
-			client, ctx, cancel, err := s.getSessionProxyClient(ctx, plmnid)
-			if err != nil {
-				glog.Errorf("failed connect to Session Proxy for PLMNID '%s': %v", plmnid, err)
-				resultChan <- genUpdateErrorResp(req)
-				return
+	for plmnid, proxyMap := range reqMap {
+		for _, req := range proxyMap {
```
uri200: I think in the `_` you have RAT type, you don't need the code starting at 102

uri200 (changes requested):
let's get back for some more changes

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
anusha-konda: Hi @emakeev , to echo @uri200 , UpdateSessionRequest from sessiond populates updates of 4G and 5G. Map of Maps is used to maintain  plmn to GWServiceType(Feg proxy) mappings i.e., [PLMN][GWServiceType].  And then relays the request to SessionProxy or N7N40Proxy of FeG.

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
anusha-konda: @uri200 , here ```_``` represents gateway_registry.GwServiceType but not RATType
```
	FegSessionProxy gateway_registry.GwServiceType = "session_proxy"
	FegN7N40Proxy   gateway_registry.GwServiceType = "n7_n40_proxy"
```
Other than RATType_TGPP_NR, as all other RATType requests are forwarded to session_proxy (as referred below), I thought to segregate the requests based on GWServiceType rather than RATType.
```
func getSessionControllerProxyType(r protos.RATType) gateway_registry.GwServiceType {
	switch r {
	case protos.RATType_TGPP_NR:
		return FegN7N40Proxy
	default:
		return FegSessionProxy
	}
}
```
Please let me know your thoughts on this.

anusha-konda: @uri200 , please let me know if any more comments on this.

uri200:
```
reqMap := map[string]map[gateway_registry.GwServiceType]
```


In this loop you don't need RatType at all. You already used RatType to select  GwServiceType  at line 80

Function `getSessionControllerProxyClient` should never use RatType but `gwType`. That function should just only do the selection depending on `gwType` flag.

```
	for plmnid, proxyMap := range reqMap {
		for gwType, req := range proxyMap {
		...

		go func(plmnid string,  gwType gateway_registry.GwServiceType, eq *protos.UpdateSessionRequest) {
				var ratType protos.RATType

				client, ctx, cancel, err := s.getSessionControllerProxyClient(ctx, gwType, plmnid)
				if err != nil {
					glog.Errorf("failed connect to %s for PLMNID '%s': %v", gwType, plmnid, err)
					resultChan <- genUpdateErrorResp(req)
					return
				}
			....
		
```

The whole point of the multidimensional map is the dimensions is giving you the coordinates where to send the message. If you end up not using one of the dimensions then it means something is wrong or that the dimension was not needed.

anusha-konda: @uri200 , Done. Please check

uri200:
Hey @GANESH-WAVELABS  Can you review this PR or include other members of your team to do so?

feg/cloud/go/services/feg_relay/gw_to_feg_relay/servicers/session_proxy_relay.go
uri200: Hey @anusha-konda , can you forward this pr to @GANESH-WAVELABS  or other members on Wavelabs so they can review it too please?

GANESH-WAVELABS (commented):
LGTM

uri200:
@shivesh-wavelabs  can you rebase?

uri200 (approved):

emakeev (approved):

shivesh-wavelabs:
> @shivesh-wavelabs can you rebase?

done.

uri200:
Hey @shivesh-wavelabs  this one is failing

image

uri200:
make sure you run `.build.py -c` on orc8r so test are run

shivesh-wavelabs:
> Hey @shivesh-wavelabs this one is failing
> 
> image

fixed

PR #11049, opened 2022-01-04
fix(agw): Added code to log emm cause in string format

rsarwad:
fix(agw): Added code to log emm cause in string format

## Summary
Modified code to print/log emm cause in string format instead of integer value
## Test Plan
Executed s1ap sanity test suite and unit test case

### DIFF
--- a/lte/gateway/c/core/oai/lib/3gpp/3gpp_24.301.h
+++ b/lte/gateway/c/core/oai/lib/3gpp/3gpp_24.301.h
@@ -507,4 +507,5 @@ typedef struct nr_ue_security_capability_s {
 #define ESM_CAUSE_MESSAGE_NOT_COMPATIBLE 101
 #define ESM_CAUSE_PROTOCOL_ERROR 111
 
+extern const char* emm_cause_str[256];
 #endif /* FILE_3GPP_24_301_SEEN */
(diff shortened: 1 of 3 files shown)


uri200:
Hey @rsarwad 

Note the idea is to print the message coming from grpc response. 
Meaning, I would like to enhance the messages at FEG relay to say where request failed. Then that error message is sent back to MME. That is the message I would like to 

In s8_porxy we already have nice messages being returned like [this one](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fcloud%2Fgo%2Fservices%2Ffeg_relay%2Fgw_to_feg_relay%2Fservicers%2Fs8_relay.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=50&start_col=16&end_row=50&end_col=16). I want to do the same for the rest of the services, this way MME logs will tell us something more, withouth having to chase the error around.


Or are we using the error returned by GRPC as an 3gpp error itself?

rsarwad:
> 

Okay, you meant to add more prints on handling grpc response messages, covering error scenarios?

ulaskozat (approved):
LGTM

PR #11060, opened 2022-01-05
fix(ci): Fix provisioning of magma_test vm

quentinDERORY:

## Summary

Remove old reference to magma etagecom

## Test Plan
Provisioned magma_test vm successfully

## Additional Information


andreilee (approved):

themarwhal (approved):

orc8r/tools/ansible/roles/pkgrepo/tasks/debian.yml
```diff
@@ -25,6 +25,11 @@
     url: https://artifactory.magmacore.org:443/artifactory/api/gpg/key/public
     state: present
 
+- name: Remove old magma etagecom sources
```
uri200, line 28: why do we have to remove them?
they should never be added on a brand new vm

PR #10792, opened 2021-12-10
feat(orc8r): Terraform Integration with cert-manager

ShubhamTatvamasi:
Signed-off-by: Shubham Tatvamasi <shubhamtatvamasi@gmail.com>

## Additional Information

- [ ] This change is backwards-breaking

PR #11078, opened 2022-01-06
fix(orc8r): Revert additonal labels from k8s services to avoid incompatibilities with deployed objects

joary:
An existing service has imutable selectors, the additonal labels included in PR10085 creates an incompatibility with existing services.

## Summary

Disable the labels added to k8s services objects from PR10085

## Additional Information


uri200:
can you take a look if  there is something else from this PR https://github.com/magma/magma/pull/10792 that needs to be reverted please and add it here?

uri200 (approved):

PR #11069, opened 2022-01-06
fix(agw): logging is formatted for sentry usage

nstng:
Signed-off-by: Nils Semmelrock <nils.semmelrock@tngtech.com>

## Summary

Current: Sentry groups affected log message by complete message. Here: split log string so that sentry is able to aggregate among log arguments. 

## Test Plan

local test against private sentry -> messages are aggregated

## Additional Information

Analogously to https://github.com/magma/magma/pull/9751 (might not have been found by linter because log message was created as variable).

- [ ] This change is backwards-breaking

PR #11095, opened 2022-01-07
fix(agw): Rendering strings is done by logging framework

jheidbrink:
## Summary

[pull/11069](https://github.com/magma/magma/pull/11069) fixed Sentry event aggregation for an error log that was observed in the lab. This PR extends this to several error logs that we haven't so far seen in the wild, but that would equally fail to be correctly aggregated by Sentry.

## Test Plan

Tested in a local repl that the new function works

## Additional Information

Based on [pull/11069](https://github.com/magma/magma/pull/11069), so should be merged after that one

The lines changed here were found with the regex `\.error\([^'"]+` (with "full words only" checked)


Neudrino (approved):

themarwhal (approved):

ardzoht (approved):
lgtm

PR #11102, opened 2022-01-07
fix(pipelined): ip_solicitation: Use link local address for RA

### LINKED ISSUE #11100: [AGW] v1.7 UE IPv6 address allocation not working

### Your Environment

- **Version:** v1.7
- **Affected Component:** AGW
- **Affected Subcomponent:** IP alloc
- **Deployment Environment:** e.g. AWS (orc8r), local docker-compose (orc8r), bare-metal (AGW), Vagrant (AGW)

### Describe the Issue
* UE asks for IPv6 PDN acess
* MME sends an attach accept with IPv6 PDN type. In Attach Accept, sends interface identifier and no prefix.
* UE does a router solitiation
* MME responds with a router advertisement and this time includes a prefix (/64) the UE is supposed to use. 
* UE doesn't use the prefix and eventually just goes idle. 

**Expected behavior**
UE should be able to access northbound networks using the IPv6 prefix sent in the Router Advertisement message. 

Theories: 
* In the attached PCAP, packet 120 is the `Router  Solicitation` sent by the UE. The source address is the link local address of the UE while the destination is the multicast all routers address (ff02::2). However, in packet 121, the `Router Advertisement` message contains the site-local address of the MME. Per [RFC4861](https://datatracker.ietf.org/doc/html/rfc4861#section-4.2) this *_must_* to be the link-local address of the interface that is sending the RA message. 

* One other potential theory is that there is no prefix being sent in the activate pdp context message. This seems less plausible as the need for the address is to create a link local address and it seem the UE is able to create it's own link local address. not entirely sure here.

```
3) The MS receives the IPv6 address produced by the GGSN in the Activate PDP Context Accept. The MS extracts
the interface identifier from the address received and stores it. The MS shall use this interface identifier to build
its link-local address and may also use it for building its full IPv6 address, as describe in step 5. The MS shall
ignore the prefix contained in the address received in the Activate PDP Context Accept. The processing of the
Activate PDP Context Accept is otherwise as specified in clause "PDP Context Activation Procedure". 
```
[source](https://www.etsi.org/deliver/etsi_ts/123000_123099/123060/10.03.00_60/ts_123060v100300p.pdf)

**Screenshots**

**Additional context**
[ipv6_address_allocation_filtered.pcapng.zip](https://github.com/magma/magma/files/7830434/ipv6_address_allocation_filtered.pcapng.zip)

pshelar:
Per RFC4861 (section-4.2) this must to be the link-local address
of the interface that is sending the RA message.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/ipv6_solicitation.py
+++ b/lte/gateway/python/magma/pipelined/app/ipv6_solicitation.py
@@ -12,6 +12,7 @@ limitations under the License.
 """
 from collections import namedtuple
 
+from magma.common.misc_utils import get_link_local_ipv6_from_if
 from magma.pipelined.app.base import ControllerType, MagmaController
 from magma.pipelined.ipv6_prefix_store import get_ipv6_interface_id
 from magma.pipelined.openflow import flows
@@ -25,6 +26,12 @@ from ryu.controller import dpset, ofp_event
 from ryu.controller.handler import MAIN_DISPATCHER, set_ev_cls
 from ryu.lib.packet import ether_types, ethernet, icmpv6, in_proto, ipv6, packet
 from ryu.ofproto.inet import IPPROTO_ICMPV6
+
+MAX_ROUTE_LIFE_TIME = int(hex(0xffffffff), 16)
+MAX_ROUTE_ADVT_LIFE_TIME = int(hex(0xffff), 16)
+MAX_HOP_LIMIT = 255
+ROUTE_PREFIX_HDR_FLAG_L = 0x4
+ROUTE_PREFIX_HDR_FLAG_A = 0x2
 
 
 class IPV6SolicitationController(MagmaController):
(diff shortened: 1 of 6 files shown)


lte/gateway/python/magma/pipelined/app/ipv6_solicitation.py
```diff
@@ -190,13 +206,19 @@ def _send_router_advertisement(
             icmpv6.icmpv6(
                 type_=icmpv6.ND_ROUTER_ADVERT,
                 data=icmpv6.nd_router_advert(
+                    ch_l=MAX_HOP_LIMIT,     # Hope limit
```
koolzz: rm comment, seems obvious by variable name

koolzz (approved):
lgtm, but remove 'ipv6_router_addr' from
 `xwf/gateway/configs/pipelined.yml/`
`cwf/gateway/configs/pipelined.yml`
`lte/gateway/python/magma/pipelined/tests/test_ipv6_solicitation.py`

pshelar:
> ipv6_router_addr

I would keep it in yml, its still supported config option.

koolzz:
> > ipv6_router_addr
> 
> I would keep it in yml, its still supported config option.

ah right, yeah maybe just comment it out for consistency

pshelar:
> > > ipv6_router_addr
> > 
> > 
> > I would keep it in yml, its still supported config option.
> 
> ah right, yeah maybe just comment it out for consistency

done.

hcgatewood (approved):

uri200 (approved):

PR #11107, opened 2022-01-10
[WIP] Python support in DevContainer

alexzurbonsen:

## Summary

Enabling DevContainer for Python (installing Python dependencies, generating proto and swagger code, IDE support: Python IntelliSense, etc.)

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-# Note: You can use any Debian/Ubuntu based image you want. 
+# Note: You can use any Debian/Ubuntu based image you want.
 FROM ubuntu:focal as magma-mme-builder
 
 # [Option] Install zsh
@@ -23,8 +23,8 @@ RUN apt-get update \
     # Clean up
     && apt-get autoremove -y && apt-get clean -y && rm -rf /var/lib/apt/lists/* /tmp/library-scripts/
 
-# Setting the ENTRYPOINT to docker-init.sh will configure non-root access to 
-# the Docker socket if "overrideCommand": false is set in devcontainer.json. 
+# Setting the ENTRYPOINT to docker-init.sh will configure non-root access to
+# the Docker socket if "overrideCommand": false is set in devcontainer.json.
 # The script will also execute CMD if you need to alter startup behaviors.
 ENTRYPOINT [ "/usr/local/share/docker-init.sh" ]
 CMD [ "sleep", "infinity" ]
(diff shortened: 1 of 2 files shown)


.devcontainer/Dockerfile
```diff
@@ -310,14 +312,85 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git
...
+RUN git clone --branch $CODEGEN_VERSION https://github.com/swagger-api/swagger-codegen /var/tmp/codegen
+
+WORKDIR /var/tmp/codegen
+RUN ../apache-maven-$MVN_VERSION/bin/mvn clean package -DskipTests
+
+RUN for SWAGGER_SRC in lte orc8r; \
+     do \
+     mkdir -p $PYTHON_LIB/gen/$SWAGGER_SRC/swagger/specs \
+     && touch $PYTHON_LIB/gen/$SWAGGER_SRC/swagger/__init__.py \
+     && touch $PYTHON_LIB/gen/$SWAGGER_SRC/swagger/specs/__init__.py \
```
jheidbrink, line 375: Have you tried this without those `__init__.py`s? We might not need those with Python 3.

jheidbrink: Also, I think it might be a good idea to migrate this to Bazel now.

.devcontainer/devcontainer.json
```diff
@@ -18,10 +18,19 @@
 		"stackbuild.bazel-stack-vscode-cc",
 		"augustocdias.tasks-shell-input",
 		"ryuta46.multi-command",
+		"zxh404.vscode-proto3",
+		"ms-python.python",
+		"ms-python.vscode-pylance",
+		"njpwerner.autodocstring",
```
jheidbrink, line 24: I think it would be useful to extract the changes to `devcontainer.json` into a very small PR that will already lift Python support from "nonexistent" to "rudimentary". That part could probably already be merged.

alexzurbonsen: Sounds good. I will prepare a devcontainer.json PR.

.devcontainer/Dockerfile
alexzurbonsen: > Have you tried this without those `__init__.py`s? We might not need those with Python 3.

For some reason leaving out the __init__.py files does cause problems for some proto and swagger files but not for others. Meaning they are not resolved properly anymore.

Issue #11091, opened 2022-01-07
Migrate Python services to pull Sentry config from shared config


themarwhal:
More details in the parent task.

Python services should now be reading the DSN values from the shared mconfig field. The configs should still be overriden if values are set in control_proxy. 

I think this might be done already. (@sebathomas / @alexzurbonsen please close the issue if it is :) )

PR #11108, opened 2022-01-10
chore(agw): add logs to indicate sentry config source for dsn and sample rate

alexzurbonsen:
## Summary

add logs to indicate sentry config source for dsn and sample rate and a additional comment to explain the reasoning behind the logic of the function

## Test Plan

None

### DIFF
--- a/orc8r/gateway/python/magma/common/sentry.py
+++ b/orc8r/gateway/python/magma/common/sentry.py
@@ -90,8 +90,8 @@ def get_sentry_dsn_and_sample_rate(sentry_mconfig: mconfigs_pb2.SharedSentryConf
     """Get Sentry configs with the following priority
 
     1) control_proxy.yml (if sentry_python_url is present)
-    2) shared mconfig (i.e. first streamed mconfig from orc8r,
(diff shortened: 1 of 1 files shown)


themarwhal:
Ah sorry, I think I miswrote in the issue. The order should be like this
1. If the control proxy value is set, take that value (as this is a local config override per-AGW)
2. if not 1, take the mconfig value

I think we may eventually want to migrate to option 2 only, but for that we need to have the shared mconfig configurable in the NMS (our oc8r UI).

So i think the original implementation might have been good? But your note about taking these values from the same source seems important. 

Also, if you could add some logs (if there aren't already) about where these values are being taken so that it's clear to log readers, that would be awesome. :)

jheidbrink (approved):

alexzurbonsen:
Oh, yes, then the original implementation was good.

I will add the note and the logs though! :)

alexzurbonsen:
I adapted the PR title and description to suit the new purpose. Closing  #11091.

themarwhal (approved):

PR #11111, opened 2022-01-10
fix(agw): Handle failed IP allocation

### LINKED ISSUE #11097: Frequent Sentry Error: Handle empty list gracefully
In [mobilityd](https://github.com/magma/magma/blob/fefdc85c021303047d60173ef3098b8c84eab9d6/lte/gateway/python/magma/mobilityd/rpc_servicer.py#L237), we don't check if the returned `ipv6_response.ip_list` is empty. Probably we should at least raise a more meaningful exception.

Sentry Issue: [LAB-AGWS-PYTHON-48](https://sentry.io/organizations/lf-9c/issues/2882186535/?referrer=github_integration)

```
IndexError: list index out of range
  File "grpc/_server.py", line 435, in _call_behavior
    response_or_iterator = behavior(argument, context)
  File "/usr/local/lib/python3.8/dist-packages/magma/mobilityd/rpc_servicer.py", line 237, in AllocateIPAddress
    ipv6_addr = ipv6_response.ip_list[0]
  File "/usr/local/lib/python3.8/dist-packages/google/protobuf/internal/containers.py", line 209, in __getitem__
    return self._values[key]

Exception calling application: list index out of range
```

jheidbrink:
fixes #11097

Signed-off-by: Jan Heidbrink <jan.heidbrink@tngtech.com>

## Summary

Currently when either IPv4 or IPv6 allocation fails, we raise an IndexError. Now we only log a warning here, as the caller of the `AllocateIPAddress` endpoint will receive meaningful context about the failure.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/mobilityd/rpc_servicer.py
+++ b/lte/gateway/python/magma/mobilityd/rpc_servicer.py
@@ -233,13 +233,17 @@ class MobilityServiceRpcServicer(MobilityServiceServicer):
                 composite_sid + ",ipv6", IPAddress.IPV6,
                 context, request,
             )
-            ipv4_addr = ipv4_response.ip_list[0]
-            ipv6_addr = ipv6_response.ip_list[0]
-            # Get vlan from IPv4 Allocate response
-            resp = AllocateIPAddressResponse(
-                ip_list=[ipv4_addr, ipv6_addr],
-                vlan=ipv4_response.vlan,
-            )
+            try:
+                ipv4_address = ipv4_response.ip_list[0]
+                ipv6_address = ipv6_response.ip_list[0]
+            except IndexError:
+                logging.warning("IPv4/IPv6 IP address allocation not successful")
+                resp = AllocateIPAddressResponse()
+            else:
+                resp = AllocateIPAddressResponse(
+                    ip_list=[ipv4_address, ipv6_address],
+                    vlan=ipv4_response.vlan,
+                )
         else:
             resp = AllocateIPAddressResponse()
 


lte/gateway/python/magma/mobilityd/rpc_servicer.py
```diff
@@ -233,12 +233,18 @@ def AllocateIPAddress(self, request, context):
                 composite_sid + ",ipv6", IPAddress.IPV6,
                 context, request,
             )
-            ipv4_addr = ipv4_response.ip_list[0]
-            ipv6_addr = ipv6_response.ip_list[0]
-            # Get vlan from IPv4 Allocate response
-            resp = AllocateIPAddressResponse(
-                ip_list=[ipv4_addr, ipv6_addr],
-                vlan=ipv4_response.vlan,
+
```
ardzoht: remove whitespace?

lte/gateway/python/magma/mobilityd/rpc_servicer.py
```diff
@@ -233,12 +233,18 @@ def AllocateIPAddress(self, request, context):
...
             )
-            ipv4_addr = ipv4_response.ip_list[0]
-            ipv6_addr = ipv6_response.ip_list[0]
-            # Get vlan from IPv4 Allocate response
-            resp = AllocateIPAddressResponse(
-                ip_list=[ipv4_addr, ipv6_addr],
-                vlan=ipv4_response.vlan,
+            
+            try:
+                ipv4_addr = ipv4_response.ip_list[0]
```
ardzoht: We should handle this error separately, given we use different allocators for IPv4 and IPv6, this means IPv4 can still be successful, with this current change we will throw away both of the allocations if one of them fails. 

How about checking for each list's content, if it's empty we will not access the list to avoid the `IndexError`, if it's not empty we will append the `ipv4_addr` or `ipv6_addr` (or both) content to the final `ip_list` and use that to set the content of `AllocateIPAddressResponse`. 

Wdyt?

ardzoht (changes requested):

lte/gateway/python/magma/mobilityd/rpc_servicer.py
jheidbrink: Sounds good to me! My initial motivation for this approach was that it might make it clearer to the caller that something went wrong. Also I wasn't sure if there is a scenario where someone requests both an IPv4 and an IPv6 address and can still reasonably proceed when only receiving one of the two.

lte/gateway/python/magma/mobilityd/rpc_servicer.py
```diff
@@ -233,13 +233,16 @@ def AllocateIPAddress(self, request, context):
...
-            # Get vlan from IPv4 Allocate response
-            resp = AllocateIPAddressResponse(
-                ip_list=[ipv4_addr, ipv6_addr],
-                vlan=ipv4_response.vlan,
-            )
+            ip_addresses = list(ipv4_response.ip_list)[:1] + list(ipv6_response.ip_list)[:1]
+            if len(ip_addresses) < 2:
+                logging.warning("IPv4/IPv6 IP address allocation not successful")
+            if ip_addresses:
+                resp = AllocateIPAddressResponse(
```
jheidbrink, line 243: @ardzoht I tested this locally with a Python grpc client, and this doesn't work if one of the two allocations fails. The reason is that we delegate the allocation of the IP addresses to the `_get_allocate_ip_response` function, and that function sets the statuscode for the response, making the client code raise an exception. It looks to me like the `_get_allocate_ip_response` function shouldn't be called twice within one grpc request. However, I think changing that is out of scope of this issue, as the issue is only about getting rid of the `IndexError`. I suggest I revert to my initial version that returns both addresses or none, and I create a separate issue for handling the case of only one successfully allocated address.

ardzoht: @jheidbrink Makes sense, thanks for double verifying the issue. If we can open a separate issue for the one-only allocation that sounds good by me and I will accept the PR with the previous implementation.

jheidbrink: Sounds good! I force-pushed the old version and will create an issue for the one-only allocation tomorrow.

ardzoht (approved):

lte/gateway/python/magma/mobilityd/rpc_servicer.py
jheidbrink: For the record: Alex and I chatted about this and decided to keep the current behaviour of returning either both addresses or none, reason being mainly that the [code](https://github.com/magma/magma/blob/59bb7281a17aa81559b6c7c9fca6e4a22d36d9b3/lte/gateway/c/core/oai/lib/mobility_client/MobilityClientAPI.cpp#L292-L311) that handles the response doesn't deal with the case of receiving just one out of two requested addresses.

PR #11121, opened 2022-01-11
fix(amf): Enhancement of logs during UE authentication failure

Akshayp77:
Signed-off-by: Akshayp77 <akshay.patidar@wavelabs.ai>

fix(amf): Enhancement of logs during authentication failure

## Summary
Enhancement of logs during UE authentication failure.
In case authentication failure log the RES,XRES, AUTN , RAND and Serving network name.

## Test Plan
Tested the authentication failure scenario using UERANSIM.

## Additional Information
https://app.zenhub.com/workspaces/magma--5g-sa-60871e99b9326f001696b1ee/issues/magma/magma/11120

Reference logs from UERANSIM testing
log

### DIFF
--- a/lte/gateway/c/core/oai/tasks/amf/amf_authentication.cpp
+++ b/lte/gateway/c/core/oai/tasks/amf/amf_authentication.cpp
@@ -35,6 +35,7 @@ extern "C" {
 
 #define AMF_CAUSE_SUCCESS (1)
 #define MAX_5G_AUTH_VECTORS 1
+#define SNNI_FMT "5G:mnc%03d.mcc%03d.3gppnetwork.org"
 
 using magma5g::AMFClientServicer;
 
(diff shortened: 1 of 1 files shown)


lte/gateway/c/core/oai/tasks/amf/amf_authentication.cpp
```diff
@@ -551,6 +555,31 @@ int amf_proc_authentication_complete(
...
+      OAILOG_STREAM_HEX(
+          OAILOG_LEVEL_INFO, LOG_NAS_AMF, "Received RES*in AUTH RSP:",
+          (char*) (msg->autn_response_parameter.response_parameter),
+          AUTH_RAND_SIZE);
+      PLMN_T_TO_MCC_MNC(
+          amf_ctx->originating_tai.plmn, mcc, mnc, mnc_digit_length);
+      snprintf(
+          snni_buffer, sizeof(snni_buffer),
+          "5G:mnc%03d.mcc%03d.3gppnetwork.org", mnc, mcc);
+      OAILOG_INFO(LOG_NAS_AMF, "Serving network name: %s\n", snni_buffer);
```
lionelgo, line 572: I would just suggest to try something like:
#define SNNI_FMT "5G:mnc%03d.mcc%03d.3gppnetwork.org"
char snni_buffer[sizeof(SNNI_FMT) - 1] = {};
...
snprintf(
          snni_buffer, sizeof(snni_buffer),
          SNNI_FMT, mnc, mcc);

So that if somebody wants to change the trace, he will not have to make mistakes in counting the size of formatted string to dimension the array. But your code is safe anyway.

lionelgo (approved):

lte/gateway/c/core/oai/tasks/amf/amf_authentication.cpp
Akshayp77: Done

ulaskozat (approved):
LGTM

PR #11138, opened 2022-01-12
test(agw): Additional s1sim test for TAU with eps bearer context status IE

pruthvihebbani:
test(agw): Additional s1sim tests for TAU with eps bearer context status 

## Summary

Added new test for TAU with eps bearer context status IE and included related tests to sanity suite

## Test Plan

- Verified new test case multiple times
- Verified sanity suite


ulaskozat (approved):
LGTM

PR #11140, opened 2022-01-12
fix(dp): handle logging when cbsd does not exist

xbend:
Fixed an exception when trying to log requests/responses that belong to
a cbsd which does not exist in the db.

## Summary

* Added check for cbsd entity when logging requests and responses
* if cbsd does not exist, log empty fields into log table

### DIFF
--- a/dp/cloud/python/magma/configuration_controller/response_processor/response_db_processor.py
+++ b/dp/cloud/python/magma/configuration_controller/response_processor/response_db_processor.py
@@ -135,9 +135,16 @@ class ResponseDBProcessor(object):
         request.state = self.request_states_map[RequestStates.PROCESSED.value]
 
     def _log_response(self, session: Session, response: DBResponse):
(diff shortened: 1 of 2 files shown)


xbend:
@hcgatewood @ardzoht 
I think we should extend the allowed PR scope with new `dp` value, as dp module is now part of Magma.
Can we do this as part of this PR or should we extend the github workflow in a separate pr?

xbend:
@hcgatewood @ardzoht
"dp" scope is now part of semantic PR scope list. CI is passing all checks. This change is ready to merge.

ardzoht (approved):
lgtm, cc @hcgatewood for orc8r approval

xbend:
cc @andreilee for orc8r approval, as we haven't heard from @hcgatewood for a while.

hcgatewood (approved):

xbend:
@ardzoht 
We are ready to merge.

PR #11146, opened 2022-01-13
fix(feg_hello): remove timezone from timestamps

uri200:

## Summary

Removing timezone from timestamps

## Test Plan
terravm test

## Additional Information


ulaskozat (approved):
LGTM

PR #11147, opened 2022-01-13
fix(s8_proxy): inject u_agw_teid to create session response

uri200:

## Summary
Added missing User plane teid from AGW.
This teid is given by FEG_RELAY so it needs to be included in the Create Session Response so the AGW knows its own user plane teid

## Test Plan
Teravm

## Additional Information


emakeev (approved):
LGTM

PR #11165, opened 2022-01-14
chore(deps): Upgrade aiohttp from 3.6.2 to 3.8.0

spikey979:

chore(deps): Upgrade aiohttp from 3.6.2 to 3.8.0

## Summary

There is  an open redirect vulnerability in python aiohttp v3.6.2 lib. A maliciously crafted link to an aiohttp-based web-server could redirect the browser to a different website. This security problem has been fixed in v3.7.4. but in that version there is a bug that causes some unit test to failing. There is a new version 3.8.0 that we shuld try.

## Test Plan

ran magma/lte/gateway make test_python

PR is for security alert: https://github.com/magma/magma/security/dependabot/lte/gateway/python/setup.py/aiohttp/open

### DIFF
--- a/lte/gateway/python/setup.py
+++ b/lte/gateway/python/setup.py
@@ -125,7 +125,7 @@ setup(
         'freezegun>=0.3.15',
         'pycryptodome>=3.9.9',
         'pyroute2==0.5.14',
-        'aiohttp==3.6.2',
+        'aiohttp==3.8.0',
         'jsonpointer>=1.14',
         'ovs>=2.13',
         'prometheus-client>=0.3.1',
--- a/lte/gateway/python/setup.py
+++ b/lte/gateway/python/setup.py
@@ -125,7 +125,7 @@ setup(
         'freezegun>=0.3.15',
         'pycryptodome>=3.9.9',
         'pyroute2==0.5.14',
-        'aiohttp==3.8.0',
+        'aiohttp',
         'jsonpointer>=1.14',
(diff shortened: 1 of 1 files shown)


spikey979:
I’ve run python tests on local machine (magma/lte/gateway make test_python) and all test passed. But on PR, in section “Run mme tests with Debug build type”, tests failed…

lte/gateway/python/setup.py
```diff
@@ -125,7 +125,7 @@
         'freezegun>=0.3.15',
         'pycryptodome>=3.9.9',
         'pyroute2==0.5.14',
-        'aiohttp==3.6.2',
+        'aiohttp==3.8.0',
```
ardzoht:
```suggestion
        'aiohttp',
```
Unfortunately due to our VMs mismatch on python versions (magma VM on 3.8 and magma_test VM on 3.5) the `aiohttp 3.8.0` package is not available on the test VM (maximum version that comes up is `3.6.3`), so we can leave this package without version so on the magma VM the latest version is installed, which testing locally should install the `3.8.1` version.

Note: This will be resolved eventually when we migrate to a containerized versions for our dev and testing environments, cc @arunuke

ardzoht (changes requested):

lte/gateway/python/setup.py
spikey979: I made changes...

ardzoht (approved):
lgtm

PR #11167, opened 2022-01-14
docs: Magma settlement proposal

Arsenii-Oganov:
Signed-off-by: Arsenii Oganov <arsenii@freedomfi.com>

## Summary

Proposal to introduce Settlement service for Magma

### DIFF
--- a/docs/readmes/proposals/p024_magma_settlement_service.md
+++ b/docs/readmes/proposals/p024_magma_settlement_service.md
@@ -1,0 +1,281 @@
+# Proposal: Settlement Service for Magma
+
+Author(s): [@arsenii-oganov]
+
+Last updated: 01/14/2022
+
+## 1.0 Objectives
+
+The objective of this work is to extend Magma and the Helium blockchain to add support for settlement interfaces to bridge the gap between traditional roaming settlement and HIP 27s proposal of packet purchasing OUI managed by the helium network operator.
+Software built to accomplish this (aka settlement domain service) will be open source under BSD-3-Clause and will reside in the github repository of the DeWi Alliance. Some changes / additions required to the Magma software will be committed to the Magma software repository under the governance of the Linux foundation, such that it can be effectively maintained in the future releases, but will remain openly available to all Helium community participants.
+As a result of this effort, any MNO, MVNO or 3rd party Gateway Manufacturer will be able to use the software to operate their own roaming interconnect with a network roaming into Helium.
+Out of Scope: The commercial aspects of how an MNO/MVNO purchases data credits on the Helium network through a session purchaser operator is outside the scope of this document and is a commercial arrangement between the clearing house, session purchaser operator and the MNO.
+
+## 2.0 Background
+
+The accounting and settlement of a Helium cellular network is unlike traditional cellular settlement. Unlike traditional cellular networks, a Helium cellular network contains radio and accounting entities (e.g. Miners) owned by private parties not affiliated with the core network operator and/or the subscriber home network provider. Additionally, the subscriber home network operator in a Helium network needs a reward path to pay individual radio/miner owners for the data consumed during subscriber sessions that traverse their deployment. A trust-limited accounting system must be implemented to ensure accounting data integrity and secure the reward path.
+When creating this accounting system, the goal has been to avoid significant (or any) deviations from traditional 3GPP interfaces to allow maximum portability and lower the barriers to entry for vendors.
+The systems centers around the new usage of a standard component, the Online Charging System (OCS). The OCS traditional role is quota management and realtime session management and accounting. In this system, the signaling of standard OCS is repurposed to provide control and accounting functions to a new Helium entity called the Session Purchaser.
+The Session Purchasers job is to account and enforce sessions against a subscribers home network operator account on the helium network. The implementation below describes the role and signaling associated with a Session Purchaser.
+
+## 3.0 Implementation scope
+
+### 3.1 High - level solution architecture
+
+![arch](https://user-images.githubusercontent.com/93994458/149513109-47b80417-7488-49c4-a2f0-ff94e6c57ebf.png)
+
+### 3.2 End to end scenarios
+
+Below will be described e2e scenarios needs to be implemented in scope of this project (Note: scenarios may be decomposed into different use-cases depends on actors involved and/or business case):
+
+- State channel management procedure
+
+The procedure describes Opening, Update and Closure of State channel management call flow.
+The detailed call flow for the procedure is specified in section 3.2.1 “State channel management procedure”.
+
+- UE attach procedure
+
+The procedure in which the UE registers to the network, and receives the usage quota to create the EPS Bearer between the UE and the PGW, in order to be able to send and receive data.
+The detailed call flow for the procedure is specified in section 3.2.2 “UE attach procedure”.
+
+- UE quota update procedure
+
+The procedure describes a next quota request by Threshold from the previously received quota from SGW to OCS.
+The detailed call flow for the procedure is specified in section 3.2.3 “UE quota update procedure”.
+
+- UE initiated session termination procedure
+
+The procedure in which the UE initiates session termination. The procedure includes the last quota usage report from SGW to OCS.
+The detailed call flow for the procedure is specified in section 3.2.4 “UE initiated session termination procedure”.
+
+- Network initiated termination procedure
+
+The procedure in which the Network (S/PGW) initiates session termination. The procedure includes the last quota usage report from SGW to OCS.
+The detailed call flow for the procedure is specified in section 3.2.5 “Network initiated termination procedure”.
+
+- TAP-Out generation and exporting procedure
+
+The procedure to generate TAP3 files and transfer the files to the DCH (HPLMN) partner.
+
+- RAP processing procedure
+
+The procedure to return rejected TAP files and records to the visited network operator for corrections.
+
+#### 3.2.1 State channel management procedure
+
+Following diagram shows e2e State Channel life cycle management that will be in real life. It consists of existing functionality (marked with black), new functionality to be implemented in magma (marked with blue).
+
+![State channel management procedure (1)](https://user-images.githubusercontent.com/93994458/149513469-0380c35c-d109-4c27-8014-d05b6cbd9a5e.png)
+
+This scenario consists of the steps below:
+
+- Step [1-2] The State channel opening procedure. The Helium Handler sends Open State Channel Request to open the State Channel. State Channel is opened once the positive Response is received from State Channel.
+- Step [3] On this Step the Handler sends the Used Data Credits for each Used Data Quota per Miner pub key.
+- Step [4-5] The State channel closure procedure. The closure procedure contains two type of triggers - all the Data credits have been used from the State channel and end of an Epoch.
+
+#### 3.2.2 UE attach procedure
+
+Following diagram shows e2e UE attach scenario that will be in real life. It consists of existing functionality (marked with black), new functionality to be implemented in magma (marked with blue).
+The attach procedure implies that State Channel is already in open state.
+
+![Attach call flow](https://user-images.githubusercontent.com/93994458/149513666-63ae0f4b-c8be-4cfc-a739-909723fca3aa.png)
+
+This scenario consists of the steps below:
+
+- Step [1 - 2] Initial attach procedure in a mobile network. The UE establishes radio link synchronization with an eNB, the UE creates a connection for data delivery via an Attach Request message sent to the eNB, the eNB then forwards the attach request to an MME/SGW.
+- Step [3] The SGW sends CCR-I to the Session Purchaser requesting volume quota.
+- Step [4] Session Purchaser checks if the IMSI-prefix (i.e. the HPLMN operator) has balance to answer with granted data units on CCR-I Request which has been received on Step [3].
+- Step [5] In response to a CCR-I (Step [3]), the Session Purchaser returns a CCA-I message that indicates success (DIAMETER_SUCCESS) or failure (DIAMETER AUTHORIZATION REJECTED) depending on whether the HPLMN operator has sufficient credit for the requested services. In case of HPLMN operator has sufficient credit the CCA contains the Granted Service Unit in Bytes.
+- Step [6] The SGW sends signed Session Init Event to Miner. The message is signed with SGW PubKey.
+- Step [7] The Miner sends signed Session Init Event to Helium Handler. The message is signed with Miner PubKey.
+- Step [8 - 9] The Create Session Request and Response messages to establish UE requested PDN connectivity through the SGW and Home PGW.
+- Step [10-15] Establishing radio link between UE and eNodeB for packet data transfer.
+- Step [16-17] Packet data between UE and Internet.
+
+#### 3.2.3 UE quota update procedure
+
+Following diagram shows e2e UE update scenario that will be in real life. It consists of existing functionality (marked with black), new functionality to be implemented in magma and helium (marked with blue).
+
+![Update call flow](https://user-images.githubusercontent.com/93994458/149513805-f78e1c18-bb12-4f1f-abf9-0a0ea5e09d20.png)
+
+This scenario consists of the steps below:
+
+- Step [1-2] Packet data transfer is ongoing between UE and Internet.
+- Step [3] The threshold of the quota is configured on SGW (e.g. 95% of granted service unit).
+- Step [4] In this Step SGW sends Session Update Request with used service unit.
+- Step [5] Session Purchaser checks if the HPLMN operator has balance to answer with granted data units on CCR-U which has been received on Step [4].
+- Step [6] In response to a CCR-U (Step [5]), the Session Purchaser returns a CCA-U message that indicates success (DIAMETER_SUCCESS) or failure (DIAMETER AUTHORIZATION REJECTED) depending on whether the HPLMN operator has sufficient credit for the requested services.
+- Step [7] The SGW sends signed data usage to Miner. The message is signed with SGW PubKey.
+- Step [8] The Miner sends signed data usage to Helium Handler. The message is signed with Miner PubKey.
+- Step [9] The Helium Handler updates State Channel with signed data usage received on Step [8]
+- Step [10-11] Packet data continuation between UE and Internet.
+
+#### 3.2.4 UE initiated session termination procedure
+
+Following diagram shows e2e UE initiated session termination scenario that will be in real life. It consists of existing functionality (marked with black), new functionality to be implemented in magma and helium (marked with blue).
+
+![UE Terminate call flow](https://user-images.githubusercontent.com/93994458/149513885-75facecc-54ef-4710-b6de-65b4881ddebb.png)
+
+This scenario consists of the steps below:
+
+- Step [1-2] UE detach procedure in a mobile network. The UE closes radio link with an eNB and MME/SGW.
+- Step [3] In this Step SGW sends Session Termination Request with last used service unit.
+- Step [4] The Session Purchaser update the HPLMN operator balance with last used data units.
+- Step [5] The Session Purchaser sends Response on CCR-T (Step [3]) to close the session.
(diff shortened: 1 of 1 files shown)


docs/readmes/proposals/p024_magma_settlement_service.md
```diff
@@ -0,0 +1,280 @@
...
+
+Miner API update (session-forwarder <-> Miner) :
+`service api {
+rpc session_usage(session_usage_update) returns (orc8r.Void) {}
+}`
+
+#### 3.3.3 Helium Miner ↔︎ Network Server (router/helium-handler)
+
+For each 3gpp session Miner should forward session usage stats from Session-forwarder to Network Server.
+`service api {
```
electronjoe: Multi-line block quotes (here and elsewhere in this document) are handled by [Fenced Code Blocks](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#fenced-code-blocks). This will improve readability of what is otherwise displayed as a single line.

docs/readmes/proposals/p024_magma_settlement_service.md
```diff
@@ -0,0 +1,280 @@
...
+
+#### 3.2.1 State channel management procedure
+
+Following diagram shows e2e State Channel life cycle management that will be in real life. It consists of existing functionality (marked with black), new functionality to be implemented in magma (marked with blue).
+
+![State channel management procedure (1)](https://user-images.githubusercontent.com/93994458/149513469-0380c35c-d109-4c27-8014-d05b6cbd9a5e.png)
+
+This scenario consists of the steps below:
+
+1. Step [1-2] The State channel opening procedure. The Helium Handler sends Open State Channel Request to open the State Channel. State Channel is opened once the positive Response is received from State Channel.
```
electronjoe: Here and elsewhere. Maybe use bulleted list here instead of ordered list, as it's a bit confusing with the steps numbered 1-5 and bullets 1-3?

docs/readmes/proposals/p024_magma_settlement_service.md
Arsenii-Oganov: agree. done

docs/readmes/proposals/p024_magma_settlement_service.md
Arsenii-Oganov: agree. done

electronjoe (approved):
Per proposal process, reviewed for just formatting or typos, discussion of the proposal will take place on the associated GH Issue.

docs/readmes/proposals/p024_magma_settlement_service.md
```diff
@@ -0,0 +1,281 @@
...
+The detailed call flow for the procedure is specified in section 3.2.1 “State channel management procedure”.
+
+- UE attach procedure
+
+The procedure in which the UE registers to the network, and receives the usage quota to create the EPS Bearer between the UE and the PGW, in order to be able to send and receive data.
+The detailed call flow for the procedure is specified in section 3.2.2 “UE attach procedure”.
+
+- UE quota update procedure
+
+The procedure describes a next quota request by Threshold from the previously received quota from SGW to OCS.
```
uri200, line 43: Per my understanding,  CCR are sent by the PCEF which inbound roaming may be at PGW

docs/readmes/proposals/p024_magma_settlement_service.md
```diff
@@ -0,0 +1,281 @@
...
+#### 3.2.3 UE quota update procedure
+
+Following diagram shows e2e UE update scenario that will be in real life. It consists of existing functionality (marked with black), new functionality to be implemented in magma and helium (marked with blue).
+
+![Update call flow](https://user-images.githubusercontent.com/93994458/149513805-f78e1c18-bb12-4f1f-abf9-0a0ea5e09d20.png)
+
+This scenario consists of the steps below:
+
+- Step [1-2] Packet data transfer is ongoing between UE and Internet.
+- Step [3] The threshold of the quota is configured on SGW (e.g. 95% of granted service unit).
```
uri200, line 104: Currently on inbound roaming we do not use sessiond. We create a flow directly and it is the other operators PGW who does keep track of the data.

To achieve that you could try to refactor inbound roaming to use sessiond. Then sessiond create an appropriate flow for that UE with a hardcoded pass all rule tracking for all the traffic.

That would requiere changes on mme, sessiond and pipelined (to use the proper API call for roaming subscribers)

uri200: sessiond would then send the updates to session_proxy or to your own implementation of session_proxy (session_purchaser ?)

docs/readmes/proposals/p024_magma_settlement_service.md
```diff
@@ -0,0 +1,281 @@
...
+The following figure shows the integration interfaces between Magma domain and OCS /DCH.
+
+![Integration interfaces](https://user-images.githubusercontent.com/93994458/149514277-062e0834-cb8c-49af-b0e4-867c4664c694.png)
+
+#### 3.3.1 DIAMETER Gy interface specification
+
+Interface transport protocol: TCP/Diameter/DCCA.
+The 3GPP based Gy interface between Session Proxy and Session Purchaser to implement real-time quota management for the packet data service for the roaming subscribers. While this interface resembles Gy in almost every way, this is a non-standard use of the Gy protocol defined to allow accounting at the operator level and allow the session purchaser a control point to approve or deny sessions at initiation or at update points throughout a session.
+The Session Purchaser is the Diameter Credit Control server, which provides the online charging data to the Session Proxy based on volume quota.
+The connection between the Session Proxy (client) and Session Purchaser (server) is TCP based. There are a series of message exchanges to manage quota in real-time.
```
uri200, line 172: My suggestion is Session Purchaser to be implemented as a proper OCS, meaning, it should return CCA (it can be a hardcoded default resonse).  But that may allow you to use the system as is (sessiond and session proxy)

[here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Amagma%2Fmagma.git&branch=master&file=feg%2Fgateway%2Fservices%2Ftestcore%2Focs%2Fmock_ocs%2Fccr_handler.go&editor=JetBrains&version=v1.2.1&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2.2&start_row=28&start_col=4&end_row=28&end_col=4) you have a mock OCS. that may easily be extended to be your session purchaser

docs/readmes/proposals/p024_magma_settlement_service.md
```diff
@@ -0,0 +1,281 @@
...
+
+#### 3.3.2 MME/SGW ↔︎ Helium Miner
+
+Session lifecycle events and usage reports should be delivered from each deployed AGW to the Network-Server (helium-handler), which uses that information to update a state channel. The Helium Miner container does such communication between hotspot(AGW) and Network Server. So, a trusted channel should be formed between MME/SGW and Miner to let a Miner container communicate with the Helium network on behalf of AGW. For having a trusted channel between AGW and Miner new component is introduced - a session-forwarder. Session-forwarder is a mediator between Magma-specific components deployed on AGW and a Helium Miner.
+To let Miner trust Session-forwarder, session-forwarder is deployed on the same AGW where software is signed and Linux Secure Boot is used to verify all software components being loaded. So, Miner can trust to all received session-related messages. Miner just does forward all message to helium-handler.
```
uri200, line 177: there is no session concept on inbound roaming. We do not use sessiond.

To have session concept you will have to change current inbound roaming flow which is
```
Attach 					UE -> 	MME 
Create session Request				MME -> s8_poxy
Create session Response  			MME <- s8_proxy
create flow						MME -> ovs
Attach Accepted				UE <-	MME
```

You will need something like

```
Attach 					UE -> 	MME 
Create session Request				MME -> s8_poxy
Create session Response        			MME <- s8_proxy
create session					MME -> 	sessiond
create_session_purchaser			 -> 	sessiond -> session_proxy -> session_purchaser
create_session_purchaser_acctp.                  <- 	sessiond <- session_proxy <- session_purchaser
create flow						sessiond -> pipelined -> ovs
Attach Accepted				UE <-      MME <- 	sessiond
```

docs/readmes/proposals/p024_magma_settlement_service.md
Arsenii-Oganov: @uri200  thats correct, but in case of Inbound roaming we are building S8 connection between our SGW to Home network PGW. And we need to manage quotas for Helium blockchain on our side, so the only proper place to handle it - is our SGW.

docs/readmes/proposals/p024_magma_settlement_service.md
Arsenii-Oganov: that was our initial approach to  refactor sessiond to use it to push data to session_proxy.
If you may advise us better approach - you are welcome.

docs/readmes/proposals/p024_magma_settlement_service.md
Arsenii-Oganov: absolutely we'll use standard OCS as session purchaser

docs/readmes/proposals/p024_magma_settlement_service.md
Arsenii-Oganov: ok, we'll investigate s8_proxy

uri200:
Those are changes I think are required to involve sessiond on inbound roaming. Note this is a 10000 feet view, so many other details need to be checked.

1- MME
~~For inbound roaming, right now MME creates a new GTP to GTP tunnel on OVS after receiving create session response from s8. That flow is GTP to GTP flow from enb to PGW. In case to move this to sessiond, this step may not be needed. Note that for the non inbound roaming MME still creates an entry at table 0 for that subscriber (question is, do we still need to create an entry on table 0 from MME same way?)~~
MME will keep the logic to create table 0 flow on OVS. MME will need to add a call to sessiond to CreateSessionRequest. Same proces we do on non roaming proces. Probably we will need to add a field on that call to identify the roaming user.

2- Sessiond
Sessiond will need to have the concept of inbound roaming session and then use a specific new API call to pipelined to create an enforcement flow for that specific IP and IMSI

Right now MME has PGW address defined in its configuration, ~~that may need to change too~~ , I don't think this field is required

3- pipelined
~~Pipelined will need some new code to handle OVS inbound roaming (GTP to GTP) tunnel creation~~
Same code can be used. The flow is identified by IMSI and IP. An all pass rule can be set. This traffic can be tracked by sessiond using the same RG.

4- other changes on sessiond
We may need other changes on the Gx/Gy interface depending how the system is defined (use of policydb + OCS)

@pshelar can you comment in here please?

pshelar:
> Those are changes I think are required to involve sessiond on inbound roaming. Note this is a 10000 feet view, so many other details need to be checked.
> 
> 1- MME For inbound roaming, right now MME creates a new GTP to GTP tunnel on OVS after receiving create session response from s8. That flow is GTP to GTP flow from enb to PGW. In case to move this to sessiond, this step may not be needed. Note that for the non inbound roaming MME still creates an entry at table 0 for that subscriber (question is, do we still need to create an entry on table 0 from MME same way?)
> 
We still need the table0 flows, except we do not set the skip enforcement flag for s8 traffic.

> 2- Sessiond Sessiond will need to have the concept of inbound roaming session and then use a specific new API call to pipelined to create a GTP to GTP tunnel
> 
SessionD would create enforcement flow to allow and track the traffic from UE.
This can be wildcard flow that allows all traffic to/from the UE.

> Right now MME has PGW address defined in its configuration, that may need to change too
> 
> 3- pipelined Pipelined will need some new code to handle OVS inbound roaming (GTP to GTP) tunnel creation
> 
There is no need to change pipelineD in current form, current active-flows API would work fine.

> 4- other changes on sessiond We may need other changes on the Gx/Gy interface depending how the system is defined (use of policydb + OCS)
> 
> @pshelar can you comment in here please?

ardzoht (approved):
lgtm

uri200:
Thanks @pshelar I will correct my previous [message](https://github.com/magma/magma/pull/11167#issuecomment-1016799640)

uri200 (approved):

Neudrino:
Affected by #11350.

docs/readmes/proposals/p024_magma_settlement_service.md
```diff
@@ -0,0 +1,281 @@
...
+
+- State channel management procedure
+
+The procedure describes Opening, Update and Closure of State channel management call flow.
+The detailed call flow for the procedure is specified in section 3.2.1 “State channel management procedure”.
+
+- UE attach procedure
+
+The procedure in which the UE registers to the network, and receives the usage quota to create the EPS Bearer between the UE and the PGW, in order to be able to send and receive data.
+The detailed call flow for the procedure is specified in section 3.2.2 “UE attach procedure”.
```
uri200, line 39: Note to track a session with OCS, sessiond needs a RG.  Tracking of sessions are not based just on IMSI, but on each RG

On regular LTE, PCRF stores the rules allowed for each susbscriber and provides the RG.  Then that IMSI and RG are used to request and report traffic to OCS

Session Purchaser has to have the concept of RG in order to talk to sessiond (even if that is hardcoded)

docs/readmes/proposals/p024_magma_settlement_service.md
```diff
@@ -0,0 +1,281 @@
...
+- Step [3] The SGW sends CCR-I to the Session Purchaser requesting volume quota.
+- Step [4] Session Purchaser checks if the IMSI-prefix (i.e. the HPLMN operator) has balance to answer with granted data units on CCR-I Request which has been received on Step [3].
+- Step [5] In response to a CCR-I (Step [3]), the Session Purchaser returns a CCA-I message that indicates success (DIAMETER_SUCCESS) or failure (DIAMETER AUTHORIZATION REJECTED) depending on whether the HPLMN operator has sufficient credit for the requested services. In case of HPLMN operator has sufficient credit the CCA contains the Granted Service Unit in Bytes.
+- Step [6] The SGW sends signed Session Init Event to Miner. The message is signed with SGW PubKey.
+- Step [7] The Miner sends signed Session Init Event to Helium Handler. The message is signed with Miner PubKey.
+- Step [8 - 9] The Create Session Request and Response messages to establish UE requested PDN connectivity through the SGW and Home PGW.
```
uri200, line 91: this has to happen before any interaction with sessiond
So step 8-9 should happen after step 02.

docs/readmes/proposals/p024_magma_settlement_service.md
```diff
@@ -0,0 +1,281 @@
...
+#### 3.2.2 UE attach procedure
+
+Following diagram shows e2e UE attach scenario that will be in real life. It consists of existing functionality (marked with black), new functionality to be implemented in magma (marked with blue).
+The attach procedure implies that State Channel is already in open state.
+
+![Attach call flow](https://user-images.githubusercontent.com/93994458/149513666-63ae0f4b-c8be-4cfc-a739-909723fca3aa.png)
+
+This scenario consists of the steps below:
+
+- Step [1 - 2] Initial attach procedure in a mobile network. The UE establishes radio link synchronization with an eNB, the UE creates a connection for data delivery via an Attach Request message sent to the eNB, the eNB then forwards the attach request to an MME/SGW.
```
uri200, line 85: is the HSS involved at all on authentication?

docs/readmes/proposals/p024_magma_settlement_service.md
uri200: Note that we need a GX policy and GY.  So this solution should be configuration-wise. There should not be any modification on sessiond/feg to bypass that.  Either you provision subscribers at orc8r or you implement a service that responds to both Gx.Gy to provide with the proper response (GX will respond with a rule and RG and GY will respond with credit for that RG) 

On the orc8r level you could try to explore omni rules to see if they will work. Omnirules are rules defined on orc8r that will apply to any subscribers that try to send a request through session_proxy.  That means that all your CCR may save GX request and will do the Gy request using that omni rule.

uri200: Note you can just use a standard PCRF/OCS like sig scale and modify it.

You can also look into our mock_PCRF and mock_OCS which are almost functional PCRF and OCS

uri200:
Note that to be able to support a wide range of use cases we need to follow 3gpp as much as possible. So changes on well know Magma call flows for the fact of adding customized features may shrink Magma scope. We should be cautious with those modifications. 

Anything that can be done as a configuration, should be done externally to the codebase  Anything that can be done by a core 3GPP service should be considered instead of adding code to Magma. Addition of separate/independent services should be fine but changes on current AGW/FEG services should be based on current Magma philosophy and be 3GPP friendly.
_______________________________________________

To understand how meaningful this proposal is, I think we should separate into two different proposals:

Proposal 1: Addition of enforcement to inbound roaming (ie, include sessiond on inbound roaming call flow)
Proposal 2: Addition of mining services and FFI specific interfaces (ie creation of miner, session forwarder, helim handler and modification of Magma API to interact with them)

**Proposal 1** has potential to anybody using inbound roaming (which will be the majority of operators/users of Magma). Magma doesn't have a way to track/enforce usage of subscribers in inbound roaming. Even this is not a 3gpp feature it can be useful to have control over roaming users. Having sessiond involved on inbound roaming may be helpful in the future in case of adding outbound roaming. 

However, that is only useful when this feature is built as **stand alone feature**. Meaning, we should be able to use enforcement service without the need to have any other FFI interface or miner service. 

So Proposal 1 may include modifications on mme, sessiond and creation of session purchaser (which should be implemented as a partial OCS so it can be extended in the future as a full OCS). I think creation of CDR and CDR engine needs a better specification, should be taken as a sub proposal and worked in parallel to the other features mentioned here.  

All those modifications should not impact current  inbound roaming or non inbound roaming use cases and we should be able to enable/disable usage of sessiond on inbound roaming.

We should be able to test this feature end to end ([for example extending Federated Integ Test](https://docs.magmacore.org/docs/next/feg/s1ap_federated_tests#s1ap-federated-integration-tests)), so a new [mock_pgw](https://github.com/magma/magma/tree/master/feg/gateway/services/s8_proxy/servicers/mock_pgw#L117-L117) and some modifications on S1AP tester wrapper may be also needed

In picture below you have a suggested call flow for attach for this case.  Note this call flow follows 3GPP/Magma call flow and we just add sessiond/CCRs.  This call flow is still missing interactions between feg or pipelined that should not be ignored.

inbound_roaming_with_sessiond

**Proposal 2** has a limited potential compared to Proposal 1.  Those services will only  interact with FFI so it is difficult to say how much impact they have to the whole Magma community. In my opinion they should be part of a different proposal and evaluated separately. 

Note that once Proposal 1 is done, other services like Miner, Helim handler can be built on top of that without disturbing main Magma call flow.


Concerns
- Is HSS involved on attach? Are the subscribers provisioned on remote HSS?
- Why can't we use same CCR U from sessiond on CDR/TAP engine
- FEG is a host of interfaces, CDR engine should be just translating, not storing data.
- Why do you need both Gy and TAP? can't just you use one? Can't your new OCS/session purchaser build those TAPs?

Arsenii-Oganov:
@uri200 I agree with your direction, that these shall be at least 2 separate and independent PRs.

uri200:
> @uri200 I agree with your direction, that these shall be at least 2 separate and independent PRs.

Note I am not talking about PR but proposals and probably two different grants. 

On the PR front, we expect this to be many multiple small.PR each with it's proper testing. Note we will not accept huge PR specially when this is modifying core systems.

Arsenii-Oganov:
In this case we just need to add Proposal 1, cause current proposal fully represents Proposal 2 do you agree?

uri200:
You can use this proposal for 2, but note that call flows will be impacted by the new proposal 1 so you will need to modify this one accordingly.

As mentioned, the idea of proposal 1 is to create an independent functionality to include enforcement in inbound roaming, with no mining/helium dependency.

Arsenii-Oganov:
Agree

ardzoht:
This also helps with keeping the scope of changes separate in terms of functionality. As @uri200 mentioned Proposal 1 should be focused on the design of all involved changes to inbound roaming / magma services. And Proposal 2 will be for defining the changes for session-forwarder and Helium miner.

Issue #10437, opened 2021-11-18
feat: get domain name in orc8r 


reginawang3495:
to be used in orc8r/cloud/go/services/bootstrapper/servicers/cloud_registration.go

https://github.com/magma/magma/blob/0480647742f57e99853f04fc86f8f831a06bd0d6/orc8r/cloud/go/services/bootstrapper/servicers/registration/cloud_registration.go#L139-L142

PR #11169, opened 2022-01-14
feat: Get domain name in orc8r

christinewang5:
## Summary 
Get orc8r domain name for simple gateway registration
Resolving #10437 

## Test
* Getting domain name while running locally via Docker
image
* Getting domain name while deploying on Minikube & AWS 
gh_logs
gh_pod_description

### DIFF
--- a/orc8r/cloud/docker/docker-compose.yml
+++ b/orc8r/cloud/docker/docker-compose.yml
@@ -17,6 +17,7 @@ services:
 #      DATABASE_SOURCE: "magma_dev:magma_dev@(maria)/magma_dev"
       SQL_DIALECT: psql
       SERVICE_REGISTRY_MODE: yaml
+      ORC8R_DOMAIN_NAME: magma.test
       VERSION_TAG: LOCAL-DEV
       HELM_VERSION_TAG: LOCAL-DEV
       MAGMA_PRINT_GRPC_PAYLOAD: 0
--- a/orc8r/cloud/go/services/bootstrapper/bootstrapper/main.go
+++ b/orc8r/cloud/go/services/bootstrapper/bootstrapper/main.go
@@ -15,8 +15,10 @@ package main
 
 import (
 	"crypto/rsa"
+	"errors"
 	"flag"
 	"io/ioutil"
+	"os"
 	"time"
 
 	"github.com/golang/glog"
@@ -92,9 +94,15 @@ func createRegistrationServicers(srv *service.OrchestratorService) (protos.Cloud
 		glog.Fatalf("failed to get rootCA: %+v", err)
 	}
 
+	domainName, err := getDomainName()
+	if err != nil {
+		glog.Errorf("failed to get domainName: %+v", err)
+	}
+
 	timeoutDurationInMinutes := srv.Config.MustGetInt(bootstrapper_config.TokenTimeoutDurationInMinutes)
 	timeout := time.Duration(timeoutDurationInMinutes) * time.Minute
-	cloudRegistrationServicer, err := registration.NewCloudRegistrationServicer(store, rootCA, timeout, true)
+
+	cloudRegistrationServicer, err := registration.NewCloudRegistrationServicer(store, rootCA, domainName, timeout, true)
 	if err != nil {
 		glog.Fatalf("error creating cloud registration servicer: %+v", err)
 	}
@@ -110,4 +118,12 @@ func getRootCA() (string, error) {
 		return "", err
 	}
 	return string(body), nil
+}
+
+func getDomainName() (string, error) {
+	domainName, ok := os.LookupEnv("ORC8R_DOMAIN_NAME")
+	if !ok {
+		return "", errors.New("failed to get orc8r domain name")
+	}
+	return domainName, nil
 }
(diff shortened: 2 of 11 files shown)


andreilee (commented):
You can run a manual test with a Docker setup. Otherwise you'll want to bring up a Magma setup on AWS to check the terraform changes.

reginawang3495:
Not sure if I'm overstepping/if it's right for this case, but I previously[ deployed on minikube](https://magma.github.io/magma/docs/next/orc8r/dev_minikube) to test the helm charts and such!

related: https://github.com/magma/magma/pull/10259

orc8r/cloud/go/services/bootstrapper/servicers/registration/cloud_registration.go
```diff
@@ -139,7 +147,16 @@ func (c *cloudRegistrationServicer) generateAndSaveTokenInfo(networkID string, l
...
-// TODO(#10437)
-func getDomainName() string {
-	return NotImplementedWarning
+func getDomainName() (string, error) {
+	// Set env variable manually if running under test conditions
+	if strings.HasSuffix(os.Args[0], ".test") {
+		os.Setenv(Orc8rDomainNameKey, Orc8rDomainNameTest)
+		return Orc8rDomainNameTest, nil
+
+	}
```
christinewang5: This is so sketch, but I tried setting the Env variable in the unit tests, but it doesn't seem to get passed along to here. I'm not sure what environment the tests are running in

andreilee: If you can't test this function properly, it's probably better to just mock out the entirety of `getDomainName` during tests, instead of including this janky logic here

andreilee: That said, it's a good investment to figure out the peculiarities of the test environment (or runtime env, etc.), otherwise we risk having workarounds like these pile up, or worse, making poor architectural decisions.

orc8r/cloud/go/services/bootstrapper/servicers/registration/cloud_registration.go
```diff
@@ -28,6 +31,8 @@ import (
 
 // NotImplementedWarning is pulled out of getDomainName for ease of testing
 const NotImplementedWarning = "warning: not implemented"
```
hcgatewood: Think we can delete this const now

orc8r/cloud/go/services/bootstrapper/servicers/registration/cloud_registration.go
hcgatewood: Ha, yes, this is pretty jank

Instead, let's pass in the domain name to the servicer upon servicer creation. Then have 1 constructor for the servicer that takes the domain name as a string arg, and 1 "helper" constructor that (a) reads the domain name from the env var then (b) calls the normal constructor by passing in the read value

hcgatewood (changes requested):
Awesome! Couple small comments, but other than that this looks great 💯

orc8r/cloud/go/services/bootstrapper/servicers/registration/cloud_registration.go
christinewang5: Ok cool, that sounds much better, thanks for suggestion 😊

orc8r/cloud/docker/docker-compose.yml
```diff
@@ -17,6 +17,7 @@ services:
 #      DATABASE_SOURCE: "magma_dev:magma_dev@(maria)/magma_dev"
       SQL_DIALECT: psql
       SERVICE_REGISTRY_MODE: yaml
+      ORC8R_DOMAIN_NAME: LOCAL-DEV
```
hcgatewood: So this should probably be the URL that the gateway currently uses to reach the Orc8r. And clarifying which one that is is probably a good exercise for during the CLI PR -- so maybe let's push merging this PR until we clarify that, thoughts?

orc8r/cloud/go/services/bootstrapper/bootstrapper/main.go
```diff
@@ -92,9 +94,15 @@ func createRegistrationServicers(srv *service.OrchestratorService) (protos.Cloud
 		glog.Fatalf("failed to get rootCA: %+v", err)
 	}
 
+	domainName, err := getDomainName()
+	if err != nil {
+		glog.Fatalf("failed to get domainName: %+v", err)
```
hcgatewood: Let's just make this an error. Because on-prem deployments may not have this var set, and it's not a show-stopping error

orc8r/cloud/helm/orc8r/values.yaml
```diff
@@ -156,7 +156,8 @@ controller:
     repository: docker.artifactory.magmacore.org/controller
     tag: 1.6.1
     pullPolicy: IfNotPresent
-
+    env:
+      orc8r_domain_name: magma.test
```
hcgatewood, line 160: This should probably match the recommendation from the docker-compose file

hcgatewood (approved):

PR #11193, opened 2022-01-18
feat(orc8r): Mini Magma Orchestrator

### LINKED ISSUE #11045: [Proposal] Mini Magma Orchestrator
# Mini Magma Orchestrator
## Problem

A lot of community members struggle when it comes to deploying Magma Orchestrator on Virtual Machines or Bare Metal and AWS is too costly for them to start with.

## Solution

We can build an automated process that will deploy Magma Orchestrator with a single bash script, the way we do today with Magma Access Gateway.

We will have to limit our resources to the following.

4 CPU/vCPU
8GB RAM
100GB Storage
1 Public/Private IP

If we deploy this solution on AWS it should not cost more than $100/month.
https://calculator.aws/#/createCalculator/EC2

Once after testing Mini Magma Orchestrator they can deploy terraform managed Orc8r on AWS.

## Non-goals

As everything will be running inside a single node the solution won't be fault-tolerant.

ShubhamTatvamasi:

## Summary

As part of the **[Proposal] Mini Magma Orchestrator** #11045

Here is the deployment code for the single node Magma Orchestrator.

## Test Plan

Deployment steps are mentioned in README.md

## Additional Information


orc8r/cloud/deploy/mini-magma/hosts.yml
```diff
@@ -0,0 +1,7 @@
+---
+all:
+  hosts: 192.168.5.70
```
mattymo, line 4: This IP will change per-deployment. I would add a comment here that it needs to be updated (just in case people don't read the readme completely)

ShubhamTatvamasi: Sure @mattymo, I will update it with a comment as per your suggestion.

hcgatewood:
Deferring to @arunuke to review as @magma/approvers-orc8r-infra, since I have voiced my mixed feelings on this plan

orc8r/cloud/deploy/mini-magma/deploy-orc8r.yml
```diff
@@ -0,0 +1,16 @@
+---
```
maxhbr, line 1: the files here are missing license headers

maxhbr (changes requested):

ShubhamTatvamasi:
Super old PR, will try to create a fresh one in future.

PR #11195, opened 2022-01-18
feat: Add cli for simple gateway registration

christinewang5:
## Summary
register CLI for gateways

## Testing
manually tested by running CLIs 
image

Note: this PR is dependent on #11169 & #11692

### DIFF
--- a/orc8r/cloud/go/service/middleware/unary/identity_decorator_bypass.go
+++ b/orc8r/cloud/go/service/middleware/unary/identity_decorator_bypass.go
@@ -19,4 +19,5 @@ package unary
 var identityDecoratorBypassList = map[string]struct{}{
 	"/magma.orc8r.Bootstrapper/GetChallenge": {},
 	"/magma.orc8r.Bootstrapper/RequestSign":  {},
+	"/magma.orc8r.Registration/Register":     {},
 }
--- a/orc8r/gateway/python/magma/common/service_registry.py
+++ b/orc8r/gateway/python/magma/common/service_registry.py
@@ -19,6 +19,7 @@ from magma.configuration.exceptions import LoadConfigError
 from magma.configuration.service_configs import load_service_config
 
 GRPC_KEEPALIVE_MS = 30 * 1000
+ROOTCA_PATH = '/var/opt/magma/certs/rootCA.pem'
 
 
 class ServiceRegistry:
@@ -84,6 +85,31 @@ class ServiceRegistry:
         Removes all the entries in the registry
         """
         ServiceRegistry.get_registry()["services"] = {}
+
+    @staticmethod
+    def get_bare_bootstrap_rpc_channel(domain_name, port, rootca_path=None):
+        """
+        Return an RPC channel to bootstrap service in CLOUD, without referencing a
+        control proxy config.
+
+        Returns:
+            grpc channel
+        """
+        rootca_path = rootca_path or ROOTCA_PATH
+        with open(rootca_path, 'rb') as rootca_f:
+            rootca = rootca_f.read()
+            ssl_creds = grpc.ssl_channel_credentials(
+                root_certificates=rootca,
+            )
+
+        authority = 'bootstrapper-controller.%s' % domain_name
+        grpc_options = [('grpc.default_authority', authority)]
+        channel = grpc.secure_channel(
+            target='%s:%s' % (authority, port),
+            credentials=ssl_creds,
+            options=grpc_options,
+        )
+        return channel
 
     @staticmethod
     def get_bootstrap_rpc_channel():
--- a/orc8r/gateway/python/scripts/register.py
+++ b/orc8r/gateway/python/scripts/register.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python3
+"""
+Copyright 2020 The Magma Authors.
+
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import argparse
+import sys
+import textwrap
+from typing import List
+
+import checkin_cli
+import snowflake
+from magma.common.cert_utils import load_public_key_to_base64der
+from magma.common.service_registry import ServiceRegistry
+from magma.configuration.service_configs import save_override_config
+from orc8r.protos.bootstrapper_pb2 import (
+    ChallengeKey,
+    RegisterRequest,
+    RegisterResponse,
+)
+from orc8r.protos.bootstrapper_pb2_grpc import RegistrationStub
+from orc8r.protos.identity_pb2 import AccessGatewayID
+
+
+def register_handler(client: RegistrationStub, args: List[str]) -> RegisterResponse:
+    """
+    Register a device and retrieves its control proxy
+    Args:
+        client: Registration stub
+        args: command line arguments
+    Returns:
+        RegisterRequest: register request, used for printing after function returns
+        RegisterResponse: response from gRPC call, either error or the control_proxy
+    """
+    req = RegisterRequest(
+        token=args.token,
+        hwid=AccessGatewayID(
+            id=snowflake.snowflake(),
+        ),
+        challenge_key=ChallengeKey(
+            key=load_public_key_to_base64der("/var/opt/magma/certs/gw_challenge.key"),
+            key_type=ChallengeKey.KeyType.SOFTWARE_ECDSA_SHA256,
+        ),
+    )
+
+    res = client.Register(req)
+    if res.HasField("error"):
+        raise Exception(res.error)
+
+    return req, res
+
+
+def main():
+    """Register a gateway"""
+    parser = argparse.ArgumentParser(description="Register a gateway.")
+    parser.add_argument(
+        "domain",
+        metavar="DOMAIN_NAME",
+        type=str,
+        help="orc8r's domain name",
+    )
+    parser.add_argument(
+        "token",
+        metavar="REGISTRATION_TOKEN",
+        type=str,
+        help="registration token after API call",
+    )
+    parser.add_argument(
+        "--ca-file",
+        type=str,
+        help="orc8r's root CA file",
+    )
+    parser.add_argument(
+        "--cloud-port",
+        type=str,
+        help="orc8r's port",
+    )
+    parser.add_argument(
+        "--no-control-proxy",
+        action="store_true",
+        help="disables writing the control proxy file",
+    )
+    args = parser.parse_args()
+    chan = ServiceRegistry.get_bare_bootstrap_rpc_channel(
+        args.domain,
+        '8444' if not args.cloud_port else args.cloud_port,
+        args.ca_file,
+    )
+    client = RegistrationStub(chan)
+    try:
+        req, res = register_handler(client, args)
+        msg = textwrap.dedent(
+            """
+            > Registered gateway
+            Hardware ID
+            -----------
+            {}
+            Challenge Key
+            -----------
+            {}
+            Control Proxy
+            -----------
+            {}
+            """,
+        )
+        print(msg.format(req.hwid, req.challenge_key, res.control_proxy))
+    except Exception as e:
+        msg = textwrap.dedent(" > Error: {} ")
+        print(msg.format(e))
+        sys.exit(1)
+
+    if not args.no_control_proxy:
+        save_override_config("control_proxy", res.control_proxy)
+
+
+if __name__ == "__main__":
+    main()
+    print("> Running checkin_cli")
+    checkin_cli.main()
--- a/orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
+++ b/orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
@@ -33,12 +33,7 @@ func NewRegistrationServicer() protos.RegistrationServer {
 }
 
 func (r *RegistrationService) Register(c context.Context, request *protos.RegisterRequest) (*protos.RegisterResponse, error) {
-	nonce, err := NonceFromToken(request.Token)
-	if err != nil {
-		return nil, err
-	}
-
-	deviceInfo, err := r.GetGatewayDeviceInfo(context.Background(), nonce)
+	deviceInfo, err := r.GetGatewayDeviceInfo(context.Background(), request.Token)
 	if err != nil {
 		clientErr := makeErr(fmt.Sprintf("could not get device info from token %v: %v", request.Token, err))
 		return clientErr, nil


orc8r/cloud/docker/controller/supervisord.conf
```diff
@@ -31,7 +31,7 @@ stdout_events_enabled=true
 stderr_events_enabled=true
 
 [program:bootstrapper]
-command=/usr/bin/envdir /var/opt/magma/envdir /var/opt/magma/bin/bootstrapper -cak=/var/opt/magma/certs/bootstrapper.key -logtostderr=true -v=0
+command=/usr/bin/envdir /var/opt/magma/envdir /var/opt/magma/bin/bootstrapper -cak=/var/opt/magma/certs/bootstrapper.key -logtostderr=true -v=20
```
hcgatewood: Nit: revert these 2

orc8r/cloud/docker/docker-compose.yml
```diff
@@ -17,6 +17,7 @@ services:
 #      DATABASE_SOURCE: "magma_dev:magma_dev@(maria)/magma_dev"
       SQL_DIALECT: psql
       SERVICE_REGISTRY_MODE: yaml
+      ORC8R_DOMAIN_NAME: LOCAL-DEV
```
hcgatewood: This should probably be `magma.test`

orc8r/cloud/go/service/middleware/unary/identity_decorator_bypass.go
```diff
@@ -17,6 +17,8 @@ package unary
 // bypass Identity verification checks.
 // For now, the only service in this category is Bootstrapper.
 var identityDecoratorBypassList = map[string]struct{}{
-	"/magma.orc8r.Bootstrapper/GetChallenge": {},
-	"/magma.orc8r.Bootstrapper/RequestSign":  {},
+	"/magma.orc8r.Bootstrapper/GetChallenge":              {},
+	"/magma.orc8r.Bootstrapper/RequestSign":               {},
+	"/magma.orc8r.Registration/Register":                  {},
+	"/magma.orc8r.CloudRegistration/GetGatewayDeviceInfo": {},
```
hcgatewood: Shouldn't need to add this here -- "CloudRegistration" is called by "Registration", rather than directly from gw

Screen Shot 2022-02-23 at 2 32 47 PM

orc8r/cloud/go/services/bootstrapper/bootstrapper/main.go
```diff
@@ -92,9 +94,15 @@ func createRegistrationServicers(srv *service.OrchestratorService) (protos.Cloud
 		glog.Fatalf("failed to get rootCA: %+v", err)
 	}
 
+	domainName, err := getDomainName()
+	if err != nil {
+		glog.Fatalf("failed to get domainName: %+v", err)
```
hcgatewood: Let's just make this an err, rather than fatal

orc8r/gateway/python/magma/common/service_registry.py
```diff
@@ -85,6 +85,34 @@ def reset():
         """
         ServiceRegistry.get_registry()["services"] = {}
 
+    @staticmethod
+    def get_bare_rpc_channel(rootca_path=None):
+        """
+        Returns a RPC channel to bootstrap service in CLOUD before having
+        control proxy config.
```
hcgatewood:
```suggestion
        Return an RPC channel to bootstrap service in CLOUD, without referencing a
        control proxy config.
```

orc8r/gateway/python/magma/common/service_registry.py
```diff
@@ -85,6 +85,34 @@ def reset():
         """
         ServiceRegistry.get_registry()["services"] = {}
 
+    @staticmethod
+    def get_bare_rpc_channel(rootca_path=None):
```
hcgatewood: If we're gonna have this only target bootstrapper, then should reflect that in the func name

orc8r/gateway/python/magma/common/service_registry.py
```diff
@@ -85,6 +85,34 @@ def reset():
...
+        """
+        Returns a RPC channel to bootstrap service in CLOUD before having
+        control proxy config.
+
+        Returns:
+            grpc channel
+        """
+        ip = 'bootstrapper-controller.magma.test'
```
hcgatewood: The `magma.test` should be passed in as an arg

orc8r/gateway/python/magma/common/service_registry.py
```diff
@@ -85,6 +85,34 @@ def reset():
...
+        port = '7444'
```
hcgatewood: Same thing for port

orc8r/gateway/python/magma/common/service_registry.py
```diff
@@ -85,6 +85,34 @@ def reset():
...
+
+        rootca_path = '/var/opt/magma/certs/rootCA.pem' if not rootca_path else rootca_path
+        with open(rootca_path, 'rb') as rootca_f:
+            rootca = rootca_f.read()
+            ssl_creds = grpc.ssl_channel_credentials(
+                root_certificates=rootca,
+            )
+
+        authority = 'bootstrapper-controller.magma.test'
```
hcgatewood: Same

orc8r/gateway/python/scripts/generate_nghttpx_config.py
```diff
@@ -62,7 +62,7 @@ def get_context():
 
 def main():
     logging.basicConfig(
-        level=logging.INFO,
+        level=logging.DEBUG,
```
hcgatewood: Revert before merging

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,168 @@
...
+from magma.common.cert_utils import load_public_key_to_base64der
+from orc8r.protos.bootstrapper_pb2 import (
+    RegisterRequest,
+    RegisterResponse,
+    GetGatewayDeviceInfoRequest,
+    GetGatewayDeviceInfoResponse,
+)
+from orc8r.protos.bootstrapper_pb2_grpc import (
+    RegistrationStub,  # Register
+    CloudRegistrationStub,  # Register
```
hcgatewood: Fmu why the comments?

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,168 @@
...
+)
+from magma.common.service_registry import ServiceRegistry
+
+
+def bare_grpc_wrapper(func):
```
hcgatewood: Maybe like "bare_cloud_grpc_wrapper"

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    Wraps a function with a gRPC wrapper which creates a RPC client to
+    the service and handles any RPC Exceptions.
+
+    Usage:
+    @cloud_grpc_wrapper
```
hcgatewood: Match to func name

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,168 @@
...
+    
+    Returns:
+        RegisterResponse: response from gRPC call, either error or the control_proxy
+    """
+    req = RegisterRequest()
+    req.token = args.token
+    req.hwid.id = snowflake.snowflake()
+    req.challenge_key.key = load_public_key_to_base64der(
+        "/var/opt/magma/certs/gw_challenge.key")
+    req.challenge_key.key_type = 0
```
hcgatewood: Can you create this as a literal rather than procedurally? E.g. how you construct a literal struct in golang

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,168 @@
...
+
+    res = client.Register(req)
+    if res.HasField("error"):
+        raise Exception(res.error)
+
+    print("Successfully registered gateway")
+    print("Hardware ID\n-----------\n%s" % req.hwid)
+    print("Challenge Key\n-----------\n%s\n" % req.challenge_key)
+    print("Control Proxy\n-----------\n%s\n\n" % res.control_proxy)
```
hcgatewood: Use one big formatted string e.g. https://github.com/magma/magma/blob/33cd34820810ed71947aca9a6011182496c71324/orc8r/gateway/python/scripts/checkin_cli.py#L146-L152

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,168 @@
...
+
+    return res
+
+
+@bare_grpc_wrapper
+def get_gateway_device_info_handler(client: CloudRegistrationStub, args: List[str]) -> GetGatewayDeviceInfoResponse:
```
hcgatewood: As noted, this isn't necessary

And then, because we are only making a single request to the cloud, we may not need to use the "bare_grpc_wrapper" that you created -- just make the request manually, e.g. https://github.com/magma/magma/blob/33cd34820810ed71947aca9a6011182496c71324/orc8r/gateway/python/scripts/checkin_cli.py#L175-L194

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,168 @@
...
+    Returns:
+        None
+    """
+    os.system("sudo mkdir /var/opt/magma/configs")
+    os.system("sudo touch /var/opt/magma/configs/control_proxy.yml")
+    os.system("sudo chmod 777 /var/opt/magma/configs/control_proxy.yml")
+
+    with open('/var/opt/magma/configs/control_proxy.yml', 'w+') as control_proxy_f:
+        control_proxy_f.write(control_proxy)
+    control_proxy_f.close()
```
hcgatewood: I think we can do this without resorting to shell cmds. Maybe something like https://stackoverflow.com/questions/12517451/automatically-creating-directories-with-file-output

hcgatewood (changes requested):
Woo looks great, couple small things but happy we got it working!

orc8r/cloud/docker/docker-compose.yml
christinewang5: Idk why I rebased this PR onto the domain name PR, but I just undid the rebase and fixed all these in the original domain name PR.

orc8r/gateway/python/scripts/register.py
christinewang5: Yeah so I was originally doing that, but because we are trying to create a file from root, it threw a `Permission Denied`. 
I think it needs sudo permissions to make a dir in the `/var` folder, was also trying to chmod and setting umask, neither worked.

https://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-in-python

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,141 @@
...
+    parser.add_argument(
+        "--no_control_proxy",
+        action="store_true",
+        help="disables writing the control proxy file"
+    )
+    args = parser.parse_args()
+
+    chan = ServiceRegistry.get_bare_bootstrap_rpc_channel(
+        args.domain,
+        '7444' if args.domain == 'magma.test' else '8444',
```
christinewang5: not sure if there is a better way to tell whether we are in test env or not??

orc8r/gateway/python/scripts/register.py
christinewang5: Hey @ardzoht! 
Was wondering if there are better ways to grant write access to a file from the AGW root. Are there unintended side effects of doing a chmod 777 like this? Thanks in advance!

ardzoht: @christinewang5 I'd recommend to look up for the function handlers on `service_registry`, I think we can make use of these that should accomplish the same thing: https://sourcegraph.com/github.com/magma/magma/-/blob/orc8r/gateway/python/magma/configuration/service_configs.py?L47:5

The only caveat is that this probably will require running the script with sudo permissions, but it should work and the code is cleaner.

christinewang5: Thanks for the pointer! @ardzoht 
Could you please advise on how to run the script with sudo permissions in the magtivate virtualenv? 
When I run it, I have to put the whole abs path there like (`sudo /home/vagrant/magma/orc8r/gateway/python/scripts/register.py`) and even then it throws a cannot find the module snowflake.

I tried several things like running with the -E option [1] and also changing /etc/sudoers which seems sketch [2] without avail, but it seems like running with sudo is using a diff python environment. 
[1] [https://stackoverflow.com/questions/50315645/cannot-run-python-script-using-sudo](https://stackoverflow.com/questions/50315645/cannot-run-python-script-using-sudo)
[2] [https://stackoverflow.com/questions/25346171/cant-import-module-when-using-root-user](https://stackoverflow.com/questions/25346171/cant-import-module-when-using-root-user)

orc8r/gateway/python/scripts/register.py
hcgatewood: I'd say let's make it a CLI arg, default to 8444. Then in test env you have to pass e.g. `--cloud-port 7444`

orc8r/gateway/python/magma/common/service_registry.py
```diff
@@ -85,6 +85,33 @@ def reset():
...
+        """
+        Return an RPC channel to bootstrap service in CLOUD, without referencing a
+        control proxy config.
+
+        Returns:
+            grpc channel
+        """
+        ip = 'bootstrapper-controller.%s' % domain_name
+
+        rootca_path = '/var/opt/magma/certs/rootCA.pem' if not rootca_path else rootca_path
```
ardzoht: can we define the filepath as a constant?

orc8r/gateway/python/magma/common/service_registry.py
```diff
@@ -85,6 +85,33 @@ def reset():
...
+        with open(rootca_path, 'rb') as rootca_f:
+            rootca = rootca_f.read()
+            ssl_creds = grpc.ssl_channel_credentials(
+                root_certificates=rootca,
+            )
+
+        authority = 'bootstrapper-controller.%s' % domain_name
```
ardzoht, line 105: is `authority` the same as `ip`?

christinewang5: in this case, yes, will just use `authority` for uniformity

ardzoht (approved):
lgtm

orc8r/gateway/python/magma/common/service_registry.py
```diff
@@ -85,6 +86,31 @@ def reset():
...
+    @staticmethod
+    def get_bare_bootstrap_rpc_channel(domain_name, port, rootca_path=None):
+        """
+        Return an RPC channel to bootstrap service in CLOUD, without referencing a
+        control proxy config.
+
+        Returns:
+            grpc channel
+        """
+        rootca_path = ROOTCA_PATH if not rootca_path else rootca_path
```
hcgatewood: I believe you can do

```suggestion
        rootca_path = rootca_path or ROOTCA_PATH
```

hcgatewood: So if no rootca path is passed in, we want to use the default system credentials. I *think* that means you don't have to pass any ssl creds, but not sure how it works. But if you want, we can leave it like this for now, and I'll test it when I get around to making the Orc8r<>Let'sEncrypt work e2e

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,124 @@
...
+        type=str,
+        help="orc8r's root CA file",
+    )
+    parser.add_argument(
+        "--cloud-port",
+        type=str,
+        help="orc8r's port",
+    )
+    parser.add_argument(
+        "--no_control_proxy",
```
hcgatewood: Your args are mixing dash-case and camel_case -- let's use dash-case since it's historically the default for CLIs

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,124 @@
...
+        metavar="DOMAIN_NAME",
+        type=str,
+        help="orc8r's domain name",
+    )
+    parser.add_argument(
+        "token",
+        metavar="REGISTRATION_TOKEN",
+        type=str,
+        help="registration token after API call",
+    )
```
hcgatewood, line 76: Is there a way to mark these arguments as required (and fail/throw err when they're not present) ?

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,124 @@
...
+    try:
+        res = register_handler(client, args)
+    except Exception as e:
+        msg = textwrap.dedent(" > Error: {} ")
+        print(msg.format(e))
+        sys.exit(1)
+
+    if not args.no_control_proxy:
+        save_override_config("control_proxy", res.control_proxy)
+
```
hcgatewood, line 123: Final thing we wanna do here is run the "checkin_cli.py" functionality, so that the operator can have confidence that the registration was fully successful

hcgatewood (changes requested):
Looking good! Couple small things

orc8r/gateway/python/scripts/register.py
christinewang5: Soo because these are considered `positional arguments` , python's `argparse` already throws an error when these are not present

hcgatewood: Oh okay cool. And the "--xyz" args are magically non-positional, then? Nice

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,128 @@
...
+            key=load_public_key_to_base64der("/var/opt/magma/certs/gw_challenge.key"),
+            key_type=ChallengeKey.KeyType.SOFTWARE_ECDSA_SHA256,
+        ),
+    )
+
+    res = client.Register(req)
+    if res.HasField("error"):
+        raise Exception(res.error)
+
+    msg = textwrap.dedent(
```
hcgatewood: Nit: I'd prob put this printed msg in the main func, to keep this func as "stateless" as possible

orc8r/gateway/python/scripts/register.py
```diff
@@ -0,0 +1,128 @@
...
+        sys.exit(1)
+
+    if not args.no_control_proxy:
+        save_override_config("control_proxy", res.control_proxy)
+
+
+if __name__ == "__main__":
+    main()
+    print("> Running checkin_cli")
+    checkin_cli.main()
```
hcgatewood, line 128: Oh dang does this work? Nice good to know

hcgatewood (approved):
Lgtm!

orc8r/gateway/python/scripts/register.py
christinewang5: Yeah looks a bit sketch but it does 🤷🏻‍♀️

PR #11210, opened 2022-01-19
docs(agw): Document Sentry client filters

sebathomas:
## Summary

Add documentation for client filters introduced in #10276.

## Additional Information

### DIFF
--- a/docs/readmes/lte/configure_sentry.md
+++ b/docs/readmes/lte/configure_sentry.md
@@ -29,7 +29,7 @@ To configure Sentry, you will need to create a pull request to update the config
 
 Reporting for Python services, MME, and SessionD will *only* be enabled if the corresponding URL fields are non-empty.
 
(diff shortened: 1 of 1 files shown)


alexzurbonsen (commented):
As far as I can see, the `/networks/{network_id}/gateways/{gateway_id}` endpoint currently doesn't allow the sentry configs to be set. The schema 'network_sentry_config' is currently only used in the other two endpoints.
We might want to adapt the existing docs as well.

sebathomas:
@alexzurbonsen Good catch, changed the existing documentation too.

themarwhal (approved):

docs/readmes/lte/configure_sentry.md
```diff
@@ -44,6 +44,28 @@ sentry_upload_mme_log: false
 sentry_sample_rate: 1.0
 ```
 
+## Client-side filters for error messages
+
+Another configuration option is a list of error messages that should be filtered and not be sent to Sentry. This is done to avoid spamming Sentry with errors that can occur frequently but that are not bugs to be monitored.
```
themarwhal, line 49: Maybe we can say that this is similar in concept to the server side fliter configuration made available by Sentry.io: https://docs.sentry.io/product/accounts/quotas/#inbound-data-filters

sebathomas: Added some information about the server-side filters.

Neudrino:
Affected by #11350.

Issue #11092, opened 2022-01-07
Migrate C/C++ services to pull Sentry config from shared config


themarwhal:
There's more information in the parent task that might be useful. But tldr is that we're moving out the sentry configuration values to be specified under a 'shared_config' section that is shared by all services.

1. Modify SessionD so that Sentry DSN value is read from [shared_mconfig](https://github.com/magma/magma/blob/master/orc8r/protos/mconfig/mconfigs.proto#L25)
    a. [Code pointers for SessionD](https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/sessiond_main.cpp#L246-L250)

2. Modify MME so that Sentry DSN value is read from [shared_mconfig](https://github.com/magma/magma/blob/master/orc8r/protos/mconfig/mconfigs.proto#L25)
    a. [Original PR that added Sentry config parsing for MME](https://github.com/magma/magma/pull/8143)

3. Mark these fields as [deprecated](https://docs.magmacore.org/docs/next/contributing/contribute_conventions#grpc):
    - [x] SessionD: https://github.com/magma/magma/blob/master/lte/protos/mconfig/mconfigs.proto#L200
    - [x] MME: https://github.com/magma/magma/blob/master/lte/protos/mconfig/mconfigs.proto#L369

PR #11212, opened 2022-01-19
chore(sessiond): move sentry config to shared_config

Neudrino:
## Summary

Made changes suggested in https://github.com/magma/magma/issues/11092 for sessiond.
Depends on https://github.com/magma/magma/pull/10904.

## Test Plan

- build sessiond local with make and bazel
- manual change of `/etc/opt/magma/gateway.mconfig` did show up at start of sessiond

## Additional information

- A refactoring will be done in https://github.com/Neudrino/magma/pull/2. (This is a Github behaviour test and the link will be updated accordingly.)


lte/gateway/c/session_manager/CMakeLists.txt
```diff
@@ -199,10 +199,12 @@ add_library(SESSION_MANAGER
     ${PROTO_SRCS}
     ${PROTO_HDRS})
 
+find_package(MAGMA_CONFIG REQUIRED)
```
sebathomas: Not sure if this is really required, MAGMA_CONFIG was already used before the change.

sebathomas (commented):
Looking good to me, just a minor question.

lte/gateway/c/session_manager/sessiond_main.cpp
```diff
@@ -17,6 +17,7 @@
 #include <glog/logging.h>
 #include <grpcpp/impl/codegen/completion_queue.h>
 #include <lte/protos/mconfig/mconfigs.pb.h>
```
ajahl: My understanding is that `<lte/protos/mconfig/mconfigs.pb.h>`and `<orc8r/protos/mconfig/mconfigs.pb.h>` are headers of the local project. Then instead of `<...>`, you should use `"..."` here  because the preprocessor searches only along the specified path and ignores standard libraries.
(see https://docs.microsoft.com/en-us/cpp/preprocessor/hash-include-directive-c-cpp?view=msvc-170)

sebathomas (approved):

ajahl (approved):

themarwhal (approved):
Thanks! 😊

lte/gateway/c/session_manager/sessiond_main.cpp
```diff
@@ -106,6 +108,12 @@ static magma::mconfig::SessionD load_mconfig() {
   return mconfig;
 }
 
+static magma::mconfig::SharedMconfig load_shared_mconfig() {
```
themarwhal, line 111: Should we move this function into the mconfig library so that it can be shared with other services?

lte/gateway/c/session_manager/sessiond_main.cpp
```diff
@@ -227,6 +235,16 @@ long get_quota_exhaust_termination_time(const YAML::Node& config) {
   return quota_exhaust_termination_on_init_ms;
 }
 
+sentry_config_t get_sentry_configuration(
+    magma::mconfig::SharedMconfig mconfig) {
```
themarwhal, line 239: Maybe this function can be pulled into the sentry library

lte/gateway/c/session_manager/sessiond_main.cpp
Neudrino: Yes. I would refactor this within the follow up PR handling MME. I realized that MME is pure C code, which does not allow for some C++ features used here (e.g. objects). So a change to generalize will anyway be necessary for the [MME effort](https://github.com/magma/magma/pull/11257). A refactoring will be done in https://github.com/Neudrino/magma/pull/2 (I will update this link depending on the Github behaviour...).

lte/gateway/c/session_manager/sessiond_main.cpp
Neudrino: This would also be affected by the refactoring envisioned and probably necessary to support the MME case, described above. So the take here was to create a minimal viable solution and see, how we can abstract with the next one.

themarwhal (approved):

PR #11213, opened 2022-01-19
chore(sctpd): Remove signal mask for SEGV

sebathomas:
## Summary

Now that SCTPd supports Sentry, the mask should be removed in order to catch segfaults (see #9509).

## Additional Information

### DIFF
--- a/lte/gateway/c/sctpd/src/sctpd.cpp
+++ b/lte/gateway/c/sctpd/src/sctpd.cpp
@@ -41,7 +41,6 @@ using magma::sctpd::SctpdUplinkClient;
 int signalMask(void) {
   sigset_t set;
   sigemptyset(&set);
-  sigaddset(&set, SIGSEGV);
   sigaddset(&set, SIGINT);
   sigaddset(&set, SIGTERM);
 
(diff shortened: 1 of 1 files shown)


themarwhal (approved):

themarwhal:
For some context, we made the same change for MME in the past to enable Sentry. (https://fb.quip.com/McOkAbf5kgEk, https://github.com/magma/magma/pull/6606)

sebathomas:
Also added a call to `shutdown_sentry` as recommended in the [native documentation](https://sentry-docs-git-sentry-ruby-40.sentry.dev/platforms/native/).

lte/gateway/c/sctpd/src/sctpd.cpp
```diff
@@ -41,7 +41,6 @@ using magma::sctpd::SctpdUplinkClient;
 int signalMask(void) {
   sigset_t set;
   sigemptyset(&set);
-  sigaddset(&set, SIGSEGV);
```
ssanadhya, line 44: We should also remove this from line 61 below. Further, since signalMask is already called at the beginning of `main()`, we can get rid of the sigprocmask call on line 65.

ssanadhya (changes requested):

lte/gateway/c/sctpd/src/sctpd.cpp
sebathomas: Deleted them too.

ssanadhya (approved):
LGTM

PR #11216, opened 2022-01-19
chore(agw): bazel test and coverage is supported for mobilityd

### LINKED ISSUE #9881: bazel: Bazelify Python mobilityd

### LINKED ISSUE #11163: Bazel test and coverage for Python target py_test are supported conveniently
Note: this issue builds on https://github.com/magma/magma/pull/10341.

### Verb test

The `py_test` rule executes a specified source file (generally a test file) as a script, i.e., `__name__ == "__main__"`. It follows:
1. If there is no script part (esp. no if-name-main construct) then basically nothing is executed. As the file was executed successfully, Bazel shows PASSED for the test file.
2. If there is a script part, e.g., if-name-main with a call to a test framework (unittest, nose, pytest, ...) then tests are executed. I.e., each test file can use some test framework with some arguments.

**Proposal for requirements:**
* Even if a test file has no if-name-main construct, tests should be executed. Especially Bazel should not state PASSED if no existing test was executed.
* All tests should use the same test framework with the same base set of arguments. If needed, it should be possible to pass special arguments for a certain test.

**Approach:**
One way to solve this is to use a Python wrapper that executes a test file and uses a fixed test framework. An approach where a wrapper is created on-the-fly by Bazel can be seen in the Selenium repo [1]. The wrapper is created by introducing a rule `pytest_test` that wraps `py_test`. For a Python test `<TEST>.py` a wrapper `<TEST>-runner.py` is created. The `py_test` rule is now used with `<TEST>.py` and `<TEST>-runner.py` where `<TEST>-runner.py` is executed (main argument for `py_test`).

### Verb coverage

Coverage is currently not supported ootb for Python, see [2].

**Proposal for requirements:**
* Bazel coverage works the same for Python targets as for existing C/Go targets.
* Created coverage data can be used in CI - if there are divergences to the current coverage data (nosetest framework via make), then these are analyzed and fixed or founded.

**Approach:**
There is a workaround mentioned in the bazelbuild repo [3]. A fork of the coverage module with patched lcov handling is used to create coverage data with Bazel. We were not able to get this approach working. ~~The patched coverage module was not compatible to calling modules - the fork might be outdated.~~ (See more detailed analysis below in discussion) [4] seems to be the origin of this approach.

In [4] another approach [5] is suggested where a decorator is used to create the lcov coverage data. This approach can easily be combined with the test wrapper in [1]. It is important that all intermediately created coverage data is named correctly in the correct folders so that Bazel coverage works correctly. Experiments show that Bazel coverage works with this approach as expected. But note that this approach should be removed if [2] is resolved.

### Open Questions

* Should test coverage be included in CI in this stage?
	* Coverage is only generated for mobilityd.
	* Test coverage is uploaded to codecov.io - should a new endpoint be used?
* Should Python tests be cleaned-up in the sense that if-name-main constructs with a call to a test framework are removed?
	* This is currently inconsistent, e.g., cases without script part or cases with a call to the unittest framework.
	* The current workflow via make ignores the script parts. The tests are executed by nosetest test detection.
* What is the preferred Python test framework?
	* Nosetest is currently in use, pytest seems to be more widespread. Opinion of authors: pytest seems to produce more detailed output.

[1] https://github.com/SeleniumHQ/selenium/blob/1990a43a745669e87acb39331a0742979ecc728a/py/private/pytest.bzl
[2] https://github.com/bazelbuild/rules_python/issues/43
[3] https://github.com/bazelbuild/bazel/blob/0d3f9acadb94eec69f0481f61a564ce351d6e04e/site/docs/coverage.md
[4] https://github.com/bazelbuild/bazel/issues/10660
[5] https://github.com/bazelbuild/bazel/issues/10660#issuecomment-924223507

nstng:

## Summary

See https://github.com/magma/magma/issues/11163 for the general issue.
Also see https://github.com/magma/magma/issues/9881.

**Done here:**
* Provide support for bazel test and coverage for mobilityd.

**Not done here:**
* CI coverage generation (and push of coverage to codecov.io)
* Make mobilityd "sudo tests" runnable

Regarding "sudo tests":
* "sudo tests" needs to be executed by root
* sighted "sudo tests" ...
  * have hard-coded vm paths, i.e., seem to be designed for execution on the vm-setup
  * call systemctl for magma@* services
  * write outside of the repo, e.g. /etc/...
* in general these tests have more of an integration test character
* e.g., for lte defined in lte/gateway/python/defs.mk
  * all tests in pipelined/tests and two of the tests in mobilityd/tests
  * assumption: tests in mobilityd are annotated with `@unittest.skipIf(os.getuid(), reason="needs root user")` because the folder holds sudo and non-sudo tests (in comparison with pipelined/tests)
* "sudo tests" cannot be executed in bazel in their current state
  * even as root (or fakeroot [1]) bazel does not allow to write outside of bazel (e.g., /etc/...)
  * bazel does not allow access to systemctl - !needs to be verified!
  * further restriction (see above) restricts "sudo tests" to run on the vm-setup
* Handling of mobilityd "sudo tests" in this PR
  * tests are tagged as "manual" [1] -> not expanded in `bazel test //<path>:all` but can be executed directly `bazel test //<path>:my_sudo_test`
  * tests are also tagged as "sudo_test" for easy execution in CI (if possible in the future): `bazel test --test_tag_filters=sudo_test $(bazel query "attr(tags, manual, //lte/gateway/python/magma/mobilityd/tests/...)")`
  * these tests will currently fail with errors!

[1] requires-fakeroot, manual on https://docs.bazel.build/versions/main/be/common-definitions.html

## Test Plan

CI, manually run
* bazel test //lte/gateway/python/magma/mobilityd/tests:all
  * bazel-out/_coverage/_coverage_report.dat is **not** created
* bazel coverage //lte/gateway/python/magma/mobilityd/tests:all
  *  bazel-out/_coverage/_coverage_report.dat is created ...
  * ... and has meaningful data. Check by, e.g., genhtml -o genhtml bazel-out/_coverage/_coverage_report.dat

## Additional Information


bazel/external/requirements.txt
```diff
@@ -4,12 +4,80 @@
 #
 #    pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in
 #
+attrs==21.4.0 \
```
themarwhal, line 7: let's add this file here so we don't have to review it in code review: https://github.com/magma/magma/blob/master/.gitattributes

ajahl (approved):

bazel/python_utils/coverage_decorator.py
```diff
@@ -0,0 +1,91 @@
...
+from coverage_lcov import converter
+
+# environment variable COVERAGE - should be 1 iff bazel coverage was used
+COVERAGE_ENV = "COVERAGE"
+# environment variable COVERAGE_DIR - bazel sandbox folder where coverage of traget is created
+COVERAGE_DIR_ENV = "COVERAGE_DIR"
+# environment variable COVERAGE_MANIFEST - bazel instrumented files, created via --instrumentation_filter
+COVERAGE_MANIFEST_ENV = "COVERAGE_MANIFEST"
+# environment variable COVERAGE_OUTPUT_FILE - location of generated lcov file
+COVERAGE_OUTPUT_FILE_ENV = "COVERAGE_OUTPUT_FILE"
```
themarwhal, line 37: Where are these environment variables set? In what scenarios would we change these values?

lte/gateway/python/magma/mobilityd/tests/BUILD.bazel
```diff
@@ -0,0 +1,100 @@
...
+    "lte",
+    "mobilityd",
+]
+
+pytest_test(
+    name = "ip_alloc_dhcp_test",
+    srcs = ["ip_alloc_dhcp_test.py"],
+    tags = COMMON_TAGS + [
+        "manual",
+        "sudo_test",
```
themarwhal: Maybe we should have a file that defines all allowed values for flags for python unit tests? I can imagine this set of flags exploding as people get creative.

bazel/python_utils/coverage_decorator.py
nstng: @themarwhal Made this more clear in documentation. In short: environment variables are set by Bazel when 'bazel coverage' is executed. I do not see a scenario in our setup where we would change the values. They are currently read in code because this is a workaround - that should be reverted when bazel supports coverage for Python ootb.

lte/gateway/python/magma/mobilityd/tests/BUILD.bazel
nstng: @themarwhal Added bazel/testing.md with a little general test and coverage documentation and usage of tags. Is this what you meant?

bazel/testing.md
```diff
@@ -0,0 +1,74 @@
...
+)
+```
+Tags can be used to filter tests to be executed:
+`bazel test --test_tag_filters=unittest,-my_tag ...`
+Execute all tests that are tagged with "unittest" but exclude all that are tagged with "my_tag".
+
+### Proposal For Tags Used In Magma
+
+ * All tests are tagged by their respective root folder, e.g., all tests in lte have the tag "**lte**".
+ * All tests are tagged by their respective module/service, e.g., all mobilityd tests are tagged by "**mobilityd**".
```
themarwhal: Do we need these to be tags? Maybe pointing out that tests can be run like this is sufficient: `bazel test //lte/gateway/python/...:*` or `bazel test //lte/gateway/python/magma/mobilityd/...:*`

bazel/testing.md
```diff
@@ -0,0 +1,74 @@
...
+ * All unit tests are tagged with "**unittest**".
```
themarwhal: It might be easier for developers if we just default to tests being unittests unless it is marked otherwise? We might see developers forget to add this tag in the future.

bazel/testing.md
```diff
@@ -0,0 +1,74 @@
...
+ * All "sudo tests" (Python specific tests that need to be executed as root) are tagged with "**sudo_test**" and "**manual**" (see below).
```
themarwhal: I was thinking we can define all the special ones like `sudo_test` in a separate .bzl file (somewhere in the bazel/ directory) that other BUILD files can import from. We can better ensure we avoid typos and case mismatch that way. This would help us keep documentation updated if we ask people to propoerly document each new tag added in code reviews.

themarwhal (changes requested):
Overall coverage/test changes look great! Thanks for an awesome test doc :) Left some comments about tagging.

lte/gateway/python/magma/mobilityd/tests/BUILD.bazel
themarwhal: Yes the doc looks great! Thanks. I meant something like creating a .bzl file that creates all these common tags as enum/consts. That way developers can look at that file to see what tags are available.

themarwhal: cc @nstng

themarwhal:
Also, it would be great if you could make the tests runnable in the bazel-base docker container as well. I think it's currrently missing this system dependency (iproute2). I use this docker container to keep track of the minimum set of dependencies required to run tests / build. Thanks!

nstng:
> Overall coverage/test changes look great! Thanks for an awesome test doc :) Left some comments about tagging.

@themarwhal Instead of answering each tagging comment, one general answer here. I understand that you prefer a strategy that results in a small set of tags that are also handled via imports of constants. So that tags stay consistent and developers might not need to set any tag at all (e.g., for unit tests). I like that approach.

Changes are rebased.

nstng:
> Also, it would be great if you could make the tests runnable in the bazel-base docker container as well. I think it's currrently missing this system dependency (iproute2). I use this docker container to keep track of the minimum set of dependencies required to run tests / build. Thanks!

@themarwhal We were not able to reproduce a problem in the bazel-base docker image. bazel build ... and bazel test ... did complete without any failures. (The two mobilityd sudo_tests failed because they are currently only runable on the vm - but this was expected). Do you have more info?

themarwhal:
> > Also, it would be great if you could make the tests runnable in the bazel-base docker container as well. I think it's currrently missing this system dependency (iproute2). I use this docker container to keep track of the minimum set of dependencies required to run tests / build. Thanks!
> 
> 
> 
> @themarwhal We were not able to reproduce a problem in the bazel-base docker image. bazel build ... and bazel test ... did complete without any failures. (The two mobilityd sudo_tests failed because they are currently only runable on the vm - but this was expected). Do you have more info?

Ah ok! Maybe I had a bad image then. My bad. If it's tested, I'm happy! 😊

themarwhal (approved):

themarwhal (approved):

themarwhal:
🥳 
test-mobilityd-gif

koolzz (approved):

ardzoht (approved):
lgtm

PR #11217, opened 2022-01-19
feat: vscode support for python in devcontainer

alexzurbonsen:
## Summary

Add basic vscode support for python in devcontainer by adding specific configuration options to devcontainer.json

Following vscode extensions are added:
- [python extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python): provides functionality for IntelliSense (via pylance), linting, code formatting, refactoring etc.
- [pylance](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) (requires the python extension): auto import completions, code navigation, function signature help, contextual document highlighting, linting, optional type-checking and more
- [autodocstring](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring): enables docstring generation, style by default: 'google'

## Test Plan

Tested locally.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -18,6 +18,9 @@
 		"stackbuild.bazel-stack-vscode-cc",
 		"augustocdias.tasks-shell-input",
 		"ryuta46.multi-command",
+		"ms-python.python",
+		"ms-python.vscode-pylance",
+		"njpwerner.autodocstring",
 	],
 	"image": "ghcr.io/magma/magma/devcontainer:sha-206a67a",
 	"settings": {
@@ -59,7 +62,12 @@
 					"clangd.restart",
 				],
 			}
-		]
+		],
+		"python.analysis.extraPaths": [
+			"/workspaces/magma/orc8r/gateway/python/",
+			"/workspaces/magma/lte/gateway/python/",
+		],
+		"python.defaultInterpreterPath": "/usr/bin/python",
 	},
 	"mounts": [
 		"source=/var/run/docker.sock,target=/var/run/docker-host.sock,type=bind"


.devcontainer/devcontainer.json
```diff
@@ -59,8 +63,22 @@
...
+			"/workspaces/magma/lte/gateway/python/",
+		],
+		"python.defaultInterpreterPath": "/usr/bin/python",
+        "python.formatting.provider": "autopep8",
+		"python.formatting.autopep8Args": [
+			// This should be the same set of flags as ones specified in `lte/gateway/precommit.py`
+			"--select=W191,W291,W292,W293,W391,E131,E2,E3",
+		],
+        "[python]": {
+          "editor.formatOnSave": true,
```
jheidbrink: I guess we do that in this section so that non-python files are not autoformatted. Do you know why we don't do the same in the vscode config for the VM?

jheidbrink (approved):

.devcontainer/devcontainer.json
alexzurbonsen: Yes, exactly. Not sure why we don't have it in the VM. I did this because a language agnostic formatOnSave had effects that I didn't want (for example when working on the Dockerfile of the devcontainer it introduces unnecessary changes).

Neudrino (approved):

.devcontainer/devcontainer.json
```diff
@@ -18,6 +18,10 @@
 		"stackbuild.bazel-stack-vscode-cc",
 		"augustocdias.tasks-shell-input",
 		"ryuta46.multi-command",
+		"zxh404.vscode-proto3",
```
electronjoe: Could we get a quick description of each of these plugins and what they are used for / enabling / required for in the PR description? It can be hard to prune plugins down the road without some understanding.  E.g. my first question was why the proto3 for a python enablement PR (is this general proto3 support? Does an existing proto-* plugin not exist - etc).

electronjoe (changes requested):

electronjoe:
Big thank you for Python VSCode support PRs - improvements in our VSCode config are highly desirable and a great community accelerator.

.devcontainer/devcontainer.json
alexzurbonsen: Of course, added a description. And I deleted the proto extension, since it is indeed not necessary at this stage. We can pick it up at a later stage again, when introducing generation of proto python files for example.

electronjoe (approved):

PR #11219, opened 2022-01-19
fix(agw): Increase unlimited rate from 200Mbps/100Mbps for DL/UL to 4Gbps/2Gbps for DL/UL

### LINKED ISSUE #11051: Increase unlimited rate from 200Mbps/100Mbps for DL/UL to 4Gbps/2Gbps for DL/UL
Data plans set up in NMS/orc8r are used as UE AMBR rates by AGW. The unlimited definition for data plans for downlink and uplink are specified as 200 and 100Mbps. These rates are too restrictive and should be pushed up to Gbps ranges.

Impacted components: NMS, Orc8r, AGW.

ankitMeta1:
Signed-off-by: Ankit Bhadoria <abhadoria@fb.com>

fix(agw): Increase unlimited rate from 200Mbps/100Mbps for DL/UL to 4Gbps/2Gbps for DL/UL

## Summary

changed the default values for ambr DL/UL speed to 4 Gbps/2Gbps 

## Test Plan
tested the whole setup locally. if the default value for DL/UL is not defined in gateway.config, under subProfiles section for subscriberdb, then the default values in subscribedb.yml is used.

Screen Shot 2022-01-19 at 1 39 27 PM

Screen Shot 2022-01-13 at 5 27 55 PM

But once we define a dataplan and that value is propagated down to gateway, then the corresponding value for the data plan is being used:-

Screen Shot 2022-01-19 at 3 22 43 PM

Screen Shot 2022-01-19 at 3 32 09 PM

Screen Shot 2022-01-19 at 3 31 53 PM

Screen Shot 2022-01-19 at 3 01 29 PM


nms/packages/magmalte/app/components/network/DataPlanConfig.js
```diff
@@ -100,7 +100,7 @@ function DataPlanConfig(props: Props) {
         <TableCell>
           {profile.max_dl_bit_rate === DATA_PLAN_UNLIMITED_RATES.max_dl_bit_rate
             ? 'Unlimited'
-            : profile.max_dl_bit_rate / BITRATE_MULTIPLIER + ' Mbps'}
+            : (profile.max_dl_bit_rate * ONE_GB_TO_MB) / BITRATE_MULTIPLIER + ' Mbps'}
```
ulaskozat: this does not look right to me. The unit of `max_dl_bit_rate` is still BPS and this will show in Mbps, right? why not leave the BITRATE_MULTIPLIER as before then?

ulaskozat:
What is the unit expectation on entering the values for the data plan? For most deployments Mbps would still be the most used unit.

nms/packages/magmalte/app/components/network/DataPlanConst.js
```diff
@@ -15,9 +15,11 @@
  */
 
 const ONE_MEGABYTE = 1000000;
+const ONE_GIGABYTE = 1000000000;
 export const DEFAULT_DATA_PLAN_ID = 'default';
-export const BITRATE_MULTIPLIER = ONE_MEGABYTE;
+export const BITRATE_MULTIPLIER = ONE_GIGABYTE;
+export const ONE_GB_TO_MB = 1000;
```
ulaskozat: are these really needed?

nms/packages/magmalte/app/components/network/DataPlanConfig.js
ankitMeta1: we divide by the bps BITRATE_MULTIPLIER, which is a GB, but the message here prints the rate in Mbps, so i converted Gbps to Mbps.

ankitMeta1:
> What is the unit expectation on entering the values for the data plan? For most deployments Mbps would still be the most used unit.

Right now, NMS doesnt have any utility to create data plans, it is still done via swagger ui, where the data rate is defined in Bps, and not Mbps. i added the changes so that the default data rates are the same in subscriberdb, and NMS.

nms/packages/magmalte/app/components/network/DataPlanConfig.js
ankitMeta1: this is also premature, because NMS right now doesnt have a way to define or edit a data plan, but i want to keep NMS upto date with the changes in subscriberdb. so that there is no discrepancy once they implement it.

ulaskozat: This is quite error prone. max_dl_bit_rate unit is Bps and you want to show it in Mbps. It does not make sense to me to multiply and divide again for a simple unit conversion. I still do not get why the BITRATE_MULTIPLIER is modified.

ankitMeta1: Just to clarify i didnt decide to show it in Mbps. Thats how it is in the existing code(https://github.com/magma/magma/blob/90383141be1a47f56dafd58b9c5029f166d2b971/nms/packages/magmalte/app/components/network/DataPlanConfig.js#L103)

I can change the BITRATE_MULTIPLIER to its original value, if most data rates will be in Mbps, but since we dont have a UI for that in NMS irght now, i am not sure, how it looks like? in swagger API i defined the data rate in Bps, then should we have dataRate in Bps in NMS UI too? 

But, According to the current code, it looks like we are expecting the NMS UI to take the data rate in Mbps.
https://github.com/magma/magma/blob/90383141be1a47f56dafd58b9c5029f166d2b971/nms/packages/magmalte/app/components/network/DataPlanEditDialog.js#L54

andreilee (approved):

ulaskozat (approved):
LGTM

PR #11052, opened 2022-01-05
test(agw): Add non-nat datapath integ test

pshelar:
Following PR adds non-NAT datapath test, it uses mconfig
to disable NAT and runs traffic test.
gateway mconfig is sorted using python to avoid changes
to the file while toggling the NAT related flags.

Signed-off-by: Pravin B Shelar <pbshelar@fb.com>

## Additional Information

- [ ] This change is backwards-breaking

PR #11222, opened 2022-01-20
fix(agw): Added missing test case to non-sanity list

VinashakAnkitAman:
## Title
fix(agw): Added missing test case to non-sanity list

## Summary
The test case s1aptests/test_attach_detach_non_nat_dp_ul_tcp.py was missing from the defs.mk file.  The test case was added as part of PR #11052. This PR adds the test case to nonsanity and also formats the test script.

## Test plan


pshelar (approved):

ssanadhya (approved):
LGTM

PR #11226, opened 2022-01-20
fix(agw): Added flag to ignore individual CI test case failure

VinashakAnkitAman:
## Title
fix(agw): Added flag to ignore individual CI test case failure

## Summary
The CI run currently stops when any test case under lte integ test fails. This PR adds a flag to ignore the failure of individual test case, so that all the test cases can be run. Even with some of the test cases failing, we should run all the test cases at least once to understand the overall impact of recent changes. This change helps in getting more coverage with some flaky test cases as well. However this feature needs more changes in log parsing to get the complete report of passing and failing test cases.

An example of the integ test execution terminal log where 5 test cases are failing in between, but without stopping on the failure test case, next test case has been executed till all the test cases listed for sanity suite have been executed at least once to extend coverage for all the scenarios.
[integ_test_logs.txt](https://github.com/magma/magma/files/8041221/55-integ_test_logs.txt)

## Test plan
Verified with different sets of passing and failing test cases


ardzoht (approved):
lgtm

ardzoht:
@VinashakAnkitAman please rebase on top of master to fix lte test GH job.

PR #11233, opened 2022-01-20
fix(agw): Update the control plane and user plane teids to directoryd for inbound user

rsarwad:

## Summary
fix(agw): Update the correct control plane and user plane teids to directoryd for inbound user. 
## Test Plan
Executed s1ap sanity test suite and unit test case


uri200 (approved):

ardzoht (approved):
lgtm

PR #11241, opened 2022-01-20
feat: Federated Integ Test automated fab script

uri200:

## Summary

This PR adds some commands to lte/fabfile to build and run `federated_integ_test`. 

We also have added an entry to show how to build and run the test on our documentation that will be seen at `Technical Reference/FEG`

Note we currently only support running this test using AGW over Vagrant VM. In the future we may change that to support non vagrant agw.

## Test Plan
Run test locally
```
cd magma/lte/gateway
fab federated_integ_test
```

Checked new manual on docusaurus docker container
image

## Additional Information


lte/gateway/python/integ_tests/federated_tests/s1aptests/test_attach_detach.py
```diff
@@ -0,0 +1,65 @@
...
+
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
```
uri200, line 12: This script is a copy of the `test_attach_detach.py` from lte integ test.

We will modify it in the future once we configure HSS remotely (right now it is configured manually)

docs/readmes/feg/s1ap_federated_test.md
```diff
@@ -0,0 +1,165 @@
...
+- Start AGW, Orc8r and FEG
+- Configure Orc8r with AGW and FEG
+- Run connectivity tests between all 3 components
+- Run a basic attach/detach test
+
+```
+cd magma/lte/gatewat
+fab federated_integ_test
+```
+
```
uri200, line 61: after this line, the text below was previously on the README. I just moved it here

lte/gateway/fabfile.py
```diff
@@ -217,6 +219,31 @@ def s1ap_setup_cloud():
...
+    vagrant_setup(
+        'magma_trfserver', destroy_vm, force_provision=provision_vm)
+
+    vagrant_setup(
+            'magma_test', destroy_vm, force_provision=provision_vm,
+        )
+    execute(_make_integ_tests)
+    execute(run_integ_tests, "federated_tests/s1aptests/test_attach_detach.py")
+
+
```
uri200, line 250: this is just the steps to run the test. Note they are different than integ_test because they only support Vagrant AGW for now

koolzz (approved):

lte/gateway/fabfile.py
```diff
@@ -310,7 +337,8 @@ def run_integ_tests(tests=None):
     command in test machine:
     $ make integ_test TESTS=s1aptests/test_attach_detach.py
     """
-    test_host = vagrant_setup("magma_test", destroy_vm=False)
+    test_host = vagrant_setup("magma_test", destroy_vm=False,
```
ardzoht: what is the reason for this change?

uri200: yes and not: it looks like vagrant_setup was defaulting `force_provision` to string `False`., which caused to be True on the comparison

I am going to change instead force_provision at vagrant_setup to use boolean value by default. 

This is the problem with types on fab: the arguments we pass through the command line always come as strings.

ardzoht (approved):
lgtm

emakeev (approved):

electronjoe (approved):

themarwhal (approved):

PR #11254, opened 2022-01-21
feat(agw): Scaffold subscriberdb service in src go for load testing.

mstre123:

## Summary

SubscriberDB rpcs backed with a map in src/go for comparative load testing.

For demonstration purposes only.

## Test Plan

Change port in subscriberDB load test to point to port 6002 to test this impl.

## Additional Information


uri200:
Can we move this to `/magma/lte/gateway/go`?

I think the `scr` folder made sense when we were thinking on refactoring big chunks of the code. But now we are just adding services or test. I think, for our own benefit, we should maintain the (not ideal) folder structure we have in magma.

mstre123:
leaving this one as pending - template for intern project

PR #11257, opened 2022-01-24
chore(mme): move sentry config to shared_config

Neudrino:
## Summary

Made changes suggested in https://github.com/magma/magma/issues/11092 for mme.
Depends on https://github.com/magma/magma/pull/10904.

## Test Plan

- build mme local with `make build_common build_oai`
  - :heavy_check_mark: in devcontainer
  - :heavy_check_mark: in VM (CMakeList setup rather confusing :confounded: :dizzy_face: )
- CI

## Additional information

- A refactoring will be done in https://github.com/Neudrino/magma/pull/2. (This is a Github behaviour test and the link will be updated accordingly.)


orc8r/gateway/c/common/sentry/SentryWrapper.cpp
```diff
@@ -43,6 +46,16 @@
...
 
+void get_sentry_configuration(sentry_config_t* sentry_config) {
+  magma::mconfig::SharedMconfig mconfig;
+  magma::load_service_mconfig_from_file(SHARED_MCONFIG, &mconfig);
+  sentry_config->sample_rate = mconfig.sentry_config().sample_rate();
+  std::strncpy(sentry_config->url_native,
+          mconfig.sentry_config().dsn_native().c_str(), MAX_URL_LENGTH - 1);
+  sentry_config->url_native[MAX_URL_LENGTH - 1] = '\0';
+  return;
+}
```
sebathomas, line 59: :+1:

orc8r/gateway/c/common/sentry/includes/SentryWrapper.h
```diff
@@ -36,6 +36,11 @@ typedef struct sentry_config {
   bool add_debug_logging;
 } sentry_config_t;
 
+/**
+ * @brief Get the sentry configuration object
```
sebathomas:
```suggestion
 * @brief Get the sentry configuration object from the shared mconfig
```

orc8r/gateway/c/common/sentry/includes/SentryWrapper.h
```diff
@@ -36,6 +36,11 @@ typedef struct sentry_config {
   bool add_debug_logging;
 } sentry_config_t;
 
+/**
+ * @brief Get the sentry configuration object from shared mconfig
+ */
+void get_sentry_configuration(sentry_config_t* sentry_config);
```
themarwhal: `construct_sentry_config_from_mconfig` might be more descriptive. Also, why not just return the struct as a return value rather than using an output parameter? `sentry_config_t get_sentry_configuration();`

Neudrino: I have renamed the method. :heavy_check_mark:

Neudrino: Thanks for the hint regarding the struct! Your suggestion was also my first thought of implementation. However, as I am coming more from the Python kind of language, we discussed this in the team. Within this discussion I have been advised that the concept of ownership is to be valued in C (as well). We furthermore discussed how memory management is to be taken care of manually in C, and if the objects will be available in different scopes using the different implementations.

My understanding is so far, that
- both approaches would work fine in this case.
- a created entity inside the method would only live outside if it is passed by value with the return statement, which
  - is the case for structs in C by default,
  - is not generally the case for objects in C++.
- The concepts of ownership suggests to create objects where they are used to avoid memory leaks.

From what I got so far of the discussion over the different implementations, my feeling is, that they are matter of taste.
The current implementation being slightly more general (as also working for C++ objects) and more focusing on the ownership principle.
Your suggested implementation being more C specific (using peculiarities of structs) and probably some compiler magic to avoid having two copies of the entity in memory when passing by value.

Did we miss anything, which makes one or the other implementation superior?

themarwhal: Yes I think they are both acceptible. I prefer return values as it looks cleaner, but I don't have a strong opinion either way. We do point to the [google C++ guidelines](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs) for these types of things [(our contributing guidelines](https://docs.magmacore.org/docs/next/contributing/contribute_conventions#c)), so we should follow it unless we have good reasons to do otherwise.

Neudrino: Thanks, @themarwhal, very fair point! I refactored to follow the Google CPP style guidelines. :heavy_check_mark:

ajahl (approved):
lgtm

themarwhal (approved):

Neudrino:
Added another commit to [deprecate the sentry_config in mme](https://github.com/magma/magma/blob/master/lte/protos/mconfig/mconfigs.proto#L372).

sebathomas (approved):

orc8r/gateway/c/common/sentry/SentryWrapper.cpp
```diff
@@ -157,6 +172,8 @@ void sentry_log_error(const char* message) {
 
 #else
 
+sentry_config_t construct_sentry_config_from_mconfig() {}
```
themarwhal: Can we silence these warnings?

lte/gateway/c/core/oai/oai_mme/CMakeLists.txt
```diff
@@ -31,9 +31,24 @@ endif ()
...
+create_proto_dir("orc8r" ORC8R_CPP_OUT_DIR)
+
+list(APPEND PROTO_SRCS "")
+list(APPEND PROTO_HDRS "")
+
+set(SMGR_ORC8R_CPP_PROTOS mconfig/mconfigs)
+generate_cpp_protos("${SMGR_ORC8R_CPP_PROTOS}" "${PROTO_SRCS}"
+  "${PROTO_HDRS}" ${ORC8R_PROTO_DIR} ${ORC8R_CPP_OUT_DIR})
+
+find_package(MAGMA_LOGGING REQUIRED)
```
ardzoht, line 45: why is this needed?

ardzoht: ok note for me: saw that the library is included as a dependency for `MConfigLoader` functions

ardzoht (approved):
lgtm, looks like you will need to regenerate protos as insync-check job is failing

orc8r/gateway/c/common/sentry/SentryWrapper.cpp
sebathomas: I think you can return an empty config instead of disabling the linter:
```cpp
sentry_config_t construct_sentry_config_from_mconfig() {
    return {0};
}
```

PR #11258, opened 2022-01-24
fix(agw): Fixed issue for timers not restored after mme restart

### LINKED ISSUE #11237: [Sanity_Broken] Some test cases are failing on latest master codebase
### Your Environment

- **Version:** Local Development Setup
- **Affected Subcomponent:** Magma MME/SPGW service

### Setup Details
The S1APTester ([https://github.com/facebookexperimental/S1APTester](https://github.com/facebookexperimental/S1APTester)) was used along with Magma EPC in Vagrant based setup to reproduce the issue.

### Describe the Issue
The test case s1aptests/test_mobile_reachability_timer_with_mme_restart.py and s1aptests/test_implicit_detach_timer_with_mme_restart.py are failing on latest master codebase. The test case expects the service request to be rejected after MME restart, however the service request goes successful with initial context setup and the test case fails with assertion.

The terminal error log is as follows:
```
************************* Sending Service request for UE id  1
************************* send SCTP SHUTDOWN
Deleting Ue Context from Traffic Handler
Keys left in Redis (list should be empty)[

]
Entries in s1ap_imsi_map (should be zero): 1
Entries left in hashtables (should be zero): 0
Entries in mme_ueip_imsi_map (should be zero): 0
F
======================================================================
FAIL: The test case validates Mobile Reachability Timer resumes the
----------------------------------------------------------------------
  File "/home/vagrant/magma/lte/gateway/python/integ_tests/s1aptests/test_mobile_reachability_timer_with_mme_restart.py", line 131, in test_mobile_reachability_timer_with_mme_restart
    )
AssertionError: 53 != 25
-------------------- >> begin captured logging << --------------------
root: ERROR: Could not load magmad yml config for mconfig modules
urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): 192.168.60.142:3333
urllib3.connectionpool: DEBUG: http://192.168.60.142:3333 "GET /stats/switches HTTP/1.1" 200 16
urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): 192.168.60.142:3333
urllib3.connectionpool: DEBUG: http://192.168.60.142:3333 "GET /stats/switches HTTP/1.1" 200 16
root: INFO: Adding subscriber : IMSI001010000000001
root: DEBUG: s1ap response expected, received: 12, 12
root: DEBUG: s1ap message expected, received: 68, 68
root: INFO: Deleting subscriber : IMSI001010000000001
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
Ran 1 test in 734.634s

FAILED (failures=1)
Makefile:42: recipe for target 'integ_test' failed
make: *** [integ_test] Error 1
```

```
************************* Sending Service request for UE id  1
************************* send SCTP SHUTDOWN
Deleting Ue Context from Traffic Handler
Keys left in Redis (list should be empty)[

]
Entries in s1ap_imsi_map (should be zero): 1
Entries left in hashtables (should be zero): 0
Entries in mme_ueip_imsi_map (should be zero): 0
F
======================================================================
FAIL: The test case validates implicit detach timer resumes the
----------------------------------------------------------------------
  File "/home/vagrant/magma/lte/gateway/python/integ_tests/s1aptests/test_implicit_detach_timer_with_mme_restart.py", line 139, in test_implicit_detach_timer_with_mme_restart
    response.msg_type, s1ap_types.tfwCmd.UE_SERVICE_REJECT_IND.value,
AssertionError: 53 != 25
-------------------- >> begin captured logging << --------------------
root: ERROR: Could not load magmad yml config for mconfig modules
urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): 192.168.60.142:3333
urllib3.connectionpool: DEBUG: http://192.168.60.142:3333 "GET /stats/switches HTTP/1.1" 200 16
urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): 192.168.60.142:3333
urllib3.connectionpool: DEBUG: http://192.168.60.142:3333 "GET /stats/switches HTTP/1.1" 200 16
root: INFO: Adding subscriber : IMSI001010000000001
root: DEBUG: s1ap response expected, received: 12, 12
root: DEBUG: s1ap message expected, received: 68, 68
root: INFO: Deleting subscriber : IMSI001010000000001
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
Ran 1 test in 777.274s

FAILED (failures=1)
Makefile:42: recipe for target 'integ_test' failed
make: *** [integ_test] Error 1
```

As per the MME log, the mobile reachability timer is not getting restored after MME restart and hence subsequent steps of mobile reachability timer expiry, implicit detach timer expiry and implicit UE context release is not happening. Because of this, UE context is not cleared and Service request is passing, which leads to test case failure.

**To Reproduce**
1) Run the test cases using make integ_test TESTS=<testcase_name>

VinashakAnkitAman:
## Title
fix(agw): Fixed issue for timers not restored after mme restart

## Summary
Some timer related tests were failing on latest master as part of the Github issue #11237. The test cases were failing because they expect a service reject after MME restart considering that the UE context would have been deleted after Mobile reachability/Implicit detach timer expiry. However, the service request was passing and leading to assertion failure for the test cases. This PR fixes the issue by fixing following 3 linked problems with this scenario:
a) The mobile reachability and implicit detach timer values were not set after MME restart
b) The timer value was stored in milli sec and compared with lapsed time in sec
c) Implicit detach timer data was not stored in redis db before MME restart because of no UE state change

## Test plan


lte/gateway/c/core/oai/tasks/mme_app/mme_app_timer_management.cpp
```diff
@@ -52,23 +52,23 @@ void mme_app_resume_timer(
     struct ue_mm_context_s* const ue_mm_context_pP, time_t start_time,
     nas_timer_t* timer, zloop_timer_fn timer_expiry_handler, char* timer_name) {
   OAILOG_FUNC_IN(LOG_MME_APP);
-  time_t current_time = time(NULL);
-  time_t lapsed_time  = current_time - start_time;
+  time_t current_time      = time(NULL);
+  time_t lapsed_time_in_ms = (current_time - start_time) * 1000;
```
ardzoht:
```suggestion
  time_t elapsed_time_in_ms = (current_time - start_time) * 1000;
```

lte/gateway/c/core/oai/tasks/mme_app/mme_app_timer_management.cpp
```diff
@@ -52,23 +52,23 @@ void mme_app_resume_timer(
...
   OAILOG_DEBUG(LOG_MME_APP, "Handling :%s timer \n", timer_name);
 
   /* Below condition validates whether timer has expired before MME recovers
    * from restart, so MME shall handle as timer expiry
    */
-  if (timer->msec <= lapsed_time) {
+  if (timer->msec <= lapsed_time_in_ms) {
```
ardzoht:
```suggestion
  if (timer->msec <= elapsed_time_in_ms) {
```

lte/gateway/c/core/oai/tasks/mme_app/mme_app_timer_management.cpp
```diff
@@ -52,23 +52,23 @@ void mme_app_resume_timer(
...
     timer_expiry_handler(mme_app_task_zmq_ctx.event_loop, timer->id, NULL);
     OAILOG_FUNC_OUT(LOG_MME_APP);
   }
-  uint32_t remaining_time_in_msecs = timer->msec - lapsed_time;
+  uint32_t remaining_time_in_msecs = timer->msec - lapsed_time_in_ms;
```
ardzoht:
```suggestion
  uint32_t remaining_time_in_msecs = timer->msec - elapsed_time_in_ms;
```

lte/gateway/c/core/oai/tasks/mme_app/mme_app_timer_management.cpp
```diff
@@ -52,23 +52,23 @@ void mme_app_resume_timer(
...
   OAILOG_DEBUG(
       LOG_MME_APP,
       "Current_time :%ld %s timer start time :%ld "
-      "lapsed time:%ld remaining time:%d (ms) \n",
-      current_time, timer_name, start_time, lapsed_time,
+      "lapsed time:%ld (ms) remaining time:%d (ms) \n",
```
ardzoht:
```suggestion
      "elapsed time:%ld (ms) remaining time:%d (ms) \n",
```

lte/gateway/c/core/oai/tasks/mme_app/mme_app_timer_management.cpp
```diff
@@ -52,23 +52,23 @@ void mme_app_resume_timer(
...
+      current_time, timer_name, start_time, lapsed_time_in_ms,
```
ardzoht:
```suggestion
      current_time, timer_name, start_time, elapsed_time_in_ms,
```

lte/gateway/c/core/oai/tasks/mme_app/mme_app_state_converter.cpp
```diff
@@ -714,6 +714,14 @@ void MmeNasStateConverter::proto_to_ue_mm_context(
   // Initialize timers to INVALID IDs
   state_ue_mm_context->mobile_reachability_timer.id = MME_APP_TIMER_INACTIVE_ID;
   state_ue_mm_context->implicit_detach_timer.id     = MME_APP_TIMER_INACTIVE_ID;
+  state_ue_mm_context->mobile_reachability_timer.msec =
```
ardzoht, line 717: what is the reason to update these on state recovery? per my understanding these are set as inactive timers as we are not saving timers in redis. tagging @ulaskozat to keep me honest here.

ardzoht (changes requested):

lte/gateway/c/core/oai/tasks/mme_app/mme_app_state_converter.cpp
ulaskozat: For registered users, when they switch to idle mode, we should be saving the timestamp and upon restart we should update it with the residual time.

ulaskozat: so yes, we do not need to save or recover this here. Lines 734-741 are the relevant states to recover.

ulaskozat: looks benign though as it will be reset during resume_timer.

VinashakAnkitAman: Hi Alex.. mobile_reachability_timer.msec contains the maximum time for the mobile reachability timer to expire and calculated completely on hardcoded configuration values from files mme.conf.template and mme_app_ue_context.h. This is used as a pre-configured value when starting the mobile reachability timer and has no dependency on whether the timer is inactive or active at that point. If you see the below code from file "tasks/mme_app/mme_app_location.c", this value of mobile_reachability_timer.msec is being set but the mobile_reachability_timer is still inactive. 

```
  ue_mm_context->mobile_reachability_timer.id = MME_APP_TIMER_INACTIVE_ID;
  ue_mm_context->implicit_detach_timer.id     = MME_APP_TIMER_INACTIVE_ID;
#if !MME_UNIT_TEST
  ue_mm_context->mobile_reachability_timer.msec =
      ((mme_config.nas_config.t3412_min) +
       MME_APP_DELTA_T3412_REACHABILITY_TIMER) *
      60000;
  ue_mm_context->implicit_detach_timer.msec =
      (ue_mm_context->mobile_reachability_timer.msec) +
      MME_APP_DELTA_REACHABILITY_IMPLICIT_DETACH_TIMER * 60000;

```
> these are set as inactive timers as we are not saving timers in redis

Now for the above point, the timer id not stored in redis db and set to inactive when restored after mme restart is because the timer id is generated only after we start a timer. So, when MME comes up we need to restart the timer if it was running earlier and at that time the timer ID will be freshly regenerated. But this is true for timer IDs only. For timer msec value, if you see for most of the timers it is already set here in this file starting from line no. 725 to 732

Just for additional information, the timer id with active or inactive flag is not an identifier for understanding whether a timer was running or not before mme restart. The actual identifier here are timer_started variables (e.g., state_ue_mm_context->time_mobile_reachability_timer_started), which is zero until a timer is started and timer start time is capture in this variable.

VinashakAnkitAman: > For registered users, when they switch to idle mode, we should be saving the timestamp and upon restart we should update it with the residual time.

Hi Ulas.. As in MME code for now, we don't update any of these variables with the "residual time", The (timer).msec variale will always hold the maximum timer value for the timer to expire and (timer)_started variable will always hold the time when the timer was actually started and same value will be restored till the timer expires even after MME restart. What actually changes is that we create a temporary variable to store the remaining time and call the StartTimer function to start the timer with this newly created variable. Please see the code segment below to resume timer from file "tasks/mme_app/mme_app_timer_management.cpp"

```
  time_t current_time       = time(NULL);
  time_t elapsed_time_in_ms = (current_time - start_time) * 1000;
  uint32_t remaining_time_in_msecs = timer->msec - elapsed_time_in_ms;

  magma::lte::MmeUeContext::Instance().StartTimer(
           remaining_time_in_msecs, TIMER_REPEAT_ONCE, timer_expiry_handler,
           arg)) == -1) 
```

> so yes, we do not need to save or recover this here. Lines 734-741 are the relevant states to recover.

Since we use this (timer).msec value to calculate the remaining time for the timer to run, we need to recover this (timer).msec value somewhere before actually resuming the timer.. Either in this function or in the function to resume timer after verifying the timer is started or not.

> looks benign though as it will be reset during resume_timer.

It is not being reset anywhere as it was set at only one place ,i.e., after ULA in the code "tasks/mme_app/mme_app_location.c" so far. I have added it here and not in resume timer function to have this value stored even if MME was restarted when the UE was in active mode. Otherwise this would only serve if MME was restarted when UE was in IDLE mode. And yes it is of no harm to add it here as by the time we definitely have the UE context.

ulaskozat: a UE in active mode should not have this timer running.

ulaskozat: anyhow, in general it is a good idea to initialize these rather than leaving them dangling (cc @ardzoht)

VinashakAnkitAman: Yes Ulas.. This mobile reachability timer/implicit detach timer will not be running while UE is in active mode.. But (timer).msec would have been initialized for these timers before MME restart during ULA and after MME restart (timer).msec value will not be re-initialized if UE is in active mode.. So after MME restart, if UE goes to idle mode, it will try to use this value to start the timer, but it will fail since this is not set. This is why restoring this data here will serve both scenarios of MME restart whether UE is in active mode or idle mode before MME restart..

ardzoht (approved):
lgtm

ardzoht:
@VinashakAnkitAman `MmeAppProcedureTest.TestMobileReachabilityTimer` unit test failing, can you take a look?

PR #11263, opened 2022-01-24
feat(orc8r): add orc8r_version field in magmad

ankitMeta1:
## Summary

This PR is part 2 of the changes required, part 1 is adding the orc8r_version field in proto, and part 3 is comparing orc8r_version with AGW version and print a warning.

## Test Plan
checked the change locally.
Screen Shot 2022-01-24 at 1 41 30 PM

### DIFF
--- a/orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
+++ b/orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
@@ -40,30 +40,31 @@ type builderServicer struct {
 	orc8Version string
 }
 
-func NewBuilderServicer(orc8rVersion *string) builder_protos.MconfigBuilderServer {
-	if orc8rVersion == nil {
-		orc8rVersion, res := os.LookupEnv("VERSION_TAG")
-		if !res {
-			glog.Error("Couldn't get value for VERSION_TAG Env variable.")
-		}
-		return &builderServicer{orc8rVersion}
-	}
+func NewBuilderServicerWithOrc8rVersion(orc8rVersion *string) builder_protos.MconfigBuilderServer {
 	return &builderServicer{*orc8rVersion}
+}
+
+func NewBuilderServicer() builder_protos.MconfigBuilderServer {
+	orc8rVersion, ok := os.LookupEnv("VERSION_TAG")
+	if !ok {
+		glog.Error("Couldn't get value for VERSION_TAG Env variable.")
+	}
+	return NewBuilderServicerWithOrc8rVersion(&orc8rVersion)
 }
 
 func (s *builderServicer) Build(ctx context.Context, request *builder_protos.BuildRequest) (*builder_protos.BuildResponse, error) {
 	ret := &builder_protos.BuildResponse{ConfigsByKey: map[string][]byte{}}
 
-	config, err := (&baseOrchestratorBuilder{s.orc8Version}).Build(request.Network, request.Graph, request.GatewayId)
+	configs, err := (&baseOrchestratorBuilder{s.orc8Version}).Build(request.Network, request.Graph, request.GatewayId)
 	if err != nil {
 		return nil, errors.Wrapf(err, "builder error")
 	}
-	for key, configToRet := range config {
+	for key, config := range configs {
 		_, ok := ret.ConfigsByKey[key]
 		if ok {
 			return nil, fmt.Errorf("builder received duplicate config for key: %v", key)
 		}
-		ret.ConfigsByKey[key] = configToRet
+		ret.ConfigsByKey[key] = config
 	}
 
 	return ret, nil
@@ -120,7 +121,7 @@ func (b *baseOrchestratorBuilder) Build(network *storage.Network, graph *storage
 
 func getMagmadMconfig(
 	gateway *configurator.NetworkEntity, graph *configurator.EntityGraph, gatewayConfig *models.MagmadGatewayConfigs,
-	orc8Version string) (*mconfig_protos.MagmaD, error) {
+	orc8rVersion string) (*mconfig_protos.MagmaD, error) {
 	version, images, err := getPackageVersionAndImages(gateway, graph)
 	if err != nil {
 		return nil, err
@@ -136,7 +137,7 @@ func getMagmadMconfig(
 		Images:                  images,
 		DynamicServices:         gatewayConfig.DynamicServices,
 		FeatureFlags:            gatewayConfig.FeatureFlags,
-		Orc8RVersion:            orc8Version,
+		Orc8RVersion:            orc8rVersion,
 	}
 
 	return ret, nil
--- a/orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
+++ b/orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
@@ -36,37 +36,42 @@ import (
 	mconfig_protos "magma/orc8r/lib/go/protos/mconfig"
 )
 
-var localBuilders = []mconfig.Builder{
-	&baseOrchestratorBuilder{},
+type builderServicer struct {
+	orc8Version string
 }
 
-type builderServicer struct{}
-
-func NewBuilderServicer() builder_protos.MconfigBuilderServer {
-	return &builderServicer{}
+func NewBuilderServicer(orc8rVersion *string) builder_protos.MconfigBuilderServer {
+	if orc8rVersion == nil {
+		orc8rVersion, res := os.LookupEnv("VERSION_TAG")
+		if !res {
+			glog.Error("Couldn't get value for VERSION_TAG Env variable.")
+		}
+		return &builderServicer{orc8rVersion}
+	}
+	return &builderServicer{*orc8rVersion}
 }
 
 func (s *builderServicer) Build(ctx context.Context, request *builder_protos.BuildRequest) (*builder_protos.BuildResponse, error) {
 	ret := &builder_protos.BuildResponse{ConfigsByKey: map[string][]byte{}}
 
-	for _, b := range localBuilders {
-		partialConfig, err := b.Build(request.Network, request.Graph, request.GatewayId)
-		if err != nil {
-			return nil, errors.Wrapf(err, "sub-builder %+v error", b)
+	config, err := (&baseOrchestratorBuilder{s.orc8Version}).Build(request.Network, request.Graph, request.GatewayId)
+	if err != nil {
+		return nil, errors.Wrapf(err, "builder error")
+	}
+	for key, configToRet := range config {
+		_, ok := ret.ConfigsByKey[key]
+		if ok {
+			return nil, fmt.Errorf("builder received duplicate config for key: %v", key)
 		}
-		for key, config := range partialConfig {
-			_, ok := ret.ConfigsByKey[key]
-			if ok {
-				return nil, fmt.Errorf("builder received partial config for key %v from multiple sub-builders", key)
-			}
-			ret.ConfigsByKey[key] = config
-		}
+		ret.ConfigsByKey[key] = configToRet
 	}
 
 	return ret, nil
 }
 
-type baseOrchestratorBuilder struct{}
+type baseOrchestratorBuilder struct {
+	orc8Version string
+}
 
 func (b *baseOrchestratorBuilder) Build(network *storage.Network, graph *storage.EntityGraph, gatewayID string) (mconfig.ConfigsByKey, error) {
 	networkID := network.ID
@@ -92,7 +97,7 @@ func (b *baseOrchestratorBuilder) Build(network *storage.Network, graph *storage
 	vals := map[string]proto.Message{}
 	if gateway.Config != nil {
 		gatewayConfig := gateway.Config.(*models.MagmadGatewayConfigs)
-		vals["magmad"], err = getMagmadMconfig(&gateway, &nativeGraph, gatewayConfig)
+		vals["magmad"], err = getMagmadMconfig(&gateway, &nativeGraph, gatewayConfig, b.orc8Version)
 		if err != nil {
 			return nil, err
 		}
@@ -115,7 +120,7 @@ func (b *baseOrchestratorBuilder) Build(network *storage.Network, graph *storage
 
 func getMagmadMconfig(
 	gateway *configurator.NetworkEntity, graph *configurator.EntityGraph, gatewayConfig *models.MagmadGatewayConfigs,
-) (*mconfig_protos.MagmaD, error) {
+	orc8Version string) (*mconfig_protos.MagmaD, error) {
 	version, images, err := getPackageVersionAndImages(gateway, graph)
 	if err != nil {
 		return nil, err
@@ -131,13 +136,9 @@ func getMagmadMconfig(
 		Images:                  images,
 		DynamicServices:         gatewayConfig.DynamicServices,
 		FeatureFlags:            gatewayConfig.FeatureFlags,
+		Orc8RVersion:            orc8Version,
 	}
-	orc8rVersion, res := os.LookupEnv("VERSION_TAG")
-	if !res {
-		glog.Error("Couldn't get value for VERSION_TAG Env variable.")
-	} else {
-		ret.Orc8RVersion = orc8rVersion
-	}
+
 	return ret, nil
 }
 
--- a/orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
+++ b/orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
@@ -16,8 +16,10 @@ package servicers
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/go-openapi/swag"
+	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
 	"github.com/pkg/errors"
 
@@ -118,7 +120,11 @@ func getMagmadMconfig(
 	if err != nil {
 		return nil, err
 	}
-
+	orc8rVersion, res := os.LookupEnv("VERSION_TAG")
+	if !res {
+		glog.V(1).Info("Couldn't get value for VERSION_TAG Env variable.")
+		orc8rVersion = "0.0.0"
+	}
 	ret := &mconfig_protos.MagmaD{
 		LogLevel:                protos.LogLevel_INFO,
 		CheckinInterval:         int32(gatewayConfig.CheckinInterval),
@@ -129,6 +135,7 @@ func getMagmadMconfig(
 		Images:                  images,
 		DynamicServices:         gatewayConfig.DynamicServices,
 		FeatureFlags:            gatewayConfig.FeatureFlags,
+		Orc8RVersion:            orc8rVersion,
 	}
 
 	return ret, nil
--- a/orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
+++ b/orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
@@ -123,7 +123,7 @@ func getMagmadMconfig(
 	orc8rVersion, res := os.LookupEnv("VERSION_TAG")
 	if !res {
 		glog.V(1).Info("Couldn't get value for VERSION_TAG Env variable.")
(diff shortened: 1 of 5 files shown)


orc8r/protos/mconfig/mconfigs.proto
```diff
@@ -92,6 +92,9 @@ message MagmaD {
 
     // List of dynamic_services
     repeated string dynamic_services = 10;
+
+    //orc8r version
+    string orc8r_version = 11;
```
andreilee: Can you include info here in the comments about what are the valid range of values?

andreilee: Since there's special logic for returning version `0.0.0` when we can't determine the Orc8r version, that info should also be included in this proto definition.

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -118,7 +120,11 @@ func getMagmadMconfig(
 	if err != nil {
 		return nil, err
 	}
-
+	orc8rVersion, res := os.LookupEnv("VERSION_TAG")
+	if !res {
+		glog.V(1).Info("Couldn't get value for VERSION_TAG Env variable.")
```
andreilee: This should be an error log. Missing this log message could lead to subtle bugs if unaddressed, depending on how we set up the AGW behavior

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -118,7 +120,11 @@ func getMagmadMconfig(
...
+		orc8rVersion = "0.0.0"
```
andreilee: A `0.0.0` placeholder for a missing version works, but it looks close enough to being a valid version that I can see somebody being unsure if it's a placeholder or not. Switch this to a value that's completely unmistakeable as a placeholder, and specify that it's a possible value in the `mconfigs.proto` definition.

andreilee (changes requested):
About the test plan - can you verify that it's working with `VERSION_TAG` specified?

With log messages, please choose the logging level according to [this](https://sematext.com/blog/logging-levels/#toc-log-level-hierarchy-what-are-the-most-common-logging-levels-how-to-choose-them-2).

`proto` definitions should ideally be written verbosely enough that somebody making AGW changes can use them without reading Orc8r code. This is an ideal that we don't currently meet, but getting closer to that point is preferable.

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
ankitMeta1: there is a part 2 to this change, where we will change AGW to show a warning when the version differs by more than 1 minor version.  
i am not sure it warrants to be an error, if the version tag is missing, it doesnt mess with the orc8r functionality. if AGW's functionality depends on this tag, then AGW should treat the absence of orc8r_version as an error and not the orc8r

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
ankitMeta1: what value do you suggest? 0.0.-1?

ankitMeta1:
> About the test plan - can you verify that it's working with `VERSION_TAG` specified?
> 
> With log messages, please choose the logging level according to [this](https://sematext.com/blog/logging-levels/#toc-log-level-hierarchy-what-are-the-most-common-logging-levels-how-to-choose-them-2).
> 
> `proto` definitions should ideally be written verbosely enough that somebody making AGW changes can use them without reading Orc8r code. This is an ideal that we don't currently meet, but getting closer to that point is preferable.

i did post a snapshot showing the magmad section of gateway.mconfig with orc8rVersion field. does that not suffice?

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -118,7 +120,11 @@ func getMagmadMconfig(
 	if err != nil {
 		return nil, err
 	}
-
+	orc8rVersion, res := os.LookupEnv("VERSION_TAG")
```
hcgatewood: At a high level, we want our components (here, servicers), to be as stateless as possible -- where everything that breaks the statelessness is passed in as a dependency during component creation/initialization

That to say: ideally we would only have to read this env var once (during servicer creation), rather than once per request

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
hcgatewood: To me an error log means "something is wrong, but we'll keep going as best we can". So the question is: do we expect VERSION_TAG to always be present? If yes, it's an error, if no, it's an info

orc8r/protos/mconfig/mconfigs.proto
```diff
@@ -92,6 +92,11 @@ message MagmaD {
 
     // List of dynamic_services
     repeated string dynamic_services = 10;
+
+    //orc8r version
+    // version should be in X.X.X format with only non-negative numbers allowed for X.
+    // A special value of 0.0.-1 will be returned, in case oc8r version is not found.
```
hcgatewood: Imo using 0.0.-1 as a sentinel value is overkill -- let's just not set the value, which will result in an empty string being the sentinel value. This feels more ergonomic/idiomatic for golang/protos

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
```diff
@@ -483,6 +488,7 @@ func TestBaseOrchestratorMconfigBuilder_Build(t *testing.T) {
 				},
 				DynamicServices: nil,
 				FeatureFlags:    nil,
+				Orc8RVersion:    "0.0.-1",
```
hcgatewood: Looks like all these tests show the version not being set. Let's add 1 test where we set the version and observe its value in the mconfig

Note that this will potentially be easier given my comments about passing the orc8r_version in as an input rather than always reading it as an env variable

hcgatewood (changes requested):
Looking good! Couple small things

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
ankitMeta1: ideally if we want our services to be stateless, then we should store everything breaking statelessness, in a distributed store(redis?), and every instance reads what it needs from there. Because if i were to run another instance of orc8r in a different environment(on a different vm, machine etc), that environment may or may not have this env variable set.

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
ankitMeta1: thats a good question. Do we? For me error is serious business, requiring attention before something catastrophic happens, or first thing i look for when something catastrophic happens.

if we were to stop AGW from running in case the versions differ by more than 1, then it would be important enough to be present, and if its not present, then log error. right now we only print warning on AGW, and nothing else, so it doesnt look like this env variable is that important.

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
hcgatewood: It's a good point for sure. In a larger, more FB-style system it would make sense to forego env variables in favor of a remote config server. For our use-case, however, env vars are nice and simple, and we only deploy to k8s (in production), so we can set the env vars in the charts pretty easily

Anyways, the point I'm making in the prev comment is not that the *servicer* (grpc), not the service (k8s), should be stateless. As in, after you create the grpc servicer, it should operate statelessly if at all possible

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
hcgatewood: At AGW we're only going to print a warning.

At Orc8r, I think we should go ahead and assume the tag will be present. In which case, it would be an error if/when it's missing

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
```diff
@@ -32,6 +33,93 @@ import (
...
+			},
+		},
+	}
+	actual, err := buildBaseOrchestrator(&nw, &graph, "gw1")
+	assert.NoError(t, err)
+	assert.Equal(t, expected, actual)
+}
+
+func TestBaseOrchestratorMconfigBuilderWithoutOrc8rVersionEnvVariable_Build(t *testing.T) {
+
```
hcgatewood: Nit: rm extra line

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
```diff
@@ -32,6 +33,93 @@ import (
 	mconfig_protos "magma/orc8r/lib/go/protos/mconfig"
 )
 
+func testUtilForTestSharedConfig(t *testing.T, orc8rVersion string) {
```
hcgatewood: Nit here and below

- The "shared config" points more to the `shared_mconfig` rather than the magmad mconfig
- Fn names to generally describe "what" not "why", and be verbs rather than nouns

So maybe something like `testMagmadOrc8rVersion`

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
```diff
@@ -32,6 +33,93 @@ import (
...
+			CheckinTimeout:          10,
+			DynamicServices:         []string{},
+			FeatureFlags:            map[string]bool{},
+		},
+	}
+	graph := configurator.EntityGraph{
+		Entities: []configurator.NetworkEntity{gw},
+	}
+
+	expected := map[string]proto.Message{
```
hcgatewood: Since we only care about testing the Orc8rVersion in this test, we can grab just that specific mconfig and compare it, rather than specifying and comparing the entire map

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -34,37 +36,42 @@ import (
...
-	&baseOrchestratorBuilder{},
+type builderServicer struct {
+	orc8Version string
 }
 
-type builderServicer struct{}
-
-func NewBuilderServicer() builder_protos.MconfigBuilderServer {
-	return &builderServicer{}
+func NewBuilderServicer(orc8rVersion *string) builder_protos.MconfigBuilderServer {
```
hcgatewood: I would personally prefer separating this out into two separate functions, as discussed, that way 1 is "always stateless" and the other is "always stateful". In this hybrid impl, its statefulness requires understanding/thinking through the implementation

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -34,37 +36,42 @@ import (
...
+	if orc8rVersion == nil {
+		orc8rVersion, res := os.LookupEnv("VERSION_TAG")
```
hcgatewood: Nit

```suggestion
		orc8rVersion, ok := os.LookupEnv("VERSION_TAG")
```

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -34,37 +36,42 @@ import (
...
-		if err != nil {
-			return nil, errors.Wrapf(err, "sub-builder %+v error", b)
-		}
-		for key, config := range partialConfig {
-			_, ok := ret.ConfigsByKey[key]
-			if ok {
-				return nil, fmt.Errorf("builder received partial config for key %v from multiple sub-builders", key)
-			}
-			ret.ConfigsByKey[key] = config
+	config, err := (&baseOrchestratorBuilder{s.orc8Version}).Build(request.Network, request.Graph, request.GatewayId)
```
hcgatewood: From offline: removing useless aggregation in followup PR

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -34,37 +36,42 @@ import (
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "builder error")
+	}
+	for key, configToRet := range config {
```
hcgatewood: Nit: rather than "config" and "configToRet", something like "configs" and "config", respectively, is more descriptive (shows the pluralization)

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
hcgatewood: Nit: files should generally read top-down, meaning, generally, when I "jump to reference" I'm moving further down in a file. This makes fore a more uniform reading experience

Here specifically, it's offputting to see a util at the top of a test file, since casual readers care principally about the actual tests. So let's drop this guy at least after the tests he's used in

hcgatewood (approved):
Lgtm! Thanks for the change 🙂  Could small nits, then msg me and I'll merge it for ya

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
ankitMeta1: this is a util function function, for the test function testing shared config(to avoid code duplication), and not the actual test itself, so i am not sure if the above name is applicable.

ankitMeta1: it was done to avoid code duplication, which almost never is a good idea. this util function makes for a smaller code footprint, avoid any problems that occur with code duplication.

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
ankitMeta1: I personally believe that having single point of entry makes it easy to follow the code flow(which is very useful when debugging a time-sensitive issue), but i don't have very strong feelings about this one, so i can change it.

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
hcgatewood: Let me try to clarify -- 1. "shared config" makes me think of the shared_mconfig, meaning it's a name overload, meaning it's productive to choose a different name that doesn't cause the semantic overload

2. See https://github.com/magma/magma/wiki/Contributing-Code-Conventions#style for verbifying function names

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
hcgatewood: There's still 1 flow, so the conceptual complexity will be ~same. Separate is preferable if you're amenable. Here's an example of where we currently do this in the codebase: https://github.com/magma/magma/blob/13b485a9b6da5874761919c840e4c0a8be81641d/orc8r/cloud/go/obsidian/swagger/spec_servicer.go#L49-L52

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
hcgatewood: To clarify: I'm definitely supportive of having this fn exist 🙂  Just preference to move its declaration down beneath the fns that call it

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -129,6 +136,7 @@ func getMagmadMconfig(
 		Images:                  images,
 		DynamicServices:         gatewayConfig.DynamicServices,
 		FeatureFlags:            gatewayConfig.FeatureFlags,
+		Orc8RVersion:            orc8Version,
```
christinewang5:
```suggestion
		Orc8rVersion:            orc8rVersion,
```

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -113,7 +120,7 @@ func (b *baseOrchestratorBuilder) Build(network *storage.Network, graph *storage
 
 func getMagmadMconfig(
 	gateway *configurator.NetworkEntity, graph *configurator.EntityGraph, gatewayConfig *models.MagmadGatewayConfigs,
-) (*mconfig_protos.MagmaD, error) {
+	orc8Version string) (*mconfig_protos.MagmaD, error) {
```
christinewang5:
```suggestion
	orc8rVersion string) (*mconfig_protos.MagmaD, error) {
```

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -34,37 +36,42 @@ import (
...
 		}
+		ret.ConfigsByKey[key] = configToRet
 	}
 
 	return ret, nil
 }
 
-type baseOrchestratorBuilder struct{}
+type baseOrchestratorBuilder struct {
+	orc8Version string
```
christinewang5:
```suggestion
	orc8rVersion string
```

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
```diff
@@ -34,37 +36,42 @@ import (
 	mconfig_protos "magma/orc8r/lib/go/protos/mconfig"
 )
 
-var localBuilders = []mconfig.Builder{
-	&baseOrchestratorBuilder{},
+type builderServicer struct {
+	orc8Version string
```
christinewang5:
```suggestion
	orc8rVersion string
```

christinewang5 (commented):
Just fixed some typos 👍

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go
ankitMeta1: since go doesnt allow fn overloading i will have to use different names, so now some trying to debug will have two starting points, depending on whether they provide orc8r version or not. but like i said, i dont have strong feelings about it, so i changed it

orc8r/cloud/go/services/orchestrator/servicers/builder_servicer_test.go
ankitMeta1: i am not sure, if its a universal thing. coming from c/c++ background, even the compiler would expect declarations before they are used. example:- if you are using a class not declared in scope then you do something called forward declaration(https://www.geeksforgeeks.org/what-are-forward-declarations-in-c/)

Even from a casual human reading perspective, almost everyone reads code from top to down, so it makes sense to have util delcared before it is being used.

We can chat offline more, if you would like :)

hcgatewood: Correct that in C/C++ context this is expected. However in Go-land, this is not a specific requirement, so we are free to make a decision based on readability. In my experience, and with the goal of the code being approachable, a top-down approach is preferable -- i.e. provide summary first, then fill in some details, then finally provide lowest-layer details

For example, when debugging Go code, I read the golang code first, and only drop down into the generated assembly if a higher-resolution understanding is required. Similarly, when reading Go code, I want to be able to skim a summary of a file by looking just at the top part of the file, delving down further in the file if a higher-resolution understanding is required. Finally, consider e.g. http://gaia.cs.umass.edu/kurose_ross/online_lectures.htm, which breaks an entire field's intro class into progressively more-detailed descriptions, starting with the highest level and then progressively filling in the missing details

That to say -- when given the choice of whether to put (a) the finest details or (b) the summary first, I'd argue it's most natural to go with option (b). Which would then point to the idea that e.g. this private function definition should be placed after its usage, because this private function definition provides impl details rather than a summary of this file's purpose

hcgatewood: In any case, given this is a small consideration, I think default is to go with the existing style (as much as it exists), which is option (b). If you feel strongly enough to change the overarching style, we can handle it in a future PR 🙂

PR #11269, opened 2022-01-25
fix(feg): Log entries created from user input

spikey979:

fix(feg): Log entries created from user input

## Summary

   If unsanitized user input is written to a log entry, a malicious user may be able to forge new log entries. I added functions for url and response body sanitizing before they are written to a log entry.

## Test Plan

ran /magma/feg/gateway/docker/ ./build.py -c and /magma/feg/gateway/sbi/sbi_test.go
all test passed

## Additional Information

PR is for security alerts:
https://github.com/magma/magma/security/code-scanning/59?query=ref%3Arefs%2Fheads%2Fmaster
https://github.com/magma/magma/security/code-scanning/58?query=ref%3Arefs%2Fheads%2Fmaster
https://github.com/magma/magma/security/code-scanning/57?query=ref%3Arefs%2Fheads%2Fmaster

### DIFF
--- a/feg/gateway/sbi/sbi_logger.go
+++ b/feg/gateway/sbi/sbi_logger.go
@@ -15,8 +15,10 @@ package sbi
 
 import (
 	"fmt"
+	"html"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -31,7 +33,7 @@ func (logger *SbiLogger) LogRequest(method string, url *url.URL, reqBody []byte,
 	if len(reqBody) != 0 {
 		extraReqInfo = fmt.Sprintf("\nBody = %s", string(reqBody))
 	}
-	glog.V(2).Infof("Request %s %v %s\n", method, url.Path, extraReqInfo)
+	glog.V(2).Infof("Request %s %v %s\n", method, sanitizeUrlPath(url.Path), sanitizeString(extraReqInfo))
 }
 
 // LogResponse logs http response related info after receiving the response from the server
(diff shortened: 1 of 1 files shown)


themarwhal (approved):

mstre123:
notice a few prs about the log sanitization issue.
not a big fan of creating a sanitize func each time it comes up.
Can we fix this further up the chain?

lucasgonze:
The logging library we're using (github.com/golang/glog) doesn't have sanitization features built-in. There also isn't a hook to set a custom sanitization function. We'd have to create a wrapper module around glog. A simple but flawed workaround is to search the codebase for glog callers to verify that nobody is skipping the sbi_logger wrapper. 

I did a security review of our logging function and found lots of ways to improve. 

glog isn't well maintained - only 27 commits over its lifetime. The leading competitor, https://github.com/Sirupsen/logrus,  has 1,218 commits. We can switch to logrus just by changing the import statement. 

Changes to the sanitization function:
* Rather than stripping CR and LF, strip any character that returns false for the golang unicode.IsPrint() method
* Check for values longer than 1024 characters
* Check for zero-length url.Path
* Check for non UTF-8

lucasgonze:
> not a big fan of creating a sanitize func each time it comes up. Can we fix this further up the chain?

@mstre123 Are you thinking of changing the caller of LogRequest / LogResponse, or the library called by LogRequest / LogResponse?

spikey979:
@lucasgonze you're right, I will make the changes you have proposed regarding sanitization function.

spikey979:
@mstre123 would the solution be to put such a function in the orc8r/lib/go/util/util.go file, and then use it where needed?

lucasgonze:

Thoughts?

spikey979:
I solve all “Log enteries created from user input” code scanning allerts in one new PR: https://github.com/magma/magma/pull/11738

PR #11251, opened 2022-01-21
fix(pipelined): Fix eBPF init logic and arg passing

koolzz:
Signed-off-by: Nick Yurchenko <koolzz@fb.com>

## Summary
- NO_VLAN is actual string recieved from mobilityd
- IP address need to be converted to string first

## Additional Information

- [ ] This change is backwards-breaking

PR #11272, opened 2022-01-25
fix(pipelined): Revert Fix eBPF init logic and arg passing

koolzz:
Reverts magma/magma#11251


themarwhal (approved):
forcemerging as it reverts a breaking change

Issue #8748, opened 2021-08-25
Show warning when gateway version incompatible with controller


hcgatewood:
Currently we expect gateway versions to be between 0 and 1 minor version less than the Orc8r version -- so e.g. if Orc8r is 1.6.x then AGW can be 1.5.y and 1.6.y.

However, we don't provide any warning logs when this is violated

Deliverables:

1. Gateway warning: magmad's mconfig should contain Orc8r's version info, which magmad can then use to print a warning for incompatible versions

Versions requirement ref: https://magma.github.io/magma/docs/next/orc8r/deploy_intro#release-versioning

Code ref: https://github.com/magma/magma/pull/8280/files#diff-1d0f7bb5997c0e2cd1182c0fe249cf61e7b5bf0650f236e08266027be34a0559

themarwhal:
@hcgatewood This seems like an interesting task. What does the Orc8r version look like? Any pointers to what it looks like now?

hcgatewood:
@themarwhal updated description with pointer: https://github.com/magma/magma/pull/8280/files#diff-1d0f7bb5997c0e2cd1182c0fe249cf61e7b5bf0650f236e08266027be34a0559

@andreilee updated Orc8r to have access to its version info, so nominally this should be pretty achievable

PR #11274, opened 2022-01-25
feat(agw): do AGW and orc8r version comparison

ankitMeta1:
## Summary
 
Part 3 of Issue(https://github.com/magma/magma/issues/8748). 

In this change, i compare the orc8r_version received from orc8r with the AGW version, and if they differ by more than 1 minor version, I print a warn message.

## Test Plan
changed the proto definition locally, generated the code and tested.

https://gist.github.com/ankitbhadoria21/b60152df431bb3a3f10f9bd81dd5e942

### DIFF
--- a/orc8r/gateway/python/magma/magmad/config_manager.py
+++ b/orc8r/gateway/python/magma/magmad/config_manager.py
@@ -14,8 +14,7 @@ limitations under the License.
 import asyncio
 import logging
 import re
-from collections import namedtuple
-from typing import Any, List
+from typing import Any, List, NamedTuple, Optional
 
 import magma.magmad.events as magmad_events
 from magma.common.service import MagmaService
@@ -31,7 +30,7 @@ from orc8r.protos.mconfig_pb2 import GatewayConfigsDigest
 CONFIG_STREAM_NAME = 'configs'
 SHARED_MCONFIG = 'shared_mconfig'
 MAGMAD = 'magmad'
-VersionInfo = namedtuple('VersionInfo', 'agw_version orc8r_version')
+VersionInfo = NamedTuple('VersionInfo', [('agw_version', Optional[re.Match]), ('orc8r_version', Optional[re.Match])])
 
 
 class ConfigManager(StreamerClient.Callback):
--- a/orc8r/gateway/python/magma/magmad/config_manager.py
+++ b/orc8r/gateway/python/magma/magmad/config_manager.py
@@ -25,6 +25,7 @@ from magma.configuration.mconfig_managers import (
 from magma.magmad.service_manager import ServiceManager
 from orc8r.protos.mconfig import mconfigs_pb2
 from orc8r.protos.mconfig_pb2 import GatewayConfigsDigest
+import re
 
 CONFIG_STREAM_NAME = 'configs'
 SHARED_MCONFIG = 'shared_mconfig'
@@ -144,5 +145,28 @@ class ConfigManager(StreamerClient.Callback):
         for srv in self._services:
             if srv in mconfig.configs_by_key:
                 configs_by_key[srv] = mconfig.configs_by_key.get(srv)
+
+        # version should be in 1.1.1 format with only positive numbers allowed
+        VERSION_REGEX = re.compile(r"[0-9]+\.(?P<minor_version>[0-9]+)\.[0-9]+")
+
+        agw_version = self._magmad_service.version
+        unpacked_mconfig = mconfig.configs_by_key.get(MAGMAD)
+        if unpacked_mconfig.Is(mconfigs_pb2.MagmaD.DESCRIPTOR):
+            magmad_parsed = mconfigs_pb2.MagmaD()
+            unpacked_mconfig.Unpack(magmad_parsed)
+            orc8r_version = magmad_parsed.orc8r_version
+
+            agw_version_parsed = VERSION_REGEX.match(agw_version)
+            orc8r_version_parsed = VERSION_REGEX.match(orc8r_version)
+
+            if not agw_version_parsed:
+                logging.warning("Gateway version: %s not valid" % agw_version)
+            elif not orc8r_version_parsed:
+                logging.warning("orchestrator version: %s not valid" % orc8r_version)
+            else:
+                agw_minor = int(agw_version_parsed.group('minor_version'))
+                orc8r_minor = int(orc8r_version_parsed.group('minor_version'))
+                if agw_minor - orc8r_minor <= -1:
+                    logging.warning("Gateway is more than one minor version behind orc8r. Please consider updating it.")
 
         magmad_events.processed_updates(configs_by_key)
--- a/orc8r/gateway/python/magma/magmad/config_manager.py
+++ b/orc8r/gateway/python/magma/magmad/config_manager.py
@@ -78,7 +78,7 @@ class ConfigManager(StreamerClient.Callback):
 
     def _warn_if_versions_are_incompatible(self, agw_version, unpacked_mconfig):
         # version should be in X.X.X format with only non-negative numbers allowed for X
-        VERSION_REGEX = re.compile(r"[0-9]+\.(?P<minor_version>[0-9]+)\.[0-9]+")
+        version_regex = re.compile(r"[0-9]+\.(?P<minor_version>[0-9]+)\.[0-9]+")
 
         # unpack the magmad structure to get orce_version field
         if unpacked_mconfig.Is(mconfigs_pb2.MagmaD.DESCRIPTOR):
@@ -86,8 +86,8 @@ class ConfigManager(StreamerClient.Callback):
             unpacked_mconfig.Unpack(magmad_parsed)
             orc8r_version = magmad_parsed.orc8r_version
 
-            agw_version_parsed = VERSION_REGEX.match(agw_version)
-            orc8r_version_parsed = VERSION_REGEX.match(orc8r_version)
+            agw_version_parsed = version_regex.match(agw_version)
+            orc8r_version_parsed = version_regex.match(orc8r_version)
 
             # agw_version is not in expected format
             if not agw_version_parsed:
--- a/orc8r/gateway/python/magma/magmad/config_manager.py
+++ b/orc8r/gateway/python/magma/magmad/config_manager.py
@@ -14,6 +14,7 @@ limitations under the License.
 import asyncio
 import logging
 import re
+from collections import namedtuple
 from typing import Any, List
 
 import magma.magmad.events as magmad_events
@@ -30,6 +31,7 @@ from orc8r.protos.mconfig_pb2 import GatewayConfigsDigest
 CONFIG_STREAM_NAME = 'configs'
 SHARED_MCONFIG = 'shared_mconfig'
 MAGMAD = 'magmad'
+VersionInfo = namedtuple('VersionInfo', 'agw_version orc8r_version')
 
 
 class ConfigManager(StreamerClient.Callback):
@@ -95,12 +97,12 @@ class ConfigManager(StreamerClient.Callback):
             if not orc8r_version_parsed:
                 logging.warning("Orchestrator version: %s not valid" % orc8r_version)
 
-            return (agw_version_parsed, orc8r_version_parsed)
+            return VersionInfo(agw_version_parsed, orc8r_version_parsed)
         logging.error(
-            "Expecting MagmaD Structure, but received a different structure: %s." %
+            "Expecting MagmaD mconfig structure, but received a different structure: %s." %
             unpacked_mconfig.type_url,
         )
-        return (None, None)
+        return VersionInfo(None, None)
 
     def process_update(self, stream_name, updates, resync):
         """
@@ -176,7 +178,9 @@ class ConfigManager(StreamerClient.Callback):
 
             agw_version = self._magmad_service.version
             unpacked_mconfig = mconfig.configs_by_key.get(MAGMAD)
-            agw_version_parsed, orc8r_version_parsed = self._warn_if_versions_are_incompatible(agw_version, unpacked_mconfig)
+            version_info = self._warn_if_versions_are_incompatible(agw_version, unpacked_mconfig)
+            agw_version_parsed = version_info.agw_version
+            orc8r_version_parsed = version_info.orc8r_version
 
             if agw_version_parsed and orc8r_version_parsed:
                 agw_minor = int(agw_version_parsed.group('minor_version'))
--- a/orc8r/gateway/python/magma/magmad/config_manager.py
+++ b/orc8r/gateway/python/magma/magmad/config_manager.py
@@ -75,7 +75,7 @@ class ConfigManager(StreamerClient.Callback):
         )
         return mconfig_digest_proto
 
-    def _warn_if_versions_are_incompatible(self, agw_version, unpacked_mconfig):
+    def _parse_versions_and_log_warning(self, agw_version, unpacked_mconfig):
         # version should be in X.X.X format with only non-negative numbers allowed for X
         version_regex = re.compile(r"[0-9]+\.(?P<minor_version>[0-9]+)\.[0-9]+")
 
@@ -177,7 +177,7 @@ class ConfigManager(StreamerClient.Callback):
 
             agw_version = self._magmad_service.version
             unpacked_mconfig = mconfig.configs_by_key.get(MAGMAD)
-            version_info = self._warn_if_versions_are_incompatible(agw_version, unpacked_mconfig)
+            version_info = self._parse_versions_and_log_warning(agw_version, unpacked_mconfig)
             agw_version_parsed = version_info.agw_version
             orc8r_version_parsed = version_info.orc8r_version
 
--- a/orc8r/gateway/python/magma/magmad/config_manager.py
+++ b/orc8r/gateway/python/magma/magmad/config_manager.py
@@ -96,10 +96,11 @@ class ConfigManager(StreamerClient.Callback):
                 logging.warning("Orchestrator version: %s not valid" % orc8r_version)
 
             return (agw_version_parsed, orc8r_version_parsed)
-        else:
-            logging.error("Expecting MagmaD Structure, but received a different structure: %s." %
-                unpacked_mconfig.type_url)
-            return (None, None)
+        logging.error(
+            "Expecting MagmaD Structure, but received a different structure: %s." %
+            unpacked_mconfig.type_url,
+        )
+        return (None, None)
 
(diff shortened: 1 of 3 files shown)


orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -145,4 +146,29 @@ def did_mconfig_change(serv_name):
             if srv in mconfig.configs_by_key:
                 configs_by_key[srv] = mconfig.configs_by_key.get(srv)
 
+        # version should be in 1.1.1 format with only positive numbers allowed
```
themarwhal: I would pull all the versioning stuff into a separate function with some descriptive name (`_warn_if_versions_are_incompatible`? Or something nicer :) I've never been good with naming)

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -145,4 +146,29 @@ def did_mconfig_change(serv_name):
...
+        VERSION_REGEX = re.compile(r"[0-9]+\.(?P<minor_version>[0-9]+)\.[0-9]+")
+
+        agw_version = self._magmad_service.version
+        unpacked_mconfig = mconfig.configs_by_key.get(MAGMAD)
```
themarwhal: Could we maybe add a few sentences on what these are checking?  And let's restructure the if conditions so that we don't have too many nested ones.

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -145,4 +146,29 @@ def did_mconfig_change(serv_name):
...
+            orc8r_version = magmad_parsed.orc8r_version
+
+            agw_version_parsed = VERSION_REGEX.match(agw_version)
+            orc8r_version_parsed = VERSION_REGEX.match(orc8r_version)
+
+            if not agw_version_parsed:
+                logging.warning("Gateway version: %s not valid" % agw_version)
+
+            if not orc8r_version_parsed:
+                logging.warning("orchestrator version: %s not valid" % orc8r_version)
```
themarwhal: Might be good to consider the case where the AGW is not connected to an Orc8r. In that case, what would the orc8r_version_parsed value be? and could we adjust the log in that case to be extra helpful?

Also,
```suggestion
                logging.warning("Orchestrator version: %s not valid" % orc8r_version)
```

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -145,4 +146,29 @@ def did_mconfig_change(serv_name):
             if srv in mconfig.configs_by_key:
                 configs_by_key[srv] = mconfig.configs_by_key.get(srv)
 
+        # version should be in 1.1.1 format with only positive numbers allowed
+        VERSION_REGEX = re.compile(r"[0-9]+\.(?P<minor_version>[0-9]+)\.[0-9]+")
+
+        agw_version = self._magmad_service.version
+        unpacked_mconfig = mconfig.configs_by_key.get(MAGMAD)
+        if unpacked_mconfig.Is(mconfigs_pb2.MagmaD.DESCRIPTOR):
```
themarwhal: would we ever hit the corresponding else case? might be worth a log if it's unexpected

orc8r/gateway/python/magma/magmad/config_manager.py
ankitMeta1: this value is read from gateway.mconfig file, so if the AGW is not connected to AGW, it will read whatever is there in AGW gateway.mconfig(depending on which gateway.mconfig file takes precedence)

orc8r/gateway/python/magma/magmad/config_manager.py
ankitMeta1: it could since the config will be streamed by Orc8r, if it decides to stream down a structure other than MagmaD, maybe we can hit the else case, unless we have some protection built-in to the code when we process updates on magmad?

orc8r/gateway/python/magma/magmad/config_manager.py
ankitMeta1: i am not sure if i see the nested condition, all if conditions looks to be 1 level deep. Arent they?

orc8r/gateway/python/magma/magmad/config_manager.py
themarwhal: Yes I am wondering about the default value in the gateway.mconfig. (So the value in lte/gateway/configs/gateway.mconfig)  If the orc8r_version for the default mconfig is empty, then do we not log anything? (this seems fine to me). If the default version is some dummy value, then it might be confusing to for a user to see a log saying it's invalid.

ankitMeta1: there is no default value defined in proto definition, and its not a optional field, so some value has to be defined for orc8r_version, looking at documentation its most likely gonna be a empty string(https://developers.google.com/protocol-buffers/docs/proto3#default)

orc8r/gateway/python/magma/magmad/config_manager.py
themarwhal: It's not that it's overly nested but there's just a lot of conditions with some nests.  It would be helpful for future readers if there's a summary of what it does and what it's checks for. 

It would be good to pull this logic into a separate function, and a comment or two can't hurt. My preference is to always check the negative case (or the case that involves less logic) so that functions can return early, but that decision is up to you.

ankitMeta1: moved the checks to a different function with more helpful comments :)

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -73,6 +74,34 @@ def get_request_args(self, stream_name: str) -> Any:
         )
         return mconfig_digest_proto
 
+    def _warn_if_versions_are_incompatible(self, agw_version, unpacked_mconfig):
+        # version should be in X.X.X format with only non-negative numbers allowed for X
+        VERSION_REGEX = re.compile(r"[0-9]+\.(?P<minor_version>[0-9]+)\.[0-9]+")
```
ardzoht: It would be good to declare this as constant outside the function

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -73,6 +74,34 @@ def get_request_args(self, stream_name: str) -> Any:
...
+
+            # agw_version is not in expected format
+            if not agw_version_parsed:
+                logging.warning("Gateway version: %s not valid" % agw_version)
+
+            # orc8r_version is not in expected format
+            if not orc8r_version_parsed:
+                logging.warning("Orchestrator version: %s not valid" % orc8r_version)
+
+            return (agw_version_parsed, orc8r_version_parsed)
```
ardzoht: I would recommend to define a new `NamedTuple` with both properties and return that, which also helps with adding type hints

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -73,6 +74,34 @@ def get_request_args(self, stream_name: str) -> Any:
...
+        logging.error(
+            "Expecting MagmaD Structure, but received a different structure: %s." %
```
ardzoht:
```suggestion
            "Expecting MagmaD mconfig structure, but received a different structure: %s." %
```

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -73,6 +74,34 @@ def get_request_args(self, stream_name: str) -> Any:
...
+            unpacked_mconfig.type_url,
+        )
+        return (None, None)
```
ardzoht: Update this as well with the `NamedTuple` obj

ardzoht (changes requested):

orc8r/gateway/python/magma/magmad/config_manager.py
ankitMeta1: its not used outside of this function, so i dont think its worth the code pollution?

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -29,6 +31,7 @@
 CONFIG_STREAM_NAME = 'configs'
 SHARED_MCONFIG = 'shared_mconfig'
 MAGMAD = 'magmad'
+VersionInfo = namedtuple('VersionInfo', 'agw_version orc8r_version')
```
ardzoht:
```suggestion
VersionInfo = NamedTuple('VersionInfo', 
              [('agw_version', str), ('orc8r_version', str)],
)
```
prefer `typing.NamedTuple` usage

ardzoht (changes requested):

ardzoht (approved):
lgtm

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -145,4 +175,16 @@ def did_mconfig_change(serv_name):
...
+            unpacked_mconfig = mconfig.configs_by_key.get(MAGMAD)
+            version_info = self._warn_if_versions_are_incompatible(agw_version, unpacked_mconfig)
+            agw_version_parsed = version_info.agw_version
+            orc8r_version_parsed = version_info.orc8r_version
+
+            if agw_version_parsed and orc8r_version_parsed:
+                agw_minor = int(agw_version_parsed.group('minor_version'))
+                orc8r_minor = int(orc8r_version_parsed.group('minor_version'))
+                if agw_minor - orc8r_minor <= -1:
+                    logging.warning("Gateway is more than one minor version behind orc8r. Please consider updating it.")
```
themarwhal, line 188: Should we just pull these into `_warn_if_versions_are_incompatible`?  Seems cleaner that way.
Or rename `_warn_if_versions_are_incompatible` -> `_parse_version_info_from_mconfig`?

themarwhal (approved):
lgtm! apart from the comment about the logging function

themarwhal:
Looks like there are some pylint checks failing for `lte-test` btw

ankitMeta1:
> Looks like there are some pylint checks failing for `lte-test` btw

those warnings are unrelated to this PR, we can create another issue for formatting the code.

orc8r/gateway/python/magma/magmad/config_manager.py
ankitMeta1: `_warn_if_versions_are_incompatible` is quite descriptive ,but i think naming is very hard, so if it works then its fine.

themarwhal:
> > Looks like there are some pylint checks failing for `lte-test` btw
> 
> those warnings are unrelated to this PR, we can create another issue for formatting the code.

Ah I see.. I wonder how that slipped through CI. This is a required check so we'll have to fix that first.

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -73,6 +75,34 @@ def get_request_args(self, stream_name: str) -> Any:
...
+            magmad_parsed = mconfigs_pb2.MagmaD()
+            unpacked_mconfig.Unpack(magmad_parsed)
+            orc8r_version = magmad_parsed.orc8r_version
+
+            agw_version_parsed = version_regex.match(agw_version)
+            orc8r_version_parsed = version_regex.match(orc8r_version)
+
+            # agw_version is not in expected format
+            if not agw_version_parsed:
+                logging.warning("Gateway version: %s not valid" % agw_version)
```
ardzoht:
```suggestion
                logging.warning("Gateway version: %s not valid", agw_version)
```

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -73,6 +75,34 @@ def get_request_args(self, stream_name: str) -> Any:
...
+
+            # orc8r_version is not in expected format
+            if not orc8r_version_parsed:
+                logging.warning("Orchestrator version: %s not valid" % orc8r_version)
```
ardzoht:
```suggestion
                logging.warning("Orchestrator version: %s not valid", orc8r_version)
```

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -73,6 +75,34 @@ def get_request_args(self, stream_name: str) -> Any:
...
+
+            return VersionInfo(agw_version_parsed, orc8r_version_parsed)
+        logging.error(
+            "Expecting MagmaD mconfig structure, but received a different structure: %s." %
```
ardzoht:
```suggestion
            "Expecting MagmaD mconfig structure, but received a different structure: %s.",
```

ankitMeta1:
> > > Looks like there are some pylint checks failing for `lte-test` btw
> > 
> > 
> > those warnings are unrelated to this PR, we can create another issue for formatting the code.
> 
> Ah I see.. I wonder how that slipped through CI. This is a required check so we'll have to fix that first.

maybe a bootcamp task?(https://github.com/magma/magma/issues/11367) or should we tackle that first?

ardzoht (commented):
@ankitbhadoria21 @themarwhal I think the pylint tests are complaining because of the `%` usage, which the `Use lazy % formatting in logging functions` message does not indicate the actual issue, I posted some comments on how to fix those for the moment but it might be a good idea to revise this pylint warning.

orc8r/gateway/python/magma/magmad/config_manager.py
themarwhal: Yes, I think the naming is good, but it is a bit confusing if we have logic to log warnings about versions inside and outside of the function. So we should have a single function that includes all version warning logs, or split them up into smaller logics with descriptive names.

themarwhal:
> @ankitbhadoria21 @themarwhal I think the pylint tests are complaining because of the `%` usage, which the `Use lazy % formatting in logging functions` message does not indicate the actual issue, I posted some comments on how to fix those for the moment but it might be a good idea to revise this pylint warning.

Oh! that's confusing :/ Thanks Alex!

ankitMeta1:
> @ankitbhadoria21 @themarwhal I think the pylint tests are complaining because of the `%` usage, which the `Use lazy % formatting in logging functions` message does not indicate the actual issue, I posted some comments on how to fix those for the moment but it might be a good idea to revise this pylint warning.

My bad, i was looking at the this thing here https://github.com/magma/magma/runs/5011466174?check_suite_focus=true, let me fix what the other issue is.

orc8r/gateway/python/magma/magmad/config_manager.py
ankitMeta1: or we can name it more descriptively. i dont agree with breaking it further into smaller function, i think this function represents a very tightly couple functionality. it is parsing the versions it received, and logs its way through its processing steps, to make it easier for people to debug, its actual purpose is not to log but to parse, and log messages are there to assist with debugging a future problem.

themarwhal: Yep renaming sounds good!

uri200:
This PR may be causing issues. See the log of magmad 

```
Feb 04 15:03:00 tvm1-agw2 magmad[1709442]: INFO:root:Checking for upgrade...
Feb 04 15:03:00 tvm1-agw2 magmad[1709442]: INFO:root:Service is currently on package version 1.7.0-1644004493-e40a420b, ignoring upgrade to 0.3.14-123456789-deadbeef because it is either equal or behind.
Feb 04 15:03:00 tvm1-agw2 magmad[1709442]: INFO:root:Checkin Successful! Successfully sent states to the cloud!
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: INFO:root:Processing config update tvm1-agw2
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:13 tvm1-agw2 magmad[1709442]: WARNING:root:Orchestrator version:  not valid
Feb 04 15:03:41 tvm1-agw2 magmad[1709442]: INFO:root:[SyncRPC] Got heartBeat from cloud
Feb 04 15:04:01 tvm1-agw2 magmad[1709442]: INFO:root:Setting logging level to INFO
```

Note magmad seems to be restarting `Setting logging level to INFO`

uri200:
@ankitbhadoria21 would it be possible to reduce that verbose log?  

Also why does the service keeps setting the log level to info `Setting logging level to INFO`?  Was this change (maybe it was like that before, but per my understanding all services check its yml config once at start)

ankitMeta1:
This code is behaving as expected, the issue(https://github.com/magma/magma/issues/8748) wants to print a log warning if the orc8r version and agw version differs by more than 1 minor version. This warning is being printed because the changes for orc8r to send the version have not been merged yet.

uri200:
Thanks @ankitbhadoria21. This issue is fixed here https://github.com/magma/magma/pull/11448

PR #11275, opened 2022-01-25
fix(pipelined): Fix eBPF init logic and arg passing

koolzz:

## Summary
 - NO_VLAN is actual string recieved from mobilityd
 - IP address need to be converted to string first

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
+++ b/lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
@@ -24,8 +24,10 @@ from threading import Thread
 
 import netifaces
 from bcc import BPF
+from lte.protos.mobilityd_pb2 import IPAddress
 from magma.pipelined.mobilityd_client import get_mobilityd_gw_info
 from pyroute2 import IPDB, IPRoute, NetlinkError, NetNS, NSPopen
+from scapy.layers.inet6 import getmacbyip6
 from scapy.layers.l2 import getmacbyip
 
 LOG = logging.getLogger("pipelined.ebpf")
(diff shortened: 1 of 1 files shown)


lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
```diff
@@ -87,9 +92,13 @@ def __init__(self, sgi_if_name: str, s1_if_name: str, gw_ip: str, enabled=True,
         self.sgi_if_name = sgi_if_name
         self.s1_if_name = s1_if_name
         self.ul_src_mac = self._get_mac_address(sgi_if_name)
-        self.ul_gw_mac = self._get_mac_address_of_ip(gw_ip)
         self.sgi_if_index = self._get_ifindex(self.sgi_if_name)
-        self.enabled = enabled
+        self.ul_gw_mac = self._get_mac_address_of_ip(gw_ip)
+
+        if self.ul_gw_mac is None:
```
koolzz, line 100: Added this because of the `IP: 2020::1` edge case

pshelar: what do you mean?

koolzz: When running integ test there seems to be an ipv6 ip passed which produces no mac lookup

pshelar: I see, lets check for IPv4 GW address and then only initialize it the ebpf manager object.

koolzz: I can fail the check if its not ipv4 but this logic can handle IPv6 already, it just handles an edge case where the passed IP address can't be looked up and matched to a mac. Lmk if you think we should drop the ipv6 support completely for now though

pshelar (approved):
LGTM. Thanks.

PR #10628, opened 2021-12-02
chore(agw): Reduce loglevel of error-treating functions

jheidbrink:
Fixes #10408

## Summary

As part of a larger effort to look at frequent error logs observed in monitoring (see #10507), the goal of this PR is to avoid multiple different error logs as a result of a single error. The initial error should still be logged as such but the subsequent log statements are just a reaction to the error and not errors themselves.

This reduces some loglevels from Error to Info or Debug. All log statements touched here are executed while handling a RpcError ("Checkin Error") that is already logged with loglevel Error.

## Test Plan

None

## Additional Information

- [ ] This change is backwards-breaking

PR #11279, opened 2022-01-26
feat(agw): Implement opt-out for Sentry, remove other options

sebathomas:
## Summary

As an intermediate solution with maximum configurability, Sentry could be disabled or enabled per Python service and run in a special opt-in mode that only tracked a few error log messages. After enabling Sentry in the HIL, we got a better estimation of the logging volume and issues, and these options were only rarely needed.

To reduce complexity, we remove the Sentry options again and replace them by an opt-out tag that can be used to exclude error logs from Sentry monitoring.

See the discussion in #10027 for more details.

## Test Plan

Added unit tests and executed a manual end-to-end test with a Sentry test account.

### DIFF
--- a/lte/gateway/python/magma/subscriberdb/client.py
+++ b/lte/gateway/python/magma/subscriberdb/client.py
@@ -31,6 +31,7 @@ from lte.protos.subscriberdb_pb2 import (
 from magma.common.grpc_client_manager import GRPCClientManager
 from magma.common.rpc_utils import grpc_async_wrapper
 from magma.common.sdwatchdog import SDWatchdogTask
+from magma.common.sentry import EXCLUDE_FROM_ERROR_MONITORING
 from magma.common.service_registry import ServiceRegistry
 from magma.subscriberdb.metrics import (
     SUBSCRIBER_SYNC_FAILURE_TOTAL,
@@ -439,11 +440,18 @@ class SubscriberDBCloudClient(SDWatchdogTask):
 
 
 def _log_grpc_error(err: grpc.RpcError) -> None:
-    if err.code() == grpc.StatusCode.INTERNAL or err.code() == grpc.StatusCode.UNIMPLEMENTED:
+    if err.code() in {grpc.StatusCode.INTERNAL, grpc.StatusCode.UNIMPLEMENTED}:
         logging.error(
             'Internal Error in subscriberdb-Orc8r communication. '
             'If AGW and Orc8r are deployed at a commit hash, '
             'upgrade AGW and Orc8r to latest commit hash.',
+            extra=EXCLUDE_FROM_ERROR_MONITORING,
+        )
+    elif err.code() in {grpc.StatusCode.UNAVAILABLE, grpc.StatusCode.DEADLINE_EXCEEDED}:
+        logging.error(
+            'Internal Error in subscriberdb-Orc8r communication. '
+            'Cannot connect to server: [%s] %s', err.code(), err.details(),
+            extra=EXCLUDE_FROM_ERROR_MONITORING,
         )
     else:
         logging.error(
--- a/orc8r/gateway/python/magma/common/sentry.py
+++ b/orc8r/gateway/python/magma/common/sentry.py
@@ -14,8 +14,6 @@ import logging
 import os
 import re
 from dataclasses import dataclass
-from enum import Enum
-from functools import wraps
 from typing import Any, Callable, Dict, List, Optional
 
 import sentry_sdk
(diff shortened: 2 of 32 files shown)


lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -444,6 +445,13 @@ def _log_grpc_error(err: grpc.RpcError) -> None:
             'Internal Error in subscriberdb-Orc8r communication. '
             'If AGW and Orc8r are deployed at a commit hash, '
             'upgrade AGW and Orc8r to latest commit hash.',
+            extra=EXCLUDE_FROM_ERROR_MONITORING,
+        )
+    elif err.code() == grpc.StatusCode.UNAVAILABLE or err.code() == grpc.StatusCode.DEADLINE_EXCEEDED:
+        logging.error(
+            'Internal Error in subscriberdb-Orc8r communication. '
+            'Cannot connect to server: [%s] %s', err.code(), err.details(),
+            extra=EXCLUDE_FROM_ERROR_MONITORING,
```
sebathomas, line 454: Only two services were using the opt-in mode in the lab: magmad and subscriberdb.
* For subscriberdb, the frequent error are excluded here.
* For magmad, the frequent errors were already handled by #10628 so opt-in mode should no longer be necessary.

lte/gateway/python/magma/subscriberdb/client.py
```diff
@@ -439,11 +440,18 @@ def _detach_deleted_subscribers_done(self, delete_future):
...
             'Internal Error in subscriberdb-Orc8r communication. '
             'If AGW and Orc8r are deployed at a commit hash, '
             'upgrade AGW and Orc8r to latest commit hash.',
+            extra=EXCLUDE_FROM_ERROR_MONITORING,
+        )
+    elif err.code() in {grpc.StatusCode.UNAVAILABLE, grpc.StatusCode.DEADLINE_EXCEEDED}:
+        logging.error(
+            'Internal Error in subscriberdb-Orc8r communication. '
+            'Cannot connect to server: [%s] %s', err.code(), err.details(),
+            extra=EXCLUDE_FROM_ERROR_MONITORING,
```
sebathomas, line 454: Only two services were using the opt-in mode in the lab: magmad and subscriberdb.
* For subscriberdb, the frequent errors are excluded here (see #10387 for more discussion).
* For magmad, the frequent errors were already handled by #10628 so opt-in mode should no longer be necessary.

orc8r/gateway/python/magma/common/sentry.py
```diff
@@ -164,38 +119,22 @@ def _filter_excluded_messages(event: Event, hint: Hint, patterns_to_exclude: Lis
...
-            event: Event, hint: Hint,
-    ) -> Optional[Event]:
-        return _filter_excluded_messages(event, hint, patterns_to_exclude)
-
-    if sentry_status == SentryStatus.SEND_SELECTED_ERRORS:
-        return filter_excluded_and_unmarked_messages
-    if patterns_to_exclude:
-        return filter_excluded_messages
-    return None
+    return filter_excluded_and_unmarked_messages
```
Neudrino: Should this now be `filter_excluded_and_marked_messages`, right? I.e. without the **`un`**. Also in line 127.

orc8r/gateway/python/magma/common/sentry.py
```diff
@@ -164,38 +119,22 @@ def _filter_excluded_messages(event: Event, hint: Hint, patterns_to_exclude: Lis
     return event
 
 
-def _get_before_send_hook(
-        sentry_status: SentryStatus,
-        patterns_to_exclude: List[str],
-) -> Optional[SentryHook]:
+def _get_before_send_hook(patterns_to_exclude: List[str]) -> Optional[SentryHook]:
 
     def filter_excluded_and_unmarked_messages(
```
alexzurbonsen, line 124: This should be called '..._and_marked_messages' now, I suppose?

Neudrino (approved):
:broom: :rocket: Thanks, looks like a big cleanup! :100:

alexzurbonsen (approved):

orc8r/gateway/python/magma/common/sentry.py
sebathomas: Good catch, fixed.

lte/gateway/python/magma/subscriberdb/client.py
themarwhal: Thanks for the context!

themarwhal (approved):
Thanks for the cleanup (and a detailed explanation on why). It would be good to document in a sentry monitoring doc somewhere on this mechanism. (Maybe it exists already, please remind me :) ).  If we could clearly outline a step-by-step guide on going from detecting a noisy/unhelpful error in Sentry from opting it out / modifying the error message, that would be useful for future contributors.

Neudrino:
We are currently building up (= WIP :wink:) the [documentation around Sentry in the Github wiki](https://github.com/magma/magma/wiki/Error-monitoring-and-triaging-workflow). I suggest we add it there as well.

pshelar (approved):

jheidbrink (approved):

PR #11285, opened 2022-01-26
fix(orc8r): Format policy action & effect enums

christinewang5:
The policy protos do not decode nicely, specifically the action and effect enum fields. Added a format function so that they form helpful messages

# Before
horrible_enums

# After
Screen Shot 2022-01-26 at 1 13 25 PM

### DIFF
--- a/orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
+++ b/orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
@@ -17,11 +17,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"strconv"
 
 	"github.com/go-openapi/strfmt"
 	"github.com/labstack/echo"
-	"github.com/pkg/errors"
 
 	"magma/orc8r/cloud/go/obsidian"
 	"magma/orc8r/cloud/go/services/certifier"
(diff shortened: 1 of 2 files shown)


orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
```diff
@@ -177,7 +177,41 @@ func loginHandler(c echo.Context) error {
 	if err != nil {
 		return obsidian.MakeHTTPError(err, http.StatusInternalServerError)
 	}
-	return c.JSON(http.StatusOK, res.PolicyLists)
+	return c.JSON(http.StatusOK, policyListProtoToModel(res.PolicyLists))
+}
+
+func policyListProtoToModel(policyLists []*protos.PolicyList) interface{} {
```
hcgatewood: 1. Break this fn down into (a) the part that loops over each policy, and (b) a call into a sub-function that convernts a single policy proto to a single policy swagger object
2. Make the return type the swagger type, rather than an iface

hcgatewood (approved):
Small change, lgtm though

orc8r/cloud/go/services/certifier/obsidian/handlers/handlers.go
christinewang5: Ok done, also just moved all these protos to model fns to a helper file, thought it was cleaner

hcgatewood (commented):
The one downside to this helpers mechanism is now the swagger models pkg can't import from the protos pkg. Not end of the world though, think it's still cleaner this way. Just something to consider. Ideally we would decide at a high level whether, and in which direction, protos/models can import from each other

PR #11287, opened 2022-01-26
fix(agw): refactor QoSManager

### LINKED ISSUE #11242: pipelined: qos: get rid of get_qos_manager() from qos manager

ankitMeta1:
## Summary

Addressing issue(https://github.com/magma/magma/issues/11242) to refactor QosManager, to create a single instance in service_manager, clean the static function and variables from the class, and pass the QosManager instance  as context to apps.

## Test Plan
ran unit test locally(make test_python)
ran make test_python_service too, but its taking longer on local machine, hoping github cli would be faster.

### DIFF
--- a/lte/gateway/python/magma/pipelined/qos/common.py
+++ b/lte/gateway/python/magma/pipelined/qos/common.py
@@ -236,6 +236,9 @@ class QosManager:
         return True
 
     def __init__(self, loop, config):
+        if 'qos' not in config.keys():
+            LOG.error("qos field not provided in config")
+            return
         self._qos_enabled = config["qos"]["enable"]
         if not self._qos_enabled:
             return
--- a/lte/gateway/python/magma/pipelined/qos/common.py
+++ b/lte/gateway/python/magma/pipelined/qos/common.py
@@ -169,33 +169,28 @@ class SubscriberState(object):
         return 0
 
 
-class QosManager(object):
-    # TODO: convert QosManager to singleton class.
-    qos_mgr = None
+class QosManager:
     # protect QoS object create and delete across all QoSManager Objects.
     lock = threading.Lock()
 
-    @staticmethod
-    def get_qos_manager(datapath, loop, config):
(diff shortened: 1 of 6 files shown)


koolzz (commented):
not a big fan of initializing the same object 3 times, 

>>> ran make test_python_service too, but its taking longer on local machine, hoping github cli would be faster.

just as a heads up, ci doens't run the whole pipelined unit test suite so need to run that locally unfortunately

ankitMeta1:
> not a big fan of initializing the same object 3 times,
> 
> > > > ran make test_python_service too, but its taking longer on local machine, hoping github cli would be faster.
> 
> just as a heads up, ci doens't run the whole pipelined unit test suite so need to run that locally unfortunately

not sure, if i understand the comment, which object are we initializing 3 times?

koolzz:
> > not a big fan of initializing the same object 3 times,
> > > > > ran make test_python_service too, but its taking longer on local machine, hoping github cli would be faster.
> > 
> > 
> > just as a heads up, ci doens't run the whole pipelined unit test suite so need to run that locally unfortunately
> 
> not sure, if i understand the comment, which object are we initializing 3 times?

ah 2 times I suppose its not in enf_stats. `self._qos_mgr.init_impl(datapath)` but thats probably not worth a larger refactor

koolzz (approved):

lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -489,7 +494,7 @@ def add_subscriber_qos(
 
     def remove_subscriber_qos(self, imsi: str = "", del_rule_num: int = -1):
         with QosManager.lock:
-            if not self._qos_enabled or not self._initialized:
+            if not self._qos_enabled or not self._initialized or not self._is_impl_initialized():
```
pshelar, line 497: @ankitbhadoria21, we dont need separate _is_impl_initialized(), can you post another PR to use _initialized to keep track of the initialization.

Issue #6368, opened 2021-04-23
AGW: python: get rid of aioeventlet lib dependency


pshelar:
aioeventlet is not well maintained lib. It does not support python 3.8. we have to apply unofficial patch on master branch to make it work on python 3.8. so it would be nice to get rid of this dependency.
This task involves:
1. investigate aioeventlet use case in magma services
2. explore alternatives
3. write a proposal or start discussion for better alternative
4. post PRs to remove this dependency.

Neudrino:
Further issues with the package are:
- There is only one version from 2017 available on pypi. https://pypi.org/project/aioeventlet/
- The available version has a metadata mismatch, which makes new pip versions complain and refuse to install.
- There is no more documentation online under [the official link](https://aioeventlet.readthedocs.io/), but only a 404.
- The [original code on Bitbucket](https://bitbucket.org/haypo/aioeventlet) is not available any more.

As far as I can see, the package is only used in the following two files:
- https://github.com/magma/magma/blob/master/lte/gateway/python/magma/pipelined/main.py#L22
- https://github.com/magma/magma/blob/master/lte/gateway/python/magma/pipelined/service_manager.py#L42

The usage is limited to:
- https://github.com/magma/magma/blob/master/lte/gateway/python/magma/pipelined/main.py#L59
- https://github.com/magma/magma/blob/master/lte/gateway/python/magma/pipelined/service_manager.py#L653-L672

There is [a comment](https://github.com/magma/magma/blob/master/lte/gateway/python/magma/pipelined/main.py#L57-L58) connected with the purpose of `aioeventlet`.

Neudrino:
For reference:
There seems to be a custom build for `aioeventlet` from a fork of the original repository done in https://github.com/magma/magma/blob/master/third_party/build/bin/aioeventlet_build.sh. The repository taken is https://github.com/openstack-archive/deb-python-aioeventlet.

Additionally there are patches to `aioeventlet` introduced in https://github.com/magma/magma/pull/6404 to fix earlier issues.
The patches are supposed to be sym-linked from https://github.com/magma/magma/tree/master/third_party/build/patches/aioeventlet to https://github.com/magma/magma/tree/master/lte/gateway/deploy/roles/magma/files/patches. However, this seems not to work correctly on my machine (Arch).

Neudrino:
## The `aioeventlet` problem.

[...]

PR #11292, opened 2022-01-27
chore: install lte and orc8r python packages in devcontainer

alexzurbonsen:
## Summary

Installs lte and orc8r python package in DevContainer analogous to the make file's `install_egg` target.

The aioeventlet package has to be installed separately because `pip install` fails due to inconsistent version meta data.
The subsequent patch of the package is necessary since the version downloaded from jfrog only contains one of two patches.

`PYTHONPATH` can be deleted, since it is integrated in bazel build now (see https://github.com/magma/magma/pull/10341/commits/bb2244a642ab34263ab3bcfdad353bf90469c730).

## Test Plan

Built Dockerfile locally and check that imports of python external dependencies work.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,3 @@
-# Note: You can use any Debian/Ubuntu based image you want. 
 FROM ubuntu:focal as magma-mme-builder
 
 # [Option] Install zsh
@@ -23,8 +22,8 @@ RUN apt-get update \
     # Clean up
     && apt-get autoremove -y && apt-get clean -y && rm -rf /var/lib/apt/lists/* /tmp/library-scripts/
 
-# Setting the ENTRYPOINT to docker-init.sh will configure non-root access to 
-# the Docker socket if "overrideCommand": false is set in devcontainer.json. 
+# Setting the ENTRYPOINT to docker-init.sh will configure non-root access to
+# the Docker socket if "overrideCommand": false is set in devcontainer.json.
 # The script will also execute CMD if you need to alter startup behaviors.
 ENTRYPOINT [ "/usr/local/share/docker-init.sh" ]
 CMD [ "sleep", "infinity" ]
@@ -35,7 +34,7 @@ ENV MAGMA_ROOT=/workspaces/magma
 ENV BUILD_TYPE=Debug
 ENV C_BUILD=/workspaces/magma/build/c
 ENV DEBIAN_FRONTEND=noninteractive
-ENV PYTHONPATH=$MAGMA_ROOT/lte/gateway/python/:$MAGMA_ROOT/orc8r/gateway/python/
+
 
 RUN echo "Install general purpose packages" && \
     apt-get update && \
@@ -72,6 +71,7 @@ RUN echo "Install general purpose packages" && \
         perl \
         pkg-config \
         python3 \
+        python3-pip \
         python-is-python3 \
         redis-server \
         ruby \
@@ -88,6 +88,7 @@ RUN echo "Install general purpose packages" && \
         update-alternatives --install /usr/bin/clang-format clang-format /usr/lib/llvm-7/bin/clang-format 10 && \
         update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/lib/llvm-11/bin/clang-tidy 10 && \
         update-alternatives --install /usr/bin/clang-apply-replacements clang-apply-replacements /usr/lib/llvm-11/bin/clang-apply-replacements 10
+
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get install -y clangd-12
 
@@ -253,7 +254,7 @@ RUN git clone https://github.com/facebook/folly && cd folly && \
     make install && \
     cd / && \
     rm -rf folly
-    
+
 ##### Build and install libgtest and gmock
 RUN cd /usr/src/googletest && \
     mkdir build && \
@@ -296,8 +297,8 @@ RUN git clone https://github.com/include-what-you-use/include-what-you-use && \
 #### libfluid is requird to build MME with OVS support
 COPY third_party /tmp/third_party/
 RUN /tmp/third_party/build/bin/libfluid_build.sh && \
-     find . -name magma-libfluid_0.1* -exec dpkg -i {} \; && \
-     rm -rf /tmp/*
+    find . -name magma-libfluid_0.1* -exec dpkg -i {} \; && \
+    rm -rf /tmp/*
 
 ##### FreeDiameter
 COPY lte/gateway/c/core/oai/patches/ /tmp/
@@ -309,14 +310,43 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git 
     cmake -DDISABLE_SCTP:BOOL=ON ../ && \
     make -j"$(nproc)" && \
     make install && \
-    cd / && \
-    rm -rf freediameter
+    rm -rf /freediameter /tmp/*
 
 ##### Go language server support for vscode
 RUN go get -v golang.org/x/tools/gopls
 
 #### Update shared library configuration
 RUN ldconfig -v
+
+
+##### Install Python requirements
+RUN pip3 install --quiet --upgrade --no-cache-dir "setuptools==49.6.0"
+
+### install python3-aioeventlet from the magma apt repo (from .circleci/config.yml)
+COPY orc8r/tools/ansible/roles/pkgrepo/files/jfrog.pub /tmp/
+RUN apt-key add /tmp/jfrog.pub && \
+    echo "deb https://facebookconnectivity.jfrog.io/artifactory/list/dev-focal/ focal main" > /etc/apt/sources.list.d/fbc.list && \
+    apt-get update -y && \
+    apt-get install -y python3-aioeventlet && \
+    rm -f /tmp/* && \
+    rm -rf /var/lib/apt/lists/*  # delete lists that are downloaded by apt-get update
+
+# add patch that is missing in jfrog version of aioeventlet (it only comes with 1 of 2 patches)
+COPY lte/gateway/deploy/roles/magma/files/patches/aioeventlet_fd_exception.patch /tmp/
+RUN patch -N -s -f /usr/local/lib/python3.8/dist-packages/aioeventlet.py < /tmp/aioeventlet_fd_exception.patch && \
+    rm -rf /tmp/*
+
+### install eggs (lte, orc8r)
+COPY /lte/gateway/python/ ${MAGMA_ROOT}/lte/gateway/python/
+WORKDIR ${MAGMA_ROOT}/lte/gateway/python/
+RUN pip3 install --quiet --upgrade --no-build-isolation --no-cache-dir --verbose --editable .[dev] && \
+    rm -rf lte.egg-info
+
+COPY /orc8r/gateway/python/ ${MAGMA_ROOT}/orc8r/gateway/python/
+WORKDIR ${MAGMA_ROOT}/orc8r/gateway/python/
+RUN pip3 install --quiet --upgrade --no-build-isolation --no-cache-dir --verbose --editable .[dev] && \
+    rm -rf orc8r.egg-info
+
 
 RUN ln -s $MAGMA_ROOT/bazel/bazelrcs/devcontainer.bazelrc /etc/bazelrc
 


.devcontainer/Dockerfile
```diff
@@ -310,14 +312,47 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git
...
 #### Update shared library configuration
 RUN ldconfig -v
 
+
+##### Install Python requirements
+RUN pip3 install --quiet --upgrade --no-cache-dir "setuptools==49.6.0"
+
+### install python3-aioeventlet from the magma apt repo (from .circleci/config.yml)
+COPY orc8r/tools/ansible/roles/pkgrepo/files/jfrog.pub ${MAGMA_ROOT}/orc8r/tools/ansible/roles/pkgrepo/files/jfrog.pub
+RUN cat ${MAGMA_ROOT}/orc8r/tools/ansible/roles/pkgrepo/files/jfrog.pub | sudo apt-key add - \
```
jheidbrink: This is probably run as root, so we don't need sudo.

.devcontainer/Dockerfile
```diff
@@ -310,14 +312,47 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git
...
+    && echo "deb https://facebookconnectivity.jfrog.io/artifactory/list/dev-focal/ focal main" | sudo tee /etc/apt/sources.list.d/fbc.list \
+    && sudo apt-get update -y  \
+    && sudo apt-get install -y python3-aioeventlet \
+    && sudo rm -rf /var/lib/apt/lists/*  # delete lists that are downloaded by apt-get update
+
+ARG PATCH_DIR=lte/gateway/deploy/roles/magma/files/patches
```
jheidbrink: As this is only used once, we could just inline it. Same with DIST_PACKAGES_DIR.

.devcontainer/Dockerfile
```diff
@@ -310,14 +312,47 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git
...
+ARG DIST_PACKAGES_DIR=/usr/local/lib/python3.8/dist-packages
+
+# add missing patch to aioeventlet
```
jheidbrink: We initially thought that python3-aioeventlet from the jfrog artifactory would already contain both patches, but then found that it only contained the first patch. We could mention that in a comment here, otherwise people might wonder why we still have to apply a patch here.

jheidbrink (approved):

.devcontainer/Dockerfile
alexzurbonsen: Yes, true. (I already tried to include the remaining patches and didn't inline the variables again...)

.devcontainer/Dockerfile
alexzurbonsen: Thanks. There is also a note in the PR description.

.devcontainer/Dockerfile
```diff
@@ -310,14 +312,44 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git
...
+RUN cat ${MAGMA_ROOT}/orc8r/tools/ansible/roles/pkgrepo/files/jfrog.pub | apt-key add - \
+    && echo "deb https://facebookconnectivity.jfrog.io/artifactory/list/dev-focal/ focal main" | tee /etc/apt/sources.list.d/fbc.list \
+    && apt-get update -y  \
+    && apt-get install -y python3-aioeventlet \
+    && rm -rf /var/lib/apt/lists/*  # delete lists that are downloaded by apt-get update
+
+# add patch that is missing in jfrog version of aioeventlet (it only comes with 1 of 2 patches)
+COPY lte/gateway/deploy/roles/magma/files/patches/aioeventlet_fd_exception.patch /tmp/
+RUN patch -N -s -f /usr/local/lib/python3.8/dist-packages/aioeventlet.py < /tmp/aioeventlet_fd_exception.patch \
+    && rm -rf /tmp/*
```
alexzurbonsen: There are two patches for aioeventlet in `lte/gateway/deploy/roles/magma/files/patches`. The jfrog aioeventlet package only contains one of them. Here we add the second one.

.devcontainer/Dockerfile
```diff
@@ -310,14 +312,45 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git
     make -j"$(nproc)" && \
     make install && \
     cd / && \
-    rm -rf freediameter
+    rm -rf freediameter \
+    rm -rf /tmp/*
```
electronjoe: Thank you for improving upon this!

.devcontainer/Dockerfile
themarwhal: Should we update the one in jfrog to have both of the patches? Is that easy to do? (cc @tmdzk / @quentinDERORY )

Neudrino: It is rather hard to find, how the custom deb-packages for aioeventlet were provided. If this was done with [the buildscript in the third party folder](https://github.com/magma/magma/blob/0fb795e1fdd2d01e8dff9dcb9b2123208c0874d2/third_party/build/bin/aioeventlet_build.sh#L44) I would have expected that all patches from [the respective folder](https://github.com/magma/magma/tree/master/third_party/build/patches/aioeventlet) were applied.
(Side comment: Afaiks devops privileges are required to upload to https://facebookconnectivity.jfrog.io anyway.)

In any way, I agree that aioeventlet is a big pain and the efforts to improve should be taken some-when. Those efforts are tracked in #6368. While removal of aioeventlet might be desirable, it is not easy. However, also meaningful improvements to the current situation would be bigger topics (https://github.com/magma/magma/issues/6368#issuecomment-1025645600), which I would see clearly out of scope of this PR and effort.

Our goal here was to quickly improve the devcontainer support for Python with low effort. By staying rather close to the current Makefile setup we try bring missing functionality to the devcontainer, following the approach "First make it work, than make it pretty."

.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Note: You can use any Debian/Ubuntu based image you want. 
+# Note: You can use any Debian/Ubuntu based image you want.
```
Neudrino: While this statement was here before, I doubt, it was tested with Trusty Tahr (or any other old image).
So while only a whitespace change was made, I would see justification to completely delete this line.

.devcontainer/Dockerfile
```diff
@@ -310,14 +312,45 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git
     make -j"$(nproc)" && \
     make install && \
     cd / && \
-    rm -rf freediameter
+    rm -rf freediameter \
```
Neudrino: Logial operation missing: `;`, `&&` , `||`
Suggesting unconditional deletion.

.devcontainer/Dockerfile
```diff
@@ -310,14 +312,45 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git
     make -j"$(nproc)" && \
     make install && \
     cd / && \
```
Neudrino: Why change the folder?
This can be refactored, if absolute path are used in deletion. This might even be considered safer.

.devcontainer/Dockerfile
```diff
@@ -310,14 +312,45 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git
...
+RUN pip3 install --quiet --upgrade --no-cache-dir "setuptools==49.6.0"
+
+### install python3-aioeventlet from the magma apt repo (from .circleci/config.yml)
+COPY orc8r/tools/ansible/roles/pkgrepo/files/jfrog.pub /tmp/
+RUN apt-key add /tmp/jfrog.pub \
+    && echo "deb https://facebookconnectivity.jfrog.io/artifactory/list/dev-focal/ focal main" > /etc/apt/sources.list.d/fbc.list \
+    && apt-get update -y  \
+    && apt-get install -y python3-aioeventlet \
+    && rm -f /tmp/* \
+    && rm -rf /var/lib/apt/lists/*  # delete lists that are downloaded by apt-get update
```
Neudrino: Inconsistency regarding style: `&&` should be either pre- or post-pended everywhere.
I prefer the prepended form, due to better visibility of the operation and less chance to miss it (see above).
However, check with the project guidelines please.

Neudrino (changes requested):
Fix 1 bug please. I value consistency.

.devcontainer/Dockerfile
```diff
@@ -35,7 +35,7 @@ ENV MAGMA_ROOT=/workspaces/magma
 ENV BUILD_TYPE=Debug
 ENV C_BUILD=/workspaces/magma/build/c
 ENV DEBIAN_FRONTEND=noninteractive
-ENV PYTHONPATH=$MAGMA_ROOT/lte/gateway/python/:$MAGMA_ROOT/orc8r/gateway/python/
```
Neudrino, line 38: Could you also comment, why this is no longer necessary, please. A note to the PR description would be valuable.

.devcontainer/Dockerfile
```diff
@@ -35,7 +34,7 @@ ENV MAGMA_ROOT=/workspaces/magma
...
```
alexzurbonsen, line 38: Deleted, because it is handled in bazel build directly now (see https://github.com/magma/magma/commit/bb2244a642ab34263ab3bcfdad353bf90469c730)

.devcontainer/Dockerfile
themarwhal: Thanks for looking into this! Yes, I agree with make it work then make it pretty, as long as we are filing issues to track the potential clean up opportunities (which you are doing, thanks!).

Maybe we can sync with @pshelar if he knows how aioeventlet patches are handled.

.devcontainer/Dockerfile
```diff
@@ -309,15 +310,44 @@ RUN git clone https://github.com/OPENAIRINTERFACE/opencord.org.freeDiameter.git
...
+COPY /lte/gateway/python/ ${MAGMA_ROOT}/lte/gateway/python/
+WORKDIR ${MAGMA_ROOT}/lte/gateway/python/
+RUN pip3 install --quiet --upgrade --no-build-isolation --no-cache-dir --verbose --editable .[dev] \
+    && rm -rf lte.egg-info
+
+COPY /orc8r/gateway/python/ ${MAGMA_ROOT}/orc8r/gateway/python/
+WORKDIR ${MAGMA_ROOT}/orc8r/gateway/python/
+RUN pip3 install --quiet --upgrade --no-build-isolation --no-cache-dir --verbose --editable .[dev] \
+    && rm -rf orc8r.egg-info
+
```
themarwhal: This is probably just me not understanding how our Python setups work, but what do these do? Are these dependencies (?) that cannot be handled with Bazel?

alexzurbonsen: Yes, it indeed installs the dependencies of lte and orc8r python code (that is, all packages listed in the respective setup.py files under `install_requires` and `extras_require={ 'dev': [ ...`). Sometime in the future these can and will all be handeled by bazel. For now though, this would be a way to achieve more Python support in the devcontainer instantly.

Background: This is essentially what happens in the make file's `install_egg` target (see https://github.com/magma/magma/blob/master/orc8r/gateway/python/python.mk), just baked directly into the container.

.devcontainer/Dockerfile
alexzurbonsen: I added a second commit to make style more consistent.

.devcontainer/Dockerfile
```diff
@@ -81,48 +83,74 @@ RUN echo "Install general purpose packages" && \
...
-        update-alternatives --install /usr/bin/clang-format clang-format /usr/lib/llvm-7/bin/clang-format 10 && \
-        update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/lib/llvm-11/bin/clang-tidy 10 && \
-        update-alternatives --install /usr/bin/clang-apply-replacements clang-apply-replacements /usr/lib/llvm-11/bin/clang-apply-replacements 10
+        wget \
+        && gem install fpm \
+        && update-alternatives --install /usr/bin/clang clang /usr/lib/llvm-11/bin/clang 10 \
+        && update-alternatives --install /usr/bin/clang++ clang++ /usr/lib/llvm-11/bin/clang++ 10 \
+        && update-alternatives --install /usr/bin/clang-format clang-format /usr/lib/llvm-7/bin/clang-format 10 \
+        && update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/lib/llvm-11/bin/clang-tidy 10 \
+        && update-alternatives --install /usr/bin/clang-apply-replacements clang-apply-replacements /usr/lib/llvm-11/bin/clang-apply-replacements 10
```
Neudrino: For consistency this should probably only have 4 white-spaces indentation.

Neudrino (commented):
Thanks so much for the big cleanup! :broom: :shield: :sparkles:

themarwhal (approved):

electronjoe (approved):

.devcontainer/Dockerfile
```diff
@@ -81,48 +83,74 @@ RUN echo "Install general purpose packages" && \
...
+        wget \
+    && gem install fpm \
+    && update-alternatives --install /usr/bin/clang clang /usr/lib/llvm-11/bin/clang 10 \
+    && update-alternatives --install /usr/bin/clang++ clang++ /usr/lib/llvm-11/bin/clang++ 10 \
+    && update-alternatives --install /usr/bin/clang-format clang-format /usr/lib/llvm-7/bin/clang-format 10 \
+    && update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/lib/llvm-11/bin/clang-tidy 10 \
+    && update-alternatives --install /usr/bin/clang-apply-replacements clang-apply-replacements /usr/lib/llvm-11/bin/clang-apply-replacements 10
+
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get install -y clangd-12
```
Neudrino, line 93: Sorry, that I did not notice earlier. I know you did not touch it, however, in terms of general cleanup, this could also be moved to the list line 44ff, as I see no reason to have it separate. This would also contribute towards layer optimization.

.devcontainer/Dockerfile
```diff
@@ -131,193 +159,223 @@ RUN ["/bin/bash", "-c", "if [[ -v GIT_PROXY ]]; then git config --global http.pr
...
+    && tar xf gnutls-3.1.23.tar.xz \
+    && cd gnutls-3.1.23 \
+    && mkdir build && cd build \
+    && CFLAGS=-D_IO_ftrylockfile ../configure --with-libnettle-prefix=/usr/local \
+    && make -j"$(nproc)" \
+    && make install \
+    && ldconfig -v \
+    && cd / \
+    && rm -rf nettle* \
+    && rm -rf gnutls*
```
Neudrino: ~~As before: `rm -rf /nettle* /gnutls*`~~
As before: `rm -rf /gnutls*`

.devcontainer/Dockerfile
```diff
@@ -131,193 +159,223 @@ RUN ["/bin/bash", "-c", "if [[ -v GIT_PROXY ]]; then git config --global http.pr
...
+RUN wget --progress=dot:giga https://ftp.gnu.org/gnu/nettle/nettle-2.5.tar.gz \
+    && tar -xf nettle-2.5.tar.gz \
+    && cd nettle-2.5 \
+    && mkdir build \
+    && cd build/ \
+    && ../configure --disable-openssl --enable-shared --libdir=/usr/local/lib \
+    && make -j"$(nproc)" \
+    && make install \
+    && ldconfig -v \
+    && cd / \
```
Neudrino: There are two separate packages combined in one RUN here. I would see the structure to be more modular, if we could separate this in nettle and gnutls with to RUN statements.
The cleanup would than need to be moved to the appropriate section, of course.
You might need to check if `--with-libnettle-prefix=/usr/local` is a blocker for that, however, `--libdir=/usr/local/lib` suggests, that the first install only needs to complete successfully.

.devcontainer/Dockerfile
```diff
@@ -131,193 +159,223 @@ RUN ["/bin/bash", "-c", "if [[ -v GIT_PROXY ]]; then git config --global http.pr
...
-    rm -rf grpc
+RUN git clone --recurse-submodules -b v1.35.0 https://github.com/grpc/grpc \
+    && cd grpc \
+    && mkdir -p cmake/build \
+    && cd cmake/build \
+    && cmake -DgRPC_INSTALL=ON -DgRPC_BUILD_TESTS=OFF -DBUILD_SHARED_LIBS=ON ../.. \
+    && make -j"$(nproc)" \
+    && make install \
+    && cd / \
+    && rm -rf grpc
```
Neudrino: :heavy_plus_sign:

.devcontainer/Dockerfile
```diff
@@ -131,193 +159,223 @@ RUN ["/bin/bash", "-c", "if [[ -v GIT_PROXY ]]; then git config --global http.pr
...
-    cd / && \
-    rm -rf libprotobuf-mutator
+RUN git clone -b v1.0 https://github.com/google/libprotobuf-mutator \
+    && mkdir -p libprotobuf-mutator/build \
+    && cd libprotobuf-mutator/build \
+    && cmake .. -GNinja -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DCMAKE_BUILD_TYPE=Debug \
+    && ninja \
+    && ninja install \
+    && cd / \
+    && rm -rf libprotobuf-mutator
```
Neudrino: :heavy_plus_sign:

.devcontainer/Dockerfile
```diff
@@ -131,193 +159,223 @@ RUN ["/bin/bash", "-c", "if [[ -v GIT_PROXY ]]; then git config --global http.pr
...
-    make install && \
-    ldconfig
+RUN git clone https://gitlab.eurecom.fr/oai/asn1c.git \
+    && cd asn1c \
+    && git checkout f12568d617dbf48497588f8e227d70388fa217c9 \
+    && autoreconf -iv \
+    && ./configure \
+    && make -j"$(nproc)" \
+    && make install \
+    && ldconfig
```
Neudrino: We probably should also remove this source after install.

.devcontainer/Dockerfile
```diff
@@ -131,193 +159,223 @@ RUN ["/bin/bash", "-c", "if [[ -v GIT_PROXY ]]; then git config --global http.pr
...
-    ldconfig -v
+RUN cd /usr/src/googletest \
+    && mkdir build \
+    && cd build \
+    && cmake -DBUILD_SHARED_LIBS=ON .. \
+    && echo "Build gtest and gmock" \
+    && make \
+    && echo "Install gtest and gmock" \
+    && make install \
+    && ldconfig -v
```
Neudrino: We probably should also remove this source after install.

.devcontainer/Dockerfile
```diff
@@ -131,193 +159,223 @@ RUN ["/bin/bash", "-c", "if [[ -v GIT_PROXY ]]; then git config --global http.pr
...
+    && cd include-what-you-use \
+    && git checkout 0.15 \
+    && cd .. \
+    && mkdir build_iwyu \
+    && cd build_iwyu \
+    && cmake -G "Unix Makefiles" -DIWHYU_LLVM_ROOT_PATH=/usr/lib/llvm-11 ../include-what-you-use/ \
+    && make \
+    && make install \
+    && rm -rf /include-what-you-use \
+    && rm -rf /build_liwyu
```
Neudrino: `rm -rf /include-what-you-use /build_liwyu`

Neudrino (commented):
Sorry, I noticed some more cleanup things. :innocent: 
But if we do a cleanup now, it will not pop-up later repeatedly. :wink:

Neudrino:
Could we do something about the linter warnings?

Neudrino:
After offline discussion:
I am OK with reverting all cleanup and only committing the original changes.

However, I still believe, that newly added code should be consistent with existing one and have good style where possible (and it does not interfere with the first item).

Neudrino (approved):

PR #11294, opened 2022-01-27
chore(nms): Upgrade webpack from 4.46.0 to 5.1.0

spikey979:

chore(nms): Upgrade webpack from 4.46.0 to 5.1.0

## Summary

   The npm package elliptic before version 6.5.4 are vulnerable to Cryptographic Issues via the secp256k1 implementation in elliptic/ec/key.js. There is no check to confirm that the public key point passed into the derive function actually exists on the secp256k1 curve. This results in the potential for the private key used in this implementation to be revealed after a number of ECDH operations are performed.
   Elliptic package is indirectly used in package webpack v4.46.0. So webpack needs to be upgraded to version 5.1.0 which no longer uses package elliptic. 

## Test Plan

ran in /magma/nms folder:
yarn run eslint
yarn run flow
yarn run test

## Additional Information

PR is fix for dependabot alert: https://github.com/magma/magma/security/dependabot/nms/yarn.lock/elliptic/open

### DIFF
--- a/nms/packages/magmalte/package.json
+++ b/nms/packages/magmalte/package.json
@@ -96,9 +96,9 @@
     "swagger-js-codegen": "^1.13.0",
     "terser-webpack-plugin": "^1.2.3",
     "url-loader": "^4.1.1",
-    "webpack": "^4.46.0",
+    "webpack": "^5.1.0",
     "webpack-bundle-analyzer": "^3.9.0",
-    "webpack-cli": "^3.2.3",
+    "webpack-cli": "^4.9.2",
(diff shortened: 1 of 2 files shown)


lucasgonze:
The CI error is flagging webpack, which suggests that the bump failed because of the commit. 

image

The semver indicates a breaking change. With luck there is a 4.x webpack with fixed dependency. Otherwise the fix will require a substantive look at the code.

lucasgonze:
Linking to issue to make tracking easier: https://github.com/magma/security/issues/56

Issue #11300, opened 2022-01-27
MME: unit test fails with LSAN when DEBUG_IS_ON is set to 1


themarwhal:
When I briefly looked into https://github.com/magma/magma/issues/10320 in December, I found that this diff results in a memory leak 
![263802953_332398298375997_2171707110115550014_n](https://user-images.githubusercontent.com/37634144/151409980-dc5abc6c-1213-4439-a65c-68bf86847efa.png)

Scott pointed to me this CMake doc (https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html)  which makes this behavior very puzzling.

Repro steps: Apply the patch and run `make -C lte/gateway test_oai BUILD_TYPE=RelWithDebInfo`

themarwhal:
Actually... is this still a case?

Issue #10320, opened 2021-11-15
[mme] Remove deprecated `DEBUG_IS_ON` compiler flag for oai/mme


electronjoe:
It has been discovered that the `oai/mme` will not build if `DEBUG_IS_ON` is undefined.  This is in part because some macro definitions (e.g. `c/core/oai/common/log.h:OAILOG_DEBUG`) are not defined if `DEBUG_IS_ON` is undefined, yet these macros are sometimes used even with `DEBUG_IS_ON` undefined -> leading to code that cannot be compiled.

Given that it is not possible to compile the MME without `DEBUG_IS_ON` set any longer, it seems reasonable to remove this functionality (which is no used any longer).

PR #11304, opened 2022-01-27
chore: fix CMake style in MME 

themarwhal:
## Summary
Back in December, I was looking into https://github.com/magma/magma/issues/10320 and noticed that the DEBUG_IS_ON flag config are inconsistent. `-DDUBUG_IS_ON` and `-DDEBUG_IS_ON=1` are identical according to cmake doc so fixing it to avoid confusion. See https://github.com/magma/magma/issues/11300

## Test Plan
CI mme test job checks both DEBUG and RELWITHDEBINFO build types

## Additional Information


ssanadhya (approved):
Thanks for the cleanup!

PR #11311, opened 2022-01-28
fix(agw): Added response timeout for stuck test cases

### LINKED ISSUE #10596: [lte][agw] Add retries support for LTE integ tests
**Feature description**

Current process of execution of LTE integ tests on CI is really flaky and is dependent on different external factors which leads to a very low number of success runs. CWAG integ tests already support retries for the list of tests, however, they are written in Go and thus we cannot use the `gotest` flags to achieve retries the same way. Thus we need a different approach to define retry logic.

**Implementation steps**
1. Install flaky nosetest plugin https://opensource.box.com/flaky/ 
2. Mark all LTE integ tests with flaky flags on nosetests Make command
2.1 Max number of retries = 3, Minimum pass threshold = 1
3. Clean_restart implementation to clean state, define callback function 
3.1 How to cover for stuck / segfault integ tests issues? (AI: Check on a method to exit test)
4. Add API call to s1ap tester to reset configuration for further retries
5. Add API call for restarting trfserver daemon on traffic server VM

Note: It's important to keep visibility of failures, this should not mask any failure on any test

**Exit criteria**
Current pass rate on CI for some integ tests goes from 70%/80% to 95% (on last 30 days - without taking into account CI infra related issues), we are aiming to increase this percentage for all integ tests to at least 95%.

VinashakAnkitAman:
## Title
fix(agw): Added response timeout for stuck test cases

## Summary
Some of the test cases seem to stuck for a long time waiting for some response messages from the mme. The reason of test cases to be stuck is mostly either the test case is expecting some message which it is not getting from the AGW or the AGW services have restarted/crashed and unable to generate responses towards S1APTester. This PR adds a timeout for the response message call so that after a maximum allocated wait time, the test case can fail with assertion and lead to execution of further steps.

## Test plan


lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -179,9 +180,37 @@ def get_ip(self, ue_id):
...
+        try:
+            return self._msg.get(True, timeout)
+        except Empty:
+            if assert_on_timeout is True:
+                raise AssertionError(
+                    "Timeout ("
+                    + str(timeout)
+                    + " sec) occurred while waiting for response message",
+                )
+            else:
```
ardzoht: remove else case

ardzoht (changes requested):

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -179,9 +180,37 @@ def get_ip(self, ue_id):
...
+        """
+        if timeout is None:
+            # Default maximum wait time is 900 sec (15 min)
+            timeout = S1ApUtil.MAX_RESP_WAIT_TIME
+
+        # Wait until callback is invoked or timeout occurred
+        try:
+            return self._msg.get(True, timeout)
+        except Empty:
+            if assert_on_timeout is True:
```
ardzoht:
```suggestion
            if assert_on_timeout:
```

ardzoht (approved):
minor nit on if condition, please address, overall LGTM

PR #11312, opened 2022-01-28
fix: Set feg-precommit label on github action

uri200:

## Summary

FEG precommit includes lint, so we don't need to run it separately
```
from feg Makefile

precommit: tools fmt lint test vet
```

Also FEG shouldn't be running code at cwf

## Test Plan

Linter executed here 

image

## Additional Information

### DIFF
--- a/.github/workflows/feg-workflow.yml
+++ b/.github/workflows/feg-workflow.yml
@@ -43,7 +43,7 @@ jobs:
           name: pr
           path: pr/
 
-  feg-lint-precommit:
+  feg-precommit:
     needs: path_filter
     if: ${{ needs.path_filter.outputs.should_not_skip == 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/feg-workflow.yml
```diff
@@ -60,21 +60,15 @@ jobs:
...
         run: |
               ./circleci/golang_before_install.sh
-      - name: Run go mod download with retry
-        uses: nick-invision/retry@v2
+      - name: make feg precommit
         if: always()
-        id: feg-lint-init
-        with:
-          command: cd ${MAGMA_ROOT}/cwf/gateway && go mod download
-          timeout_minutes: 10
```
uri200, line 69: maybe i am missing something but why is this `/cwf/gateway` necessary on feg

PR #11317, opened 2022-01-28
chore(agw): remove impl initialized function and avoid multiple initialization

ankitMeta1:
## Summary

remove _is_impl_initialized and put a initialized check on init_impl

## Test Plan

ran unit test locally


lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -180,24 +180,21 @@ def init_impl(self, datapath):
         """
         Takese in datapath, and initializes appropriate QoS manager based on config
         """
+        if self._initialized:
```
pshelar, line 183: can you also move lock to object attribute from class attribute, you could create separate PR to address this.

pshelar (approved):
LGTM

PR #11320, opened 2022-01-28
fix(mme): Update GetENBState to remove S1AP state mgr dependency

### LINKED ISSUE #8450: [agw][BE] S1AP gRPC servicer state access violation
### Your Environment

- **Version:** master
- **Affected Subcomponent:** MME

### Describe the Issue

- Currently, enodebd sends out a `GetENBState` gRPC request for obtaining the current eNB state from s1ap task, this is a periodic call that is done every minute through the service303's `get_operational_states` function.
- This results in a crash due to the gRPC call done by enodebd comes before `s1ap_state` getting read and populated from redis, there is an assertion done by the StateManager that expects the state to be not null. 

**Proposed changes**

> To fix the resulting crash due to the `AssertFatal` on the S1APStateManager, `GetENBState` can be modified to be a push (or streaming) model. 
> This gRPC call will be initiated by the MME and should only be done after all S1AP task state in the service is populated.
> This will also allow to maintain the `get_s1ap_state` retrieved call for now, to ensure state sync and ownership of the task's state. 

- Current service definition:
```
service S1apService {
    // Returns state of the S1 connected eNBs
    rpc GetENBState (magma.orc8r.Void) returns (EnbStateResult) {}
}
```

- Updated service definition (with streaming):
```
service S1apService {
    // Pushes a stream of S1 connected eNBs map
    rpc ListEnbState (magma.orc8r.Void) returns (stream EnbStateResult) {}
}
```

- `enodebd/s1ap_service.py` gRPC client updates will be needed to reflect consuming of the `EnbStateResult` stream.

ardzoht:

## Summary

- Removes `get_s1ap_state` usage from S1AP GRPC service which was causing segfault due to `s1ap_state` nullptr, closes #8450 
- Refactors `S1apServiceImpl::GetENBState` for using direct access to redis to construct GRPC response 
- Formatting changes

## Test Plan

- Testing on local environment to ensure eNB state is retrieved by `enodebd` service
- Pending: Add unit test / integ test for `GetENBState` endpoint

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/grpc_service/S1apServiceImpl.cpp
+++ b/lte/gateway/c/core/oai/tasks/grpc_service/S1apServiceImpl.cpp
@@ -16,54 +16,51 @@
  */
 #include "lte/gateway/c/core/oai/tasks/grpc_service/S1apServiceImpl.h"
 
+#include <memory>
 #include <string>
 
-extern "C" {
-#include "lte/gateway/c/core/oai/common/assertions.h"
-#include "lte/gateway/c/core/oai/lib/hashtable/hashtable.h"
-#include "lte/gateway/c/core/oai/common/log.h"
-#include "lte/gateway/c/core/oai/include/s1ap_state.h"
-}
+#include "lte/protos/oai/s1ap_state.pb.h"
+#include "orc8r/protos/common.pb.h"
+#include "redis_utils/redis_client.h"
 
 using grpc::ServerContext;
 using grpc::Status;
-using magma::EnbStateResult;
-using magma::S1apService;
+using magma::lte::EnbStateResult;
+using magma::lte::S1apService;
+using magma::lte::oai::S1apState;
+using magma::orc8r::Void;
 
 namespace magma {
-using namespace lte;
-using namespace orc8r;
+namespace lte {
 
-S1apServiceImpl::S1apServiceImpl() {}
+S1apServiceImpl::S1apServiceImpl() : client_(nullptr) {}
+
+void S1apServiceImpl::init(std::shared_ptr<RedisClient> client) {
+  client_ = client;
+}
 
 Status S1apServiceImpl::GetENBState(
     ServerContext* context, const Void* request, EnbStateResult* response) {
   OAILOG_DEBUG(LOG_UTIL, "Received GetENBState GRPC request\n");
 
-  // Get state from S1APStateManager
-  // TODO: Get state through ITTI message from S1AP task, as it's read only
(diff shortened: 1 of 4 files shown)


lte/gateway/c/core/oai/tasks/grpc_service/S1apServiceImpl.cpp
```diff
@@ -16,54 +16,51 @@
...
-    }
+  S1apState s1ap_state_msg = S1apState();
+  auto rc                  = client_->read_proto("s1ap_state", s1ap_state_msg);
+  if (rc != RETURNok) {
+    OAILOG_DEBUG(
+        LOG_UTIL,
+        "Failure while reading s1ap_state from redis. Cancelling GetENBState "
+        "request");
+    return Status::CANCELLED;
+  }
```
ulaskozat, line 54: This type of design has two major issues:

1. GRPC servicer is not the owner of the state and it should not rely on Redis backend to fetch it. The actual owner may evolve and change how/when these states are updated/refactored on Redis.

2. This breaks the parity between stateful and stateless. E.g., if someone disables stateless feature, eNB state will no longer be read.

ulaskozat (changes requested):
@ardzoht I did not raise this concern during the last meeting, but the change you are doing here needs to be discussed at a higher architectural level.

ulaskozat:
The motivating idea for this particular change is to eliminate unnecessary ITTI messages. Then, why don't we consider making this a push rather than pull model? i.e., MME makes the rpc call to deliver the eNB states.

ardzoht:
> The motivating idea for this particular change is to eliminate unnecessary ITTI messages. Then, why don't we consider making this a push rather than pull model? i.e., MME makes the rpc call to deliver the eNB states.

There are currently no ITTI messages being passed, motivating idea for this is that there is an issue when the RPC call is done to MME and the `s1ap_state` is null: https://sentry.io/organizations/lf-9c/issues/2897744515/?project=5754002&referrer=github_integration (this is happening on service initialization, likely due to the gRPC service being initialized and the S1AP task not having populated state yet), and so the service crashes.

Couple of remarks on the gRPC service and how state is used:

1. Push model seems like a good approach given that the MME initiates communication, but with the expectation this should be done once all local state on MME is populated.
2. Ideally an approach to take for all state throughout MME is to define redis as a first class citizen for storage (no sync between local state and Redis, no stateless vs stateful modes), obviously this is out of scope for this change but something to keep in mind while thinking on the ownership of the states.
cc @ssanadhya

PR #11327, opened 2022-01-31
chore(mobilityd): Exclude invalid IP block error from monitoring

### LINKED ISSUE #10846: Frequent Sentry Error: Invalid IP block format in mobilityd mconfig
Sentry Issue: [LAB-AGWS-PYTHON-42](https://sentry.io/organizations/lf-9c/issues/2857488203/?referrer=github_integration)

```
Invalid IP block format: 192.168.127.0/16
```

sebathomas:
## Summary

User configuration errors don't need to be tracked in error monitoring.

In principle it would also be possible to set `strict=False` in the IP address conversion. However, it is probably better to be strict here instead of trying to interpret ambiguous IP ranges given by the user.

Fixes #10846.

## Additional Information


ardzoht (approved):
lgtm

PR #11347, opened 2022-02-01
test(agw): Use magma-test VM for running trf server

pshelar:
This creates separate namespace to run trf-server on magma-test
VM thus eliminating need of running magma-trfserver in
separate VM.
This would further reduce requirement of running magma integ-tests.

## Additional Information

### DIFF
--- a/lte/gateway/deploy/roles/trfserver/files/traffic_server.py
+++ b/lte/gateway/deploy/roles/trfserver/files/traffic_server.py
@@ -522,8 +522,15 @@ class TrafficTestDriver(object):
     def _get_macs(self):
         ''' Retrieves the MAC addresses of the associated test servers, based
         on the information of the instances '''
+        dev = 'eth2'
         ip = pyroute2.IPRoute()
-        mac = ip.link('get', index=ip.link_lookup(ifname='eth2')[0])[0] \
+        # In case of trf-server in namespace, the namespace only has eth3
+        # Therefore if eth2 is missing,  use eth3.
+        # This check is done for compatibility for namespace based trf-serve
+        res = ip.link_lookup(ifname=dev)
+        if len(res) == 0:
+            dev = 'eth3'
+        mac = ip.link('get', index=ip.link_lookup(ifname=dev)[0])[0] \
             .get_attr('IFLA_ADDRESS')
 
         return (mac,) * len(self._instances)
--- a/lte/gateway/Vagrantfile
+++ b/lte/gateway/Vagrantfile
@@ -120,6 +120,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     # UE trfgen network
     magma_test.vm.network "private_network", ip: "192.168.128.11", nic_type: "82540EM"
     #config.ssh.private_key_path = "~/.ssh/vagrant.key"
+    magma_test.vm.network "private_network", ip: "192.168.129.43", nic_type: "82540EM"
     config.ssh.forward_agent = true
 
     magma_test.vm.provider "virtualbox" do |vb|
--- a/lte/gateway/fabfile.py
+++ b/lte/gateway/fabfile.py
@@ -250,7 +250,7 @@ def federated_integ_test(
 
 def integ_test(
     gateway_host=None, test_host=None, trf_host=None,
-    destroy_vm='True', provision_vm='True',
+    destroy_vm='True', provision_vm='True', use_trf_ns='False',
 ):
     """
     Run the integration tests. This defaults to running on local vagrant
@@ -272,6 +272,7 @@ def integ_test(
 
     destroy_vm = bool(strtobool(destroy_vm))
     provision_vm = bool(strtobool(provision_vm))
+    use_trf_ns = bool(strtobool(use_trf_ns))
 
     # Setup the gateway: use the provided gateway if given, else default to the
     # vagrant machine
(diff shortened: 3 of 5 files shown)


ardzoht:
@pshelar What is the motivation for this change? Is this intended for CI integ test improvement? 

I would assume that this still allows the user to use the separate VMs to run traffic tests? @VinashakAnkitAman and me were entertaining this idea of running traffic server in the `magma_test` VM however it didn't made much sense to us from the implementation perspective.

Also if we're going to go forward with this please include the evidence of test plan.

pshelar:
> @pshelar What is the motivation for this change? Is this intended for CI integ test improvement?
> 
This would eliminate trf-server VM. so all integ tests would needs just two VM, magma and magma-test.

> I would assume that this still allows the user to use the separate VMs to run traffic tests? @VinashakAnkitAman and me were entertaining this idea of running traffic server in the `magma_test` VM 

This PR does same, runs trf-server in magma-test VM.

> however it didn't made much sense to us from the implementation perspective.

Can you elaborate on this?

> Also if we're going to go forward with this please include the evidence of test plan.

I ran `fab integ_test:destroy_vm=False,use_trf_ns=True` which works.

VinashakAnkitAman:
Thanks @pshelar for the details above. May be I have missed some core parts of discussion here, so wanted to know a bit more. Since trfserver VM was having minimum hardware requirement and load for simply running the iperf server, I was curious to know that do we see some benefits other than saving hardware resources as well. Because probably going further we have a plan to have containerized setups where one magma-test and one TRFServer would be sufficient to serve many AGW instances in the network. So hardware resources would be minimally required in any way combinedly for magma-test and TRFServer.

Why I am asking above is that _the current 3 VM setup seems to be an ideal scenario for a real deployable environment where we have UE/ENB talking on one interface side of AGW and the data server talking on another_. Moreover, since the UE is allocated a NATted IP which does not have direct access to data server and the data goes encapsulated as GTP packet via SPGW tunnel to the Trfserver and vice versa. So issues with data capture, tunnel setup and data encapsulation are comparatively easily identifiable with failing data test cases in current architecture. But combining the magma-test and TRFServer in one VM/container would have both the interfaces in same VM, and will probably have more potential risk of bypassing the SPGW data tunnel and might lead to some fake successes/failures at times when running test scripts.

That's why just wanted to understand if combining the 2 VMs is really of much need or help.

pshelar:
> Thanks @pshelar for the details above. May be I have missed some core parts of discussion here, so wanted to know a bit more. Since trfserver VM was having minimum hardware requirement and load for simply running the iperf server, I was curious to know that do we see some benefits other than saving hardware resources as well. Because probably going further we have a plan to have containerized setups where one magma-test and one TRFServer would be sufficient to serve many AGW instances in the network. So hardware resources would be minimally required in any way combinedly for magma-test and TRFServer.
> 
It is pretty wasteful to create and run entire VM for single light weight process. Note we are destroying and creating VMs on each CI run.

I do not see any issues  using proposed approach to extend it to test against multiple AGWs in future. 

> Why I am asking above is that _the current 3 VM setup seems to be an ideal scenario for a real deployable environment where we have UE/ENB talking on one interface side of AGW and the data server talking on another_. Moreover, since the UE is allocated a NATted IP which does not have direct access to data server and the data goes encapsulated as GTP packet via SPGW tunnel to the Trfserver and vice versa. So issues with data capture, tunnel setup and data encapsulation are comparatively easily identifiable with failing data test cases in current architecture. But combining the magma-test and TRFServer in one VM/container would have both the interfaces in same VM, and will probably have more potential risk of bypassing the SPGW data tunnel and might lead to some fake successes/failures at times when running test scripts.
> 

This PR separates the trf-server using  separate network namespace, there is no direct route between simulated UE and trf-server. The networking between UE, agw and trf-server is same. trf-server is using same network as trf-server-VM

 so I do not understand how it can result in fake successes ? Can you give concrete example?

> That's why just wanted to understand if combining the 2 VMs is really of much need or help.

This PR is basically keeping same design as we have today, except using two VMs. I dont see any downside of doing this. we could push this PR as option to run integ-tests using two VMs. This PR does not break existing integ VMs setup.

ardzoht:
@pshelar Can we then include these changes to help CI runs to use only 2 VMs, but still keep the option to spin up the trfserver VM for local environments as a separate setup?

pshelar:
> @pshelar Can we then include these changes to help CI runs to use only 2 VMs, but still keep the option to spin up the trfserver VM for local environments as a separate setup?

This change does not get rid of trf-server, you have to set a flag to run integ-test with 2 VMs `fab integ_test:destroy_vm=False,use_trf_ns=True`
So even after applying this PR, CI as well as local testing can continue to use separate trf-server VM with current defaults.

uri200:
Hey @pshelar , would make sense to move this to the magma vm instead? If we can create those network spaces, maybe we can think in the future moving magma_test S1AP to the dev and run it all in one single vm?

lte/gateway/deploy/roles/trfserver/files/traffic_server.py
```diff
@@ -522,8 +522,14 @@ def __init__(self, server, instances):
     def _get_macs(self):
         ''' Retrieves the MAC addresses of the associated test servers, based
         on the information of the instances '''
+        dev = 'eth2'
         ip = pyroute2.IPRoute()
-        mac = ip.link('get', index=ip.link_lookup(ifname='eth2')[0])[0] \
+        # incase of missing eth2, use eth3.
+        # This check is done for compatibility for namespace based trf-serve
```
uri200, line 529: so is `eth2` used when server is run on magma_trfserver and `eth3` when server is run on `magma_test`?

If so, can me note this here just in case this needs to be changed in the future?

pshelar:
> Hey @pshelar , would make sense to move this to the magma vm instead? If we can create those network spaces, maybe we can think in the future moving magma_test S1AP to the dev and run it all in one single vm?

I explored it, But it made networking alot complicated for datapath tests for NAT and NonNAT. I would like to go there gradually.

maxhbr:
@pshelar this PR has conflicts

MoritzThomasHuebner:
Closing this because of inactivity

PR #11365, opened 2022-02-01
fix(pipelined): Fix Unit Test arg init for qos, add ebpf guards

koolzz:

## Summary
 -  Qos init seems to fail in unit tests because in qos init we don't initiliaze all the args because of the premature return statements (introduced here https://github.com/magma/magma/pull/11287, cc @ankitbhadoria21 )
 - Also noticed some ebpf issues with nat_iface lookup in tests, added some checcks

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
+++ b/lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
@@ -52,6 +52,9 @@ def get_ebpf_manager(config):
     else:
         enabled = False
     gw_info = get_mobilityd_gw_info()
+    if not ('nat_iface' in config and 'enodeb_iface' in config):
+        LOG.info("eBPF manager: Missing nat_iface/ennodeb_iface")
+        return None
     for gw in gw_info:
         if gw.ip.version != IPAddress.IPV4:
             continue
--- a/lte/gateway/python/magma/pipelined/qos/common.py
+++ b/lte/gateway/python/magma/pipelined/qos/common.py
@@ -178,6 +178,8 @@ class QosManager(object):
         """
         Takese in datapath, and initializes appropriate QoS manager based on config
         """
+        if not self._qos_enabled:
+            return
         if self._initialized:
             return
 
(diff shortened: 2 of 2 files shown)


themarwhal:
force merging to unblock lte-integ-test on CI

ankitMeta1:
I am actually curious about this one @koolzz, looking at the changes you made here and I made in my mentioned PR, it looks like the the check I added to see whether 'qos' key exists in config or not, in the QoSManager constructor caused it?, I added it because we were accessing the 'qos' section from the config, and if it doesn't exist, then an exception will be thrown, so we should probably check for it, before accessing.

The old code operates like this, it tries to access 'qos' field and call setup, but if an exception is thrown and we will not call setup in QoSManager. So I am just wondering if previously the test was previously silently not running?

koolzz:
> I am actually curious about this one @koolzz, looking at the changes you made here and I made in my mentioned PR, it looks like the the check I added to see whether 'qos' key exists in config or not, in the QoSManager constructor caused it?, I added it because we were accessing the 'qos' section from the config, and if it doesn't exist, then an exception will be thrown, so we should probably check for it, before accessing.
> 
> The old code operates like this, it tries to access 'qos' field and call setup, but if an exception is thrown and we will not call setup in QoSManager. So I am just wondering if previously the test was previously silently not running?

Yeah the error was in the  init_impl function as the variable that we were checki was not initialized

ankitMeta1:
> > I am actually curious about this one @koolzz, looking at the changes you made here and I made in my mentioned PR, it looks like the the check I added to see whether 'qos' key exists in config or not, in the QoSManager constructor caused it?, I added it because we were accessing the 'qos' section from the config, and if it doesn't exist, then an exception will be thrown, so we should probably check for it, before accessing.
> > The old code operates like this, it tries to access 'qos' field and call setup, but if an exception is thrown and we will not call setup in QoSManager. So I am just wondering if previously the test was previously silently not running?
> 
> Yeah the error was in the init_impl function as the variable that we were checki was not initialized

Cool. then I think my other hypothesis is also right. becuase I didn't reorder the initialization of variables in  the constructor, I just added a condition to check for 'qos'. So in the previous code before my change, if you were to try to access 'qos' from config, it will throw exception(https://github.com/magma/magma/pull/11287/files#diff-afaf023d5c72e158db82116e628ecf30043e30cb73d8482163e56150eef37c6fL244) and then we won't call setup(https://github.com/magma/magma/pull/11287/files#diff-afaf023d5c72e158db82116e628ecf30043e30cb73d8482163e56150eef37c6fL183-L184) or initialize QosStore. So I don't think those tests were running correctly before? maybe they were silently off? @koolzz @pshelar

PR #11369, opened 2022-02-01
fix(agw): Replace master with main for S1apTester repo

ssanadhya:

## Summary

The script to update S1apTester library on Magma test VM was failing since it was referring to `main` branch of [S1APTester repo](https://github.com/facebookexperimental/S1APTester) as `master`. This change fixes this terminology.

## Test Plan

On `magma_test VM`, verified that:
- `update_package.sh <pull request> ` executes without error
- a new `~/s1ap-tester/bin/libtfw.so` is generated after the above script succeeds


themarwhal (approved):

ardzoht (approved):
lgtm

PR #11377, opened 2022-02-02
fix(agw): Updating priority of default rule in AGW

bhuvaneshne:
## Summary

Default rule was previously installed with priority 2. 
This is inverted before flows are pushed to OVS - Which resulted in very high priority for "this" rule.
This caused side effects when multiple policies are tested with differing priorities. 
Fix involves changing priority to a very high value (Lowering priority). 

## Test Plan

Bring up end to end setup. Attach UE with atleast two policies. Dump flows in OVS and inspect.

pipelined_cli.py debug display_flows | grep "table=enforcement(main_table)" 

image

Check if the default rules are installed with low priority in OVS
(See screenshot above): Default rules are installed with PRIORITY=5 (very low)
The other rules (defaultBearerPolicy, dedicatedBearerTCP) are installed with higher priorities

### DIFF
--- a/lte/gateway/python/magma/policydb/default_rules.py
+++ b/lte/gateway/python/magma/policydb/default_rules.py
@@ -29,8 +29,9 @@ def get_allow_all_policy_rule(
         # Don't set the rating group
         # Don't set the monitoring key
         # Don't set the hard timeout
+        # Setting priority to a low value
         id=policy_id,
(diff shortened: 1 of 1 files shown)


maxhbr:
This was today mentioned as important in https://magmacore.slack.com/archives/C0191LLL03S/p1673942453345879

PR #11383, opened 2022-02-02
docs(orc8r): Removed install orc8r managed page

christinewang5:
## Summary
`Managed Orc8r Deployment via Orcl` is no longer used and is not the primary method to deploy, so is generally confusing to leave in the docs.


themarwhal (approved):

hcgatewood (changes requested):
Awesome! Looks like there are a couple references to this page as well, both in other pages and in the sidebars.json. Can you remove those as well?

themarwhal:
Removing `ready2merge` as it still needs Hunter's stamp

hcgatewood (approved):

PR #11393, opened 2022-02-02
chore: downgrade platform-server from 0.3.0 to 0.2.3

lucasgonze:
## Summary

This is debugging to find the commit that introduced a problem. This PR should never be merged.

## Additional Information


lucasgonze:
Debugging complete. Fix will be in https://github.com/facebookincubator/fbc-js-core/pull/159.

PR #11401, opened 2022-02-03
chore(deps): Upgrade eventlet from v0.30.2 to v0.31.0

spikey979:

chore(deps): Upgrade eventlet from v0.30.2 to v0.31.0

## Summary

A websocket peer may exhaust memory on Eventlet side by sending very large websocket frames. Malicious peer may exhaust memory on Eventlet side by sending highly compressed data frame. Version 0.31.0 restricts websocket frame to reasonable limits. Eventlet package should be upgraded to v0.31.0 or later.

## Test Plan

ran magma/lte/gateway make test_python

## Additional Information

This is for dependabot alert: https://github.com/magma/magma/security/dependabot/lte/gateway/python/setup.py/eventlet/open

### DIFF
--- a/lte/gateway/python/setup.py
+++ b/lte/gateway/python/setup.py
@@ -120,7 +120,7 @@ setup(
         'idna==2.8',
         'python-dateutil==2.8.1',
         'six>=1.12.0',
-        'eventlet==0.30.2',
+        'eventlet==0.31.0',
         'h2>=3.2.0',
         'hpack>=3.0',
         'freezegun>=0.3.15',


ardzoht:
@spikey979 thanks for this change. could I ask to also verify the installation of this in `magma_test` VM? 

unfortunately we have a mismatch on both of these VMs regarding the installed python versions, and there are some packages that are available in 3.8 that are not available in 3.5. 

I don't expect this one to be the case but just to be safe let's double check in that front, with that confirmation I can proceed and accept the PR.

ardzoht (changes requested):
please also verify installation on `magma_test` VM

PR #11405, opened 2022-02-03
fix(dp): single DB queries in RC active mode service

xbend:
When building database state for Active Mode Controller,
RC active mode service should use a single select query when fetching
database data required to build the Domain Proxy state, to resolve result inconsistency when a DB commit was
issued in between the selects.

## Summary

* RC: modified DB state query as one compound select for all state-related entities


m-trojanowski (approved):
LGTM!

PR #11406, opened 2022-02-03
chore: generate Python proto files in devcontainer

alexzurbonsen:
## Summary

Generate Python proto files in DevContainer analogously to the make files (by using `gen_protos.py` and `gen_prometheus_proto.py` modules).

Add extension to VSCode:
 - [proto3 extension](https://marketplace.visualstudio.com/items?itemName=zxh404.vscode-proto3): understands proto (syntax highlighting, code completion, code formatting and more)

## Test Plan

Built DevContainer locally. Checked that proto modules can be imported.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -347,6 +347,23 @@ WORKDIR ${MAGMA_ROOT}/orc8r/gateway/python/
 RUN pip3 install --quiet --upgrade --no-build-isolation --no-cache-dir --verbose --editable .[dev] && \
     rm -rf orc8r.egg-info
 
+#### protos
+ARG PYTHON_LIB=/usr/local/lib/python3.8
+
+COPY /protos/ ${MAGMA_ROOT}/protos/
+COPY /lte/protos/ ${MAGMA_ROOT}/lte/protos/
+COPY /orc8r/protos/ ${MAGMA_ROOT}/orc8r/protos/
+COPY /feg/protos/ ${MAGMA_ROOT}/feg/protos/
+COPY /dp/protos/ ${MAGMA_ROOT}/dp/protos/
+WORKDIR $MAGMA_ROOT
+RUN pip install mypy-protobuf && \
+    mkdir $PYTHON_LIB/gen && \
+    for PROTO_SRC in orc8r lte feg dp; \
+    do \
+    python protos/gen_protos.py ${PROTO_SRC}/protos ${MAGMA_ROOT},orc8r/protos/prometheus ${MAGMA_ROOT} $PYTHON_LIB/gen && \
+    python protos/gen_prometheus_proto.py ${MAGMA_ROOT} $PYTHON_LIB/gen; \
+    done && \
+    echo "$PYTHON_LIB/gen" > $PYTHON_LIB/dist-packages/magma_gen.pth
 
 RUN ln -s $MAGMA_ROOT/bazel/bazelrcs/devcontainer.bazelrc /etc/bazelrc
 
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -21,6 +21,7 @@
 		"ms-python.python",
 		"ms-python.vscode-pylance",
 		"njpwerner.autodocstring",
+		"zxh404.vscode-proto3",
 	],
 	"image": "ghcr.io/magma/magma/devcontainer:sha-206a67a",
 	"settings": {


.devcontainer/Dockerfile
```diff
@@ -336,17 +336,30 @@ COPY lte/gateway/deploy/roles/magma/files/patches/aioeventlet_fd_exception.patch
 RUN patch -N -s -f /usr/local/lib/python3.8/dist-packages/aioeventlet.py < /tmp/aioeventlet_fd_exception.patch && \
     rm -rf /tmp/*
 
+COPY . ${MAGMA_ROOT}
```
jheidbrink: Could we copy just the needed files? That might reduce the docker image size, and also express more clearly which files the following steps depend on.

.devcontainer/Dockerfile
```diff
@@ -336,17 +336,30 @@ COPY lte/gateway/deploy/roles/magma/files/patches/aioeventlet_fd_exception.patch
...
+
+WORKDIR $MAGMA_ROOT
+RUN mkdir $PYTHON_LIB/gen && \
+    for PROTO_SRC in orc8r lte feg; \
+    do \
+    python protos/gen_protos.py ${PROTO_SRC}/protos ${MAGMA_ROOT},orc8r/protos/prometheus ${MAGMA_ROOT} $PYTHON_LIB/gen && \
+    python protos/gen_prometheus_proto.py ${MAGMA_ROOT} $PYTHON_LIB/gen; \
+    done && \
+    echo "$PYTHON_LIB/gen" > $PYTHON_LIB/dist-packages/magma_gen.pth && \
+    find $PYTHON_LIB/gen -type d | tail -n +2 | sed '/__pycache__/d' | xargs -I % touch "%/__init__.py"
```
jheidbrink: Have you tried if this works without creating those `__init__.py`s?

Neudrino (approved):

.devcontainer/Dockerfile
```diff
@@ -336,17 +336,30 @@ COPY lte/gateway/deploy/roles/magma/files/patches/aioeventlet_fd_exception.patch
...
     rm -rf lte.egg-info
 
-COPY /orc8r/gateway/python/ ${MAGMA_ROOT}/orc8r/gateway/python/
 WORKDIR ${MAGMA_ROOT}/orc8r/gateway/python/
 RUN pip3 install --quiet --upgrade --no-build-isolation --no-cache-dir --verbose --editable .[dev] && \
     rm -rf orc8r.egg-info
 
+#### protos
+ARG PYTHON_LIB=/usr/local/lib/python3.8
+
```
themarwhal, line 352: This is orthogonal to the Bazel work right? But to enable python dev work with the existing build system in the devcontainer?

alexzurbonsen: Yes, that would be the idea for this one as well.

.devcontainer/Dockerfile
alexzurbonsen: Yes, I tried. It didn't work for some of the files. I would have to dig deeper to understand why. But it might be worth it.

alexzurbonsen: Ok, couldn't reproduce the error. I think it works as well. `module` in `from package_name import module` is found and the IDE forwards you to the code if you wish.  But it doesn't understand how to navigate to `package_name`. That might be because it is a namespace package and it is missing the `__init__.py`  to forward you to, not sure.

.devcontainer/Dockerfile
alexzurbonsen: Yes, works. Located the relevant sources, as far as i can see.

.devcontainer/Dockerfile
alexzurbonsen: An example from my VSCode:

Screenshot 2022-02-03 at 18 28 20

jheidbrink (approved):

.devcontainer/Dockerfile
```diff
@@ -347,6 +347,21 @@ WORKDIR ${MAGMA_ROOT}/orc8r/gateway/python/
...
+COPY /protos/ ${MAGMA_ROOT}/protos/
+COPY /lte/protos/ ${MAGMA_ROOT}/lte/protos/
+COPY /orc8r/protos/ ${MAGMA_ROOT}/orc8r/protos/
+WORKDIR $MAGMA_ROOT
+RUN pip install mypy-protobuf && \
+    mkdir $PYTHON_LIB/gen && \
+    for PROTO_SRC in orc8r lte feg; \
```
themarwhal: we seem to only copy the lte and orc8r directories above? Is that intentional?

alexzurbonsen: No, thanks for noticing! I checked again and saw that there is also a new target for proto generation, `dp`. Included it as well.

themarwhal (approved):

PR #11422, opened 2022-02-04
feat(agw): Add initial UE stats tracking to eBPF

koolzz:

## Summary

 - Add IMSI to dl map
 - Add basic byte counting to ebpf UL/DL

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/gtpv1-u/ebpf_dl_map.h
+++ b/lte/gateway/c/core/oai/tasks/gtpv1-u/ebpf_dl_map.h
@@ -23,6 +23,8 @@
 struct bpf_map_val {
   uint32_t ip;
   uint32_t tei;
+  uint64_t bytes;
+  uint8_t user_data[64];
 };
 
 int get_map_fd() {
@@ -30,9 +32,11 @@ int get_map_fd() {
 }
 
 void add_ebpf_dl_map_entry(
-    int hash_fd, struct in_addr ue, struct in_addr enb, uint32_t o_tei) {
-  struct bpf_map_val val = {htonl(enb.s_addr), o_tei};
-  uint32_t nkey          = htonl(ue.s_addr);
+    int hash_fd, struct in_addr ue, struct in_addr enb, uint32_t o_tei,
+    Imsi_t imsi) {
+  struct bpf_map_val val = {htonl(enb.s_addr), o_tei, 0, {}};
+  memcpy(val.user_data, imsi.digit, sizeof(imsi.digit));
+  uint32_t nkey = htonl(ue.s_addr);
   bpf_map_update_elem(hash_fd, &nkey, &val, 0);
 }
 
--- a/lte/gateway/python/magma/pipelined/ebpf/ebpf_dl_handler.c
+++ b/lte/gateway/python/magma/pipelined/ebpf/ebpf_dl_handler.c
@@ -26,6 +26,8 @@ struct dl_map_key {
 struct dl_map_info {
   u32 remote_ipv4;
   u32 tunnel_id;
+  u64 bytes;
+  u8 user_data[64];
 };
 
 // The map is pinned so that it can be accessed by pipelined or debugging tool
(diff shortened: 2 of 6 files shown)


ardzoht (approved):
lgtm

lte/gateway/python/magma/pipelined/ebpf/ebpf_dl_handler.c
```diff
@@ -91,5 +83,7 @@ int gtpu_egress_handler(struct __sk_buff* skb) {
     return TC_ACT_OK;
   }
 
+  fwd->bytes += skb->len;
+
```
pshelar, line 88: we need lock to account this correctly, can you add TODO here

lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
```diff
@@ -268,16 +280,19 @@ def print_dl_map(self):
             ue_ip = self._unpack_ip(k.ue_ip)
             remote_ipv4 = self._unpack_ip(v.remote_ipv4)
             tunnel_id = socket.ntohl(v.tunnel_id)
+            imsi = self._unpack_imsi(v.imsi)
+            bytes = v.bytes
 
             print(
```
ardzoht, line 286: shouldn't these be logging instead of print?

lte/gateway/python/magma/pipelined/tests/test_ebpf_dl_dp.py
```diff
@@ -126,7 +128,7 @@ def testEbpfDlFrw1(self):
         cls.sendPacket(cls.inner_src_ip, cls.inner_dst_ip)
         self.assertEqual(len(cls.packet_cap1), 0)
 
-        cls.ebpf_man.add_dl_entry(cls.inner_dst_ip, cls.gtp_pkt_dst, cls.gtp_tunnel_id)
+        cls.ebpf_man.add_dl_entry(cls.inner_dst_ip, cls.gtp_pkt_dst, cls.gtp_tunnel_id, imsi)
```
ardzoht: legit warning?

lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
koolzz: they're print for now as we mostly use the script directly for debugging

lte/gateway/c/core/oai/tasks/gtpv1-u/ebpf_dl_map.h
```diff
@@ -23,15 +23,18 @@
 struct bpf_map_val {
   uint32_t ip;
   uint32_t tei;
+  uint8_t imsi[16];
```
pshelar: Lets not add 3GPP terminology in eBPF, we can define this as binary blob that can be used in user-space to store data.

koolzz: sure should I just rename this to something like?
```
 uint8_t user_data[16];
```

pshelar: lets keep it at the end of this struct and extend it to 64 bytes.

lte/gateway/c/core/oai/tasks/gtpv1-u/ebpf_dl_map.h
```diff
@@ -23,15 +23,18 @@
...
+  uint64_t bytes;
```
pshelar, line 26: this shld be 64bit to be same as kernel stats type.

koolzz: not following its already uint64?

lte/gateway/c/core/oai/tasks/gtpv1-u/ebpf_dl_map.h
koolzz: 64 bytes seems like a lot of space, is there any concern that we're wasting unnecessary memory? IMSI is only 16 bytes, what else are we going to fit there in the future

lte/gateway/c/core/oai/tasks/gtpv1-u/ebpf_dl_map.h
pshelar: ah.. I misread it.

lte/gateway/c/core/oai/tasks/gtpv1-u/ebpf_dl_map.h
pshelar: rule number, version number, direction, etc.... I am pretty sure we will exhaust the 64 bytes, but if you want we can start at 16bytes.

lte/gateway/python/magma/pipelined/ebpf/ebpf_dl_handler.c
```diff
@@ -26,6 +26,8 @@ struct dl_map_key {
 struct dl_map_info {
   u32 remote_ipv4;
   u32 tunnel_id;
+  u64 bytes;
+  u8 user_data[16];
 };
```
pshelar, line 31: We need to define this struct in single header file, its already diverged compared to OAI header definition.

koolzz: yeah I'm exploring this option

pshelar (approved):

PR #11434, opened 2022-02-04
chore(orc8r): Added query params tests for REST API auth

christinewang5:
# Summary
* Added query params tests for REST API auth
* Slightly refactored CreateTestUser so that it takes in user creds and the policy and does everything at once

# Test
* Ran the unit test to check if right


hcgatewood (changes requested):
Cool these are helpful. Can you also add 1 additional case --

- A token has read permissions for network some_network
- GET /deadbeef?network_id=some_network (same network as above)
- Should return deny, and add comment explaining that we're ensuring network_id is only pulled from the path params, not the query params

hcgatewood (approved):

PR #11438, opened 2022-02-06
fix(subscriberdb): Handling sqlite lock failure

panyogesh:
Fix:
  1. Handling the OperationalError case for sqlite3
  2. Dumping the process holding the lock
  3. Return the Server too busy error message to AMF in case  if db is locked

 Test:
         1. Simulated the issue with
             connection.isolation_level = 'EXCLUSIVE'
             connection.execute('BEGIN EXCLUSIVE')
          2. Tested with the lock and without lock on simulator

 Logs:
       1. For process holding the lock
           subscriberdb[2785574]:
              root      2784266 F.... (root)python3
         subscriberdb[2785574]:
             root      2785490 F.... (root)python3.8

      Feb  6 19:44:58 magma-dev-focal subscriberdb[2784266]:
         ERROR:root:Sqlite3 DB is locked for 222456000000121:
         IMSI222456000000121

## Additional Information


panyogesh:
[Test-Logs-Subscriberdb-Lock-Feb-7.txt](https://github.com/magma/magma/files/8011179/Test-Logs-Subscriberdb-Lock-Feb-7.txt)

panyogesh:
7/23 Test  #7: test_itti ..................................   Passed    4.55 sec
      Start  8: test_amf_app

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -22,7 +22,12 @@
 from magma.subscriberdb.sid import SIDUtils
 from orc8r.protos.digest_pb2 import Digest, LeafDigest
 
-from .base import BaseStore, DuplicateSubscriberError, SubscriberNotFoundError
+from .base import (
```
ardzoht: Let's use absolute imports for all python code

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -22,7 +22,12 @@
...
+    BaseStore,
+    DuplicateSubscriberError,
+    SubscriberNotFoundError,
+    SubscriberServerTooBusy,
+)
 from .onready import OnDataReady, OnDigestsReady
```
ardzoht: can you also update this import to absolute method?

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -264,6 +269,14 @@ def get_subscriber_data(self, subscriber_id):
                 row = res.fetchone()
                 if not row:
                     raise SubscriberNotFoundError(subscriber_id)
+        except sqlite3.OperationalError:
+            # Print the process holding the lock
+            db_parts = db_location.split(":", 1)
+            if (len(db_parts) == 2) and db_parts[1]:
+                path_str = db_parts[1].split("?")
+                import os
```
ardzoht: I don't think there's much benefits into using conditional imports here, unless we have any explicit evidence that there could be performance impact, let's keep this import on the top of the file

lte/gateway/python/magma/subscriberdb/store/sqlite.py
```diff
@@ -264,6 +269,14 @@ def get_subscriber_data(self, subscriber_id):
...
+                os.system('fuser -uv ' + path_str[0])
```
ardzoht: update this to `subprocess.call()` or `Popen()`, retrieve the stdout and then log it using `logging` library

ardzoht (changes requested):

panyogesh:
[Test-Logs-Subscriberdb-Lock-Feb-8.txt](https://github.com/magma/magma/files/8022623/Test-Logs-Subscriberdb-Lock-Feb-8.txt)

Attaching the updated logs.

lte/gateway/python/magma/subscriberdb/store/sqlite.py
panyogesh: Comments addressed.

panyogesh: Comments addressed.

lte/gateway/python/magma/subscriberdb/store/sqlite.py
panyogesh: Comments addressed.

lte/gateway/python/magma/subscriberdb/store/sqlite.py
panyogesh: Comments addressed.

lte/gateway/python/magma/subscriberdb/store/sqlite.py
panyogesh: Comments addressed.

ardzoht:
@panyogesh How frequent is the issue happening on your testing? Also does this happen with reaching a certain amount of attached subscribers / how does the load look like?

ardzoht (dismissed):
lgtm, left some questions regarding the failure itself

ardzoht (changes requested):
@panyogesh How frequent is the issue happening on your testing? If this is happening really frequently have we tested the performance impact of dropping running processes over time?

panyogesh:
> @panyogesh How frequent is the issue happening on your testing? If this is happening really frequently have we tested the performance impact of dropping running processes over time?

@ardzoht : Interestingly this issue is coming only on one setup from TerravM which is the regression setup. There is no fixed pattern but on an average 2 testcases fails out of 95 cases and these are random failures.  Same script when executed on other TVM setups or in scale test setups (where we run continuous registration/deregistration for 4-5 hours) this issue is not seen. So far we have not done any performance impact as in other setups issue is not reproduced.

ardzoht (approved):
lgtm

PR #11440, opened 2022-02-07
fix(amf): Unit test case fix of stateless amf

Rajeshs23:
Signed-off-by: Rajeshs23 <rajesh.sure@wavelabs.ai>

    fix(amf): Unit test case fix of stateless amf
## Summary
Since Stateless feature PR and IPV6 PR are merged parallel there is test case error in stateless test case where we need to specify the pdu type which previously not needed. So this lead to the error. It is modified and given specific type. Now this is resolved and test cases are passing.

## Test Plan

- Verified UT cases

## Additional Information
Error occurred:
testcasesfail
Unit Test cases passing after resolving:
testcasepass


ardzoht (approved):
lgtm

ardzoht:
Merging to unblock GH unit test job failures

PR #11442, opened 2022-02-07
feat(ci): Set token permissions to read only

spikey979:

feat(ci): Set token permissions to read only

## Summary

The OpenSSF Scorecard flagged a number of Dangerous-Workflow items. This check "Determines if the project's GitHub Action workflows avoid dangerous patterns." For detailed background see https://github.com/ossf/scorecard/blob/5d3f19838078ad86630f1f80c5ad051249594a1a/docs/checks.md#token-permissions
There were 42 instances of non read-only tokens detected in GitHub workflows. [See token-permissions.txt](https://github.com/ospoco/magma-issue-tracker/files/7946382/token-permissions.txt)

In all files I added top level permissions set to read-all. I also used online tool https://app.stepsecurity.io/

## Additional Information
This is for issue: https://github.com/ospoco/magma-issue-tracker/issues/54

### DIFF
--- a/.github/workflows/agw-docker-load-test.yaml
+++ b/.github/workflows/agw-docker-load-test.yaml
@@ -13,6 +13,8 @@ on:  # yamllint disable-line rule:truthy
 
 concurrency: ${{ github.workflow }}
 
+permissions: read-all
+
 jobs:
   docker-load-test:
     name: agw docker load tests
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -21,6 +21,8 @@ env:
   BAZEL_CACHE: bazel-cache
   BAZEL_CACHE_REPO: bazel-cache-repo
 
+permissions: read-all
+
 jobs:
   path_filter:
     runs-on: ubuntu-latest
--- a/.github/workflows/amis-workflow.yml
+++ b/.github/workflows/amis-workflow.yml
@@ -9,6 +9,8 @@ on:  # yamllint disable-line rule:truthy
 env:
   MAGMA_VERSION: "1.7.0"
 
+permissions: read-all
+
 jobs:
   publish-amis-to-marketplace:
     name: publish-amis-to-marketplace job
--- a/.github/workflows/autolabel-pullrequests.yml
+++ b/.github/workflows/autolabel-pullrequests.yml
@@ -29,6 +29,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
+permissions: read-all
+
 jobs:
   AutoLabelPR:
     runs-on: ubuntu-latest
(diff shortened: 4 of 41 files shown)


electronjoe (changes requested):
To my eye, this PR may be doing a couple different things (and it might be easier to review / roll back / etc if they were separated).

One large function of this PR appears to be the pinning of ~all GitHub Action external workflow references to a particular git hash.  I suspect this is more secure (preventing upstream tampering of what v2 means) - but the PR description as it stands now does not discuss this large set of changes specifically (and they seem not necessary for the PR topic at hand).

Can we split these out into their own PR with a description / rationale?

varunsh-coder:
Happy to see you used https://app.stepsecurity.io/. The code for that is at https://github.com/step-security/secure-workflows. Please do create an issue if you see areas of improvement. Also, I am working on automating scorecard requirements, so if you have pain points etc, please let me know. Thanks!

spikey979:
@electronjoe, yes you're right. I made the changes you suggested, so this PR only contains the changes described here.

electronjoe (approved):

spikey979:
I don't think error "undefined: os.ReadFile / os.WriteFile" has anything to do with the changes I made. I haven't changed file magma/src/go/capture/gen/main.go and everything seems ok to me in the file.

themarwhal:
> 

@mstre123 I seem to also see this error when running locally inside the Magma VM. Any clue why it is failing?
@quentinDERORY Please review.

mstre123:
> > 
> 
> @mstre123 I seem to also see this error when running locally inside the Magma VM. Any clue why it is failing? @quentinDERORY Please review.

looks like the error is being thrown here
https://github.com/magma/magma/blob/master/src/go/capture/gen/main.go#L100
so that is trying to read https://github.com/magma/magma/blob/master/src/go/capture/gen/main.go#L40
"/home/vagrant/magma/lte/gateway/python/integ_tests/defs.mk"
which is still on the branch
https://github.com/spikey979/magma/blob/token_permissions/lte/gateway/python/integ_tests/defs.mk

i wonder if the permission change in this diff breaks how we do os reads and writes.

mstre123:
oh actually took a closer look 
```
note: module requires Go 1.17
```

spikey979:
@mstre123,  regarding the error in file src/go/capture/gen/main.go -> undefined: os.ReadFile / os.WriteFile…
The test that failed referred to the .github/workflows/golang-build-test.yml file, so I first removed the changes I made in that file, however the test still failed with the same error.
After that, in file src/go/capture/gen/main.go I replaced os.WriteFile and os.ReadFile commands with ioutil.WriteFile and ioutil.ReadFile commands and now the test pass.
According to the Go documentation, commands os.ReadFile / os.WriteFile should be identical to ioutil.WriteFile / ioutil.ReadFile, so I have no idea why the os. commands failed the test.

themarwhal:
@tmdzk / @quentinDERORY ?

quentinDERORY (approved):

PR #11448, opened 2022-02-07
chore(agw): print version mismatch message only once for magmad process update

ankitMeta1:
## Summary

print version mismatch message only once for magmad process update
 
## Test Plan

make test_python_service MAGMA_SERVICE

### DIFF
--- a/orc8r/gateway/python/magma/magmad/config_manager.py
+++ b/orc8r/gateway/python/magma/magmad/config_manager.py
@@ -175,16 +175,16 @@ class ConfigManager(StreamerClient.Callback):
             if srv in mconfig.configs_by_key:
                 configs_by_key[srv] = mconfig.configs_by_key.get(srv)
 
-            agw_version = self._magmad_service.version
-            unpacked_mconfig = mconfig.configs_by_key.get(MAGMAD)
-            version_info = self._parse_versions_and_log_warning(agw_version, unpacked_mconfig)
-            agw_version_parsed = version_info.agw_version
-            orc8r_version_parsed = version_info.orc8r_version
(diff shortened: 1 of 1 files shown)


uri200 (approved):

themarwhal:
Re-running the agw-workflow to make it pass (see https://github.com/magma/magma/issues/11362)

themarwhal:
Enabling automerge so that it'll merge when the agw-workflow hopefully passes.

orc8r/gateway/python/magma/magmad/config_manager.py
```diff
@@ -175,16 +175,16 @@ def did_mconfig_change(serv_name):
...
+        unpacked_mconfig = mconfig.configs_by_key.get(MAGMAD)
+        version_info = self._parse_versions_and_log_warning(agw_version, unpacked_mconfig)
+        agw_version_parsed = version_info.agw_version
+        orc8r_version_parsed = version_info.orc8r_version
+
+        if agw_version_parsed and orc8r_version_parsed:
+            agw_minor = int(agw_version_parsed.group('minor_version'))
+            orc8r_minor = int(orc8r_version_parsed.group('minor_version'))
+            if agw_minor - orc8r_minor <= -1:
+                logging.warning("Gateway is more than one minor version behind orc8r. Please consider updating it.")
```
hcgatewood, line 188: Unless I'm misreading, this logic doesn't account for whether the major versions are equal -- e.g. Orc8r 2.3 and AGW 1.7 would not output this warning, while Orc8r 2.9 and AGW 1.7 would. This feels like an unexpected behavior

Let me know if I'm misreading, thoughts?

PR #11450, opened 2022-02-07
fix: modify poll interval to check for new version in magmad

uri200:

## Summary

Magmad is getting clogged with lots of tasks to be run with asycio. That causes high latencies that are jeopardizing other tasks: for example syn_grpc sometimes takes 10s to resopnd back a messge

This change aims to  increases the poll time to check if version of AGW should be updated. That will leave more cycles on magmad for other tasks

## Additional Information

### DIFF
--- a/orc8r/gateway/python/magma/magmad/metrics.py
+++ b/orc8r/gateway/python/magma/magmad/metrics.py
@@ -24,7 +24,7 @@ from magma.magmad.check.network_check import ping
 from orc8r.protos.mconfig import mconfigs_pb2
 from prometheus_client import Counter, Gauge
 
-POLL_INTERVAL_SECONDS = 10
+POLL_INTERVAL_SECONDS = 100
 
 MAGMAD_PING_STATS = Gauge(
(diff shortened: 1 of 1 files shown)


orc8r/gateway/python/magma/magmad/metrics.py
```diff
@@ -24,7 +24,7 @@
 from orc8r.protos.mconfig import mconfigs_pb2
 from prometheus_client import Counter, Gauge
 
-POLL_INTERVAL_SECONDS = 10
+POLL_INTERVAL_SECONDS = 100
```
themarwhal, line 27: did you mean to commit `orc8r/gateway/python/magma/magmad/tests/_trial_temp/_trial_marker`

themarwhal (approved):

orc8r/gateway/python/magma/magmad/metrics.py
uri200: oops thats wrong 
Thanks!

PR #11451, opened 2022-02-07
fix(pipelined): qos tc class setup.

pshelar:
Setup class before using it.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/qos/common.py
+++ b/lte/gateway/python/magma/pipelined/qos/common.py
@@ -278,9 +278,9 @@ class QosManager(object):
             return
         if self._clean_restart:
             LOG.info("Qos Setup: clean start")
+            self.impl.setup()
             self.impl.destroy()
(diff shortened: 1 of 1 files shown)


lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -278,9 +278,9 @@ def _setupInternal(self):
             return
         if self._clean_restart:
             LOG.info("Qos Setup: clean start")
+            self.impl.setup()
```
koolzz, line 281: setup and then right after destroy?

pshelar: This is to cleanup the tc objects. I could post separate PR to fix the init sequence.

koolzz: sure as long as it works

koolzz (approved):

PR #11455, opened 2022-02-08
chore(nms): Upgrade postcss to v8.4.5

spikey979:

chore(nms): Upgrade postcss to v8.4.5

## Summary

The package postcss before 8.2.13 are vulnerable to Regular Expression Denial of Service (ReDoS) via getAnnotationURL() and loadAnnotation() in lib/previous-map.js. The vulnerable regexes are caused mainly by the sub-pattern /*\s* sourceMappingURL=(.*). 

Package **postcss** is indirectly used in packages: **autoprefixer**, **css-loader**, **postcss-loader** and **postcss-flexbugs-fixes**, so the following upgrades were made:
- autoprefixer v7.1.6 to v10.4.2
- css-loader v1.0.1 to v6.6.0
- postcss-loader v3.0.0 to v6.2.1
- postcss-flexbugs-fixes v4.2.1 to v5.0.2

After changes, postcss package was upgraded to v8.4.5

## Test Plan

Ran in /magma/nms folder:
yarn run eslint
yarn run flow
yarn run test

## Additional Information

This is for dependabot alert: 
https://github.com/magma/magma/security/dependabot/docs/docusaurus/yarn.lock/postcss/open


lucasgonze:
@spikey979 The failure in the CI seems to be meaningful. It is caused by a conflict with mapbox-gl. I worked on this postcss bump in the past and gave up because of the amount of work it would take to reconcile Magma code with current versions of these components. What needs to happen is substantive Javascript. 

My feeling is that the components magma is using are really old and will continue to introduce problems. It's worth the trouble to move to the latest versions of these upstream libraries.

Options:
1. An FE dev needs to take this ticket
2. You or I need to engage at the level of code

lucasgonze:
The CI errors look very familiar:

```
TypeError: this.getOptions is not a function
    at Object.loader (/usr/src/node_modules/postcss-loader/dist/index.js:40:24)
 @ /usr/src/node_modules/mapbox-gl/dist/mapbox-gl.css 2:12-141 9:17-24 13:15-22
```

I looked into this in the past. It calls for active coding rather than a version bump.

PR #11465, opened 2022-02-08
chore(ci): remove circleci

themarwhal:

## Summary
* Removed circleci and .circleci/ as we no longer use CircleCI
* Moved scripts used from circleci to .github/workflows/scripts

Discussion: https://magmacore.slack.com/archives/C01AWMH0EJW/p1641945001007000

## Test Plan
CI

## Additional Information
TODO: remove CircleCI integration from the repository


.github/workflows/backport.yml
```diff
@@ -1,44 +0,0 @@
----
```
quentinDERORY: Why do we want to remove this workflow?

themarwhal: I'm not sure if it's used anymore - happy to add it back if it is. I think now the approach we are taking is that the original PR author is responsible for raising a separate PR to previous branches.

themarwhal: cc @tmdzk lmk if this is still used!

themarwhal: Actually, I'll just add it back into this PR so that I'm not doing too many things here. But it would still be good to know whether to remove it or keep it for the future.

electronjoe (approved):

amarpad (approved):

hcgatewood (approved):

tmdzk (approved):

PR #11500, opened 2022-02-10
chore: use virtualenv to install external python dependencies and make generated proto files available there

alexzurbonsen:
## Summary

Currently internal and external python dependencies as well as generated proto files are installed into the system python 3 interpreter. Installing dependencies globally should usually be avoided but seemed fine here, since we are doing it inside a container which is itself the environment. It turned out, that this hinders the bazel migration efforts, since they rely on a "clean" system interpreter (any dependency installed into the interpreter used for bazel migration, is taken for granted by bazel and not included as a dependency).

This change installs most python dependencies into a virtual environment (except for aioeventlet for example) and sets the default python interpreter in the DevContainer to be the one of the virtual environment.

Furthermore, the directory `/home/vscode/build/python/lib/python3.8/site-packages` is added as a `python.analysis.extraPath` to enable Pylance to resolve imports regardless of the interpreter that is chosen by the user.

Setting `python.terminal.activateEnvironment` to `true` (which is a default made explicit here) causes VSCode to source the environment in every terminal that is opened while the virtual env interpreter is selected. 

In case another interpreter is selected the user can manually source the environment by using the `magtivate` command.    

## Test Plan

Build and run locally.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -30,6 +30,8 @@ CMD [ "sleep", "infinity" ]
 
 ARG GIT_PROXY
 ARG FEATURES=mme_oai
+ARG PYTHON_VERSION=3.8
+# PYTHON_VERSION must be in sync with "python.defaultInterpreterPath" and "python.analysis.extraPaths" in .devcontainer/devcontainer.json
 ENV MAGMA_ROOT=/workspaces/magma
 ENV BUILD_TYPE=Debug
 ENV C_BUILD=/workspaces/magma/build/c
@@ -70,7 +72,7 @@ RUN echo "Install general purpose packages" && \
         ninja-build \
         perl \
         pkg-config \
-        python3 \
+        python${PYTHON_VERSION} \
         python3-pip \
         python-is-python3 \
         redis-server \
@@ -81,6 +83,7 @@ RUN echo "Install general purpose packages" && \
         tzdata \
         unzip \
         vim \
+        virtualenv=20.0.17-1ubuntu0.4 \
         wget && \
         gem install fpm && \
         update-alternatives --install /usr/bin/clang clang /usr/lib/llvm-11/bin/clang 10 && \
@@ -320,7 +323,14 @@ RUN ldconfig -v
 
 
 ##### Install Python requirements
-RUN pip3 install --quiet --upgrade --no-cache-dir "setuptools==49.6.0"
+
+### create virtualenv
+ARG PYTHON_VENV=/home/vscode/build/python
+ENV PYTHON_VENV_EXECUTABLE=${PYTHON_VENV}/bin/python${PYTHON_VERSION}
+# PYTHON_VENV must by in sync with "python.defaultInterpreterPath", "python.analysis.extraPaths" and magtivate path in "postCreateCommand" in .devcontainer/devcontainer.json
+
+RUN virtualenv --system-site-packages --python=/usr/bin/python${PYTHON_VERSION} ${PYTHON_VENV}
+RUN ${PYTHON_VENV_EXECUTABLE} -m pip install --quiet --upgrade --no-cache-dir "setuptools==49.6.0"
 
 ### install python3-aioeventlet from the magma apt repo
 COPY orc8r/tools/ansible/roles/pkgrepo/files/jfrog.pub /tmp/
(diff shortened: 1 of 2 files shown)


.devcontainer/Dockerfile
```diff
@@ -339,16 +346,21 @@ RUN patch -N -s -f /usr/local/lib/python3.8/dist-packages/aioeventlet.py < /tmp/
 ### install eggs (lte, orc8r)
 COPY /lte/gateway/python/ ${MAGMA_ROOT}/lte/gateway/python/
 WORKDIR ${MAGMA_ROOT}/lte/gateway/python/
-RUN pip3 install --quiet --upgrade --no-build-isolation --no-cache-dir --verbose --editable .[dev] && \
+RUN . $PYTHON_VENV/bin/activate && \
```
Neudrino: I would prefer to use `source` instead of `.` because it is much more explicit and readable.

.devcontainer/Dockerfile
```diff
@@ -320,7 +320,14 @@ RUN ldconfig -v
...
 ##### Install Python requirements
-RUN pip3 install --quiet --upgrade --no-cache-dir "setuptools==49.6.0"
+RUN pip3 install --quiet --upgrade --no-cache-dir "setuptools==49.6.0" "virtualenv==20.0.17"
+# use same virtualenv version that is installed in VM
+# TODO: install virtualenv with apt instead?
+
+### create virtualenv
+ARG PYTHON_VENV=/home/vscode/build/python
+
+RUN virtualenv --system-site-packages --python=/usr/bin/python3.8 $PYTHON_VENV
```
Neudrino: If we use `/usr/bin/python3.8` in multiple places (also below), we should probably make it an `ARG` (or `ENV`, depending...).

.devcontainer/devcontainer.json
```diff
@@ -64,11 +64,13 @@
...
 			}
 		],
+		"python.terminal.activateEnvironment": true, // Activate Python Environment in Terminal created using the Extension.
 		"python.analysis.extraPaths": [
 			"/workspaces/magma/orc8r/gateway/python/",
 			"/workspaces/magma/lte/gateway/python/",
+			"/home/vscode/build/python/lib/python3.8/site-packages",
 		],
-		"python.defaultInterpreterPath": "/usr/bin/python",
+		"python.defaultInterpreterPath": "/home/vscode/build/python/bin/python3.8",
```
Neudrino: If this needs to be in sync with the entry line 382 from file above, we should probably add a comment in both places to indicated this coupling (if there is no chance to enforce it).

.devcontainer/devcontainer.json
```diff
@@ -84,7 +86,7 @@
 	"mounts": [
 		"source=/var/run/docker.sock,target=/var/run/docker-host.sock,type=bind"
 	],
-	"postCreateCommand": "sudo ln -s ${containerWorkspaceFolder}/lte/gateway/configs /etc/magma",
+	"postCreateCommand": "sudo ln -s ${containerWorkspaceFolder}/lte/gateway/configs /etc/magma; echo \"\nalias magtivate='source /home/vscode/build/python/bin/activate'\" >> ~/.bashrc",
```
Neudrino, line 89: Afaiks from [documentation](https://code.visualstudio.com/docs/remote/devcontainerjson-reference), this field is also accepting an json array. Thus we should probably create one and have multiple entries there instead of spaghetti bash. :wink:

.devcontainer/Dockerfile
alexzurbonsen: Unfortunately, during the Docker build stage we don't have a shell which understands `source`. Thus the `.`

.devcontainer/Dockerfile
alexzurbonsen: Couldn't find another use of this explicit string but put `lib/python3.8/gen` in a `GEN_DIR` variable. Hope that also helps.

Neudrino: Maybe we can further consolidate. I would assume that the `python3.8` part is connected somehow.
You would still need to keep that in sync manually. Can we design it in a way, that there cannot happen a manual misconfiguration by changing one place but not another?

Neudrino: The target would be, that we cannot break the setup by forgetting to update some string somewhere.

Neudrino: How about
```
ARG PYTHON_VERSION="3.8"
ARG PYTHON_SYSTEM_INTERPETER="/whatever/python${PYTHON_VERSION}"
ARG PYTHON_VENV_PATH="/other/path/python${PYTHON_VERSION}"
ARG ...
```
You could also make the whole statement `python3.8` a variable instead of only the version, of course.

alexzurbonsen: Got it! There also seems to be the possibility to reference environment variables in the devcontainer.json. That might help as well. Looking into it.

.devcontainer/devcontainer.json
alexzurbonsen: I looked into ways to enforce it, but doesn't seem to possible currently. Will add comments.

Some more detail: It seems like passing Docker environment variables to settings like analysis.extraPaths and the like is not possible. They can be referenced by `${containerEnv:SOME_VARIABLE}` but only inside the `remoteEnv` property (see [docs](https://code.visualstudio.com/docs/remote/devcontainerjson-reference#_variables-in-devcontainerjson) here) if I understand correctly. This property makes the variables available in terminals and for VSCode. But that doesn't mean they can be referenced as environment variables inside devcontainer.json.  (It is even not supported to use them in a settings.json, which could have been a possible workaround, see [this issue](https://github.com/microsoft/vscode/issues/2809#issuecomment-1031011012).)

More general: inside settings like `python.defaultInterpreter` only predefined VSCode variables seem to be supported. (But the documentation I saw was not too explicit about this, see for example  [python.defaultInterpreter](https://code.visualstudio.com/docs/python/settings-reference#_general-settings) and [predefined variables](https://code.visualstudio.com/docs/editor/variables-reference).)

Neudrino: Thanks for the detailed research and explanation! :pray: 

Doing such things across different files with different technologies is always a big hassle and often plainly not possible. Expecting something like it was the intention of suggesting simple comments in this case. Thanks for checking better methods anyway. :+1:

.devcontainer/devcontainer.json
alexzurbonsen: Tried it, but only got one command to work via an array. So I kept the single string.

Externalize the code into a shell script would be another, see [multiple postCreateCommands](https://github.com/microsoft/vscode-remote-release/issues/3527), but not something we want at this stage, I suppose.

.devcontainer/Dockerfile
alexzurbonsen: I created an `ARG PYTHON_VERSION="3.8"`. With that we have captured all the variable parts of the paths in variables I suppose (assuming that python will be installed via apt and the virtual environment is created with virtualenv) Let me know what you think.

.devcontainer/Dockerfile
```diff
@@ -320,7 +322,15 @@ RUN ldconfig -v
 
 
 ##### Install Python requirements
-RUN pip3 install --quiet --upgrade --no-cache-dir "setuptools==49.6.0"
+RUN pip3 install --quiet --upgrade --no-cache-dir "setuptools==49.6.0" "virtualenv==20.0.17"
+# use same virtualenv version that is installed in VM
+# TODO: install virtualenv with apt instead?
```
Neudrino: This is probably a one liner change. Could we incorporate that in this PR as well instead of the one-liner comment, please?
`sudo apt install virtualenv=20.0.17-1ubuntu0.4`

.devcontainer/Dockerfile
```diff
@@ -333,40 +343,46 @@ RUN apt-key add /tmp/jfrog.pub && \
...
+    pip3 install --quiet --upgrade --no-build-isolation --no-cache-dir --verbose --editable .[dev] && \
     rm -rf orc8r.egg-info
 
-# install formatter autopep8
-RUN pip install --no-cache-dir autopep8
+### install formatter autopep8
+RUN . ${PYTHON_VENV}/bin/activate && \
+    pip3 install --no-cache-dir autopep8
+
+# TODO Does virtualenv have to be sourced in every RUN statement or could I just use $PYTHON_VENV/bin/pip3 instead? First option seems cleaner, the second one comes with less code.
```
Neudrino: If the second is possible, this might be preferable in this case and is imho clear enough. I would guess this is easy to test locally.

.devcontainer/devcontainer.json
Neudrino: OK. You might want to add a comment and link to the respective issue than?

.devcontainer/Dockerfile
alexzurbonsen: Do I understand correctly, that it is preferable to install virtualenv with apt in this case? Changing it.

Neudrino: You could install either way. I slightly prefer apt, however pip would also work for me.

Out of scope (of this PR), but interesting:
Is there a special necessity for that specific version of setuptools?
Otherwise, if an previous release would also work, we could also install setuptools with `apt install python3-setuptools=45.2.0-1`.
Advantage would be, that we can remove the complete pip3 install step.

alexzurbonsen: It is the version that is pinned in setup.py files. (Though in the `python.mk` file a comment says that it has to be "newer than 41.0.1" - but not sure if that is still valid.)

Neudrino: The restriction you mentioned (>41.0.1) would be met installing via apt. So this could be worth a try in a separate follow up PR.
But as mentioned before, out of scope here. :wink:

.devcontainer/Dockerfile
alexzurbonsen: Tested it and it is possible. To be on the safe side I am now using `path/to/python -m pip` for which I introduced a variable `PYTHON_VENV_EXECTUABLE` after all ;)

Neudrino (approved):

themarwhal (approved):

PR #11520, opened 2022-02-10
chore(ctraced): ctraced is bazelified

### LINKED ISSUE #11413: bazel: Bazelify Python ctraced
**General Info:**
* the service is located in `orc8r/gateway/python/magma/ctraced`.
* `lte/gateway/python/magma/redirectd/BUILD.bazel` can be used for orientation (https://github.com/magma/magma/pull/11316)

**Acceptance criteria:**
* service can be run via `bazel run //orc8r/gateway/python/magma/ctraced:ctraced`
  * on dev-container: possible errors are understood (can happen because of missing magma context)
  * on vm: service runs without errors
* only src files that are needed are added to target
* only dependencies that are needed are added to target
* transitive dependencies are added to respective transitive libraries
* note: service has no explicit tests

christinewang5:
ran `bazel run //orc8r/gateway/python/magma/ctraced:ctraced`

image

https://github.com/magma/magma/issues/11413

### DIFF
--- a/orc8r/gateway/python/magma/ctraced/BUILD.bazel
+++ b/orc8r/gateway/python/magma/ctraced/BUILD.bazel
@@ -1,0 +1,41 @@
+# Copyright 2022 The Magma Authors.
+
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_python//python:defs.bzl", "py_binary", "py_library")
+
+ORC8R_ROOT = "../../"
+
+py_binary(
+    name = "ctraced",
+    srcs = ["main.py"],
+    imports = [ORC8R_ROOT],
+    # legacy_creat_init = False is required to fix issues in module import, see https://github.com/rules-proto-grpc/rules_proto_grpc/issues/145
+    legacy_create_init = False,
+    main = "main.py",
+    python_version = "PY3",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//orc8r/gateway/python/magma/common:sentry",
+        "//orc8r/gateway/python/magma/common:service",
+        "//orc8r/protos:ctraced_python_grpc",
+    ],
+)
+
+py_library(
+    name = "ctraced_lib",
+    srcs = [
+        "command_builder.py",
+        "rpc_servicer.py",
+        "trace_manager.py",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [],
+)
--- a/orc8r/protos/BUILD.bazel
+++ b/orc8r/protos/BUILD.bazel
@@ -254,6 +254,16 @@ python_proto_library(
 )
 
 proto_library(
+    name = "ctraced_proto",
+    srcs = ["ctraced.proto"],
+)
+
+python_grpc_library(
+    name = "ctraced_python_grpc",
+    protos = [":ctraced_proto"],
+)
+
+proto_library(
     name = "bootstrapper_proto",
     srcs = ["bootstrapper.proto"],
     deps = [


christinewang5:
Hey @nstng! Have you seen this log before when running bazel run for other services?
```
INFO:root:The 'orc8r' distribution was not found and is required by the application
```

orc8r/gateway/python/magma/ctraced/BUILD.bazel
```diff
@@ -0,0 +1,49 @@
...
+]
+
+
+py_binary(
+    name = "ctraced",
+    srcs = SRCS,
+    imports = [
+        LTE_ROOT,
+        ORC8R_ROOT,
+    ],
```
themarwhal: @nstng What is the best way to determine whether only ORC8R_ROOT is needed or both?

orc8r/gateway/python/magma/ctraced/BUILD.bazel
```diff
@@ -0,0 +1,49 @@
...
+    # legacy_creat_init = False is required to fix issues in module import, see https://github.com/rules-proto-grpc/rules_proto_grpc/issues/145
+    legacy_create_init = False,
+    main = "main.py",
+    python_version = "PY3",
+    visibility = ["//visibility:public"],
+    deps = DEPS,
```
themarwhal: let's just embed DEPS and SRCS... (We're planning on cleaning up the rest of services soon)

nstng:
> Hey @nstng! Have you seen this log before when running bazel run for other services?
> 
> ```
> INFO:root:The 'orc8r' distribution was not found and is required by the application
> ```

Hi @christinewang5, yes, your service needs access to an orc8r setup. You can can run your service in the magma-vm and additionally setup the orc8r. See https://docs.magmacore.org/docs/basics/quick_start_guide. In some cases this was causing more service interaction and made additional missing dependencies visible. In your case it looks to me like your service has already started (listening on address 127...). But testing this is in general a good idea.

orc8r/gateway/python/magma/ctraced/BUILD.bazel
nstng: @christinewang5 @themarwhal

In general: if a source file in @MAGMA_ROOT/orc8r/gateway/python is needed, then an orc8r root is needed. Same for lte sources.
More specific: ctraced is an orc8r service so an ORC8R_ROOT will be needed. LTE_ROOT only if there are lte dependencies - and if the they transitive then it is not easy to see.
From logging: in this case if ORC8R_ROOT is missing, you get
`    from magma.common.sentry import sentry_init
ModuleNotFoundError: No module named 'magma'`
This is, bazel can not even find the magma module. As the common sub-module resides in the orc8r, this indicates the missing import. On the other hand, if you add the orc8r root but remove the sentry dependency, you get:
`    from magma.common.sentry import sentry_init
ModuleNotFoundError: No module named 'magma.common.sentry'`
This is, bazel knows about the magma module and is missing a source file.

nstng: @christinewang5 in this case LTE_ROOT is not needed. (for testing this manually, see previous reply)

orc8r/gateway/python/magma/ctraced/BUILD.bazel
```diff
@@ -0,0 +1,49 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@python_deps//:requirements.bzl", "requirement")
```
nstng: `requirement` is not used - the load line can be removed. If you do not see this marked in your ide then an extension might be missing (yellow underscore). I assume "bazel-stack-vscode" provides this feature.

orc8r/gateway/python/magma/ctraced/BUILD.bazel
```diff
@@ -0,0 +1,49 @@
...
+ORC8R_ROOT = "../../"
+
+LTE_ROOT = ORC8R_ROOT + "../../../lte/gateway/python"
+
+DEPS = [
+    "//orc8r/gateway/python/magma/common:sentry",
+    "//orc8r/protos:mconfigs_python_proto",
+    "//orc8r/protos:mconfig_python_proto",
+    "//orc8r/gateway/python/magma/common:service",
+    "//orc8r/protos:ctraced_python_proto",
```
nstng: The proto libraries are a subset of the grpc libraries, i.e., the proto libraries are not needed here - this line can be removed and also the ctraced_python_proto target from the lte/proto package.

orc8r/protos/BUILD.bazel
```diff
@@ -253,6 +253,21 @@ python_proto_library(
     protos = [":identity_proto"],
 )
 
+proto_library(
+    name = "ctraced_proto",
+    srcs = ["ctraced.proto"],
+)
+
+python_proto_library(
```
nstng: See comment above, this target is not needed.

orc8r/gateway/python/magma/ctraced/BUILD.bazel
nstng: (This is always a little misleading, because the first missing dependency is the proto library and if afterwards the grpc library is needed then the proto library becomes obsolete.)

orc8r/gateway/python/magma/ctraced/BUILD.bazel
```diff
@@ -0,0 +1,49 @@
...
+load("@rules_python//python:defs.bzl", "py_binary")
+
+ORC8R_ROOT = "../../"
+
+LTE_ROOT = ORC8R_ROOT + "../../../lte/gateway/python"
+
+DEPS = [
+    "//orc8r/gateway/python/magma/common:sentry",
+    "//orc8r/protos:mconfigs_python_proto",
+    "//orc8r/protos:mconfig_python_proto",
```
nstng: The two dependencies seem to be unnecessary.
`mconfigs_python_proto` is a dependency in `main.py` but seems to be added transitively - we are still unsure how to handle these, because it is not easy to see if a direct dependency is already covered transitively. In most cases we have not added these.
I do not see a usage of `mconfig_python_proto`. Should be removed.

orc8r/gateway/python/magma/ctraced/BUILD.bazel
nstng: And afterwards `bazel run //:buildifier` should be able to sort the entries.

orc8r/gateway/python/magma/ctraced/BUILD.bazel
```diff
@@ -0,0 +1,49 @@
...
+
+SRCS = [
+    "command_builder.py",
+    "main.py",
+    "rpc_servicer.py",
+    "trace_manager.py",
+]
+
+
+py_binary(
```
nstng, line 16: Sorry for not updating the issue - we started to follow a new design pattern. The binary only has the main.py as a source and all other source files as a dependent py_library. See `orc8r/gateway/python/magma/magmad` for an example.

nstng (changes requested):

orc8r/gateway/python/magma/ctraced/BUILD.bazel
christinewang5: Hmm when I run `bazel run //orc8r/gateway/python/magma/ctraced:ctraced` it still asks for LTE_ROOT


image

nstng: Hi, could it be that you just removed the initialization of LTE_ROOT, but not the usage in the deps list?

christinewang5: Ah yes, thanks for that catch

nstng (approved):
Looks good

themarwhal (approved):

PR #11529, opened 2022-02-11
feat(dp): Soft cbsd deletion

jkmar:
## Summary

Previously backend service just deleted cbsds from the database and inserted deregistration request directly.
However this approach was not compliant with dp architecture, since:
- AMC component should be responsible for generating SAS requests
- some component (mainly CC) still needed data after cbsd deleted

The solution was to add special delete flag to cbsd, and let backend just change this flag instead of deleting cbsd directly.
Cbsds marked as deleted would still be deleted, but this time by AMC.

## Test Plan
Each updated package has its own unit tests.
Also integration tests with the domain proxy application were updated.


xbend (approved):
LGTM

WojSad (approved):

PR #11530, opened 2022-02-11
fix(dp): introducing mutable_cbsd entity in swagger definitions

WojSad:

fix(dp): introducing mutable_cbsd entity in swagger definitions

## Summary

there are now 2 types of cbsd entities: cbsd (for GET requests)
and mutable_cbsd (for POST/PUT requests)

## Test Plan

Tests updated


src/go/go.mod
```diff
@@ -23,19 +23,3 @@ require (
 	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.1.0
 	google.golang.org/protobuf v1.27.1
 )
-
```
jkmar: I think we shouldn't commit this

jkmar (approved):
LGTM, just remove rebase with master and remove src/go/go.mod changes from commit

PR #11535, opened 2022-02-11
chore(orc8r): Removed thanos acl variable from terraform

christinewang5:
Removed the thanos variable bc it was causing an issue when deploying to dev and thanos itself is turned off 

image

### DIFF
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/thanos.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/thanos.tf
@@ -15,7 +15,6 @@ resource "aws_s3_bucket" "thanos_object_store_bucket" {
   count = var.thanos_enabled ? 1 : 0
 
   bucket = var.thanos_object_store_bucket_name
-  acl = "private"
   tags = {
     Name = "Thanos Object Store"
   }


arunuke (commented):
The larger question is - do we want to completely remove all Thanos variables (as this fix currently shows) or remove just the acl variable from the thanos terraform file (which, according to the error message, is automatically set based on the applied configuration)

Copying @hcgatewood in case he has a preference on one over the other. the removal of 'acl' appears to be the simpler fix, but removing thanos because it's not used could be our long term direction as well.

hcgatewood (commented):
I could be wrong but I don't think this will fix the problem (?). Feel free to correct me if you validated this

Instead of removing the variables, basically it seems like there's a line in our thanos tf file(s) that is deprecated, and we need to remove that specific line. cc @arunuke per his comment in the Slack 

P.S. next week I'll go through and make sure no one is using this, then I'll remove the whole thing 🙂

hcgatewood:
Right just seeing @arunuke's comment. Yeah we just want to remove the ACL line for now 👍👍

arunuke:
This is some vulcan level mind-meld @hcgatewood :-).

I propose we make a change just to the 'acl' variable in the tf file (ie, completely remove it since it is automatically set) for this PR.

christinewang5:
Hey just deleted only the acl variable. Kind of understood the Slack comment but this makes sense, thank you 😊

arunuke (approved):
LGTM !!

hcgatewood (approved):

ShubhamTatvamasi (approved):

PR #11546, opened 2022-02-12
chore(mme): Add bazel build targets for mme's protos/grpcs

### LINKED ISSUE #11545: Cleanup and commit numerous BUILD.bazel for proto / grpc needs
Add BUILD.bazel labels / support for all necessary protobuf / grpc outputs consumed by the MME build.

electronjoe:
Closes issue #11545.

### DIFF
--- a/feg/protos/BUILD.bazel
+++ b/feg/protos/BUILD.bazel
@@ -10,6 +10,7 @@
 # limitations under the License.
 
 load("@rules_proto//proto:defs.bzl", "proto_library")
+load("@rules_proto_grpc//cpp:defs.bzl", "cpp_grpc_library", "cpp_proto_library")
 load("@rules_proto_grpc//python:defs.bzl", "python_proto_library")
 
 package(default_visibility = ["//visibility:public"])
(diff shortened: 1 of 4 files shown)


themarwhal (approved):

feg/protos/BUILD.bazel
```diff
@@ -25,3 +26,61 @@ python_proto_library(
...
+
+cpp_proto_library(
+    name = "s6a_proxy_cpp_proto",
+    protos = [":s6a_proxy_proto"],
+)
+
+cpp_grpc_library(
+    name = "s6a_proxy_cpp_grpc",
+    protos = [":s6a_proxy_proto"],
+    deps = [":s6a_proxy_cpp_proto"],
```
nstng, line 43: For Python this dependency is not necessary - the grpc rule creates proto and grpc libraries. Is this needed for cpp? (Im an not sure though if `s6a_proxy_cpp_proto` is pulled from the cache if it was already compiled.)

feg/protos/BUILD.bazel
```diff
@@ -25,3 +26,61 @@ python_proto_library(
...
+proto_library(
+    name = "csfb_proto",
+    srcs = ["csfb.proto"],
+    deps = ["//orc8r/protos:common_proto"],
+)
+
+cpp_proto_library(
+    name = "csfb_cpp_proto",
+    protos = [":csfb_proto"],
+    deps = ["//orc8r/protos:common_cpp_proto"],
```
nstng, line 55: Similar to question above: for Python this dependency is not needed (but, yes, we have some targets where this pattern is currently used). My understanding: the dependency in csfb_proto suffices, .i.e., in this case protoc will compile also common.proto.

PR #11553, opened 2022-02-12
chore(mme): Removes un-necessary HAS_SCTPD check

### LINKED ISSUE #11552: Remove un-necessary compiler define and check of HAVE_LIBSCTP

electronjoe:
Closes #11552.

Additional preprocessor based defines in the MME make builds
environmentally dependent in complex ways, makes maintenance of build
harder, and in this case does nothing to ensure `libsctp` is actually
installed.

## Test plan

CI tests which include `make build_oai` and `make test_oai`.


ssanadhya:
@electronjoe , is the assumption here that dev environments (virtual machine/containers config) should ensure that dependent packages are installed? 
The release build will include the dependencies in the package information itself.

electronjoe:
> @electronjoe , is the assumption here that dev environments (virtual machine/containers config) should ensure that dependent packages are installed?
> The release build will include the dependencies in the package information itself.

As the implementation in Magma stands right now (for libsctpd), there is actually no tie between the dependent package being present and compilation success. We straight up set the HAS_SCTPD check flag in `CMakeLists.txt` and then we check it in compile time (so it has no effect).

There are a few options if we want to do something about dependencies being present.

1) environmental check that the shared lib is present at compile time (no guarantee for runtime location though). Existing CMake code base does not do this. Our Bazel build prototype actually does validate presence.
or
2) Execution of unit tests that depend upon the library loading dynamically (and ensure success of unit tests).
or
3) Statically compile dependencies into our binary (my preference and we are doing this for some deps, but may be hard to do for all.

3) has the added benefit of making runtime environment simpler to manage across containers, VMs, etc as we no longer depend upon local setup.

LMK if that makes sense @ssanadhya!

ssanadhya:
> > @electronjoe , is the assumption here that dev environments (virtual machine/containers config) should ensure that dependent packages are installed?
> > The release build will include the dependencies in the package information itself.
> 
> As the implementation in Magma stands right now (for libsctpd), there is actually no tie between the dependent package being present and compilation success. We straight up set the HAS_SCTPD check flag in `CMakeLists.txt` and then we check it in compile time (so it has no effect).
> 
> There are a few options if we want to do something about dependencies being present.
> 
> 1. environmental check that the shared lib is present at compile time (no guarantee for runtime location though). Existing CMake code base does not do this. Our Bazel build prototype actually does validate presence.
>    or
> 2. Execution of unit tests that depend upon the library loading dynamically (and ensure success of unit tests).
>    or
> 3. Statically compile dependencies into our binary (my preference and we are doing this for some deps, but may be hard to do for all.
> 4. has the added benefit of making runtime environment simpler to manage across containers, VMs, etc as we no longer depend upon local setup.
> 
> LMK if that makes sense @ssanadhya!

Thanks for the clarification @electronjoe ! Approving this PR as the existing checks are not working as intended. 

For the alternate approaches, I think option 1 gives us a first line of defense (compile-time), which is nice and then integration tests will give us run-time check, so we should be good.

Option 3 can inflate the size of our binary, no? Further, as you said, it may not be straightforward to do for all dependencies.

I believe option 4 is just a follow-up on option 3.

ssanadhya (approved):
LGTM

electronjoe:
> I believe option 4 is just a follow-up on option 3.

Oops sorry! You caught my error. I meant to say "Option three ..." there was not supposed to be an option four.

Issue #4868, opened 2021-02-10
Correct includes in AGW c/c++ codebase (remove unused, insert undeclared)


electronjoe:
This can be addressed by careful / repeated application of [include-what-you-use](https://include-what-you-use.org/).

Un-necessary includes have multiple negative effects including:

- Un-necessarily increases scope of compilation for changes
- Introduces cyclical dependency chains that hinder some analysis and build toolchains
- Make tracing dependencies harder for new developers

And missing includes:

- Cause unexpected build failures if and when the compilation includes change
- Make it unclear what dependencies exist between compilation units

electronjoe:
Just noting that in addition to command line tools, there is a [vs-code plugin](https://marketplace.visualstudio.com/items?itemName=pokowaka.pokowaka-iwyu) for iwyu. Need to determine if they are equivalent or not.

electronjoe:
Also note the following online resources for setting this up in CMake / Bazel:

- CMake setup question on [Stack-Overflow](https://stackoverflow.com/questions/30951492/how-to-use-the-tool-include-what-you-use-together-with-cmake-to-detect-unused-he)
- [Bazel rule](https://github.com/erenon/bazel_clang_tidy)

electronjoe:
As further motivation for this effort - In attempting to build in a new docker environment (using existing Dockerfile) - I'm finding multiple header files that are missing critical includes and won't compile. I'm fixing them up now - the ones discovered by this build - but clearly there is a problem here that is sensitive to build environment in a negative way.

electronjoe:
I have a prototype PR that is achieving IWYU.  You can see the (enormous) example output in [This](https://gist.github.com/electronjoe/78ad1fec672d4fbe5f442e0b38be7d13) GH Gist.

I followed instructions [Here](https://github.com/include-what-you-use/include-what-you-use) after getting Clang building our Repo (Prototype work for Issue #4865 ).

Some thoughts:

- Findings seem to be ?exclusively? for cpp files?
  - I thought c was also included... Investigate
- IWYU is analyzing the outputs of protoc
  - I could not find a way to mask these (yet)
  - We obviously won't be fixing up these
- If we find the // WHY INCLUDED comments annoying (I like them) we can disable
- IWYU offers a python script to auto-apply findings
  - Initially I think this is not feasible / too many breakages
  - But possibly if applied to a smaller sub-directory? Explore.

electronjoe:
[...]

PR #11569, opened 2022-02-13
chore(mme): Delete dead code impl of int_to_hex_string

### LINKED ISSUE #11568: Verify and delete dead code in itti_msg_to_proto_msg.cpp

electronjoe:
Fixes #11568.

Discovered during Bazel Build effort of MME.

## Test Plan

CI tests which include `make build_oai` and `make test_oai`.

### DIFF
--- a/lte/gateway/c/core/oai/lib/sms_orc8r_client/itti_msg_to_proto_msg.cpp
+++ b/lte/gateway/c/core/oai/lib/sms_orc8r_client/itti_msg_to_proto_msg.cpp
@@ -16,8 +16,6 @@
  */
 
 #include <iomanip>
-#include <iostream>
-#include <string>
 
 #include "lte/protos/mconfig/mconfigs.pb.h"
 #include "orc8r/gateway/c/common/config/includes/MConfigLoader.h"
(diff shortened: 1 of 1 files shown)


ajahl (commented):
https://github.com/magma/magma/blob/1530cdf8bd25370873947da342b09356b7d5fb4e/lte/gateway/c/core/oai/lib/sms_orc8r_client/itti_msg_to_proto_msg.cpp#L19-L20

I guess, there is no need to include <iostream> and  <string> anymore.
Testet with Codespaces.

Neudrino:
> https://github.com/magma/magma/blob/1530cdf8bd25370873947da342b09356b7d5fb4e/lte/gateway/c/core/oai/lib/sms_orc8r_client/itti_msg_to_proto_msg.cpp#L19-L20
> 
> I guess, there is no need to include and anymore. Testet with Codespaces.

I can confirm that the PR including the suggested change by @ajahl builds and test runs in the VM for me.

electronjoe:
> https://github.com/magma/magma/blob/1530cdf8bd25370873947da342b09356b7d5fb4e/lte/gateway/c/core/oai/lib/sms_orc8r_client/itti_msg_to_proto_msg.cpp#L19-L20
> 
> I guess, there is no need to include and anymore. Testet with Codespaces.

Thanks for noticing the un-necessary includes!

I really want to get IncludeWhatYouUse integration working for us (#4868)... we plan to do so and it would have made Bazel work much easier (in fact it still will make bazel work much easier - as we work to break cycles).

electronjoe:
> gateway/c/core/oai/lib/sms_orc8r_client/itti_msg_to_proto_msg.cpp

Fix applied!

ajahl (approved):

ssanadhya (approved):
LGTM

PR #11604, opened 2022-02-15
chore: cleanup asn1c.bzl

themarwhal:

## Summary
1. linkstatic=True is needed to force `cc_asn1_library` to be always linked statically. (See https://docs.bazel.build/versions/main/be/c-cpp.html#cc_library.linkstatic). I will still have to investigate why dynamic linking is broken, this will be a workaround for now. 
2. found out there is a flag to avoid generating the example file, which removes the need for us to manually remove the file

## Test Plan
CI and tested the patch on @electronjoe MME branch

Added the following in BUILD.bazel and built all
```
diff --git a/BUILD.bazel b/BUILD.bazel
index 68333281e..ebcd2a991 100644
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -11,6 +11,8 @@
 
 load("@bazel_gazelle//:def.bzl", "gazelle")
 load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier", "buildifier_test")
+load("//bazel:asn1c.bzl", "cc_asn1_library")
+
 
 buildifier(
     name = "buildifier",
@@ -70,3 +72,9 @@ filegroup(
 # TODO: Remove when we move proto generation to bazel - this prevents gazelle from causing import issues in the meantime.
 # gazelle:exclude **/*.proto/
 gazelle(name = "gazelle")
+
+cc_asn1_library(
+    name = "asn1_r15",
+    asn1_file = "lte/gateway/c/core/oai/tasks/s1ap/messages/asn1/r15/s1ap-15.6.0.asn1",
+    prefix = "S1ap_",
+)
```

## Additional Information

### DIFF
--- a/bazel/asn1c.bzl
+++ b/bazel/asn1c.bzl
@@ -19,7 +19,7 @@ def _get_dir_name(name):
     return name + ".c"
 
 def _contruct_substitution_commands(ctx, dir_path):
-    """Returns a list of commands that replaces certain strings with another. 
+    """Return a list of commands that replaces certain strings with another. 
 
     The replacement configuration is taken from ctx.attr.substitutions.
     """
@@ -37,33 +37,23 @@ def _contruct_substitution_commands(ctx, dir_path):
     return substitution_commands
 
 def _construct_file_filter_commands(ctx, dir_path):
-    """Returns a list of commands that removes any files that fit the condition described below.
-
-    1. any file that does not end with .c or .h
-    2. any file that is specified in ctx.attr.files_to_remove
-    """
+    """Return a list of commands that removes any files that are not .c or .h"""
     filter_commands = []
 
     # cc_library will complain if there are files that are not one of (.c, .cc, .cpp, .h, ...)
     filter_files_template = "ls -d {dir}/* | grep --invert-match --extended-regexp \'{choose}\' | xargs  --no-run-if-empty rm "
     choose = "**.[.][ch]$"
     filter_commands.append(filter_files_template.format(dir = dir_path, choose = choose))
-
-    # remove any files specified in ctx.attr
-    for file_name in ctx.attr.files_to_remove:
-        filter_commands.append("rm {dir}/{file} || true".format(dir = dir_path, file = file_name))
     return filter_commands
 
 def _construct_asn1c_commands(ctx, dir_path):
-    """Returns a string of command that runs asn1c"""
-    prefix = ctx.attr.prefix
+    """Return a string of command that runs asn1c"""
     flags = ctx.attr.flags
     asn1_file = ctx.attr.asn1_file.files.to_list()[0].path
     asn1c_command_template = "{asn1c} {flags} -D {dir} {input}"
     return [
         asn1c_command_template.format(
             asn1c = ctx.executable._asn1c.path,
-            prefix = prefix,
             flags = flags,
             dir = dir_path,
             input = asn1_file,
@@ -71,7 +61,7 @@ def _construct_asn1c_commands(ctx, dir_path):
     ]
 
 def _gen_with_asn1c_impl(ctx):
-    """generate files by running asn1c
+    """Generate files by running asn1c
 
     Args:
         ctx: passed through Bazel
@@ -112,9 +102,6 @@ def _get_attrs():
             allow_single_file = [".asn1"],
             doc = """The asn file asn1c should use to generate files""",
         ),
-        "files_to_remove": attr.string_list(
-            doc = """A list of file names to remove after code generation""",
-        ),
         "flags": attr.string(
             doc = """Command line flags passed to asn1c""",
         ),
@@ -145,11 +132,10 @@ def cc_asn1_library(
         prefix):
     """Create a CC library of generated asn1 files.
 
-    This library wraps up generated files from these 4 actions:
+    This library wraps up generated files from these 3 actions:
       1. generate .c/.h files by running asn1c with the given input: asn1_file
       2. remove all non .c/.h files
-      3. remove converter-example.c
-      4. apply string substitutions
+      3. apply string substitutions
 
     Args:
         name: the name of rule
@@ -158,7 +144,7 @@ def cc_asn1_library(
     """
     gen_name = name + "_genrule"
 
-    flags = "-pdu=all -fcompound-names -fno-include-deps -gen-PER"
+    flags = "-pdu=all -fcompound-names -fno-include-deps -gen-PER -no-gen-example"
 
     # Taken from https://github.com/magma/magma/blob/14c1cf643a61d576b3d24642e17ed3911d19210d/lte/gateway/c/core/oai/tasks/s1ap/CMakeLists.txt#L35
     # The original PR (PR2707) doesn't give an explanation on why this is necessary.
@@ -173,8 +159,6 @@ def cc_asn1_library(
         flags = flags,
         prefix = prefix,
         substitutions = substitutions,
-        # We want to remove converter-example.c as it does not compile
-        files_to_remove = ["converter-example.c"],
     )
 
     cc_library(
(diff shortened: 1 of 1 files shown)


bazel/asn1c.bzl
```diff
@@ -173,13 +164,13 @@ def cc_asn1_library(
...
-        files_to_remove = ["converter-example.c"],
     )
 
     cc_library(
         name = name,
         srcs = [gen_name],
         # This is needed so that the CCInfo (header/include info) can be used
         deps = [gen_name],
+        # TODO: add a comment here about why this is importatnt
+        linkstatic = True,
```
electronjoe, line 171: I realize this is a draft and you are experimenting! But does this just result in only a static as1n_16.a output? Which then forces google test targets to static link? Is that ~normal? Maybe it is. I thought perhaps the dynamic linking of google test against the code under test was for speed of compilation / linking / execution tests once you have built the target.

electronjoe: I did 7 minutes of google-fu and I cannot find advisement on Bazel / cc_test on the merits of dynamic vs static linking for the target libraries...

themarwhal: @electronjoe  Yes, so this would make it so that .a files are produced for asn library (https://docs.bazel.build/versions/main/be/c-cpp.html#cc_library.linkstatic)

> The linkstatic attribute has a different meaning if used on a [cc_library()](https://docs.bazel.build/versions/main/be/c-cpp.html#cc_library) rule. For a C++ library, linkstatic=True indicates that only static linking is allowed, so no .so will be produced. linkstatic=False does not prevent static libraries from being created. The attribute is meant to control the creation of dynamic libraries.


I am having difficulty understanding why dynamic linking is broken for the asn1 library, so I thought this might be an OK workaround as it only forces static linking for the ASN libraries.

themarwhal: Setting `linkstatic=True` in cc_binary (True by default) or cc_test (False by default) would force all dependencies to be statically linked. Setting `linkstatic=True` in cc_library forces that particular library to be always statically linked. (Which is what you are saying above, I think)

themarwhal: I am still trying to understand why the dynamic linking is broken for the custom rule though... But I haven't made much progress there yet.

electronjoe: @themarwhal and I discussed and (as she already realized, but I just came to realize) forcing linkstatic=True for just asn1c is not nearly such a large penalty as my prototype solution to this issue (which was to set linkstatic=True for all MME tests - which impacts far more libraries / slows things down a lot).

TL;DR; your approach to force static linking for asn1c sounds OK to me at this point.

bazel/asn1c.bzl
```diff
@@ -148,8 +137,7 @@ def cc_asn1_library(
     This library wraps up generated files from these 4 actions:
```
nstng: nitpicking: "3 actions"

nstng:
Cannot command on collapsed lines :/, so I try to do it here :)
```
asn1c_command_template = "{asn1c} {flags} -D {dir} {input}"
    return [
        asn1c_command_template.format(
            asn1c = ctx.executable._asn1c.path,
            prefix = prefix,
            flags = flags,
            dir = dir_path,
            input = asn1_file,
        ),
    ]
```
Prefix seems to be not used in the template - this is imho already done via setting `"ASN1C_PREFIX": ctx.attr.prefix,`

nstng:
Experiment
```
lte/gateway/c/core/oai/tasks/ngap/messages/asn1/r16/BUILD.bazel

load("//bazel:asn1c.bzl", "cc_asn1_library")
cc_asn1_library(
    name = "foo",
    asn1_file = "r16.asn1",
    prefix = "myprefix",
)

bazel build lte/gateway/c/core/oai/tasks/ngap/messages/asn1/r16:foo

/workspaces/magma/lte/gateway/c/core/oai/tasks/ngap/messages/asn1/r16/BUILD.bazel:3:16: Compiling lte/gateway/c/core/oai/tasks/ngap/messages/asn1/r16/foo_genrule.c/converter-example.c failed: (Exit 1): gcc failed: error executing command /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -MD -MF ... (remaining 18 argument(s) skipped)
```
Was the `converter-example.c` removal essential?

With the pre-change state my test target is not failing.

By re-adding the `converter-example.c` removal I can confirm that `libfoo.so` and `libfoo.so-2.params` are not generated when `linkstatic = True` is used.

themarwhal:
> Experiment
> 
> ```
> lte/gateway/c/core/oai/tasks/ngap/messages/asn1/r16/BUILD.bazel
> 
> load("//bazel:asn1c.bzl", "cc_asn1_library")
> cc_asn1_library(
>     name = "foo",
>     asn1_file = "r16.asn1",
>     prefix = "myprefix",
> )
> 
> bazel build lte/gateway/c/core/oai/tasks/ngap/messages/asn1/r16:foo
> 
> /workspaces/magma/lte/gateway/c/core/oai/tasks/ngap/messages/asn1/r16/BUILD.bazel:3:16: Compiling lte/gateway/c/core/oai/tasks/ngap/messages/asn1/r16/foo_genrule.c/converter-example.c failed: (Exit 1): gcc failed: error executing command /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -MD -MF ... (remaining 18 argument(s) skipped)
> ```
> 
> Was the `converter-example.c` removal essential?
> 
> With the pre-change state my test target is not failing.
> 
> By re-adding the `converter-example.c` removal I can confirm that `libfoo.so` and `libfoo.so-2.params` are not generated when `linkstatic = True` is used.

The removal is necessary for compiling (or so I remember), but I did not realize there was an option to not generate them at all in the first place. :)

themarwhal:
> asn1c_command_template

OH! good catch :) I think i was trying to put it here first,but then decided to do it directly with `run_shell`. Will clean this up :)

nstng:
> The removal is necessary for compiling (or so I remember), but I did not realize there was an option to not generate them at all in the first place. :)

With my experiment above the file `converter-example.c` is being generated for me - and building the target fails. What am I missing?

themarwhal:
> > The removal is necessary for compiling (or so I remember), but I did not realize there was an option to not generate them at all in the first place. :)
> 
> With my experiment above the file `converter-example.c` is being generated for me - and building the target fails. What am I missing?

oh I see what you are saying now... I forgot that we deleted the Asn1c library targets from the repository temporarily... I may not have been testing this at all :o will try out a build. (Edit: I had a bad copy and paste between branches :/ fix incoming!)

bazel/asn1c.bzl
```diff
@@ -158,7 +144,7 @@ def cc_asn1_library(
     """
     gen_name = name + "_genrule"
 
-    flags = "-pdu=all -fcompound-names -fno-include-deps -gen-PER"
+    flags = "-pdu=all -fcompound-names -fno-include-deps -gen-PER -no-gen-example"
```
themarwhal, line 147: @nstng This was missing in the previous iteration.

nstng: nice, building my test target is working now for me

nstng (approved):
Reviewed general code changes and tested building a target. LGTM
I can not really comment on the performance part as I do not know how this rule will be used and what the impact of static linking can cause in this context.

themarwhal:
> Reviewed general code changes and tested building a target. LGTM
> I can not really comment on the performance part as I do not know how this rule will be used and what the impact of static linking can cause in this context.

I do not 100% understand this too, but this will only affect `cc_test` as `cc_binary`s default to linking statically. Hopefully if we're only statically linking this library link performance will be not an issue. :)

PR #11613, opened 2022-02-15
chore: Generate Python Swagger models in DevContainer

alexzurbonsen:
## Summary

Generate Swagger models for Python in DevContainer analogous to the make file (swagger target in `orc8r/gateway/python/python.mk`).

## Test Plan

Built and run locally.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -70,6 +70,7 @@ RUN echo "Install general purpose packages" && \
         lldb \
         make \
         ninja-build \
+        openjdk-8-jdk \
         perl \
         pkg-config \
         python${PYTHON_VERSION} \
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -333,40 +344,78 @@ RUN apt-key add /tmp/jfrog.pub && \
...
+
+WORKDIR /var/tmp/
+
+RUN wget http://mirrors.ibiblio.org/apache/maven/maven-3/${MVN_VERSION}/binaries/apache-maven-${MVN_VERSION}-bin.tar.gz && \
+    tar --extract --verbose --file apache-maven-${MVN_VERSION}-bin.tar.gz
+
+RUN git clone --branch ${CODEGEN_VERSION} https://github.com/swagger-api/swagger-codegen /var/tmp/codegen
+
+WORKDIR /var/tmp/codegen
+RUN ../apache-maven-${MVN_VERSION}/bin/mvn clean package -DskipTests
```
Neudrino: Can we avoid the `..`, please? I would see using parent directories as rather unsafe and not a good practice.
Easiest way to do it, might be to swap those lines, as you are in the correct folder before.

.devcontainer/Dockerfile
```diff
@@ -333,40 +344,78 @@ RUN apt-key add /tmp/jfrog.pub && \
...
+    do \
+    mkdir -p ${PYTHON_VENV}/${GEN_DIR}/${SWAGGER_SRC}/swagger/specs && \
+    # Copy swagger specs over to the build directory,
+    # so that eventd can access them at runtime
+    cp /tmp/${SWAGGER_SRC}/swagger/* ${PYTHON_VENV}/${GEN_DIR}/${SWAGGER_SRC}/swagger/specs && \
+    # Generate the files
+    ls ${PYTHON_VENV}/${GEN_DIR}/${SWAGGER_SRC}/swagger/specs \
+    | grep -e ".*\.yml" \
+    | xargs -t -I% /usr/bin/java -jar ${SWAGGER_CODEGEN_JAR} generate \
+    -i ${PYTHON_VENV}/${GEN_DIR}/${SWAGGER_SRC}/swagger/specs/% \
```
Neudrino: This feels somehow so broken to me, that I cannot resist in making a suggestion to improve.
I would guess, glob patterns come to mind, which should do in most cases.
How about
```bash
    ls ${PYTHON_VENV}/${GEN_DIR}/${SWAGGER_SRC}/swagger/specs/*.yml \
    | xargs -t -I% /usr/bin/java -jar ${SWAGGER_CODEGEN_JAR} generate \
    -i % \
```

.devcontainer/Dockerfile
```diff
@@ -333,40 +344,78 @@ RUN apt-key add /tmp/jfrog.pub && \
...
+    -o ${PYTHON_VENV}/${GEN_DIR}/${SWAGGER_SRC}/swagger \
+    -l python \
+    -Dmodels && \
+    # Flatten and clean up directory
+    mv ${PYTHON_VENV}/${GEN_DIR}/${SWAGGER_SRC}/swagger/swagger_client/* ${PYTHON_VENV}/${GEN_DIR}/${SWAGGER_SRC}/swagger/ && \
+    rmdir ${PYTHON_VENV}/${GEN_DIR}/${SWAGGER_SRC}/swagger/swagger_client && \
+    rm -r ${PYTHON_VENV}/${GEN_DIR}/${SWAGGER_SRC}/swagger/test && \
+    rm -r /tmp/${SWAGGER_SRC}/swagger/; \
```
Neudrino: Not sure, what is created here, however, if we only need the content of `swagger_client`, we might reduce the copying a lot by doing something like
```bash
    -o /tmp/${SWAGGER_SRC}/swagger/ \
    -l python \
    -Dmodels && \
    # Flatten and clean up directory
    mv /tmp/${SWAGGER_SRC}/swagger/swagger_client/* ${PYTHON_VENV}/${GEN_DIR}/${SWAGGER_SRC}/swagger/ && \
    rm -r /tmp/${SWAGGER_SRC}/swagger/; \
```

Neudrino (commented):
The block copied over from makefile seems very ugly to me. :confused:
While we might not clean it up completely, we could do some slight improvements, such that it is not jumping at people too much.

.devcontainer/Dockerfile
Neudrino: We could even remove the `/swagger` subdir everywhere in `/tmp`, because I do not see any meaningful usage of it.

.devcontainer/Dockerfile
alexzurbonsen: Thanks. I will look into it!

alexzurbonsen: Had to add another pipe to extract the filename (Because `ls` returns the full path of the file). I didn't find an option in `ls` to avoid this. Another option would be to `cd` into the directory first. But I guess we don't want that either.

.devcontainer/Dockerfile
alexzurbonsen: I chose a slightly different approach: through out the `cp` statement that used to be in L403 within the loop. Since we have a docker `COPY` statement for every item in the loop anyways, I just do the copying there. That gets rid of the /tmp/... directories entirely and also of the `mkdir` statement.

I checked: the swagger generation creates the following directories `swagger_client/models` `docs` and `test`. Apparently we want to keep `models` and `docs`. Thus we delete `test` and move `models` one level up (plus any other files in `swagger_client`, I only checked for the orc8r and there `swagger_client` was empty except for `models`)

Neudrino (approved):

sebathomas (approved):
Tested it on my laptop and the swagger generation works. I can open the files in VS Code after rebuilding the devcontainer with these changes. :+1:

PR #11616, opened 2022-02-15
fix(dp): Do not do any operations for a cbsd when state change is pending

jkmar:
## Summary

Previously AMC ignored pending requests when deleting cbsd.
That behavior could let to race condition in the following scenario:
1. Someone creates cbsd
2. AMC reads state and generates registration request
3. CC reads registration request and sends it to SAS
4. Someone deletes cbsd
5. AMC reads state and deletes cbsd
6. SAS gets processed registration request and registers cbsd
After that cbsd is deleted, but registered in SAS.

The solution was to do the following:
1. RC should send request type of pending requests
2. AMC should not do anything for cbsd if it has pending state change requests
3. CC should send requests for cbsd marked as deleted


xbend (approved):
LGTM

WojSad (approved):

PR #11619, opened 2022-02-15
fix(sessiond):  sessiond searches session based on IPv4 and IPv6

uri200:

## Summary

`aggregate_records` on sessions was looking for sessions with IPv4. This caused IPv6 sessions not to be found

This pr modifies how `find_session` searches when `IMSI_AND_UE_IPV4_OR_IPV6_OR_UPF_TEID` option is selected.

Note that now `SessionSearchCriteria` object has four different keys: IMSI, IPv4, IPv6, TEID. So the session will be found if any of those combinations matches
```
IMSI + IPv4
IMSI + IPv6
IMSI + TEID (for 5G)
```

## Test Plan
```
make sm_precommit
```

Tested in Spirent (phy_06)

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/SessionStore.h
+++ b/lte/gateway/c/session_manager/SessionStore.h
@@ -59,8 +59,9 @@ struct SessionSearchCriteria {
   std::string imsi;
   SessionSearchCriteriaType search_type;
   std::string secondary_key;
+  std::string tertiary_key;
   uint32_t secondary_key_unit32;
-  uint32_t tertiary_key_unit32;
+  uint32_t quaternary_key_unit32;
 
   SessionSearchCriteria(const std::string p_imsi,
                         SessionSearchCriteriaType p_type,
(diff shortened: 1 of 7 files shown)


lte/gateway/c/session_manager/SessionStore.h
```diff
@@ -69,19 +70,21 @@ struct SessionSearchCriteria {
...
                         const std::string p_secondary_key,
-                        const uint32_t tertiary_key_unit32)
+                        const std::string p_tertiary_key,
+                        const uint32_t p_quaternary_key_unit32)
       : imsi(p_imsi),
         search_type(p_type),
         secondary_key(p_secondary_key),
-        tertiary_key_unit32(tertiary_key_unit32) {}
+        tertiary_key(p_tertiary_key),
+        quaternary_key_unit32(p_quaternary_key_unit32) {}
```
uri200, line 87: now this `SessionSearchCriteria` can search by ipv4, ipv6 and TEID separately

themarwhal (approved):

Issue #10297, opened 2021-11-13
fix(orc8r): missing library on macs with M1 chip


lucasgonze:
### Your Environment

- **Version:** latest 
- **Affected Component:** Orchestrator
- **Affected Subcomponent:** ${MAGMA_ROOT}/orc8r/cloud/docker/build.py
- **Deployment Environment:** local build on an OS X machine with an M1 chip

### Describe the Issue

Linker error in build.py in orc8r/cloud/docker on OS X machines with an M1 chip

**To Reproduce**

1. Clone latest magma/magma
2. cd ${MAGMA_ROOT}/orc8r/cloud/docker && ./build.py -t
3. see error `qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory`

**Screenshots**
![image](https://user-images.githubusercontent.com/111581/141599198-cec44b18-401f-45eb-8a30-cd36949648a1.png)

**Additional context**

This ticket on nodejs is a correct diagnosis of the cause: 
[Library used on Linux image not found on M1 ARM64 image](https://github.com/nodejs/help/issues/3239)

The suggested fix is simple - in the Dockerfile, specify the platform flag `FROM --platform=linux/amd64`. However there needs to be a mechanism to vary the Dockerfile for this platform

Issue #11608, opened 2022-02-15
Upgrade MME Docker environment to current Ubuntu LTS version


Neudrino:
## Issue

- See https://github.com/magma/magma/blob/master/lte/gateway/docker/mme/Dockerfile.ubuntu18.04
- Currently Ubuntu 18.04 LTS Bionic Beaver is used.
- The Project migrated to clang-format-11, which is not available for Bionic.
- clang-format-11 is available on Ubuntu 20.04 LTS Focal Fossa.

For details see also https://github.com/magma/magma/pull/11549.

## Solution

- Upgrade the docker image to focal fossa.
- Ensure that all third party dependencies are also working.

themarwhal:
@rdefosse for coordination

shahrahman-fb:
Closing this issue as part of a repository cleanup effort. This issue has had no activity for over 2 years.

If this issue is still relevant, please reopen it or file a new issue with updated information. We are actively maintaining the project again and want to focus on current, actionable issues.

See: #15833 (Phase 1 cleanup), #15828 (tracking issue)

PR #11620, opened 2022-02-15
fix(orc8r): fix orc8r docker build on M1 Macs

emakeev:

## Summary

The change fixes orc8r build on Mac M1 (issue #10297)

## Test Plan

build.py -g; build.py -a ...

## Additional Information

### DIFF
--- a/orc8r/cloud/docker/controller/Dockerfile
+++ b/orc8r/cloud/docker/controller/Dockerfile
@@ -1,7 +1,9 @@
 # ------------------------------------------------------------------------------
 # Base: for tests, precommit, codegen, etc.
 # ------------------------------------------------------------------------------
-FROM ubuntu:xenial as base
+ARG PLATFORM=linux/amd64
+
+FROM --platform=$PLATFORM ubuntu:xenial as base
 
 ENV GO111MODULE on
 ENV GOPATH ${USER}/go


orc8r/cloud/docker/controller/Dockerfile
```diff
@@ -1,7 +1,9 @@
 # ------------------------------------------------------------------------------
 # Base: for tests, precommit, codegen, etc.
 # ------------------------------------------------------------------------------
-FROM ubuntu:xenial as base
+ARG PLATFORM=linux/amd64
+
+FROM --platform=$PLATFORM ubuntu:xenial as base
```
ShubhamTatvamasi, line 6: Do we have any plans to upgrade from **xenial** to **focal**?

Neudrino: https://github.com/magma/magma/issues/11608 is a similar (but yet also different) issue open for MME. In general it strongly depends on the used external dependency, if it is easy to upgrade or not. I might be worthwhile to open a separate issue for this image.

However that my be, an image upgrade is probably not in scope of this PR?
Thoughts, @emakeev ?

ShubhamTatvamasi: You are right @Neudrino. LGTM.

ShubhamTatvamasi (approved):

orc8r/cloud/docker/controller/Dockerfile
emakeev: this PR just unblocks current M1 based builds. There is probably cleaner long term solution, but - I'd leave it to the owners to figure out.

ShubhamTatvamasi: @emakeev I have enabled Auto-merge on this PR. Just rebase it with the master so all tests can pass.

PR #11627, opened 2022-02-16
fix(ctraced): missing dependency is added

LKreutzer:

## Summary

- Missing bazel dependency is added in module ctraced.

## Test Plan

-  Run ctraced: 
   - `bazel run //orc8r/gateway/python/magma/ctraced:ctraced`

## Additional Information

### DIFF
--- a/orc8r/gateway/python/magma/ctraced/BUILD.bazel
+++ b/orc8r/gateway/python/magma/ctraced/BUILD.bazel
@@ -23,9 +23,9 @@ py_binary(
     python_version = "PY3",
     visibility = ["//visibility:public"],
     deps = [
+        ":ctraced_lib",
         "//orc8r/gateway/python/magma/common:sentry",
         "//orc8r/gateway/python/magma/common:service",
-        "//orc8r/protos:ctraced_python_grpc",
     ],
 )
 
@@ -36,6 +36,6 @@ py_library(
         "rpc_servicer.py",
         "trace_manager.py",
     ],
-    visibility = ["//visibility:public"],
-    deps = [],
+    visibility = ["//visibility:private"],
+    deps = ["//orc8r/protos:ctraced_python_grpc"],
 )


nstng (approved):
Somehow missed it in the original pr https://github.com/magma/magma/pull/11520.
Done in pairing.

orc8r/gateway/python/magma/ctraced/BUILD.bazel
```diff
@@ -36,6 +36,6 @@ py_library(
         "rpc_servicer.py",
         "trace_manager.py",
     ],
-    visibility = ["//visibility:public"],
-    deps = [],
+    visibility = ["//visibility:private"],
```
Neudrino, line 39: Just out of curiosity:
Is setting the libraries to private part of the new strategy/philosophy coming with the split of `main.py` and "the rest"?

Neudrino, line 39: Is visibility private not be the default visibility setting, such that this line can be omitted completely?

Neudrino (approved):
Thanks for the quick fix! :pray:

orc8r/gateway/python/magma/ctraced/BUILD.bazel
nstng: No, usually the <service>_lib is used in tests, but here are no tests. This is just set to make it more clear that this lib is only used here. We decided to use this more in the last couple of days. Will be added more regularly in existing places when touched.

nstng: Yes, it is the default and could be omitted. I prefer the better readability - but can be discussed.

Neudrino: Fine for me either way! :+1: 
Not omitting is certainly more explicit. Not sure if an experienced Bazel user needs it though?

However that may be, do we want to define and document a policy, what we see best practice?
Maybe it is worthwhile to start a new section in the [Contribute Conventions Wiki](https://github.com/magma/magma/wiki/Contributing-Code-Conventions#languages) collecting all findings on that part?

Neudrino (approved):

orc8r/gateway/python/magma/ctraced/BUILD.bazel
nstng: ui, did not know this one - will add some lines.

themarwhal (approved):

PR #10953, opened 2021-12-21
docs: migration from Docusaurus to GitHub wiki

Neudrino:
## Summary

- This is part of the effort from https://github.com/magma/magma/issues/9848.
- To avoid having diverging documentation the content was copied over and linked in the appropriate places for a seamless transition.
- Please see **https://github.com/magma/magma/wiki/Contributing-Code** for the migration target.
- While the migration is not completed with this, it is a first valuable step and imho finalizable.

Issue #9848, opened 2021-10-23
[Docs] Restructuring and extending the Magma Contributor Guide


ildikov:
### Your Environment

- **Affected Component:** Documentation

### Describe the Issue

As discussed in issue #9601 the current documentation that provides information that help contributors to contribute to the Magma project needs some enhancements. This bug report is created to capture the target structure of the documentation along with the tasks that are required to get us there.

The desired table of contents for the new Contributor Guide is the following:
Contributor Guide:

- [x] Introduction
- [x] Overview of the Magma governance and community structure - Migrated by @Neudrino to https://github.com/magma/magma/wiki/Overview-of-the-Community-Structure-and-Governance
- [x] Releases - @jaredmullane 
- [x] Communication tools and processes - @electronjoe 
- [x] ~~Overview of the contribution workflow~~ - skipped for now as there are multiple ways to contribute; it can be added to relevant chapters
- [x] Tools overview - Done by @electronjoe at https://github.com/magma/magma/wiki/Tools.
- [x] Accounts setup - @ShubhamTatvamasi 
- [x] Tracking and fixing bugs
    - newly added: https://github.com/magma/magma/wiki/Error-monitoring-and-triaging-workflow
- [x] Adding and tracking proposals - @dcheline 
- [x] Contributing code - @Neudrino 
    - [x] Development environment - @Neudrino 
    - [x] Coding guidelines - @Neudrino 
    - [x] Developing Workflow - @Neudrino 
      - [x] Commit and PR guidelines - @Neudrino 
      - [ ] Testing your code locally - @ShubhamTatvamasi 
    - [x] Code Review Process - @Neudrino 
      - [x] CI/CD, GitHub Checks - @Neudrino 
- [x] Contributing documentation - @jaredmullane 
    - [x] Overview of documentation structure and sources
    - [x] Conventions
    - [x] Versioning documentation
        - [x] Versioning conventions, when a new version is created and who can create it
        - [x] Versioning process (https://docs.magmacore.org/docs/next/docs/docs_overview)

Tasks to complete the changes:

- [x] Create the new structure for the Contributor Guide
  - New structure is created in GH Wiki: https://github.com/magma/magma/wiki/Contributor-Guide 
- [ ] Move content over from other parts of the Docusaurus that belongs to the Contributor Guide: ie: proposal process description
- [x] Outline further tasks to polish the new and existing chapters of the Contributor Guide (one task per chapter unless more granularity is needed)

### Dependencies

[...]

PR #11634, opened 2022-02-16
docs: Migration of contributers guide to Github Wiki

Neudrino:
## Summary

- This is part of the effort from https://github.com/magma/magma/issues/9848.
- This is a follow up to https://github.com/magma/magma/pull/10953.
- To avoid having diverging documentation the content was copied over and linked in the appropriate places for a seamless transition.
- Please see **https://github.com/magma/magma/wiki/Contributing-Code** for the migration target.
- This should about finalize the migration of the topic "Contributing Code".

### DIFF
--- a/docs/readmes/contributing/contribute_onboarding.md
+++ b/docs/readmes/contributing/contribute_onboarding.md
@@ -6,107 +6,11 @@ hide_title: true
 
 # Developer Onboarding
 
-This document walks a new developer through the process of onboarding to the Magma project.
-
-## Project overview
-
-Magma is an open-source software platform that gives network operators an open, flexible, and extendable mobile core network solution. Our mission is to connect the world to a faster network by enabling service providers to build cost-effective and extensible carrier-grade networks.
-
-In more approachable terms, Magma is a collection of software that makes running things like a cellular network affordable and customizable. With Magma's current rate of growth, increasing stability and reliability of the platform are all huge wins.
-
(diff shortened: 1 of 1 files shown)


themarwhal (approved):
Changes look good!

Do we have a strategy on how searching for an entry would work with GitHub Wikis? Is there a search bar?

Also, I think we should do the same for versioned docs too :)

ildikov (approved):
Thank you @Neudrino !

electronjoe (approved):

Neudrino:
> Do we have a strategy on how searching for an entry would work with GitHub Wikis? Is there a search bar?

You can use the Github global search bar with a search restricted to the Repository. On the left hand-side you will than find a category "Wikis".

grafik

Neudrino:
> Also, I think we should do the same for versioned docs too :)

Good point. Maybe we can do a back port, once all people finished their part in the migration?
Or the problem might solve itself, if 1.7 is out and we have a new "default" version of the DocuSaurus?

themarwhal:
> > Do we have a strategy on how searching for an entry would work with GitHub Wikis? Is there a search bar?
> 
> You can use the Github global search bar with a search restricted to the Repository. On the left hand-side you will than find a category "Wikis".
> 
> grafik

ah! didn't realize this was an option :D glad i asked!

themarwhal:
> > Also, I think we should do the same for versioned docs too :)
> 
> Good point. Maybe we can do a back port, once all people finished their part in the migration? Or the problem might solve itself, if 1.7 is out and we have a new "default" version of the DocuSaurus?

Yeah I think we can do a full cleaning sweep once the migration is fully complete? (I would rather have it be gone entirely so that there's no chance for confusion :) )

PR #11641, opened 2022-02-16
fix(dp): fix migrations

jkmar:
## Summary

There where 3 migrations (A, B, C)
But both B and C had A as a previous version which was wrong
and caused migration script to fail.

This commit fixes this issue (now previous version of C is B).


WojSad (approved):
LGTM

xbend (approved):

PR #11652, opened 2022-02-16
refactor(orc8r): Separate split-concern service protos

christinewang5:
separated split-concern servicers to protected and southbound 

## Testing
* ran relevant unit tests


hcgatewood:
To answer your question: we want to split both the proto definition and the implementation

Though with the caveat that, if there's shared impl between the top-level proto implementations, then it's fine to have shared functionality between the two servicers. The main goal is to separate the proto definitions, and move the impl of those definitions to the appropriate `**/servicers/{protected,southbound}`

orc8r/cloud/go/services/metricsd/servicers/protected/servicer.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, label := range metric.Label {
+		if label.GetName() == labelName {
+			label.Value = &labelValue
+			labelAdded = true
+		}
+	}
+	if !labelAdded {
+		metric.Label = append(metric.Label, &prom_proto.LabelPair{Name: strPtr(labelName), Value: &labelValue})
+	}
+}
```
christinewang5: I repeated the implementation of this function across all protected & southbound servers. Couldn't find a good way to share the implementation

christinewang5:
Hey @hcgatewood! Could u let me know if this is on the right track of what u were looking for? 
I essentially split it into two `MetricsControllerSouthboundServer` & `MetricsControllerProtectedServer` and then wrapped the two implementations in`MetricsControllerServer`.

orc8r/protos/metricsd.proto
```diff
@@ -240,10 +240,13 @@ message PushedMetricsContainer {
...
   // Report a collection of metrics from a service
   rpc Collect(MetricsContainer) returns (Void) {}
 
+}
+
+service MetricsControllerProtected {
   // Push a collection of metrics to metricsd
   rpc Push(PushedMetricsContainer) returns (Void) {}
-}
+}
```
hcgatewood: Nit: preserve eof newline

orc8r/cloud/go/services/metricsd/metricsd/main.go
```diff
@@ -48,7 +48,8 @@ func main() {
 	}
 
 	controllerServicer := servicers.NewMetricsControllerServer()
-	protos.RegisterMetricsControllerServer(srv.GrpcServer, controllerServicer)
+	protos.RegisterMetricsControllerSouthboundServer(srv.GrpcServer, controllerServicer.Southbound)
+	protos.RegisterMetricsControllerProtectedServer(srv.GrpcServer, controllerServicer.Protected)
```
hcgatewood: Yeah for reference -- this is the key point here, that we're registering these servicers separately. That way, we'll be able to register them to different GrpcServers when we add that functionality

hcgatewood (changes requested):
@christinewang5 this looks great. Couple small things but otherwise lgtm

1. For the proto naming, let's do `MetricsController` and `CloudMetricsController`, for naming consistency across the codebase, and to practice backwards compatibility (not an issue here, but changing the service name could conceivably cause a breakage)
2. Can just delete `orc8r/cloud/go/services/metricsd/servicers/servicer.go`, as I don't believe there's a need to aggregate the 2 servicers -- can treat them as fully separate
3. For the functionality that you duplicated -- consider making that a public method declared in the southbound servicer's pkg, and importing it into the protected servicer's pkg. Either way is fine but that's prob my preference

hcgatewood (changes requested):
Okay so originally I thought this would be fine for a single big PR, but now I'm thinking it will be better to split it up actually

Let's do the following PRs
1. Split all the servicers (just the proto definitions, don't need to move to southbound/protected yet)
2. Handle the southbound/protected move, one PR per servicer

This will make it easier to review, and the main benefit will be we can convince Git that you just moved the files, rather than rn where it thinks you rewrote a lot of the files

I think you can accomplish these changes without having to throw out this PR -- might have to do a bit of Git maneuvering is all, lmk if have any questions on that and happy to help

hcgatewood (changes requested):
Unless I'm mistaken, I think as part of this split, you'll also need to initialize the cloud/protected version of each servicer in the relevant places where the existing/southbound servicer is initialized

I believe the failing tests will corroborate that assumption, lmk if not

But basically this PR is trying to do "as little as possible" to split up the servicers into the southbound and protected servicers, while ensuring tests still pass. Then the other PRs you've raised will handle actually moving servicers into their appropriate southbound/protected directories

Lmk if that makes sense, happy to chat about it as well

christinewang5:
> Unless I'm mistaken, I think as part of this split, you'll also need to initialize the cloud/protected version of each servicer in the relevant places where the existing/southbound servicer is initialized
> 
> I believe the failing tests will corroborate that assumption, lmk if not
> 
> But basically this PR is trying to do "as little as possible" to split up the servicers into the southbound and protected servicers, while ensuring tests still pass. Then the other PRs you've raised will handle actually moving servicers into their appropriate southbound/protected directories
> 
> Lmk if that makes sense, happy to chat about it as well

I think it is rather difficult to do as little as possible and still make the tests pass. Even if I did initialize the cloud/protected version of each servicer everywhere, I would also have to move the implementations of all the methods too, which I am doing in the subsequent PRs, which I can also do, but idk what the point is in splitting in that case bc the next few PRs will actually just be moving the file to another folder. 

What if I stacked all the PRs and make the last one pass all the tests? Lmk!

feg/cloud/go/services/health/servicers/health_test_service.go
```diff
@@ -1,9 +1,7 @@
 /*
 Copyright 2020 The Magma Authors.
-
```
hcgatewood, line 3: ?

orc8r/cloud/go/services/configurator/servicers/southbound.go
```diff
@@ -41,6 +41,16 @@ func NewSouthboundConfiguratorServicer(factory storage.ConfiguratorStorageFactor
 	return &sbConfiguratorServicer{factory}, nil
 }
 
+type cloudSbConfiguratorServicer struct {
+	factory storage.ConfiguratorStorageFactory
+}
+
+func NewCloudSouthboundConfiguratorServicer(factory storage.ConfiguratorStorageFactory) (cfg_protos.CloudSouthboundConfiguratorServer, error) {
```
hcgatewood: Eh we can leave this one as-is, no need to make it protected -- it's the same data as the GetMconfig, it just happens to only be used internally

orc8r/cloud/go/services/state/servicers/servicer.go
```diff
@@ -225,7 +237,28 @@ func (srv *stateServicer) getStates(_ context.Context, req *protos.GetStatesRequ
...
+		return nil, internalErr(err, "GetStates (get) blobstore get many")
+	}
+
+	err = store.Commit()
+	if err != nil {
+		return nil, internalErr(err, "GetStates (get) blobstore commit transaction")
+	}
+
+	return &protos.GetStatesResponse{States: blobsToStates(blobs)}, nil
+}
```
hcgatewood, line 259: Why adding this in? Unless I'm reading wrong, looks unusued

hcgatewood (commented):
Okay awesome! Looking about right. Couple small things, lmk your thoughts

orc8r/cloud/go/services/state/servicers/servicer.go
christinewang5: used on line 69 of this file

hcgatewood: Ah gotcha. I guess I'm just confused why GitHub thinks this block is new? Weird. But whatever, seems reasonable given my skimming of the original file 👍

orc8r/cloud/go/services/configurator/servicers/southbound.go
christinewang5: Do you mean not to split the proto at all?!

hcgatewood: Correct, we don't need to split the configurator proto because it's already split into norhtbound.proto and southbound.proto (protected and southbound, respectively)

orc8r/cloud/go/services/state/servicers/servicer.go
christinewang5: Oh I just copied the method for both cloudStateServicer & stateServicer. Don't think there is a good way to share the impl of methods on structs, like I can't import one to another...

hcgatewood: Ah gotcha. Okay yeah is fine

hcgatewood (approved):

themarwhal (approved):

PR #11666, opened 2022-02-17
chore: change devcontainer image sha to master

alexzurbonsen:
## Summary

Update DevContainer image sha to `master` (the latest image does not come with a sha, thus switch to `master`).

The change in docker image tagging was activated with https://github.com/magma/magma/pull/11508.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -23,7 +23,7 @@
 		"njpwerner.autodocstring",
 		"zxh404.vscode-proto3",
 	],
-	"image": "ghcr.io/magma/magma/devcontainer:sha-e05d447",
+	"image": "ghcr.io/magma/magma/devcontainer:master",
 	"settings": {
 		"terminal.integrated.shell.linux": "/bin/bash",
 		"files.watcherExclude": {
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -15,7 +15,7 @@ on:  # yamllint disable-line rule:truthy
     # Run four times a day to build bazel cache
     - cron: '0 0,6,12,18 * * *'
 env:
-  DEVCONTAINER_IMAGE: "ghcr.io/magma/magma/devcontainer:sha-206a67a"
+  DEVCONTAINER_IMAGE: "ghcr.io/magma/magma/devcontainer:master"
   BAZEL_CACHE: bazel-cache
   BAZEL_CACHE_REPO: bazel-cache-repo
 


.devcontainer/devcontainer.json
```diff
@@ -23,7 +23,7 @@
 		"njpwerner.autodocstring",
 		"zxh404.vscode-proto3",
 	],
-	"image": "ghcr.io/magma/magma/devcontainer:sha-e05d447",
+	"image": "ghcr.io/magma/magma/devcontainer:master",
```
themarwhal, line 26: How frequently would this value update?

Neudrino: _TLDR:_ Not often (or not at all).

_Long:_
We noticed that with our late interest in devContainer improvements, we need to update the image (sha) quite often to make the improvements available. Also there have been cases, where the Bazel team had additional requirements or was impeded, due to some changes not getting in the "production devcontainer". Thus we wanted to go towards a more rolling release scheme for the devcontainer image for at least as long as we are frequently improving it.
If desired, we can easily go back to a sha based manual approach by reverting #11508, once frequent changes are no longer required.

We might want to discuss some more CI improvements in order to stay with a rolling release scheme for devcontainer, which I would see beneficial though.

themarwhal: @Neudrino If this means that the devcontainer gets rebuilt only when the underlying is updated, then I'm ok with that. (That shouldn't be happening too often anyways).

Neudrino: @themarwhal, sorry, but I am not sure I completely get what you are asking. :sweat_smile: 

A rebuild of the container is defined by [the Github Action](https://github.com/magma/magma/blob/master/.github/workflows/docker-builder-devcontainer.yml), of course. However, this did not change to before. I.e. it rebuilds on schedule every 6 hours as well as on any change made on master to the respective files.
(If the rebuild every 6 hours is sensible is probably debatable but out of scope here.)

If you are asking, when functional changes will occur, than:
- if a reviewed merge to master happens, or
- if the base image changes (assuming, we keep the scheduled rebuild).

Neudrino (approved):

.devcontainer/devcontainer.json
themarwhal: Ah sorry! I should have been more clear. I'm not sure if you've seen this, but if you have a codespace open and rebase ontop of a commit (which is often master) that has a different underlying codespace image or configuration, you get a little pop up that says whether the user would like to rebuild the devcontainer. This is great if the container has actually been updated, but not too great if not. (The rebuild takes about a few minutes). 

So I just wanted to avoid GitHub codespaces rebuilding the dev env everytime a new commit is pushed. 

I do not know this will happen with this change though, just raising my concerns. I also think we probably do not have to rebuild the image every 6 hours. :) So maybe this is a way to counter this too.

themarwhal: cc @Neudrino

Neudrino: Now I understand that it is all about the Github Codespace integration. I probably got through the different scenarios and the expected behaviour (from my understanding).

In general, the GHCR image-stream-tags and the git repository branch names are different entities, which merely can have the same name. I.e. The Magma repo git branch `https://github.com/magma/magma/tree/master/` does not necessarily be in full sync with GHCR image stream tag `ghcr.io/magma/magma/devcontainer:master`.
Of course the are not disconnected, but our Github actions automation is configured in a way, that the code on the Github repo master branch results in an GHCR image with the tag master.

Now what does that mean for VScode?
As far as I know, VScode detects any changes in the `devcontainer.json` (and the Dockerfile and docker-compose.yaml, too, according to documentation) and offers the little pop-up you mentioned to update the codespace. As consequence I would suspect the following behaviour using the master tag.

- Same as before: You do whatever for a (git) operation without changing the above mentioned files (i.e. a rebase on master). Your VScode would not notice a change of the devcontainer and would *not* prompt to rebuild.
- Same as before: You do a change in the above mentioned files, you would be prompted with the pop-up.
- Difference: In the rare case that you would have an image change without changing the files above, you would not immediately get a new devcontainer, but it would only be delivered on the next start of the devcontainer.

So VScode only cares if the configuration changes, not, if the actual image changes. And a rebase on master is a git operation, which does not affect the image-stream-tag master on GHCR at all, afaiks.

> So I just wanted to avoid GitHub codespaces rebuilding the dev env everytime a new commit is pushed.

No, if you do not change the respective files, I would say. This is the same behaviour as before, afaik.

themarwhal: Great. This sounds good to me. cc @electronjoe for visibility

Neudrino: Anyway, if this turns out to deviate from the expectations, we can easily make a small change to revert the behaviour. :wink:

electronjoe: Lets give it a try!

electronjoe (approved):

PR #11667, opened 2022-02-17
fix(orc8r): Try fix Helm deploy bug with numeric version

hcgatewood:
## Summary

Env var was unquoted, which normally is fine, but when we tried to deploy an all-numeric hash, threw a type err. To fix, we'll quote the value to ensure it's always interpreted as a string

## Test Plan

do_it_live.jpg

## Additional Information


hcgatewood:
Force-merging to unblock staging deployment

PR #11669, opened 2022-02-17
chore(sessiond): remove #if 0 block

themarwhal:

## Summary
Removing dead code - @GANESH-WAVELABS is this good to be removed? This was added 8 months ago but was never removed.

## Additional Information


GANESH-WAVELABS:
Thanks Marie! for remembering It has been long did not notice this, we can clean such type of dead code. We will take care of newly introducing code not to happen such dead code.

uri200 (approved):

Issue #9714, opened 2021-10-16
bazel: Bazelify MME


themarwhal:
Issue tracking all blockers to migrate C/C++ for production: https://github.com/magma/magma/issues/10890

# Blocking and Needs Owner
- [x] https://github.com/magma/magma/issues/11638
- [x] https://github.com/magma/magma/issues/11897 
- [x] #12355

# Non-Blocking and Needs Owner
- [x] https://github.com/magma/magma/issues/11936 (blocker for coverage collection in CI)
- [x] https://github.com/magma/magma/issues/12181
- [x] https://github.com/magma/magma/issues/12182
- [x] https://github.com/magma/magma/issues/12183
- [x] https://github.com/magma/magma/issues/12184
- [x] https://github.com/magma/magma/issues/12185

### MME dependency Bazel-ification 

Non-blocking because we can let Bazel use system libraries (dynamically linked) - but this approach is not preffered for reasons of dependency consistency challenges across environments, build reproduction, and container size - among other reasons.

- [x] https://github.com/magma/magma/issues/9710 (Not a blocker exactly, this is a nice to have for completeness)
- [x] https://github.com/magma/magma/issues/9728 (small)
- [x] https://github.com/magma/magma/issues/10796 (small)
- [x] https://github.com/magma/magma/issues/10797 (small)
- [x] https://github.com/magma/magma/issues/10800 (small)
- [x] https://github.com/magma/magma/issues/10798 (small)

# Final Clean Up (Once Bazel-ified)
- [x]  https://github.com/magma/magma/issues/11736
- [x]  https://github.com/magma/magma/issues/11760
- [x] https://github.com/magma/magma/issues/11959
- [x] #11767 (small) (@ajahl @vktng  @nstng )
- [x] Migrate code coverage generation in CI agw-workflow
- [x] #12153

## MME dependency cycle refactoring tasks
- [x] #11892

# In Progress
- [x] #12057 (@jheidbrink  @LKreutzer )
- [x] https://github.com/magma/magma/issues/11656 (@electronjoe)

# Completed!
- [x] #12232
- [x] #12233 
- [x] #11861 (hard) (@ajahl @vktng  @nstng )
- [x] https://github.com/magma/magma/issues/11537
- [x] https://github.com/magma/magma/issues/10320
- [x] https://github.com/magma/magma/issues/11532
- [x] https://github.com/magma/magma/issues/11646 (@themarwhal)
- [x] #11881
- [x] #11899 (@electronjoe)
[...]

PR #11682, opened 2022-02-18
chore(mme): Removes dangerous wildcard namespace imports

electronjoe:
In debugging some build issues with Bazel build (for #9714), I decided to apply best
practices to these import statements and ONLY import the precise symbols
necessary for the local reference.  Per Google C++ Style Guide.

Broadly we should audit all such imports in the code base and trim them down to only what is needed.


ulaskozat (approved):
LGTM

PR #11696, opened 2022-02-18
Revert "Revert "Revert "fix(deps): Fixing CPU architecture hard codings to enable aarch64"""

quentinDERORY:
Reverts magma/magma#11671


themarwhal:
Triple revert!!

PR #11701, opened 2022-02-19
fix(mme): Provide cpp compatible variant of PARENT_STRUCT

### LINKED ISSUE #11585: Verify change to PARENT_STRUCT macro in common_defs.h

electronjoe:
Closes #11585.

The existing macro fails to compile In Bazel build effort because some paths to the containing header are not guarded by `extern "C"`. In C++, this macro's `typeof` use must instead be `decltype`.  Missing `extern "C"` linkage is a common issue with the Magma code-base and quite challenging to fixup. It will be aided by rename of CPP headers to .hpp (see GH issue associated with that effort).

I did attempt to fixup all the missing `extern "C"` calls, but then ran into new errors on the macro compilation that appeared again related to extern "C" use.

## Test Plan

Bazel build success on prototype branch.

CI tests include `make test_oai` and `make build_oai`.

### DIFF
--- a/lte/gateway/c/core/oai/common/common_defs.h
+++ b/lte/gateway/c/core/oai/common/common_defs.h
@@ -51,11 +51,21 @@
 #define COUNT_OF(x) \
   ((sizeof(x) / sizeof(0 [x])) / ((size_t)(!(sizeof(x) % sizeof(0 [x])))))
 
+#ifndef __cplusplus
 #define PARENT_STRUCT(cOnTaiNeD, TyPe, MeMBeR)                    \
   ({                                                              \
     const typeof(((TyPe*)0)->MeMBeR)* __MemBeR_ptr = (cOnTaiNeD); \
     (TyPe*)((char*)__MemBeR_ptr - OFFSET_OF(TyPe, MeMBeR));       \
   })
+#endif
+
+#ifdef __cplusplus
+#define PARENT_STRUCT(cOnTaiNeD, TyPe, MeMBeR)                      \
+  ({                                                                \
+    const decltype(((TyPe*)0)->MeMBeR)* __MemBeR_ptr = (cOnTaiNeD); \
+    (TyPe*)((char*)__MemBeR_ptr - OFFSET_OF(TyPe, MeMBeR));         \
+  })
+#endif
 
 #define OAI_MAX(a, b)       \
   ({                        \


lte/gateway/c/core/oai/common/common_defs.h
```diff
@@ -51,11 +51,21 @@
...
     const typeof(((TyPe*)0)->MeMBeR)* __MemBeR_ptr = (cOnTaiNeD); \
     (TyPe*)((char*)__MemBeR_ptr - OFFSET_OF(TyPe, MeMBeR));       \
   })
+#endif
+
+#ifdef __cplusplus
+#define PARENT_STRUCT(cOnTaiNeD, TyPe, MeMBeR)                    \
+  ({                                                              \
+    const decltype(((TyPe*)0)->MeMBeR)* __MemBeR_ptr = (cOnTaiNeD); \
+    (TyPe*)((char*)__MemBeR_ptr - OFFSET_OF(TyPe, MeMBeR));       \
```
maxhbr:
```suggestion
    (TyPe*)((char*)__MeMBeR_ptr - OFFSET_OF(TyPe, MeMBeR));       \
```
I do not understand the naming convention, but that looks inconsistent.

lte/gateway/c/core/oai/common/common_defs.h
```diff
@@ -51,11 +51,21 @@
...
+#ifndef __cplusplus
 #define PARENT_STRUCT(cOnTaiNeD, TyPe, MeMBeR)                    \
   ({                                                              \
     const typeof(((TyPe*)0)->MeMBeR)* __MemBeR_ptr = (cOnTaiNeD); \
     (TyPe*)((char*)__MemBeR_ptr - OFFSET_OF(TyPe, MeMBeR));       \
   })
+#endif
+
+#ifdef __cplusplus
+#define PARENT_STRUCT(cOnTaiNeD, TyPe, MeMBeR)                    \
```
sebathomas: Is e.g. `cOnTaiNeD` some kind of naming convention for C++ macros?

Neudrino: The Magma Code Conventions (for C) can be found [in the Wiki](https://github.com/magma/magma/wiki/Contributing-Code-Conventions#c) and are referring to the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html).

I was not able to find anything advising this type of naming on first glance. So imho, this should be either cleaned up (_preferred_) or a direct link to the section demanding that formatting should be given.

electronjoe: My intent was to make it easy to see what I had and had not changed in this hairy macro.  So changing names was something I decided not to do in this PR (there were some 0.o over the naming conventions here).  I do not know WHY this form of naming conventions were adopted, nor what the rules are - but OAI appears to be "consistent" across their macros [examples can be found here](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/magma/magma%24+%23define.*%28&patternType=regexp).  Happy to rename the fields / fixup naming in this PR if you prefer, however.

lte/gateway/c/core/oai/common/common_defs.h
electronjoe: How in the world is this compiling... Now I question things.

lte/gateway/c/core/oai/common/common_defs.h
Neudrino: Ok. I understand that no one knows, why it was introduced in first place, but it is widely used in OAI.
So renaming might just be out-of-scope in this PR to keep it consistent with the rest of the component.

lte/gateway/c/core/oai/common/common_defs.h
electronjoe: AAH the UI is what is confusing me here. This is a SUGGESTION not a snippet of my diff (which appears it would not compile if applied). So you are suggesting that your pattern match estimate of the naming convention should result in `__MeMBeR_ptr` (based upon their use of `MemBeR` probably)? That seems reasonable to assume. I would fall back to "I was trying to not change things other than the critical change in this PR". How about I post a follow-up PR to change all the naming of this macro?

Neudrino: :crossed_fingers:

electronjoe: Once this PR is merged I'll float [this renaming PR](https://github.com/electronjoe/magma/commit/415bb8e81c0e90f60dbbba7b37dde53f7a7d738a)

lte/gateway/c/core/oai/common/common_defs.h
electronjoe: Once this PR is merged I'll float [this renaming PR](https://github.com/electronjoe/magma/commit/415bb8e81c0e90f60dbbba7b37dde53f7a7d738a)

electronjoe:
Seeing some lte test failures with python traces. I think I will rebase on master and see if they go away, then look more deeply if they do not.

sebathomas (approved):

ardzoht (approved):
lgtm

lte/gateway/c/core/oai/common/common_defs.h
lionelgo: This was done like this, in a so silly way to avoid collisions with whatever,  whatever would be the (consistent) naming convention.

PR #11722, opened 2022-02-21
feat(dp): Restart SAS flow on cbsd update

jkmar:
## Summary

When cbsd is updated it should restart SAS flow (be deregistered and registered again),
since updated fields can affect SAS requests (e.g serial number).

This feature is implemented by adding is_updated flag to cbsd db table.
When orc8r dp service updates cbsd it also changes that flag to true.
Active Mode Controller then deregisters cbsds with that flag set,
and only after deregistration it removes the flag.

The following condition should always hold:
when cbsd is updated then:
- either is_updated flag is set
- or cbsd is deregistered.


xbend (approved):

dp/cloud/python/magma/radio_controller/tests/test_utils/active_mode_config_builder.py
```diff
@@ -3,10 +3,10 @@
 from typing import Optional
 
 from dp.protos.active_mode_pb2 import (
-    ActiveModeConfig,
```
WojSad, line 6: since ActiveModeConfig is out of the picture, maybe let's rename the class file altogether or move this code to db_cbsd_builder.py?
Also changing the builder class name seems reasonable

WojSad (approved):

dp/cloud/python/magma/radio_controller/tests/test_utils/active_mode_config_builder.py
jkmar: Changed to `ActiveModeCbsdBuilder` (we already have `DbCbsdBuilder`, so just `CbsdBuilder` wasn't enough)

PR #11731, opened 2022-02-22
chore(monitord): monitord is bazelified

### LINKED ISSUE #11657: bazel: Bazelify Python monitord

vktng:

## Summary

- Module monitord is bazelified including the tests.

## Test Plan

- Run monitord:
  - `bazel run //lte/gateway/python/magma/monitord:monitord`
- Run tests:
  - `bazel test //lte/gateway/python/magma/monitord/tests:all`

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/monitord/BUILD.bazel
+++ b/lte/gateway/python/magma/monitord/BUILD.bazel
@@ -1,0 +1,54 @@
+# Copyright 2022 The Magma Authors.
+
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_python//python:defs.bzl", "py_binary", "py_library")
+
+MAGMA_ROOT = "../../../../../"
+
+ORC8R_ROOT = "{}orc8r/gateway/python".format(MAGMA_ROOT)
+
+LTE_ROOT = "{}lte/gateway/python".format(MAGMA_ROOT)
+
+py_binary(
+    name = "monitord",
+    srcs = ["main.py"],
+    imports = [
+        LTE_ROOT,
+        ORC8R_ROOT,
+    ],
+    # legacy_creat_init = False is required to fix issues in module import, see https://github.com/rules-proto-grpc/rules_proto_grpc/issues/145
+    legacy_create_init = False,
+    main = "main.py",
+    visibility = ["//visibility:private"],
+    deps = [
+        ":monitord_lib",
+        "//lte/protos:mconfigs_python_proto",
+        "//lte/protos:mobilityd_python_proto",
+        "//orc8r/gateway/python/magma/common:sentry",
+        "//orc8r/gateway/python/magma/common:service",
+    ],
+)
+
+py_library(
+    name = "monitord_lib",
+    srcs = [
+        "cpe_monitoring.py",
+        "icmp_job.py",
+        "icmp_state.py",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [
+        "//lte/protos:mobilityd_python_grpc",
+        "//orc8r/gateway/python/magma/common:rpc_utils",
+        "//orc8r/gateway/python/magma/common:sentry",
+        "//orc8r/gateway/python/magma/magmad/check/network_check:ping",
+    ],
+)
--- a/lte/gateway/python/magma/monitord/tests/BUILD.bazel
+++ b/lte/gateway/python/magma/monitord/tests/BUILD.bazel
@@ -1,0 +1,31 @@
+# Copyright 2022 The Magma Authors.
+
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("//bazel:python_test.bzl", "pytest_test")
+
+MAGMA_ROOT = "../../../../../../"
+
+ORC8R_ROOT = "{}orc8r/gateway/python".format(MAGMA_ROOT)
+
+LTE_ROOT = "{}lte/gateway/python".format(MAGMA_ROOT)
+
+pytest_test(
+    name = "test_icmp_monitor",
+    srcs = ["test_icmp_monitor.py"],
+    imports = [
+        LTE_ROOT,
+        ORC8R_ROOT,
+    ],
+    deps = [
+        "//lte/gateway/python/magma/monitord:monitord_lib",
+        "//lte/protos:mobilityd_python_proto",
+    ],
+)


lte/gateway/python/magma/monitord/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+    ],
+    # legacy_creat_init = False is required to fix issues in module import, see https://github.com/rules-proto-grpc/rules_proto_grpc/issues/145
+    legacy_create_init = False,
+    main = "main.py",
+    visibility = ["//visibility:private"],
+    deps = [
+        ":monitord_lib",
+        "//lte/protos:mconfigs_python_proto",
+        "//lte/protos:mobilityd_python_proto",
+        "//orc8r/gateway/python/magma/common:rpc_utils",
```
LKreutzer: ~~Looking at the .py files it seems that the rpc_utils are required by cpe_monitoring.py and not main.py and should thus be moved to the py_library deps?~~

Looking at the .py files it seems that the rpc_utils are required only by cpe_monitoring.py and not main.py?

lte/gateway/python/magma/monitord/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_python//python:defs.bzl", "py_binary", "py_library")
+
+LTE_ROOT = "../../"
+
+ORC8R_ROOT = LTE_ROOT + "../../../orc8r/gateway/python"
```
electronjoe: nitpick: Here and elsewhere in this PR.  Using existing DEFINES to build off of them makes a lot of sense if half of a construction is likely to change and is based on another construction (not sure I'm wording that well).  But here, ORCR_ROOT really has no meaningful tie to LTE_ROOT and I would suggest you define ORC8R_ROOT without reference to LTE_ROOT. It implies relationship that doesn't exist?

lte/gateway/python/magma/monitord/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+
+py_binary(
+    name = "monitord",
+    srcs = ["main.py"],
+    imports = [
```
electronjoe, line 23: Question (maybe @nstng ?) : I have forgotten how we are python Bazelifying and am wondering why the `imports` subsection exists - I presume this is to allow discovery of in-repo python libraries imported by this python without full repo-relative paths.  Did we not find a way to avoid this and use deps = section instead? I am guessing that's not possible until we fully Bazel transition?

If the above is correct, how do we feel about a TODO on each import clause for python BUILD.bazel files noting that the imports section will be deprecated / fixed up once we are fully over to Bazel with a GH issue associated in the TODO? @themarwhal has been doing this for our C/C++ ticket items that needed improvement down the road (so people copy pasting realize it's not optimal yet, and it's easy to discover the work down the road).

lte/gateway/python/magma/monitord/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+)
+
+py_library(
+    name = "monitord_lib",
+    srcs = [
+        "cpe_monitoring.py",
+        "icmp_job.py",
+        "icmp_state.py",
+    ],
+    visibility = ["//visibility:public"],
```
electronjoe, line 47: Question (again, perhaps for @nstng): I recall asking this question previously but don't know how it settled.  In my time at Google we typically embedded visibility information in the package declaration at top of BUILD.bazel (rarely placing it in individual rules).  I get that this is less granular and - if e.g. package is public - could lead to default behavior that is not what is intended. But it seemed generally fine (and reduces a lot of per-target boilerplate). What are your thoughts?

lte/gateway/python/magma/monitord/BUILD.bazel
nstng: You are right - we should change this here (and in other services when we touch them again - I was about to create general clean-up findings issue anyway).

lte/gateway/python/magma/monitord/BUILD.bazel
nstng: `imports` adds entries to the bazel runtime PYTHONPATH - somewhere this needs to be specified . Problem here (or in general) is that these paths needs to be releative. The reason is imho that bazel does the builds in a sandbox where the absolute folder structure is unknown (`/tmp/bazel/sandbox/42/...`). Also the added `LTE_ROOT` and `ORC8R_ROOT` depends on the folder structure, i.e., it could be `../../` or `../../../`. I would really like to clean this up but currently have no good idea. - not sure if this answers your question, if not, let me know.

lte/gateway/python/magma/monitord/BUILD.bazel
nstng: We already have a couple of cases where some libraries are private, e.g., we started to use the pattern to have a `py_binary` only reference the `main.py` and add other source files as a `py_library` (this greatly improves the readability of dependencies). If the `py_library` is not tested then there is currently no reason to make it public.
Also I was thinking about to introduce grouped visibility - i.e., a target can only be used by a certain subset of packages.
In my opinion a strict visibility policy can mitigate uncontrolled growth over time. But if you prefer to reduce the boilerplate then we can change this - my opinion is not that strong on this topic.

lte/gateway/python/magma/monitord/BUILD.bazel
nstng: Had a brainstorming in the team because we all are not happy with the current solution. What do you think about the following? - should make it more clear in our opinion

```
MAGMA_ROOT = "../../../../../"
ORC8R_ROOT = "{}orc8r/gateway/python".format(MAGMA_ROOT)
LTE_ROOT = "{}lte/gateway/python".format(MAGMA_ROOT)
```

LKreutzer (changes requested):

nstng:
FYI iputils-ping is missing in the devcontainer - we will add another PR for this that needs to be handled before.

vktng:
> FYI iputils-ping is missing in the devcontainer - we will add another PR for this that needs to be handled before.

PR created: https://github.com/magma/magma/pull/11744

lte/gateway/python/magma/monitord/BUILD.bazel
nstng: Had a brainstorming in the team because we all are not happy with the current solution. What do you think about the following? - should make it more clear in our opinion

```
MAGMA_ROOT = "../../../../../"
ORC8R_ROOT = "{}orc8r/gateway/python".format(MAGMA_ROOT)
LTE_ROOT = "{}lte/gateway/python".format(MAGMA_ROOT)
```

lte/gateway/python/magma/monitord/BUILD.bazel
nstng: Had a brainstorming in the team because we all are not happy with the current solution. What do you think about the following? - should make it more clear in our opinion

```
MAGMA_ROOT = "../../../../../"
ORC8R_ROOT = "{}orc8r/gateway/python".format(MAGMA_ROOT)
LTE_ROOT = "{}lte/gateway/python".format(MAGMA_ROOT)
```

Neudrino: The suggestion would also make "the level" of the Python root compared to the repository root more clear. :+1: 
This might be especially helpful, once/if a bigger code restructuring would be desired / done somewhere in future.

lte/gateway/python/magma/monitord/BUILD.bazel
vktng: Done!

LKreutzer (approved):
Tested in dev-container, magma-vm and bazel-base.

lte/gateway/python/magma/monitord/BUILD.bazel
electronjoe: I didn't realize such statements were possible in Starlark! But I guess since it's Python based it makes sense that it is capable.  LGTM. Thanks for putting thought into this.

lte/gateway/python/magma/monitord/BUILD.bazel
electronjoe: Can you take a look at other bazel python projects on GitHub and see if they all have imports statements? I feel like this should not be necessary (but maybe I'm wrong with respect to Python).  Does Bazel python not infer an import of the root of the repo? (that would be my expectation, based on how other languages e.g. Golang are handled w.r.t. imports).

lte/gateway/python/magma/monitord/BUILD.bazel
electronjoe: I don't presently have strong opinions either. On a project like Magma I would generally assume that libraries can be used anywhere in the repo (we are not trying to limit dependencies by other teams). But I agree that some files make no sense to be used by others (e.g. simple main.py files).  For test purposes, I often see use of `testonly = 1,` to prevent use in production code (better than visibility for this purpose).

TL;DR; feel free to continue doing this, my MME bazelification is using package-level visibility so far.

electronjoe (approved):
% your proposed changes this LGTM

lte/gateway/python/magma/monitord/BUILD.bazel
nstng: I get a lot of hits via `file:.(?:bzl|BUILD|bazel)$ py_binary AND imports =`. Usually only one entry is used - but I guess we have a special case at magma because the Python `magma` module exists in multiple sub-folders (`lte/...` and `orc8r/...`).

While searching and thinking about it ... it might be possible to move the imports down to the libraries. This way the binaries and the tests do not need imports. But I fear that this could result in the pythonpath/sys.path becoming confusing  "Why Python has/has not access to this library?!".

lte/gateway/python/magma/monitord/BUILD.bazel
nstng: We will discuss this again in the team. And we will have a look at and start using `testonly = 1` (did not had this on my radar - thank you)

vktng: Added a revise task to our backlog.

nstng (approved):
in pairing

PR #11744, opened 2022-02-22
chore: iputils-ping is added to the devcontainer (needed by python services)

vktng:

## Summary

First step to add iputils-ping dependency to the devcontainer (needed for python services).

## Additional Information

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -56,6 +56,8 @@ RUN echo "Install general purpose packages" && \
         g++ \
         g++-9 \
         gcc-9 \
+        # iputils-ping is a dependency of python services (e.g. magmad)
+        iputils-ping \
         lcov \
         gdb \
         git \


LKreutzer (approved):
Lgtm

nstng (approved):
in pairing

.devcontainer/Dockerfile
```diff
@@ -56,6 +56,8 @@ RUN echo "Install general purpose packages" && \
         g++ \
         g++-9 \
         gcc-9 \
+        # dependency of python services (e.g. magmad)
```
electronjoe: Nit / feedback: is there any way to make this comment more clearly pertain to just iputils-ping? As it stands now I might wonder if I were a novice looking at this list.  Is it possible to `iputils-ping \ # dependency of python services (e.g. magmad)`? If not, maybe we say on line 59 `# iputils-ping is a dependency of python services (e.g. magmad)`? What do you think.

vktng: I would have preferred the first solution, but the highlighting was not correct, so we opted for the second option.

Neudrino (approved):

themarwhal (approved):

PR #11746, opened 2022-02-22
chore(agw): upload journal log to sentry

ajahl:

## Summary

The PR is a prototype for uploading journal log information to Sentry when a c service crashes.

Approach: We used the `before_send` callback method that is called by the Sentry SDK before a crash report is sent. The callback extracts certain information from `journalctl` log and provide it stored in a separate log file,  [magma_services.log](https://github.com/magma/magma/files/8124561/magma_services.log). The crash report contains the log file as attachment. 

## Additional Information


orc8r/gateway/c/common/sentry/SentryWrapper.cpp
```diff
@@ -36,6 +36,7 @@
 #define SHARED_MCONFIG "shared_mconfig"
 #define SHOULD_UPLOAD_MME_LOG "sentry_upload_mme_log"
 #define MME_LOG_PATH "/var/log/mme.log"
+#define SYS_LOG_PATH "/var/log/magma_services.log"
```
alexzurbonsen: While reading this again: should we rename `SYS_LOG_PATH` to something that reveals that we are writing to this path and not reading from it? The current name might suggest that it is the path to the syslog file, analogous to  `MME_LOG_PATH`.

ajahl: We write the file first and afterwards read it via Sentry SDK. I renamed in `JOURNAL_LOG_PATH`.

alexzurbonsen (approved):

Neudrino:
Thanks @themarwhal and @ardzoht for the insights shared today!

There were two items which I suggest to address separately in respective follow up PRs:
- providing SCTPd logs, and
- making the number of lines configurable via streamed mconfig.

orc8r/gateway/c/common/sentry/SentryWrapper.cpp
```diff
@@ -115,6 +116,14 @@ std::string get_snowflake() {
   return buffer.str();
 }
 
+static sentry_value_t before_send(sentry_value_t event, void* hint,
+                                  void* data) {
+  std::string command = "journalctl -u magma@* -n 1000 > ";
```
sebathomas: To get sctpd as well:
```suggestion
  std::string command = "journalctl -u sctpd -u magma@* -n 1000 > ";
```

Neudrino (approved):

orc8r/cloud/go/services/orchestrator/obsidian/models/swagger.v1.yml
```diff
@@ -1451,6 +1451,11 @@ definitions:
       upload_mme_log:
         type: boolean
         example: false
+      number_of_lines_in_log:
+        type: number
+        format: uint32
+        minimum: 0
```
hcgatewood: This might be unnecessary given the object's type (?)

hcgatewood (approved):
Cloud changes lgtm

themarwhal (approved):
Woohoo!

alexzurbonsen (changes requested):
We should also modify the go code that translates from swagger to proto, such that it reflects the changes.

sebathomas:
Right. I think that's:
* lte/cloud/go/services/lte/servicers/builder_servicer.go
* orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go

ajahl:
> We should also modify the go code that translates from swagger to proto, such that it reflects the changes.


> Right. I think that's:
> 
> * lte/cloud/go/services/lte/servicers/builder_servicer.go
> * orc8r/cloud/go/services/orchestrator/servicers/builder_servicer.go

It's done. I modified the go files.

sebathomas (approved):
Tried it on my machine and confirmed that the number of lines is forwarded to the mconfig file in the AGW.

alexzurbonsen (approved):

sebathomas:
Just noticed the `yarn test` error in the CI. Rerunning to see if it's flaky.

sebathomas (changes requested):
Nope, I think it's consistently red. We should fix that before merging.

sebathomas (approved):
This was not broken by the PR, it's also red on master. Opened another ticket to track that:
#11895

PR #11747, opened 2022-02-22
chore: pull build cache on codespace start

themarwhal:

## Summary
Add two extra commands that get executed on Codespace creation to pull in cache for Bazel build. The cache is pulled from S3 which is populated by a daily GitHub Action job. (See https://github.com/magma/magma/actions/workflows/bazel-cache-push.yml)

While we save ~3 minutes on `bazel build //...`, we will spend about 5-10 seconds more pulling in the cache on Codespace startup. I believe this will not affect restarting a codespace that has already been created. 

## Numbers
Running `bazel build //...` on a new Codespace (16 cores) on master takes 4m 42s
Screen Shot 2022-02-22 at 3 44 32 PM

Running `bazel build //...` on a new Codespace (16 cores) off of this PR takes 1m 23s
Screen Shot 2022-02-22 at 3 50 27 PM

A Codespace can be spun up by clicking on the Code button like this
Screen Shot 2022-02-22 at 3 47 12 PM

## Test Plan
Tested by spinning up a Codespace from this PR.

## Additional Information

### DIFF
--- a/.devcontainer/post-create-commands.sh
+++ b/.devcontainer/post-create-commands.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# $1: should be containerWorkspaceFolder from https://code.visualstudio.com/docs/remote/devcontainerjson-reference
+
+sudo ln -s "$1"/lte/gateway/configs /etc/magma
+echo "alias magtivate='source /home/vscode/build/python/bin/activate'" >> ~/.bashrc
+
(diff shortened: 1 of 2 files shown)


.devcontainer/post-create-commands.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+sudo ln -s $1/lte/gateway/configs /etc/magma
+echo \"\nalias magtivate='source /home/vscode/build/python/bin/activate'\" >> ~/.bashrc
```
themarwhal: fixing...

electronjoe (approved):

Issue #11743, opened 2022-02-22
Bazel technical debt


LKreutzer:
### Bazel technical debt

This issue collects technical debts from the ongoing bazelification effort originating from new insights. Changes will be applied according to the scout rule or as dedicated refactoring if appropriate.

## Prioritized backlog

- Refactor WORKSPACE.bazel (has grown really big).
- State of `requirements.in` should be validated vs `setup.py`s.
- Revise visibility of targets. Global visibility vs local visibility. (see https://github.com/magma/magma/issues/12531)
- Use external swagger rule if one is released.
- Use external pytest rule if one is released (also needs to support `bazel coverage`).
- Centralize all bazel (rules, etc) dependencies into one file - having one point where upgrades can be organized or pinned versions can be documented.
- Update Lupa and fakeredis version when python restrictions are resolved.
- Use python rules>=0.8.0 when python3.8 for arm is supported.

## Done

- test_only = 1 -> test_only = True #12473 
- Use `testonly = 1` in test only libraries. #11802
- Fix typo everywhere `legacy_creat_init` #11803
- Inline all lists in BUILD.bazel files (so the buildifier can format them). #11789
- Use pattern for imports: #11778
```
MAGMA_ROOT = "../../../../../"
ORC8R_ROOT = "{}orc8r/gateway/python".format(MAGMA_ROOT)
LTE_ROOT = "{}lte/gateway/python".format(MAGMA_ROOT)
```
- Split py_library from py_binary for each service. I.e. move the .py files (other than main.py) into a separate py_library. (@LKreutzer) https://github.com/magma/magma/pull/12154
- Add [`size`](https://docs.bazel.build/versions/main/be/common-definitions.html#test.size) attribute to python test targets (default is 'medium', with 5 minute timeout). https://github.com/magma/magma/pull/12196
- Figure out why magmad tests do not need imports (swagger generated files needs to be checked). (@LKreutzer) https://github.com/magma/magma/pull/12225

nstng:
outdated - already covered by other issues

PR #11748, opened 2022-02-22
chore: pin Bazel build tools version to 5.0.1

themarwhal:

## Summary
Pinning it to a specific version so that the sha256 value does not change. Having a fixed version / sha256 will reduce the need for Bazel to fetch the repository often. (I think it might be everytime the Bazel server restarts? I am actually not sure). 

Now that GitHub Codespaces pulls in build cache on container start (https://github.com/magma/magma/pull/11747), this should mean that buildifier is never built from scratch unless the version is updated. 

## Additional Information

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -61,8 +61,9 @@ gazelle_dependencies(go_repository_default_config = "@//:WORKSPACE.bazel")
 
 http_archive(
     name = "com_github_bazelbuild_buildtools",
-    strip_prefix = "buildtools-master",
-    url = "https://github.com/bazelbuild/buildtools/archive/master.zip",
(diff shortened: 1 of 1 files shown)


nstng:
@themarwhal Out of scope comment: maybe we should think about moving all version (and I assume also sha) strings into a separate file so that we have a centralized point with all needed bazel version information. Should make it easier to define a regular "check if there a new versions" task or to document versions that are pinned by design.

[Edit] Went ahead and added it to https://github.com/magma/magma/issues/11743 - if you have arguments against such change, let me know or feel free to remove it.

nstng (approved):
lgtm - tested local build (working and no complains about the sha)

PR #11749, opened 2022-02-22
chore: add a role to symlink clang-format-11 in Magma VM

themarwhal:

## Summary
We recently updated our default clang-format from version 7 to 11. However, we missed that the role to symlink `/usr/bin/clang-format` to `clang-format-11` is run only for preburn. @tmdzk has offered to upgrade the base image soon, but in the meantime I will move it to a provision step.

## Test Plan
provisioned both magma and magma_test

After provisioning, ran `/usr/bin/clang-format --version` to make sure it points to clang-format-11, not clang-format-7

```bash
vagrant@magma-dev-focal:~/magma$ /usr/bin/clang-format --version
Ubuntu clang-format version 11.0.0-2~ubuntu20.04.1
```

## Additional Information

### DIFF
--- a/lte/gateway/deploy/roles/dev_common/tasks/main.yml
+++ b/lte/gateway/deploy/roles/dev_common/tasks/main.yml
@@ -396,11 +396,6 @@
     grub-mkconfig -o /boot/grub/grub.cfg
   when: preburn
 
-- name: Create links for clang-format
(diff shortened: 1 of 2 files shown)


ssanadhya (approved):
Thanks for the quick fix!

PR #11759, opened 2022-02-23
fix(mme): Move the zmq context into truly shared file in mme_app_task

electronjoe:
In breaking up these unit tests to run more concurrently, I had linking
errors in finding this symbol.  Turns out we have defined the symbol in
one place (test utilities header file) but actually instantitated it in
a non-shared c file (at least not part of utilities).

This PR moves the declaration to a more sensible place, the
mme_app_test_util.cpp file.

## Test Plan

Execution of CI tests including `make build_oai` and `make test_oai`.


ardzoht (approved):
lgtm

ssanadhya (approved):
lgtm

PR #11773, opened 2022-02-24
feat(orc8r): EPS Authentication Service - a gRPC version of S6a type authentication

emakeev:

EPS Authentication Service a gRPC equivalent of S6a based authentication.

## Summary

The service mimics FeG's S6a proxy protocol, but relies on internal subscriberdb for client credentials vs S6a proxy's reliance on HSS. This service was recovered & refactored from previously broken & depreciated magma codebase.

It was tested using original unit tests & it's not yet enabled in prod.

## Test Plan

unit tests

## Additional Information


uri200 (approved):

lte/protos/BUILD.bazel
```diff
@@ -354,6 +354,16 @@ cpp_grpc_library(
     deps = [":sms_orc8r_cpp_proto"],
 )
 
+proto_library(
```
nstng, line 357: Tested locally that
`bazel build //lte/protos:diam_errors_proto` and
`bazel build //lte/protos:diam_errors_cpp_proto`
are building. ~~This is the only bazel change here though. These targets are currently not used in any bazel binary or test target. I recommend to remove these changes for now. The bazel proto libraries should be added when `diam_errors.proto` is used in other bazel targets - fyi using bazel is an ongoing effort.~~
[Edit] Ok, just realized that `diam_errors.proto` is used in `subscriberauth.proto` and needs to be added as a dependency. I will check what needs to be changed and answer below.

nstng, line 357: This comment is actually for line 367:
`diam_errors_proto` needs to be added as a dependency, i.e., after line 369 add
`deps = [":diam_errors_proto"],`

can be tested on vm or dev container via `bazel build lte/protos:subscriberauth_proto`. Or build all proto libraries via `bazel build lte/protos/...`.

nstng, line 357: This comment is actually for line 372:
`diam_errors_cpp_proto` needs to be added as a dependency, i.e., after line 374 add
`deps = [":diam_errors_cpp_proto"]`

can be tested on vm or dev container via `bazel build lte/protos:subscriberauth_cpp_proto`. Or build all proto libraries via `bazel build lte/protos/...`.

nstng (changes requested):

hcgatewood (approved):

lte/protos/BUILD.bazel
emakeev: It doesn't seem to be enough, I still see build errors.
I am not sure how to fix the build since I cannot run VirtualBox on M1 Mac & don't see relevant magma bazel build instructions...

nstng: Hi, the Bazel build seems to be working again [1]. Thank you for using the suggested fixes :) The other failing Jobs seem to be not related to this issue - the Bazel-ification is an ongoing effort, i.e., required CI builds are not affected yet. Feel free to join #bazel in the Magma slack if you have more questions.

[1] https://github.com/magma/magma/actions/runs/1896125985

nstng (approved):
Bazel changes lgtm

ardzoht (approved):
`lte/.../` changes lgtm

PR #11313, opened 2022-01-28
feat(agw): add upgrade function for eNB

weiyongchao0412:
Signed-off-by: weiyongchao <weiyongchao@baicells.com>

## Summary

- Add command and message for upgrading eNB.
- We can send upgrade command to eNB from eNB's upgrade page in NMS.

### Unittest classes and methods for rts model have been added, we run the unittest at local environment.

1. agw's make test result:
make-test-result

2. unittest result for upgrade function:
unittest-of-manual-download

### Upgrade Progress:

1. eNB's version info before upgrade:
enodeb-info-before-upgrade

2. Upgrading page on NMS:
nms-enodeb-upgrade-dialog

3. eNB's version info after upgrade:
enodeb-info-after-upgrade

## Additional Information

Support Nova 227/ Nova243 software version:3.6.6

We can use this tool FileServer(https://github.com/weiyongchao0412/FileServer) to get upgrading file's various infos to input on NMS eNB upgrade page.
Sample preview:
fileserver-preview

- [ ] This change is backwards-breaking

PR #11775, opened 2022-02-24
feat(agw): add upgrade function for eNB

weiyongchao0412:
Signed-off-by: weiyongchao0412 <weiyongchao@baicells.com>

## Summary

- Add command and message for upgrading eNB.
- We can send upgrade command to eNB from eNB's upgrade page in NMS.

### Unittest classes and methods for qrtb model have been added, we run the unittest at local environment.

1. agw's make test result:
make-test-result

2. unittest result for upgrade function:
unittest-of-manual-download

### Upgrade Progress:

1. eNB's version info before upgrade:
enodeb-info-before-upgrade

2. Upgrading page on NMS:
nms-enodeb-upgrade-dialog

3. eNB's version info after upgrade:
enodeb-info-after-upgrade

## Additional Information

Support Nova 436q software version:2.9.2+

We can use this tool FileServer(https://github.com/weiyongchao0412/FileServer) to get upgrading file's various infos to input on NMS eNB upgrade page.
Sample preview:
fileserver-preview


electronjoe (commented):
Is this a dupe with https://github.com/magma/magma/pull/11313? If so can we close one of them?

weiyongchao0412:
> Is this a dupe with #11313? If so can we close one of them?

Sorry for the confusion. #11313 is adding upgrade function for Nova 227/Nova 243 devices. This PR is for Nova 436Q devices. The code implement is different. For your reference. Thanks.

lte/gateway/python/magma/enodebd/state_machines/enb_acs.py
```diff
@@ -218,3 +221,9 @@ def is_enodeb_connected(self) -> bool:
     @abstractmethod
     def stop_state_machine(self) -> None:
         pass
+
+    def download_asap(self, url: str, user_name: str, password: str, target_file_name: str, file_size: int, md5: str) -> None:
+        """
+        Send a request to download&upgrade the eNodeB ASAP
+        """
+        pass
```
andreilee, line 229: ## 1
This should probably be marked as `@abstractmethod`. While this has been implemented for QRTB, if this method is called on other eNodeB handlers within `enodebd` service, this method will fail silently.


## 2
Some additional documentation is warranted here regarding the arguments. Does the `url` refer specifically to a password protected download link? Must it always, or is it possible for it refer to a download URL that is not password protected? And in that case, what use is the `user_name` and `password` - what would the caller pass in? There are some opinionated decisions being made here about exactly what is hiding behind this URL, but it's not clear to me as a reviewer that every download URL will conform to what's being required here.

One possible alternate function signature could be:
```suggestion
    def download_asap(self, download_metadata: Any) -> None:
        """
        Send a request to download&upgrade the eNodeB ASAP
        """
        # QRTB handler would check the type of download_metadata, and handle appropriately
        # Other eNodeB handlers might expect downloads to be specified differently, so maybe it needs its own type
        pass
```
This would require `enodebd_cli.py` to be more complicated to compensate, of course, but this is just one option. Either way, I think this method signature can't remain the way it is.

lte/protos/enodebd.proto
```diff
@@ -95,6 +95,26 @@ message SingleEnodebStatus {
...
+  string target_file_name = 6;
+  uint32 delay_seconds = 7;
+  string success_url = 8;
+  string failure_url = 9;
+  string device_serial = 10;
+  string md5 = 11;
+}
+
+message DownloadResponse {
+  uint32 status = 1;
```
andreilee, line 113: What does each number indicate for the status? This should be well documented, and nobody should need to dig through the code to figure this out.

lte/protos/enodebd.proto
```diff
@@ -95,6 +95,26 @@ message SingleEnodebStatus {
...
+  string start_time = 2;
+  string complete_time = 3;
```
andreilee, line 115: How are these specified? Is there a particular format?

lte/protos/enodebd.proto
```diff
@@ -117,6 +137,9 @@ service Enodebd {
   // Reboot every connected eNodeB
   rpc RebootAll (magma.orc8r.Void) returns (magma.orc8r.Void);
 
+  // Download file to eNodeB
+  rpc Download (DownloadRequest) returns (DownloadResponse);
```
andreilee, line 141: Considering that this is specifically for upgrading eNB firmware, that context should be included in the documentation here.

andreilee (changes requested):
Thanks for your contributions here! While this looks good overall, I will ask that you pay special attention to interface definitions, like the `.proto` definitions, and `enb_acs.py`.

PR #11777, opened 2022-02-24
chore(state): state service is bazelified

### LINKED ISSUE #11729: bazel: Bazelify Python state

LKreutzer:

## Summary

- Module state is bazelified including the tests.
- The state service has a unique feature: additional proto libraries can be loaded by adding them to `state.yml`. This means we need to depend on all proto libraries in `lte/protos` and `lte/protos/oai`. Other proto libraries should not be needed (discussed with @ardzoht).
 - This is solved by providing targets that provide all proto files from these folders. We know this is sort of an anti-pattern, but seems to be the solution that involves the least effort on the developer side for future changes. 

## Test Plan
In dev-container, magma-vm and bazel-base:

-  Run state: 
   - `bazel run //orc8r/gateway/python/magma/state:state`
- Run tests: 
  - `bazel test //orc8r/gateway/python/magma/state/tests:all`
- Adding  `state.yml` dependency
  - Add the following code: 
   ```
   - proto_file: "lte.protos.sctpd_pb2"
     proto_msg: "InitReq"
     redis_key: "mobilityd_ipdesc_record"
     state_scope: "network"
   ```
  - the following logging should not occur with a `bazel run` on state
    `ERROR:root:No module named 'lte.protos.sctpd_pb2'` 

## Additional Information

### DIFF
--- a/lte/protos/BUILD.bazel
+++ b/lte/protos/BUILD.bazel
@@ -454,3 +454,32 @@ python_grpc_library(
     name = "subscriberauth_python_grpc",
     protos = [":subscriberauth_proto"],
 )
+
+# This proto library is exclusively meant for the state service.
+# Users can manually add proto libraries to state.yml.
+# This means that all proto libraries need to be available
+# in builds of the state service.
+proto_library(
+    name = "all_proto",
+    srcs = glob(["*.proto"]),
+    visibility = ["//visibility:private"],
+    deps = [
+        "//orc8r/protos:common_proto",
+        "//orc8r/protos:digest_proto",
+        "//orc8r/protos:service303_proto",
+        "@protobuf//:field_mask_proto",
+        "@protobuf//:timestamp_proto",
+    ],
+)
+
+# See comment at :all_proto
+python_proto_library(
+    name = "all_python_proto",
+    protos = [":all_proto"],
+    visibility = ["//orc8r/gateway/python/magma/state:__pkg__"],
+    deps = [
+        "//orc8r/protos:common_python_proto",
+        "//orc8r/protos:digest_python_proto",
+        "//orc8r/protos:service303_python_proto",
+    ],
+)
--- a/lte/protos/oai/BUILD.bazel
+++ b/lte/protos/oai/BUILD.bazel
@@ -11,6 +11,7 @@
 
 load("@rules_proto//proto:defs.bzl", "proto_library")
 load("@rules_proto_grpc//cpp:defs.bzl", "cpp_proto_library")
+load("@rules_proto_grpc//python:defs.bzl", "python_proto_library")
 
 package(default_visibility = ["//visibility:public"])
 
@@ -131,4 +132,22 @@ proto_library(
 cpp_proto_library(
     name = "ngap_state_cpp_proto",
     protos = [":ngap_state_proto"],
+)
+
+# This proto library is exclusively meant for the state service.
+# Users can manually add proto libraries to state.yml.
+# This means that all proto libraries need to be available
+# in builds of the state service.
+proto_library(
+    name = "all_proto",
+    srcs = glob(["*.proto"]),
+    visibility = ["//visibility:private"],
+    deps = ["@protobuf//:timestamp_proto"],
+)
+
+# See comment at :all_proto
+python_proto_library(
+    name = "all_python_proto",
+    protos = [":all_proto"],
+    visibility = ["//orc8r/gateway/python/magma/state:__pkg__"],
 )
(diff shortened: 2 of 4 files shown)


electronjoe (approved):
Thank you! Looks good =D

orc8r/gateway/python/magma/state/BUILD.bazel
```diff
@@ -0,0 +1,69 @@
...
+    srcs = ["main.py"],
+    imports = [ORC8R_ROOT],
+    # legacy_creat_init = False is required to fix issues in module import, see https://github.com/rules-proto-grpc/rules_proto_grpc/issues/145
+    legacy_create_init = False,
+    main = "main.py",
+    python_version = "PY3",
+    visibility = ["//visibility:public"],
+    deps = [
+        ":garbage_collector",
+        ":keys",
```
nstng: keys and redis not needed (discussed in pairing)

orc8r/gateway/python/magma/state/BUILD.bazel
```diff
@@ -0,0 +1,69 @@
...
+py_binary(
+    name = "state",
+    srcs = ["main.py"],
+    imports = [ORC8R_ROOT],
+    # legacy_creat_init = False is required to fix issues in module import, see https://github.com/rules-proto-grpc/rules_proto_grpc/issues/145
+    legacy_create_init = False,
+    main = "main.py",
+    python_version = "PY3",
+    visibility = ["//visibility:public"],
+    deps = [
```
nstng, line 27: discussed in pairing - state.yml dependencies missing

nstng (changes requested):

orc8r/gateway/python/magma/state/BUILD.bazel
LKreutzer: Pushed.

orc8r/gateway/python/magma/state/BUILD.bazel
LKreutzer: Changes pushed.

lte/protos/oai/BUILD.bazel
```diff
@@ -66,6 +90,12 @@ proto_library(
     deps = [":common_types_proto"],
 )
 
+python_proto_library(
+    name = "std_3gpp_types_python_proto",
+    protos = [":std_3gpp_types_proto"],
+    deps = [":common_types_python_proto"],
+)
+
```
Neudrino: Out of curiosity: Why does this library has a dependency, while the below equivalent cpp library does not?
In most other cases this seems to be "symmetric".

LKreutzer: Obsolete

LKreutzer:
@electronjoe pushed some needed changes, you might want to have another look, especially regarding the `glob` statements and visibility of the protos targets. See `lte/protos/BUILD.bazel` and `lte/protos/oai/BUILD.bazel`, targets `all_proto` and `all_python_proto`. Also see updated PR description for details.

Neudrino (approved):

lte/protos/BUILD.bazel
```diff
@@ -454,3 +454,32 @@ python_grpc_library(
...
     protos = [":subscriberauth_proto"],
 )
+
+# This proto library is exclusively meant for the state service.
+# Users can manually add proto libraries to state.yml.
+# This means that all proto libraries need to be available
+# in builds of the state service.
+proto_library(
+    name = "all_proto",
+    srcs = glob(["*.proto"]),
```
electronjoe, line 464: My first thought was: `Generally in Bazel globbing of source is an anti-pattern (though OK if you are rapidly standing something up - generally needs to be refactored to remove globbing`) - but then with Sourcegraph search I found [171 instances of this type of globbing for protos](https://sourcegraph.com/search?q=context:global+srcs+%3D+glob%28%5B%22*.proto%22%5D%29%2C&patternType=literal) including some Google repos, so that seems OK.  I guess what still isn't confirmed and feels strange to me is to have more than one Bazel label referencing the same source (this feels a bit anti-DRY).  I do appreciate that you put the comment here so that others don't lazily include this target if un-necessary.

The RISK IMO is that people lazily use all_proto and then have much larger than necessary build graphs when iterating simple unit tests on targets needing only one or two protos.

Aaah And i see you have dealt with this risk by making visibility of all_python_proto only visible to `python/magma/state:__pkg__`.

So this all seems reasonable to me =D

electronjoe (approved):

lte/protos/BUILD.bazel
electronjoe: Edit: Can you imagine dealing with this cleanly and safely like this in lesser build systems? Toss this on the Bazel-win pile.

Neudrino: > Aaah And i see you have dealt with this risk by making visibility of all_python_proto only visible to `python/magma/state:__pkg__`.

:heavy_plus_sign: I have to agree, that this idea is quite good! :rocket:

nstng (approved):
in pairing

vktng (approved):

PR #11782, opened 2022-02-24
feat(agw): Add Neighbor Freq parameters and Neighbor Cell parameters to config the eNB from NMS.

hakeen0618:
Signed-off-by: hanhufeng <hanhufeng@baicells.com>

1. Add Neighbor Freq parameters, including:
- PLMN,Enodeb Neighbor Cell PLMN ID Consists of MCC and MNC.
- CELLID, Enodeb Neighbor Cell Identifier that identifies a cell.
- EARFCN
- PCI
- TAC
- Qoffset
- CIO
2. Add Neighbor Cell parameters, as following:
- EARFCN
- QoffsetRange
- QRxLevMinSib5
- PMax
- TReselectionEUTRA
- TReselectionEUTRASFMedium
- ReselThreshHigh
- ReselThreshLow
- Reselection_priority
 
## Test Plan

The parameter base on enodeb module,  unit test had already passed. 
code review from the cli command as following:
Orc8r: ./build --tests
Agw:
./precommit.py --build 
./precommit.py --lint --diff
nms:
yarn run eslint
yarn run flow
yarn run test 

Test  the parameter distribution process through the swagger API ,and test the adding adjacent cells and freq from NMS, the BTS side has seen the results is OK!

test enb devices : Nova 436q      software version:3.6.6  device_class : mBS31005    hardware version : B01

## Additional Information


lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
```diff
@@ -2456,10 +2456,155 @@ definitions:
           - 'Baicells Nova-233 G2 OD FDD'
           - 'Baicells Nova-243 OD TDD'
           - 'Baicells Neutrino 224 ID FDD'
+          - 'Baicells Nova-436q TDD'
           - 'Baicells ID TDD/FDD'
           - 'NuRAN Cavium OC-LTE'
           - 'FreedomFi One'
         x-nullable: false
+      NeighborFreqList:
```
andreilee: snake case
```suggestion
      neighbor_frequency_list:
```

lte/cloud/go/services/lte/servicers/protected/builder_servicer.go
```diff
@@ -467,6 +470,54 @@ func getEnodebConfigsBySerial(nwConfig *lte_models.NetworkCellularConfigs, gwCon
...
+			enbNeighborCell.QOffset = neighbor.QOffset
+			enbNeighborCell.Cio = neighbor.Cio
+			enbNeighborCell.Enable = neighbor.Enable
+			enbNeighborCell.Plmn = neighbor.Plmn
+			index := strconv.Itoa(int(neighbor.Index))
+			ret[index] = enbNeighborCell
+		}
+		return ret
+	}
+	return nil
```
andreilee: This should return an error if the config type is unmanaged - then handle that error in line 430/431

lte/gateway/python/magma/enodebd/device_config/configuration_init.py
```diff
@@ -332,6 +351,133 @@ def _set_tdd_subframe_config(
...
+                if neighbor_cell_x.cio:
+                    cfg.set_parameter_for_object(
+                        BaicellsParameterName.NEIGHBOR_CELL_CIO_N % i, neighbor_cell_x.cio,
+                        desired_object_name,
+                    )
+
+
+def _set_neighbor_freq_list(
+        cfg: EnodebConfiguration,
+        neighbor_freq_list: Any,
```
andreilee, line 418: Could you specify the type for `neighbor_freq_list`?

andreilee (changes requested):
Thanks for submitting these changes! I've added a few comments, mostly nit-level.

lte/gateway/python/magma/enodebd/device_config/configuration_init.py
hakeen0618: Yes, I will modify the type Any to pb2.Neighborfreqlist type.

lte/cloud/go/services/lte/servicers/protected/builder_servicer.go
hakeen0618: This had a  condition in line 419/420, when the condition is true, the config type must be mangaged - so it would not need to modify.

lte/cloud/go/services/lte/obsidian/models/swagger.v1.yml
hakeen0618: The variable should be neighbor_frequency_list, had modify and update.

lte/cloud/go/services/lte/servicers/protected/builder_servicer.go
hakeen0618: Already delete the if condition .

hakeen0618: This had a condition in line 419/420, when the condition is true, the config type must be mangaged - so already delete the if condition.

andreilee (changes requested):
Overall this looks great! One thing that I will require here is more comprehensive changes to `baicells_qrtb_tests.py`. The state machine for `BaiellsQRTBHandler` should be attempting to set the configuration for Neighbor Freq and Neighbor Cell parameters and that should be explicitly checked in the tests - whether the generated TR-069 messages match that.

PR #10759, opened 2021-12-09
fix(amf): Convert MLOG to OAILOG for all core transaction logic for n…

Chandhu-Wavelabs:
Signed-off-by: chandhu-wavelabs <chandhu.naga@wavelabs.ai>

## Summary

MLOGs in nas5g are converted to OAI LOGs and unwanted DEBUG logs are removed, 

## Test Plan

Tested with UERANSIM, NAS5G log level : DEBUG, INFO, WARNING and ERROR are working fine.
Sample_logs
UT_result1
UT_result2

## Additional Information

- [ ] This change is backwards-breaking

PR #11788, opened 2022-02-24
chore: remove LOG_WITH_GLOG compilation flag

themarwhal:
## Summary
https://github.com/magma/magma/pull/10759 removed OAI's dependency to magma_logging, which means that we can just log with GLOG by default. (Which was already the case for sctpd/liagentd/sessiond...)

This PR removes the conditional flag to toggle between glog and std::cout to avoid creating a dependency like OAI had to the library before in the future. (Before the aforementioned PR, OAI was not able to compile with LOG_WITH_GLOG set to 1)

## Test Plan
CI

## Additional Information

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -9,7 +9,6 @@ build:production --config=lsan --compilation_mode=dbg --strip=never --copt=-O3
 
 # C/C++ CONFIGS
 build --cxxopt=-std=c++14
-build --cxxopt=-DLOG_WITH_GLOG=1
 build --experimental_strict_action_env
 
 # MAGMA VM CONFIGS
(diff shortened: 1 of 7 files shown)


ardzoht (approved):
lgtm

nstng (approved):
Bazel changes lgtm.

I still see some `LOG_WITH_GLOG` in `CMakeLists.txt` files. I assume they are not relevant?

themarwhal:
> Bazel changes lgtm.
> 
> I still see some `LOG_WITH_GLOG` in `CMakeLists.txt` files. I assume they are not relevant?

Yes they are no longer relevant. Will clean those up in a separet PR.

PR #11836, opened 2022-02-28
feat(orc8r): Added protected gRPC ports to service registry

christinewang5:
# Summary
* changed service registry infra to support handling two grpc ports
* certifier service now runs on is own protected port
* updated relevant unit tests 

---- 
# Test Plan
added logs from servicer and from client to make sure they are the intended  ports 

## Deploying locally via Docker
```
directoryd stderr | E0314 18:06:02.775625      70 registry.go:118] christinewang -- AddServices LOCATION certifier @ localhost:9086 (protected: localhost:9186)
feg_relay stderr | E0314 18:06:03.079159      24 registry.go:118] christinewang -- AddServices LOCATION certifier @ localhost:9086 (protected: localhost:9186)
metricsd stderr | E0314 18:06:03.622512      31 registry.go:118] christinewang -- AddServices LOCATION certifier @ localhost:9086 (protected: localhost:9186)
obsidian stderr | E0314 18:06:03.658297      67 registry.go:118] christinewang -- AddServices LOCATION certifier @ localhost:9086 (protected: localhost:9186)
...
certifier stderr | E0314 18:06:04.299479      44 registry.go:118] christinewang -- AddServices LOCATION certifier @ localhost:9086 (protected: localhost:9186)
certifier stderr | E0314 18:06:04.299783      44 registry.go:118] christinewang -- AddServices LOCATION service_registry @ localhost:9180 (protected: localhost:9190)
...
dev_setup stderr | E0314 18:06:04.515080     149 registry.go:549] christinewang -- GetProtectedConnectionImpl -- GetProtectedServiceAddresss -- SERVICE certifier ADDR localhost:9186
```
=> grpc now has 9180 and 9190, certifier now runs on 9086 and 9186
## Deploying via Minikube & AWS
* logs from certifier pod
```
E0317 18:54:20.874366       1 registry.go:118] christinewang5 -- AddServices LOCATION certifier @ localhost:9086 (protected: localhost:9186)
...
E0317 18:54:20.893767       1 service.go:216] christinewang -- protected service port 9186
E0317 18:54:20.893997       1 service.go:200] christinewang -- service port 9086
```
this one's similar to the Docker logs, but has the correct ports as well
* logs from `kubectl exec -it --namespace orc8r deploy/orc8r-orchestrator -- \
/var/opt/magma/bin/accessc \
add-existing -admin -cert /var/opt/magma/certs/admin_operator.pem \
admin_operator`, which makes client calls to certifier service
```
E0317 15:24:59.006689      27 client_api.go:56] christinewang service_registry GetServiceAddress -- SERVICE certifier PROTECTED: true
**E0317 15:24:59.008837      27 registry.go:547] christinewang -- GetProtectedConnectionImpl -- GetProtectedServiceAddresss -- SERVICE certifier ADDR orc8r-certifier:9190**
E0317 15:24:59.028600      27 registry.go:201] christinewang -- GetServiceAddress service registry mode: k8s
E0317 15:24:59.028710      27 registry.go:469] christinewang -- GetConnectionImpl -- first try for existing connection -- SERVICE service_registry
E0317 15:24:59.028983      27 client_api.go:56] christinewang service_registry GetServiceAddress -- SERVICE accessd PROTECTED: false
E0317 15:24:59.029797      27 registry.go:478] christinewang -- GetConnectionImpl -- GetServiceAddresss -- SERVICE accessd ADDR orc8r-accessd:9180
```

Also ran certifier client unit tests at `orc8r/cloud/go/services/certifier/client_api_test.go`


cwf/k8s/cwf_operator/pkg/registry/registry.go
```diff
@@ -68,7 +68,7 @@ func (r *k8sConnectionRegistry) GetConnection(service string, port int) (*grpc.C
 	ctx, cancel := context.WithTimeout(context.Background(), GrpcMaxTimeoutSec*time.Second)
 	defer cancel()
 
-	return r.GetConnectionImpl(ctx, serviceAddr)
```
hcgatewood, line 71: From offline: let's use an enum here (so e.g. {southbound, internal}) rather than true/false. Is ready for change and more readable

feg/cloud/go/services/health/client_api.go
```diff
@@ -1,9 +1,7 @@
 /*
 Copyright 2020 The Magma Authors.
-
```
hcgatewood: Nit: can you fix your copyright setting in IntelliJ so that this doesn't happen

orc8r/cloud/go/services/service_registry/servicers/protected/docker_servicer.go
```diff
@@ -116,6 +117,13 @@ func (s *DockerServiceRegistryServicer) GetServiceAddress(ctx context.Context, r
 	// Since orc8r mesh services are attached to the same docker network,
 	// the services don't need to expose any ports. We can use a standardized
 	// port for the gRPC service address.
+	if req.Protected {
```
hcgatewood: From offline: this servicer is dead code, no need to edit it

orc8r/cloud/go/tools/migrations/m005_certifier_to_blobstore/main.go
```diff
@@ -76,7 +76,7 @@ func main() {
 func manuallyVerifyCertifierMigration() error {
 	// NOTE: we hard-code the certifier server location and bypass certifier/client_api.go
 	// to avoid using application code as much as possible (registry etc.)
-	certSrvAddr := "localhost:9086"
```
hcgatewood: From offline: not worth changing the migrations

orc8r/lib/go/registry/global_registry.go
```diff
@@ -117,6 +117,12 @@ func GetServicePort(service string) (int, error) {
 	return globalRegistry.GetServicePort(service)
 }
 
+// GetProtectedServicePort returns the listening port for the protected RPC service.
+// The service needs to be added to the registry before this.
+func GetProtectedServicePort(service string) (int, error) {
```
hcgatewood: From offline: consider converting these to using the same funcs but with enums. This should help avoid a lot of duplicated code

orc8r/protos/service_registry.proto
```diff
@@ -55,6 +55,7 @@ message FindServicesResponse {
 
 message GetServiceAddressRequest {
   string service = 1;
+  bool protected = 2;
```
hcgatewood: Same -- let's make an enum

cwf/gateway/services/uesim/test_init/test_service_init.go
```diff
@@ -26,9 +26,9 @@ import (
 
 func StartTestService(t *testing.T) {
 	factory := test_utils.NewSQLBlobstore(t, "uesim_test_service_blobstore")
-	srv, lis := test_utils.NewTestService(t, registry.ModuleName, registry.UeSim)
+	srv, lis, _ := test_utils.NewTestService(t, registry.ModuleName, registry.UeSim)
```
hcgatewood, line 29: So this PR introduces a lot of incidental changes due to changing the signature of NewTestService etc. Normally, we'd stack those changes in a lower diff, but since github makes it so painful I'm fine w lumping everything together

hcgatewood (commented):
Looks good! Couple small notes but overall seems like the right way to go about it

jkmar (approved):
Domain Proxy changes look ok

orc8r/lib/go/service/service.go
```diff
@@ -133,33 +139,84 @@ func NewServiceWithOptionsImpl(moduleName string, serviceName string, serverOpti
...
+	errChan := make(chan error)
+	go func() {
+		errChan <- service.run()
+		wg.Done()
+	}()
+	go func() {
+		errChan <- service.runProtected()
+		wg.Done()
+	}()
+	wg.Wait()
```
hcgatewood: I don't think this logic is quite right -- consider if 1 of the servicers returns an err and the other continues to run. Then only one of the wg.Done will fire, so the wg.Wait will never complete, so the 1 err will never be returned from this func

Lmk if I'm misreading

orc8r/lib/go/service/service.go
```diff
@@ -133,33 +139,84 @@ func NewServiceWithOptionsImpl(moduleName string, serviceName string, serverOpti
...
+	go func() {
+		if plis != nil {
+			err := service.ProtectedGrpcServer.Serve(plis)
+			if err != nil {
+				glog.Fatalf("failed to run test service: %v", err)
+			}
+		}
+		wg.Done()
+	}()
+	wg.Wait()
```
hcgatewood: Same here

orc8r/protos/service_registry.proto
```diff
@@ -53,8 +53,14 @@ message FindServicesResponse {
...
 }
 
+enum ServiceType {
+  SOUTHBOUND = 0;
+  PROTECTED = 1;
+}
+
 message GetServiceAddressRequest {
   string service = 1;
+  ServiceType serviceType = 2;
```
hcgatewood: serviceType -> service_type

(Ref: https://developers.google.com/protocol-buffers/docs/style)

cwf/gateway/registry/local_registry.go
```diff
@@ -14,6 +14,7 @@ limitations under the License.
 package registry
 
 import (
+	protos2 "magma/orc8r/lib/go/protos"
```
hcgatewood: Nit: I'm thinking "lib_protos" to (a) keep the refactor easy while still (b) using a descriptive name

hcgatewood (changes requested):
Cool lgtm minus the waitgroup logic

Also I was thinking maybe we should keep the certifier servicer change in this PR, to ensure the logic change is working. But I think it's fine this way actually -- we can make the certifier switchover in a separate PR 🙂

orc8r/lib/go/service/service.go
christinewang5: I think the wg has to be there for the two services to run simultaneously, otherwise two go routines will just fire and then it will proceed with the main go routine and exit. 

Or I could do something like 
```
go func () {
    service.runProtected()
}

service.run() 
```
And that would be equivalent I believe, because the last line would block the main go routine
Was loosely referencing this https://medium.com/rungo/running-multiple-http-servers-in-go-d15300f4e59f

Though it does only return one error, I changed it so that it returns both.

themarwhal (approved):
stamping for cwf/feg, deferring to @hcgatewood for lgtm

orc8r/cloud/go/service/service.go
```diff
@@ -134,14 +134,22 @@ func (s *OrchestratorService) Run() error {
...
+		if lis != nil {
+			serverErr <- s.GrpcServer.Serve(lis)
+		}
+	}()
+
+	go func() {
+		if plis != nil {
+			serverErr <- s.ProtectedGrpcServer.Serve(plis)
+		}
 	}()
```
hcgatewood: Nit: these if statements could be inverted ("if not nil then go func") for slight optimization, but is nbd

hcgatewood (approved):
Awesome! Lgtm 👍

hcgatewood:
Enabling auto-merge, assuming manually tested with reasonable confidence 🙂

PR #11845, opened 2022-02-28
chore: Add --compilation_mode=dbg by default to bazelrc

themarwhal:

## Summary
* --compilation_mode=dbg adds the compilation flag (-g) to include debug information. We include the debug symbols for production, so we would benefit from having this for development as well. 
* One thing that is note worthy is that this will enable -g for building dependencies as well. 
* Also adding `--test_output=errors` as default. This behavior is usually what we want 

## Additional Information

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -4,12 +4,15 @@ startup --output_base=/tmp/bazel
 build --announce_rc
 build --color=yes
 
-build:development --compilation_mode=fastbuild
-build:production --config=lsan --compilation_mode=dbg --strip=never --copt=-O3
+build:production --config=lsan --strip=never --copt=-O3
 
 # C/C++ CONFIGS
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -10,6 +10,7 @@ build:production --config=lsan --compilation_mode=dbg --strip=never --copt=-O3
 # C/C++ CONFIGS
 build --cxxopt=-std=c++14
 build --experimental_strict_action_env
+build --compilation_mode=dbg
```
nstng, line 12: nitpicking: comment for line 8: is `--compilation_mode=dbg` now redundant?

nstng (approved):

.bazelrc
themarwhal: yeah true, will clean this up

PR #11852, opened 2022-02-28
refactor(orc8r): Split Metricsd to protected and southbound

christinewang5:
## Additional Information


themarwhal (approved):

orc8r/cloud/go/services/metricsd/servicers/southbound/servicer.go
```diff
@@ -0,0 +1,125 @@
...
+		return &protos.Void{}, err
+	}
+	for _, e := range metricsExporters {
+		err := e.Submit(metricsToSubmit)
+		if err != nil {
+			glog.Error(err)
+		}
+	}
+	return new(protos.Void), nil
+}
```
hcgatewood, line 74: Nit here and below: newlines between func defns

orc8r/cloud/go/services/metricsd/servicers/protected/servicer.go
```diff
@@ -134,39 +87,11 @@ func preprocessCloudMetrics(family *prom_proto.MetricFamily, hostName string) ex
 	}
 	for _, metric := range family.Metric {
 		metric.Label = protos.GetDecodedLabel(metric)
-		addLabel(metric, metrics.CloudHostLabelName, hostName)
+		servicers.AddLabel(metric, metrics.CloudHostLabelName, hostName)
```
hcgatewood: Nit: move "helpers" like this to their own pkg -- we generally don't want to import logical fns like this from a servicers pkg (is confusing)

orc8r/cloud/go/services/metricsd/exporters/remote_exporter.go
```diff
@@ -40,3 +40,27 @@ func (r *remoteExporter) getExporterClient() (protos.MetricsExporterClient, erro
 	}
 	return protos.NewMetricsExporterClient(conn), nil
 }
+
+type TestMetricExporter struct {
```
hcgatewood: Nit: let's move this to a test (or test helpers) pkg. Generally speaking, don't want to put test-specific code in the same pkg as "normal" code, to avoid confusion

(Of course sometimes there's no other option, or it makes the most sense. But in this case I think we can drop it somewhere else)

hcgatewood (approved):
Lgtm, couple nits but otherwise msg when ready to merge

orc8r/cloud/go/services/metricsd/metricsd/main.go
```diff
@@ -47,8 +47,8 @@ func main() {
 		glog.Fatalf("Error creating orc8r service for metricsd: %s", err)
 	}
 
-	controllerServicer := servicers.NewMetricsControllerServer()
-	protos.RegisterMetricsControllerServer(srv.GrpcServer, controllerServicer)
+	controllerServicer := protected.NewCloudMetricsControllerServer()
+	protos.RegisterCloudMetricsControllerServer(srv.GrpcServer, controllerServicer)
```
hcgatewood: !

hcgatewood (changes requested):
Oh oops just kidding -- don't forget to add the new southbound servicer to the main.go as well!

(Both in main.go and in any test service startups)

orc8r/cloud/go/services/metricsd/exporters/remote_exporter.go
christinewang5: i just replicated this in the two places it was used in `servicer_southbound_test.go` and `servicer_protected_test.go`

orc8r/cloud/go/services/metricsd/servicers/protected/servicer.go
christinewang5: cool also replicated these helpers in both servicers

hcgatewood (approved):

Issue #11591, opened 2022-02-14
Find method for replacement of compile-time {PACKAGE_BUGREPORT, PACKAGE_NAME, PACKAGE_VERSION}


electronjoe:
Some of our logging macros in the MME leverage the above compiler defines as part of log line outputs.  This is problematic because in our Bazel build, having one large MME core target of `lte/gateway/c/core:core` (due to cyclic dependencies) - necessitates ~one definition at compile time for these defines (unless one does the awful and uses headers to over-write the defines - which ends up fragile and depends upon header ordering).

We will likely need a new way of expressing this logging output behavior - either splitting these MACROs into per-domain macros with static log line outputs... Or modifying our logging framework (a goal but maybe now is not the time).

I suppose an initial question is whether these log line outputs are even necessary or used functionally today (before we spend effort ensuring their maintenance).  We appear to have ~3 domains differentiated in logging - by differing CMakeLists.txt defines / header file over-writes today in MME.

electronjoe:
Discussion with Alex, Shruti, Ulas and Marie - we may just replace these with constant strings to mark the important fields.

electronjoe:
Regarding `PACKAGE_VERSION` I'm wondering if this is used in some way to bind a build release version into our outputs. I see something over in `.github/workflows/amis-workflow.yml`? [Sourcegraph](https://sourcegraph.com/github.com/magma/magma/-/blob/.github/workflows/amis-workflow.yml?L88:56&subtree=true).

nstng:
Our understanding:
* These values are only set via `CMakeList.txt` s. There are only three cases. `PACKAGE_VERSION` and `PACKAGE_BUGREPORT` seem to be only dummy values, i.e., piratically unused.
* Could be solved by creating packages - should this be tried or do you see the circular dependencies as happening here?

themarwhal:
> Our understanding:
> 
> * These values are only set via `CMakeList.txt` s. There are only three cases. `PACKAGE_VERSION` and `PACKAGE_BUGREPORT` seem to be only dummy values, i.e., piratically unused.
> * Could be solved by creating packages - should this be tried or do you see the circular dependencies as happening here?

Yes I think we were worried about circular dependencies here. (But maybe it might not be a problem, I am not sure)

themarwhal:
I will investigate using per_file_copt for this!

PR #11862, opened 2022-03-01
chore(mme): Adds Bazel build target for oai/mme

electronjoe:
Builds (unvalidated / untested) oai_mme binary BUT has a number of limitations, enumerated in GH Issues and enumerated in TODOs within the Bazel build label. This PR is preparing us for unit test commits which will stack on top, supporting execution of ~all MME unit tests via Bazel.

Limitations are dominated by precompiler defines which we set for the entire OAI Core target (and which sometimes have differentiation by sub-modules in existing CMake builds - for better and for worse).

### Pre-Req for `bazel-base` container

- [x] Requires #11889 for successful `bazel build //lte/gateway/c/core:oai_mme` from within `bazel-base` container.
- [x] Requires #11899 for successful `bazel build //lte/gateway/c/core:oai_mme` from within `bazel-base` and `devcontainer` containers while supporting Magma test VM.

## Test Plan

- [x] Bazel build of new label in Devcontainer
- [x] Bazel build of new label in [bazel-base](https://github.com/magma/magma/tree/master/experimental/bazel-base) container
- [x] Bazel build of new label in Magma Test VM - by proxy support =D

`bazel build //lte/gateway/c/core:oai_mme`
`bazel test //lte/gateway/c/core:oai_mme` (builds with differing precompiler define allowing unit tests)


lte/gateway/c/core/BUILD.bazel
```diff
@@ -0,0 +1,1163 @@
...
+        "-DASN1_MINIMUM_VERSION=923",  # TODO(#11860) ASN1_MINIMUM_VERSION should just be removed
+        "-DPACKAGE_BUGREPORT=\\\"PACKAGE_BUGREPORT\\\"",  # TODO(#11591): needs deprecation
+        "-DPACKAGE_NAME=\\\"PACKAGE_NAME\\\"",  # TODO(#11591): needs deprecation
+        "-DPACKAGE_VERSION=\\\"0.1\\\"",  # TODO(#11591): needs deprecation
+        "-DCMAKE_BUILD_TYPE=\\\"CMAKE_BUILD_TYPE\\\"",  # TODO(#11591): needs deprecation
+        "-DS1AP_DEBUG_LIST=false",  # TODO: Determine if this is even used / merits commit to code base
+        "-DSCTP_DUMP_LIST=false",  # TODO: Determine if this is even used / merits commit to code base
+        "-DTRACE_HASHTABLE=false",  # TODO: Determine if this is even used / merits commit to code base
+        "-DTRACE_3GPP_SPEC=false",  # TODO: Determine if this is even used / merits commit to code base
+        "-DMME_UNIT_TEST=true",  # TODO(#11861) MME_UNIT_TEST prevents use in production, needs fix
```
electronjoe: Now un-necessary, remove

themarwhal (approved):

lte/gateway/c/core/BUILD.bazel
```diff
@@ -0,0 +1,1163 @@
...
+    prefix = "S1ap_",
+)
+
+cc_asn1_library(
+    name = "asn1_r16",
+    asn1_file = "oai/tasks/ngap/messages/asn1/r16/r16.asn1",
+    prefix = "Ngap_",
+)
+
+// TODO(#11892): Giant Bazel label is bad practice, break this up
```
nstng: needs # for comment - breaks the bazel builds in CI

lte/gateway/c/core/BUILD.bazel
```diff
@@ -0,0 +1,1163 @@
...
+        "oai/tasks/sms_orc8r/sms_orc8r_messages.h",
+    ],
+    copts = [
+        "-DEMBEDDED_SGW",  # TODO(#11638): Feature flags fixups
+        "-DS6A_OVER_GRPC",  # TODO(#11638): Feature flags fixups
+        "-DASN1_MINIMUM_VERSION=923",  # TODO(#11860) ASN1_MINIMUM_VERSION should just be removed
+        "-DPACKAGE_BUGREPORT=\\\"PACKAGE_BUGREPORT\\\"",  # TODO(#11591): needs deprecation
+        "-DPACKAGE_NAME=\\\"PACKAGE_NAME\\\"",  # TODO(#11591): needs deprecation
+        "-DPACKAGE_VERSION=\\\"0.1\\\"",  # TODO(#11591): needs deprecation
+        "-DCMAKE_BUILD_TYPE=\\\"CMAKE_BUILD_TYPE\\\"",  # TODO(#11591): needs deprecation
```
nstng: regarding TODO: is this really #11591 related?

lte/gateway/c/core/BUILD.bazel
```diff
@@ -0,0 +1,1163 @@
...
+        "-DS1AP_DEBUG_LIST=false",  # TODO: Determine if this is even used / merits commit to code base
```
nstng: Should these also be tracked in issues?

nstng (changes requested):

lte/gateway/c/core/BUILD.bazel
```diff
@@ -0,0 +1,1163 @@
+# Copyright 2021 The Magma Authors.
```
nstng: nit: 2022?

lte/gateway/c/core/BUILD.bazel
```diff
@@ -0,0 +1,1163 @@
...
+    name = "oai_mme",
+    srcs = [
+        "oai/oai_mme/oai_mme.c",
+    ],
+    deps = [
+        ":core",
+    ],
+)
+
+cc_library(
```
nstng, line 1152: Why not in `oai/common/BUILD.bazel`?
Ahh I see - then all common src files need to be declared there - OK

nstng:
What is the current expectation for running the binary? Currently I get (on devcontainer)
```
./bazel-bin/lte/gateway/c/core/oai_mme
Segmentation fault (core dumped)
```

lte/gateway/c/core/BUILD.bazel
electronjoe: Yes - you got it. Per discussion in #11892 we cannot have a module in oai/common/ unless all files can be in oai/common/ and unfortunately there are cycles with some components in oai/common that require inclusion in `//lte/gateway/c/oai/core:core` today.  What I would advise (and we could do this as soon as Bazelification is complete and CMake is turned down) is to move `oai/common/sentry_log.{h,cpp}` to a new directory and then give it a module in that directory to untangle this issue.

lte/gateway/c/core/BUILD.bazel
electronjoe: Whoops! Added that late last night and didn't re-test LOL. Thanks.

lte/gateway/c/core/BUILD.bazel
electronjoe: Good point. No I think it's a separate issue (we really don't have these semantic build types in Bazel).  I'll create a new issue to track.

lte/gateway/c/core/BUILD.bazel
electronjoe: Good point. Done!

electronjoe:
> What is the current expectation for running the binary? Currently I get (on devcontainer)

The only binary testing I have done is to run it by hand (as you have) and note the "normal" behavior in which the binary crashes trying to load configuration files that are missing.  I did not explore further as I wanted to have all unit tests healthy first (that seemed a more efficient first step) and we haven't completed that task yet.  Ultimately we'll want to pass this binary through the lte integ test before it's considered healthy.

I could omit this binary from the PR but without it vast portions of linking success are not confirmed / validated (the `//lte/gateway/c/core:core` `cc_library` will only compile the c/cpp but will not attempt linking without being consumed by a target). So it's more a test than a functional binary that is tested.  I investigated labeling the `cc_binary` of `//lte/gateway/c/core:oai_mme` as `testonly = True,` - but this doesn't do what we want here.

To your point I can at least immediately call out in this PR description that the build artifact is untested? I could also drop a GH issue in BUILD.bazel if you like. Let me know what you think.

electronjoe:
I will update this PR's `system_libraries.BUILD` paths for `libnettle` and `libgnutls` once #11899 merges, for final merge here.

nstng:
> > What is the current expectation for running the binary? Currently I get (on devcontainer)
> 
> I could omit this binary from the PR but without it vast portions of linking success are not confirmed / validated (the `//lte/gateway/c/core:core` `cc_library` will only compile the c/cpp but will not attempt linking without being consumed by a target). So it's more a test than a functional binary that is tested. I investigated labeling the `cc_binary` of `//lte/gateway/c/core:oai_mme` as `testonly = True,` - but this doesn't do what we want here.
> 
> To your point I can at least immediately call out in this PR description that the build artifact is untested? I could also drop a GH issue in BUILD.bazel if you like. Let me know what you think.

Feel free to leave the binary target in the build file - I was just curious if it is already runable. As parts of this will be changed I think it is a good starter for future work to have the binary already present. (We have the call out in this PR now by our discussion :) ).

nstng (approved):

PR #11885, opened 2022-03-01
chore(ci): modify Bazel build and test job so that we only start from…

themarwhal:
… scratch on master

## Summary
1. Noticed that the Bazel build and test job was starting from scratch very often on master. This is probably because the cutoff value to remove the cache is now too small. (Maybe due to all the python dependencies we're adding). 
2. Modified the logic so that we only rebuild the cache on master. This change makes it so that PR authors are not affected by the occasional slow build.
3. Separated out the cache wiping logic for Bazel-cache-repo and bazel-cache. Bazel-cache-repo should rarely need a cache wipe as it only holds the source code of our dependencies, which should not change frequently.

## Test Plan
CI (will monitor master)

## Additional Information

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -81,20 +81,41 @@ jobs:
         # thereby only ever using the last successful cache version. This solution will result in a
         # few slower CI actions around the time cache is detected to be too large, but it should
         # incrementally improve thereafter.
-        - name: Ensure cache size
+        - name: Ensure cache size BAZEL_CACHE
+          # Only run on master to avoid slow build on PRs
+          if: github.event_name == 'push'
           env:
             BAZEL_CACHE_DIR: .${{ env.BAZEL_CACHE }}
-            BAZEL_CACHE_REPO_DIR: .${{ env.BAZEL_CACHE_REPO }}
+            # Use a 7GB threshold since actions/cache compresses the results, and Bazel caches seem
+            # to only increase by a few hundred megabytes across changes for unrelated branches.
+            # Uncompressed cache on master is looking to be around 6GB (from observing jobs on master)
+            BAZEL_CACHE_CUTOFF_MB: 7000
           run: |
             # See https://stackoverflow.com/a/27485157 for reference.
             EXPANDED_BAZEL_CACHE_PATH="${BAZEL_CACHE_DIR/#\~/$HOME}"
+
+            CACHE_SIZE_MB=$(du -smc $EXPANDED_BAZEL_CACHE_PATH | cut -f1)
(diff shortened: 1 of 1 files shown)


.github/workflows/bazel.yml
```diff
@@ -81,20 +81,41 @@ jobs:
         # thereby only ever using the last successful cache version. This solution will result in a
         # few slower CI actions around the time cache is detected to be too large, but it should
         # incrementally improve thereafter.
-        - name: Ensure cache size
+        - name: Ensure cache size BAZEL_CACHE
+          # Only run on master to avoid slow build on PRs
```
electronjoe, line 85: Good idea!

electronjoe (approved):

nstng (approved):
lgtm

nstng:
Just to be sure - in order to monitor if the thresholds should be adapted, one can monitor https://github.com/magma/magma/actions/workflows/bazel.yml and check if there are "too much" long running runs, right?

themarwhal:
> Just to be sure - in order to monitor if the thresholds should be adapted, one can monitor https://github.com/magma/magma/actions/workflows/bazel.yml and check if there are "too much" long running runs, right?

Yep that's correct. Will document this.

.github/workflows/bazel.yml
```diff
@@ -81,20 +81,41 @@ jobs:
...
-            BAZEL_CACHE_REPO_DIR: .${{ env.BAZEL_CACHE_REPO }}
+            # Use a 7GB threshold since actions/cache compresses the results, and Bazel caches seem
+            # to only increase by a few hundred megabytes across changes for unrelated branches.
+            # Uncompressed cache on master is looking to be around 6GB
+            BAZEL_CACHE_CUTOFF_MB: 7000
           run: |
             # See https://stackoverflow.com/a/27485157 for reference.
             EXPANDED_BAZEL_CACHE_PATH="${BAZEL_CACHE_DIR/#\~/$HOME}"
+
+            CACHE_SIZE_MB=$(du -smc $EXPANDED_BAZEL_CACHE_PATH | grep total | cut -f1)
```
Neudrino:
```suggestion
            CACHE_SIZE_MB=$(du -sm $EXPANDED_BAZEL_CACHE_PATH | cut -f1)
```
If we give only one argument to `du`, we should be able to avoid `--total / -c` and with that the grep, afaiks.

Neudrino: _FYI a side note:_ `du` seems to be localized. I.e. on my German system `du -c` produces a line with the string `insgesamt` instead of `total`.

PR #11911, opened 2022-03-02
chore(mme): Cleanup irrelevant logs and unused variable

ssanadhya:

## Summary

Remove some irrelevant logs, such as:
- Recovery message is not handled by S1ap task since it does not have any pending timers at restart, downgrading the error to debug
` ERROR S1AP   tasks/s1ap/s1ap_mme.c           :0297    Unknown message ID 11:RECOVERY_MESSAGE`
- Thread id reading SPGW task 
- Unused variable in `test_pgw_pco.cpp`

## Test Plan

`make test_oai`


ardzoht (approved):
lgtm

PR #11924, opened 2022-03-03
feat(dp): Add total count to cbsd list

jkmar:
## Summary

Previously cbsd list returned only list of cbsds,
but this made it impossible to do pagination,
since the total number of pages was unknown.


m-trojanowski (approved):

WojSad (approved):

jkmar:
Bazel is broken even on master, and fails CI are not related to changes of this pr at all. Bazel fails even on pr, which just changes the documentation: #11940

PR #11939, opened 2022-03-03
chore: try separating out build and test for bazel

themarwhal:

## Summary
Out current hypothesis is that if we run `bazel build //...` and `bazel test //...` back to back, we exhaust the available memory allowed in the GitHub action nodes. See discussion: https://magmacore.slack.com/archives/C02HRAXJEBB/p1646322360021719

I'm wondering if the test only action env added in https://github.com/magma/magma/pull/11874 is causing a very large cache as it produces different set of build actions for `bazel build //...` and `bazel test //...`. (This was expected in theory, but we did not expect CI to break :p ). 

From testing on this PR, it helps to make the two actions run in parallel. This might be a good quick remediation solution.

## Test Plan
CI

## Additional Information

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -47,101 +47,163 @@ jobs:
               - '**/*.bazel'
               - '**/*.bzl'
 
-  bazel_build_and_test:
+  bazel_build:
     needs: path_filter
     # Only run workflow if this is a scheduled run on master branch,
     # or a pull_request that skip-duplicate-action wants to run again.
     if: |
       (github.event_name == `schedule` && github.ref == 'refs/heads/master')
         || ${{ needs.path_filter.outputs.files_changed == 'true' }}
-    name: Bazel Build and Test Job
+    name: Bazel Build Job `bazel build //...`
     runs-on: ubuntu-latest
     steps:
-        - name: Check Out Repo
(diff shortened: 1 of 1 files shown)


.github/workflows/bazel.yml
```diff
@@ -135,8 +135,70 @@ jobs:
...
+    needs: path_filter
+    # Only run workflow if this is a scheduled run on master branch,
+    # or a pull_request that skip-duplicate-action wants to run again.
+    if: |
+      (github.event_name == `schedule` && github.ref == 'refs/heads/master')
+        || ${{ needs.path_filter.outputs.files_changed == 'true' }}
+    name: Bazel Test Job `bazel test //...`
+    runs-on: ubuntu-latest
+    steps:
+        - name: Check Out Repo
```
electronjoe: Is the indentation linting finding real?

electronjoe (approved):

electronjoe:
Does TNG have any advice on GH Action re-use for this PR? Is it possible?

.github/workflows/bazel.yml
themarwhal: weird! will investigate

themarwhal: looks like the entire file was off by two

PR #11953, opened 2022-03-03
chore: add lte/gateway/c/core:oai_mme to VScode configs

themarwhal:

## Summary
Add the new OAI target to bzl config. This will allow us to generate the necessary compile_commands.json file that include OAI to enable clangd navigation.

## Test Plan
Genearted a codespace from this PR, generated a compilation database file and ran clangd.

Select: "Bzl: Bazel C/C++ Generate Compilation Database"
Screen Shot 2022-03-03 at 6 57 56 PM

The generation should pop up and should take about less than a minute
Screen Shot 2022-03-03 at 6 58 51 PM

Restart clangd
Screen Shot 2022-03-03 at 6 59 17 PM

No more squiggly lines, code navigation works
Screen Shot 2022-03-03 at 7 00 28 PM

## Additional Information

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -69,7 +69,8 @@
 			"//lte/gateway/c/session_manager:sessiond",
 			"//lte/gateway/c/sctpd/src:sctpd",
 			"//lte/gateway/c/li_agent/src:liagentd",
-			"//lte/gateway/c/connection_tracker/src:connectiond"
+			"//lte/gateway/c/connection_tracker/src:connectiond",
+			"//lte/gateway/c/core:oai_mme",
 		],
 		"multiCommand.commands": [
 			{
--- a/vscode-workspaces/workspace.magma-vm-workspace.code-workspace
+++ b/vscode-workspaces/workspace.magma-vm-workspace.code-workspace
@@ -45,6 +45,7 @@
 			"//lte/gateway/c/sctpd/src:sctpd",
 			"//lte/gateway/c/li_agent/src:liagentd",
 			"//lte/gateway/c/connection_tracker/src:connectiond",
+			"//lte/gateway/c/core:oai_mme",
 		],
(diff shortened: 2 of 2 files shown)


nstng:
Nice improvement, thank you! This will help with working on the mme tests.
Questions for the user workflow:
* when is a `bzl gen` or `clangd` restart necessary?
  * rebuild of devcontainer, adding new functions, ...

nstng (approved):
LGTM - tested in dev container

LKreutzer (approved):
Tested locally.

electronjoe (approved):

electronjoe (changes requested):
The sign-off of the PR is to `Signed-off-by: GitHub [noreply@github.com](mailto:noreply@github.com)`

I was about to merge this so we could party on it =D

themarwhal:
> Nice improvement, thank you! This will help with working on the mme tests. Questions for the user workflow:
> 
> * when is a `bzl gen` or `clangd` restart necessary?
>   
>   * rebuild of devcontainer, adding new functions, ...

@nstng 
`bzl gen` is necessary everytime a new codespace starts (for the most part). The task outputs a `compile_commands.json` file at the root directory and that is all clangd needs. I would guess you would probably have to rerun if you add a new dependency, I'm not sure if clangd is smart enough to figure out dependencies about new files (maybe?). Definitely do not need to trigger it for modifying an existing file.

`clangd` restart is needed everytime the compilation commands file is modified. (So whenever `bzl gen` is run, I always restart it afterwards).

We might be able to improve this workflow a bit by generating the compile_commands file on codespaces create, but if we go that route it might just be simpler to write our own script. (envoy has a script that does the same thing: https://sourcegraph.com/github.com/envoyproxy/envoy/-/blob/tools/gen_compilation_database.py)

^^ this approach might end up being better so that we can include test targets easily too.

electronjoe (approved):

PR #11963, opened 2022-03-04
chore(mobilityd): Fix type errors shown by mypy

sebathomas:
## Summary

I tried static type checking of mobilityd using mypy and fixed some of the errors it found. Mostly these were minor issues (missing return statements or null checks).

The only bigger change is the introduction of dedicated types for IP addresses and networks. Before this PR, they referred to functions as variable types, which is not a valid type hint.

## Test Plan

```
cd $MAGMA_ROOT/lte/gateway
mypy --ignore-missing-imports python/magma/mobilityd/
```

Error count on master:
```
Found 108 errors in 16 files (checked 31 source files)
```

Error count after the changes:
```
Found 24 errors in 5 files (checked 31 source files)
```

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/mobilityd/dhcp_desc.py
+++ b/lte/gateway/python/magma/mobilityd/dhcp_desc.py
@@ -94,6 +94,7 @@ class DHCPDescriptor:
         if self.state == DHCPState.OFFER or self.state == DHCPState.REQUEST \
                 or self.state == DHCPState.ACK:
             return self.ip
+        return None
 
     def ip_is_allocated(self) -> bool:
         """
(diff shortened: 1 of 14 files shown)


ardzoht (approved):
lgtm

themarwhal:
Thanks for this Sebastian!!! Is there a way to add linters / test coverage for these types of things?

sebathomas:
> Thanks for this Sebastian!!! Is there a way to add linters / test coverage for these types of things?

I'd love to have it in the CI. It would be easy to set up if we could fix all warnings first and then keep it green. Getting to a green state might take a lot of work though.

I'm not sure if there is any support for a mode with existing issues, e.g. where it checks that the number of errors doesn't increase so you could lower the allowed error count over time.

I'll look into it a bit more...

PR #11975, opened 2022-03-04
fix(agw): Fixed wrong LTE integ test status in CI dashboard

### LINKED ISSUE #11972: [CI] LTE integ test continuous runs do not reflect failures on the GH workflow
### Your Environment

- **Version:** master
- **Affected Component:** CI

### Describe the Issue

- With the addition of https://github.com/magma/magma/pull/11226, the LTE integ test runs on CI are now continuous. Any failure on a test case will not block the run to subsequent test cases. 
- Due to this change, the workflow of LTE integ test is publishing success on the dashboard even though the failure is reported. This is because the workflow does not take into account the overall results of the test cases but rather the status is published based on the success of the reports being published.
- The GH workflow https://github.com/magma/magma/blob/master/.github/workflows/lte-integ-test.yml needs to be updated with an additional step to obtain the results of the test cases and update the report on CI based on those. 

**Screenshots**
image

VinashakAnkitAman:
## Title
fix(agw): Fixed wrong LTE integ test status in CI dashboard

## Summary
The recent CI runs always show successful status even if some of the test cases are failing (Reference Github Issue: #11972 ). This PR keeps the status of the LTE integ test in a file which is then read after CI execution to decide the final verdict.

## Test plan
Locally tested with  set of passing and failing test cases


lte/gateway/python/integ_tests/Makefile
```diff
@@ -45,7 +45,8 @@ integ_test: $(PYTHON_BUILD)/setupinteg_env $(BIN)/nosetests
 ifdef TESTS
 	$(call execute_test,$(TESTS))
 else
-	$(foreach test,$(EXTENDED_TESTS) $(PRECOMMIT_TESTS),$(call execute_test,$(test));)
+	echo "pass" > $(MAGMA_ROOT)/test_status.txt
+	$(foreach test,$(JUNK_TESTS),$(call execute_test,$(test));)
```
ardzoht: where does `JUNK_TESTS` come from?

VinashakAnkitAman: Fixed it. That was a mistake while committing the file was overwritten by some temporary changes.

ardzoht (approved):

tmdzk (approved):

PR #12009, opened 2022-03-08
chore: upgrade yarn to 2+

lucasgonze:
## Additional Information


lucasgonze:
This was an experiment to see what would happen if I upgraded yarn. Result: some actions need to be adjusted.

Closing for now.

PR #12012, opened 2022-03-08
chore: Fix type errors and mypy.ini

nicarl:
## Summary

I added some more type hints based on the work from #11963. 
Furthermore, I created an MVP for a `mypy.ini`, which is a first step for adding a type checking step to the CI. mypy currently ignores quite some folders and files, this can be fixed step by step.

Locking down mypy settings in a `mypy.ini` is generally recommended as it specifies the rules mypy uses. They are currently rather loose. After decreasing the list of excluded files and folders one can increase the strictness of mypy.

We should check that these changes are compatible with https://github.com/magma/magma/pull/12008

## Test Plan

``` bash
cd $MAGMA_ROOT
mypy lte/gateway/python/
```
Before:
```
python/integ_tests/s1aptests/test_attach_detach.py: error: Duplicate module named "test_attach_detach" (also at "python/integ_tests/federated_tests/s1aptests/test_attach_detach.py")
python/integ_tests/s1aptests/test_attach_detach.py: note: Are you missing an __init__.py? Alternatively, consider using --exclude to avoid checking one of them.
Found 1 error in 1 file (errors prevented further checking)
```

After:
``` 
Success: no issues found in 424 source files
```

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
+++ b/lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
@@ -24,9 +24,10 @@ import logging
 from copy import deepcopy
 from ipaddress import ip_address, ip_network
 from threading import Condition
-from typing import List
+from typing import List, Optional, cast
 
 from magma.mobilityd.ip_descriptor import IPDesc, IPState, IPType
+from typing_extensions import TypeGuard
 
 from .dhcp_client import DHCPClient
 from .dhcp_desc import DHCPDescriptor, DHCPState
@@ -205,17 +206,17 @@ class IPAllocatorDHCP(IPAllocator):
 
                 if retry_count % DEFAULT_DHCP_REQUEST_RETRY_FREQUENCY == 0:
                     self._dhcp_client.send_dhcp_packet(
-                        mac, vlan,
+                        mac, str(vlan),
                         DHCPState.DISCOVER,
                     )
                 self.dhcp_wait.wait(timeout=DEFAULT_DHCP_REQUEST_RETRY_DELAY)
 
-                dhcp_desc = self._dhcp_client.get_dhcp_desc(mac, vlan)
+                dhcp_desc = self._dhcp_client.get_dhcp_desc(mac, str(vlan))
 
                 retry_count = retry_count + 1
 
-            return dhcp_desc
+            return cast(DHCPDescriptor, dhcp_desc)
 
 
-def dhcp_allocated_ip(dhcp_desc) -> bool:
+def dhcp_allocated_ip(dhcp_desc: Optional[DHCPDescriptor]) -> TypeGuard[DHCPDescriptor]:
     return dhcp_desc is not None and dhcp_desc.ip_is_allocated()
--- a/lte/gateway/python/magma/mobilityd/dhcp_client.py
+++ b/lte/gateway/python/magma/mobilityd/dhcp_client.py
@@ -91,8 +91,8 @@ class DHCPClient:
     def send_dhcp_packet(
         self, mac: MacAddress, vlan: str,
         state: DHCPState,
-        dhcp_desc: DHCPDescriptor = None,
-    ):
+        dhcp_desc: Optional[DHCPDescriptor] = None,
+    ) -> None:
         """
         Send DHCP packet and record state in dhcp_client_state.
 
@@ -113,7 +113,7 @@ class DHCPClient:
             )
             self._msg_xid = self._msg_xid + 1
             pkt_xid = self._msg_xid
-        elif state == DHCPState.REQUEST:
+        elif state == DHCPState.REQUEST and dhcp_desc:
             dhcp_opts = [
                 ("message-type", "request"),
                 ("requested_addr", dhcp_desc.ip),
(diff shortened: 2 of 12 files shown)


lte/gateway/python/mypy.ini
```diff
@@ -0,0 +1,21 @@
+[mypy]
+python_version = 3.7
```
nicarl: Here I was not sure, 3.7. was a wild guess.

sebathomas: It's 3.8. :)

lte/gateway/python/mypy.ini
```diff
@@ -0,0 +1,21 @@
...
+    ^python/magma/enodebd/ | 
+    ^python/magma/subscriberdb/protocols/diameter/ | 
+    ^python/magma/kernsnoopd/ | 
+    ^python/precommit.py$ | 
+    ^python/scripts/agw_health_cli.py$ |
+    ^python/load_tests.common.py$ |
+    ^python/scripts/generate_oai_config.py$ |
+    ^python/integ_tests/s1aptests/s1ap_utils.py$ |
+    ^python/integ_tests/s1aptests/test_attach_detach_non_nat_dp_ul_tcp.py$ |
+    ^python/integ_tests/s1aptests/util/traffic_messages.py$ |
```
sebathomas: Except the integration tests, which are still executed on 3.5 Might make sense to keep them excluded for the moment.

lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
```diff
@@ -24,9 +24,10 @@
 from copy import deepcopy
 from ipaddress import ip_address, ip_network
 from threading import Condition
-from typing import List
+from typing import List, Optional, cast
 
 from magma.mobilityd.ip_descriptor import IPDesc, IPState, IPType
+from typing_extensions import TypeGuard
```
sebathomas, line 30: I didn't know mypy had introduced type guards. :partying_face: 

We'll probably need the typing_extensions package in `setup.py` to use it though.

sebathomas (commented):
Thanks so much for looking into this! Looking good, except a few minor points. Just started the CI.

lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
sebathomas: Looks like I was wrong, no additional dependency needed.

sebathomas (changes requested):
The existence of __init__.py in integ_tests breaks the integration tests. Best not touch that folder for now.

nicarl:
> The existence of **init**.py in integ_tests breaks the integration tests. Best not touch that folder for now.

I deleted the `__init__.py` and excluded the integration tests from the type checking.

sebathomas (approved):
Thank you! :+1:
(Everything looks good to me, just started the CI again.)

lte/gateway/python/mypy.ini
```diff
@@ -0,0 +1,19 @@
+[mypy]
+python_version = 3.8
+ignore_missing_imports = True
+show_error_codes = True
+warn_unused_configs = True
+ignore_errors = False
```
Neudrino: Could we add an option `namespace_packages = True` here to deal with packages without the `__init__.py`.

lte/gateway/python/mypy.ini
```diff
@@ -0,0 +1,19 @@
...
+    ^python/magma/subscriberdb/protocols/diameter/ | 
+    ^python/magma/kernsnoopd/ | 
+    ^python/precommit.py$ | 
+    ^python/scripts/agw_health_cli.py$ |
+    ^python/load_tests.common.py$ |
+    ^python/scripts/generate_oai_config.py$ |
+    ^python/setup.py$ |
+    ^python/scripts/state_cli.py$ |
+    ^python/integ_tests/ 
+    )
```
Neudrino: Why placing this file within `lte/gateway/python/` and not in the repository root?
If at all, the Github action (#12008) would probably expect it to be at the repository root in order to load it (not tested so far).

Neudrino: _minor styling_: There are some trailing whitespaces in this section.

lte/gateway/python/mypy.ini
Neudrino: You would probably need to adopt the `    ^python/magma/subscriberdb/protocols/diameter/ |` line and remove the `diameter` subfolder from it, though.

Neudrino: And maybe also the
```
install_types = True
non_interactive = True
```

lte/gateway/python/mypy.ini
nicarl: I moved the file to the repository root :+1:

lte/gateway/python/mypy.ini
nicarl: As discussed yesterday I moved the flags into the `mypy.ini`

Neudrino (approved):

nicarl:
The PR is ready for merging from my side. Unfortunately, I can not set the corresponding label.

lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
ardzoht: @sebathomas seems like we do need it, on .deb magma package there's an issue due to the missing dependency

PR #12017, opened 2022-03-08
chore(ci): Fix GHA cache pulling logic

themarwhal:

## Summary
The cache key and restore-keys fields should match. Otherwise, we could be pulling in the wrong cache. Due to this bug on master, we're pulling in any cache element that matches the pattern, which can be wrong.

For example this job should be pulling in the cache item for test_all, but pulls in the gcc warning cache instead. 
https://github.com/magma/magma/runs/5472340333?check_suite_focus=true
Screen Shot 2022-03-08 at 6 02 51 PM

## Test Plan
CI

## Additional Information


tmdzk (approved):

nstng (approved):
compared logging of different runs - LGTM :+1:

PR #12022, opened 2022-03-09
fix(orc8r): Update gateway device information during registration

christinewang5:
Previously, the gateway was registered to the device service but the gateway itself did not have the updated device info (i.e. hardware id and challenge key, etc.). This caused checkin_cli.py to fail bc the gateway was half registered.

Fixed so that the device field is updated upon calling register.py

Tested by spinning up orc8r, this was after running the register.py script
image

### DIFF
--- a/orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
+++ b/orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
@@ -11,6 +11,7 @@ import (
 	"magma/orc8r/cloud/go/orc8r"
 	"magma/orc8r/cloud/go/serdes"
 	"magma/orc8r/cloud/go/services/bootstrapper"
+	"magma/orc8r/cloud/go/services/configurator"
 	"magma/orc8r/cloud/go/services/device"
 	"magma/orc8r/cloud/go/services/orchestrator/obsidian/models"
 	"magma/orc8r/cloud/go/services/tenants"
@@ -44,6 +45,12 @@ func (r *RegistrationService) Register(c context.Context, request *protos.Regist
 		return clientErr, nil
 	}
 
+	err = updateGatewayDevice(c, deviceInfo, request.Hwid, request.ChallengeKey)
+	if err != nil {
+		clientErr := makeErr(fmt.Sprintf("error updating gateway: %v", err))
+		return clientErr, nil
+	}
+
 	err = r.RegisterDevice(*deviceInfo, request.Hwid, request.ChallengeKey)
 	if err != nil {
 		clientErr := makeErr(fmt.Sprintf("error registering device: %v", err))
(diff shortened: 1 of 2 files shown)


orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
```diff
@@ -114,3 +119,43 @@ func makeErr(errString string) *protos.RegisterResponse {
...
+func updateGatewayDevice(ctx context.Context, deviceInfo *protos.GatewayDeviceInfo, hwID *protos.AccessGatewayID, challengeKey *protos.ChallengeKey) error {
+	networkID := deviceInfo.NetworkId
+	gatewayID := deviceInfo.LogicalId
+
+	ent, err := configurator.LoadEntity(
+		ctx,
+		networkID, orc8r.MagmadGatewayType, gatewayID,
+		configurator.EntityLoadCriteria{},
+		serdes.Entity,
+	)
```
hcgatewood, line 140: I think there's a "load entity by physical ID" func

orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
```diff
@@ -114,3 +119,43 @@ func makeErr(errString string) *protos.RegisterResponse {
...
+	status, err := wrappers.GetGatewayStatus(ctx, networkID, ent.PhysicalID)
+	if err != nil && err != merrors.ErrNotFound {
+		return err
+	}
+	gw := (&models.MagmadGateway{}).FromBackendModels(ent, device, status)
+
+	_, err = configurator.UpdateEntities(ctx, networkID, gw.ToEntityUpdateCriteria(ent), serdes.Entity)
+	if err != nil {
+		return err
+	}
```
hcgatewood: We should be able to do this with the "mutable" version of the gateway object -- i.e. without having to pull/integrate gateway state

hcgatewood (changes requested):
Looks good! 2 small things

orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
christinewang5: I was referencing how the update gateway handlers does it here https://github.com/christinewang5/magma/blob/cc9780ca2f8bef44d42a034ac744af07b7a445aa/orc8r/cloud/go/services/orchestrator/obsidian/handlers/gateway_handlers.go#L340-L362

But I was searching for the mutable version and it seems like there are only models for specific gateway types like cwf, lte, fed, etc. Lmk if there is a way to get the mutable version just for a generic gateway?

orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
christinewang5: I think the func `LoadEntityForPhysicalID` gets the entity by the physical ID, but since the gateway is not yet registered to a physical ID, idt it would work... unless there is some load entity by physical ID func u are referencing?

hcgatewood: Oh true you're right, I misread "gateway ID" as meaning "hardware ID" -- a +1 for why functions with large # of arguments is an anti-pattern 🙂

orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
hcgatewood: Ah true weird that there isn't a mutable version of the magmad gateway. So let's still do it this way, except -- we can pass in the "status" as nil, since we don't care about that when we update the ent

hcgatewood (approved):
Lgtm, just fix the 1 thing where we don't need to read in the gw status

PR #12029, opened 2022-03-09
chore: Push test cache for Magma VM to S3

themarwhal:

## Summary
Magma VM cache: Improve Magma VM cache hosted on S3 by including test cache as well.
Devconainer cache:  Python tests do not pass when run with --config=asan/lsan. Modifying the cache population to only include C/C++ targets

https://github.com/magma/magma/runs/5473144787?check_suite_focus=true

Screen Shot 2022-03-09 at 11 56 35 AM
## Test Plan
Tested commands locally with magma VM

## Additional Information

### DIFF
--- a/.github/workflows/bazel-cache-push.yml
+++ b/.github/workflows/bazel-cache-push.yml
@@ -65,6 +65,14 @@ jobs:
         run: |
           cd lte/gateway
           vagrant ssh -c 'cd ~/magma; bazel build //...' magma
+      - name: Test all with Bazel `bazel test //... --cache_test_results=no`
+        run: |
+          cd lte/gateway
+          vagrant ssh -c 'cd ~/magma; bazel test //... --cache_test_results=no' magma
+      - name: Test C/C++ with ASAN `bazel test //orc8r/gateway/c/... //lte/gateway/c/... --config=asan`
+        run: |
+          cd lte/gateway
+          vagrant ssh -c 'cd ~/magma; bazel test //orc8r/gateway/c/... //lte/gateway/c/... --config=asan --cache_test_results=no' magma
       - name: Upload .bazel-cache and .bazel-cache-repo to S3
         run: |
           tar -zcvf ${{ env.BAZEL_CACHE_MAGMA_VM_TAR }} ${{ env.BAZEL_CACHE }}/
--- a/.github/workflows/bazel-cache-push.yml
+++ b/.github/workflows/bazel-cache-push.yml
@@ -57,22 +57,22 @@ jobs:
         run: |
           cd lte/gateway
           vagrant up magma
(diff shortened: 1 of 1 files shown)


.github/workflows/bazel-cache-push.yml
```diff
@@ -65,6 +65,14 @@ jobs:
         run: |
           cd lte/gateway
           vagrant ssh -c 'cd ~/magma; bazel build //...' magma
+      - name: Test all with Bazel `bazel test //... --cache_test_results=no`
```
nstng: why the `--cache_test_results=no`? from my understanding this is not relevant for building caches - and only adds to the runtime.

.github/workflows/bazel-cache-push.yml
```diff
@@ -65,6 +65,14 @@ jobs:
...
+        run: |
+          cd lte/gateway
+          vagrant ssh -c 'cd ~/magma; bazel test //... --cache_test_results=no' magma
```
nstng: more a conceptual question: do we want to let the job fail if a bazel run fails here? the job is designed to build caches - the build and the tests are verified in another workflow.

.github/workflows/bazel-cache-push.yml
```diff
@@ -92,9 +100,8 @@ jobs:
           run: |
             cd /workspaces/magma
             bazel build //...
-            bazel test //...
-            bazel test //... --config=asan
-            bazel test //... --config=lsan
+            bazel test //... --cache_test_results=no
```
nstng: just curious: is there a reason why the run here executes all statements vs for the vm where all statements are split in different steps?

themarwhal: good q! just inconsistent. Will fix!

.github/workflows/bazel-cache-push.yml
themarwhal: Oh i think you might be right!

> Note that test results are always saved in Bazel's output tree, regardless of whether this option is enabled, so you needn't have used --cache_test_results on the prior run(s) of bazel test in order to get cache hits. The option only affects whether Bazel will use previously saved results, not whether it will save results of the current run.

I will double check locally to see if it affects the cache size, then remove if it doesn't

.github/workflows/bazel-cache-push.yml
themarwhal: I think for now, failing is OK. I wouldn't have caught the issue on master if I didn't see everything red :D

nstng: Agree :)

electronjoe (approved):
LGTM % @nstng feedback

nstng (approved):

PR #11969, opened 2022-03-04
fix(agw): Creating unique port number for all dedicated bearers

rsarwad:
fix(agw): Creating unique port number for all dedicated bearers

## Summary
On master code base, test case, test test_attach_detach_maxbearers_twopdns.py was failing because, as part of this TC, first PDN is created as part of attach and the dedicated bearer is created with 4 packet filters with port number 5001, 5002, 5003 and 5004.
In next step, secondary PDN is created along with 8 dedicated bearers starting same port numbers.
Since port numbers are same; most of the packet filters have loaded to secondary PDN's default bearer.
Now added logic to assign unique port number for all packet filters of dedicated bearer
## Test Plan
Executed test_attach_detach_maxbearers_twopdns.py multiple times

## Additional Information
Checking for other test case for which same port numbers are assigned for different dedicated bearer

PR #12038, opened 2022-03-09
fix(mme): Fix network address in create bearer packet filter

ssanadhya:

## Summary

The test case [test_attach_detach_maxbearers_twopdns.py](https://github.com/magma/magma/blob/master/lte/gateway/python/integ_tests/s1aptests/test_attach_detach_maxbearers_twopdns.py) started failing after the changes in https://github.com/magma/magma/commit/b4fc6908954e49e3b9e47fa3cfe0e4ff2c9fdea2 . 
The new `parseIpv4Network` logic behaves differently from `folly::IPAddress::tryCreateNetwork` when the input address is of the form "192.168.129.42/16". Specifically, the output is:

- folly::IPAddress::tryCreateNetwork("192.168.129.42/16") -> (192.168.0.0, 16)
- parseIpv4Network("192.168.129.42/16") -> (192.168.129.42, 16)

Since the default downlink flow rules in `s1ap_utils.py` include rules for "192.168.129.42/32" and "192.168.129.42/16", the above logic , in addition with the overlapping port number range (as highlighted in https://github.com/magma/magma/pull/11969) , was causing an overlap on flow rules. This was causing the number of created rules to be 28, while the expected count was 33 ( = 1 default + 8 bearers * 4 rules_per_bearer). 

This can be seen by dumping all the DL flows from OVS for `test_attach_detach_maxbearers_twopdns.py` on:

https://github.com/magma/magma/commit/2d306616f554475b0af0d357a9a481c53b17e70b , which shows [33 downlink flows ](https://github.com/magma/magma/files/8218894/dlflows_2d30.txt)

https://github.com/magma/magma/commit/b4fc6908954e49e3b9e47fa3cfe0e4ff2c9fdea2, which shows [28 downlink flows](https://github.com/magma/magma/files/8218897/dlflows_b4fc.txt)

The changes in https://github.com/magma/magma/pull/11969 circumvent the issue by expanding the port number range so that the flow rules do not overlap, but the IP address match is still incorrect.

This change:
1. fixes the logic in `fillIpv4` to treat the IP address returned from `parseipv4network` as a string and apply the mask when copying that address into a packet filter.
2. Adds a debug log in `s1ap_utils.py` to print the number of DL flows created vs expected
3. Include the PDN count also in generating the unique port number for each bearer.

## Test Plan

- Functional test with `test_attach_detach_maxbearers_twopdns.py`
Before the change, MME Debug log reports:
```
DEBUG UTIL tasks/grpc_service/SpgwServiceIm:0358  Network Address: 192.168.129.42 Network Mask: 255.255.0.0
```
After the change
```
DEBUG UTIL tasks/grpc_service/SpgwServiceIm:0359  Network Address: 192.168.0.0 Network Mask: 255.255.0.0
```

- Sanity test
`make integ_test`

### DIFF
--- a/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
+++ b/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
@@ -467,6 +467,8 @@ class S1ApUtil(object):
                     },
                 },
             )
+            print ("Created ", len(total_dl_ovs_flows_created),
+                    "Expected ", total_num_dl_flows_to_be_verified)
             assert (
                 len(total_dl_ovs_flows_created)
                 == total_num_dl_flows_to_be_verified
--- a/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
+++ b/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
@@ -468,7 +468,7 @@ class S1ApUtil(object):
                 },
             )
             print ("Created ", len(total_dl_ovs_flows_created),
-                    "Expected ", total_num_dl_flows_to_be_verified)
+                   "Expected ", total_num_dl_flows_to_be_verified)
             assert (
                 len(total_dl_ovs_flows_created)
                 == total_num_dl_flows_to_be_verified
--- a/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
+++ b/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
@@ -467,8 +467,10 @@ class S1ApUtil(object):
                     },
                 },
             )
-            print ("Created ", len(total_dl_ovs_flows_created),
-                   "Expected ", total_num_dl_flows_to_be_verified)
+            print(
+                "OVS DL flows created ", len(total_dl_ovs_flows_created),
+                "expected ", total_num_dl_flows_to_be_verified,
+            )
             assert (
                 len(total_dl_ovs_flows_created)
                 == total_num_dl_flows_to_be_verified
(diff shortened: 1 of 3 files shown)


ulaskozat (approved):
LGTM

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -467,6 +467,8 @@ def _verify_dl_flow(self, dl_flow_rules=None):
                     },
                 },
             )
+            print ("Created ", len(total_dl_ovs_flows_created),
```
ardzoht: should we describe on the log that these are for number of OVS flows?

ardzoht (approved):

themarwhal:
We do have some unit test coverage for the parse function. Is it worth adding the previously failing test case here? https://sourcegraph.com/github.com/magma/magma/-/blob/lte/gateway/c/core/oai/test/spgw_task/test_spgw_service_impl.cpp?L27 or maybe it is the fill function that is in question here? I do not know, so I will defer to @LKreutzer who added the unit tests for this :)

pruthvihebbani (approved):

LKreutzer (approved):
LGTM. @themarwhal we will add some test cases (on Monday as I am travelling)!

themarwhal:
> LGTM. @themarwhal we will add some test cases (on Monday as I am travelling)!

@LKreutzer thanks! sounds great :) Safe travels!

ssanadhya:
> LGTM. @themarwhal we will add some test cases (on Monday as I am travelling)!

Agree with @themarwhal , it will be good to add test coverage for the fill function here. I did not modify the parsing function since it is working correctly as a parser in this case. The logic has to be vetted in the fillIpv4 function. Thanks for taking this up @LKreutzer

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
ssanadhya: Addressed in latest version

PR #12052, opened 2022-03-10
chore(mobilityd): Exclude run-out-of-ips from error monitoring

### LINKED ISSUE #12049: Sentry(mobilityd): Run out of available IP addresses
Sentry Issue: [LAB-AGWS-PYTHON-96](https://sentry.io/organizations/lf-9c/issues/3086078194/?referrer=github_integration)

```
Run out of available IP addresses
```

sebathomas:
## Summary

This is not a bug in Magma that needs to be monitored.
Fixes #12049.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/mobilityd/ip_allocator_pool.py
+++ b/lte/gateway/python/magma/mobilityd/ip_allocator_pool.py
@@ -26,7 +26,10 @@ import logging
 from copy import deepcopy
 from typing import List
 
+from magma.common.sentry import EXCLUDE_FROM_ERROR_MONITORING
 from magma.mobilityd.ip_descriptor import IPDesc, IPState, IPType
+from magma.mobilityd.mobility_store import MobilityStore
+from magma.mobilityd.utils import IPAddress, IPNetwork
 
 from .ip_allocator_base import (
     IPAllocator,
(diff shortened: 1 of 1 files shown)


Neudrino (approved):

lte/gateway/python/magma/mobilityd/ip_allocator_pool.py
```diff
@@ -230,7 +231,10 @@ def alloc_ip_address(self, sid: str, vlan_id: int) -> IPDesc:
             ip_desc.state = IPState.ALLOCATED
             return ip_desc
         else:
-            logging.error("Run out of available IP addresses")
+            logging.error(
+                "Run out of available IP addresses",
+                extra=EXCLUDE_FROM_ERROR_MONITORING,
```
pshelar, line 236: This is legitimate error that should be reported by sentry so that operator can fix the configuration. why is it excluded?

ardzoht: +1 I think this should not be excluded

sebathomas: Normally we try to exclude application errors that don't indicate something wrong in the code and are more related to server maintenance. In this case, it's not a very frequent error so we can also leave it if you prefer.

PR #12056, opened 2022-03-10
feat(dp): WIP - do not merge - elasticsearch integration

WojSad:

feat(dp): Elasticsearch integration (inserting data to new index in ES)

## Summary

Changed logic for storing Domain Proxy related logs (requests and responses to and from SAS, to and from CBSD).
From now on they are no longer stored in the DB but in Elasticsearch through Fluentd.
Amended respective configuration files.
Added test Elasticsearch and Fluentd deployments (they are now built alongside Domain Proxy) 

## Test Plan

Added necessary mocks to Unit tests
Added Integration test for storing logs in ES


dp/cloud/python/magma/configuration_controller/config.py
```diff
@@ -53,6 +54,18 @@ class Config(object):
         'SAS_CERT_PATH', '/backend/configuration_controller/certs/ca.crt',
     )
 
+    # Elasticsearch
+    ELASTICSEARCH_INDEX = os.environ.get('ELASTICSEARCH_INDEX', 'dp')
```
jkmar: Can we have a separate module for elastic search?
It doesn't have to be a separate app (with Dockerfile), it can just be simple python module with:
- config (so that CC and RC don't have to have duplicated parts)
- one method publish request (or something like that, used by both RC and CC)

dp/cloud/python/magma/db_service/migrations/versions/f09f644ccc67_.py
```diff
@@ -0,0 +1,55 @@
...
+Create Date: 2022-03-10 14:38:53.050143
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = 'f09f644ccc67'
+down_revision = 'f84efd78a4de'
```
jkmar, line 14: Remember to update it when merging, since this has to point to the latest migration and it might change. Unfortunately we currently don't have any tests for it, so if it is broken, CI won't notice it.

dp/tools/deployment/vendor/config/fluent-conf.yml
```diff
@@ -0,0 +1,47 @@
+apiVersion: v1
```
jkmar, line 2: I would suggest fixing all yamlint errors

dp/cloud/python/magma/test_runner/tests/test_active_mode_controller.py
```diff
@@ -68,6 +80,39 @@ def given_cbsd_provisioned(self):
...
+        self.assertEqual(1, actual_log_types["registrationRequest"])
+        self.assertEqual(1, actual_log_types["registrationResponse"])
+        self.assertEqual(1, actual_log_types["spectrumInquiryRequest"])
+        self.assertEqual(1, actual_log_types["spectrumInquiryResponse"])
+        self.assertEqual(1, actual_log_types["spectrumInquiryResponse"])
+        self.assertEqual(1, actual_log_types["grantRequest"])
+        self.assertEqual(1, actual_log_types["grantResponse"])
+        self.assertEqual(1, actual_log_types["heartbeatRequest"])
+        self.assertEqual(1, actual_log_types["heartbeatResponse"])
+        # The number of GetCBSDStateRequest may differ between tests, so only asserting they have been logged
```
jkmar, line 112: How about some filtering to have exact results, (or maybe it is impossible)?

dp/skaffold.yaml
```diff
@@ -72,6 +74,8 @@ deploy:
     defaultNamespace: default
     manifests:
       - "./tools/deployment/vendor/fake_sas.yml"
+      - "./tools/deployment/vendor/elasticsearch.yml"
```
jkmar, line 77: Just a question - will orc8r and dp use the same elasticsearch?

dp/cloud/python/magma/radio_controller/services/dp/service.py
```diff
@@ -98,15 +105,19 @@ def CBSDRegister(self, request: CBSDRequest, context) -> CBSDStateResult:
         """
 
         logger.info(f"Registering CBSD {request.serial_number=}")
+        request_type = result_type = 'CBSDRegister'  # noqa: WPS429
```
jkmar: Do we actually need 2 variables for that (there are several places with the same code actually)? One should be enough, especially since they are not changed anywhere, unless I'm missing something

WojSad: i only did it for readibility. Will make it into one var called message_type

dp/skaffold.yaml
WojSad: yes, however this one is only going to be used by DP for testing (like fake sas)

dp/cloud/python/magma/fluentd_client/client.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class FluentdClient(object):
+    """
+    Client class for Fluentd communication.
+    """
+    def __init__(self, config: Optional[Config] = None):
+        self.config = config or get_config()
+
+    def log_db_request(self, db_request: DBRequest) -> None:
```
jkmar: The current approach is fine (separate fluent package), maybe one improvement would be to just have one method here: `send_dp_log`, which takes custom `dp_log` type defined in this package, that way we would have just one method and not depend on database or grpc here.
For example:
```
@dataclass
class DPLog:
    from: str
    to: str
    name: str
    message: str
    serial_number: str
    network_id: str
    fcc_id: str

def send_dp_log(self, log: DPLog):
    resp = self._post_to_fluentd(asdict(log))
```

dp/tools/deployment/vendor/config/fluent-conf.yml
```diff
@@ -0,0 +1,52 @@
...
+kind: ConfigMap
+metadata:
+  name: fluentd-config
+data:
+  fluent.conf: |
+    <system>
+      log_level debug
+    </system>
+
```
jkmar: One last yaml error here

jkmar (approved):

dp/cloud/python/magma/fluentd_client/config.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Configuration class for Configuration Controller
+    """
+    # General
+    LOG_LEVEL = os.environ.get('LOG_LEVEL', 'DEBUG')
+
+    # Elasticsearch
+    ELASTICSEARCH_INDEX = os.environ.get('ELASTICSEARCH_INDEX', 'dp')
+
+    # Fluentd
```
xbend, line 28: It looks like the orchestrator configuration variables that define where orc8r fluentd is are: `FLUENTD_SERVICE_HOST` and (`FLUENTD_SERVICE_PORT` or `FLUENTD_SERVICE_PORT_FORWARD`)

xbend (changes requested):

xbend:
Closing. Implemented as part of https://github.com/magma/magma/pull/12311

Issue #11968, opened 2022-03-04
mme: memory leak with init_call_data


themarwhal:
As soon as I ran`make integ_test` locally, I'm seeing
```bash
ar 04 15:49:58 magma-dev-focal mme[37541]: TASK_SCTP terminated
Mar 04 15:50:00 magma-dev-focal mme[37541]: [ITTI][D] Some threads are still running, force exit
Mar 04 15:50:00 magma-dev-focal mme[37541]: Cleaning up SPGW configsE: 22-03-04 15:50:00 [37541]dangling 'PUSH' socket created at /home/vagrant/magma/lte/gateway/c/core/oai/lib/itti/intertask_interface.c:206
Mar 04 15:50:00 magma-dev-focal mme[37541]: E: 22-03-04 15:50:00 [37541]dangling 'PUSH' socket created at /home/vagrant/magma/lte/gateway/c/core/oai/lib/itti/intertask_interface.c:206
Mar 04 15:50:00 magma-dev-focal mme[37541]: E: 22-03-04 15:50:00 [37541]dangling 'PULL' socket created at /home/vagrant/magma/lte/gateway/c/core/oai/lib/itti/intertask_interface.c:214
Mar 04 15:50:00 magma-dev-focal mme[37541]: =================================================================
Mar 04 15:50:00 magma-dev-focal mme[37541]: ==37541==ERROR: LeakSanitizer: detected memory leaks
Mar 04 15:50:00 magma-dev-focal mme[37541]: Direct leak of 728 byte(s) in 1 object(s) allocated from:
Mar 04 15:50:00 magma-dev-focal mme[37541]:     #0 0x7f0dc153d3d1 in operator new(unsigned long) (/lib/x86_64-linux-gnu/liblsan.so.0+0x103d1)
Mar 04 15:50:00 magma-dev-focal mme[37541]:     #1 0x55f9e4298c7e in magma::S6aProxyResponderAsyncService::init_call_data() /home/vagrant/magma/lte/gateway/c/core/oai/tasks/async_grpc_service/grpc_async_service_task.cpp:91
Mar 04 15:50:00 magma-dev-focal mme[37541]:     #2 0x55f9e4297934 in magma::AsyncService::wait_for_requests() /home/vagrant/magma/lte/gateway/c/core/oai/tasks/async_grpc_service/grpc_async_service_task.cpp:55
[...]

PR #12062, opened 2022-03-10
fix(agw): Modified code to stop completion queue and clear the messages from it when mme service is stopped

rsarwad:
fix(agw): Modified code to stop completion queue and clear the messages from it when mme service is stopped

## Summary
Provided fixes for the issue, https://github.com/magma/magma/issues/11968
Modified the code to stop the completion queue after server is shutdown. Previously the completion queue was stopped while terminating async grpc itti thread. 

## Test Plan
Executed s1ap sanity test suite and S6a-CLR and S6a-Reset messages are verified on Tera VM


ardzoht (approved):
lgtm thanks for the fix!

PR #12081, opened 2022-03-11
ci(orc8r): Added Terraform Lint CI Check

### LINKED ISSUE #11996: Create CI job for checking terraform format checks
Setup CI on GitHub Actions for checking terraform format on all `.tf` files with the following command. Also check if we have something in marketplace that we can use. 
```bash
terraform fmt
```

ShubhamTatvamasi:

## Additional Information

### DIFF
--- a/.github/workflows/terraform-lint.yml
+++ b/.github/workflows/terraform-lint.yml
@@ -1,0 +1,16 @@
+name: Terraform Lint
+
+on: [push, pull_request]
+
+jobs:
+  terraform-lint:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Terraform Lint Check
+      run: |
+        if [[ -n "$(terraform fmt -check -recursive -diff)" ]]; then
+          echo "Some terraform files need be formatted, run 'terraform fmt' to fix"
+          exit 1
+        fi
--- a/.github/workflows/terraform-lint.yml
+++ b/.github/workflows/terraform-lint.yml
@@ -15,11 +15,11 @@ jobs:
   terraform-lint:
     runs-on: ubuntu-latest
     steps:
-    - name: Checkout
-      uses: actions/checkout@v3
-    - name: Terraform Lint Check
-      run: |
-        if [[ -n "$(terraform fmt -check -recursive -diff)" ]]; then
-          echo "Some terraform files need be formatted, run 'terraform fmt' to fix"
-          exit 1
-        fi
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Terraform Lint Check
+        run: |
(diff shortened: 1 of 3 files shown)


themarwhal (approved):
lgtm!

.github/workflows/terraform-lint.yml
```diff
@@ -0,0 +1,25 @@
...
+on:
+  push:
+    branches:
+      - master
+      - 'v1.*'
+  pull_request:
+    branches:
+      - master
+      - 'v1.*'
+
```
themarwhal, line 13: should we add some concurrency configs? @quentinDERORY

ShubhamTatvamasi:
For common tasks such as lint check, should we start using actions from marketplace?
https://github.com/marketplace/actions/terraform-lint-action

ShubhamTatvamasi:
I have updated code, now using action from marketplace.

themarwhal (approved):

Neudrino (commented):
> For common tasks such as lint check, should we start using actions from marketplace? https://github.com/marketplace/actions/terraform-lint-action

I agree that using Actions from Marketplace might be a good idea.
However, on top of that we might want to stick to the standard [ReviewDog](https://github.com/reviewdog/reviewdog/blob/master/README.md) provided linters where possible, as they provide excellent integration with Github out of the box, and avoid custom code here.

In particular for this case this would mean to us [tflint](https://github.com/marketplace/actions/run-tflint-with-reviewdog).
This would also have the advantage that it could also easily be integrated with [our existing set (and setup) of reviewdog](https://github.com/magma/magma/blob/master/.github/workflows/reviewdog-workflow.yml).

.github/workflows/terraform-lint.yml
```diff
@@ -0,0 +1,22 @@
+---
+name: Terraform Lint
+
+on:
```
Neudrino: yamllint...
```suggestion
on:  # yamllint disable-line rule:truthy
```

Neudrino (changes requested):

Neudrino (approved):
More linters are always good and this PR LGTM in general.

We might consider to us [tflint](https://github.com/marketplace/actions/run-tflint-with-reviewdog) and integrate it with our [our existing set (and setup) of reviewdog](https://github.com/magma/magma/blob/master/.github/workflows/reviewdog-workflow.yml) linters to benefit from the excellent Github integration of reviewdog :dog:.

.github/workflows/terraform-lint.yml
Neudrino:
```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number}}
  cancel-in-progress: true
```
Maybe something like this in alignment with the other linters.

PR #12086, opened 2022-03-11
chore: fix mypy static type checking issues

Neudrino:
## Summary

:information_source: Commit wise review suggested.

Further cleanup for mypy introduction.
Follow up on:
- https://github.com/magma/magma/pull/11963
- https://github.com/magma/magma/pull/12012

## Test Plan

``` bash
cd $MAGMA_ROOT
mypy lte/gateway/python/
```

Before:
```
Found 14 errors in 6 files (checked 179 source files)
```

After:
``` 
Success: no issues found in 179 source files
```


lte/gateway/python/setup.py
```diff
@@ -22,7 +22,7 @@
 
 setup(
     name='lte',
-    version=VERSION,
+    version=VERSION,  # type: ignore[arg-type]
```
nicarl: Mypy is actually right here. The type of `VERSION` is `Optional[str]`. I Would rather raise and error if `PKG_VERSION` is not set as env variable

lte/gateway/python/magma/kernsnoopd/handlers.py
```diff
@@ -92,7 +93,7 @@ def _get_cmdline(self, pid: int) -> list:
         return psutil.Process(pid=pid).cmdline()
 
     @lru_cache(maxsize=1024)
-    def _ip_addr_to_str(self, family: int, daddr: (int, int)) -> str:
+    def _ip_addr_to_str(self, family: int, daddr: Tuple[int, int]) -> str:  # type: ignore[return]
```
nicarl: The correct type annotation should be:
```python
def _ip_addr_to_str(self, family: Literal[AddressFamily.AF_INET6, AddressFamily.AF_INET], daddr: Tuple[int, int]) -> str:

File: lte/gateway/python/magma/kernsnoopd/snooper.py [hunk 3]
```diff
@@ -65,7 +66,7 @@ def __init__(
 
         super().__init__(interval=collect_interval, loop=service_loop)
         self._bpf = None
-        self._handlers = []
+        self._handlers: List[Any] = []
```
nicarl: I would avoid `Any` if possible.
`List[ByteCounter]` should be the correct type

lte/gateway/python/load_tests/common.py
```diff
@@ -84,7 +75,7 @@ def benchmark_grpc_request(
     output_file: str,
     num_reqs: int,
     address: str,
-    import_path: str = None,
+    import_path: str,
```
Neudrino, line 78: It turns out, that the import path is not actually optional, but always given by the caller.
I.e. the logic to "intelligently" determine some protobuf path can be removed.

pshelar (approved):
LGTM

lte/gateway/python/magma/kernsnoopd/handlers.py
Neudrino: As far as I can see, the mypy error does not complain on the types here, but that the method does not return for all given pathways.
I.e. I would see the following options
- not checking the AF_INET6 type explicitly
```python
        if family == AF_INET:
            return inet_ntop(AF_INET, pack('I', daddr[0]))
        else:
            # noinspection PyTypeChecker
            return inet_ntop(AF_INET6, self.Addr(*daddr))
```
- implement another return
```python
        if family == AF_INET:
            return inet_ntop(AF_INET, pack('I', daddr[0]))
        elif family == AF_INET6:
            # noinspection PyTypeChecker
            return inet_ntop(AF_INET6, self.Addr(*daddr))
        else:
            raise Exception("No valid socket family given!")
```
- another return with a `pass` instead of a `raise`.

lte/gateway/python/setup.py
Neudrino: As `0.0.0` was mentioned as default in case of `None` in the comment, I adopted to set the value directly.

lte/gateway/python/magma/kernsnoopd/snooper.py
Neudrino: Thanks a lot! Changed.

themarwhal (approved):

lte/gateway/python/magma/kernsnoopd/handlers.py
nicarl: Yes it complains about not returning after the `elif` case. If you use `family: Literal[AddressFamily.AF_INET6, AddressFamily.AF_INET]` and enforce correct typing you can use:
```python
if family == AF_INET:
    return inet_ntop(AF_INET, pack('I', daddr[0]))
# family has to be AF_INET6 at this point if everyone respects the types :-)
return inet_ntop(AF_INET6, self.Addr(*daddr))
```
Your current solution is also fine :+1:

sebathomas (approved):

PR #12095, opened 2022-03-12
fix(agw): Add typing-extensions as Python dependency

sebathomas:
## Summary

The import was introduced as part of #12012 to improve the typing of Python code. Apparently, it causes issues on bare-metal AGWs.

## Test Plan

Executed `make run` on my VM and the Python services started up. Unfortunately, I have no bare-metal AGW available to test it.

## Additional Information


themarwhal (approved):

ardzoht (approved):

ardzoht:
Thanks for the fix @sebathomas !

PR #12100, opened 2022-03-14
chore(mme): add unit test for fillIpv4

LKreutzer:

## Summary
- Unit test for `SpgwServiceImpl::fillIpv4` is added (see  #12038)
  - `SpgwServiceImpl::fillIpv4` is a private function. The only public function that can lead to calls in `fillIpv4` is `SpgwServiceImpl::CreateBearer` and several conditions need to be fulfilled in order to lead to `fillIpv4` calls. 
  - It seems that the `SpgwServiceImpl::fillIpv4` function is not called in the existing unit tests - all OAI unit tests succeed even if e.g. `raise(SIGSEGV);` is added to the `SpgwServiceImpl::fillIpv4` function. The same if true for the `SpgwServiceImpl::CreateBearer` function which is the only C/C++ function that can lead to calls of the `fillIpv4` function (via `fillUpPacketFilterContents`).
  - `SpgwServiceImpl::fillIpv4` is made public in order to enable direct testing and to limit extensive overhead from indirect testing via `SpgwServiceImpl::CreateBearer`.

## Test Plan
- Run tests `make test_oai`
- CI

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/grpc_service/SpgwServiceImpl.h
+++ b/lte/gateway/c/core/oai/tasks/grpc_service/SpgwServiceImpl.h
@@ -87,24 +87,27 @@ class SpgwServiceImpl final : public SpgwService::Service {
 
   ipv4_network_t parseIpv4Network(const std::string& ipv4network_str);
 
- private:
-  /*
-   * Fill up the packet filter contents such as flags and flow tuple fields
-   * @param pf_content: packet filter content to be filled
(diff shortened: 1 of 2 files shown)


lionelgo (approved):
A public wrapper dedicated to test bool fillIpv4(...) would not change the existing interfaces.
... but would add an interface on the other side.

themarwhal (approved):
thanks :)

PR #12112, opened 2022-03-14
chore: Use per_file_copt for MME unit test flag

themarwhal:

## Summary
Moving the MME unit test compile flag to be handled with --per_file_copt (Thanks TNG team (edit: @ajahl :) ) for finding this!)

I've run `bazel build //...` and then `bazel test //...`  successively with NO CACHE and Bazel seems to build less with this PR. Visibly, I was able to see that this PR leads to only the OAI directory being rebuilt with the test flag. Curiously, the cache size does not change between master and this PR. However, the build time seems to differ by some amount.

| branch | bazel build //... fresh | bazel test //... after build all | du -sh .bazel-cache after build and test |
|-|-|-|-|
| master | 415.338s | 344.743s | 5.8G |
| this PR | 399.585s | 158.011s | 5.8G |

## Test Plan
CI

## Additional Information

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -31,14 +31,14 @@ common:devcontainer --repository_cache=/workspaces/magma/.bazel-cache-repo
 build:devcontainer --define=folly_so=1
 
 # TEST CONFIGS
-#  Bazel test runtime default: PATH=/bin:/usr/bin:/usr/local/bin
-#  Some python tests require access to /usr/sbin binaries (e.g. route, ifconfig)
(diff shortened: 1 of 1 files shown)


electronjoe (approved):
Good to have an example of this in the `.bazelrc`, regardless of gains!

themarwhal:
Ah I see the edit by Nils Thanks @ajahl for finding the flag :) And thank you @nstng for pointing me to Alex!

PR #12119, opened 2022-03-15
[WIP] Update to Go 1.17

alexzurbonsen:
## Summary

WIP

This is a cleaned up branch (at least it is a little cleaner).

## Additional Information


jkmar:
If go version is updated in every module then why not use the latest one (1.18)?

alexzurbonsen:
> If go version is updated in every module then why not use the latest one (1.18)?

Already looking into it. 1.18 wasn't released, when we started to work on the update.

PR #12122, opened 2022-03-15
chore: bazel dependencies are cleaned up 

nstng:

### Problem
* we explicitly use `https://github.com/protocolbuffers/protobuf` v3.15.0
* via `https://github.com/rules-proto-grpc/rules_proto_grpc` we also use v3.15.3
* apart from using different versions, this leads to >100mb extra cache

### Solution Here
* remove explicit dependency of protobuf v3.15.0
* use implicit protobuf v3.15.3 from grpc rules when needed

## Test Plan

CI

## Additional Information

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -10,7 +10,7 @@
 # limitations under the License.
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
-load("//bazel:third_party_repositories.bzl", "grpc", "protobuf")
+load("//bazel:third_party_repositories.bzl", "grpc")
 
 http_archive(
     name = "rules_python",
@@ -93,12 +93,6 @@ load("@rules_cc//cc:repositories.bzl", "rules_cc_dependencies")
 rules_cc_dependencies()
 
 ### PROTO / GRPC DEPENDENCIES ###
-protobuf()
-
-load("@protobuf//:protobuf_deps.bzl", "protobuf_deps")
-
-protobuf_deps()
-
 grpc()
 
 load("@rules_proto_grpc//:repositories.bzl", "rules_proto_grpc_repos", "rules_proto_grpc_toolchains")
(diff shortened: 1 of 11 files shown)


electronjoe (approved):
Nice job untangling something rather nuanced / challenging.  This is absolutely better - having just one protobuf variant flying around is a good call.  My initial concern was simply traceability now that explicit protobuf dep has been removed. E.g. that `rules_proto_grpc/../3.1.1.` yields proto version `v3.15.3`. I see you have mentioned this in the comment. Don't know if there is any other comment location that may persist where we would want to document this versioning? Down the road I could see people being like: "what version of protoc are we using" and having trouble figuring it out. edit: is there a simple way (bazel query or other) to determine this protobuf version - that you could leave a command for in this PR description?

themarwhal (approved):
ohh! I learn something new everyday! Thanks for the cleanup!

themarwhal:
> Nice job untangling something rather nuanced / challenging. This is absolutely better - having just one protobuf variant flying around is a good call. My initial concern was simply traceability now that explicit protobuf dep has been removed. E.g. that `rules_proto_grpc/../3.1.1.` yields proto version `v3.15.3`. I see you have mentioned this in the comment. Don't know if there is any other comment location that may persist where we would want to document this versioning? Down the road I could see people being like: "what version of protoc are we using" and having trouble figuring it out. edit: is there a simple way (bazel query or other) to determine this protobuf version - that you could leave a command for in this PR description?

I vaguely remember reading somewhere that Bazel 5.0.0's module? approach is an attempt at dealing with these types of transitive dependencies... But yes, I think at least some sort of query might be helpful to figure out.

nstng:
> > edit: is there a simple way (bazel query or other) to determine this protobuf version - that you could leave a command for in this PR description?

> ... But yes, I think at least some sort of query might be helpful to figure out.

I did not figure out a canonical way :/ ... best I can currently offer:

* What external dependencies are there?
  * `bazel query //external:* | sort`
* Where is an external dependency?
  * ``"`bazel info output_base`/external/com_google_protobuf"``
* What version has the dependency?
  * depends on the dependency where to find it, e.g., ``cat "`bazel info output_base`/external/com_google_protobuf/js/package.json" | grep version``

Of course this is only relevant for transitive external dependencies. For direct external dependencies this is specified in `bazel/<prefix>_repositories.bzl`. But centralizing meta data into a single file is in the backlog (e.g., as in https://github.com/rules-proto-grpc/rules_proto_grpc/blob/2e8be8e27cc82203794f14dbdcf37189b02ab722/repositories.bzl#L11)

PR #12135, opened 2022-03-15
chore: migrate gcc-warnings job to use per_file_copts

themarwhal:

## Summary
This PR adds a max_gcc_warnings config to bump up GCC warnings flags for CI annotation. In the previous iteration, I had to manually modify the boringssl source file to remove -Werror so that build did not fail. With this approach, we do not have to touch external dependencies as we only apply the expanded set of GCC flags to Magma C/C++ files. This is also nice in terms of cache bloating.

This provides a quicker workflow of just adding `--config=max_gcc_warnings` to any target. 

## Test Plan
Ran `bazel build //lte/gateway/c/... --config=max_gcc_warnings` locally to see more warning output
Checking the compile log uploaded by the CI job.

## Additional Information

### DIFF
--- a/.github/workflows/gcc-problems.yml
+++ b/.github/workflows/gcc-problems.yml
@@ -56,9 +56,6 @@ jobs:
       - path_filter
     if: ${{ needs.path_filter.outputs.should_not_skip == 'true' }}
     name: Build all Bazelified C/C++ targets
-    env:
-      C_CPP_FLAGS: "-Wextra -Wshadow -Wimplicit-fallthrough -Wduplicated-cond -Wduplicated-branches -Wlogical-op -Wnull-dereference -Wformat=2 -Wstrict-overflow=4 -Wuninitialized -Wshift-overflow=2"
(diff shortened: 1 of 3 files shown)


.github/workflows/gcc-problems.yml
```diff
@@ -56,9 +56,6 @@ jobs:
       - path_filter
     if: ${{ needs.path_filter.outputs.should_not_skip == 'true' }}
     name: Build all Bazelified C/C++ targets
-    env:
-      C_CPP_FLAGS: "-Wextra -Wshadow -Wimplicit-fallthrough -Wduplicated-cond -Wduplicated-branches -Wlogical-op -Wnull-dereference -Wformat=2 -Wstrict-overflow=4 -Wuninitialized -Wshift-overflow=2"
-      C_ONLY_FLAGS: "-Wjump-misses-init"
```
themarwhal, line 61: there is no --per_file_c_only_copt option it seems? will double check tomorrow

electronjoe (approved):

PR #12137, opened 2022-03-16
chore: Migrate PACKAGE_* compile defs to be handled with per_file_copt

### LINKED ISSUE #11591: Find method for replacement of compile-time {PACKAGE_BUGREPORT, PACKAGE_NAME, PACKAGE_VERSION}

themarwhal:

## Summary
Address https://github.com/magma/magma/issues/11591

This PR moves out the copts that set PACKAGE_BUGREPORT, PACKAGE_VERSION, and PACKAGE_NAME from oai's BUILD.bazel to bazelrc. Ideally, we would handle these by creating libraries for each task and setting the copts there. As that is not feasible for the initial MME bazelification due to cyclic dependencies, per_file_copt gives us an easy way to set copts depending on the file path. 

I'm limiting this PR to Bazel related changes only, but I am thinking these define variables can be cleaned up a little bit more. We can probably get rid of PACKAGE_BUGREPORT and PACKAGE_VERSION as they don't seem to contain useful values.  

## Test Plan
Run `bazel build //lte/gateway/c/core/... -s` and visibly confirm the compilation flags are correct. 

## Additional Information

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -37,8 +37,16 @@ build --test_env=PATH=/bin:/usr/bin:/usr/local/bin:/usr/sbin
 # Use MAGMA_ROOT from the host system in tests.
 # Needed by python tests (e.g. freedomfi_one_tests in enodebd)
 build --test_env=MAGMA_ROOT
+
+# MME specific compile time defines
 # Compile mme libraries with unit test flag
 test --per_file_copt=^lte/gateway/c/core/.*$@-DMME_UNIT_TEST
+# TODO: deprecate these flags used for logging if possible
+build --per_file_copt=^lte/gateway/c/core/.*$@-DPACKAGE_BUGREPORT=\"TBD\"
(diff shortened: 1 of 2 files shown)


nstng (approved):
Nice idea to use per_file for this. Tested locally, lgtm

PR #12143, opened 2022-03-16
chore: bazel file linter is applied in ci

nstng:

## Summary

Forgot again to apply the buildifier in a previous PR :/
Should we add a check to the Bazel CI workflow?
Added a check in the test job - I would see it as a test, but are open to move it to a separate job.

linterFail

## Test Plan

CI
* see first commit fith failure run https://github.com/magma/magma/runs/5570173878?check_suite_focus=true
* second commit with fixes TBD

## Additional Information

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -185,3 +185,12 @@ jobs:
           run: |
             cd /workspaces/magma
             bazel test //... --test_output=errors --cache_test_results=no
+      - name: Run `bazel run //:check_starlark_format`
+        uses: addnab/docker-run-action@v2
+        with:
+          image: ${{ env.DEVCONTAINER_IMAGE }}
+          # TODO: Remove work-around mount of Github workspace to /magma (https://github.com/addnab/docker-run-action/issues/11)
+          options: -v ${{ github.workspace }}:/workspaces/magma/
+          run: |
+            cd /workspaces/magma
+            bazel run //:check_starlark_format
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -25,7 +25,7 @@ buildifier_test(
     timeout = "short",
     srcs = ["//:starlark_files"],
     lint_mode = "warn",
-    lint_warnings = ["all"],
+    lint_warnings = ["-unused-variable"],
     verbose = True,
 )
 
--- a/bazel/cpp_repositories.bzl
+++ b/bazel/cpp_repositories.bzl
@@ -69,7 +69,7 @@ def cpp_repositories():
         commit = "d8326b2bba945a435f299e7526c403d7a1f68c1f",
         remote = "https://github.com/jupp0r/prometheus-cpp.git",
         shallow_since = "1485901529 +0100",
-        repo_mapping = {"@protobuf" : "@com_google_protobuf"},
+        repo_mapping = {"@protobuf": "@com_google_protobuf"},
     )
 
     # cpp_redis dependency
--- a/bazel/go_repositories.bzl
+++ b/bazel/go_repositories.bzl
@@ -1,6 +1,21 @@
+# Copyright 2021 The Magma Authors.
+
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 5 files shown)


.github/workflows/bazel.yml
```diff
@@ -183,3 +183,12 @@ jobs:
           run: |
             cd /workspaces/magma
             bazel test //... --test_output=errors --cache_test_results=no
+      - name: Run `bazel run //:check_starlark_format`
```
themarwhal, line 188: yeah I haven't had a chance to look into this, but `check_starlark_format` **should** be run as part of bazel test //... but it doesn't seem to catch all issues. Is this identical as running `bazel run //:buildifier` and then checking for modified files?

nstng: My understanding:
* `buildifier_test` will supersede `buildifier` with `mode` `diff` (deprecated)
* I would have also expect that a `buildifier_test` target should be run with `bazel test`, but I was not able to get the target to fail reliable with `test`. E.g., formmatting issues in `$MAGMA_ROOT/BUILD.bazel` are detected via `bazel test`. But not in other build files. `bazel run` seems to be reliable.

> Is this identical as running bazel run //:buildifier and then checking for modified files?

This is my understanding. I.e., same as `bazel run //:buildifier` with `mode` `diff`

themarwhal: Thanks for the summary :) so is the difference triggering it with run vs test?

themarwhal (approved):

.github/workflows/bazel.yml
nstng: yes, run works in my experiments - but the test also seems to fail for linter warnings. I will have an additional look. I could try to remove the linting from the check - do you still need/use the `check_starlark_format` in it's current state? If yes, I can create a new one without the linting.
FYI: the linter warnings are semi-helpful. The complaints seem to be for starlark syntax in general, i.e., bazel specific syntax is also in the warnings, e.g., unused name argument in custom rules.

themarwhal: Yes let's address the lint warnings and keep it on. We only seem to have a few!

.github/workflows/bazel.yml
```diff
@@ -183,3 +183,12 @@ jobs:
...
+        uses: addnab/docker-run-action@v2
+        with:
+          image: ${{ env.DEVCONTAINER_IMAGE }}
+          # TODO: Remove work-around mount of Github workspace to /magma (https://github.com/addnab/docker-run-action/issues/11)
+          options: -v ${{ github.workspace }}:/workspaces/magma/ -v ${{ github.workspace }}/lte/gateway/configs:/etc/magma
```
themarwhal: nit, i'm guessing we only need
```suggestion
          options: -v ${{ github.workspace }}:/workspaces/magma/
```

.github/workflows/bazel.yml
nstng: IMHO some we can not really address - because this is a starlark issue, but not a bazel issue. E.g., bazel wants to always have a name attribute in custom functions. But the linter complains that the argument is unused.
I would suggest that I either take a look if this linter warning can be excluded, or exclude the linting completely.

themarwhal: ahh i see. I'm ok with not having linting if we can't address or exclude some errors. thanks for looking into this :)

themarwhal: Maybe we can selectively enable the ones we want here (List of available tags https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md )

.github/workflows/bazel.yml
nstng: top, thank you

.github/workflows/bazel.yml
nstng: ok, with the second attempt I now understand how the exclude mechanics work :)
-> excluded the unused variable lint warning and fixed doc string warnings

I still see behavior in the buildifier that I do not understand, e.g., needed function doc string in `go_repositories.bzl` but not in `python_repositories.bzl`, no property ordering in `go_repositories.bzl`, ...
But I think this change is still a valid improvement.

electronjoe (approved):
Thanks @nstng! I myself keep forgetting buildifier also =D

Issue #11736, opened 2022-02-22
Delete unit test main wrappers which are no longer used


electronjoe:
Once the MME is transitioned to Bazel build, we can execute on this.

Bazel's `cc_test` rules do not require definition of a `main()` function - one is inferred by gtest.  We can therefore go back and delete all these test files which serve only this purpose.  E.g. `oai/test/sgw_s8_task/sgw_s8_test.cpp` and friends.

PR #12148, opened 2022-03-16
chore(mme): clean up MME unit tests to remove unused main function 

themarwhal:

## Summary
This PR covers a few MME unit test cleanups
1. When applicable, use TEST over TEST_F
2. If a main function is not necessary, remove it. If it is, add a comment indicating that. 

I am finding that some main functions are completely unused, but some are used to initialize MME's log thread. It's unclear to me which ones are required for Bazel (tracked https://github.com/magma/magma/issues/11736) , but I've removed all the ones that are not needed by CMake for now.

I think a good follow-up for this PR would be to identify exactly which test cases have a dependency on the log thread. It might be good to separate them out into separate files if they are not separate already.
## Test Plan
CI and manually running make test_oai in the Magma VM.

## Additional Information


ssanadhya (approved):
Lgtm

ulaskozat (approved):
LGTM. Good clean up!

PR #12142, opened 2022-03-16
chore(mme): Commits failing Bazel core oai tests with manual tag

electronjoe:
## Summary

Work towards #11656.

This PR commits four oai core tests as Bazel test targets, but
as each of them fail in some manner across `vanilla`, `asan` or `lsan`
configurations we use the `manual` tag so that they are not run in CI.

Manual tag use is not best practice generally, as it can lead to hidden
tests that are broken and don't get attention. But since we have a
focused team working on fixing these tests - and it seems valuable to
have the team working on master and not my branch - here we propose
committing them with a `manual` tag.

## Missing Coverage

I executed the script provided by @nstng in https://github.com/magma/magma/issues/11767#issuecomment-1055742007 with the following output:

```
@electronjoe ➜ /workspaces/magma (pr-giant-bazel-mme-rebase ✗) $ /tmp/script.sh 
Source files not found in BUILD.bazel:
mme_app_serialization.cpp
mme_app_serialization.h
test_mme_app_ue_context.c
mme_app_test.cpp
s6a_test.cpp
s6a_recv_async_grpc_messages.cpp
ngap_test.cpp
test_spgw_state_converter.cpp
test_spgw_service_impl.cpp
sgw_s8_test.cpp
s1ap_test.cpp
amf_app_test.cpp
test_amf_encode_decode.cpp
```

Note that some of these files are "wrapper" gtest files that CMake uses and we do not use in Bazel, covered in #11736.

Others are tests that need to be brought into Bazel:
- test_amf_encode_decode.cpp #12146
- test_spgw_service_impl.cpp #12105
- test_mme_app_ue_context.c #12147
- s6a_recv_async_grpc_messages.cpp #12174

Are these dead code @ssanadhya?

- mme_app_serialization.cpp
- mme_app_serialization.h

We will want to re-run the script after a number of these have been dealt with.

## Test Plan

Validated that we do not break any Bazel build or test with this PR.
Validated that tests can be run manually and fail per their GH Issues.

Signed-off-by: Scott Moeller <electronjoe@gmail.com>

PR #12150, opened 2022-03-16
fix(mme): Fix spgw_procedures test for Bazel

### LINKED ISSUE #11984: Explore failing //lte/gateway/c/core/oai/test/spgw_task:spgw_procedures_test
# Problem
* https://github.com/magma/magma/blob/331551e2289d30fdbd6c811e8d2a46d95fd5c6b2/lte/gateway/c/core/oai/test/spgw_task/test_spgw_procedures.cpp#L623 fails with Bazel with [1].
  * with make this test is successful
* The matcher https://github.com/magma/magma/blob/331551e2289d30fdbd6c811e8d2a46d95fd5c6b2/lte/gateway/c/core/oai/test/spgw_task/test_spgw_procedures.cpp#L72 is triggered, but the `memcmp` of the arrays failes. With Bazel the passed array (not the array from `arg`) seems to start with five non-deterministic values (from this point of view).
* Creating a copy of the matcher array before defining the `EXPECT_CALL` works for Bazel (also with asan and lsan) and make, i.e., in between something happens with the array when running with Bazel.
  * My understanding of the google test matchers: expected values are passed, i.e., variables should be constants that are not changed by test and/or tested logic. The solution above seems reasonable. Understanding the root cause is still open.
* Unclear if related: during testing was observed that removing https://github.com/magma/magma/blob/331551e2289d30fdbd6c811e8d2a46d95fd5c6b2/lte/gateway/c/core/oai/test/spgw_task/test_spgw_procedures.cpp#L361 and lines until the end of the function results in a still green test (with make) - obviously, because additional testing and asserts are removed. But, removing now https://github.com/magma/magma/blob/331551e2289d30fdbd6c811e8d2a46d95fd5c6b2/lte/gateway/c/core/oai/test/spgw_task/test_spgw_procedures.cpp#L359 results in similar behavior as with Bazel (the "random" pattern of the array is slightly different).

[1]
```
lte/gateway/c/core/oai/test/spgw_task/test_spgw_procedures.cpp:365: EXPECT_CALL(*mme_app_handler, mme_app_handle_nw_init_bearer_deactv_req( check_params_in_deactv_bearer_req( sample_gx_nw_init_ded_bearer_deactv_req.no_of_bearers, sample_gx_nw_init_ded_bearer_deactv_req.ebi )))...
  Expected arg #0: check params in deactv bearer req (1, 0x7ffe36289f89)
           Actual: 20-byte object <01-06 00-00 00-00 00-00 00-00 00-00 01-00 00-00 00-00 00-00>
         Expected: to be called once
           Actual: never called - unsatisfied and active
```

jheidbrink:
Fixes #11984

## Summary

Moved one `EXPECT_CALL` from the `deactivate_dedicated_bearer` to the `TestDedicatedBearerDeactivation` `TEST_F` macro. We found that this fixes the test when run with Bazel. Also works with `asan` and `lsan` options.

## Test Plan

The `bazel test` commands have to be run on the `pr-giant-bazel-mme-rebase` branch (with these changes cherry-picked).

```
bazel test lte/gateway/c/core/oai/test/spgw_task:spgw_procedures_test --runs_per_test=100
bazel test lte/gateway/c/core/oai/test/spgw_task:spgw_procedures_test --config=asan
bazel test lte/gateway/c/core/oai/test/spgw_task:spgw_procedures_test --config=lsan
cd $MAGMA_ROOT/lte/gateway && make test_oai
```

## Additional Information

We have not understood why `bazel test` fails without this change. With `print` debugging, we were able to see that the `MATCHER_P2` doesn't seem to get the parameters that we would expect. Out of curiosity, we tested if we can outsource the `EXPECT_CALL` to another function, but that also fails (with the same error as without this change). It seems we have to call it directly from the `TEST_F` macro.

### DIFF
--- a/lte/gateway/c/core/oai/test/spgw_task/test_spgw_procedures.cpp
+++ b/lte/gateway/c/core/oai/test/spgw_task/test_spgw_procedures.cpp
@@ -342,14 +342,6 @@ void SPGWAppProcedureTest ::deactivate_dedicated_bearer(
       &sample_gx_nw_init_ded_bearer_deactv_req, test_imsi_str,
       DEFAULT_EPS_BEARER_ID, ded_eps_bearer_id);
 
-  // check that MME gets a bearer deactivation request
-  EXPECT_CALL(*mme_app_handler,
-              mme_app_handle_nw_init_bearer_deactv_req(
-                  check_params_in_deactv_bearer_req(
-                      sample_gx_nw_init_ded_bearer_deactv_req.no_of_bearers,
-                      sample_gx_nw_init_ded_bearer_deactv_req.ebi)))
-      .Times(1);
-
   return_code = spgw_handle_nw_initiated_bearer_deactv_req(
       &sample_gx_nw_init_ded_bearer_deactv_req, test_imsi64);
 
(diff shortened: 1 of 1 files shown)


themarwhal:
After this merges (https://github.com/magma/magma/pull/12142), please remove the manul tag as a follow up. Thanks :)

LKreutzer (approved):
In pairing.

ssanadhya (approved):
lgtm

Issue #10492, opened 2021-11-22
Toward Magma Distributed Tracing


hcgatewood:
# Toward Magma Distributed Tracing

## tl;dr

Distributed tracing unlocks developers with **enhanced debugging tools** and empowers operators with **radically improved visibility** into their deployments. Imagine real-time, dev-or-prod, passively-generated request tracing — **across Orc8r and gateways**.

If you zoom in on [the picture below](https://www.digitalocean.com/community/tutorials/how-to-implement-distributed-tracing-with-jaeger-on-kubernetes), you’ll see a cross-service, end-to-end understanding of a request’s path through an example application. This includes **services visited, latencies, annotations, and more** — both for application-level code, as well as for DB operations.

<img width="1680" alt="trace-detail-ss (1)" src="https://user-images.githubusercontent.com/8029544/142932167-d0a040c8-13d4-4b01-9fa4-742ca450100d.png">

## Intro

This document aims to provide a guiding outline for how to progressively outfit the Magma project with distributed tracing functionalities.

In particular, we propose

* A particular set of technologies
* An architecture to handle reporting gateway traces

## Context

This section provides context on the distributed tracing landscape, for the chosen set of technologies. Note that we’re proposing to use both Jaeger and OpenTelemetry together as complementary technologies.

For an intro to distributed tracing, see the [short tracing intro](https://www.youtube.com/watch?v=WqMLZF0_Tzc&ab_channel=NewRelic) or [comprehensive Jaeger intro](https://www.youtube.com/watch?v=UNqilb9_zwY&ab_channel=CNCF%5BCloudNativeComputingFoundation%5D).

### Why distributed tracing

The Magma project has grown large enough that debugging, especially root-causing performance bottlenecks, has become unwieldy. Distributed tracing provides a mechanism for an online, real-time understanding of how requests pass through a series of services. This will unlock developers with enhanced debugging tools and empower operators with radically improved visibility into their deployments.

### [Jaeger](https://www.jaegertracing.io/)

CNCF graduated project. End-to-end solution for outfitting a project with distributed tracing. Compatible with OpenTracing (and [mostly+imminently compatible with OpenTelemetry](https://www.jaegertracing.io/docs/1.21/opentelemetry/)). Supports multiple storage backends, defaulting to Elasticsearch. Also supports [tunable sampling rates and patterns](https://www.jaegertracing.io/docs/1.22/sampling/), to limit network and storage pressure.

Includes the following components

[...]

PR #12151, opened 2022-03-16
chore: Upgrade to Go 1.18

### LINKED ISSUE #4055: Upgrade to current Go version
https://golang.org/dl/

### LINKED ISSUE #12559: Un-pin `golang.org/x/net`
To upgrade protobuf in #12546 we pinned `golang.org/x/net` with replace directives, since the version chosen by `go mod tidy` required Go 1.17.

In Go upgrade these replace directives should be removed again.

alexzurbonsen:
## Summary

+--------------------------------------+
  Note: Merge together with #12422.  
+--------------------------------------+

This PR upgrades Magma to Go 1.18. While there are various good reasons to upgrade Go to a recent version, the initial motivation for us was the implementation of the distributed tracing proposal (#10492). The needed Open Telemetry modules rely on Go >= 1.16.

Closes https://github.com/magma/magma/issues/4055, closes https://github.com/magma/magma/issues/12559.

We recommend to review commit by commit. (Many changes are in generated code, which we put into the `./build.py --generate` commit).

#### go-swagger
With Go various other modules have to be upgraded as well. The most influential change is the version bump in `go-swagger` from `v0.21.0` to `v0.29.0` (currently latest). It was necessary to get swagger generation going.

The upgrade in `go-swagger` causes changes in generated files and requires minor changes in code and tests at various places to maintain previous functionality. They are bundled in the following commits:
- _snapshot tests: RewriteTest and GenerateTest: substitute expected..._
- _fix build and test errors due to swagger generation..._

#### go install and go get
Behaviour of `go install` and `go get` changes in Go >= 1.16 (see https://go.dev/doc/go1.16 for `go install` and https://go.dev/doc/go1.18 for `go get`).

If used within a module, a version suffix has to be used to ignore module requirements (e.g. `go install somepackage@v1.2.3`). Without version suffix an entry in `go.mod` file is required. This causes commands like `make tools` to fail with Go >=1.16. We thus updated existing tools.go files and introduced new tools.go files in commit _Update existing tools.go files..._

Commit _Replace go get with go install..._ is due to changes of `go get` behaviour.

#### Dockerfiles 
Dockerfiles are updated to use Go 1.18 in commits _Update Dockerfiles to Go 1.18_ and _update orc8r_deployer dockerfile..._.

VM and DevContainer can be upgraded independently once this PR is merged (see #12422 and #12424).

#### CI
Commit _Update Go to 1.18 in CI_ switches to Go 1.18 in various workflows. We also filed two separate PRs with CI Updates in #12352 and #12305 as prerequisite for this PR.

#### mockery and go-acc
An upgrade of mockery to v2 and go-acc to v0.2.8 is also necessary and done in commits _upgrade mockery to v2..._ and _upgrade go-acc to get it running..._. Since we changed some mocked functions to use pointers commit _generate new mockery files_ is needed.

#### go-openapi in cwf_operator module
To fix an error similar to this one https://github.com/kubernetes-sigs/kustomize/tree/master/cmd/depprobcheck certain versions of go-openapi modules were pinned in commit _Replace versions of go-openapi/spec…_ to resolve incomaptible versions.

## Test Plan

- [x] CI (Also ran various checks locally, but they are all included in the CI.)
- [x] local tests with running orc8r, AGW and NMS: tested various workflows like
   - end2end workflow of updating config yaml file in NMS
   - setting configs with swagger
   - propagation of mconfig to AGW
- [x] Precommit checks in feg/gateway/docker and orc8r/cloud/docker, `./build.py -c`
- [x] cwag precommit test, `vagrant up cwag && vagrant ssh cwag && cd magma/cwf/gateway && make precommit` (relies on first doing `vagrant up cwag --provision` with https://github.com/magma/magma/pull/12422)
- [x] cwag integ test: https://docs.magmacore.org/docs/cwf/dev_testing#run-integration-tests
- [x] Federated integ test: https://docs.magmacore.org/docs/next/feg/s1ap_federated_tests#s1ap-federated-integration-tests
- [x] Build and deploy a K8s Orc8r (can be locally using Minikube)

## Open point for follow-up PR
- Run `make gen` in `orc8r/cloud/api/v1/`. This generates changes in roughly 1000 files, might be easier to review this one if done in a seprate PR.

## Additional Information

- Done in pairing with @sebathomas, @voisey, @ajahl, @Neudrino.
- Once this PR is merged you will have to run `vagrant up magma --provision` to install Go 1.18 on your VM.


feg/cloud/go/tools.go
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+*/
+
+package main
+
+// Package tools is a convention to ensure `go mod tidy` adds development tool
+// packages to go.mod / go.sum.
+
+import (
+	_ "magma/orc8r/cloud/go/tools/swaggergen"
```
uri200, line 23: why is this file needed here? 
tools directory contain just tools but they are not touching swagger

feg/gateway/services/s8_proxy/servicers/s8_proxy_test.go
```diff
@@ -623,7 +624,9 @@ func TestS8proxyCreateSessionNoProtocolConfigurationOptions(t *testing.T) {
 
 	// check PCO
 	assert.NoError(t, err)
-	assert.Empty(t, csRes.ProtocolConfigurationOptions)
+	emptyProcConfOpts := &protos.ProtocolConfigurationOptions{ConfigProtocol: 0, ProtoOrContainerId: nil}
+	equal := proto.Equal(emptyProcConfOpts, csRes.ProtocolConfigurationOptions)
+	assert.True(t, equal)
```
uri200: why is Empty not valid anymore

feg/gateway/services/swx_proxy/client_api_test.go
```diff
@@ -91,17 +93,21 @@ func standardSwxProxyTest(t *testing.T) {
 		t.Fatalf("GRPC SAR Register Error: %v", err)
 		return
 	}
-	assert.Equal(t, &protos.RegistrationAnswer{SessionId: regRes.GetSessionId()}, regRes)
-	t.Logf("GRPC Register SAA: %#+v", *regRes)
+	equal := proto.Equal(&protos.RegistrationAnswer{SessionId: regRes.GetSessionId()}, regRes)
+	assert.True(t, equal)
```
uri200: why are we adding lines of code like this to the tests?

feg/gateway/services/swx_proxy/client_api_test.go
```diff
@@ -91,17 +93,21 @@ func standardSwxProxyTest(t *testing.T) {
...
+	regResJSON, _ := orcprotos.MarshalIntern(regRes)
+	t.Logf("GRPC Register SAA: %v", string(regResJSON))
```
uri200: those logs are just for debugging. I would not Marshall them 
Adding that function all around the code may slow down test execution on the long run

feg/gateway/services/testcore/hss/servicers/hss_test.go
```diff
@@ -155,7 +157,9 @@ func TestHomeSubscriberServer_GetSubscriberDataGrpc(t *testing.T) {
 	assert.EqualError(t, err, "rpc error: code = NotFound desc = Subscriber '100' not found")
 
 	reply, err := client.AddSubscriber(context.Background(), &sub)
-	assert.Equal(t, orcprotos.Void{}, *reply)
+	replyCopy := proto.Clone(reply).(*orcprotos.Void)
+	equal := proto.Equal(&orcprotos.Void{}, replyCopy)
+	assert.True(t, equal)
```
uri200: is this something needed to fix any new linter warning?

feg/gateway/services/testcore/pcrf/mock_pcrf/mock_driven_pcrf_test.go
```diff
@@ -297,3 +300,82 @@ func getMockReAuthHandler() gx.PolicyReAuthHandler {
 		}
 	}
 }
+
```
uri200: lets add the name of the function at the beginning of the comment

```
// equalRDefSlices compare two slices with *fegproto...
```

feg/gateway/services/testcore/pcrf/mock_pcrf/mock_driven_pcrf_test.go
```diff
@@ -297,3 +300,82 @@ func getMockReAuthHandler() gx.PolicyReAuthHandler {
...
+// compare two slices with *fegprotos.RuleDefinition message pointers. Slices can be unsorted.
+// If duplicates exist, each has to appear the same number of times in both lists.
+// Equality of elements is determined with proto.Equal
+func equalRDefSlices(expect []*fegprotos.RuleDefinition, actual []*fegprotos.RuleDefinition) bool {
```
uri200: can you use full name please `equalRuleDefinitionSlices`

feg/gateway/services/testcore/pcrf/mock_pcrf/mock_driven_pcrf_test.go
```diff
@@ -297,3 +300,82 @@ func getMockReAuthHandler() gx.PolicyReAuthHandler {
...
+	for _, act := range actual {
+		for j, exp := range expect {
+			if contains(matched, j) {
+				continue
+			} else if proto.Equal(exp, act) {
+				matched = append(matched, j)
+				break
+			}
+		}
+	}
```
uri200: This algorithm is a bit confusing. Why not to use a map with the rule name as a key and a pointer to the rule?

You build two maps and then you can compare key by key.

Note we can not install two rules with the same name, so that must trigger an error

feg/gateway/services/testcore/pcrf/mock_pcrf/mock_driven_pcrf_test.go
```diff
@@ -297,3 +300,82 @@ func getMockReAuthHandler() gx.PolicyReAuthHandler {
...
 }
+
+// compare two slices with *fegprotos.RuleDefinition message pointers. Slices can be unsorted.
+// If duplicates exist, each has to appear the same number of times in both lists.
+// Equality of elements is determined with proto.Equal
+func equalRDefSlices(expect []*fegprotos.RuleDefinition, actual []*fegprotos.RuleDefinition) bool {
+	if len(expect) != len(actual) {
+		return false
+	}
+	var matched []int
```
uri200: you can use a map to implement a set, this way you don't need `contains`

```
var matched map[int]struct{} 
```

https://yourbasic.org/golang/implement-set/

feg/gateway/tools/s6a_cli/main.go
```diff
@@ -272,7 +274,7 @@ func air(cmd *commands.Command, args []string) int {
 				errChann <- errCli
 				return
 			}
-			fmt.Printf("Sending AIR to %s:\n%s\n%+#v\n\n", peerAddr, json, *req)
+			fmt.Printf("Sending AIR to %s:\n%s\n%+#v\n\n", peerAddr, json, proto.Clone(req).(*protos.AuthenticationInformationRequest))
```
uri200: does go 1.18 have issues with pointers?

feg/gateway/tools/s8_cli/main.go
```diff
@@ -521,5 +521,8 @@ func main() {
 	}
 	args := os.Args[2:]
 	cmd.Flags().Parse(args)
-	os.Exit(cmd.Handle(args))
+	// "If a test calls os.Exit, the test binary will exit, and go test will report a failure." https://github.com/golang/go/issues/45619
+	if i := cmd.Handle(args); i != 0 {
+		os.Exit(i)
+	}
```
uri200: maybe it would be best just remove the os.Exit(0) we have in this cli and just use return
And then just run the command in this line
```
cmd.Handle(args)
```

feg/gateway/services/eap/providers/aka/servicers/handlers/auth_utils.go
```diff
@@ -101,14 +102,15 @@ func getSwxVector(s *servicers.EapAkaSrv, imsi string, resyncInfo []byte) (*tgpp
 	if ans == nil {
 		return nil, status.Error(codes.Internal, "Error: Nil SWx Response")
 	}
+	ansJSON, _ := orcprotos.MarshalIntern(ans)
```
emakeev: why would you take a performance hit of needlessly serializing message if it's only needed  (pls, see below regarding needing it) on error?

feg/gateway/services/eap/providers/aka/servicers/handlers/auth_utils.go
```diff
@@ -101,14 +102,15 @@ func getSwxVector(s *servicers.EapAkaSrv, imsi string, resyncInfo []byte) (*tgpp
...
 	if len(ans.SipAuthVectors) == 0 {
-		return nil, status.Errorf(codes.Internal, "Error: Missing/empty SWx Auth Vector: %+v", *ans)
+		return nil, status.Errorf(codes.Internal, "Error: Missing/empty SWx Auth Vector: %v", string(ansJSON))
```
emakeev: why do you need marshaled string? please see: https://go.dev/play/p/fZBzU5CGtWr
you can just pick a format you like to print.

feg/gateway/services/eap/providers/aka/servicers/handlers/auth_utils.go
```diff
@@ -101,14 +102,15 @@ func getSwxVector(s *servicers.EapAkaSrv, imsi string, resyncInfo []byte) (*tgpp
...
 	}
 	av := ans.SipAuthVectors[0] // Use first vector for now
 	ra := av.GetRandAutn()
 	if len(ra) < aka.RandAutnLen {
 		return nil, status.Errorf(codes.Internal,
-			"Invalid SWx RandAutn len (%d, expected: %d) in Response: %+v", len(ra), aka.RandAutnLen, *ans)
+			"Invalid SWx RandAutn len (%d, expected: %d) in Response: %v", len(ra), aka.RandAutnLen, string(ansJSON))
```
emakeev: same as above

feg/gateway/services/eap/providers/sim/servicers/handlers/auth_utils.go
```diff
@@ -96,14 +97,15 @@ func getSwxVectors(_ *servicers.EapSimSrv, imsi string) (*tgppAuthResult, error)
 		if ans == nil {
 			return nil, status.Error(codes.Internal, "Error: Nil SWx Response")
 		}
+		ansJSON, _ := orcprotos.MarshalIntern(ans)
```
emakeev: No need to marshal struct before you even know if you need it & in 99.9% of cases - you won't need it...

feg/gateway/services/eap/providers/sim/servicers/handlers/auth_utils.go
```diff
@@ -96,14 +97,15 @@ func getSwxVectors(_ *servicers.EapSimSrv, imsi string) (*tgppAuthResult, error)
...
 		if len(ans.SipAuthVectors) == 0 {
-			return nil, status.Errorf(codes.Internal, "Error: Missing/empty SWx Auth Vector: %+v", *ans)
+			return nil, status.Errorf(codes.Internal, "Error: Missing/empty SWx Auth Vector: %v", string(ansJSON))
```
emakeev: what's wrong with %v, %+v or %#v ?

feg/gateway/services/eap/providers/sim/servicers/handlers/auth_utils.go
```diff
@@ -96,14 +97,15 @@ func getSwxVectors(_ *servicers.EapSimSrv, imsi string) (*tgppAuthResult, error)
...
 		}
 		for _, v := range ans.GetSipAuthVectors() {
 			ra := v.GetRandAutn()
 			if len(ra) < sim.RAND_LEN {
 				return nil, status.Errorf(codes.Internal,
-					"Invalid SWx RandAutn len (%d, expected: %d) in Response: %+v", len(ra), sim.RandAutnLen, *ans)
+					"Invalid SWx RandAutn len (%d, expected: %d) in Response: %v", len(ra), sim.RandAutnLen, string(ansJSON))
```
emakeev: same as above

feg/gateway/services/swx_proxy/servicers/config.go
```diff
@@ -83,7 +84,8 @@ func GetSwxProxyConfig() []*SwxProxyConfig {
 		}
 	}
 
-	glog.V(2).Infof("Loaded %s configs: %+v", SwxProxyServiceName, *configsPtr)
+	configsPtrJSON, _ := orcprotos.MarshalIntern(configsPtr)
+	glog.V(2).Infof("Loaded %s configs: %v", SwxProxyServiceName, string(configsPtrJSON))
```
emakeev: Please avoid needlessly marshaling message if you ONLY need it for debug printout.

feg/gateway/services/testcore/hss/hss/main.go
```diff
@@ -68,8 +69,9 @@ func main() {
 	}
 	// Start diameter server
 	startedChan := make(chan string, 1)
+	configJSON, _ := orcprotos.MarshalIntern(servicer.Config)
 	go func() {
-		log.Printf("Starting home subscriber server with configs:\n\t%+v", *servicer.Config)
+		log.Printf("Starting home subscriber server with configs:\n\t%v\n", string(configJSON))
```
emakeev: what's wrong with %v & Co.?

feg/gateway/services/testcore/hss/servicers/config.go
```diff
@@ -86,7 +87,8 @@ func GetHSSConfig() (*mconfig.HSSConfig, error) {
 		}, err
 	}
 
-	glog.V(2).Infof("Loaded %s configs: %+v\n", hssServiceName, *configsPtr)
+	configsPtrJSON, _ := orcprotos.MarshalIntern(configsPtr)
+	glog.V(2).Infof("Loaded %s configs: %v\n", hssServiceName, string(configsPtrJSON))
```
emakeev: use %v here

feg/gateway/tools/s6a_cli/main.go
emakeev: Why do you need to clone to print?

feg/gateway/tools/s6a_cli/main.go
```diff
@@ -282,11 +284,11 @@ func air(cmd *commands.Command, args []string) int {
 			}
 			json, errCli = orcprotos.MarshalIntern(r)
 			if errCli != nil {
-				errCli = fmt.Errorf("Marshal Error %v for result: %+v", errCli, *r)
+				errCli = fmt.Errorf("Marshal Error %v for result: %+v", errCli, proto.Clone(r).(*protos.AuthenticationInformationAnswer))
```
emakeev: why do you need to clone to Marshal?

feg/gateway/tools/s6a_cli/main.go
```diff
@@ -282,11 +284,11 @@ func air(cmd *commands.Command, args []string) int {
...
 				errChann <- errCli
 				return
 			}
-			fmt.Printf("Received AIA:\n%s\n%+v\n", json, *r)
+			fmt.Printf("Received AIA:\n%s\n%+v\n", json, proto.Clone(r).(*protos.AuthenticationInformationAnswer))
```
emakeev: why do you need to clone to print?

feg/gateway/tools/s8_cli/main.go
```diff
@@ -521,5 +521,8 @@ func main() {
 	}
 	args := os.Args[2:]
 	cmd.Flags().Parse(args)
-	os.Exit(cmd.Handle(args))
+	// "If a test calls os.Exit, the test binary will exit, and go test will report a failure." https://github.com/golang/go/issues/45619
```
emakeev: does the issue affect s8_cli tool? The issue affects Go tests not applications. So, why this change?

~/magma/feg/gateway/tools$ go test ./...
?   	magma/feg/gateway/tools/gw_csfb_service_cli	[no test files]
?   	magma/feg/gateway/tools/gw_s6a_service_cli	[no test files]
?   	magma/feg/gateway/tools/gx_client_cli	[no test files]
?   	magma/feg/gateway/tools/gy_client_cli	[no test files]
?   	magma/feg/gateway/tools/hello_cli	[no test files]
?   	magma/feg/gateway/tools/hss_cli	[no test files]
?   	magma/feg/gateway/tools/s6a_cli	[no test files]
?   	magma/feg/gateway/tools/s8_cli	[no test files]
?   	magma/feg/gateway/tools/swx_cli	[no test files]

~/magma/feg/gateway/tools$ go version
go version go1.18 darwin/arm64

feg/gateway/tools/swx_cli/main.go
```diff
@@ -224,7 +226,7 @@ func sendSar(addr string, client swxClient) int {
 	if err != nil {
 		fmt.Print("Unable to convert request to JSON for printing; Still attempting to send request...")
 	} else {
-		fmt.Printf("Sending SAR (REGISTER) to %s:\n%s\n%+#v\n\n", addr, json, *req)
+		fmt.Printf("Sending SAR (REGISTER) to %s:\n%s\n%+#v\n\n", addr, json, proto.Clone(req).(*protos.RegistrationRequest))
```
emakeev: clone?

feg/gateway/tools/swx_cli/main.go
```diff
@@ -256,10 +258,10 @@ func sendMar(addr string, client swxClient) int {
 	}
 	json, err = orcprotos.MarshalIntern(res)
 	if err != nil {
-		fmt.Printf("Marshal Error %v for result: %+v", err, *res)
+		fmt.Printf("Marshal Error %v for result: %+v", err, proto.Clone(res).(*protos.AuthenticationAnswer))
```
emakeev: clone?

feg/gateway/tools/swx_cli/main.go
```diff
@@ -256,10 +258,10 @@ func sendMar(addr string, client swxClient) int {
...
 		return 3
 	}
-	fmt.Printf("Received successful MAA:\n%s\n%+v\n", json, *res)
+	fmt.Printf("Received successful MAA:\n%s\n%+v\n", json, proto.Clone(res).(*protos.AuthenticationAnswer))
```
emakeev: clone?

lte/cloud/go/services/eps_authentication/servicers/eps_auth_suite_test.go
```diff
@@ -122,13 +122,14 @@ func addSubscriber(store storage.SubscriberDBStorage, sd *lteprotos.SubscriberDa
 		Config: &sdb_models.SubscriberConfig{},
 	}
 	if sd.GetLte() != nil {
+		SubProfile := sdb_models.SubProfile(sd.GetSubProfile())
```
emakeev: lowercase?

lte/cloud/go/services/subscriberdb/obsidian/models/conversion.go
```diff
@@ -144,8 +144,9 @@ func (m *MutableSubscriber) FromEnt(ent configurator.NetworkEntity, policyProfil
 		model.StaticIps = config.StaticIps
 		// If no profile in backend, return "default"
 		// TODO(8/21/20): enforce this at the API layer (and include data migration)
-		if model.Lte.SubProfile == "" {
-			model.Lte.SubProfile = "default"
+		if model.Lte.SubProfile != nil && *model.Lte.SubProfile == "" {
```
emakeev: if model.Lte.SubProfile == nil || *model.Lte.SubProfile == "" {  ...

orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
```diff
@@ -66,9 +67,10 @@ func (r *RegistrationService) Register(c context.Context, request *protos.Regist
 	return res, nil
 }
 
-func RegisterDevice(deviceInfo protos.GatewayDeviceInfo, hwid *protos.AccessGatewayID, challengeKey *protos.ChallengeKey) error {
+func RegisterDevice(deviceInfo *protos.GatewayDeviceInfo, hwid *protos.AccessGatewayID, challengeKey *protos.ChallengeKey) error {
+	deviceInfoCopy := proto.Clone(deviceInfo).(*protos.GatewayDeviceInfo)
```
emakeev: you seem to clone every proto message everywhere. Why? Anything changed for Go 1.18 in this regard?

emakeev (changes requested):

emakeev:
would it be possible to split this PR into set of smaller PRs? 
For example: PR1 - swagger upgrade, PR2 - proto upgrade, PR3 - Go 1.18 upgrade.

It'd keep things simple & easier to review.

feg/cloud/go/tools.go
alexzurbonsen: `go install` behaviour changes in Go >= 1.16. Without this file and the line added `make swagger_tools` won't work in feg/cloud/go
https://github.com/magma/magma/blob/5bca245d7be0e53eb2c78dac408a6c3210300369/orc8r/cloud/go/module.mk#L92-L93

feg/gateway/services/swx_proxy/client_api_test.go
alexzurbonsen: These lines are necessary because of the new protobuf implementation. Messages now contain a `sync.Mutex` such that copying is detected in govet checks. Let me copy an example of such a failing check for clarity:

```
services/tenants/servicers/protected/servicer_test.go:154:18: copylocks: call of assert.Equal copies lock value: magma/orc8r/cloud/go/services/tenants/protos.GetControlProxyResponse contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex (govet)
	assert.Equal(t, *controlProxy, sampleGetControlProxyRes)
```

This can be avoided by using `proto.Equal` for comparison.

Also the assert.Equal comparison takes into account internal proto fields, which causes the comparison to fail often.

feg/gateway/services/s8_proxy/servicers/s8_proxy_test.go
alexzurbonsen: same as below

feg/gateway/services/swx_proxy/client_api_test.go
alexzurbonsen: Unfortunately passing the message directly with a "%v" format string is failing the govet check as well due to disallowed copying. Thus this workaround.

feg/gateway/services/testcore/hss/servicers/hss_test.go
alexzurbonsen: Yes, see the example above.

feg/gateway/services/testcore/pcrf/mock_pcrf/mock_driven_pcrf_test.go
alexzurbonsen: Thanks for the explanation. The original function `assert.ElementsMatch` allows for duplicate elements such that we thought we would have to allow for that as well. Will change it accordingly.

feg/gateway/services/testcore/pcrf/mock_pcrf/mock_driven_pcrf_test.go
alexzurbonsen: Thanks, that is much better...

feg/gateway/tools/s6a_cli/main.go
alexzurbonsen: It is protobuf again. See the more detailed description in the summary of this PR as well. This is done to pass copylocks check of govet.

feg/gateway/services/eap/providers/sim/servicers/handlers/auth_utils.go
alexzurbonsen: This is again due to the new protobuf implementation. Unfortunately passing the message directly with a format specifier like "%v" does not pass govet`s coyplocks check. See PR summary for more detailed description.
I copy an example of another case here for clarity:

```
services/s6a_proxy/servicers/s6a_proxy_test.go:148:28: copylocks: call of t.Logf copies lock value: magma/feg/cloud/go/protos.PurgeUEAnswer contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex (govet)
		t.Logf("GRPC PUA: %#+v", *puResp)
```

feg/gateway/services/eap/providers/sim/servicers/handlers/auth_utils.go
alexzurbonsen: see comment above

feg/gateway/services/testcore/hss/hss/main.go
alexzurbonsen: see comment above

feg/gateway/services/testcore/hss/servicers/config.go
alexzurbonsen: see comment above

feg/gateway/tools/s6a_cli/main.go
alexzurbonsen: Doesn't pass copylocks check otherwise (unfortunately)

feg/gateway/tools/s6a_cli/main.go
alexzurbonsen: same as above

feg/gateway/tools/s6a_cli/main.go
alexzurbonsen: see comment above

feg/gateway/tools/swx_cli/main.go
alexzurbonsen: see comment above

feg/gateway/tools/swx_cli/main.go
alexzurbonsen: see comment above

feg/gateway/tools/swx_cli/main.go
alexzurbonsen: see comment above

orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
alexzurbonsen: see comment above

feg/gateway/services/eap/providers/sim/servicers/handlers/auth_utils.go
alexzurbonsen: moving it inside the if block

feg/gateway/services/eap/providers/aka/servicers/handlers/auth_utils.go
alexzurbonsen: please see comment below https://github.com/magma/magma/pull/12151#discussion_r846662123

feg/gateway/services/eap/providers/aka/servicers/handlers/auth_utils.go
alexzurbonsen: please see here https://github.com/magma/magma/pull/12151#discussion_r846662123

uri200:
You may also need to take a look at `wrappers` module used on CWF integ test 

```
&wrappers.StringValue{Value: "2M"},
```


btw you will need to run precommit check at `feg/gateway/docker`  and `orc8r/cloud/docker` 
```
./build.py -c
```
And at cwg `/cwf/gateway`
```
(maybe you can do it without vagrant, but just in case)
vagrant up cwag
vagrant ssh cwag
cd magma/cwf/gateway
make precommit
```

feg/gateway/services/testcore/hss/servicers/hss_test.go
uri200: Using 

```
	assert.True(t, proto.Equal(&orcprotos.Void{}, reply))
```
With your PR, I am able to pass `go vet ./...` and lint

```
root@docker-desktop:/magma/feg/gateway# go version
go version go1.18 linux/amd64
root@docker-desktop:/magma/feg/gateway# go vet ./...
root@docker-desktop:/magma/feg/gateway# make lint
/usr/sbin/golangci-lint
-> golangci-lint already installed
golangci-lint run --config /magma/.golangci.yml
root@docker-desktop:/magma/feg/gateway#
```

Am I looking at the wrong place

feg/gateway/services/swx_proxy/client_api_test.go
uri200: If I print the pointer itself I am able to bypass the govet issue

```
	t.Logf("GRPC Register SAA: %#+v", regRes)
```
And the log result looks like the same (just for the & indicating it is a pointer)

With Pointer 

image


without the pointer 

image

feg/gateway/services/swx_proxy/client_api_test.go
uri200: I guess if you use the pointer itself you will not be copying the mutex inside the proto. 

If I print the pointer (`regRes` , not `*regRes`) I am not triggering the linter issue

```
	t.Logf("GRPC Register SAA: %#+v", regRes)
```
And the log result looks like the same (just for the & indicating it is a pointer)

With Pointer (`*regRes`)

image


without the pointer (`regRes`)

image

feg/gateway/services/eap/providers/sim/servicers/handlers/auth_utils.go
uri200: please try to use `puResp` instead of `*puResp`

lte/cloud/go/services/subscriberdb/obsidian/models/conversion.go
maxhbr: hey @emakeev , you can use the <code>```suggestion</code> block, to make it very easy to see the changes you are suggesting:

```suggestion
		if model.Lte.SubProfile != nil || *model.Lte.SubProfile == "" {
```

uri200:
As mentioned by @emakeev , I would suggest to place those

feg/gateway/tools/s8_cli/main.go
uri200: note s8_cli has a test file that is running with 
```
go test -tags cli_test magma/feg/gateway/tools/...
```

or you can run it doing `make test`

```
root@docker-desktop:/magma/feg/gateway# go test -tags cli_test magma/feg/gateway/tools/...
?   	magma/feg/gateway/tools/gw_csfb_service_cli	[no test files]
?   	magma/feg/gateway/tools/gw_s6a_service_cli	[no test files]
?   	magma/feg/gateway/tools/gx_client_cli	[no test files]
?   	magma/feg/gateway/tools/gy_client_cli	[no test files]
?   	magma/feg/gateway/tools/hello_cli	[no test files]
?   	magma/feg/gateway/tools/hss_cli	[no test files]
?   	magma/feg/gateway/tools/s6a_cli	[no test files]
ok  	magma/feg/gateway/tools/s8_cli	(cached)
?   	magma/feg/gateway/tools/swx_cli	[no test files]
```


Note the s8_cli test was created a s a workaround because we were using go 1.12. Naybe go 1.18 have better testing tool for testing cli commands

feg/gateway/services/eap/providers/aka/servicers/handlers/auth_utils.go
ajahl: Is moved into the error statement.

uri200:
You should also try to run the federated test. That test will run agw, orc8r and feg and you will see if there is something broken between gateways after your changes.

Note this and other end to end test that are will be impacted by this change are not run on CI


https://docs.magmacore.org/docs/next/feg/s1ap_federated_tests#s1ap-federated-integration-tests

feg/gateway/services/eap/providers/sim/servicers/handlers/auth_utils.go
emakeev: would using ans instead of *ans work for you? Passing by ref/pointer won't copy.

feg/gateway/services/eap/providers/aka/servicers/handlers/auth_utils.go
emakeev: note that you can use pointer to struct with %v. would switching from *ans to ans work 4 you?

ajahl: works

feg/gateway/tools/s8_cli/main.go
ajahl: running `go test -tags cli_test magma/feg/gateway/tools/...` produces the error: 

```
--- FAIL: TestS8Cli (0.01s)
panic: unexpected call to os.Exit(0) during test [recovered]
        panic: unexpected call to os.Exit(0) during test
```

so we change the file as you described above (https://github.com/magma/magma/pull/12151#discussion_r846363251).

feg/gateway/services/eap/providers/aka/servicers/handlers/auth_utils.go
uri200: Pleaae for better traceability, let the owner of the comment hit the resolve button.

cwf/gateway/integ_tests/test_runner.go
```diff
@@ -477,32 +479,32 @@ func (tr *TestRunner) WaitForEnforcementStatsForRuleGreaterThanOrDoesNotExist(im
...
+		return nil, true
 	}
+	txBytes := record.BytesTx
+	if record.BytesTx < min {
+		return record, false
+	}
+	fmt.Printf("\t\u2713 %s, %s now passed %d > %d in enforcement stats!(%d%%)\n",
+		imsi, ruleID, txBytes, min, 100*txBytes/min)
+	return record, true
+}
```
uri200: what is the change in this function?

feg/gateway/services/aaa/radius/config.go
```diff
@@ -37,7 +39,7 @@ var defaultConfigs = &mconfig.RadiusConfig{
 func ValidateConfigs(cfg *mconfig.RadiusConfig) *mconfig.RadiusConfig {
 	res := &mconfig.RadiusConfig{}
 	if cfg != nil {
-		*res = *cfg
+		res = proto.Clone(cfg).(*mconfig.RadiusConfig)
```
uri200: This seems wrong. We are cloning an existing object when we validate the config. 
Please can you try to play with the code to remove that clone function

feg/gateway/services/testcore/pcrf/mock_pcrf/mock_driven_pcrf_test.go
```diff
@@ -297,3 +300,81 @@ func getMockReAuthHandler() gx.PolicyReAuthHandler {
...
+			}
+		}
+	}
+	return len(matched) == len(expect)
+}
+
+// equalUMISlices compares two slices with *fegprotos.UsageMonitoringInformation message pointers. Slices can be unsorted.
+// If duplicates exist, each has to appear the same number of times in both lists.
+// Equality of elements is determined with proto.Equal
+func equalUMISlices(expect []*fegprotos.UsageMonitoringInformation, actual []*fegprotos.UsageMonitoringInformation) bool {
```
uri200: please use full names

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
```diff
@@ -3350,7 +3351,7 @@ func TestAPNResource(t *testing.T) {
 		ParamNames:             []string{"network_id"},
 		ParamValues:            []string{"n0"},
 		ExpectedStatus:         500, // this would actually make more sense as a 400, but it's a non-trivial fix
-		ExpectedErrorSubstring: `could not find entities matching [type:"apn" key:"apn0" ]`,
+		ExpectedErrorSubstring: "could not find entities matching" + tests.Separator + "[type:\"apn\"" + tests.Separator + "key:\"apn0\"]",
```
uri200: is this change triggered by the linter?

lte/cloud/go/services/policydb/obsidian/handlers/policy_handlers.go
```diff
@@ -337,7 +337,7 @@ func UpdateRule(c echo.Context) error {
 	if err := rule.ValidateModel(reqCtx); err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, err)
 	}
-	if ruleID != string(rule.ID) {
+	if ruleID != string(*rule.ID) {
```
uri200, line 340: why is rule id  now a pointer?

Note policy rule is used across the system so a change like that may impact other CPP services like sessiond

uri200 (changes requested):
Please run also cwag integ test to verify nothing is broken across gateways

https://docs.magmacore.org/docs/cwf/dev_testing#run-integration-tests

feg/gateway/tools/s8_cli/main.go
voisey: See https://github.com/magma/magma/pull/12151/commits/321cf30f4591665c175a058e4c74ce96567a24ef

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
alexzurbonsen: The problem here was that the tests are failing because an extra whitespace between "type" and "key" fields for example. This is not fixed by removing the whitespace since it appears somewhat randomly. We think it is similar to this issue https://github.com/golang/protobuf/issues/1269. The inserted separator is the workaround.

The comparison now takes the expected error (sub)string, splits it at the separator, and checks wether the parts are contained in the actual message.

feg/gateway/services/aaa/radius/config.go
alexzurbonsen: Can we just use `cfg` directly here or is there a need to copy its values to `res`?

lte/cloud/go/services/policydb/obsidian/handlers/policy_handlers.go
sebathomas: We don't completely understand why some fields (like rule ID and subprofile) became pointers in the generated code after the swaggergen library upgrade. However, we didn't change anything in the model so I don't think it should have any impact on CPP services.

feg/gateway/services/swx_proxy/client_api_test.go
alexzurbonsen: Oh yes...

cwf/gateway/integ_tests/test_runner.go
ajahl: It is only a remove whitespace in `100*txBytes`, triggered by the precommit command I guess

alexzurbonsen:
> would it be possible to split this PR into set of smaller PRs? For example: PR1 - swagger upgrade, PR2 - proto upgrade, PR3 - Go 1.18 upgrade.
> 
> It'd keep things simple & easier to review.

Splitting this PR into smaller pieces isn't really possible since changes depend on each other (e.g. the upgraded go-swagger module requires go 1.17) and GitHub has no concept of stacks.

sebathomas:
> Please run also cwag integ test to verify nothing is broken across gateways
> 
> https://docs.magmacore.org/docs/cwf/dev_testing#run-integration-tests

They are green after adapting the Go version for the cwag VMs. Pushed the changes to #12422.

```
[vagrant@127.0.0.1:2202] out: DONE 40 tests, 11 skipped in 697.958s
[vagrant@127.0.0.1:2202] out: 

[vagrant@127.0.0.1:2202] Executing task '_clean_up'
[vagrant@127.0.0.1:2202] run: pkill go
[localhost] local: pwd
[localhost] local: vagrant status magma_trfserver
[localhost] local: vagrant ssh-config magma_trfserver
[vagrant@127.0.0.1:2201] run: pkill iperf3 > /dev/null &
Integration Test Passed for "all"!
Disconnecting from vagrant@127.0.0.1:2200... done.
Disconnecting from vagrant@127.0.0.1:2201... done.
Disconnecting from vagrant@127.0.0.1:2202... done.
```

ajahl:
> You should also try to run the federated test. That test will run agw, orc8r and feg and you will see if there is something broken between gateways after your changes.
> 
> Note this and other end to end test that are will be impacted by this change are not run on CI
> 
> https://docs.magmacore.org/docs/next/feg/s1ap_federated_tests#s1ap-federated-integration-tests

The s1ap federated integration tests finished successfully.

voisey:
> You may also need to take a look at `wrappers` module used on CWF integ test
> 
> ```
> &wrappers.StringValue{Value: "2M"},
> ```
> 
> btw you will need to run precommit check at `feg/gateway/docker` and `orc8r/cloud/docker`
> 
> ```
> ./build.py -c
> ```
> 
> And at cwg `/cwf/gateway`
> 
> ```
> (maybe you can do it without vagrant, but just in case)
> vagrant up cwag
> vagrant ssh cwag
> cd magma/cwf/gateway
> make precommit
> ```

Thanks for the suggestions! Each of the three precommit tests you mentioned are working on this branch.

alexzurbonsen:
> You may also need to take a look at `wrappers` module used on CWF integ test
> 
> ```
> &wrappers.StringValue{Value: "2M"},
> ```

@uri200 Can you explain a bit more what you had in mind here? In `cwf/gateway` protobuf has already been at version `v1.5.2` on master. And couldn't find this particular line anywhere outside of `cwf/gateway`. Thank you!

uri200:
Then it is fine. I am off this week. I Will take a look after I come back

alexzurbonsen:
Thanks for the review efforts so far! Is there any input from other approvers groups? @m-trojanowski, @tmdzk and @hcgatewood from @magma/approvers-orc8r-infra, @magma/approvers-domain-proxy, @magma/approvers-ci.

Would be great to see this PR go through. It enables the Jaeger/Distributed Tracing efforts :)

.github/workflows/feg-workflow.yml
```diff
@@ -88,7 +88,7 @@ jobs:
         if: always()
         id: feg-precommit
         run: |
-          go get gotest.tools/gotestsum
+          go install gotest.tools/gotestsum@latest
```
hcgatewood: Nit: I generally shy away from pinning to latest, good way to get spontaneous CI breakages

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
```diff
@@ -339,9 +339,10 @@ func listEnodebs(c echo.Context) error {
 	for _, ent := range ents {
 		enodebs[ent.Key] = (&lte_models.Enodeb{}).FromBackendModels(ent)
 	}
+	token := lte_models.PageToken(nextPageToken)
 	paginatedEnodebs := &lte_models.PaginatedEnodebs{
 		Enodebs:    enodebs,
-		PageToken:  lte_models.PageToken(nextPageToken),
+		PageToken:  &token,
```
hcgatewood, line 345: For my understanding: we can't do this inline/single-line?

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
```diff
@@ -1248,7 +1248,7 @@ func TestCreateGateway(t *testing.T) {
 		ParamNames:     []string{"network_id"},
 		ParamValues:    []string{"n1"},
 		ExpectedStatus: 500,
-		ExpectedError:  "error creating gateway: rpc error: code = Internal desc = could not find entities matching [type:\"cellular_enodeb\" key:\"dne\" ]",
+		ExpectedError:  "error creating gateway: rpc error: code = Internal desc = could not find entities matching" + tests.Separator + "[type:\"cellular_enodeb\"" + tests.Separator + "key:\"dne\"]",
```
hcgatewood: Optional: tests.Separator -> tests.Sep since we're gonna be using it a lot

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
hcgatewood: Nit: instead of escaping the quotes, use the backtick

lte/cloud/go/services/subscriberdb/storage/ip_lookup_test.go
```diff
@@ -113,13 +113,20 @@ func TestIPLookup(t *testing.T) {
...
 		assert.NoError(t, err)
 		want = []*protos.IPMapping{
 			{Ip: "ipZ", Imsi: "IMSI0", Apn: "apn0"},
 		}
-		assert.Equal(t, want, got)
+		assertEqualArrays(t, want, got)
 	})
 }
+
+func assertEqualArrays(t *testing.T, expected []*protos.IPMapping, actual []*protos.IPMapping) {
```
hcgatewood: For my understanding: why is this necessary over assert.Equal?

lte/cloud/go/services/subscriberdb/streamer/providers.go
```diff
@@ -133,8 +133,8 @@ func subscriberToMconfig(ent configurator.NetworkEntity, apnConfigs map[string]*
 		AuthOpc:  cfg.Lte.AuthOpc,
 	}
 
-	if cfg.Lte.SubProfile != "" {
-		sub.SubProfile = string(cfg.Lte.SubProfile)
+	if cfg.Lte.SubProfile != nil && *cfg.Lte.SubProfile != "" {
```
hcgatewood, line 136: Yeah the fact that all these get introduced as ptrs now is def annoying... no input from Googling/searching through GH Issues?

dp/cloud/go/active_mode_controller/internal/message_generator/sas/grant_test.go
```diff
@@ -25,6 +25,7 @@ func TestGrantRequestGenerator(t *testing.T) {
 			LowFrequencyHz:  3620 * 1e6,
 			HighFrequencyHz: 3630 * 1e6,
 		}},
+		preferences: new(active_mode.FrequencyPreferences),
```
hcgatewood: Nit: prefer `&FreqPref{}` over `new(FreqPref)`

feg/gateway/policydb/serializers_test.go
```diff
@@ -42,11 +42,8 @@ func TestPolicyDeserializer(t *testing.T) {
 	deserializedPolicy, ok := iPolicy.(*lteProtos.PolicyRule)
 	assert.True(t, ok)
 
-	// clear out meta fields
-	clearOutMetaFieldsFromPolicy(policy)
-	clearOutMetaFieldsFromPolicy(deserializedPolicy)
-
-	assert.Equal(t, policy, deserializedPolicy)
+	equal := proto.Equal(policy, deserializedPolicy)
+	assert.True(t, equal)
```
hcgatewood: Nit: consider combining this pattern into single line

feg/gateway/services/csfb/servicers/test/csfb_test.go
```diff
@@ -295,7 +308,8 @@ func TestCsfbServer_MMEResetIndication(t *testing.T) {
 	client := protos.NewCSFBFedGWServiceClient(conn)
 	reply, err := client.MMEResetIndication(context.Background(), req)
 	assert.NoError(t, err)
-	assert.Equal(t, &orcprotos.Void{}, reply)
+	equal := proto.Equal(&orcprotos.Void{}, reply)
+	assert.True(t, equal)
```
hcgatewood: I'm assuming this pattern doesn't provide good error messages -- I believe there's a way to conditionally print the values that differ in the assert call?

feg/gateway/tools/s6a_cli/main.go
hcgatewood: It's crazy to me that this is necessary... does using `%+v` or `%v` resolve the need?

My assumption is fmt.Printf is copying the value of the passed ptr, which triggers the govet warning. Not sure why fmt.Printf is doing that copying though

Like, 1. I'm not sure why this even fixes the issue, because we're still printing a protobuf either way... and 2. like, there's gotta be a sane way to print protobufs, right? Hm

At end of the day, if we're stuck with it is fine 🙂 Just trying to understand

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
```diff
@@ -211,44 +220,45 @@ func (store *sqlConfiguratorStorage) getBuilder(networkID string, filter EntityL
...
+	if criteriaCopy.PageToken != "" {
 		where = append(where, sq.Gt{entCol(entKeyCol): pageToken.LastIncludedEntity})
 	}
 
 	b := builder.Where(where)
 	return addSuffix(b), nil
 }
 
-func scanEntityRow(rows *sql.Rows, criteria EntityLoadCriteria) (NetworkEntity, error) {
+func scanEntityRow(rows *sql.Rows, criteria *EntityLoadCriteria) (*NetworkEntity, error) {
```
hcgatewood: Maybe there's some master reason why protobuf maintainers did it this way, but man this seems suspect lol. Like, we can't even pass a proto by value anymore?

hcgatewood (approved):
Reviewed through "beginning of requested feg changes", per @maxhbr instructions. Overall lgtm

The painful points are definitely all related to protobuf version bump (ugh I cry), namely: string comparison and copying/cloning. Those are both sad turn of events, but looks like you handled it appropriately as best I can tell

As far as testing, since we're changing a lot of unit tests, once we have all the dev done we'll want to do the following manual testing before merging:

- Build and deploy a K8s Orc8r (can be local using Minikube)
- FeG integ tests (cc @uri200)

If both those are successful, I'd have enough confidence to merge this

Thanks for all your hard work! 💯💯

themarwhal (approved):

uri200 (changes requested):
Please allow some time to be reviewed again.

emakeev (changes requested):
any possibility of splitting this?

There are at least 3 major upgrades in this PR:
1. proto V2 upgrade
2. swagger upgrade
3. Go 1.18 upgrade

Nobody can reasonably review a 100K lines PR (even if major part of it is autogenerated). And it also obfuscates history & blame.

maxhbr:
@emakeev : that is an interesting suggestion in https://github.com/magma/magma/pull/12526, we will look into that

emakeev:
@maxhbr, when #12526 passes required tests we can merge & you can rebase swagger & Go 1.18 parts. We usually try to keep 1 commit -> 1 PR for simplicity.

lte/cloud/go/services/subscriberdb/storage/ip_lookup_test.go
alexzurbonsen: The root problem here is the new proto implementation of messages. Messages now contain other (unexported) meta fields whose values make the comparison fail.
The easiest way to avoid that, that occurred to us, is using proto.Equal(...) or make string comparisons with .String() for example (but the latter is not recommended by protobuf due to unstable representation).

Just for illustration:
```
ip_lookup_test.go:57: 
            	Error Trace:	ip_lookup_test.go:57
            	Error:      	Not equal: 
            	            	expected: []*protos.IPMapping{(*protos.IPMapping)(0xc0001151a0), (*protos.IPMapping)(0xc000115200)}
            	            	actual  : []*protos.IPMapping{(*protos.IPMapping)(0xc000114ea0), (*protos.IPMapping)(0xc000114f60)}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -9,3 +9,1337 @@
            	            	    },
            	            	-   atomicMessageInfo: (*impl.MessageInfo)(<nil>)
            	            	+   atomicMessageInfo: (*impl.MessageInfo)({
            	            	+    GoReflectType: (*reflect.rtype)({
```

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
alexzurbonsen: Separator won in team vote.  ¯\\\_(ツ)_/¯

feg/gateway/tools/s6a_cli/main.go
alexzurbonsen: Actually, you are right. It is not necessary as we learned by now. Pointers can still be passed with `%+v` for example. It is fixed now.

Sorry, that this still appeared in the review. We didn't squash the fix up commits yet. Will do!

.github/workflows/feg-workflow.yml
ajahl: Make sense, is pinned  to `1.8.0` now.

orc8r/cloud/go/services/configurator/storage/sql_entity_load_helpers.go
alexzurbonsen: Yes, somewhat mean... It fails the `go vet -copylocks` check, since in API v2 implementation proto messages contain a `sync.Mutex`.

This comment in a related issue https://github.com/golang/protobuf/issues/1225#issuecomment-713699536 might clarify to some degree.

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
voisey: Fixed

lte/cloud/go/services/lte/obsidian/handlers/handlers.go
ajahl: Yes, the address operator `&` requires an operand that is addressable which is not given inline. It would result in `invalid composite literal type`.

feg/gateway/services/csfb/servicers/test/csfb_test.go
alexzurbonsen: That is right. We optimized a little and introduced this function:

https://github.com/alexzurbonsen/magma/blob/d3ac7bfb8dc6e4857a98888bfbe3ea0401e57092/orc8r/cloud/go/test_utils/unit_test_proto_utils.go#L52-L67

Doesn't show preview, I guess since it is on a fork. Let me copy it here:
```
// AssertMessagesEqual compares two proto.Message's with proto.Equal.
// Prints string representations of messages upon inequality.
// Note, that string representations provided by protobuf are not stable.
// Please only pay attention to the field values themselves, not to whitespaces
// in between.
func AssertMessagesEqual(t *testing.T, expected, actual proto.Message) bool {
	equal := proto.Equal(expected, actual)
	if !equal {
		assert.Fail(t, fmt.Sprintf("Not equal: \n"+
			"expected: %s\n"+
			"actual: %s\n"+
			"Note, that string representations provided by protobuf are not guaranteed to be stable and the comparison between messages does not rely on them. Please pay attention to differences in the field values only.",
			expected.String(), actual.String()))
	}
	return true
}
```

The idea is to compare messages with proto.Equal but use string representations for error output. The resulting message might be misleading due to unstable string representations by protobuf. Thus the "Note,..." in the error message.

feg/gateway/policydb/serializers_test.go
alexzurbonsen: Will use the function mentioned in [comment](https://github.com/magma/magma/pull/12151#discussion_r856188709) below instead if that is fine with you.

lte/cloud/go/services/subscriberdb/streamer/providers.go
voisey: Good question. We're currently not sure exactly what lead to this happening, so it requires some more digging. See also https://github.com/magma/magma/pull/12151#discussion_r850172178

alexzurbonsen:
@emakeev Thank you for testing it! We want to incorporate open review comments here first to not invalidate them. Once we are through we will take a look at the split.

emakeev:
> @emakeev Thank you for testing it! We want to incorporate open review comments here first to not invalidate them. Once we are through we will take a look at the split.

@alexzurbonsen #12526 is finished & tested, all unit tests & lints are OK. it incorporates most of your proto changes, please feel free to take over or land. You may get better quality reviews & comments on a smaller, more targeted PRs.

alexzurbonsen:
> @alexzurbonsen #12526 is finished & tested, all unit tests & lints are OK. it incorporates most of your proto changes, please feel free to take over or land. You may get better quality reviews & comments on a smaller, more targeted PRs.

@emakeev Please find the protobuf upgrade in #12546. Had to pin one module that required Go 1.17 to a particular version. Other than that it contains relevant changes from this PR, with some minor changes. (Hopefully makes rebasing as easy as possible).

alexzurbonsen:
@emakeev @uri200 Rebased onto the merged protobuf upgrade #12546 and adapted the PR description. Can you take another look?

fyi: the pinned version of `x/net` is removed in this PR.

cwf/gateway/go.mod
```diff
@@ -11,7 +11,6 @@
 //
 module magma/cwf/gateway
 
-// TODO remove golang.org/x/net line once Go Upgrade (https://github.com/magma/magma/pull/12151) is merged
```
voisey, line 14: Shouldn't the replacement of x/net be removed here @alexzurbonsen? To me it looks like you've removed the comment but not the relevant line.

cwf/k8s/cwf_operator/go.mod
```diff
@@ -51,10 +45,166 @@ require (
...
+
 // Pinned to kubernetes-1.16.2
-// TODO remove golang.org/x/net line once Go Upgrade (https://github.com/magma/magma/pull/12151) is merged
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+
+	github.com/go-openapi/analysis => github.com/go-openapi/analysis v0.19.12
+	github.com/go-openapi/loads => github.com/go-openapi/loads v0.19.7
+	github.com/go-openapi/spec => github.com/go-openapi/spec v0.19.8
 	golang.org/x/net => golang.org/x/net v0.0.0-20210520170846-37e1c6afe023
```
voisey: And here?

dp/cloud/go/go.mod
```diff
@@ -11,7 +11,7 @@
 //
 module magma/dp/cloud/go
 
-go 1.13
+go 1.18
 
 replace (
 	golang.org/x/net => golang.org/x/net v0.0.0-20210520170846-37e1c6afe023
```
voisey: And here?

cwf/gateway/go.mod
alexzurbonsen: Good catch.

orc8r/cloud/go/tools.go
```diff
@@ -18,11 +18,13 @@ package main
 
 // Put all binary tool dependencies in here so they can be tracked by the go
 // module.
-
+// dependencies listed here should also be listed in TOOL_DEPS in orc8r/cloud/go/Makefile
```
maxhbr: absolute paths might not be stable. Maybe better write `./Makefile`?

uri200:
Hey @alexzurbonsen  you may need to fix conflicts

image

alexzurbonsen:
Done :)

orc8r/cloud/go/services/orchestrator/obsidian/handlers/network_handlers_test.go
```diff
@@ -880,7 +880,7 @@ func Test_PutNetworkDNSHandlers(t *testing.T) {
 		ExpectedStatus: 400,
 		ExpectedError: "validation failure list:\n" +
 			"validation failure list:\n" +
-			"a_record.0 in body must be of type ipv4: \"192-88-99-142\"",
+			"records.0.a_record.0 in body must be of type ipv4: \"192-88-99-142\"",
```
Neudrino, line 883: Should that be `dns.records.0.` as [above in line 319](https://github.com/magma/magma/pull/12151/files#diff-1e3584f3517597a3aa63e19a1e9c69b3f04fd065b144d6d61543353e8330dd25R319)

Neudrino (approved):

orc8r/cloud/go/services/orchestrator/obsidian/handlers/network_handlers_test.go
alexzurbonsen: Checked again an the other tests indeed call a different URL. This one calls dns.

voisey:
Hi @magma/approvers-domain-proxy, could we please get your input on here when you have the time? :slightly_smiling_face:

jkmar (approved):

alexzurbonsen:
Thanks for the review @jkmar.

Got notified that you left a comment on enodebd_dp.proto, saying that no go code generation is required for that file. Can't locate the comment, may be since the change is outdated and was moved to the protobuf update PR recently. So let me just write here that we had to include the `option go_package ...` due to an error in the code generation process we got otherwise. (Probabaly new in the updated protobuf version.)

alexzurbonsen:
We deployed orc8r on minikube, as advised by @hcgatewood. Worked out fine (for example we tested whether some swagger calls work out as expected).

@emakeev @uri200 is there any further feedback from your side?

uri200:
> We deployed orc8r on minikube, as advised by @hcgatewood. Worked out fine (for example we tested whether some swagger calls work out as expected).
> 
> @emakeev @uri200 is there any further feedback from your side?

Hey @alexzurbonsen  in my opinion the risk is at connectivity between AGW - Orc8r - FEG. 
This can be initially tested with fedrated s1ap test. But that should be test it in a lab. 


Can you share the results of s1ap federated test?

sebathomas:
@uri200 Wasn't the lab recently closed? We ran the s1ap federated integration tests a few weeks ago on a laptop and they were successful, see: https://github.com/magma/magma/pull/12151#issuecomment-1099200148

uri200:
@sebathomas note that not having run tests after changes may pose a big risk with changes that modify endpoints. Many of those services run across gateways and currently there is no test coverage.

In my opinion, a change like this should be acompanied of a strong and repeatable testing strategy. So running S1ap integntest should be done after any change.for precaution.

Maybe it would be a good idea to consider to add S1ap federated integ test to CI. This way changes like this one would be les risky.

See that we already have LTE intef test running there.

sebathomas:
@uri200 I completely agree, we were planning to run all integration tests again before the merge in any case. The question is if there are any other review comments.

Neudrino:
> Maybe it would be a good idea to consider to add S1ap federated integ test to CI. This way changes like this one would be les risky.

This might in general be a good idea, however, I see it clearly out of scope of this PR.

Afaiks all the required and additionally requested tests were successfully run locally and such that the upgrade was verified to be good with the best methods currently available. Thus, to me the PR looks very much **ready to be merged**, as all changes were implemented and all requested checks and test available were successfully executed.

In my opinion this PR is a really great effort and benefit for the project and enables a lot of further positive development especially around orc8r.
@emakeev @uri200, could you re-review the PR please and explicitly state which changes you would still see required to be able to merge this PR, please.

uri200:
Hey @Neudrino , as mentioned in early message we need to make sure that S1ap Federated integ test is passing with the new modifications. That test is the only existing way to test interaction between gateways.

ajahl:
> > We deployed orc8r on minikube, as advised by @hcgatewood. Worked out fine (for example we tested whether some swagger calls work out as expected).
> > @emakeev @uri200 is there any further feedback from your side?
> 
> Hey @alexzurbonsen in my opinion the risk is at connectivity between AGW - Orc8r - FEG. This can be initially tested with fedrated s1ap test. But that should be test it in a lab.
> 
> Can you share the results of s1ap federated test?

@uri200 Here are the results of the S1AP federated test:

```
. /home/vagrant/build/python/bin/activate
echo "Running test: federated_tests/s1aptests/test_attach_detach.py"
Running test: federated_tests/s1aptests/test_attach_detach.py
timeout --foreground -k 930s 900s sudo -E PATH=/home/vagrant/build/python/bin:/home/vagrant/build/python/bin:/usr/lib/ccache:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin PYTHONPATH=:/home/vagrant/s1ap-tester/bin /home/vagrant/build/python/bin/nosetests --with-flaky --force-flaky --no-success-flaky-report --max-runs=3 --min-passes=1 --with-xunit --xunit-file=/var/tmp/test_results/test_attach_detach.xml -x -s federated_tests/s1aptests/test_attach_detach.py || (echo "fail" > /home/vagrant/magma/test_status.txt && exit 1)
Start time 12:46:07
tag: AUTH_TYPE tagVal: MILENAGE
************************* Enb tester config

**********************TPT open server******** 0

 Server Socket FD =[13]
************* Testing *Time elapsed(in usec): for timer expiry               :3672 usec
AGW is stateless
APN Correction configured
Health service is disabled
Using subscriber IMSI IMSI001010000000001
Using IMEI 3805468432113171
Using subscriber IMSI IMSI001010000000002
Using IMEI 3805468432113172
************************* UE device config for ue_id  1
************************* UE device config for ue_id  2
************************* Configuring IP block
************************* Waiting for IP changes to propagate
************************* S1 setup
************************* UE App config
************************* Running End to End attach for  UE id  1
ue id 1 found
PDN_TYPE: 1
IPv4 PDN_ADDRESS:192 168 128 12

*************NbUiNbuHdlUeRadioCapMsg: Send UE cap msg SUCCESS*********
************************* Running UE detach for UE id  1
Deleting Ue Context from Traffic Handler
************************* Running End to End attach for  UE id  2
ue id 2 found

*************NbUiNbuHdlUeRadioCapMsg: Send UE cap msg SUCCESS*********
PDN_TYPE: 1
IPv4 PDN_ADDRESS:192 168 128 13
************************* Running UE detach for UE id  2
Deleting Ue Context from Traffic Handler
************************* send SCTP SHUTDOWN
Keys left in Redis (list should be empty)[

]
Entries in s1ap_imsi_map (should be zero): 0
Entries left in hashtables (should be zero): 0
Entries in mme_ueip_imsi_map (should be zero): 0
.
----------------------------------------------------------------------
Ran 1 test in 10.359s

OK
sleep 1
Connection to 127.0.0.1 closed.

Done.
Disconnecting from vagrant@127.0.0.1:2200... done.
```

uri200:
you may need to rebase

uri200 (approved):

maxhbr (approved):
LGTM

orc8r/cloud/go/tools.go
alexzurbonsen: Changed this path and the path in the respective Makefile (plus analogous changes made in feg/gateway).

voisey:
Hi @emakeev, do you have any further comments on this, or would you be happy for us to proceed with getting it merged?

alexzurbonsen:
The cwag precommit tests passed again.

```
vagrant@cwag-dev:~$ make -C ${MAGMA_ROOT}/cwf/gateway precommit
make: Entering directory '/home/vagrant/magma/cwf/gateway'
go fmt ./...
integ_tests/gy_enforcement_test.go
integ_tests/gy_reauth_test.go
integ_tests/test_runner.go
services/uesim/client_api.go
services/uesim/servicers/uesim.go
go test ./services/...
?       magma/cwf/gateway/services/gateway_health/events        [no test files]
?       magma/cwf/gateway/services/gateway_health/gateway_health        [no test files]
?       magma/cwf/gateway/services/gateway_health/health/gre_probe      [no test files]
?       magma/cwf/gateway/services/gateway_health/health/service_health [no test files]
?       magma/cwf/gateway/services/gateway_health/health/system_health  [no test files]
?       magma/cwf/gateway/services/gateway_health/metrics       [no test files]
ok      magma/cwf/gateway/services/gateway_health/servicers     0.256s
ok      magma/cwf/gateway/services/uesim        0.065s
ok      magma/cwf/gateway/services/uesim/servicers      0.122s
?       magma/cwf/gateway/services/uesim/test_init      [no test files]
?       magma/cwf/gateway/services/uesim/uesim  [no test files]
go vet ./...
make: Leaving directory '/home/vagrant/magma/cwf/gateway'
```

```
vagrant@cwag-dev:~$ make -C ${MAGMA_ROOT}/cwf/gateway/integ_tests precommit
make: Entering directory '/home/vagrant/magma/cwf/gateway/integ_tests'
go fmt ./...
go test --tags=all -c ./...
make: Leaving directory '/home/vagrant/magma/cwf/gateway/integ_tests'
vagrant@cwag-dev:~$ 
```

alexzurbonsen:
The cwag-integ test also passed again:

```
...
[vagrant@127.0.0.1:2201] out: DONE 40 tests, 11 skipped in 634.018s
[vagrant@127.0.0.1:2201] out: 

[vagrant@127.0.0.1:2201] Executing task '_clean_up'
[vagrant@127.0.0.1:2201] run: pkill go
[localhost] local: pwd
[localhost] local: vagrant status magma_trfserver
[localhost] local: vagrant ssh-config magma_trfserver
[vagrant@127.0.0.1:2200] run: pkill iperf3 > /dev/null &
Integration Test Passed for "all"!
Disconnecting from vagrant@127.0.0.1:2222... done.
Disconnecting from vagrant@127.0.0.1:2200... done.
Disconnecting from vagrant@127.0.0.1:2201... done.
```

alexzurbonsen:
So all tests mentioned in the Test plan section of the PR description were executed again and passed. From this side we are good to go. :)

uri200:
Note cwf does not have any significative test, because it has minimal code.

The tests that are going to provide more signal are:

- AGW, oc8r and FEG unit test
- S1ap integ test to test logic on the AGW, specially MME.
- CWF integ test to test enforcement ( interaction between sessiond, pipelined on agw, and session proxy in feg)
- S1ap federated integ test which is an initial framework to do end to end test with all the gateways in place (agw, or8cr, feg)

emakeev:
> Hi @emakeev, do you have any further comments on this, or would you be happy for us to proceed with getting it merged?

It LGTM. Thanks!

emakeev (approved):

uri200:
Once you fix the merge conflicts, we will be able to merge it.

ajahl (approved):

voisey:
Note that I've just merged the changes from https://github.com/magma/magma/pull/12422 into here so that Go is also updated in the VMs when we merge this.

PR #12178, opened 2022-03-17
chore(mme): Preparation work to split mme_procedure_test into multiple files

themarwhal:

## Summary

Pull out test fixture used in mme_procedure_test into a separate file. This allows the fixture to be shared across multiple test files. (Wider effort is tracked at https://github.com/magma/magma/issues/12181, https://github.com/magma/magma/issues/12182)

## Test Plan
CI

## Additional Information


lionelgo (approved):

lte/gateway/c/core/oai/test/mme_app_task/BUILD.bazel
```diff
@@ -105,6 +105,16 @@ cc_test(
     ],
 )
 
+cc_library(
+    name = "mme_procedure_test_fixture",
+    srcs = ["mme_procedure_test_fixture.h"],
+    deps = [
+        ":mme_app_test_core",
+        "//lte/gateway/c/core/oai/test/mock_tasks",
+        "@com_google_googletest//:gtest_main",
```
electronjoe: Question: Is gtest actually used in this header? I don't see it imported so I wonder if gtest is needed for this label?.

Nit: If gtest is needed - we do not need the main-auto-wrapped import I think (as this is not a test needing a main() function). So we could change this line to `"@com_google_googletest//:gtest",`.

electronjoe (approved):

lte/gateway/c/core/oai/test/mme_app_task/BUILD.bazel
```diff
@@ -105,6 +105,16 @@ cc_test(
     ],
 )
 
+cc_library(
+    name = "mme_procedure_test_fixture",
+    srcs = ["mme_procedure_test_fixture.h"],
```
nstng: nit: hdrs

nstng (approved):
:+1:

lte/gateway/c/core/oai/test/mme_app_task/BUILD.bazel
themarwhal: good catch :D

lte/gateway/c/core/oai/test/mme_app_task/BUILD.bazel
themarwhal: I think it is needed to pull in `::testing::Test`, but will double check. But I did not realize there was a distinction between gtest and gtest_main :D Will double check to see what is needed. Thanks!

nstng:
btw native, asan and lsan passed locally for me (did not check for flakyness). was this test fixed by accident? :)

themarwhal:
> btw native, asan and lsan passed locally for me (did not check for flakyness). was this test fixed by accident? :)

still flaky :/ I recently saw a failure :'(

lte/gateway/c/core/oai/test/mme_app_task/BUILD.bazel
electronjoe: If it's needed for `::testing::Test` - and if I'm right - can we add an `#include <gtest>`

Yeah gtest has a couple targets and I *think* it's reasonable to only use gtest if you aren't authoring any actual tests. But yeah confirm for me =D And I do get that others might be confused by this in the future...

themarwhal: #include <gtest/gtest.h> is already included. Testing with "@com_google_googletest//:gtest" seems to work.

electronjoe: > #include <gtest/gtest.h> is already included

Aah ok, I should have ctr-f'd.

themarwhal: Yeah I think gtest is needed, testing this is a little bit of an effort because of how C++ gobbles includes :P but I *think* it is needed

PR #11245, opened 2022-01-21
fix(agw): Modified code to handle TAU Req for connected UEs

rsarwad:
fix(agw): Modified code to handle TAU Req for connected UEs
## Test Plan
Executed s1ap sanity test suite, for which test_multi_enb_complete_reset.py was failing because as part of this TC; it first connects 5 eNBs and for each eNB there shall be one UE connected. 
When complete reset is initiated for one eNB; but all UE contexts were getting deleted; Ideally we need to delete the contexts only for the UE for which eNB Reset is triggered. The changes are available in PR, https://github.com/facebookexperimental/S1APTester/pull/70

PR #12188, opened 2022-03-17
fix(mme): Update handle_ue_context_rel_timer_expiry to return gracefully

ardzoht:

## Summary

-  Changes on https://github.com/magma/magma/pull/11245 creates an issue discovered while running `test_attach_auth_failure.py` due to the timer handler returning `RETURNerror` (-1) which triggers an exit on the S1AP's ZMQ thread. 
- This PR fixes the issue by gracefully handling the return of the handler by returning `RETURNok` (0) instead.
- Updates all potential zmq's AssertFatal messages to indicate possible source of crash.

## Test Plan

- `make precommit` runs successfully 
- `make integ_test`

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/s1ap/s1ap_mme_handlers.c
+++ b/lte/gateway/c/core/oai/tasks/s1ap/s1ap_mme_handlers.c
@@ -5309,14 +5309,16 @@ static int handle_ue_context_rel_timer_expiry(zloop_t* loop, int timer_id,
   if (!s1ap_pop_timer_arg_ue_id(timer_id, &mme_ue_s1ap_id)) {
     OAILOG_WARNING(LOG_S1AP, "Invalid Timer Id expiration, Timer Id: %u\n",
                    timer_id);
-    OAILOG_FUNC_RETURN(LOG_S1AP, RETURNerror);
+    // Timer handlers need to return 0 to avoid triggering ZMQ thread exit
+    OAILOG_FUNC_RETURN(LOG_S1AP, RETURNok);
   }
   if ((ue_ref_p = s1ap_state_get_ue_mmeid(mme_ue_s1ap_id)) == NULL) {
     OAILOG_ERROR(
         LOG_S1AP,
         "Failed to find UE context for mme_ue_s1ap_id " MME_UE_S1AP_ID_FMT,
         mme_ue_s1ap_id);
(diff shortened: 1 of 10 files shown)


ssanadhya (approved):
lgtm

ssanadhya:
@ardzoht , were you able to confirm that we have to keep the non-void return type on the timer handler functions? If not, it will be good to have a void return type for all timer expiry handlers.

ardzoht:
> @ardzoht , were you able to confirm that we have to keep the non-void return type on the timer handler functions? If not, it will be good to have a void return type for all timer expiry handlers.

@ssanadhya For the purpose of unblocking CI, I have kept the signature of the functions to the same. I have created #12189 to track this investigation which I plan to do next week.

ssanadhya:
> > @ardzoht , were you able to confirm that we have to keep the non-void return type on the timer handler functions? If not, it will be good to have a void return type for all timer expiry handlers.
> 
> @ssanadhya For the purpose of unblocking CI, I have kept the signature of the functions to the same. I have created #12189 to track this investigation which I plan to do next week.

Sounds good. Thanks for creating the issue!

lte/gateway/c/core/oai/tasks/s1ap/s1ap_mme_handlers.c
```diff
@@ -5309,14 +5309,16 @@ static int handle_ue_context_rel_timer_expiry(zloop_t* loop, int timer_id,
...
+    OAILOG_FUNC_RETURN(LOG_S1AP, RETURNok);
   }
   if ((ue_ref_p = s1ap_state_get_ue_mmeid(mme_ue_s1ap_id)) == NULL) {
     OAILOG_ERROR(
         LOG_S1AP,
         "Failed to find UE context for mme_ue_s1ap_id " MME_UE_S1AP_ID_FMT,
         mme_ue_s1ap_id);
-    OAILOG_FUNC_RETURN(LOG_S1AP, RETURNerror);
+    // Timer handlers need to return 0 to avoid triggering ZMQ thread exit
+    OAILOG_FUNC_RETURN(LOG_S1AP, RETURNok);
```
rsarwad, line 5321: For this return we need to return with RETURNok
https://github.com/magma/magma/pull/12188/files#diff-cacc260c2f044ba3a92919fafccb6e8c903d33b8900a88ddd94758793792cc3bR5336

Issue #12067, opened 2022-03-10
[FeG] Gx/Gy interfaces over SCTP with connections suddenly closed, then no interface/service auto-recovered


crbertoldo:
### Your Environment

- **Version:** 1.5
- **Affected Component:** FeG
- **Affected Subcomponent:** session_proxy, FeG EC2 instance itself, and/or its underlying network
- **Deployment Environment:** e.g. AWS (orc8r), bare-metal (AGW)
- **Extra:** AWS Site-to-Site VPN between the FeG's VPC (us-west-2) and Cyan's OCS/PCRF (both on the same instance in Miami)

### Describe the Issue

We have two issues that cause the described by the title of this issue:

1. Unexpected SCTP SHUTDOWNs for Gx and/or Gy that **sometimes** is not proceeded by the interface recovery automatically or FeG/session_proxy takes a while to recover it (observations for an interval between a few seconds to an hour to do so). We see these SHUTDOWNs happening right after a packet drop for Diameter DWR towards the PCRF/OCS. But, not every time we have a drop, SHUTDOWNs happen. When there's no auto-recovery, we can see at `/proc/net/sctp/assocs` the missing association(s). Restarting the FeG containers recover the interface(s).
2. Unexpected SCTP ABORTs also for Gx and/or Gy that is **never** proceeded by the interface recovery automatically. When it happens neither restarting the FeG containers recover the interface(s). Just rebooting the FeG VM (and sometimes at least 2 or 3 times consecutively) recovers the interface(s).  Different from _1_ above, the sctp/assoc(s) keep active for the affected interface(s) while its Rx queue keeps increasing (looks like it cannot send data to the application layer, so keeps buffering). Example for Gy:
```
 ASSOC     SOCK   STY SST ST HBKT ASSOC-ID TX_QUEUE RX_QUEUE UID INODE LPORT RPORT LADDRS <-> RADDRS HBINT INS OUTS MAXRT T1X T2X RTXC wmema wmemq sndbuf rcvbuf
       0        0 2   1   3  13      2        0        0       0 22809 43447  3969  10.10.0.137 172.17.0.1 <-> *10.218.10.91 	    7500    10    10   10    0    0        1        1        0   212992   212992
       0        0 2   1   3  97      3        0    54656       0 27358 60552  3869  10.10.0.137 172.17.0.1 <-> *10.218.10.91 	    7500    10    10   10    0    0        0        1        0   212992   212992
```

**To Reproduce**
[...]

PR #12192, opened 2022-03-17
fix(feg): Diameter closed connection recovery

emakeev:

## Summary

Add connection recovery logic for externally closed diameter connection.
Current connection manager only recovers errored connections. There several scenarios when connennection can be closed 'quietly' without connection manager knowing about it - when diameter watchdog is timed out, it'll close connection unconditionally, protocol stack may closed unhealthy or idle connection, etc.

The fix is relying on diameter Close Notifier channel to detect connection closure & reconnect.

The fix can potentially resolve issue #12067

## Test Plan

unit tests, TVM

## Additional Information

### DIFF
--- a/feg/gateway/diameter/connection.go
+++ b/feg/gateway/diameter/connection.go
@@ -18,6 +18,7 @@ import (
 	"net"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/fiorix/go-diameter/v4/diam"
 	"github.com/fiorix/go-diameter/v4/diam/avp"
@@ -33,6 +34,9 @@ type messageTypeEnum uint8
 const (
 	requestMessage messageTypeEnum = 1
 	answerMessage  messageTypeEnum = 2
+
+	connectionRecoveryInterval = time.Second
+	connectionRecoveryattempts = 6
 )
 
 // Connection is representing a diameter connection that you can
@@ -42,8 +46,11 @@ type Connection struct {
 	metadata *smpeer.Metadata
 	server   *DiameterServerConfig
 	client   *sm.Client
+	disabled bool
 	mutex    sync.Mutex
 }
+
+var disabledErr = errors.New("connection disabled")
 
 func newConnection(client *sm.Client, server *DiameterServerConfig) *Connection {
 	conn := &Connection{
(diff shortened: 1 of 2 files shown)


feg/gateway/diameter/connection.go
```diff
@@ -155,16 +155,42 @@ func (c *Connection) getDiamConnection() (diam.Conn, *smpeer.Metadata, error) {
...
+			}
+			if conn.LocalAddr() != nil {
+				remoteAddress = conn.RemoteAddr().String()
+			}
+		}
+		glog.Infof("notified of %s->%s connection closure", localAddress, remoteAddress)
+	}
+	if c != nil && conn != nil && c.destroyConnection(conn) {
+		// if connection was closed not by connection management functions, recover it
+		c.getDiamConnection()
```
uri200: is it possible this fails again and then the connection is just lost?

Meaning if for some reason there is a network outage and the server is not reacheable  `getDiamConnection`  will fail and the connection will never be created. We may end up in the same place.

Maybe we need some kind of infinite loop to keep retrying every n seconds?

```	
if c != nil && conn != nil && c.destroyConnection(conn) {
		for {
		// if connection was closed not by connection management functions, recover it
			_, _, err := c.getDiamConnection()
			if err != nil {
				return
			}
			glog.Error("Failed to recover closed connection. Retrying")
			time.sleep(1time.Second)
		}
		```


Or maybe this loop could be at `getDiamConnection`

Comment (emakeev) [reply feg/gateway/diameter/connection.go]: I think, if there is a prolonged network outage, we'll just fall back to active recovery - when the client tries to send out a request, getDiamConnection is called again and will attempt to reconnect. Putting infinite loop here could be a bit heavy, maybe a limited number of retries.

Comment (uri200) [reply feg/gateway/diameter/connection.go]: Note `newConnection` suffers of this issue. It never tries to reconnect if it fails once. 
Maybe we could add a loop in that go routine and just call `newConnection` in this line  instead of `c.getDiamConnection()`

http://github.com/magma/magma/blob/master/feg/gateway/diameter/connection.go#L48-L48

**Review by uri200 — APPROVED** [review]

File: feg/gateway/diameter/connection.go [hunk 2]
```diff
@@ -155,16 +159,50 @@ func (c *Connection) getDiamConnection() (diam.Conn, *smpeer.Metadata, error) {
...
+	if c != nil && conn != nil && c.destroyConnection(conn) {
+		// if connection was closed not by connection management functions, recover it
+		retryWaitTime := connectionRecoveryInterval
+		for i := 0; i < connectionRecoveryattempts; i++ {
+			_, _, err := c.getDiamConnection()
+			if err == nil {
+				break
+			}
+			time.Sleep(retryWaitTime)
+			retryWaitTime *= 2
```
uri200, line 199: now i am just concerned about HA functionality :D

emakeev: HE disable will set connection to disabled...

PR #12196, opened 2022-03-18
chore: all python tests are (mandatory) categorized by size

### LINKED ISSUE #11743: Bazel technical debt

nstng:

## Summary

Technical debt task from https://github.com/magma/magma/issues/11743.

Bazel uses `size = "medium"` as default, i.e., a test times out after five minutes. See [docu](https://bazel.build/reference/be/common-definitions). This should be set to more granular values so that regression of runtime is tested as well.

Notes:
* I made the `size` property mandatory - this is possible because we are using a wrapper rule (`pytest_test`)  for `py_test`. I think this is reasonable, but I can remove this restriction if there are valid arguments.
* From the current test set the majority of tests can be set to "small" (one minute time out). Exceptions with "medium" are following.
* `processor_tests` "medium" - had locally actually 2/10 runs >1 minute
* `onready_test`, `store_tests`, `sync_rpc_client_tests` had an average of >25 seconds. Here I just want to make sure that there are no time outs on less powerful setups.

## Additional Information

### DIFF
--- a/bazel/python_test.bzl
+++ b/bazel/python_test.bzl
@@ -92,7 +92,7 @@ _pytest_runner = rule(
     },
 )
 
-def pytest_test(name, srcs, deps = [], args = [], data = [], imports = [], python_version = None, **kwargs):
+def pytest_test(name, srcs, size, deps = [], args = [], data = [], imports = [], python_version = None, **kwargs):
     runner_target = "%s-runner.py" % name
 
     _pytest_runner(
@@ -113,5 +113,6 @@ def pytest_test(name, srcs, deps = [], args = [], data = [], imports = [], pytho
         main = runner_target,
         legacy_create_init = False,
         imports = imports + ["."],
+        size = size,
         **kwargs
     )
--- a/lte/gateway/python/magma/enodebd/tests/BUILD.bazel
+++ b/lte/gateway/python/magma/enodebd/tests/BUILD.bazel
@@ -20,6 +20,7 @@ LTE_ROOT = "{}lte/gateway/python".format(MAGMA_ROOT)
 
 pytest_test(
     name = "baicells_old_tests",
+    size = "small",
     srcs = ["baicells_old_tests.py"],
     imports = [
         LTE_ROOT,
@@ -35,6 +36,7 @@ pytest_test(
 
 pytest_test(
     name = "baicells_qafb_tests",
+    size = "small",
     srcs = ["baicells_qafb_tests.py"],
     imports = [
         LTE_ROOT,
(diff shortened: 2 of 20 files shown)


LKreutzer (approved):

jheidbrink:
Do you have an idea why the `medium` tests take longer? I just looked at https://docs.bazel.build/versions/main/be/common-definitions.html and found the following classification:
> Unit tests are considered "small", integration tests "medium", and end-to-end tests "large" or "enormous".

jheidbrink (approved):

vktng (approved):

nstng:
> Do you have an idea why the `medium` tests take longer? I just looked at https://docs.bazel.build/versions/main/be/common-definitions.html and found the following classification:
> 
> > Unit tests are considered "small", integration tests "medium", and end-to-end tests "large" or "enormous".

No, I did not check the logic. But imho it fits to the description. One test is in 80% of the runs <1 minute. The other medium tests locally in all runs. I just used medium in order to avoid flaky behavior.

jheidbrink:
btw we can query for medium-sized python tests with `bazel query 'attr(size, medium, kind(py_test, //...))'`

themarwhal (approved):
Thanks for the cleanup!

PR #12211, opened 2022-03-21
feat(dp): Add channel preference logic to AMC

jkmar:
## Summary

The algorithm is following:
Grant attempts are ordered first by frequency and then by bandwidth.
If all preferred frequencies are exhausted, random ones are chosen.
After trying all frequencies and one round of random ones
(one for each bandwidth) spectrum inquiry is sent.

There are also some API changes in active mode proto
regarding new preferences field and adjustments in channels.


xbend (approved):
python part LGTM

WojSad (approved):

Issue #12210, opened 2022-03-21
Come up with a scalable solution for S3 cache population


themarwhal:
We want to populate cache in S3 (or anywhere) so that they can be easily pulled in by devs and GitHub Codespaces

Some commands we want to cache

- bazel test //...
- bazel build //...
- bazel test //orc8r/gateway/c/... //lte/gateway/c/... --config=asan

themarwhal:
@LKreutzer Also this too. if you already have a master task for cache related things, please close this out and point it to the other task :D

LKreutzer:
> @LKreutzer Also this too. if you already have a master task for cache related things, please close this out and point it to the other task :D

Thanks for pointing this issue out to me! This should be covered by the work we are planning to do for https://github.com/magma/magma/issues/12458 - but I will keep it open for now. Currently the plan is that the s3 bazel-remote cache will be publicly read-only accessible, which could include the application in the codespaces and elsewhere in the future.

LKreutzer:
These commands are already covered by the workflows and the remote cache that is currently set up.

PR #12212, opened 2022-03-21
chore: temporarily reduce workload on Bazel cache population job

themarwhal:

## Summary
The bazel cache push job for devcontainer has been failing for the past few days due to disk issues :( We might need a different approach for this. But for now, I'm modifying the job so that we only populate cache for the plain build and test. With our recent change in the ASAN/lSAN copt config, running with --config=asan will only trigger a rebuild in the Magma C/C++ targets, which is not great but not awful.

Issue is tracked here; https://github.com/magma/magma/issues/12210

## Test Plan
master

## Additional Information

### DIFF
--- a/.github/workflows/bazel-cache-push.yml
+++ b/.github/workflows/bazel-cache-push.yml
@@ -112,15 +112,7 @@ jobs:
           run: |
             cd /workspaces/magma
             bazel test //...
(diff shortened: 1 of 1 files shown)


nstng (approved):
Thank you for fixing this - will move discussion to slack, on how to handle this.

PR #12215, opened 2022-03-21
fix(pipelined): PROXY_TAG_REG value is changed to reg12 so that overlaps are prevented

### LINKED ISSUE #12073: datapath: fix ovs register assigments.
reg10 is used for two separate use-cases, those registers are not use simultaneously to its not issue for now. but its better to keep it separate.
ref: https://github.com/magma/magma/blob/988b8fa15590fa5e0afddce3bd799932d8bddfc7/lte/gateway/python/magma/pipelined/openflow/registers.py#L22

verification:
1. Run unit tests
2. run integ tests.

nstng:
## Summary

See https://github.com/magma/magma/issues/12073.

Reviewer: Please have an additional look at `lte/gateway/python/magma/pipelined/policy_converters.py` if the change is reasonable.

## Test Plan

CI
* (sudo) unit tests locally green
* integration tests: https://github.com/nstng/magma/runs/5627342283?check_suite_focus=true
  * note: run on forked repo, i.e., upload failed, but test run successful 

## Additional Information


pshelar (approved):
LGTM

PR #12220, opened 2022-03-22
fix(feg): FeG diameter handler blocking fix and additional logging for diameter & session_proxy

emakeev:

## Summary

Additional logging for diameter, session_proxy also a fix for long standing [bug](https://github.com/magma/magma/blame/v1.6.1/feg/gateway/diameter/diameter_client.go#L302-L303) in FeG diameter client which could block response handle indefinitely on mismatch/lost diameter response (answer).

The fix should resolve issue https://github.com/magma/magma/issues/12067

## Test Plan

unit tests

## Additional Information

### DIFF
--- a/feg/gateway/services/session_proxy/credit_control/gy/gy_client.go
+++ b/feg/gateway/services/session_proxy/credit_control/gy/gy_client.go
@@ -144,7 +144,7 @@ func (gyClient *GyClient) SendCreditControlRequest(
 		return err
 	}
 
-	glog.V(2).Infof("Sending Gy CCR message:\n%s\n", message)
+	glog.V(3).Infof("Sending Gy CCR message:\n%s\n", message)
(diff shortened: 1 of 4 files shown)


feg/gateway/services/session_proxy/credit_control/gy/gy_client.go
```diff
@@ -466,7 +466,7 @@ func getReceivedCredits(cca *CCADiameterMessage) []*ReceivedCredits {
 // Output: diam.HandlerFunc
 func getCCAHandler() diameter.AnswerHandler {
 	return func(message *diam.Message) diameter.KeyAndAnswer {
-		glog.V(2).Infof("Received Gy CCA message:\n%s\n", message)
+		glog.V(3).Infof("Received Gy CCA message:\n%s\n", message)
```
uri200, line 469: can you change Gx to 3 too. This way it will be balanced

uri200 (approved):

PR #12235, opened 2022-03-22
chore: modify bzl cc compilation db generation target to include all …

themarwhal:
…C paths

## Summary
The current config only creates compilation db information for the specified binary targets. This works well enough for the most part but excludes all files used for unit testing. This PR modifies the config to just generate everything under `lte/gateway/c/` and `orc8r/gateway/c/`. 

Now all C/C++ paths should resolve by doing the following
1. Run Cmd+Shift+P and Bzl: Bazel/C++: Generate Compilation Database
2. Run Cmd+Shift+P and clangd: Restart language server
## Test Plan
Tested locally with the Magma VM workspace to confirm complation db is produced for test targets
*no more squiggly for MME test files*
Screen Shot 2022-03-22 at 12 41 55 PM

Also spun up a GitHub Codespace off of this PR to test.

## Additional Information

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -66,11 +66,8 @@
 		"editor.formatOnSave": true,
 		// Update this field with any new Bazel targets that need compilation database generation
 		"bsv.cc.compdb.targets": [
-			"//lte/gateway/c/session_manager:sessiond",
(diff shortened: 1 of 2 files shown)


electronjoe (approved):
Nice!

Issue #12231, opened 2022-03-22
[mme]: work to standardize / sort #include statements


electronjoe:
Some of the Bazel migration work may soon move groups of files around into more meanginful / useful new locations (as an effort to remove cycles in our include dependencies broadly).  Moving files like `lte/gateway/c/core/oai/common/assertions.h` results in MANY (order 90) files being touched and updated - and without automated re-sorting of includes we are at a loss as to how to efficiently make these changes.

As an initial attempt, I've authored a go binary to sort targeted blocks of repo includes.  E.g. for any header or c/c++ file, it finds blocks of the form:

```
#include <iostream>
#include "lte/gateway/...../somefile.h"
#include "lte/gateway/..../targetfile.h"
...
#include "lte/gateway/.../stuff.h"

#include "lte/gateway/..../other.h"
```

With a target of `lte/gateway/.../targetfile.h` would sort only the following block in this source segment:

```
#include "lte/gateway/...../somefile.h"
#include "lte/gateway/..../targetfile.h"
...
#include "lte/gateway/.../stuff.h"
```

This approach doesn't "break" any existing formatting decisions - and tries to partially improve sorting. But does not finalize the sorting correctness.

PR #12241, opened 2022-03-22
chore(mme): Sorts includes for blocks including assertion.h

electronjoe:
Progress toward #12231.

This PR is part of a stack of changes:
- **1 of 2** - This PR - #12241
- **2 of 2** #12252

This PR sorts include blocks which contain within them `#include "lte/gaetway/c/core/oai/common/assertion.h"`.  In every block with this include, we sort the block (where block ends at lines missing `#include "` prefix).  This is being done because we want to move some `oai/common/` source files into a new directory in the untangling process (removing cyclic dependencies) and in so doing, we'll want to re-sort these blocks again in those PRs (but doing so was breaking the build due to e.g. missing includes or hidden sequence dependencies).

Achieved with script [here](https://gist.github.com/electronjoe/be515256c71f37a42e335c262ba5ac01). One missing include was detected and subsequently added to this PR - enabling successful `make build_oai` and `make tet_oai`.

```
go run experimental/sort_includes/sort.go -target lte/gateway/c/ -include_path "lte/gateway/c/core/oai/common/assertions.h"
```

## Test Plan

CI Workflow


ssanadhya (changes requested):
Thanks for the cleanup @electronjoe! Could you also fix the lint warning on double includes in `nas_proc.cpp` and `Detach.c`?

electronjoe:
> Thanks for the cleanup @electronjoe! Could you also fix the lint warning on double includes in `nas_proc.cpp` and `Detach.c`?

@ssanadhya done! Good point.

ssanadhya (approved):
LGTM, thanks for the cleanup!

PR #12254, opened 2022-03-23
ci: add mypy reviewdog linter

Neudrino:
## Summary

- add ReviewDog :dog: mypy linter
- added parts are based on existing content with minimal changes
- `github-pr-review` was chosen as reporter. It can later be increased to `github-pr-check`, when quality permits.

## Test Plan

- CI
  - https://github.com/magma/magma/actions/runs/2033463969 Run is green, when no Python is changed.
  - From a PR the correct annotation of PRs cannot be tested due to restrictions on write access of the used GH token. https://github.com/magma/magma/runs/5674167848?check_suite_focus=true#step:3:46

## Additional Information

- Replaces #12008 due to better ReviewDog integration

### DIFF
--- a/.github/workflows/reviewdog-workflow.yml
+++ b/.github/workflows/reviewdog-workflow.yml
@@ -137,6 +137,23 @@ jobs:
           reporter: github-pr-review  # Post code review comments.
           locale: "US"
 
+  mypy:
+    name: mypy
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code.
+        uses: actions/checkout@v2
+        with:
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+          fetch-depth: 0
+      - name: Run mypy with reviewdog
+        uses: tsuyoshicho/action-mypy@v3.6.0
+        with:
+          github_token: ${{ secrets.github_token }}
+          filter_mode: added  # Any added or changed content.
+          reporter: github-pr-review  # Post code review comments.
+
   shellcheck:
     name: shellcheck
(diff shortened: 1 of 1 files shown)


.github/workflows/reviewdog-workflow.yml
```diff
@@ -160,21 +194,17 @@ jobs:
...
       - uses: actions/checkout@v2
         with:
-          fetch-depth: 0
           ref: ${{github.event.pull_request.head.ref}}
           repository: ${{github.event.pull_request.head.repo.full_name}}
-      - name: Get changed files
-        id: py-changes
-        # Set outputs.py to be a list of modified python files
-        run: |
-          echo "::set-output name=py::$(git diff --name-only --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} | grep .py$ | xargs)"
```
Neudrino, line 175: There were the following issues with this part, such that I replaced it.
- Errors occurred on this line.
- It is not reusable for other actions.
- It is not consistent with the common approach of dorny-paths-filter.

maxhbr: If that line is prone to errors and you have a refactoring, you might also want to look at: https://github.com/magma/magma/blob/53c4848439f1a8255a2d3f63219900a8e1e1ac9b/.github/workflows/python-workflow.yml#L63

Neudrino: Good catch. However, overall a more general refactoring and consolidation of linter and formatter application might be beneficial, which I would however see outside of the scope of this PR. :wink: :arrow_right: #12291

tmdzk (approved):

Issue #11638, opened 2022-02-16
Come up with a strategy to manage build / test pipelines for all mme build FEATURES


themarwhal:
## Background
MME currently has too many ways it can be built and tested. (See full list below) We only cover a subset of them in CI, leading to a current brekage on master for unit testing the mme_oai / s6a_fd feature (https://github.com/magma/magma/issues/11617).

## Action Needed
1. We need CI coverage on all build / test commands we support
2. We should minimize the set of different compile time flags as they can get out of hand quickly.
3. It would be good to minimize any build time changes on the set of files built based on flags. (Seen here: https://sourcegraph.com/github.com/magma/magma/-/blob/lte/gateway/c/core/oai/tasks/s6a/CMakeLists.txt) We should keep them if we have a good reason to, but it would be good to minimize it so that the build logic is simpler and easier to maintain.

## Open Questions
1. What are the historical reasons for going with build time flags vs run time configs? (Is it executable size? or?)
2. Do we still want to support a separate build mode for s6a_fd? Specifically, this flag is only used for a case where we want to build the agw_of MME with s6a NOT over grpc.

### Building
- `make build_oai FEATURES='mme_oai' BUILD_TYPE=Debug`
  - ✅ Covered in CI (OAI pipeline)
- `make build_oai FEATURES='mme_oai' BUILD_TYPE=RelWithDebInfo`
  - ✅ Covered in CI (OAI pipeline)
- `make build_oai FEATURES='agw_of' BUILD_TYPE=Debug`
  - ✅ Covered in CI
- `make build_oai FEATURES='agw_of' BUILD_TYPE= RelWithDebInfo`
  - ✅ Covered in CI
- `make build_oai FEATURES='agw_of s6a_fd' BUILD_TYPE=Debug`
  - ❌ **Not covered in CI** 
- `make build_oai FEATURES='agw_of s6a_fd' BUILD_TYPE= RelWithDebInfo`
  - ❌ **Not covered in CI** 
- ~~`make build_oai FEATURES='mme' BUILD_TYPE=Debug`~~
  - ~~⚠️  **Not covered in CI**~~
  - ~~Soon to be deprecated with https://github.com/magma/magma/pull/11615~~
- ~~`make build_oai FEATURES='mme' BUILD_TYPE= RelWithDebInfo`~~
  - ~~⚠️  **Not covered in CI**~~
  - ~~Soon to be deprecated with https://github.com/magma/magma/pull/11615~~

### Testing
[...]

PR #12257, opened 2022-03-23
fix(mme): fix SEGV errors in bazel built mme

### LINKED ISSUE #12232: Fix segmentation fault when running the bazel built MME
## Fix segmentation fault when running the bazel built MME

- Steps to reproduce:
  - Build MME: `bazel build //lte/gateway/c/core:oai_mme` 
  - Run the binary manually `./bazel-bin/lte/gateway/c/core/oai_mme`
  - Results in `Segmentation fault (core dumped)`
- Needs to be fixed in order to proceed with #12057

### Details: `glog` error
When running `lldb ./bazel-bin/lte/gateway/c/core/oai_mme` and starting the run with the parameters `run -c /var/opt/magma/tmp/mme.conf -s /var/opt/magma/tmp/spgw.conf` we find with `bt`:
```
(lldb) bt
* thread #1, name = 'oai_mme', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
  * frame #0: 0x00007ffff741b9e4 libstdc++.so.6`std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::c_str() const + 4
    frame #1: 0x0000555555f59a97 oai_mme`google::VLOG2Initializer() at vlog_is_on.cc:137:44
    frame #2: 0x0000555555f59ebe oai_mme`google::InitVLOG3__(site_flag=0x0000555557216e60, site_default=0x0000555557227ec0, fname="orc8r/gateway/c/common/config/MConfigLoader.cpp", verbose_level=1) at vlog_is_on.cc:206:21
    frame #3: 0x0000555555ee07b1 oai_mme`magma::load_service_mconfig(service_name="mme", config_stream=0x00007fffffffdb70, message=0x00007fffffffddd0) at MConfigLoader.cpp:110:5

```
The SEGV originates from a `MLOG(MERROR)` call in `MConfigLoader.cpp:110:5`. The error can also be reproduced by adding either `VLOG(MERROR);` or simply `LOG(INFO);` somewhere in the `load_service_mconfig` function. Curiously the same calls work if carried out e.g. in the `test_pdu_generator` in `test_pdu_generator.cpp`.

The error is unaffected by the `glog` version, i.e. upgrading to `glog-0.5.0`. The `-lgflags` do not seem to be needed.

### LINKED ISSUE #12233: Fix illegal instruction operand error when running the bazel built MME
## Fix illegal instruction operand error when running the bazel built MME

- Steps to reproduce:
  - Build MME: `bazel build //lte/gateway/c/core:oai_mme` 
  - Run the binary manually `./bazel-bin/lte/gateway/c/core/oai_mme`
  - Results in `Segmentation fault (core dumped)`
- Needs to be fixed in order to proceed with #12057

### Details: `zmq` socket error

If we comment out the `MLOG(MERROR)` in `MConfigLoader.cpp:110:5` (see https://github.com/magma/magma/issues/12232) we find another SIGILL error when running `lldb ./bazel-bin/lte/gateway/c/core/oai_mme` and starting the run with the parameters `run -c /var/opt/magma/tmp/mme.conf -s /var/opt/magma/tmp/spgw.conf` we find with `bt`:
```
(lldb) bt
* thread #2, name = 'oai_mme', stop reason = signal SIGILL: illegal instruction operand
  * frame #0: 0x00005555557c8fd4 oai_mme`init_task_context(task_id=TASK_SHARED_TS_LOG, remote_task_ids=0x00007ffff5c428ec, remote_tasks_count='\0', msg_handler=(oai_mme`handle_message at shared_ts_log.c:165:70), task_zmq_ctx_p=0x000055555722ae60) at intertask_interface.c:217:5
```
which occurs at `intertask_interface.c:217:5` at `task_zmq_ctx_p->pull_sock` due to the fact that the function call to `zsock_new_pull` in `task_zmq_ctx_p->pull_sock = zsock_new_pull(itti_desc.tasks_info[task_id].uri);` returns `NULL`. The argument `itti_desc.tasks_info[task_id].uri` is `ipc:///run/IPC_TASK_SHARED_TS_LOG`, which seems correct.

LKreutzer:

## Summary

- Bazel built MME is linked **dynamically** and with `EMBEDDED_SGW` set.
- Minimal changes that **make the bazel built MME binary runnable** and fix:
  - #12232 
  -  #12233
- Note: The `glog` dependency needs to be linked dynamically (see https://github.com/google/glog/issues/53), ideally `linkstatic = False` should thus only apply to `glog`. 
- Note: `bazel run` on MME needs runtime options `-c /var/opt/magma/tmp/mme.conf -s /var/opt/magma/tmp/spgw.conf`. Can these be set at the `cc_binary`?

## Test Plan

- Run MME with bazel:
  `bazel run lte/gateway/c/core:oai_mme -- -c /var/opt/magma/tmp/mme.conf -s /var/opt/magma/tmp/spgw.conf`

## Additional Information

### DIFF
--- a/lte/gateway/c/core/BUILD.bazel
+++ b/lte/gateway/c/core/BUILD.bazel
@@ -1138,6 +1138,10 @@ cc_binary(
     srcs = [
         "oai/oai_mme/oai_mme.c",
     ],
+    copts = [
+        "-DEMBEDDED_SGW",  # TODO(#11638): Feature flags fixups
+    ],
+    linkstatic = False,
     deps = [
         ":core",
     ],


electronjoe (approved):

electronjoe:
Do you know why LinkStatic=False was needed?

Aah I see in your PR Description

> Note: The glog dependency needs to be linked dynamically (see https://github.com/google/glog/issues/53), ideally linkstatic = False should thus only apply to glog.

lte/gateway/c/core/BUILD.bazel
```diff
@@ -1138,6 +1138,10 @@ cc_binary(
     srcs = [
         "oai/oai_mme/oai_mme.c",
     ],
+    copts = [
+        "-DEMBEDDED_SGW",  # TODO(#11638): Feature flags fixups
```
themarwhal, line 1142: Thanks for looking into this issue!! I think eventually (maybe soon), we should pull these copts to per_file_copts that can be controlled by a build time flag. Where the build time flag corresponds to the FEATURES we support in Makefile. see https://github.com/magma/magma/issues/11638

So something like `bazel build <MME TARGET> --config=of_mme`

lte/gateway/c/core/BUILD.bazel
```diff
@@ -1138,6 +1138,10 @@ cc_binary(
...
+    ],
+    linkstatic = False,
```
themarwhal, line 1144: :'(

themarwhal (approved):

lte/gateway/c/core/BUILD.bazel
themarwhal: I wonder why we haven't had this issue for other binaries that depend on glog.

LKreutzer: We experimented with a dummy C++ service in bazel that uses `glog` and static linking: as long as there are no logging calls in static variables the error does not occur. We were able to replicate the SEGV in the dummy service by calling logging in a static variable, see the example given in https://github.com/google/glog/issues/53.

PR #12264, opened 2022-03-24
fix(orc8r): Fixed simple gateway registration cli

christinewang5:
* added function save_root_ca to use the provided value of the root ca from the command line argument
* added function save_control_proxy because previously save_override_config was assuming that the control proxy was json decoded which formatted the writing incorrectly
* fixed the challenge key, before it was being base64 encoded twice
* added a sleep so that the bootstrapper manager has sufficient time to run bootstrap before running checkin cli

* manually tested

### DIFF
--- a/orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
+++ b/orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
@@ -2,6 +2,7 @@ package registration
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 
 	"github.com/go-openapi/strfmt"
@@ -67,8 +68,11 @@ func (r *RegistrationService) Register(c context.Context, request *protos.Regist
 }
 
 func RegisterDevice(deviceInfo protos.GatewayDeviceInfo, hwid *protos.AccessGatewayID, challengeKey *protos.ChallengeKey) error {
-	gatewayRecord := createGatewayDevice(hwid, challengeKey)
-	err := device.RegisterDevice(context.Background(), deviceInfo.NetworkId, orc8r.AccessGatewayRecordType, hwid.Id, gatewayRecord, serdes.Device)
+	gatewayRecord, err := createGatewayDevice(hwid, challengeKey)
+	if err != nil {
+		return err
+	}
+	err = device.RegisterDevice(context.Background(), deviceInfo.NetworkId, orc8r.AccessGatewayRecordType, hwid.Id, gatewayRecord, serdes.Device)
 	return err
 }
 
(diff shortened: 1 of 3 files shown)


orc8r/gateway/python/scripts/register.py
```diff
@@ -107,22 +135,34 @@ def main():
...
         print(msg.format(e))
         sys.exit(1)
 
-    if not args.no_control_proxy:
-        save_override_config("control_proxy", res.control_proxy)
+    print("> Waiting {} seconds for next bootstrap".format(
+        BootstrapManager.LONG_BOOTSTRAP_RETRY_INTERVAL.total_seconds()))
+    print()
 
+    time.sleep(BootstrapManager.LONG_BOOTSTRAP_RETRY_INTERVAL.total_seconds())
```
christinewang5: Idk if this is necessary, just that if we run the script without this, it'll fail. Lmk if there are better alternatives though

orc8r/cloud/go/services/bootstrapper/servicers/registration/registration.go
```diff
@@ -114,12 +118,16 @@ func makeErr(errString string) *protos.RegisterResponse {
...
 // createGatewayDevice creates the gateway device model
-func createGatewayDevice(hwID *protos.AccessGatewayID, challengeKey *protos.ChallengeKey) *models.GatewayDevice {
-	challengeKeyBase64 := strfmt.Base64(challengeKey.Key)
+func createGatewayDevice(hwID *protos.AccessGatewayID, challengeKey *protos.ChallengeKey) (*models.GatewayDevice, error) {
+	decodedKey, err := base64.StdEncoding.DecodeString(string(challengeKey.Key))
+	if err != nil {
+		return nil, err
+	}
+	challengeKeyBase64 := strfmt.Base64(decodedKey)
 	return &models.GatewayDevice{
```
hcgatewood: Nit: I don't like returning literal structs because I prefer always a single-line return for readability. So i.e. create a variable := the struct literal, then return the variable (and nil)

orc8r/gateway/python/scripts/register.py
```diff
@@ -77,7 +102,7 @@ def main():
     parser.add_argument(
         "--ca-file",
         type=str,
-        help="orc8r's root CA file",
+        help="orc8r's root CA file contents",
```
hcgatewood: I think let's leave this the way it was before actually -- it's kinda awkward to paste the entire root CA contents into the CLI cmd. (Or did you have thoughts on why this is preferable?)

hcgatewood (changes requested):
Looks good! Couple small notes. Also can you also ensure you can pass the cloud_port as a number (ie not quoted)

orc8r/gateway/python/scripts/register.py
christinewang5: I was just thinking that we never really use the contents from the response body `registration_info: {"rootCA": "...", ....}` and it should probably get passed in somewhere. 
Before I assumed it was taking in a filepath, but does the `rootCA.pem` live anywhere but the `/var/opt/magma/certs` directory? I think it might be weird for the operator to know to paste the `rootCA.pem` to a random location and then point to this file through the script? Idk lemme know your thoughts.

hcgatewood: Yeah so the idea is they would save the root CA cert to the file system, then point to it from the cli invocation. I think that's a bit cleaner, just cause pasting the whole contents of a cert into a cli invocation is a bit much

(The "win" is, if we can get moved over to using cert-manager, AWS-deployed Orc8rs can skip this step)

hcgatewood (approved):

ardzoht (approved):
agw changes lgtm

ajahl (approved):

Issue #11892, opened 2022-03-02
Super-Issue: Break up Bazel OAI Core Bazel build label


electronjoe:
## Problem Statement

You want Bazel build labels to be ~reasonably sized (as a general rule the smaller the better - I often aim for singular cpp/hpp pair per `cc_library` label).  Instead we have order 500 c/c++ files AND 500 headers in one giant label at `//lte/gateway/c/core:core` in #11862.

Many attempts were made to break up this target but repeatedly we were blocked - primarily by transitive cyclic dependencies that covered the majority of the 1,000 artifacts of this label.  **Bazel does not allow for labels to cyclically depend upon each other (the build graph must be a directed acyclic graph)**.

Fixing this is a major effort - but by breaking up the labels / fixing code cycles you will gain at least the following if not more:

- Clearer code dependencies for newcomers
- Greatly improved build and test iteration times between modifications
- Greatly improved Bazel cache behavior (e.g. Marie's remote cache fetch for CI or local dev)
- Improved Bazel build taget path semantics (build label is adjacent to source files, not in some parent directory)

## Path to Betterment

I believe the following process may be the best for refactoring and fixing these cycles / improving Bazel build for the OAI Core.

**Key Constraints**

- Bazel does not allow cycles in dependencies of build labels
- Bazel does not allow a `module` (directory with a BUILD.bazel) if a parent module already specifies artifacts within

So I think the process might look something like this:

1. Hunt for a small component (one or more header files and zero or more c/c++ files) that has no transitive include cycles with itself. Or create one by refactoring an existing cycle (techniques for this need to be learned - much easier in CPP with interfaces).
1. Move that component into a new `cc_library` in a new module location (say `lte/gateway/c/core/_meaningful_module_name` directory)
1. Generate a Bazel module in that direcctory, specify the `cc_library`
1. Update bazel rules across the repo which include this library (by adding to deps)
1. Update all include paths across the repo for the new directory location

This process enables us to move "cleaned" code into a new Bazel module and avoids the second constraint above. Further, it is likely that the code organization could use improvement anyway - this is an opportunity to do so.

See https://github.com/magma/magma/issues/11948 for an analysis of current dependency structure.

PR #12265, opened 2022-03-24
chore(mme): move some sources out of c/core/oai/common and ...

electronjoe:
...split bazel target for lib/hashtable

Working towards #11892.

The purpose of this PR is to begin the process (demonstrate a path) of untangling cyclic dependencies that have presently been balled up into one very large Bazel label at `//lte/gateway/c/core:core`.  Untangling them reduces technical debt, but also greatly improves cache behavior in Bazel, build and test iteration times for sub-components.

This PR:

1. Moves five source files with no external cyclic dependencies from `lte/gateway/c/core/oai/common` to `lte/gateway/c/core/common`.  This move is required due to Bazel constraints (you cannot create a Bazel module if some source files in that directory are already in parent-directory modules - ergo we cannot create /c/core/oai/common/BUILD.bazel module if some files in oai/common/ are specified in `:core` already).
2. Generates isolated Bazel Build targets for these sources within `lte/gateway/c/core/common/BUILD.bazel` and updates all BUILD.bazel repo-wide with need of these dependencies
3. Establishes a new CMake external project within `lte/gateway/c/core/common` and adds it as a dependency to `lte/gateway/c/core/oai/common/CMakeLists.txt`. Note that this is a shortcut - what should be done is to hunt down EVERY CMake target that needs these dependencies and add the linking directive. But CMake use is on a fast path to deprecation at this point and this solution "works" because everybody who needed these assets already links against Common.
4. With 1-3 above done it was now possible to break out `lib/hashtable` into a new separate Bazel label as well (removing it from the giant `:core` label)

Note that 1-3 are "atomic" and I don't see how I can make this PR smaller [edit: I could have started with lesser migration targets that are not so frequently used across the codebase...].  Line item 4 is optional but a very small contribution to the size of this PR.

## Test Plan

Locally validated changes in GitHub Codespaces by validating success of:

```
bazel build //lte/gateway/c/... <success>
bazel test //lte/gateway/c/... <success>
cd lte/gateway
make build_oai <success>
make test_oai <success>
```

Signed-off-by: Scott Moeller [electronjoe@gmail.com](mailto:electronjoe@gmail.com)


nstng (approved):
This looks like a blueprint for a couple of followup tasks :)

Only minor question ...
Moving sources out of the core feels wired, but I know about the restriction. Do you see this as an intermediate state that should be refactored when core/oai/common becomes a Bazel package? Should it be documented or tracked in an issue?

electronjoe:
> Moving sources out of the core feels wired

This was the feedback from @ardzoht as well - so I am going to instead move the sources to `lte/gateway/c/core/common`.

The goal would be that all source would be moved down a dir to `c/core` and `oai` would be deprecated.

Does that address your concern?

nstng:
> > Moving sources out of the core feels wired
> 
> This was the feedback from @ardzoht as well - so I am going to instead move the sources to `lte/gateway/c/core/common`.
> 
> The goal would be that all source would be moved down a dir to `c/core` and `oai` would be deprecated.
> 
> Does that address your concern?

This sounds like a good idea - with this the way forward is clear and no post refactoring is needed.

ardzoht (approved):
lgtm, synced offline with @electronjoe regarding the `lte/gateway/c/core` hierarchy, the desired structure should be files being moved out of `../c/core/oai` should be moved to `../c/core/common`

PR #12274, opened 2022-03-24
chore(agw): fix mypy issues

Neudrino:
## Summary

- get `lte/gateway/python` mypy clean
- Follow up to https://github.com/magma/magma/pull/12086

## Test Plan

- CI
- :heavy_check_mark: `${MAGMA_ROOT}$ mypy lte/gateway/python/`
  `Success: no issues found in 179 source files`
- :heavy_check_mark: `${MAGMA_ROOT}/lte/gateway/python$ make test_all`
- :heavy_check_mark: `${MAGMA_ROOT}/lte/gateway$ make test` with log output [lte-gateway-test.log](https://github.com/magma/magma/files/8379361/lte-gateway-test.log).

### DIFF
--- a/lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
+++ b/lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
@@ -24,7 +24,7 @@ import logging
 from copy import deepcopy
 from ipaddress import ip_address, ip_network
 from threading import Condition
-from typing import List, Optional, cast
+from typing import List, cast
 
 from magma.mobilityd.ip_descriptor import IPDesc, IPState, IPType
 
@@ -123,7 +123,7 @@ class IPAllocatorDHCP(IPAllocator):
         """
         mac = create_mac_from_sid(sid)
 
-        dhcp_desc = self._dhcp_client.get_dhcp_desc(mac, str(vlan_id))
+        dhcp_desc = self._dhcp_client.get_dhcp_desc(mac, vlan_id)
         LOG.debug(
             "allocate IP for %s mac %s dhcp_desc %s", sid, mac,
             dhcp_desc,
--- a/lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
+++ b/lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
@@ -129,10 +129,10 @@ class IPAllocatorDHCP(IPAllocator):
             dhcp_desc,
         )
 
-        if dhcp_allocated_ip(dhcp_desc) is not True:
+        if not dhcp_desc or not dhcp_allocated_ip(dhcp_desc):
             dhcp_desc = self._alloc_ip_address_from_dhcp(mac, vlan_id)
 
-        if dhcp_allocated_ip(dhcp_desc):
+        if dhcp_desc and dhcp_allocated_ip(dhcp_desc):
             ip_block = ip_network(dhcp_desc.subnet)
             ip_desc = IPDesc(
                 ip=ip_address(dhcp_desc.ip),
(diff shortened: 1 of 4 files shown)


ardzoht:
@Neudrino I would recommend to run `make test` as well on the magma VM. That will also verify coverage for some unit tests that are not run as part of the current GH action and are run on post-merge.

lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
```diff
@@ -200,7 +200,10 @@ def _alloc_ip_address_from_dhcp(
             dhcp_desc = None
             while (
                 retry_count < self._retry_limit
-                and dhcp_allocated_ip(dhcp_desc) is not True
+                and (
+                    (dhcp_desc is None)
+                    ^ ((dhcp_desc is not None) and not dhcp_allocated_ip(dhcp_desc))
+                )
             ):
```
pshelar, line 206: I dont see reason to make this condition complicated? why is 'dhcp_desc' None check is moved from dhcp_allocated_ip() all call sites?

Neudrino:
> @Neudrino I would recommend to run `make test` as well on the magma VM. That will also verify coverage for some unit tests that are not run as part of the current GH action and are run on post-merge.

Thanks, @ardzoht for the hint! :pray: 
Unfortunately there is no target `test` within the Makefile in `lte/gateway/python`.
The target `test_all` also runs on CI on PR (agw-workflow), so that is already covered.
The target `check` does not seem to run on CI, but the included linters always fail with hundreds of issues, which makes them quite useless. Those heaps of issues were not introduced by this PR.
The target `coverage` runs without issues locally, but only determines coverage, which is done via [codecov](https://app.codecov.io) in CI I believe.

So could you be a bit more specific, which tests exactly you mean, i.e. which makefile target?
Is that (difference and workflow) actually documented?
Maybe it would than be sensible to include those tests in the CI as well?

lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
Neudrino: This changes was triggered by a mypy issue which was tricky to remove. I realized that it could further be simplified and did so.
Moreover there was a inconsistency regarding null checks in the code, which was cleaned up with this refactoring. Those null checks are nowhere else moved to a deeper level but for this specific one. In addition having this check inside the method (combined with multiple levels of negation) made the logic harder comprehend in a lot if places. By moving this out the logic difficult to comprehend logic could be simplified and thus be made much more transparent. All usages in other places benefit from a much clearer and more consistent logical structure as well.

ardzoht:
> > @Neudrino I would recommend to run `make test` as well on the magma VM. That will also verify coverage for some unit tests that are not run as part of the current GH action and are run on post-merge.
> 
> Thanks, @ardzoht for the hint! 🙏 Unfortunately there is no target `test` within the Makefile in `lte/gateway/python`. The target `test_all` also runs on CI on PR (agw-workflow), so that is already covered. The target `check` does not seem to run on CI, but the included linters always fail with hundreds of issues, which makes them quite useless. Those heaps of issues were not introduced by this PR. The target `coverage` runs without issues locally, but only determines coverage, which is done via [codecov](https://app.codecov.io) in CI I believe.
> 
> So could you be a bit more specific, which tests exactly you mean, i.e. which makefile target? Is that (difference and workflow) actually documented? Maybe it would than be sensible to include those tests in the CI as well?

These are difficult to include on GH action as there are dependencies that are run only on the VM such as OVS dependency. The `make test` target is available under the `magma/lte/gateway` filepath. The difference against the `make test_all` under `magma/lte/gateway/python` is that this one only runs `lte->python` tests and not `orc8r->python` tests, for the purpose of this PR, `make test_all` is sufficient.

ardzoht (approved):
lgtm, this can be merged with @pshelar final approve.

pshelar (approved):

pshelar (changes requested):
Can you confirm that `make test` in let/gateway passes?

Neudrino:
> Can you confirm that `make test` in let/gateway passes?

Sure. I just ran `${MAGMA_ROOT}/lte/gateway$ make test` locally and it went trough fine, afaiks. :heavy_check_mark: 
Local log is: [lte-gateway-test.log](https://github.com/magma/magma/files/8379361/lte-gateway-test.log).
I will also update the PR description.

pshelar (approved):

Issue #12198, opened 2022-03-18
Freeze fluentd docker image gem versions


sklgromek:
Currently, fluentd_forward docker image is built together with fluentd-plugin-elasticsearch  which doesn't specify Elasticsearch ruby module  version in dependencies:
https://github.com/uken/fluent-plugin-elasticsearch/blob/1091d8b00ecfb52bbcdf5bc7c8e8a1f58f04e287/fluent-plugin-elasticsearch.gemspec#L23

As a result, the latest version of the library is installed, which is not compatible with the version of the fluent Elasticsearch plugin that is installed when building the image.

Fluentd returns the following error:
```2022-03-17 12:28:22 +0000 [info]: parsing config file is succeeded path="/fluentd/etc/fluent.conf"
/usr/lib/ruby/gems/2.5.0/gems/fluent-plugin-elasticsearch-3.5.5/lib/fluent/plugin/oj_serializer.rb:7:in `<class:Oj>': uninitialized constant Elasticsearch::Transport (NameError)
        from /usr/lib/ruby/gems/2.5.0/gems/fluent-plugin-elasticsearch-3.5.5/lib/fluent/plugin/oj_serializer.rb:6:in `<module:Serializer>'
        from /usr/lib/ruby/gems/2.5.0/gems/fluent-plugin-elasticsearch-3.5.5/lib/fluent/plugin/oj_serializer.rb:4:in `<module:Plugin>'
        from /usr/lib/ruby/gems/2.5.0/gems/fluent-plugin-elasticsearch-3.5.5/lib/fluent/plugin/oj_serializer.rb:3:in `<top (required)>'
        from /usr/lib/ruby/gems/2.5.0/gems/fluent-plugin-elasticsearch-3.5.5/lib/fluent/plugin/out_elasticsearch.rb:22:in `require_relative'
        from /usr/lib/ruby/gems/2.5.0/gems/fluent-plugin-elasticsearch-3.5.5/lib/fluent/plugin/out_elasticsearch.rb:22:in `<top (required)>'
        from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
        from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
        from /usr/lib/ruby/gems/2.5.0/gems/fluentd-1.7.4/lib/fluent/registry.rb:102:in `block in search'
[...]

PR #12284, opened 2022-03-25
fix(orc8r): Freeze fluentd docker image gem versions #12198

sklgromek:
Signed-off-by: Tomasz Gromowski <tomasz.gromowski@codilime.com>

## Summary
Fluentd elasticsearch plugin doesn't define the version of Elasticsearch gem. So always recent is installed, which doesn't work correctly with fluentd-plugin-elasticsearch version 3.5.5.
https://github.com/uken/fluent-plugin-elasticsearch/blob/96ff1706282a3bdb77533b72d3792d692530b547/fluent-plugin-elasticsearch.gemspec#L27

This change will freeze Elasticsearch  gem version to 7.13, which is compatible with both Elasticsearch and Opensearch
https://opensearch.org/docs/latest/clients/index/
It will also bump fluentd elasticsearch plugin version to latest 5.2.1

## Additional Information

### DIFF
--- a/orc8r/cloud/docker/fluentd/Dockerfile
+++ b/orc8r/cloud/docker/fluentd/Dockerfile
@@ -1,5 +1,8 @@
 FROM fluent/fluentd:v1.7-1
 USER root
-RUN ["gem", "install", "fluent-plugin-elasticsearch", "--no-document", "--version", "3.5.5"]
-RUN ["fluent-gem", "install", "fluent-plugin-multi-format-parser"]
+RUN gem install \
+    elasticsearch:7.13.0 \
+    fluent-plugin-elasticsearch:5.2.1 \
+    fluent-plugin-multi-format-parser:1.0.0 \
+    --no-document
 USER fluent
--- a/orc8r/cloud/docker/fluentd_forward/Dockerfile
+++ b/orc8r/cloud/docker/fluentd_forward/Dockerfile
@@ -1,5 +1,8 @@
 FROM fluent/fluentd:v1.7-1
 USER root
-RUN ["gem", "install", "fluent-plugin-elasticsearch", "--no-document", "--version", "3.5.5"]
-RUN ["fluent-gem", "install", "fluent-plugin-multi-format-parser"]
+RUN gem install \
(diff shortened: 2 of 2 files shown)


orc8r/cloud/docker/fluentd/Dockerfile
```diff
@@ -1,5 +1,8 @@
 FROM fluent/fluentd:v1.7-1
 USER root
-RUN ["gem", "install", "fluent-plugin-elasticsearch", "--no-document", "--version", "3.5.5"]
-RUN ["fluent-gem", "install", "fluent-plugin-multi-format-parser"]
+RUN gem install \
+    elasticsearch:7.13.0 \
+    fluent-plugin-elasticsearch:5.2.1 \
+    fluent-plugin-multi-format-parser:1.0.0 \
+    --no-document
 USER fluent
```
Neudrino: Could we add a EOL at EOF as a cleanup here, please?

Neudrino (commented):
_More general:_
If those two docker files are exactly identical, I wonder, why we need both? :thinking:

ShubhamTatvamasi:
Hi @sklgromek, can you have a look where are we using `fluentd_forward/Dockerfile` file?

sklgromek:
@ShubhamTatvamasi @Neudrino 
> _More general:_ If those two docker files are exactly identical, I wonder, why we need both? thinking

I'm not sure about it. Fluent_forward is used by helm orc8r charts. For fluentd I see reference here https://github.com/openEPC/magma/blob/53c4848439f1a8255a2d3f63219900a8e1e1ac9b/orc8r/cloud/docker/docker-compose.yml#L133. As they were identical I fixed both. One of them can probably be removed. Do you want me to do this in this PR?

Neudrino:
> Do you want me to do this in this PR?

Doing it in a separate PR would work fine for me. (As a matter of taste, I would even prefer that.)
You might also consider creating an issue for it.

Neudrino (approved):

ShubhamTatvamasi (approved):
LGTM

PR #12285, opened 2022-03-25
chore(pipelined): devcontainer and bazel-base use needed depencies

vktng:

## Summary

Add needed dependencies for devcontainer and bazel-base which are required by https://github.com/magma/magma/pull/12271

## Test Plan

No docker build failures in CI. Tested locally with #12271.

## Additional Information

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -71,6 +71,7 @@ RUN echo "Install general purpose packages" && \
         lld \
         lldb \
         make \
+        netbase \
         ninja-build \
         openjdk-8-jdk \
         perl \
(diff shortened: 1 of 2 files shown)


electronjoe (approved):
Are the Hadolint findings in the PR useful?

vktng:
> Are the Hadolint findings in the PR useful?

They make sense, but we never used version pinning in these Dockerfile-s. If we want to use this, then we should do it in a separated technical debt task. The other issues were not introduced here, but are related to the changed lines (of course this is not an argument to not fix it, but this should again be done in a separate task for all occurences). Is this ok for you?

nstng (approved):
in pairing

PR #12291, opened 2022-03-25
chore(ci): remove unstable bash code in favour of Github actions

Neudrino:
## Summary

- remove duplication in change detection in favour of dorny path filter of unstable bash hack
- This workflow only does linting and formatting, such that the non-python workflow file should not be included in the filter.

## Test Plan

- CI:
  - https://github.com/magma/magma/actions/runs/2051685689 correctly run for no changes.
  - https://github.com/magma/magma/actions/runs/2051704767 correctly fail with arbitrary new line added to python file.

## Additional Information

- Suggested in https://github.com/magma/magma/pull/12254#discussion_r835317696

### DIFF
--- a/.github/workflows/python-workflow.yml
+++ b/.github/workflows/python-workflow.yml
@@ -23,6 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       should_not_skip: ${{ steps.changes.outputs.filesChanged }}
+      files_changed: ${{ steps.changes.outputs.filesChanged_files }}
     steps:
       # Need to get git on push event
       - uses: actions/checkout@v2
@@ -32,7 +33,8 @@ jobs:
         with:
           filters: |
             filesChanged:
-              - [".github/workflows/python-workflow.yml", "**/*.py"]
+              - ["**/*.py"]
+          list-files: 'shell'
       # Need to save PR number as Github action does not propagate it with workflow_run event
       - name: Save PR number
         if: always()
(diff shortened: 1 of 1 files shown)


.github/workflows/python-workflow.yml
```diff
@@ -56,32 +58,24 @@ jobs:
...
-        name: Format and check for leftover changes
+      - name: Format and check for leftover changes
         uses: addnab/docker-run-action@v3
         with:
           image: magma/py-lint:latest
           options: -u 0 -v ${{ github.workspace }}:/code
           run: |
             echo "Running formatting tools. This should be equivalent to running './lte/gateway/python/precommit.py --format --diff' locally."
-            for file in ${{ steps.py-changes.outputs.py }};
+            for file in ${{ needs.pre_job.outputs.files_changed }};
```
maxhbr, line 75: (like before your change) this fails with spaces in filenames?

.github/workflows/python-workflow.yml
```diff
@@ -56,32 +58,24 @@ jobs:
...
             do
               set -e
-              echo ""
-              echo "Running isort on $file...";
+              echo -e "\nRunning isort on $file...";
```
maxhbr: not sure if that is an improvement

.github/workflows/python-workflow.yml
Neudrino: This should work for filenames with spaces (or other not too terrible special characters) due to [this change](https://github.com/magma/magma/pull/12291/files#diff-3159a3c5e81c321bcf5c551bad19ed48d0d2dc0109348184f6bb03f0ba5f50ceR37).
As opposed to the previous manual implementation, I would expect that the action delivers a set of properly quoted strings, the same way `ls` does nowadays.

.github/workflows/python-workflow.yml
Neudrino: For me it is.

maxhbr (approved):
looks good to me

tmdzk (approved):
LGTM

PR #12297, opened 2022-03-25
fix(mme): Reduce flakiness in MME procedure unit tests

### LINKED ISSUE #12166: mme: //lte/gateway/c/core/oai/test/mme_app_task:mme_procedures_test is flaky
## Repro steps
Run `bazel test //lte/gateway/c/core/oai/test/mme_app_task:mme_procedures_test --runs_per_test=50 `

ssanadhya:

## Summary

MME procedures test have been flaky because:
1. The timer values were reduced for all tests, leading to unexpected timer expiration and too many message retransmissions, leading to more than expected calls to s1ap_generate_dl_nas_transport()
2. The S6a handler was not starting before the test ran, leading to zero calls to s6a_* functions

This PR fixes these issues by:
1. Removing the timer override from default Setup() call and only modifying individual timer durations as part of individual tests.
2. Reordering the thread creation and adding a 10 msec wait in the setup of each test to make sure all mock tasks are running before the test begins

## Test Plan

No failure seen with the following:

`bazel test //lte/gateway/c/core/oai/test/mme_app_task/... --runs_per_test=50`
`bazel test //lte/gateway/c/core/oai/test/mme_app_task/... --runs_per_test=100`

**TODO**
Test `TestImsiAttachRejectIdentRetxFailure` is testing Identity retx with an IMSI attach, which seems incorrect. This needs to be cleaned up in a separate PR before the call to `nas_config_timer_reinit` can be replaced with a specific timer.

### DIFF
--- a/lte/gateway/c/core/oai/test/mme_app_task/mme_procedure_test_fixture.h
+++ b/lte/gateway/c/core/oai/test/mme_app_task/mme_procedure_test_fixture.h
@@ -137,8 +137,8 @@ class MmeAppProcedureTest : public ::testing::Test {
     task_sgw_s8.detach();
     task_sms_orc8r.detach();
 
-    // Sleep for 10 milliseconds to make sure all mock tasks are up before the
-    // test
+    // Sleep for 10 milliseconds to make sure all mock tasks
+    // are running before the test starts
     std::this_thread::sleep_for(std::chrono::milliseconds(10));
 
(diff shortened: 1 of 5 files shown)


pruthvihebbani (approved):

maxhbr (approved):
LGTM

pruthvihebbani:
@ssanadhya please rebase with latest master

PR #11615, opened 2022-02-15
chore(mme): scrub the mme feature from Makefile

themarwhal:
Signed-off-by: GitHub <noreply@github.com>

## Summary
Discussion: https://magmacore.slack.com/archives/C02GMJ8S8FR/p1644937945289039

The mme feature flag is no longer used. Removing it to avoid confusion.

According to Shruti:
> the gtpnl data path feature set was removed in https://github.com/magma/magma/pull/9747/files . I think the mme feature > was kept to accommodate other S6a over GRPC with non-openflow data path, but we never used that one, so it is better 
> to remove it.

In testing all build features for MME, I see that unit tests do not compile (fail to link) when compiled with the `-DS6A_OVER_GRPC=False` flag. This behavior is consistently seen on master as well, so I think this PR is not breaking any feature that is working on master. Filed a follow up task : https://github.com/magma/magma/issues/11617

## Test Plan
Triggered an LTE Integ Test run here: https://github.com/themarwhal/magma/actions/runs/1848608621

👇 Manual testing in GitHub codespaces 👇 
### default, nothing specified (agw_of)
✅ build
```
magma % make -C lte/gateway build_oai
make: Entering directory '/workspaces/magma/lte/gateway'
BAZEL_FLAGS --config=development
MAIN_FEATURE agw_of
OAI_FLAGS -DS6A_OVER_GRPC=True -DEMBEDDED_SGW=True
```
✅ test
```
magma % make -C lte/gateway test_oai
... Success
```

### manually specify agw_of (should be identical to the one above)
✅ build
```
magma % make FEATURES='agw_of' -C lte/gateway build_oai 
make: Entering directory '/workspaces/magma/lte/gateway'
BAZEL_FLAGS --config=development
MAIN_FEATURE agw_of
OAI_FLAGS -DS6A_OVER_GRPC=True -DEMBEDDED_SGW=True
```
✅ test
```
magma % make FEATURES='agw_of' -C lte/gateway test_oai 
... Success
```

### manually specify agw_of AND s6a_fd
✅ build
```
magma % make FEATURES='agw_of s6a_fd' -C lte/gateway build_oai 
make: Entering directory '/workspaces/magma/lte/gateway'
BAZEL_FLAGS --config=development
MAIN_FEATURE agw_of
OAI_FLAGS -DS6A_OVER_GRPC=False -DEMBEDDED_SGW=True
```
❌ test (also broken on master)
```
[...]

PR #12320, opened 2022-03-28
poc for what MME feature configuration could look like

themarwhal:

## Summary
I was thinking we could configure MME features like this, any thoughts? @LKreutzer / @nstng 

## Additional Information

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -31,6 +31,12 @@ build:devcontainer --disk_cache=/workspaces/magma/.bazel-cache
 common:devcontainer --repository_cache=/workspaces/magma/.bazel-cache-repo
 build:devcontainer --define=folly_so=1
 
+# MME FEATURES (default is agw_of)
+build --config=agw_of
+
+build:agw_of --per_file_copt=^lte/gateway/c/core/.*$@-DEMBEDDED_SGW=1,-DS6A_OVER_GRPC=1
+build:mme_oai --per_file_copt=^lte/gateway/c/core/.*$@-DEMBEDDED_SGW=0,-DS6A_OVER_GRPC=0
+
 # TEST CONFIGS
 # Bazel test runtime default: PATH=/bin:/usr/bin:/usr/local/bin
 # Some python tests require access to /usr/sbin binaries (e.g. route, ifconfig)
--- a/lte/gateway/c/core/BUILD.bazel
+++ b/lte/gateway/c/core/BUILD.bazel
@@ -1062,8 +1062,6 @@ cc_library(
         "oai/tasks/sms_orc8r/sms_orc8r_messages.h",
     ],
     copts = [
-        "-DEMBEDDED_SGW",  # TODO(#11638): Feature flags fixups
-        "-DS6A_OVER_GRPC",  # TODO(#11638): Feature flags fixups
         "-DCMAKE_BUILD_TYPE=\\\"CMAKE_BUILD_TYPE\\\"",  # TODO(#11897): needs alternative approach or deprecation
     ],
     deps = [
@@ -1129,9 +1127,6 @@ cc_binary(
     name = "oai_mme",
     srcs = [
         "oai/oai_mme/oai_mme.c",
-    ],
-    copts = [
-        "-DEMBEDDED_SGW",  # TODO(#11638): Feature flags fixups
     ],
     linkstatic = False,
     deps = [


nstng:
This looks like a reasonable approach to bundle and toggle build flags. I can not comment the initial setup you posted because I do not know in what situation what flag makes sense - and what the different use cases are.

Some thoughts ...
* With this is would be possible to do `bazel build //foo --config=mme_oai`
   i.e., `EMBEDDED_SGW=1, S6A_OVER_GRPC=1, S6A_OVER_GRPC=0`
   What `S6A_OVER_GRPC` setting would win?
* I assume `agw_of` and `mme_oai` are meant to be set via command line - but `embedded_sgw`, `s6a_over_grpc`, `s6a_over_fd` do not make sense alone? I.e., can we make sure these can not be set via command line? I do not assume bazel allows this kind of configuration.
* Does it make sense to consolidate the use cases and bundle the flags in code? E.g., do not have
   `agw_of ^= EMBEDDED_SGW=1, S6A_OVER_GRPC=1`
   but have `OGW_OF` in code to do the same and then setting the flag via `per_file_copt`.
* nit: is it `agw_off`?

themarwhal:
> This looks like a reasonable approach to bundle and toggle build flags. I can not comment the initial setup you posted because I do not know in what situation what flag makes sense - and what the different use cases are.
> 
> Some thoughts ...
> 
> * With this is would be possible to do `bazel build //foo --config=mme_oai`
>   i.e., `EMBEDDED_SGW=1, S6A_OVER_GRPC=1, S6A_OVER_GRPC=0`
>   What `S6A_OVER_GRPC` setting would win?
> * I assume `agw_of` and `mme_oai` are meant to be set via command line - but `embedded_sgw`, `s6a_over_grpc`, `s6a_over_fd` do not make sense alone? I.e., can we make sure these can not be set via command line? I do not assume bazel allows this kind of configuration.
> * Does it make sense to consolidate the use cases and bundle the flags in code? E.g., do not have
>   `agw_of ^= EMBEDDED_SGW=1, S6A_OVER_GRPC=1`
>   but have `OGW_OF` in code to do the same and then setting the flag via `per_file_copt`.
> * nit: is it `agw_off`?

Yes these are all great points!

I was wondering about the first case as well. I was wondering about the first point as well. I think the latest value takes, but we will have to test this. Also agree about the second point, I would want to make this as clear to the user as possible here. Only the supported options should be configurable. (So maybe we can just add the --per_file_copt settings directly under the config)

Not exactly sure what you mean by the third point?

It is `agw_of` :D I think it stands for open flow. See https://github.com/magma/magma/pull/11615

nstng:
> Not exactly sure what you mean by the third point?

There seems to be a use case for `agw_of ^= EMBEDDED_SGW=1, S6A_OVER_GRPC=1`. If we exchange now all pre-compiler statements from `#if EMBEDDED_SGW` and `#if S6A_OVER_GRPC` to `#if AGW_OF` then we can model this feature by a one-liner
`build:sgw_of --per_file_copt=^lte/gateway/c/core/.*$@-DSGW_OF=1`

imho this is just a trade off between 
**A:** complex pre compiler set-up and
**B:** less complex bazelrc config and avoiding the usage of pointless combinations of flags (or do all combinations make sense in some use case?)

themarwhal:
> > Not exactly sure what you mean by the third point?
> 
> 
> 
> There seems to be a use case for `agw_of ^= EMBEDDED_SGW=1, S6A_OVER_GRPC=1`. If we exchange now all pre-compiler statements from `#if EMBEDDED_SGW` and `#if S6A_OVER_GRPC` to `#if AGW_OF` then we can model this feature by a one-liner
> 
> `build:sgw_of --per_file_copt=^lte/gateway/c/core/.*$@-DSGW_OF=1`
> 
> 
> 
> imho this is just a trade off between 
> 
> **A:** complex pre compiler set-up and
> 
> **B:** less complex bazelrc config and avoiding the usage of pointless combinations of flags (or do all combinations make sense in some use case?)

Ah I see what you mean! This is probably something that is worth asking the MME team!

themarwhal:
@nstng Modified the config so that we only support the two options: agw_of and mme_oai. I've also confirmed that passing in --config=mme_oai properly overwrites the variables. 

But you get a warning like this:
```bash
<command-line>: warning: "EMBEDDED_SGW" redefined
<command-line>: note: this is the location of the previous definition
<command-line>: warning: "S6A_OVER_GRPC" redefined
<command-line>: note: this is the location of the previous definition
```

so maybe there is a better way,,, but i do not know yet.

nstng:
> @nstng Modified the config so that we only support the two options: agw_of and mme_oai. I've also confirmed that passing in --config=mme_oai properly overwrites the variables.

Cool!

> so maybe there is a better way,,, but i do not know yet.

* Is the warning maybe a good thing? Creates awareness that the default is overridden.
* If there is no default config, then I assume `mme_oai` is the implied default (both flags false). This would make the warning more justifiable.
* Do you know if this are all features? If yes, then we have `EMBEDDED_SGW <=> S6A_OVER_GRPC`.
* Is this a PoC for discussion, or do you aim for a merge?

themarwhal:
> > @nstng Modified the config so that we only support the two options: agw_of and mme_oai. I've also confirmed that passing in --config=mme_oai properly overwrites the variables.
> 
> Cool!
> 
> > so maybe there is a better way,,, but i do not know yet.
> 
> * Is the warning maybe a good thing? Creates awareness that the default is overridden.
> * If there is no default config, then I assume `mme_oai` is the implied default (both flags false). This would make the warning more justifiable.
> * Do you know if this are all features? If yes, then we have `EMBEDDED_SGW <=> S6A_OVER_GRPC`.
> * Is this a PoC for discussion, or do you aim for a merge?

1. Yes the warning is not terrible, but might be concerning. I would rather have a more descriptive warning :/
2. The default configuration we currently have if for agw_of (both variables set)
3. Yes after this morning's change (https://github.com/magma/magma/pull/12340), these two are the only options. 
4. I just meant it as a POC for now, but if we all agree with an approach I can turn it into a PR.

nstng:
> > * Do you know if this are all features? If yes, then we have `EMBEDDED_SGW <=> S6A_OVER_GRPC`.
>     3. Yes after this morning's change ([chore(mme): remove s6a_fd FEATURE from Makefile #12340](https://github.com/magma/magma/pull/12340)), these two are the only options.

I mean they are equivalent. In pre-compiler statements all `S6A_OVER_GRPC` can be set to `EMBEDDED_SGW` (or to a new more fitting name). This is of course out of scope here - just an observation.

.bazelrc
```diff
@@ -31,6 +31,12 @@ build:devcontainer --disk_cache=/workspaces/magma/.bazel-cache
 common:devcontainer --repository_cache=/workspaces/magma/.bazel-cache-repo
 build:devcontainer --define=folly_so=1
 
+# MME FEATURES (default is agw_of)
+build --config=agw_of
+
+build:agw_of --per_file_copt=^lte/gateway/c/core/.*$@-DEMBEDDED_SGW=1,-DS6A_OVER_GRPC=1
+build:mme_oai --per_file_copt=^lte/gateway/c/core/.*$@-DEMBEDDED_SGW=0,-DS6A_OVER_GRPC=0
```
LKreutzer, line 38: The `--config=agw_of` seems to work for the build and the run looks normal.

But for `--config=mme_oai` we get the following error already during the `bazel build //lte/gateway/c/core:oai_mme --config=mme_oai` on vm and dev-container.

```
ERROR: /home/vagrant/magma/lte/gateway/c/core/BUILD.bazel:30:11: Compiling lte/gateway/c/core/oai/tasks/sgw/pgw_config.c failed: (Exit 1): gcc failed: error executing command /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -g -MD -MF ... (remaining 405 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
lte/gateway/c/core/oai/tasks/sgw/pgw_config.c: In function 'pgw_config_process':
lte/gateway/c/core/oai/tasks/sgw/pgw_config.c:70:3: warning: implicit declaration of function 'async_system_command' [-Wimplicit-function-declaration]
   70 |   async_system_command(TASK_ASYNC_SYSTEM, PGW_ABORT_ON_ERROR,
      |   ^~~~~~~~~~~~~~~~~~~~
lte/gateway/c/core/oai/tasks/sgw/pgw_config.c:169:11: error: 'str_sgi' undeclared (first use in this function)
  169 |           str_sgi);
      |           ^~~~~~~
lte/gateway/c/core/oai/tasks/sgw/pgw_config.c:169:11: note: each undeclared identifier is reported only once for each function it appears in
Target //lte/gateway/c/core:oai_mme failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 15.705s, Critical Path: 4.09s
INFO: 62 processes: 4 internal, 58 linux-sandbox.
FAILED: Build did NOT complete successfully
```

What is the `--config=mme_oai` intended to be used for? With `-DEMBEDDED_SGW=0` I would also expect to see the issue #12233 appear.

themarwhal: Yes, this will need to be addressed :/ It is used for the OAI-Jenkins job, but I do not know other than that. @rdefosse / @lionelgo what is the use case for the flag?

themarwhal: @LKreutzer Adding this as a blocker https://github.com/magma/magma/issues/12355

nstng: Had a quick look:
* `pgw_config.c` with `EMBEDDED_SGW=0` is actually not working for the last 1.5y.
* its imo not failing with make because `EMBEDDED_SGW` also controls the set of build files - see `lte/gateway/c/core/oai/tasks/CMakeLists.txt` for this case (but this is used in other places as well)
* naive solution would be to have different targets :/
* at least this could be a pointer for libraries that can be refactored out of the mme binary target - @electronjoe would you agree? [edit:] not sure anymore, cyclic dependencies might still exist with commonly used libraries

nstng: See https://github.com/magma/magma/pull/12365 for draft PR with more information

PR #12322, opened 2022-03-28
fix(mme): Correct logic of Identity Req retransmission unit test

ssanadhya:

## Summary

As a follow up from  https://github.com/magma/magma/pull/12297, the test case `TestImsiAttachRejectIdentRetxFailure` was testing the logic of Auth Request retransmission, which is already covered in `TestImsiAttachRejectAuthRetxFailure`. 

This change modifies the test to check for Identity Request retransmissions when a Guti based attach is made.

## Test Plan

`bazel test //lte/gateway/c/core/oai/test/mme_app_task:mme_procedures_test`


ardzoht (approved):
lgtm

PR #12330, opened 2022-03-29
fix(state): add missing common:service dependency

jheidbrink:
## Summary

Add a missing dependency on `common:service` target to the state service

## Test Plan
`bazel run orc8r/gateway/python/magma/state` -> make sure it doesn't fail with `ModuleNotFoundError`

## Additional Information


themarwhal (approved):

LKreutzer (approved):
In pairing

electronjoe (approved):

PR #12339, opened 2022-03-29
chore(pipelined): blocking of local ipv6 addresses is supported

### LINKED ISSUE #12074: datapath: Add support for IPv6 local address in access control app.
there is flag to block IP traffic to local IPv4 address, we need similar support for local IPv6 address.
ref: https://github.com/magma/magma/blob/988b8fa15590fa5e0afddce3bd799932d8bddfc7/lte/gateway/configs/pipelined.yml#L53

verification:
1. Add unit test for IPv6 local address access control
2. run unit tests
3. run integ test

nstng:
## Summary
See https://github.com/magma/magma/issues/12074.

Couple of notes and questions @pshelar and @koolzz:
* `access_control.py` uses `netifaces` which is outdated
* please have a detailed look at our test: we replaced the suffix of `fe80::` addresses as these are not static in the vm setup - and it is disproportionate to make them static. Is there a test mechanic we did not see that achieves the same?
* in the ipv4 case all `127.*` addresses are blocked - from our research we found that there is not really a fixed ipv6 block for loopback addresses. For now we only blocked in this case `::1`. What is your expectation here?
* should this feature be extended to the blocking list mechanics for ipv6?

## Test Plan

Mainly `test_access_control.py`.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/access_control.py
+++ b/lte/gateway/python/magma/pipelined/app/access_control.py
@@ -13,6 +13,7 @@ limitations under the License.
 """
 import ipaddress
 from collections import namedtuple
+from enum import Enum
 
 import netifaces
 from magma.pipelined.app.base import ControllerType, MagmaController
@@ -20,7 +21,12 @@ from magma.pipelined.openflow import flows
 from magma.pipelined.openflow.magma_match import MagmaMatch
 from magma.pipelined.openflow.registers import Direction
 from ryu.lib.packet import ether_types
-from ryu.lib.packet.in_proto import IPPROTO_ICMP
+from ryu.lib.packet.in_proto import IPPROTO_ICMP, IPPROTO_ICMPV6
+
+
+class _IpVersion(Enum):
+    IPV4 = 1
+    IPV6 = 2
 
 
 class AccessControlController(MagmaController):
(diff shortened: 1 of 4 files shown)


lte/gateway/python/magma/pipelined/app/access_control.py
```diff
@@ -33,6 +35,10 @@ class AccessControlController(MagmaController):
     allowed tunnels.
     """
 
+    # lightweight enum for ipv4/6 destinction
```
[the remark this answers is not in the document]
Neudrino: More likely a misspelling of `distinction`...

lte/gateway/configs/pipelined.yml_prod
```diff
@@ -48,6 +48,7 @@ default_rule_tag: 0x1
 access_control:
   # Blocks access to all AGW local IPs from UEs.
   block_agw_local_ips: true
+  block_agw_local_ips_ipv6: true
```
pshelar: No need to have separate flag for local blocking, single flag can control both IPv4 and v6 address blocking.

lte/gateway/python/magma/pipelined/tests/snapshots/test_access_control.AccessControlTestLocalIpBlockLTEIpV6.test_blocking_ip_match.snapshot
```diff
@@ -0,0 +1,22 @@
+ cookie=0x0, table=access_control(main_table), n_packets=0, n_bytes=0, priority=100,icmp6,reg1=0x1,ipv6_dst=fe80::linkLocalSuffix actions=resubmit(,middle(main_table)),set_field:0->reg0,set_field:0->reg3
+ cookie=0x0, table=access_control(main_table), n_packets=0, n_bytes=0, priority=100,icmp6,reg1=0x1,ipv6_dst=fe80::linkLocalSuffix actions=resubmit(,middle(main_table)),set_field:0->reg0,set_field:0->reg3
+ cookie=0x0, table=access_control(main_table), n_packets=1, n_bytes=54, priority=10,ipv6,reg1=0x1,ipv6_dst=::1 actions=drop
+ cookie=0x0, table=access_control(main_table), n_packets=1, n_bytes=54, priority=10,ipv6,reg1=0x1,ipv6_dst=2020::10 actions=drop
+ cookie=0x0, table=access_control(main_table), n_packets=0, n_bytes=0, priority=10,ipv6,reg1=0x1,ipv6_dst=fe80::linkLocalSuffix actions=drop
```
pshelar, line 13: why same flows are repeated multiple times?
OVS does not keep duplicate flows in flow table.

nstng: You mean the lines with `fe80::linkLocalSuffix`? These are actually different flows - we replaced the suffix of the (dynamic) ipv6 addresses with `linkLocalSuffix` in the assert (see also note in PR description). Making all interfaces on the vm use static addresses seems disproportionate (there is imho no easy way to add a static address and remove the initial dynamic address - if you know a way, please let us know).

lte/gateway/configs/pipelined.yml_prod
nstng: :+1: will be changed

nstng: change is applied

lte/gateway/python/magma/pipelined/app/access_control.py
```diff
@@ -205,7 +231,29 @@ def _install_local_icmp_flows(self, datapath, ip_network):
...
+            return MagmaMatch(
+                direction=Direction.OUT,
+                eth_type=ether_types.ETH_TYPE_IP,
+                ipv4_dst=(
+                    ip_network.network_address,
+                    ip_network.netmask,
+                ),
+                ip_proto=IPPROTO_ICMP,
+            )
+        return MagmaMatch(
```
pshelar, line 251: better readability you could add 'else' here.

pshelar (approved):

lte/gateway/python/magma/pipelined/tests/snapshots/test_access_control.AccessControlTestLocalIpBlockLTEIpV6.test_blocking_ip_match.snapshot
pshelar: right, there is no interface to make link-local addr static or predictable.

lte/gateway/python/magma/pipelined/app/access_control.py
```diff
@@ -154,29 +159,56 @@ def _add_gre_tun_allow_flow(self, datapath, gre_ip, gre_key):
             resubmit_table=self.next_table,
         )
 
-    def _install_local_ip_blocking_flows(self, datapath):
+    def _install_local_ip_blocking_flows(self, datapath, ip_version):
```
sebathomas: Since we call this method for both versions and it's not used anywhere else, it might be easier to just block both ip versions in this method and not have the parameter.

lte/gateway/python/magma/pipelined/app/access_control.py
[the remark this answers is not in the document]
sebathomas: Could also become a real enum though.

lte/gateway/python/magma/pipelined/tests/test_access_control.py
```diff
@@ -132,11 +105,73 @@ def setUpClass(cls):
             access_control_controller_reference.result()
         cls.testing_controller = testing_controller_reference.result()
 
+    @classmethod
+    def getConfig(cls):
+        raise NotImplementedError
+
+    @classmethod
+    def getMconfig(cls):
```
sebathomas: Normally, function names should be snake case. (Unittest only has some as camel case because it started as a port of junit.)

sebathomas (approved):

nstng:
running integ tests on fork: https://github.com/nstng/magma/actions/runs/2096312195 (/)

PR #12345, opened 2022-03-30
fix: add list Subscribers call to mock HSS

uri200:

## Summary

Added List All Subscribers call to mock HSS. This method was missing to implement subscriber db same methods

Also made HSS `UpdateSubscriber` compatible with `UpdateSubscriber` on subscriberdb

This may be useful for federated integ test

## Test Plan
Added some unit test
```
make precommit
```
## Additional Information


feg/gateway/services/testcore/hss/client_api.go
```diff
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"fmt"
 
+	"magma/orc8r/lib/go/protos"
+
```
emakeev: can you pls change the order? std, github, magma ...

emakeev (approved):

PR #12351, opened 2022-03-30
ci: always pull docker image from remote to prevent out of sync tags

Neudrino:
## Summary

Github action runners are not managed by Magma and might be reused several times. The [docker documentation](https://docs.docker.com/engine/reference/commandline/run/) seems to suggest, that if a image is already present in the local docker registry, it will not be re-fetched from the remote unless explicitly told so. See the option `--pull` with default `missing` [in the docker documentation](https://docs.docker.com/engine/reference/commandline/run/). This can lead to stale images being used. 

To mitigate the issue we need to force pull the first time we use an image within the workflow.
Keep in mind that we do _not_ want to force pull in every following action as this would bring the risk to use different images in one workflow run, when a image tag is updated during that workflow run.

The `addnab/docker-run-action` required an update to version `v3` for this to work.

## Test Plan

- CI
- future observation for unexpected behaviour in CI

## Additional Information

- Thanks to @nstng for finding this and vivid discussion!

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -108,6 +108,7 @@ jobs:
         uses: addnab/docker-run-action@v2
         with:
           image: ${{ env.DEVCONTAINER_IMAGE }}
+          options: --pull always
           # Run a simple echo command to pull down the image. This makes it a bit more clear how much time is spent on building Magma and not pulling down the image.
           run: |
             echo "Pulled the devontainer image!"
@@ -171,6 +172,7 @@ jobs:
         uses: addnab/docker-run-action@v2
         with:
           image: ${{ env.DEVCONTAINER_IMAGE }}
+          options: --pull always
           # Run a simple echo command to pull down the image. This makes it a bit more clear how much time is spent on building Magma and not pulling down the image.
           run: |
             echo "Pulled the devontainer image!"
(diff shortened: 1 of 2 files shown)


electronjoe:
Looks like the OWNERS masks are confusing this GitHub workflow for a bazel BUILD file? Hence the review is only `approvers-bazel`.

.github/workflows/bazel.yml
```diff
@@ -184,7 +186,7 @@ jobs:
             cd /workspaces/magma
             bazel test //... --test_output=errors --cache_test_results=no
       - name: Run `bazel run //:check_starlark_format`
-        uses: addnab/docker-run-action@v2
+        uses: addnab/docker-run-action@v3
```
nstng, line 189: Should pull always be added here as well?
I assume it needs only to be added to the very first use of the image so it will get updated in the local repository - but for consistency I recommend to add it in every step where the image is used.

nstng (changes requested):

nstng:
Tested extensively on #12271. Image was updated correctly :+1: . We should track somewhere that this should be changed for all workflows that use the latest devcontainer image.

.github/workflows/bazel.yml
Neudrino: No, as detailed in the PR description.

nstng:
> Looks like the OWNERS masks are confusing this GitHub workflow for a bazel BUILD file? Hence the review is only `approvers-bazel`.

This file was added on purpose to `approvers-bazel` because it only contains Bazel related builds.

.github/workflows/bazel.yml
nstng: Ah, I did not see that it was only in the first for each step - ok.

nstng (approved):

PR #12353, opened 2022-03-30
chore(ci): create POC workflow to run integ tests on bazel built MME service

LKreutzer:

## Summary
- Create POC GH workflow to run integ tests on bazel-built MME service
- Note: In the current state we run into a time-out because the `bazel build //lte/gateway/c/core:oai_mme` takes too long without caching.
  -  The build takes 2-4h **without any caching**, it seems that this is mainly due to building external targets.
  - Note: Currently we skip `_run_sudo_python_unit_tests` to save some time.

## Additional Information

### DIFF
--- a/.github/workflows/lte-integ-test-bazel.yml
+++ b/.github/workflows/lte-integ-test-bazel.yml
@@ -1,0 +1,109 @@
+---
+
+name: LTE integ test bazel
+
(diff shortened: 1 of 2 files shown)


LKreutzer:
Replaced by extended bazel LTE integ tests in #12963

PR #12360, opened 2022-03-31
fix(dp): Unregister CBSD on SAS response 105

xbend:
Fixed an issue where a CBSD state remained "registered" when SAS
returned a response with reponseCode 105.
Response code 105 means the CBSD got unregistered in SAS and should
consider itself "Unregistered".
This PR changes the CBSD state to "unregistered" when any SAS response
is a 105 code.

## Summary

* Set CBSD state to "Unregistered" on each SAS response that has the responseCode 105.

## Additional Information

### DIFF
--- a/dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
+++ b/dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
@@ -52,24 +52,32 @@ def process_registration_response(obj: ResponseDBProcessor, response: DBResponse
 
     cbsd_id = response.payload.get("cbsdId", None)
     if response.response_code == ResponseCodes.DEREGISTER.value:
-        _terminate_all_grants_from_response(response, session)
+        _unregister_cbsd(response, session)
     elif response.response_code == ResponseCodes.SUCCESS.value and cbsd_id:
         payload = response.request.payload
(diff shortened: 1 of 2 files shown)


jkmar (approved):
One question:
Does that mean SAS can respond with 105 code on any type of request (not just Deregistration)?
I also wonder if there are any certification tests for that

xbend:
@jkmar 
> One question: Does that mean SAS can respond with 105 code on any type of request (not just Deregistration)? I also wonder if there are any certification tests for that

According to WINNF-TS-0016, error codes 1XX are general errors related to the SAS-CBSD protocol. Documentation does not explicitly specify that those errors can occur on any request type, but that's the assumption - as SAS can internally deregister a CBSD at any time. Any CBSD that's communicating with SAS that has some sort of state, can issue any request according to their state machine, and SAS should/will respond exactly with error 105 indicating that the CBSD has been deregistered.

FreedomFi environment got this error initially on a heartbeat response for a CBSD, and then got the same error subsequently on each spectrum inquiry that followed after.

Certification testing documentation WINNF-TS-0122 has only 1 test case `WINNF.FT.C.HBT.3` which targets handling of response code 105 for hearbeat requests, however section `6.7.1` about CBSD deregistration process says:
```
Deregistrations triggered by the SAS in a response message with a responseCode of
105 are covered in other test cases.
```

PR #12365, opened 2022-03-31
chore(mme): mme binary is split into EMBEDDED_SGW 1 and 0

nstng:

## Summary

This is currently only a PoC.

### Problem
* mme builds for features `agw_of` (`EMBEDDED_SGW=1`, `S6A_OVER_GRPC=1`) and `mme_oai` (`EMBEDDED_SGW=0`, `DS6A_OVER_GRPC=0`) are actually build with differing sets of source files.
* This is set via conditionally adding sub-directories and single files in `CMakeLists.txt` files.
* There are files that will not even compile if added. Example: `pgw_config.c` is not added in a make build if `EMBEDDED_SGW=0`. And an included `#if (!EMBEDDED_SGW)` is damaged.
* In general it might be a good idea to build artifacts only with needed files.

### Approach Here
* Only add source files needed for feature.

### Status
* Both libraries and binaries build locally, all tests are green locally.
* There is only one test that seems to need the `EMBEDDED_SGW=0` library -> `gtpv2_fteid_test` (makes sense from the `EMBEDDED_SGW` description "S11/GTPV2-C interface not present")
* Can this be solved more clean via Bazel `select`?
* Unclear if all `CMakeLists.txt` configurations are migrated correctly - this is a fast created PoC for now.
* Unclear if both targets need all header files - some that should be conditional are needed for both targets.

## Additional Information


LKreutzer (commented):
Building works for both targets and the run on `agw_of` looks normal. Running the `mme_oai` target with bazel stops without giving a meaningful reason. Running the binary directly with ` bazel-bin/lte/gateway/c/core/mme_oai  -c /var/opt/magma/tmp/mme.conf -s /var/opt/magma/tmp/spgw.conf` I still find the `Illegal instruction (core dumped)` SEGV from issue #12233 at `intertask_interface.c:215:5`, which is included in `MME_SRCS`.

nstng:
> Building works for both targets and the run on `agw_of` looks normal. Running the `mme_oai` target with bazel stops without giving a meaningful reason. Running the binary directly with ` bazel-bin/lte/gateway/c/core/mme_oai -c /var/opt/magma/tmp/mme.conf -s /var/opt/magma/tmp/spgw.conf` I still find the `Illegal instruction (core dumped)` SEGV from issue #12233 at `intertask_interface.c:215:5`, which is included in `MME_SRCS`.

@LKreutzer we see the same behavior with the artifact created by make. But it looks like this behavior is due to permission problems. If the artifact is executed with `sudo` then the error does not occur (and the bazel and make artifact produce logging almost identical - not much though).

nstng:
closed in the more up-to-date PoC #13058

PR #12369, opened 2022-04-01
feat(ci): Docker promte script for artifactory

ShubhamTatvamasi:

## Additional Information

### DIFF
--- a/orc8r/tools/docker/promote.sh
+++ b/orc8r/tools/docker/promote.sh
@@ -10,7 +10,7 @@ declare -A repositories=(
   [feg]="gateway_go gateway_python"
   [orc8r]="controller magmalte nginx"
   [agw]="agw_gateway_c agw_gateway_python ghz_gateway_c ghz_gateway_python"
-  [cwf]="cwag_go gateway_go gateway_pipelined gateway_python gateway_sessiond" # operator
+  [cwf]="cwag_go gateway_go gateway_pipelined gateway_python gateway_sessiond operator"
 )
 
 for repo in ${!repositories[@]}; do
--- a/orc8r/tools/docker/promote.sh
+++ b/orc8r/tools/docker/promote.sh
@@ -3,12 +3,12 @@
 set -ex
 
 MAGMA_TAG=1.7
-NEW_MAGMA_TAG=1.7.0
+NEW_MAGMA_TAG=1.7.1
 MAGMA_ARTIFACTORY=artifactory.magmacore.org
 
 declare -A repositories=(
-  [feg]="gateway_go gateway_python"
   [orc8r]="controller magmalte nginx"
+  [feg]="gateway_go gateway_python"
   [agw]="agw_gateway_c agw_gateway_python ghz_gateway_c ghz_gateway_python"
   [cwf]="cwag_go gateway_go gateway_pipelined gateway_python gateway_sessiond operator"
 )
--- a/orc8r/tools/docker/promote.sh
+++ b/orc8r/tools/docker/promote.sh
@@ -15,13 +15,28 @@ declare -A repositories=(
 
 for repo in ${!repositories[@]}; do
   for image in ${repositories[${repo}]}; do
+
+    # Change docker URL to Artifactory
+    sed -i "s/docker/${repo}-prod/g" ~/.docker/config.json
+
     # Pull docker image from test registry
     docker pull ${repo}-test.${MAGMA_ARTIFACTORY}/${image}:${MAGMA_TAG}
 
     # Tag docker image with new tag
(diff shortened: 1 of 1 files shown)


orc8r/tools/docker/promote.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
+set -ex
+
+MAGMA_TAG=1.7
+NEW_MAGMA_TAG=1.7.0
```
maxhbr: another place where we need to bump versions ;)

orc8r/tools/docker/promote.sh
```diff
@@ -0,0 +1,42 @@
...
+  [orc8r]="controller magmalte nginx"
+  [agw]="agw_gateway_c agw_gateway_python ghz_gateway_c ghz_gateway_python"
+  [cwf]="cwag_go gateway_go gateway_pipelined gateway_python gateway_sessiond operator"
+)
+
+for repo in ${!repositories[@]}; do
+  for image in ${repositories[${repo}]}; do
+
+    # Change docker URL to Artifactory
+    sed -i "s/docker/${repo}-prod/g" ~/.docker/config.json
```
maxhbr, line 20: that can be pulled out above line 17. And is very stateful. Might be better to write temporary configurations and pass `--config=...`.

This also feels like a very unsafe operation, with a lot of unwanted changes

orc8r/tools/docker/promote.sh
```diff
@@ -0,0 +1,42 @@
...
+    docker push ${repo}-prod.${MAGMA_ARTIFACTORY}/${image}:${NEW_MAGMA_TAG}
+    docker push ${repo}-prod.${MAGMA_ARTIFACTORY}/${image}:latest
+
+    # Remove uploaded image
+    docker rmi ${repo}-test.${MAGMA_ARTIFACTORY}/${image}:${MAGMA_TAG}
+    docker rmi ${repo}-prod.${MAGMA_ARTIFACTORY}/${image}:${NEW_MAGMA_TAG}
+    docker rmi ${repo}-prod.${MAGMA_ARTIFACTORY}/${image}:latest
+
+    # Change docker URL back to docker
+    sed -i "s/${repo}-prod/docker/g" ~/.docker/config.json
```
maxhbr, line 39: If any intermediate command times out or fails, the config will be left in an unexpected state, and this script will behave unexpected in the next run.

maxhbr (changes requested):
instead of using `sed -i` on the actual config, it should use a temporary one so that in the error case no modified configuration is left.

Also the `docker rmi` calls should run after error to clean up.

orc8r/tools/docker/promote.sh
ShubhamTatvamasi: You are right @maxhbr, let me look for a fix for this.

orc8r/tools/docker/promote.sh
ShubhamTatvamasi: @maxhbr so we are actually running this job in GitHub actions, so even if it fails, docker config inside runner is not relevant for us. What do you think we should do in this scenario?
https://github.com/ShubhamTatvamasi/magma-artifactory-docker-promote/runs/5849037781

orc8r/tools/docker/promote.sh
ShubhamTatvamasi: Version updated to 1.7.1

orc8r/tools/docker/promote.sh
maxhbr: This adds a script that looks user callable and it does not add the GH Action.

Maybe it should be moved to a different location if it is GH Action specific, e.g. `/.github/workflows/scripts`

orc8r/tools/docker/promote.sh
maxhbr: maybe something like:
```suggestion
    tmpconfig="$(mktemp -d)"
    sed "s/docker/${repo}-prod/g" ~/.docker/config.json > "$tmpconfig/config.json"
```
and then use the ` --config=$tmpconfig...` flag on every docker call.


Also the `sed`, replacing every occurrence of `docker` in the docker config feels way to broad and it should be replaced by a more stable and less error prone pattern. Or maybe even better would be to not use the default config but have it as a heredoc in this script.

maxhbr (changes requested):

Neudrino (commented):
One might also want to address the shellcheck linter comments.

orc8r/tools/docker/promote.sh
ShubhamTatvamasi: So we have to deal with 8 docker repos here. I used `sed` command to make the process easy. I understand the tmp file location and everything, but in this script we are not adding config.

ShubhamTatvamasi:
Closing the PR, as the code will change with every release. I will share the current code with the Release team.

PR #12370, opened 2022-04-01
chore(mme): redundant EMBEDDED_SGW precompiler statements are cleaned up

nstng:

## Summary

This change removes redundant precompile statements using `EMBEDDED_SGW`. These are redundant because the mme build via make does not use certain source files depending on the value of `EMBEDDED_SGW`.

See https://github.com/magma/magma/blob/81bfb14b911cd5665b72bacf4020ce96fb0db6c9/lte/gateway/c/core/oai/tasks/CMakeLists.txt#L16 the sub-directory `oai/tasks/sgw` is included iff `EMBEDDED_SGW=True`.

Why? Currently, with Bazel, mme is build with all sources. If we want to keep this concept then the change is necessary because at least `pgw_config.c` does not compile with `EMBEDDED_SGW=False`.

## Test Plan

* CI
* optional: build with Bazel with `-DS6A_OVER_GRPC=False -DEMBEDDED_SGW=False` and `-DS6A_OVER_GRPC=True -DEMBEDDED_SGW=True`

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/tasks/sgw/pgw_config.c
+++ b/lte/gateway/c/core/oai/tasks/sgw/pgw_config.c
@@ -66,18 +66,6 @@ void pgw_config_init(pgw_config_t* config_pP) {
 
 //------------------------------------------------------------------------------
 status_code_e pgw_config_process(pgw_config_t* config_pP) {
-#if (!EMBEDDED_SGW)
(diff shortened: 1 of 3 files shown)


ssanadhya (approved):
Lgtm, thanks for the cleanup!

PR #12374, opened 2022-04-01
fix(dp): Unregister CBSD on SAS reponse 103

xbend:
Fixed an issue where SAS returns responseCode 103 (INVALID_VALUE) for
cbsdId parameter for CBSDs that got deregistered on SAS side, and CBSD
remains stuck on Domain Proxy side in registered state, as SAS never
reported a DEREGISTER response code prior.

reponseCode 103 handing with "cbsdId" as the invalid parameter is now
made identical with responseCode 105 handling - CBSD gets
unregistered on Domain Proxy side.

## Summary

* Set CBSD state to "Unregistered" when spectrum inquiry or grant or hearbeat or relinquishment requests get a SAS response with responseCode 103 and responseData object contains "cbsdId" as the invalid paramterer name
* Fixed unit tests 

## Additional Information

### DIFF
--- a/dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
+++ b/dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
@@ -13,7 +13,7 @@ limitations under the License.
 
 import logging
 from datetime import datetime
-from typing import Dict, Optional
+from typing import Callable, Dict, Optional
 
 from magma.configuration_controller.custom_types.custom_types import DBResponse
 from magma.configuration_controller.response_processor.response_db_processor import (
@@ -35,6 +35,39 @@ CHANNEL_TYPE = "channelType"
 OPERATION_PARAM = "operationParam"
 
 
+def unregister_cbsd_on_response_condition(process_response_func) -> Callable:
+    """
+    Unregister a CBSD on specific SAS response code conditions.
+
+    This decorator is applied to any process response functions
+    which should react to response codes that require Domain Proxy
+    to internally unregister the CBSD.
+
+    Currently a CBSD should be marked as unregistered on Domain Proxy if:
+    * SAS returns a response with responseCode 105 (ResponseCodes.DEREGISTER)
+    * SAS returns a response with responseCode 103 (ResponseCodes.INVALID_VALUE)
+      and responseData has "cbsdId" listed as the INVALID_VALUE parameter
+
+    Parameters:
+        process_response_func: Response processing function
+
+    Returns:
+        response processing function wrapper
+    """
+    def process_response_wrapper(obj: ResponseDBProcessor, response: DBResponse, session: Session) -> None:
+        if any([
+            response.response_code == ResponseCodes.DEREGISTER.value,
+            _is_response_invalid_value_cbsd_id(response),
+        ]):
+            logger.info(f'SAS {response.payload} implies CBSD immedaite unregistration')
+            _unregister_cbsd(response, session)
+            return
+        process_response_func(obj, response, session)
+
+    return process_response_wrapper
+
+
+@unregister_cbsd_on_response_condition
 def process_registration_response(obj: ResponseDBProcessor, response: DBResponse, session: Session) -> None:
     """
     Process registration response
@@ -46,11 +79,11 @@ def process_registration_response(obj: ResponseDBProcessor, response: DBResponse
     """
 
     cbsd_id = response.payload.get("cbsdId", None)
-    if response.response_code == ResponseCodes.DEREGISTER.value:
-        _unregister_cbsd(response, session)
-    elif response.response_code == ResponseCodes.SUCCESS.value and cbsd_id:
+    if response.response_code == ResponseCodes.SUCCESS.value and cbsd_id:
         payload = response.request.payload
         cbsd = _find_cbsd_from_request(session, payload)
+        if not cbsd:
+            return
         cbsd.cbsd_id = cbsd_id
         _change_cbsd_state(cbsd, session, CbsdStates.REGISTERED.value)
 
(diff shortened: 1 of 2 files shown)


dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
```diff
@@ -91,7 +91,10 @@ def process_spectrum_inquiry_response(obj: ResponseDBProcessor, response: DBResp
         session: Database session
     """
 
-    if response.response_code == ResponseCodes.DEREGISTER.value:
+    if any([
```
jkmar: It looks like this part is duplicated in multiple functions (it appears at the beginning of almost all requests) can we extract this to a function or maybe even some decorator?

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
```diff
@@ -63,11 +63,11 @@ def process_registration_response(obj: ResponseDBProcessor, response: DBResponse
 def _find_cbsd_from_request(session: Session, payload: Dict) -> DBCbsd:
     if "cbsdSerialNumber" in payload:
         return session.query(DBCbsd).filter(
-            DBCbsd.cbsd_serial_number == payload["cbsdSerialNumber"],
+            DBCbsd.cbsd_serial_number == payload.get("cbsdSerialNumber"),
         ).scalar()
     if "cbsdId" in payload:
         return session.query(DBCbsd).filter(
-            DBCbsd.cbsd_id == payload["cbsdId"],
+            DBCbsd.cbsd_id == payload.get("cbsdId"),
```
jkmar, line 98: If there is no `cbsdId` in payload this filter will match **all** unregistered cbsd? Is that expected? I'm not sure, since we are returning scalar as if we expected only one result.

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
```diff
@@ -434,19 +444,24 @@ def _create_db_requests(
...
+            sas_response_code=response_code,
+            sas_response_data=response_data,
         )
-        return self._prepare_response_from_payload(req_type, response_payload, response_code)
+        return self._prepare_response_from_payload(req_type, response_payload, response_code, response_data)
 
     @staticmethod
-    def _prepare_response_from_payload(req_type, response_payload, response_code=None):
-        if response_code is not None:
+    def _prepare_response_from_payload(req_type, response_payload, response_code=0, response_data=None):
```
jkmar: Why we changed response code's default value from `None` to `0`? It looks like it changes semantic of this functions, since we check whether response code is `None`, so for callers which did not specify this parameter, the behavior is changed

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
```diff
@@ -348,21 +348,31 @@ def test_channel_params_set_on_grant_response(self):
         self.assertEqual(max_eirp, grant.max_eirp)
 
     @parameterized.expand([
-        (REGISTRATION_REQ, registration_requests),
-        (SPECTRUM_INQ_REQ, spectrum_inquiry_requests),
-        (GRANT_REQ, grant_requests),
-        (HEARTBEAT_REQ, heartbeat_requests),
-        (RELINQUISHMENT_REQ, relinquishment_requests),
-        (DEREGISTRATION_REQ, deregistration_requests),
+        (REGISTRATION_REQ, registration_requests, ResponseCodes.DEREGISTER.value, None, CbsdStates.UNREGISTERED.value),
```
jkmar: I'm wondering if we could somehow merge first to parameters, since the first one seems to always imply the second one, that way we would have 4 parameters instead of 5 and be more readable

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
xbend: Before constructing the query and building the cbsd filter, we are checking if the JSON payload contains the keys.
I can add failsafes for cases where the request is inavlid (does not contain neither of the keys)

xbend: Added warning log and made sure that callers are checking if this function returns a db entity.

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
xbend: Redone function call logic.
`_prepare_response_from_payload` only builds a http `requests` object from prepared payload.
Any substitutions to the payload are done (if needed) are/should be done by the direct caller.

Only a few test functions use this directly, rather than building generic responses from db_requests. And those test functions currently used fixed fixtures and not templates.

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
xbend: Created a simple map that links request types with fixtures.
All callers retrieve the fixture from the map based on request type key.

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
```diff
@@ -1,4 +1,5 @@
 from typing import Dict
+from flask import request
```
jkmar: flask?

jkmar (approved):

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
xbend: Removing. Good spot. That was some IDE shenanigans.

PR #12380, opened 2022-04-04
test(mme): Running time unittest

thhung:
## Summary

Adding the report of running time for each testcases. 

## Test Plan

make test_oai_runtime

## Additional Information


themarwhal:
Just a heads up that we support a similar functionality with the Bazel unit test support. :)

Running a unit test target will by default generate a test.xml file in $MAGMA_ROOT/bazel-testlogs .
For example,
```bash
bazel test //lte/gateway/c/core/...
# Results in
root@e320c8434361:/magma# find bazel-testlogs/ -name "test.xml"
bazel-testlogs/lte/gateway/c/core/oai/test/mme_app_task/mme_procedures_network_initiated_test/test.xml
bazel-testlogs/lte/gateway/c/core/oai/test/mme_app_task/mme_procedures_mobile_reachability_timer_test/test.xml
bazel-testlogs/lte/gateway/c/core/oai/test/mme_app_task/mme_procedures_test/test.xml
bazel-testlogs/lte/gateway/c/core/oai/test/mme_app_task/mme_app_config_test/test.xml
bazel-testlogs/lte/gateway/c/core/oai/test/mme_app_task/mme_procedures_x2_handover_test/test.xml
bazel-testlogs/lte/gateway/c/core/oai/test/mme_app_task/mme_app_ue_context_test/test.xml
bazel-testlogs/lte/gateway/c/core/oai/test/mme_app_task/mme_app_emm_encode_decode_test/test.xml
bazel-testlogs/lte/gateway/c/core/oai/test/mme_app_task/eps_quality_of_service_test/test.xml
bazel-testlogs/lte/gateway/c/core/oai/test/mme_app_task/mme_procedures_nas_timer_test/test.xml
bazel-testlogs/lte/gateway/c/core/oai/test/mme_app_task/mme_app_nas_encode_decode_test/test.xml
bazel-testlogs/lte/gateway/c/core/oai/test/mme_app_task/sgw_config_test/test.xml
...
```

where a single test.xml looks like this:
```bash
root@e320c8434361:/magma# cat bazel-testlogs/lte/gateway/c/core/oai/test/lib/lib_bstr_test/test.xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites tests="5" failures="0" disabled="0" errors="0" time="0" timestamp="2022-04-04T17:50:16" name="AllTests">
  <testsuite name="BstrTest" tests="5" failures="0" disabled="0" errors="0" time="0" timestamp="2022-04-04T17:50:16">
    <testcase name="TestBstrchrp" status="run" result="completed" time="0" timestamp="2022-04-04T17:50:16" classname="BstrTest" />
    <testcase name="TestBtrimws" status="run" result="completed" time="0" timestamp="2022-04-04T17:50:16" classname="BstrTest" />
    <testcase name="TestBdelete" status="run" result="completed" time="0" timestamp="2022-04-04T17:50:16" classname="BstrTest" />
    <testcase name="TestBiseqcstrcaseless" status="run" result="completed" time="0" timestamp="2022-04-04T17:50:16" classname="BstrTest" />
    <testcase name="TestBiseqcaselessblk" status="run" result="completed" time="0" timestamp="2022-04-04T17:50:16" classname="BstrTest" />
  </testsuite>
</testsuites>
```

The CI job for Bazel unit tests uploads them all to GitHub action like this:
https://github.com/magma/magma/blob/master/.github/workflows/agw-workflow.yml#L204

thhung:
Thanks @themarwhal for a quick review. Indeed that there is a big overlap between our tasks. If we plan to support Bazel in the future, I think we don't need this task anymore. 

@ardzoht Shall we close this PR?

themarwhal:
> Thanks @themarwhal for a quick review. Indeed that there is a big overlap between our tasks. If we plan to support Bazel in the future, I think we don't need this task anymore.
> 
> @ardzoht Shall we close this PR?

If you want a single XML file with all MME unit test results, we might need some scripting to combine all the individual XML files. So maybe the script you wrote here would be useful for the Bazel case as well!

ardzoht:
Thanks @themarwhal for the bazel context, we wanted to see if this addition would be useful for both CMake and Bazel testing. @thhung As Marie mentions, can we adapt this PR to support a final XML file report combining the individual ones?

thhung:
Hi, 
I think this is a good idea. We can combine multiple XML files into a single one. I will adapt this PR.

lte/gateway/python/scripts/runtime_report.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python3
+
+"""
+Copyright 2020 The Magma Authors.
```
themarwhal:
```suggestion
Copyright 2022 The Magma Authors.
```

lte/gateway/python/scripts/runtime_report.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description="Merging all the .xml unittest reports into a single file")
+    parser.add_argument('-i', '--input', nargs='+', help='<Required> List of xml files', required=True)
+    parser.add_argument('-o', '--output', help='Output of xml report', required=False)
+    args = parser.parse_args()
+
+    print(f"Value of env variable GTEST_OUTPUT = {os.environ['GTEST_OUTPUT']}")
```
themarwhal: what is this for?

lte/gateway/python/scripts/runtime_report.py
```diff
@@ -0,0 +1,88 @@
...
+    print(f"Processing test reports at :")
+    pprint(args.input)
```
themarwhal: With the current state of input being a long list of files, it might be more legible if you print line by line

lte/gateway/python/scripts/runtime_report.py
```diff
@@ -0,0 +1,88 @@
...
+    tree = ET.ElementTree()
+    tree._setroot(testsuites_node)
+    tree.write(output_path)
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description="Merging all the .xml unittest reports into a single file")
+    parser.add_argument('-i', '--input', nargs='+', help='<Required> List of xml files', required=True)
+    parser.add_argument('-o', '--output', help='Output of xml report', required=False)
```
themarwhal: Can you set a default value here?

lte/gateway/python/scripts/runtime_report.py
```diff
@@ -0,0 +1,88 @@
...
+    args = parser.parse_args()
+
+    print(f"Value of env variable GTEST_OUTPUT = {os.environ['GTEST_OUTPUT']}")
+    print(f"Processing test reports at :")
+    pprint(args.input)
+    print("=" * 50)
+
+    output_path = args.output if args.output else "./report_all_tests.xml"
```
themarwhal: Up to you, but I've always defaulted to using $MAGMA_ROOT for these things so that the output shows up in the same place regardless of how the script is run. If you don't want to do that, it might be good to log the abosolute path so it's clear to users.

lte/gateway/python/scripts/runtime_report.py
```diff
@@ -0,0 +1,88 @@
...
+
+    tree = ET.ElementTree()
+    tree._setroot(testsuites_node)
+    tree.write(output_path)
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description="Merging all the .xml unittest reports into a single file")
+    parser.add_argument('-i', '--input', nargs='+', help='<Required> List of xml files', required=True)
```
themarwhal: I think it might be more useful to have the input be a regex to point to multiple XML files, rather than a list of files, but I'll let @ardzoht comment on this :)

nstng:
Hi, maybe I am wrong, but this sounds like a standard problem. If this is junit xml format, should we use something off-the-shelf? E.g., quick google: https://github.com/bhovhannes/junit-report-merger

lte/gateway/python/scripts/runtime_report.py
thhung: It is redundant, actually. It shows where the reports of the Gtest are saved. I will remove it.

lte/gateway/python/scripts/runtime_report.py
thhung: Sure. Let me do that.

lte/gateway/python/scripts/runtime_report.py
thhung: It's true, given the number of test we have. I will remove it.

lte/gateway/python/scripts/runtime_report.py
thhung: Thanks for the suggestion. My plan will be:
- if $MAGMA_ROOT is existed,  I will default to using $MAGMA_ROOT. 
- Otherwise, default to a given path. Of course I will log the absolute path.

lte/gateway/python/scripts/runtime_report.py
thhung: I don't know how complex the regex, we plan to point to multiple XML files, is. But if we have a regex that the shell expansion can handle, this will work, for example: `/tmp/magma_oai/*.xml` works for my test. When I wrote this, I mean gtest will put all the reports in a single folder or we run this script from Makefile. So, maybe let's decide on the input format and I can continue from there?

ardzoht:
> Hi, maybe I am wrong, but this sounds like a standard problem. If this is junit xml format, should we use something off-the-shelf? E.g., quick google: https://github.com/bhovhannes/junit-report-merger

@nstng that library looks like it adds newer dependencies which we don't have on AGW. I agree that this looks like a general problem, however adding a third-party dep to achieve this seems a little overwill to me.

lte/gateway/python/scripts/runtime_report.py
ardzoht: I agree with @themarwhal that for this, passing a regex param seems more useful to pass multiple files as a cmd. @thhung can we adapt the script to accept regex ?

thhung: Hi, thanks for suggestions. I added an option for regex. Please let me know if it makes sense or not.

ardzoht:
@thhung I'd suggest adding a new section on https://github.com/magma/magma/blob/master/docs/readmes/lte/dev_notes.md to include an example on how to run this command on the AGW.

ardzoht (changes requested):
Please add documentation on the script usage.

docs/readmes/lte/dev_notes.md
```diff
@@ -282,3 +282,55 @@ enbs {
 }
 num_enbs: 1
 ```
+
+## Unit test report
```
ardzoht:
```suggestion
## Combining XML Unit test reports
```

docs/readmes/lte/dev_notes.md
```diff
@@ -282,3 +282,55 @@ enbs {
 }
 num_enbs: 1
 ```
+
+## Unit test report 
+
+At the moment of writing, running the unit test targets does not generate the report file by default. To get the report of all unit tests, you have two options.
```
ardzoht:
```suggestion
At the moment of writing, running the unit test targets does not generate the aggregated XML report file by default. To get the report of all unit tests, you have two options.  
```

lte/gateway/Makefile
```diff
@@ -207,6 +207,17 @@ test_c: ## Run all Bazel-ified C/C++ tests
 test_oai: ## Run all OAI-specific tests
 	$(call run_ctest, $(C_BUILD)/core, $(C_BUILD)/core/oai, $(GATEWAY_C_DIR)/core, $(OAI_FLAGS) $(OAI_TEST_FLAGS) $(OAI_NOBENCHMARK_FLAGS), $(OAI_TESTS))
 
+test_oai_runtime: export REPORT_FOLDER=${MAGMA_ROOT}/report/
+test_oai_runtime: export UNITTEST_REPORT_FOLDER=${REPORT_FOLDER}/unittest_report/
+test_oai_runtime: export MERGED_REPORT_FOLDER=${REPORT_FOLDER}/merged_report/
+test_oai_runtime: export GTEST_OUTPUT=xml:${UNITTEST_REPORT_FOLDER}
```
ardzoht, line 213: What is the reason of setting these under `test_oai_runtime`? We should set these at the top of the file as the other variables

ardzoht (changes requested):

lte/gateway/Makefile
thhung: Hi Alex, my purpose is to limit the effect only inside the `test_oai_runtime`. If we put it at the top of the file, it might affect other targets.

themarwhal (approved):

docs/readmes/lte/dev_notes.md
```diff
@@ -282,3 +282,59 @@ enbs {
...
+If you prefer a simple method, you can simply do:
+
+```bash
+cd ${MAGMA_ROOT}/lte/gateway/
+make test_oai_runtime
+```
+
+And you will get the combined report at `${MAGMA_ROOT}/report/merged_report/report_all_tests.xml`
+
+### Option 2: Manual method
```
themarwhal, line 301: What is the benefit of specifying a manual method? As in, why would a developer opt to do this manually?

ardzoht (approved):
lgtm

docs/readmes/lte/dev_notes.md
thhung: Hi @themarwhal, my intention for this manual method is just in case, in the future, a developer would like to make some changes to my script and how to use it, they can use this method as a guideline to make their changes.

PR #12383, opened 2022-04-04
fix(nms): Initially load count of subscribers correctly

### LINKED ISSUE #11929: Subscriber count is not correct in NMS UI
### Your Environment
- **Version:** v1.6.1
- **Affected Component:** NMS
- **Affected Subcomponent:** N/A

### Describe the Issue
 i have provisioned 29 subscribers. When i  see in network it is showing 25 subscribers (may be based on default page size) when i go to next page it is showing 4 subscribers on network. If any one explored or having knowledge on NMS code please let me know the exact behavior or exact fix for the same if it seems as a bug. Sharing screenshots of same for your reference.

**To Reproduce**
1. Go to NMS UI
2. Click on Subscriber and add subscriber (in my case i have uploaded subscribers via .csv file)
3. Scroll up to Network there check subscriber count (it may be displaying based on the default page size)
4. Go to subscriber and go to next row 
5. Come back to Network and check it will show number subscribers from second page/row

**Expected behavior**
It has to show total number of subscribers provisioned irrespective of page/row we are changing.
image
image (1)
image (2)
image (3)
image (4)

sebathomas:
## Summary

Fixes a bug where the number of subscribers shown in the KPI panel depends on the pagination of the subscriber table.

Does not handle changes to the number without reloading NMS.

Fixes #11929.

Thanks @Siddharthlende for investigating!

Done in pairing with @thmsschmitt.

## Test Plan

Set up NMS locally with 98 subscribers.
nmsbug

## Additional Information

### DIFF
--- a/nms/app/components/context/SubscriberContext.js
+++ b/nms/app/components/context/SubscriberContext.js
@@ -35,6 +35,7 @@ state: paginated subscribers
 sessionState: paginated subscribers session state
 metrics: subscriber metrics
 gwSubscriberMap: gateway subscriber map
+totalCount: total count of subscribers
 setState: POST, PUT, DELETE subscriber
 */
 export type SubscriberContextType = {
@@ -43,6 +44,7 @@ export type SubscriberContextType = {
   forbiddenNetworkTypes: {[string]: core_network_types},
   metrics?: {[string]: Metrics},
   gwSubscriberMap: {[gateway_id]: Array<subscriber_id>},
+  totalCount: number,
   setState?: (
     key: string,
     val?: mutable_subscriber | mutable_subscribers,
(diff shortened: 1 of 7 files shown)


sebathomas:
@andreilee Is there a good way to also get automatic updates in? I think it would be possible to do that in `setSubscriberState` in `SubscriberState.js` but that method has a lot of different update paths...

sebathomas:
The broken test was also red on the last few runs in NMS PRs. Might be fixed by https://github.com/magma/magma/pull/12384.

sebathomas:
Rebased.

nms/app/components/context/SubscriberContext.js
```diff
@@ -43,6 +43,7 @@ export type SubscriberContextType = {
   forbiddenNetworkTypes: {[string]: core_network_types},
   metrics?: {[string]: Metrics},
   gwSubscriberMap: {[gateway_id]: Array<subscriber_id>},
+  totalCount: number,
```
andreilee, line 47: nit: update the comments above

andreilee (approved):

PR #12385, opened 2022-04-04
chore(ci): github runners have more space available

nstng:

## Summary

Experimenting with the Bazel mme setup we experienced again out-of-space situations in Bazel workflows. There are github actions that clean up the used ubuntu image e.g. [1] (@themarwhal I think you suggested that one the last time we looked into space issues).
[1] does not work directly in our situation as we use `/dev/root` via docker. [1] removes dotnet, android SDKs and Haskell and this can actually be done manually very easy.

`Filesystem      Size  Used Avail Use% Mounted on`
**before cleanup**
`/dev/root        84G   55G   29G  66% /`
**after cleanup**
`/dev/root        84G   38G   47G  45% /`
**after run (Bazel Build & Test)**
`/dev/root        84G   62G   22G  74% /`

-> ~24G are filled for the run
-> without cleanup currently ~79G/84G would be used
-> with cleanup ~22G are still available after run

**Note** I left a `df -h` at the end of the workflows - this might be helpful for future debugging, but I can remove it if requested.

[1] https://github.com/easimon/maximize-build-space

## Test Plan

CI

## Additional Information

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -143,6 +143,17 @@ jobs:
     name: C/C++ unit tests with Bazel
     runs-on: ubuntu-latest
     steps:
+      - name: Maximize build space
+        shell: bash
+        run: |
+          echo "Available storage before:"
+          df -h
+          echo "Removing /usr/share/dotnet, /usr/local/lib/android, /opt/ghc"
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/ghc
+          echo "Available storage after:"
+          df -h
       - name: Check Out Repo
         # This is necessary for overlays into the Docker container below.
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


themarwhal (approved):
wahoo!

tmdzk (approved):

.github/workflows/agw-workflow.yml
```diff
@@ -213,6 +224,12 @@ jobs:
         with:
           name: Unit Test Results
           path: c-cpp-test-results/
+      - name: Build space left after run
+        shell: bash
+        run: |
+          echo "Available storage:"
+          df -h
+          echo
```
LKreutzer: Nit: What are the empty `echo`s needed for - is a newline needed?

LKreutzer (approved):

.github/workflows/agw-workflow.yml
nstng: No, they are not needed. I used a couple for esthetic reasons - and removed them. forgot about these. removed.

PR #12386, opened 2022-04-04
fix: pin genproto to specific version to avoid envoyproxy/go-control-…

uri200:
…plane update

## Summary

We are using envoy 0.9.4 on Magma. That version seems to be using specific versions of grpc and protobuf module.s 

To update `envoy/go-control-plane` we need to refactor the code because the newest versions have a total different API.

So in order to be able to use some new modules on magma we need to pin genproto version to prevent `envoy/go-control-plane ` to automatically update on go.mod and break. 

Note in the near future this will break and envoy will need to be either updated or placed on a separate module. 

## Test Plan

make precommit

## Additional Information


emakeev (approved):
LGTM

PR #12391, opened 2022-04-04
fix: run FEG on Vagrant VM on federated integ test

uri200:

## Summary

This PR modifies federated integ test fab file to run FEG inside the Magma VM

The reason for that change is that FEG uses host netowrk mode. That mode is not supported properly on MAC so we can't access to FEG services port (so we can send grpc messages from the host). To allow us to send messages and configure grpc services like mock hss, we need to move FEG execution to Vagrant vm, where docker host network mode is supported

Note this will only impact federated integ test.

## Test Plan

`fab test_connectivity`
on federated integ test
## Additional Information

### DIFF
--- a/lte/gateway/python/integ_tests/federated_tests/docker/docker-compose.override.yml
+++ b/lte/gateway/python/integ_tests/federated_tests/docker/docker-compose.override.yml
@@ -16,10 +16,10 @@ services:
 
   control_proxy:
     extra_hosts:
-      - controller.magma.test:127.0.0.1
-      - bootstrapper-controller.magma.test:127.0.0.1
-      #- controller.magma.test:10.0.2.2
-      #- bootstrapper-controller.magma.test:10.0.0.2
+      # Use 127.0.0.1 when running FEG on the host itself
+      # Use 10.0.2.2 when running FEG on Vagrant VM
+      - controller.magma.test:10.0.2.2
+      - bootstrapper-controller.magma.test:10.0.0.2
     command:
       - /bin/bash
       - -c
@@ -37,10 +37,10 @@ services:
       context: ${BUILD_CONTEXT}
       dockerfile: feg/gateway/docker/python/Dockerfile
     extra_hosts:
-      - controller.magma.test:127.0.0.1
-      - bootstrapper-controller.magma.test:127.0.0.1
-      #- controller.magma.test:10.0.2.2
-      #- bootstrapper-controller.magma.test:10.0.2.2
+      # Use 127.0.0.1 when running FEG on the host itself
+      # Use 10.0.2.2 when running FEG on Vagrant VM
+      - controller.magma.test:10.0.2.2
+      - bootstrapper-controller.magma.test:10.0.2.2
     command: python3.8 -m magma.magmad.main
 
 volumes:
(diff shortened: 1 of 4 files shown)


themarwhal (approved):

VinashakAnkitAman:
LGTM

emakeev (approved):

lte/gateway/python/integ_tests/federated_tests/docker/docker-compose.override.yml
```diff
@@ -16,10 +16,10 @@ services:
 
   control_proxy:
     extra_hosts:
-      - controller.magma.test:127.0.0.1
-      - bootstrapper-controller.magma.test:127.0.0.1
-      #- controller.magma.test:10.0.2.2
-      #- bootstrapper-controller.magma.test:10.0.0.2
+      #- controller.magma.test:127.0.0.1
```
ssanadhya: What's the purpose of the commented config here? Can we add a comment explaining the options of using localhost vs 10.0.0.2?

lte/gateway/python/integ_tests/federated_tests/fabfile.py
```diff
@@ -257,10 +290,16 @@ def test_connectivity(timeout=10):
 
     # check FEG-cloud connectivity
     print("\n### Checking FEG-Cloud connectivity ###")
+    vagrant_setup('magma', destroy_vm=False)
+    with cd(feg_docker_integ_test_path_vagrant):
+        run('docker-compose exec magmad checkin_cli.py')
+    """
```
ssanadhya: Remove commented code

orc8r/tools/fab/vagrant.py
```diff
@@ -21,8 +22,14 @@ def __ensure_in_vagrant_dir():
...
     pwd = local('pwd', capture=True)
     if not os.path.isfile(pwd + '/Vagrantfile'):
-        print("Error: Vagrantfile not found. Try executing from fbcode/magma")
-        exit(1)
+        # check if we are on a vagrant subdirectory
+        with cd(pwd):
+            if not local('Vagrant validate', capture=True):
+                print(
+                    "Error: Vagrantfile not found. "
+                    "Try executing from fbcode/magma",
```
ssanadhya: The dir path `fbcode/magma` doesn't exist any more. Is this referring to the MAGMA_ROOT?

uri200: We can remove that message. I guess it was there from before and was never modified

lte/gateway/python/integ_tests/federated_tests/docker/docker-compose.override.yml
uri200: 10.0.0.2 is used when FEG is run locally in the host. 127.0.0.1 when run on Vagrant. 
I can add a note there. But I would rather prefer to leave that comments because the knowledge of those routes is not easy to figure out. Leave them commented will provide context in case we want to switch back the test environment to the host.


Note that config file is just a test file exclusively for federated integ test. 

```
    extra_hosts:
      # Use 127.0.0.1 when running FEG on the host itself 
      # Use 10.0.2.2 when running FEG on Vagrant VM
      - controller.magma.test:10.0.2.2
      - bootstrapper-controller.magma.test:10.0.0.2
```

lte/gateway/python/integ_tests/federated_tests/fabfile.py
uri200: i will divide that function into two so we can execute that command using FEG From Host and FEG from Vagrant

ssanadhya (approved):
Lgtm

PR #12395, opened 2022-04-05
chore(dp): Remove processed requests and responses from database

jkmar:
## Summary

List of changes:
- Requests no longer have state id
- All requests are pending, after processing they are deleted
- Responses are no longer stored in database
- Responses and Request State tables were deleted
- Protocol controller integration tests were disabled


xbend (approved):
LGTM

PR #12399, opened 2022-04-05
feat(nms): Remove Material-table and use material-table/core instead

HannaFar:

## Summary

Remove Material-table and use material-table/core instead

## Additional Information


thmsschmitt:
We are using `filefy` in SubscriberTable, but we didn't include it in our dependencies and now it got removed see https://github.com/magma/magma/runs/5835150695?check_suite_focus=true

nms/packages/magmalte/app/components/ActionTable.js
```diff
@@ -30,7 +30,7 @@ import FilterList from '@material-ui/icons/FilterList';
 import FirstPage from '@material-ui/icons/FirstPage';
 import FormControl from '@material-ui/core/FormControl';
 import LastPage from '@material-ui/icons/LastPage';
-import MaterialTable from 'material-table';
+import MaterialTable from '@material-table/core';
```
thmsschmitt, line 33: Unfortunately we have to drop the `strict-local` above

HannaFar: thanks!

andreilee (approved):
Fix the yarn.lock issue before merging, please

PR #12402, opened 2022-04-05
feat(nms): Merging administration tools into nms

### LINKED ISSUE #11543: [NMS] Merging Administration Tools into NMS
### Problem
When the user wants to manage users, networks, or view audit logs, they have to switch to the Administration Tools from the NMS. The switch between two portals has a few usability problems:
- Users may not be aware of the Portal Switcher. The Administration Tools is hard to discover.
- Users may be confused about which portal they are at, especially for non-power users.
- Switching back and forth between the two portals is inefficient.

### Design proposal
In this new proposal, we are merging the Administration Tools into the NMS. The users, audit logs, and networks pages will be tabs under a parent page called Administration. It can be accessed through Account --> Administration in the navigation menu.

Here's a link to the mockup: [Figma link](https://www.figma.com/file/Ec5jkliFb4i4OYGdI9AkYk/?node-id=3%3A11)

Account
<p align = "center"><em>Access point to Administration</em></p>

Administration
<p align = "center"><em>Administration page</em></p>

thmsschmitt:
## Summary 

Move administration and password settings from separate tab into main UI.

Fixes: #11543

## Details

The administration page has been moved into the network specific UI thus the URL scheme has been changed to `/nms/<networkID>/admin/...`. If the admin page is selected the account menu button will be highlighted as suggested in the mockup. The new design for the button, menu and the account settings page itself are done as a separate PR #12430.
Screenshot 2022-04-07 at 15 46 35

The form to change the password was previously on the admin page but did not have a sidebar entry. Now it is a separate page under `/nms/<networkID>/settings`.
Screenshot 2022-04-07 at 15 46 54

If no network exists one still needs to be able to get to the admin and account settings page. Therefore in this special case the pages can still be reached under `/admin/...` and `/settings` without the network related entries in the sidebar.

Screenshot 2022-04-07 at 15 51 07
Screenshot 2022-04-07 at 15 50 39

The account settings are also available on the organizations page.
Screenshot 2022-04-07 at 15 47 25

## Testplan

Open the administration page as:
- super user with at least one network
- super user without any network

Open the account settings page:
- as super user with at least one network
- as super user without any network
- as regular user with at least one network
- as regular user without any network
- from the organizations page


nms/packages/magmalte/fbc_js_core/ui/context/AppContext.js
```diff
@@ -37,6 +37,7 @@ export type AppContextType = {
   user: User,
   showExpandButton: () => void,
   hideExpandButton: () => void,
+  isOrganizations: boolean,
```
andreilee, line 40: Could you add a comment explaining what this means?

andreilee (approved):
Looks great! Could you just add a comment about `isOrganizations`, before I merge this?

There are some refactoring/cleanup type changes that have been included in the PR here. I would prefer that those types of changes are put in a separate commit (or even PR) whenever possible.

PR #12407, opened 2022-04-06
chore(cwf, dp, feg, orc8r): Update and introduce tools.go files

alexzurbonsen:
## Summary

Update and introduce `tools.go` files. These files are used to track dev dependencies of a module. 

This is a prerequisite of the Go version upgrade #12151 (still WIP currently), since the behavior of `go install` changes in Go >= 1.16 (see https://go.dev/doc/go1.16). If used from within a module, a version suffix (e.g. `go install somepackage@v1.2.3`) has to be added to ignore the module. Without version suffix an entry in `go.mod` file is required. This causes commands like `make tools` to fail with Go >=1.16.

The changes of this PR shouldn't interfere with anything in the current Go version. They only cause dev dependencies to be explicitly tracked in modules, i.e. they are also listed in `go.mod` & `go.sum` files.

**Edit**: There seems to be interference. Probably because dev depencies are not pinned to some version which changes other versions...

## Test Plan

Ran `go mod tidy` in all go modules in magma and iterated this until all changes propagated through interdependent modules (two iterations necessary).  

CI.


alexzurbonsen:
included in #12151

PR #12411, opened 2022-04-06
fix(dp): Fix requests migration

jkmar:
## Summary

Migration generated by alembic was wrong, since it tried to
delete table when it was still referenced as foreign key.


xbend (approved):
LGTM

PR #12412, opened 2022-04-06
fix(ci): operator docker image artifactory push fix master branch

ShubhamTatvamasi:

## Summary

Same operator docker image CI fix.

https://github.com/magma/magma/pull/12377/files
https://github.com/magma/magma/pull/12412/files

## Additional Information

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -677,7 +677,8 @@ jobs:
             echo TAG="${GIT_BRANCH_VERSION:1}" >> $GITHUB_ENV
           fi
       - name: Tag and push to Jfrog Registry
-        if: github.ref == 'refs/heads/master'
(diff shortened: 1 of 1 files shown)


Neudrino (commented):
Is the PR intentionally against `master`?
I would expect for `master` the implementation is correct as is?

ShubhamTatvamasi:
> Is the PR intentionally against `master`? I would expect for `master` the implementation is correct as is?

Yes, we just don't want the same to happen with `v1.8` that why this PR is needed.

Neudrino (approved):
Thanks for the explanation. And the restriction master or release branch will be the one from the workflow itself, such that we still avoid artifacts from PRs being pushed, afaiks.

Neudrino (approved):

tmdzk (approved):

PR #12417, opened 2022-04-07
fix(nms): Fix insecure randomness

spikey979:

fix(nms): Fix insecure randomness

## Summary

Using a cryptographically weak pseudo-random number generator to generate a security-sensitive value, such as a password, makes it easier for an attacker to predict the value. A cryptographically secure pseudo-random number generator should be used instead.

## Additional Information

This is for code scanning alerts:
https://github.com/magma/magma/security/code-scanning/62
https://github.com/magma/magma/security/code-scanning/63
https://github.com/magma/magma/security/code-scanning/64
https://github.com/magma/magma/security/code-scanning/65

### DIFF
--- a/nms/fbc_js_core/auth/express.js
+++ b/nms/fbc_js_core/auth/express.js
@@ -36,6 +36,8 @@ import type {ExpressRequest, ExpressResponse} from 'express';
 import type {FBCNMSRequest} from './access';
 import type {UserType} from '../../fbc_js_core/sequelize_models/models/user';
 
+import crypto from 'crypto';
+
 const logger = logging.getLogger(module);
 const PASSWORD_FOR_LOGGING = '<SECRET>';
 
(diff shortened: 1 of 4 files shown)


thmsschmitt:
Screenshot 2022-04-08 at 13 45 47
The server runs in node. You need to use https://nodejs.org/api/crypto.html instead.

spikey979:
> Screenshot 2022-04-08 at 13 45 47 The server runs in node. You need to use https://nodejs.org/api/crypto.html instead.

If I use crypto.getRandomValues() without "winodw", in CI tests I get error: Cannot resolve name `crypto`. [cannot-resolve-name]

thmsschmitt:
You need to import it: 

```
import crypto from 'crypto';
```

also note that it has a different API then `window.crypto`. See https://nodejs.org/api/crypto.html

thmsschmitt (changes requested):
Sorry that this takes so long.

`crypto.getRandomValues` is only available since node v17.4.0 and we are using v16.14. Also only using 4 bytes isn't great.

I suggest we use something like
`crypto.randomBytes(16).toString('hex')`

thmsschmitt (approved):

spikey979:
@thmsschmitt thanks for helping

andreilee (approved):

thmsschmitt:
@spikey979 I'm waiting for the `ready2merge` label. It would also be good to rebase this once more.

PR #12424, opened 2022-04-07
chore: upgrade Go installation in DevContainer

alexzurbonsen:
## Summary

[_Merge after #12151._]

Upgrade Go to 1.18 in DevContainer.

This is necessary because of #12151.

## Test Plan

Built DevContainer locally.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -62,7 +62,6 @@ RUN echo "Install general purpose packages" && \
         gdb \
         git \
         gnupg \
-        golang \
         libclang-11-dev \
         libgmp3-dev \
         libpcap-dev \
@@ -97,6 +96,13 @@ RUN echo "Install general purpose packages" && \
         update-alternatives --install /usr/bin/clang-apply-replacements clang-apply-replacements /usr/lib/llvm-11/bin/clang-apply-replacements 10
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get install -y clangd-12
+
+# Install golang 1.18
+WORKDIR /usr/local
+RUN curl https://artifactory.magmacore.org/artifactory/generic/go1.18.linux-amd64.tar.gz --remote-name --location && \
+    tar -xzf go1.18.linux-amd64.tar.gz && \
+    rm go1.18.linux-amd64.tar.gz
+ENV PATH=$PATH:/usr/local/go/bin
 
 # Install bazel
 WORKDIR /usr/sbin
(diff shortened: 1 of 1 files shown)


themarwhal (approved):

.devcontainer/Dockerfile
```diff
@@ -98,6 +97,12 @@ RUN echo "Install general purpose packages" && \
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get install -y clangd-12
 
+# Install golang 1.18
+WORKDIR /usr/local
+RUN curl https://artifactory.magmacore.org/artifactory/generic/go1.18.linux-amd64.tar.gz --remote-name --location && \
+    tar -xzf go1.18.linux-amd64.tar.gz && \
+    cp -r go/bin/* /usr/local/bin/
+
```
Neudrino: We should remove the "extraction directory" after use here to keep the image small, imho.

.devcontainer/Dockerfile
```diff
@@ -98,6 +97,12 @@ RUN echo "Install general purpose packages" && \
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get install -y clangd-12
 
+# Install golang 1.18
+WORKDIR /usr/local
+RUN curl https://artifactory.magmacore.org/artifactory/generic/go1.18.linux-amd64.tar.gz --remote-name --location && \
+    tar -xzf go1.18.linux-amd64.tar.gz && \
```
maxhbr, line 103: this leaves `go1.18.linux-amd64.tar.gz` in `/usr/local`, please clean that up.

maxhbr (dismissed):

.devcontainer/Dockerfile
```diff
@@ -98,6 +97,13 @@ RUN echo "Install general purpose packages" && \
...
+    rm go1.18.linux-amd64.tar.gz && \
+    export PATH=$PATH:/usr/local/go/bin
```
maxhbr: This export does not work that way (it is lost after that line), instead you need to write:

```suggestion
ENV PATH=$PATH:/usr/local/go/bin
```

alexzurbonsen: Oh yes, thanks!

.devcontainer/Dockerfile
```diff
@@ -161,7 +167,7 @@ RUN wget --progress=dot:giga https://ftp.gnu.org/gnu/nettle/nettle-2.5.tar.gz &&
     rm -rf gnutls*
 
 ##### Useful for logfile modification e.g. pruning all /magma/... prefix from GCC warning logs
-RUN GOBIN="/usr/bin/" go get github.com/ezekg/xo
+RUN GOBIN="/usr/bin/" go install github.com/ezekg/xo@0f7f076932dd
```
maxhbr, line 170: do we have any process of bumping such pinned packages?

alexzurbonsen: Not as far as I know. But I wondered that as well. Let's think of something then?!
Background: I pinned it here since Hunter advised so for a different module on the go upgrade PR. But before (with go get) there was no pinning in place either and no such process necessary.

.devcontainer/Dockerfile
Neudrino: Same as https://github.com/magma/magma/pull/12424#discussion_r857551932.

.devcontainer/Dockerfile
Neudrino: Closing as same as https://github.com/magma/magma/pull/12424#discussion_r859978511.

Neudrino (approved):

maxhbr (approved):
LGTM

PR #12461, opened 2022-04-13
chore(agw): Migrate all c++ headers to .hpp extension

### LINKED ISSUE #10319: Rename C++ headers from .h to .hpp extension

electronjoe:
# Summary

This PR closes issue #10319 by migrating all Magma cpp headers to `.hpp` extension (except those within the `session_manager` sub-folders as this code base has fallen behind in maintenance of best practices). See the Issue for rationale - TL;DR; many errors in the Magma codebase are made difficult to detect due to our mixed c/c++ and lack of clear header file differentiation (by either directory or file extension).

Examples of impact:

- This migration allows improved static analysis (by analyzers that only support e.g. c++)
- This migration allows improved automation of `#include` directives (e.g. enforcing the correct `extern "C"` annotations)

## Part One: Automated renaming in `/orc8r/gateway/c/`

I authored a Go binary to automate much of the detection, mapping and fixup of `#include` statements for this PR.  The Go binary generally does the following:

1. Form a set of all `<filename>.cpp` files
    1. but omit `session_manager` directory
    2. but omit any `*test*.cpp` filename (audited, these are all tests and so no header to map)
    3. but omit manual blacklist of tests without `*test*.cpp` naming, or top level binaries
2. Form a set of all headers which will be renamed from `header.h` -> `header.hpp` composed of
    1. Any `<filename>.h` file in the repo having `<filename>.cpp` somewhere in the repo (audited for name collisions)
    2. Manual map, where necessary, to cover headers for which the `<filename>.cpp` implementations do not share name
3. Do a `git move` of every rename header file in the set
4. Fixup all `#include "path/to/new/header.h` -> `header.hpp`
5. Fixup all `BUILD.bazel` files impacted by the renames
6. Fixup all `CMakeLists.txt` files impacted by the renames

This PR was achieved by execution of [this Golang script](https://gist.github.com/electronjoe/2ab64a7e64de289b0ef88decb2ad3abf) as follows, from within a GH Codespace / Devcontainer:

```
cd /workspaces/magma/
go run hppify.go -root /workspaces/magma/ -target /workspaces/magma/orc8r/gateway/c/ -dry_run=false
```

## Part Two: manual fixup of `session_manager`

Because Session Manager has not been fixed up with full include paths relative to repo root (an outstanding task) - `hppify.go` is unable to precisely correct the renamed `.h` -> `.hpp` `#include` directives. So I fixed them by hand / in blocks using vscode replace all in files iteratively.

## Part Three: correct CMake build include copies of `*.h`

Various CMake build steps in Magma unfortunately rely on copying include directories to various build directories.  These often copy all `*.h` headers, and now need either to copy all `*.hpp` headers (if that is all that exists now) or copy `*.h` and `*.hpp` headers.

# Implications to outstanding PRs

It is anticipated that the move git semantics will enable clean rebasing for any outstanding PRs that touched the now-renamed `.hpp` files.

# Test Plan

Validated I did not break compile or test for either Bazel or CMake within the devcontainer environment (and CI).

Within the devcontainer validated success of the following:

```
bazel build //orc8r/gateway/c/...
bazel test //orc8r/gateway/c/...
bazel build //lte/gateway/c/...
bazel test //lte/gateway/c/...
cd lte/gateway
make build_oai
make test_oai
```


themarwhal (approved):
lgtm! Double checked that the moved header files preserve git blame history: https://github.com/magma/magma/blame/c04e9500ca138311463616f63c7d24b478abc2d4/lte/gateway/c/connection_tracker/src/PacketGenerator.hpp

ssanadhya (approved):
Thanks for the clean up @electronjoe ! Two more fixes to include:
1. state_manager.h is also a C++ header and can be switched to `.hpp`
2. Please address the lint warning for duplicate includes in `amf_app_ue_context_and_proc.hpp` and `amf_fsm.cpp` 

Note that the protobuf generated headers are still `.h` , e.g. `s1ap_state.pb.h` even though it is all C++.

PR #12480, opened 2022-04-15
ci(dp): Run integration tests once and improve logging from them

sklgromek:

## Summary
- Run integration tests only once
- Print  test logs on stdout
- Initcontainer to wait for elasticsearch 

## Additional Information

### DIFF
--- a/dp/Makefile
+++ b/dp/Makefile
@@ -66,13 +66,13 @@ dev_orc8r:
 
 
 .PHONY: dev_orc8r_integration_test
-dev_orc8r_integration_test: cleanup_test_db
+dev_orc8r_integration_test: cleanup_test_db _ci_init
 	kubectl delete jobs -l jobName=domain-proxy-db-setup --ignore-not-found=true
 	kubectl delete pods test-runner-orc8r --ignore-not-found=true
 	skaffold dev -p orc8r-deployment,integration-tests-no-orc8,integration-tests-orc8r-only --trigger=manual
 
 .PHONY: dev_integration_test
-dev_integration_test: cleanup_test_db
+dev_integration_test: cleanup_test_db _ci_init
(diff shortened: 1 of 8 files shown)


dp/Makefile
```diff
@@ -64,12 +64,17 @@ clean:
 dev_orc8r:
 	skaffold dev -p orc8r-deployment --trigger=manual
 
+
 .PHONY: dev_orc8r_integration_test
-dev_orc8r_integration_test:
+dev_orc8r_integration_test: cleanup_test_db
```
jkmar: Is this "private target", but prefixing it with `_`, i.e. can this be run as standalone command? There was also `orc8r_integration_tests` target which also did some setup required by this one

dp/skaffold.yaml
```diff
@@ -106,7 +106,9 @@ profiles:
               extraEnv:
                 APP_CONFIG: "TestConfig"
             db_service:
-              enabled: false
+              enabled: true
```
jkmar, line 114: I'm afraid that this change requires also changing the setup in tests themselves, since setup in tests creates all models in set up and drops them in tear down, so it races with db services.
In fact this is exactly the reason, why it was disabled in the first place.
So to sum up, this change is fine, but only if accompanied by test setup change.

Neudrino (commented):
Hi @sklgromek ,
if you are going to further follow up with this PR, there are also some yamllint issues to be addressed.

sklgromek:
> 



> Hi @sklgromek , if you are going to further follow up with this PR, there are also some yamllint issues to be addressed.

Hi @Neudrino, ok I will take care of it.

xbend:
@magma/approvers-orc8r-infra 
Please disregard this PR for review. We are trying to see if the latest commit fixes issue with smsd helm. If it does, we will provide a separate PR with just that commit.

jkmar (approved):

PR #12492, opened 2022-04-18
fix(mme): check entire supported TA list

JosiahWhite:
Signed-off-by: Josiah White <me@josiahwhite.io>

## Summary

Fixed a bug in the MME where an S1 setup request with multiple broadcast PLMNs would be rejected if the first PLMN did not match the configured PLMN.

This PR changes the checking so that the MME will accept the S1 setup request as long as a single item matches the configured PLMNs and TAC in the MME configuration.

## Test Plan

Verified with existing integration tests, added an additional test for this situation.
Confirmed with pcap before and after that this problem is solved.

## Additional Information

The supplied test case will not succeed until the S1APTester code has been patched to properly handle the `suppTA` parameter. The S1APTester PR has been opened here: https://github.com/magma/S1APTester/pull/74

### DIFF
--- a/lte/gateway/python/integ_tests/s1aptests/test_s1setup_success_secondary_plmn.py
+++ b/lte/gateway/python/integ_tests/s1aptests/test_s1setup_success_secondary_plmn.py
@@ -13,8 +13,6 @@ limitations under the License.
 
 import ctypes
 import unittest
-import time
-import os
 
 import s1ap_types
 from integ_tests.s1aptests.s1ap_utils import S1ApUtil
--- a/lte/gateway/python/integ_tests/s1aptests/test_s1setup_success_secondary_plmn.py
+++ b/lte/gateway/python/integ_tests/s1aptests/test_s1setup_success_secondary_plmn.py
@@ -1,0 +1,89 @@
+"""
+Copyright 2020 The Magma Authors.
+
+This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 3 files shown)


ardzoht (approved):
lgtm, thanks for including s1ap test coverage

rsarwad (approved):
LGTM

rsarwad:
Hi @uri200, The changes

uri200 (approved):

lionelgo (commented):
The assumption made in magma is that there is no more than one PLMN, that's it ?

JosiahWhite:
> The assumption made in magma is that there is no more than one PLMN, that's it ?

There is only one PLMN that can be configured via NMS, but the MME configuration can be modified to serve multiple PLMNs. This also fixes an edge case where a manually configured eNB with multiple PLMNs will not attach properly if the first supported TA does not match a PLMN configured on the MME.

uri200 (approved):

lte/gateway/python/integ_tests/s1aptests/test_s1setup_success_secondary_plmn.py
```diff
@@ -0,0 +1,87 @@
+"""
+Copyright 2020 The Magma Authors.
```
themarwhal, line 2: 2022* just for future :)

PR #12504, opened 2022-04-20
fix(orc8r): Upgrade docker tags on helm charts

ShubhamTatvamasi:

## Additional Information

### DIFF
--- a/orc8r/cloud/deploy/terraform/orc8r-helm-aws/examples/basic/main.tf
+++ b/orc8r/cloud/deploy/terraform/orc8r-helm-aws/examples/basic/main.tf
@@ -93,7 +93,7 @@ module "orc8r-app" {
   elasticsearch_disk_threshold = tonumber(module.orc8r.es_volume_size * 75 / 100)
 
   orc8r_deployment_type = "fwa"
-  orc8r_tag             = "1.6.1"
+  orc8r_tag             = "1.7.0"
 }
 
 output "nameservers" {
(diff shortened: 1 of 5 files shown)


ShubhamTatvamasi:
Hi @tmdzk, we have to upgrade docker tags in our helm charts. Should we also upgrade helm versions or just docker tags in values.yaml is enough? Then we can re-release helm charts with the latest docker tags.

tmdzk:
> Hi @tmdzk, we have to upgrade docker tags in our helm charts. Should we also upgrade helm versions or just docker tags in values.yaml is enough? Then we can re-release helm charts with the latest docker tags.

we should do both

ShubhamTatvamasi:
> > Hi @tmdzk, we have to upgrade docker tags in our helm charts. Should we also upgrade helm versions or just docker tags in values.yaml is enough? Then we can re-release helm charts with the latest docker tags.
> 
> we should do both

Okay, on it.

ShubhamTatvamasi:
Hi @tmdzk, now helm chart version is also updated.

tmdzk (approved):

tmdzk:
We need to backport to v1.7 right? @ShubhamTatvamasi

ShubhamTatvamasi:
> 

Yes @tmdzk, we have to backport this to v1.7 branch also. Should I create another PR for that?

maxhbr (approved):
Changes look sensible

PR #12505, opened 2022-04-20
refactor(nms): Consolidate themes

thmsschmitt:
## Summary

Colors were taken from three different themes across the app: 
- The NMS default theme
- The `fbs_js_core` theme aka symphony
- The alarms theme 

Under some circumstance the material UI defaults were used because the `fbs_js_core` was not set though a context provider but was accessed from the theme passed down by `makeStyles`.

All theme usages have been consolidated to the NMS default theme using the most similar colors. No new colors were added. The `fbs_js_core` theme and the  alarms theme were removed.

## Test Plan

Visually inspect the app and check that nothing looks off. Some differences in legacy components are to be expected.


andreilee (approved):
Looks great! (Or hopefully the same!)

PR #12508, opened 2022-04-20
chore(mme): Removes include of bits/stdc++.h

electronjoe:
This header should not be used in production environment, as it imports
all c++ headers en mass. Instead the appropriate headers should be
included individually (if any are missing).

Compilation succeeds without this include - I am unsure what headers (if any) required this include. We will eventually run `Include-What-You-Use` to fix any missing includes.


electronjoe:
@rsarwad gentle ping when you are available for review.

lionelgo (approved):

ardzoht (approved):

PR #12511, opened 2022-04-20
docs: add DCO.md, with content from https://developercertificate.org/

maxhbr:
## Summary

This adds the DCO explicitly to the repository, right now it is just implicitly mentioned.

Currently it is implied by
- the required DCO check in the CI
- the content of https://github.com/magma/magma/wiki/Contributing-Code/_compare/3791b1a38e99e519c8c72f23a8273afe6b566b97...68f40961f8bc55cea3e83b6a50f690cb35b4807f before this diff (which was also created by me. This PR tries to move that information to the main repository)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,6 @@
 See the [Contributor Guide](https://github.com/magma/magma/wiki/Contributor-Guide/)
 for information on developer onboarding, workflow, contribution conventions,
 presubmit checks and more.
+
+Please read also the [DCO](./DCO.md) for details regarding regarding the
+sign-off rules for a contribution.
--- a/DCO.md
+++ b/DCO.md
@@ -1,0 +1,37 @@
+# DCO
+
+You must sign-off all commits on the originating branch for a PR, which certifies that you wrote it or otherwise have the right to pass it on as an open-source contribution.
+The rules are pretty simple: if you can certify the below (from [developercertificate.org](https://developercertificate.org/)):
+
+> Developer's Certificate of Origin 1.1
+>
+> By making a contribution to this project, I certify that:
+>
+> (a) The contribution was created in whole or in part by me and I
+>     have the right to submit it under the open source license
+>     indicated in the file; or
+>
+> (b) The contribution is based upon previous work that, to the best
+>     of my knowledge, is covered under an appropriate open source
+>     license and I have the right under that license to submit that
+>     work with modifications, whether created in whole or in part
+>     by me, under the same open source license (unless I am
+>     permitted to submit under a different license), as indicated
+>     in the file; or
+>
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,3 +1,6 @@
 See the [Contributor Guide](https://github.com/magma/magma/wiki/Contributor-Guide/)
 for information on developer onboarding, workflow, contribution conventions,
 presubmit checks and more.
+
+Please read also the [DCO](./DCO.md) for details regarding the legal
```
ildikov: I'm not sure 'legal requirements' is the best term to use here, would be good to double check.

DCO.md
```diff
@@ -0,0 +1,37 @@
+# DCO
+
+You must sign-off all commits on the originating branch for a PR, which certifies that you wrote it or otherwise have the right to pass it on as an open-source contribution.
+The rules are pretty simple: if you can certify the below (from [developercertificate.org](https://developercertificate.org/)):
```
ildikov, line 4: I wonder if the link is enough here in case the text would get updated there?

maxhbr: It is just a pointer to the origin. The full text is copied and if it changes that has no implications.

CONTRIBUTING.md
maxhbr: Agreed, but I also had no better term for that.

Neudrino (commented):
I think we should rather add all this information to the [Contributing Code section within the Wiki](https://github.com/magma/magma/wiki/Contributing-Code/). Otherwise we end up with yet another place were documentation is distributed (next to docusaurus, wiki, confluence) and it becomes even more confusing where to find what.

maxhbr:
I already contributed it to https://github.com/magma/magma/wiki/Contributing-Code/#signed-off-commits, but this is a effort to instead move it next to the actual code. This is not just documentation, this is a legal document and similar to the license it is best placed in the root of the project. E.g. for having it in VCS.

Neudrino (approved):

maxhbr:
This was discussed in TSC in https://wiki.magmacore.org/display/HOME/April+25%2C+2022+Meeting, and the outcome is, that a review by codeowners + @rdefosse is sufficient.

CONTRIBUTING.md
rdefosse: @maxhbr 

Instead of `details regarding the legal requirements for a contribution.` we could say:

`details regarding the sign-off rules for a contribution.`

Rules is not as strong as legal requirements.

rdefosse (changes requested):

CONTRIBUTING.md
maxhbr: @rdefosse, I applied your suggestion

rdefosse (approved):

PR #12515, opened 2022-04-21
feat(orc8r): Added terraform locals block for specifying region

ShubhamTatvamasi:

## Summary

If we use `locals` terraform block then we can use it for specifying `region` at just one place.

## Additional Information


hcgatewood (approved):
I knew there had to be a solution 🙂  Thanks for the fix!

PR #7477, opened 2021-06-10
orc8r: Fix hanging Swagger UI

hcgatewood:
## Summary

nprobe service is crashing, causing Swagger UI to time out. Fix: change the swagger spec timeout to be short (5s), so it will skip the failing service and still return in time.

Resolves #7374.

## Test Plan

- [x] Manual

## Additional Information

- [ ] This change is backwards-breaking

PR #12516, opened 2022-04-21
fix(orc8r): make polling for service specs parallel

maxhbr:
## Summary

This fixes a small todo created by @hcgatewood, where the polling of service specs should be parallelized so that the timeouts of 5secs do not accumulate. They might in the worst case exceed the parent timeout of 60secs. This is related to #7477.

## Test Plan

If polling of a service fails, that result is empty in the array. To test that this is handled, the file `empty.yml` was added to the corresponding testdata.

### DIFF
--- a/orc8r/cloud/go/obsidian/swagger/poll.go
+++ b/orc8r/cloud/go/obsidian/swagger/poll.go
@@ -18,6 +18,7 @@ import (
 	"github.com/pkg/errors"
 
 	"magma/orc8r/cloud/go/obsidian/swagger/spec"
+	"magma/orc8r/cloud/go/parallel"
 )
 
 // GetCombinedSpec polls every servicer registered with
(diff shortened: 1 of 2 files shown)


hcgatewood (changes requested):
Hey @maxhbr this looks great! Though for simplicity, I was thinking we use an existing function we have -- thoughts?

https://github.com/magma/magma/blob/ff012313835a8ecb0cd4b9ac2055186468295838/orc8r/cloud/go/parallel/map.go#L30-L29

maxhbr:
Good point! ~~But this function fails on error, instead of silently swallowing it.~~ I will swallow the error in the worker function.

hcgatewood (approved):
Hm okay I didn't realize the input was a struct rather than a string. This is definitely one of those things that will (hopefully) be a whole lot cleaner with generics 🙂 

Either this or the way you had it before works for me. Thanks for the fix!

PR #12518, opened 2022-04-21
chore(dp): Make relation property level instead of table level

jkmar:
## Summary

This allows making multiple relations to the same table.

### DIFF
--- a/dp/cloud/go/services/dp/storage/db/table.go
+++ b/dp/cloud/go/services/dp/storage/db/table.go
@@ -26,7 +26,6 @@ func CreateTable(tx sq.BaseRunner, builder sqorc.StatementBuilder, metadata *Mod
 		PrimaryKey("id")
 	fields := metadata.Properties
 	tableBuilder = addColumns(tableBuilder, fields)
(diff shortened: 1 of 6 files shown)


dp/cloud/go/services/dp/storage/db/table.go
```diff
@@ -47,17 +46,13 @@ func addColumns(builder sqorc.CreateTableBuilder, fields []*Field) sqorc.CreateT
...
-		builder = builder.ForeignKey(
-			table,
-			map[string]string{column: "id"},
-			sqorc.ColumnOnDeleteCascade,
-		)
+		if field.Relation != "" {
+			builder = builder.ForeignKey(
+				field.Relation,
+				map[string]string{field.Name: "id"},
+				sqorc.ColumnOnDeleteCascade,
```
WojSad, line 53: is this the only type of action we're expecting on delete or do we want to make it customizable?

jkmar: This is actually just used in tests, so having it hardcoded like this is fine I think

WojSad (approved):

PR #12522, opened 2022-04-21
chore(sessiond): Hpp-ify SessionD header files

### LINKED ISSUE #10319: Rename C++ headers from .h to .hpp extension

themarwhal:
## Summary
Addresses https://github.com/magma/magma/issues/10319

This change is mostly automated changes by the hppify script @electronjoe authored here: https://github.com/magma/magma/pull/12461

All files are moved, not copied so git history is preserved.

## Test Plan
CI

## Additional Information


electronjoe (approved):
Yay!

nstng (approved):

LKreutzer (approved):

GANESH-WAVELABS (approved):
LGTM!

PR #12524, opened 2022-04-21
ci: Adapt LTE Integ Test VM resources to macOS runners

jheidbrink:
## Summary

This is the second attempt at https://github.com/magma/magma/pull/12501 which i had to revert. (I was referring to Github Workflow inputs that only exist in my fork)

## Test Plan

I ran the LTE Integ Test in my fork. I didn't wait for it to complete, but verified that it went past the point that broke in https://github.com/magma/magma/pull/12501.

## Additional Information

### DIFF
--- a/.github/workflows/lte-integ-test.yml
+++ b/.github/workflows/lte-integ-test.yml
@@ -43,6 +43,8 @@ jobs:
       - name: Run the integ test
         run: |
           cd lte/gateway
+          export MAGMA_DEV_CPUS=3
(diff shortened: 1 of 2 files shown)


themarwhal (approved):

ajahl (approved):

Neudrino (approved):
I do not see too much difference to the previous iteration except for the non-templating.

The semantic PR check needs to be fixed though.

PR #12525, opened 2022-04-21
chore: add a utility script to generate compilation db with Bazel

themarwhal:

## Summary
We currently rely on a VSCode plug in to generate the compilation_commands.json file with Bazel. While this useful, if we want to enable IWYU / clang-tidy based checks in CI, we will need a separate script to handle this. This PR provides a script that can generate a compilation db without relying on the plugin :)

I based this script off of a script that the envoy proxy team created. (https://github.com/envoyproxy/envoy/blob/dca672af515029d36ddfc378480e3041f4d3a9f3/tools/gen_compilation_database.py) They use an apache 2 license, so I copied that into the file and mentioned which commit I copied off of.

I am thinking it might be good have the script run everytime a devcontainer is opened so that we skip a manual step in IDE configuration, but I will think about that a bit more.

## Test Plan
I ran `dev_tools/gen_compilation_database.py` and made sure a compile_commands.json file is enabled. I ran the clangd plugin against it to double check it is a valid output.
I ran `dev_tools/gen_compilation_database.py --output_dir=/tmp` and ran the apply-iwyu script against it to also double check it is a valid DB.

## Additional Information

### DIFF
--- a/dev_tools/apply-iwyu.sh
+++ b/dev_tools/apply-iwyu.sh
@@ -13,17 +13,26 @@
 ################################################################################
 
 # How to use: 
-# 1. Run `Bzl: Bazel/C++: Generate Compilation Database` in the VSCode command palette (CMD+Shift+P) to generate the compilation database
-# 2. $MAGMA_ROOT/dev_tools/apply-iwyu.sh $PATH_TO_APPLY_IWYU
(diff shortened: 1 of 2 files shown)


electronjoe (approved):
Nice find Marie!

Issue #12532, opened 2022-04-22
Towards ARM-based Bazel Builds


nstng:
For the ongoing ARM effort we also need to make sure that Bazel builds can happen on ARM architecture. This includes local development and CI workflows.

* all Bazel workflows are using the bazel-base image
  * the workflow currently use the devcontainer image which has the downside that software is set up that is not needed by the Bazel builds. This additional software makes builds less hermetic (and thus missing dependencies are hard to track). Also some of these dependencies are currently not ready for ARM.
  * [ ] https://github.com/magma/magma/pull/12539
* the bazel base image is ARM ready
* the devcontainer image is ARM ready

nstng:
closed as out-of-scope for now - currently there is no CI arm support. A new issue can be created when/if this becomes relevant again.

PR #12539, opened 2022-04-22
chore(ci): bazel workflows use the bazel base image

nstng:

## Summary

Currently the devcontainer image is used for Bazel CI workflows. This has the disadvantage that pre-installed libraries can be conflicting the hermetic(-ish) build process of Bazel. We should use the leaner Bazel base image for the workflows.

Another advantage of the Bazel Base image is that less pre-installed libraries makes it more easy to switch to an ARM based image. See https://github.com/magma/magma/issues/12532.

Note that `bazel-cache-push.yml` was not adapted because s3 caches for the github code-space feature are created here, i.e., it makes complete sense to do this in the devcontainer image. 

## Test Plan

* CI
* `bazel.yml`, `agw-workflow.yml`, `gcc-problems.yml` have been run successfully on the forked repo -> https://github.com/nstng/magma/actions

## Additional Information

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -18,6 +18,7 @@ concurrency:
 
 env:
   DEVCONTAINER_IMAGE: "ghcr.io/magma/magma/devcontainer:latest"
+  BAZEL_BASE_IMAGE: "ghcr.io/magma/magma/bazel-base:latest"
   BAZEL_CACHE: bazel-cache
   BAZEL_CACHE_REPO: bazel-cache-repo
 
@@ -181,10 +182,10 @@ jobs:
           BAZEL_CACHE_CUTOFF_MB: 6000
         run: |
           ./.github/workflows/check-bazel-cache-dir-size.sh "$BAZEL_CACHE_DIR" "$BAZEL_CACHE_CUTOFF_MB"
-      - name: Setup Devcontainer Image
+      - name: Setup Bazel Base Image
         uses: addnab/docker-run-action@v2
         with:
-          image: ${{ env.DEVCONTAINER_IMAGE }}
+          image: ${{ env.BAZEL_BASE_IMAGE }}
           # Run a simple echo command to pull down the image. This makes it a bit more clear how much time is spent on building Magma and not pulling down the image.
           run: |
             echo "Pulled the devontainer image!"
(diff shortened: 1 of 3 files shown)


.github/workflows/agw-workflow.yml
```diff
@@ -375,6 +376,13 @@ jobs:
...
           run: |
             echo "Pulled the devontainer image!"
+      - name: Setup Bazel Base Image
+        uses: addnab/docker-run-action@v2
+        with:
+          image: ${{ env.BAZEL_BASE_IMAGE }}
+          # Run a simple echo command to pull down the image. This makes it a bit more clear how much time is spent on building Magma and not pulling down the image.
+          run: |
+            echo "Pulled the bazel base image!"
             bazel # pull down bazel, if bazel download fails we can fail before we do all the lengthy work below
```
Neudrino, line 386: Would we still want to be able to execute the parts on the devcontainer even when the bazel pull fails? I.e. only pull the image right before use? In that case we would see at least from one part of the pipeline, if it can successfully build. :man_shrugging:

Neudrino (approved):

themarwhal:
I wonder if it's still useful to have at least one CI check for building with the devcontainer? It might be good to have a check to make sure the devcontainer is still healthy.

themarwhal:
also, feel free to move the bazel-base docker directory output of experimental :)

nstng:
> I wonder if it's still useful to have at least one CI check for building with the devcontainer? It might be good to have a check to make sure the devcontainer is still healthy.

I wanted to answer that `bazel-cache-push.yml` still builds on the devcontainer - but the I realized that this job is red for a long time :( https://github.com/magma/magma/actions/workflows/bazel-cache-push.yml?page=3 - we will have a detailed look.
When this is fixed then we should implement an alarm to be more aware of this job.
Regarding long term: I agree that we should have a PR based job to check if the devcontainer is still healthy. Imho we should do this when remote caches are implemented sufficiently. Currently I fear the github caches are too small (I do not know how compatible the bazel-base and devcontainer caches are)

Lets see if https://github.com/magma/magma/pull/12548 solves this.

themarwhal (approved):

.github/workflows/agw-workflow.yml
```diff
@@ -181,33 +182,33 @@ jobs:
...
           run: |
             echo "Pulled the devontainer image!"
       - name: Create a directory to store XML test results
         uses: addnab/docker-run-action@v2
         with:
           shell: bash
-          image: ${{ env.DEVCONTAINER_IMAGE }}
+          image: ${{ env.BAZEL_BASE_IMAGE }}
           # TODO: Remove work-around mount of Github workspace to /magma (https://github.com/addnab/docker-run-action/issues/11)
-          options: -v ${{ github.workspace }}:/workspaces/magma/ -v ${{ github.workspace }}/lte/gateway/configs:/etc/magma
```
LKreutzer, line 197: Why do we need to remove the `workspaces` from the path? Is it because the `bazel-base` image does not use this concept?

LKreutzer (approved):

.github/workflows/agw-workflow.yml
nstng: The bazel-base `MAGMA_ROOT` is at `/magma`, for devcontainer it's `/workspace/magma`.

Neudrino: It might be easier to change the bazel-base container to also use `/workspace/magma` or update the devcontainer to use just `/magma` to avoid switches.
Updating the bazel-base image seems currently easier to me though.

nstng: Yes, it should be unified. I would move this out of scope here and suggest to do this in the effort to make the devcontainer build on top of the bazel base image. Sounds good?

.github/workflows/agw-workflow.yml
nstng: This seems to be a little inconsistent. Almost all jobs are executed even if the bazel binary download fails (`if: always()`). Even the bazel coverage job has `if: always()`. It should be consolidated in what case we want to do a partial upload of generated coverage - but this is getting out of scope for this task.

nstng:
> also, feel free to move the bazel-base docker directory output of experimental :)

We are currently working on building the devcontainer on top of the bazel base image - we will move the bazel base directory in this context.

PR #12543, opened 2022-04-22
chore: Generate compile_commands.json on devcontainer creation

themarwhal:
## Summary
I've recently added a script to generate compile_commands.json quickly without relying on a VSCode extension (which can be clunky sometimes). Since the generation only takes a few seconds, I'm adding the generation to devcontainer startup logic so that clangd can pick it up on startup.

## Test Plan
Testing by spinning up a codespace from this PR

## Additional Information


vktng (approved):
We have tested it in the devcontainer locally, looks good. Takes a bit more time to rebuild, but this is acceptable.

LKreutzer (approved):

vktng:
Why is this a draft? Seems like a reasonable change.

PR #12548, opened 2022-04-24
fix(ci): bazel-cache-push.yml has enough memory for build

nstng:

## Summary

The VM-part of the "Push Bazel Cache To S3" workflow is currently failing - see https://github.com/magma/magma/actions/workflows/bazel-cache-push.yml.

### Analysis
* The reason are out of memory errors https://github.com/magma/magma/runs/6143309276?check_suite_focus=true
```
FATAL: bazel ran out of memory and crashed. Printing stack trace:
java.lang.OutOfMemoryError: Java heap space
```
* Memory reserved for a VM is hardcoded - compared to docker, where the available memory from the host is used. This is, by hardcoding the VM memory correctly, the job might run through without failure. But compared to a docker setup, the VM also needs some overhead memory.
  * **Not done here:** check if the VM memory can be increased so that the job runs successfully with the current configuration.

### Solution Here
* Increase VM memory (slightly) by using mechanics from https://github.com/magma/magma/pull/12524 (thx @jheidbrink)
* Running Bazel with memory saving settings - this has negative effects on incremental builds (this should not matter much here as these are non-incremental CI builds, see https://bazel.build/docs/memory-saving-mode)

## Test Plan

The workflow only runs on master pushes. Run on local fork: https://github.com/nstng/magma/actions/runs/2216611016

## Additional Information

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,4 +1,5 @@
 startup --output_base=/tmp/bazel
+startup --host_jvm_args="-Xmx8g"
 
 # STYLE CONFIGS
 build --announce_rc
@@ -70,6 +71,11 @@ build:asan --per_file_copt=^.*/gateway/c/.*$@-fsanitize=address,-fsanitize=undef
 # LSAN
 build:lsan --linkopt=-fsanitize=leak
 build:lsan --per_file_copt=^.*/gateway/c/.*$@-fsanitize=leak,-fno-omit-frame-pointer
+
+# Config for running Bazel with limited memory
+# see https://bazel.build/docs/memory-saving-mode
+# This is helpful for builds on the VM on github runners where only 14g are available.
+build:limit_mem --discard_analysis_cache --nokeep_state_after_build --notrack_incremental_state
 
 # Config for turning up more GCC warnings
 build:max_gcc_warnings --per_file_copt=^.*/gateway/c/.*$@-Wextra,-Wshadow,-Wimplicit-fallthrough,-Wduplicated-cond,-Wduplicated-branches,-Wlogical-op,-Wnull-dereference,-Wformat=2,-Wstrict-overflow=4,-Wuninitialized,-Wshift-overflow=2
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -1,4 +1,5 @@
 startup --output_base=/tmp/bazel
+startup --host_jvm_args="-Xmx8g"
```
LKreutzer, line 2: This flag would also be used in local builds correct? I am somewhat worried about what impact this might have locally - but I could not see any `OutOfMemoryError`s when building some things like the `session_manager` or the MME core.

nstng: It is more than the default value (otherwise the test runs on my fork would have failed) - but I was not able to find in the Bazel docs what the default value actually is.

LKreutzer: I found a comment on [sourcegraph](https://sourcegraph.com/github.com/sorbet/sorbet/-/blob/.bazelrc?L6) that "the default is to use 1/4 of memory" (unclear how reliable this info is):
```
startup --host_jvm_args=-Xmx3072m # we need at least 3G to reliably build, the default is to use 1/4 of memory, which makes flaky on 8G boxes
```

Could a fixed 8GB limit be an issue for devs that have less memory available or can we assume that everyone has at least 8GB to make use of?

nstng: Discussed offline: using less memory makes the CI jobs fail. We expect dev machines to have >8gb of memory. If we get feed-back for failing CI jobs or dev setup then we can adapt this further.

LKreutzer (approved):

.github/workflows/bazel-cache-push.yml
```diff
@@ -56,23 +56,25 @@ jobs:
       - name: Bring up the Magma VM
         run: |
           cd lte/gateway
+          export MAGMA_DEV_CPUS=3
+          export MAGMA_DEV_MEMORY_MB=9216
```
Neudrino, line 60: I assume this is intentionally 8GB+1GB to allow the overhead needed for the JVM or such?
Just want to make sure...

Neudrino (approved):

.github/workflows/bazel-cache-push.yml
nstng: Right - see https://github.com/magma/magma/pull/12524 and the related description/analysis https://github.com/magma/magma/pull/12501

PR #12293, opened 2022-03-25
ci: reviewdog filter consolidation

Neudrino:
## Summary

- reduction of tech-depth for reviewdog workflow

## Test Plan

- CI:
  - without changes: https://github.com/magma/magma/actions/runs/2041381006
  - with changes: https://github.com/magma/magma/actions/runs/2041395553 (including arbitrary newlines in go code)
- Actual annotations to the PR are not available until this is on master due to Github token permission limitations.

## Additional Information

- requires https://github.com/magma/magma/pull/12254

PR #12553, opened 2022-04-25
feat(ci): consolidate linter implementations to use reviewdog for better Github integration

Neudrino:
## Summary

- add tflint to reviewdog to have this integrated with Github as well.

## Test Plan

- CI
  - no change to Terraform files: https://github.com/magma/magma/actions/runs/2219664144
  - change to Terraform files which are green: https://github.com/magma/magma/actions/runs/2219674103

## Additional information

- requires https://github.com/magma/magma/pull/12293


maxhbr (changes requested):
Changes look good, but PR title does not follow semantic commits format.

maxhbr (approved):
Ah, `ci: ` is allowed, but I would prefer to have `feat(ci):` or `fix(ci):`.

I think `ci` should be used in the context of for example `ci(orc8r):` if you do ci work for orc8r. If a feature is added to ci it should be `feat(ci)`.

maxhbr:
This adds a new top level folder that is not covered in the codeowners file. Is there a group that would be appropriate?

Neudrino:
> This adds a new top level folder that is not covered in the codeowners file. Is there a group that would be appropriate?

Maybe I misunderstand the comment, but the changed file is covered by the `@magma/approvers-ci` group.

https://github.com/magma/magma/blob/master/CODEOWNERS#L3

Neudrino:
> This adds a new top level folder that is not covered in the codeowners file. Is there a group that would be appropriate?

Maybe I misunderstand the comment, but the changed file is covered by the `@magma/approvers-ci` group.

https://github.com/magma/magma/blob/master/CODEOWNERS#L3

maxhbr (approved):
LGTM

PR #12572, opened 2022-04-27
fix(agw): Format lte/gateway/python with clang-format-11

### LINKED ISSUE #12507: ebpf: `lte/gateway/python` sub-dir is not being clang-formatted
Our new clang-format use and linter actions in Github Actions - are omitting / missing this sub-directory (which does actually include c files now).

Presently the formatting is supplied by Makefile at `lte/gateway/Makefile` but omits (accidentally) the `lte/gateway/python` sub-directory.

```
cd lte/gateway
make format_all
```

Needs repair.  Presumably so does the GH Action.

voisey:
## Summary

Closes https://github.com/magma/magma/issues/12507.

I have:
* added the formatting of `lte/gateway/python` to `lte/gateway/Makefile`
* updated the GitHub workflow, `agw_workflow.yml`, so that the formatting gets run in the CI
* formatted the relevant C files by running `cd lte/gateway && make format_all`

## Additional Information

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -454,10 +454,10 @@ jobs:
           clangFormatVersion: 11.1.0
           # taken from .clang-format
           style: file
-      - name: Check clang-format for lte/gateway/c
+      - name: Check clang-format for lte/gateway
         uses: DoozyX/clang-format-lint-action@v0.13
         with:
-          source: 'lte/gateway/c'
+          source: 'lte/gateway/c lte/gateway/python'
           extensions: 'h,hpp,c,cpp'
           clangFormatVersion: 11.1.0
           # taken from .clang-format
--- a/lte/gateway/Makefile
+++ b/lte/gateway/Makefile
@@ -165,6 +165,8 @@ format_all:
 	/usr/bin/clang-format-11 -i {} \;
 	find $(MAGMA_ROOT)/lte/gateway/c/ \( -iname "*.c" -o -iname "*.cpp" -o -iname "*.h"  -o -iname "*.hpp" \) -exec \
 	/usr/bin/clang-format-11 -i {} \;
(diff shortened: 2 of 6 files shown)


themarwhal (approved):

nstng (approved):
lgtm - actually surprised that this works on `lte/gateway/python/magma/kernsnoopd/ebpf/byte_count.bpf.c` as this is not completely valid c code. This needs to be templated before it is runable (e.g., #45 `{{PROXY_PORT}}`).

Neudrino (approved):
_Optional:_ Does `DoozyX/clang-format-lint-action` support lists in the `source:` field? If so, one might use one. If not, :cry: .

.github/workflows/agw-workflow.yml
```diff
@@ -454,10 +454,10 @@ jobs:
           clangFormatVersion: 11.1.0
           # taken from .clang-format
           style: file
-      - name: Check clang-format for lte/gateway/c
+      - name: Check clang-format for lte/gateway
         uses: DoozyX/clang-format-lint-action@v0.13
         with:
-          source: 'lte/gateway/c'
+          source: 'lte/gateway/c lte/gateway/python'
```
Neudrino, line 460: How about `source: 'lte/gateway'`? Does that not work?

voisey: I think that would work too. I just thought it would be better to be more explicit, so that it's clear that `lte/gateway` (without a `/c`) isn't a case of us not being specific enough. Otherwise someone might come along, assume the Python folder contains no C code and re-add the `/c` to the path.

Neudrino: Good reasoning. :+1:

Neudrino (approved):

electronjoe (approved):
Thanks!

PR #12574, opened 2022-04-27
fix(orc8r): Allow updating QoS profiles

### LINKED ISSUE #12307: Cannot change QoSProfile for a policy

### Your Environment

- **Version:** 1.6.1
- **Affected Component:** Orchestrator
- **Affected Subcomponent:** Orchestrator pod/service 
- **Deployment Environment:** bare-metal (AGW)

### Describe the Issue

**To Reproduce**
1. On NMS UI, Make sure you have a few QoS Profiles defined under Traffic>Policies>Profiles
2. On NMS UI go to Traffic>Policies>Policies
3. Edit a existing policy and change the QoS Profile for it. Say earlier it was Policy1 now make it Policy2.
4. Click Save.
5. Open the same policy for edit again. You can see that it still points to Policy1 and not the updated Policy2.

**Expected behavior**
The policy change should be applied and new policy should be visible in the UI.

sebathomas:
## Summary

Updating of QoS profiles was broken because the update handler did not remove old associations from the DB.
Fixes #12307.

## Test Plan

- Adapted integration test to include an update that changes between profiles.
- Started NMS and confirmed that changing profiles is now possible.

## Additional Information

### DIFF
--- a/lte/cloud/go/services/policydb/obsidian/handlers/policy_handlers_test.go
+++ b/lte/cloud/go/services/policydb/obsidian/handlers/policy_handlers_test.go
@@ -1105,7 +1105,19 @@ func TestPolicyWithQoSProfile(t *testing.T) {
 	tc := tests.Test{
 		Method:         "POST",
 		URL:            "/magma/v1/lte/n1/policy_qos_profiles",
-		Payload:        newTestQoSProfile(),
+		Payload:        newTestQoSProfileNamed("profile0"),
+		ParamNames:     []string{"network_id"},
+		ParamValues:    []string{"n1"},
+		Handler:        postProfile,
+		ExpectedStatus: 201,
+	}
+	tests.RunUnitTest(t, e, tc)
+
+	// Post other profile
+	tc = tests.Test{
+		Method:         "POST",
+		URL:            "/magma/v1/lte/n1/policy_qos_profiles",
+		Payload:        newTestQoSProfileNamed("profile1"),
 		ParamNames:     []string{"network_id"},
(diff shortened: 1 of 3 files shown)


andreilee (approved):

lte/cloud/go/services/policydb/obsidian/handlers/policy_handlers_test.go
```diff
@@ -1194,9 +1232,9 @@ func TestPolicyWithQoSProfile(t *testing.T) {
 	// Delete profile
 	tc = tests.Test{
 		Method:         "DELETE",
-		URL:            "/magma/v1/lte/n1/policy_qos_profiles/profile0",
+		URL:            "/magma/v1/lte/n1/policy_qos_profiles/profile1",
```
maxhbr, line 1235: shouldn't you also delete `profile0`?

maxhbr (commented):
Just a question.

lte/cloud/go/services/policydb/obsidian/handlers/policy_handlers_test.go
sebathomas: It's not necessary for the test logic since it's no longer associated to the policy after the PUT from line 1195. (I don't think the DELETEs are needed to clean up test data, the test generates other entities that are not removed at the end.)

maxhbr: Thanks, that clarifies it

maxhbr (approved):
LGTM

PR #12576, opened 2022-04-27
feat: adopt codeowners to include groups instead of persons

Neudrino:
## Summary

Currently there are single persons the only codeowners for some part of the code. In context of Meta moving developers to new shores, it might be sensible to replace those with appropriate approver-groups, such that development on certain components is not blocked by reduced capacity for the project Magma.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -39,6 +39,7 @@
 # agw code related to orc8r
 /orc8r/gateway/c/ @magma/approvers-agw
 /orc8r/gateway/python @magma/approvers-agw @magma/approvers-orc8r
+/src/go/ @magma/approvers-orc8r
 
 # approvers-agw-<subsection>
 /lte/gateway/c/session_manager @magma/approvers-agw-sessiond
@@ -55,12 +56,10 @@
 /xwf/ @magma/approvers-xwf
 /feg/radius @magma/approvers-xwf @magma/approvers-cwf
 
-/src/go/ @markjen
-
 /dp/ @magma/approvers-domain-proxy
 
 /docs/ @magma/approvers-docs
-/docs/readmes/proposals @electronjoe
+/docs/readmes/proposals @magma/approvers-tsc
 
 # approvers-bazel
 /.bazel-cache/ @magma/approvers-bazel
@@ -74,7 +73,7 @@
 /.bazelrc @magma/approvers-bazel
 *.bazel @magma/approvers-bazel
 
-/CODEOWNERS @amarpad @electronjoe @hcgatewood
+/CODEOWNERS @magma/approvers-tsc
 
 # Generated code excluded from code ownership
 **/go.mod @ghost


CODEOWNERS
```diff
@@ -55,12 +56,9 @@
 /xwf/ @magma/approvers-xwf
 /feg/radius @magma/approvers-xwf @magma/approvers-cwf
 
-/src/go/ @markjen
-
```
Neudrino, line 59: @markjen seems to have left the project in November 2021.

CODEOWNERS
```diff
@@ -55,12 +56,9 @@
...
 /dp/ @magma/approvers-domain-proxy
 
 /docs/ @magma/approvers-docs
-/docs/readmes/proposals @electronjoe
```
Neudrino, line 63: @electronjoe is still with the project today, but we might want to redistribute this to a group of people non-the-less.
Are there any better candidates than just the fallback @magma/approvers-docs ?

CODEOWNERS
```diff
@@ -74,7 +72,7 @@
 /.bazelrc @magma/approvers-bazel
 *.bazel @magma/approvers-bazel
 
-/CODEOWNERS @amarpad @electronjoe @hcgatewood
+/CODEOWNERS @magma/repo-magma-admin
```
Neudrino: Which one would be the appropriate group to be placed here?
(I only put a suggestion here to be able to write this comment and start a discussion.)

I believe @amarpad has a new project since November 2021 as there were only 5 PR reviews since than.
@hcgatewood @electronjoe Would you be able to suggest a appropriate group?

CODEOWNERS
maxhbr: maybe make that part of @magma/approvers-orc8r instead of dropping it? @hcgatewood: what do you think?

CODEOWNERS
maxhbr: I think this needs to be under control of someone in charge of the proposal process. Maybe that needs its own approvers group. Docs is not the right fallback

CODEOWNERS
maxhbr: maybe a tsc group?

CODEOWNERS
Neudrino: Yes. That was [my view as well](https://github.com/magma/magma/pull/12576/files#diff-fcf14c4b7b34fe7a11916195871ae66a59be87a395f28db73e345ebdc828085bR42). I moved it to be consistent with the grouping.

(~~I cannot get those inline code blocks to work somehow.~~ The inline code blocks are not rendered in preview mode, but only after submitting the comment. :confused: )
https://github.com/magma/magma/blob/11edb4f5178b15681fba1ad042f8a3a1e8733222/CODEOWNERS#L42

CODEOWNERS
```diff
@@ -39,6 +39,7 @@
 # agw code related to orc8r
 /orc8r/gateway/c/ @magma/approvers-agw
 /orc8r/gateway/python @magma/approvers-agw @magma/approvers-orc8r
+/src/go/ @magma/approvers-orc8r
```
Neudrino, line 42: I believe this folder is mainly Orc8r, right?

CODEOWNERS
Neudrino: Changed to `@magma/approvers-tsc` in expectation of the group getting set up by @wendallkaters .

CODEOWNERS
Neudrino: Changed to `@magma/approvers-tsc` in expectation of the group getting set up by @wendallkaters .

Neudrino:
Thanks to @wendallkaters for creating a [new TSC approvers group](https://github.com/orgs/magma/teams/approvers-tsc). Thus opening it as real PR for the community and concerned parties.

CODEOWNERS
maxhbr: I do not think that it is strongly related to the orc8r itself. I think that is the gateway written in golang.

CODEOWNERS
electronjoe: Yes we do want to move this off of a single person (me). approvers-tsc sounds like a  good strategy.

electronjoe (approved):
These changes make sense - thank you for taking the initiative to improve things.

ShubhamTatvamasi (approved):

maxhbr:
Approved by TSC on https://wiki.magmacore.org/display/HOME/May+23%2C+2022+Meeting

PR #12579, opened 2022-04-27
fix(agw): Ignore health_cli.py error message

gjalves:
## Summary

When using health_cli.py, remove the "Error: 'utf-8' codec can´t decode byte 0x8a in position 1530: invalid start byte" when syslog has non utf-8 data

## Test Plan

Be sure to have an invalid character in /var/log/syslog. It´s possible to do it using 

```
echo -e -n "\x8a" >> /var/log/syslog
health_cli.py
```
You should expect the command to abort with an error message without the patch, but running smoothly with it.

Signed-off-by: Gustavo Junior Alves <gustavo@radtonics.com>


ardzoht (approved):

ardzoht:
@gjalves Can you please include a test plan validation in the description on the PR?

andreilee (approved):

sebathomas:
@gjalves Just stumbled over this PR. If it is ready to merge, please set the `ready2merge` tag as a label.

PR #12581, opened 2022-04-28
fix(agw): Added flaky tests back to sanity list

VinashakAnkitAman:
## Title
fix(agw): Added flaky tests back to sanity list

## Summary
Many of the test cases were commented because of flaky behavior during sanity. Some of them have already been fixed over the course of time and some of them needs to be retested after introducing flaky retry support for failing test cases. This PR adds those flaky test cases back to sanity list.

## Test plan
Verified with sanity


ardzoht (approved):
lgtm

PR #12582, opened 2022-04-28
fix(agw): Added flaky testcase iteration count

VinashakAnkitAman:
## Title
fix(agw): Added flaky testcase iteration count

## Summary
From the CI logs it was difficult to identify which iteration of test case is executing for flaky test cases retrying on failure. This PR adds print to display the test case execution iteration count.

## Test plan
Verified with sanity


ardzoht (approved):
lgtm

PR #12587, opened 2022-04-28
refactor(nms): Upgrade React Router to version 6

thmsschmitt:
## Summary

Upgrade react-router to version 6.
There are many breaking / conceptual changes see https://reactrouter.com/docs/en/v6/upgrading/v5.

The most important ones for us are:

- `Switch` has been replaced with `Routes`.
- `useHistory` has been replaced with `useNavigate`.
- `Route`s may only be rendered inside of `Routes` or other `Route`s.
- **`Route` paths and `Link.to` paths are now relative**
- Routes with descendant routes in other components **need a trailing `*`**.
- Routes render their descendants via element prop. 

## Test Plan

Run the NMS and try to hit every possible link in the application.
Unfortunaly this is a rather risky change beause URL changes, or paths that no longer match will lead blank pages and silent failures.


andreilee (approved):
For context, why upgrade to v6 if the developers suggest waiting for the backwards compatibility package?

thmsschmitt:
> For context, why upgrade to v6 if the developers suggest waiting for the backwards compatibility package?

A compat package has been released https://github.com/remix-run/react-router/discussions/8753, but its purpose is it to ease the migration and have both version running in parallel so the end goal would still be to upgrade everything.

PR #12591, opened 2022-04-28
feat(dp): migration tests framework + poc test

WojSad:

feat(dp): migration tests framework + poc test

## Summary

Added AlembicTestCase allowing to test python migrations, check database state before and after the migration, test individual tables etc.

### DIFF
--- a/dp/cloud/python/magma/db_service/tests/unit/test_b0cad5321c88.py
+++ b/dp/cloud/python/magma/db_service/tests/unit/test_b0cad5321c88.py
@@ -1,0 +1,38 @@
+from magma.db_service.tests.alembic_testcase import AlembicTestCase
+
+
+class Testb0cad5321c88TestCase(AlembicTestCase):
+
+    def setUp(self) -> None:
+        super().setUp()
+        self.tables = [
+            'cbsd_states',
+            'domain_proxy_logs',
+            'grant_states',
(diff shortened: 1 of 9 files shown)


dp/cloud/python/magma/db_service/tests/unit/test_b0cad5321c88.py
```diff
@@ -0,0 +1,38 @@
+from magma.db_service.tests.alembic_testcase import AlembicTestCase
+
+
+class Testb0cad5321c88TestCase(AlembicTestCase):
+
+    def setUp(self) -> None:
+        super().setUp()
```
jkmar, line 7: It is possible to create different kind of test:
- specify some data before migration (that would be some records in the database)
- specify expected data after migration

For example suppose that we have the following table:
```
CREATE TABLE T (ID INT);
```
And the following records:
```
INSERT INTO T VALUES(1);
INSERT INTO T VALUES(2);
```
now when we execute the following migration:
```
ALTER TABLE T ADD COLUMN A INT;
UPDATE T SET A = ID;
```
then we should get the following result:
```
SELECT * FROM T
ID, A
1, 1
2, 2
```

So the question is can we do it with this version of a framework?

jkmar (approved):

PR #12604, opened 2022-04-29
docs: Update how to start postgres in minikube

sebathomas:
## Summary

The parameters were changed in the current version of the container.
See: https://artifacthub.io/packages/helm/bitnami/postgresql

## Test Plan

Start DB with the new command, run minikube orc8r and confirm that the services can connect to the DB.

## Additional Information

### DIFF
--- a/docs/docusaurus/versioned_docs/version-1.4.X/orc8r/dev_minikube_deployment.md
+++ b/docs/docusaurus/versioned_docs/version-1.4.X/orc8r/dev_minikube_deployment.md
@@ -43,7 +43,7 @@ $ helm install \
     postgresql \
     --create-namespace \
     --namespace magma \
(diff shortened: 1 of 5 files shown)


Neudrino (approved):

Neudrino:
Is `latest` used for the postgres?  In that case it might also fail for releases 1.7 and earlier and should also be updated in that documentation.
E.g.
- https://github.com/magma/magma/blob/master/docs/docusaurus/versioned_docs/version-1.7.0/orc8r/dev_minikube.md
- https://github.com/magma/magma/blob/master/docs/docusaurus/versioned_docs/version-1.6.X/orc8r/dev_minikube.md

sebathomas:
Changed it for all versions.

voisey (approved):

PR #12611, opened 2022-05-02
chore(dp): Improve skaffold orc8r image building stage

sklgromek:
- Adjust skaffold artifacts dependency paths
- Bump up skaffold API version

## Additional Information

### DIFF
--- a/dp/skaffold.yaml
+++ b/dp/skaffold.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: skaffold/v2beta23
+apiVersion: skaffold/v2beta28
 kind: Config
 metadata:
   name: domain-proxy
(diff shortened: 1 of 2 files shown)


xbend (approved):

dp/tools/skaffold_hooks/hooks.sh
```diff
@@ -11,7 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-MAGMA_ROOT=$(realpath ..)
+MAGMA_ROOT=$(realpath .)
```
jkmar, line 14: Is it correct? (magma root should be root of git project, but here it looks like it's pointing to dp)

sklgromek: I changed also context in skaffold build artifacts, and added dir attributes to hooks calls. So they will be called from the project root directory

jkmar: :+1:

dp/skaffold.yaml
```diff
@@ -159,58 +161,51 @@ profiles:
...
         path: /build/artifacts/-
         value:
           image: orcnginx
+          context: ..
           custom:
-            buildCommand: "tools/skaffold_hooks/hooks.sh build_nginx"
+            buildCommand: "dp/tools/skaffold_hooks/hooks.sh build_nginx"
             dependencies:
               paths:
-                - "../dp/**"
```
jkmar, line 185: Also not sure about this, since the orc8r is a monolith and it contains also a dp service

sklgromek: Here are dependencies for orcnginx artifact which is just reverse-proxy to obsidian, so dp dependency it's not necessary

jkmar (approved):

PR #12614, opened 2022-05-02
refactor(nms): Remove fbc-js-core select component

thmsschmitt:
## Summary

There was only one place left were the fbc-js-core `Select` component was used.
It was next to a native Material-UI Select and looked out of place. The fbc-js-core `Select` component was replaced with the native Material-UI Select component and all related code has been deleted.
Before
Screenshot 2022-05-02 at 15 01 22
After
Screenshot 2022-05-02 at 16 06 06

## Test Plan

I made sure that creating users with the three different roles still works. 

I also used the linter to check that no removed code was still in use.
``` 
$ yarn eslint .
...

✖ 186 problems (0 errors, 186 warnings)

✨  Done in 20.25s.
```


Neudrino (approved):

andreilee (approved):
stamped

PR #12617, opened 2022-05-03
refactor(nms): Unify button styling

thmsschmitt:
## Summary

Two different button components were used across the app, native Material UI buttons and a custom button component. The native button component was styled to resemble those buttons https://www.figma.com/file/Mf3kZVGHxdK2DPjxN2O8zT/Magma-Components?node-id=208%3A19476. The custom button component had many different style options some of which also resemble the aforementioned design. Most of the time a different style was used though. More confusing sometimes the props form the custom button component were used together with the material UI button where they had no effect. 

All button have been unified to use native Material UI buttons and the new design.
Solid blue buttons are now solid grey and blue text buttons are now grey text.

Before 
Screenshot 2022-05-02 at 16 46 48

After 
Screenshot 2022-05-02 at 16 47 16

Very few special cases have been introduced were grey text buttons didn't look right. For instance here:  
Screenshot 2022-05-02 at 18 39 05

Note that this also means that the login popup now uses the theme and has a grey button
Screenshot 2022-05-02 at 18 36 45

## Test Plan
Look at all buttons across the app and make sure they look ok.


nms/app/components/AccountSettings.js
```diff
@@ -117,7 +117,11 @@ export default function AccountSettings() {
             />
           </AltFormField>
         </List>
-        <Button onClick={onSave} disabled={!isSaveEnabled}>
+        <Button
+          onClick={onSave}
+          disabled={!isSaveEnabled}
+          variant="contained"
+          color="primary">
```
Neudrino, line 124: _Style question:_ Would it be possible and more aligned to the above `PasswordInput` to do the following? Or is this uncommon?
```javascript
        <Button
          onClick={onSave}
          disabled={!isSaveEnabled}
          variant="contained"
          color="primary"
        >
```

nms/app/components/admin/userManagement/UsersSettings.js
```diff
@@ -144,7 +144,12 @@ function UsersSettings(props: Props) {
     <div className={classes.paper}>
       <div className={classes.header}>
         <Text variant="h5">Users</Text>
-        <Button onClick={() => setShowDialog(true)}>Add User</Button>
+        <Button
+          variant="contained"
+          color="primary"
```
Neudrino, line 149: Does the primary colour needs explicit specification or would this also be used by default?

andreilee:
It looks like the "Login" button may need to be re-aligned.

nms/app/components/AccountSettings.js
thmsschmitt: Yes, this is done by the auto formatter (prettier). 
We could change the setting https://prettier.io/docs/en/options.html#bracket-line
But I'm not sure if should. You will get used it :)

nms/app/components/admin/userManagement/UsersSettings.js
thmsschmitt: It needs to be specified, the default is `default` see https://v4.mui.com/api/button/#props

Neudrino (approved):

thmsschmitt:
> It looks like the "Login" button may need to be re-aligned.

Done
Screenshot 2022-05-10 at 10 55 22

HannaFar (approved):

PR #12620, opened 2022-05-03
feat(agw): Disable SysRq key

berezovskyi-oleksandr:
## Summary

SysRq commands might lead to security vulnerabilities since it performs direct calls to the Linux kernel

See: https://www.kernel.org/doc/html/latest/admin-guide/sysrq.html

## Test Plan

Tested at the real AGW

## Additional Information


ssanadhya (approved):
Lgtm

PR #12624, opened 2022-05-03
refactor(nms): Inline fbc-js-core eslint config

thmsschmitt:
## Summary
We were still using the archived [eslint-config-fbcnms](https://github.com/magma/fbc-js-core/tree/main/fbcnms-packages/eslint-config-fbcnms) package as base config for eslint.

The `eslint-config-fbcnms` has been inlined and some cleanup has been done:
- Overwritten rules have been merged
- `eslint-config-prettier` has been used to identify rules that conflict with prettier. As a result the `comma-dangle` rule and `max-len` rule were removed. The latter would still report warnings for comments that are ignored by prettier, but since they were only warnings they were also ignored by developers ;)   
-  The dependency to `eslint-config-fbjs` was removed because it all its rule were anyhow overwritten.

## Test Plan
I made some changes that should break the linter to verify that they still do so.

### DIFF
--- a/nms/.eslintrc.js
+++ b/nms/.eslintrc.js
@@ -22,90 +22,256 @@ const combinedOpenSourcePattern = new RegExp(
   '(' + newOpenSourcePattern.source + ')|(' + openSourcePattern.source + ')',
 );
 
-module.exports.extends = ['eslint-config-fbcnms'];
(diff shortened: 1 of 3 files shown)


andreilee (approved):
Looks clean!

PR #12627, opened 2022-05-03
refactor(orc8r): Replace multi.go with hashicorp/go-multierror module

alexzurbonsen:
## Summary

This PR modifies Magma's multi error management to use the `hashicorp/go-multierror` module.

Formatting of multierrors changes due to the format used in the hashicorp module.
The new output looks like this:

```
"x errors occured:
	* error
	* error
	* error
	... \n\n"
```

whereas the former output was

```
"errors: [index: error; index: error; ...]"
```
where `error` is a string that contains the error message, and in many cases additional parts with further details abouth the error.

`hashicorp/go-multierror` is already used in this repo, often in conjunction with the `Errorf` function from [`github.com/pkg/errors`](https://github.com/pkg/errors). Since it is not longer maintained we use the standard lib `fmt.Errorf` in this PR. We also opened an issue to remove the `pkg/errors` from magma #12632.

Since now unused, we delete `multi.go` and respective tests. Plus `grpc_to_http.go`, which was already unused before.

## Test Plan

- [x] Precommit checks in orc8r/cloud and feg/gateway`./build.py -c`
- [x] Build orc8r `./build.py --all`

## Additional Information

Done in pairing with @mpfirrmann.


feg/gateway/services/session_proxy/servicers/multiple_session_controller.go
```diff
@@ -219,12 +220,14 @@ func (srv *CentralSessionControllers) Enable(
 	ctx context.Context,
 	void *orcprotos.Void,
 ) (*orcprotos.Void, error) {
-	multiError := merrors.NewMulti()
+	multiError := &multierror.Error{}
 	for i, controller := range srv.centralControllers {
 		_, err := controller.Enable(ctx, void)
-		multiError = multiError.AddFmt(err, "error(%d):", i+1)
+		if err != nil {
+			multiError = multierror.Append(multiError, errors.Errorf("error(%d): %v", i+1, err))
```
maxhbr: `multiError = multierror.Append(multiError, ...` is weird to read, with `multi{eE}rror` in different casing variants. Is there a neater way to write that or a more distinct name for the variable?

EDIT: This seems to be the expected pattern (see https://github.com/hashicorp/go-multierror/blob/9974e9ec57696378079ecc3accd3d6f29401b3a0/append.go#L39), but the similarity of names is still annoying. Maybe use `errs` as name?

feg/gateway/services/session_proxy/servicers/session_controller.go
```diff
@@ -340,20 +341,20 @@ func (srv *CentralSessionController) Enable(
...
 		if err != nil {
-			multiError.Add(fmt.Errorf("An error occurred while enabling connections; policyClient err: %s", err))
+			multierror.Append(multiError, errors.Errorf("An error occurred while enabling connections; policyClient err: %s", err))
 		}
 	}
 	if !srv.cfg.DisableGy {
 		err := srv.creditClient.EnableConnections()
 		if err != nil {
-			multiError.Add(fmt.Errorf("An error occurred while enabling connections; creditClient err: %s", err))
+			multierror.Append(multiError, errors.Errorf("An error occurred while enabling connections; creditClient err: %s", err))
```
maxhbr: at other locations you wrote the following:
```suggestion
			multiError = multierror.Append(multiError, errors.Errorf("An error occurred while enabling connections; creditClient err: %s", err))
```

this feels like unnecessary error prone syntax.

feg/gateway/services/session_proxy/servicers/session_controller.go
```diff
@@ -340,20 +341,20 @@ func (srv *CentralSessionController) Enable(
 	ctx context.Context,
 	void *orcprotos.Void,
 ) (*orcprotos.Void, error) {
-	multiError := merrors.NewMulti()
+	multiError := &multierror.Error{}
 	if !srv.cfg.DisableGx {
 		err := srv.policyClient.EnableConnections()
 		if err != nil {
-			multiError.Add(fmt.Errorf("An error occurred while enabling connections; policyClient err: %s", err))
+			multierror.Append(multiError, errors.Errorf("An error occurred while enabling connections; policyClient err: %s", err))
```
maxhbr: at other locations you wrote the following:
```suggestion
			multiError = multierror.Append(multiError, errors.Errorf("An error occurred while enabling connections; creditClient err: %s", err))
```

this feels like unnecessary error prone syntax.

maxhbr (changes requested):

sebathomas:
The insync-check is failing, you probably need some more go mod clean-up.

feg/gateway/services/session_proxy/servicers/multiple_session_controller.go
mpfirrmann: Fixed.

feg/gateway/services/session_proxy/servicers/session_controller.go
mpfirrmann: Fixed.

maxhbr:
The pr title prefix is fairly generic and does not match best practices, maybe better would be `refactor(orc8r): `.

orc8r/cloud/go/services/orchestrator/servicers/protected/indexer_servicer.go
```diff
@@ -117,39 +118,48 @@ func setSecondaryStates(ctx context.Context, networkID string, states state_type
...
 func unsetSecondaryStates(ctx context.Context, networkID string, states state_types.StatesByID) (state_types.StateErrors, error) {
 	sessionIDToIMSI, cTeidToHwId, uTeidToHwId, stateErrors := getMappings(states)
-	multiError := merrors.NewMulti()
+	errs := &multierror.Error{}
 
 	err := unsetTeids(ctx, controlPlaneTeid, networkID, cTeidToHwId)
-	multiError = multiError.Add(err)
-
+	if err != nil {
+		errs = multierror.Append(errs, err)
```
maxhbr: multierror already ignores `nil`, no need for the if block: https://github.com/hashicorp/go-multierror/blob/9974e9ec57696378079ecc3accd3d6f29401b3a0/append.go#L9

orc8r/cloud/go/services/orchestrator/servicers/protected/indexer_servicer.go
```diff
@@ -117,39 +118,48 @@ func setSecondaryStates(ctx context.Context, networkID string, states state_type
...
+	}
 	err = unsetTeids(ctx, userPlaneTeid, networkID, uTeidToHwId)
-	multiError = multiError.Add(err)
-
+	if err != nil {
```
maxhbr: same here

orc8r/cloud/go/services/orchestrator/servicers/protected/indexer_servicer.go
```diff
@@ -117,39 +118,48 @@ func setSecondaryStates(ctx context.Context, networkID string, states state_type
...
+		errs = multierror.Append(errs, err)
+	}
 	err = unsetSessionIDs(ctx, networkID, sessionIDToIMSI)
-	multiError = multiError.Add(err)
-
-	return stateErrors, multiError.AsError()
+	if err != nil {
```
maxhbr: same here

maxhbr (changes requested):

maxhbr (approved):
LGTM

uri200 (changes requested):
Why are we removing our own implementation of errors if we already have it?
Why aren't we removing hashicorp instead to avoid that module not being supported in the future? (same that happened with errors module)

Note we also try to maintain the errors in one single line. So printing them in multiple lines is something we should avoid. Otherwise debugging and parsing becomes difficult. (for example you can not use grep properly anymore and you need to look for previous lines to see the full error)

orc8r/cloud/go/services/orchestrator/servicers/protected/indexer_servicer.go
```diff
@@ -117,39 +118,48 @@ func setSecondaryStates(ctx context.Context, networkID string, states state_type
 	if len(sessionIDToIMSI) == 0 && len(cTeidToHwId) == 0 {
 		return stateErrors, nil
 	}
-	multiError := merrors.NewMulti()
+	errs := &multierror.Error{}
 	if len(sessionIDToIMSI) != 0 {
 		err := directoryd.MapSessionIDsToIMSIs(ctx, networkID, sessionIDToIMSI)
-		multiError = multiError.AddFmt(err, "failed to update directoryd mapping of session IDs to IMSIs %+v", sessionIDToIMSI)
+		if err != nil {
```
uri200: i know this has been mentioned below, but just as a placeholder: remove all ifs since `Append` will ignore nils

orc8r/gateway/go/services/magmad/service/magmad.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/aeden/traceroute"
 	"github.com/emakeev/snowflake"
 	"github.com/golang/glog"
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
```
uri200: why are we introducing an unsupported package here

https://github.com/magma/magma/issues/12632

orc8r/cloud/go/services/orchestrator/servicers/protected/indexer_servicer.go
maxhbr: this contains formatting, the `errors. Errorf` result will still not be `nil`, even if `err` was `nil`.

But this 3 line block could be extracted to a function, since they just differ in a few words.

orc8r/gateway/go/services/magmad/service/magmad.go
maxhbr: The ticket #12632 was created as a result of realizing that the unsupported package was used here. I think this is for consistence but I agree that this PR should maybe also fix (parts of) #12632.

alexzurbonsen: @uri200 The package is removed in the final commit of this PR, so this is not part of the changes. (The PR isn't nicely tailored for commit by commit review, sorry.)

maxhbr:
>  Why aren't we removing hashicorp instead to avoid that module not being supported in the future? (same that happened with errors module)

I expect that the chances that this package is better maintained than our own library function are pretty high. It is also well tested and due to its wide usage it is also probably already known to many contributors.

> Note we also try to maintain the errors in one single line.

That is a good point!

orc8r/cloud/go/services/orchestrator/servicers/protected/indexer_servicer.go
alexzurbonsen: Introduced a wrapper in this file.

uri200:
Let's wait for @emakeev , he may have some context on multi error code.

emakeev (changes requested):
Could you please explain what is the purpose of this change?
The usage pattern seems to be nearly identical,
the implementation already exists & in use and hasn't been problematic so far.

Is there any reason, we want to spend time of developers & reviewers to fix something which is not broken while there are so many real improvements Magma can benefit from?

If you just need to change the format of the errors' string (which also has to be justified, BTW), it could be accomplished by a small **4 lines PR vs this 400 lines PR** and keep flexibility to modify the format later anytime you need/want, just change func (me *Multi) Error() string to:


```
func (me *Multi) Error() string {
	if me == nil {
		return "<nil>"
	}
	switch len(me.errors) {
	case 0:
		return ""
	case 1:
		return me.errors[0].Error()
	default:
		var b bytes.Buffer
		fmt.Fprintf(&b, "%d errors occurred:\n", len(me.errors))
		fmtStr := "\t* %v\n"
		for _, e := range me.errors {
			fmt.Fprintf(&b, fmtStr, e)
		}
		b.Write([]byte("\n"))
		return b.String()
	}
}

```

```
--- a/orc8r/lib/go/merrors/multi.go
+++ b/orc8r/lib/go/merrors/multi.go
@@ -41,12 +41,12 @@ func (me *Multi) Error() string {
                return me.errors[0].Error()
        default:
                var b bytes.Buffer
-               fmtStr := "errors: [%d: %v"
-               for i, e := range me.errors {
-                       fmt.Fprintf(&b, fmtStr, i, e)
-                       fmtStr = "; %d: %v"
+               fmt.Fprintf(&b, "%d errors occurred:\n", len(me.errors))
+               fmtStr := "\t* %v\n"
+               for _, e := range me.errors {
+                       fmt.Fprintf(&b, fmtStr, e)
                }
-               b.Write([]byte("]"))
+               b.Write([]byte("\n"))
                return b.String()
        }
 }

```
or something like this...

emakeev:
> I expect that the chances that this package is better maintained than our own library function are pretty high. It is also well tested and due to its wide usage it is also probably already known to many contributors.
> 
note that it also applies to the rest of Magma codebase (merror is maintained as well as the rest of Magma code), the original implementation is simple, controlled by us, allows us to format multi errors ANY way we want or need, now and in the future and it relies only on std Go packages. 
But - the bigger question is - why fix something which is not broken? What's the incentive to do it?

orc8r/gateway/go/services/magmad/service/magmad.go
emakeev: > The ticket #12632 was created as a result of realizing that the unsupported package was used here. I think this is for consistence but I agree that this PR should maybe also fix (parts of) #12632.

Could you please elaborate? We introduce 12627 which we are not sure, we need and then we create #12632 to fix issues that #12627 will create?

maxhbr: Oh,  sorry, that was not formulated clear enough.

#12632 was created while working on #12627. But it will also be solved here for the touched code. So no new instances of #12632 will be introduced and the code commented here is actually fixed in a later commit, as @alexzurbonsen explained.

maxhbr:
Hey @emakeev,

Thank you for your comments. 

There are right now two implementations of multierrors used in Magma:
* our own implementation, used in 5 files and 
* the [hashicorp one](https://github.com/hashicorp/go-multierror), used in 12 files.

We saw an opportunity to clean up the source code and decided to go for the more commonly used one. The hashicorp library is also widely known, better documented and might rely less on tribal knowledge. Migrating to that also preserves the possibility to add arbitrary `ErrorFormat` formatters and drops ~300 lines of source code we needed to maintain. 

Regarding the formatting change:
Since more often than not, the code already uses the default hashicorp formatting with multiline error messages, we think that this would make it at least more consistent.

orc8r/gateway/go/services/magmad/service/magmad.go
alexzurbonsen: Cleaned up the commit history. Now all changes that replace merrors by hasicorp/go-multierror are in one commit.

emakeev:
> Hey @emakeev,
> 
> Thank you for your comments.
> 
> There are right now two implementations of multierrors used in Magma:
> 
> * our own implementation, used in 5 files and
> * the [hashicorp one](https://github.com/hashicorp/go-multierror), used in 12 files.
> 
> We saw an opportunity to clean up the source code and decided to go for the more commonly used one. The hashicorp library is also widely known, better documented and might rely less on tribal knowledge. Migrating to that also preserves the possibility to add arbitrary `ErrorFormat` formatters and drops ~300 lines of source code we needed to maintain.
> 
> Regarding the formatting change: Since more often than not, the code already uses the default hashicorp formatting with multiline error messages, we think that this would make it at least more consistent.

Sure, it seems to be reasonable to settle on a single implementation & since the work is already done, I don't see an issue w it. I'll leave it to @uri200 to decide if the change to multi line format is acceptable for FeG.

emakeev (approved):

uri200 (approved):

uri200:
you need to fix the insych issue 
Maybe you just need to run `./build.py -g`

PR #12631, opened 2022-05-04
feat(dp): changed python CBSD model

WojSad:

feat(dp): changed python CBSD model

## Summary
Added a number of fields needed for future single-step registration

## Test Plan

migration tests included


dp/cloud/python/magma/db_service/models.py
```diff
@@ -222,7 +223,21 @@ class DBCbsd(Base):
     preferred_frequencies_mhz = Column(
         JSON, nullable=False, server_default=sa_text("'[]'::json"),
     )
+    single_step_registration = Column(Boolean, nullable=False, server_default='false')
```
jkmar: maybe call it `single_step_enabled` to indicate that it is a flag rather than a value

dp/cloud/python/magma/db_service/models.py
```diff
@@ -222,7 +223,21 @@ class DBCbsd(Base):
...
+    cbsd_category = Column(String, nullable=False, server_default='B')
     network_id = Column(String)
+    latitude = Column(Float)
+    longitude = Column(Float)
+    height = Column(Float)
```
jkmar: it would be nice to add suffix with units to these fields

dp/cloud/python/magma/db_service/models.py
```diff
@@ -222,7 +223,21 @@ class DBCbsd(Base):
...
+    height_type = Column(String)
+    horizontal_accuracy = Column(Float)
+    antenna_azimuth = Column(Integer)
+    antenna_downtilt = Column(Integer)
+    antenna_beamwidth = Column(Integer)
+    antenna_model = Column(String)
+    eirp_capability = Column(Integer)
+    cpi_digital_signature = Column(Text)
```
jkmar, line 239: Do all of these fields have to be null? Nulls are more difficult to work with than non nulls.

jkmar (approved):

PR #12633, opened 2022-05-04
feat(ci): add federated integ test to ci

uri200:

## Summary

This PR adds Federated Integ Test to CI. This task is a modification of l[te-integ-test.](http://github.com/magma/magma/blob/master/.github/workflows/lte-integ-test.yml#L3-L3)

Even this works there are possible enhancements to make it faster or more reliable:
1. Modify this job to run in a ubuntu box. That will allow us to install docker on the ubuntu box. This may requiere changes on the fab script and on Vagrant vms to make sure they work in both, linux and MacOS (remember the test needs to run on MacOs so developers can use it locally)

2. Modify s1ap federated integ test to run Orc8r inside the magma Vagrant machine. That may work but that would make a mess on the local scenario. Also this doesn't assure that the runner takes hours to run due to low ram memory (docker over vagrant over the runner) 

3. Run S1ap Federated integ test on a AWS node that can support Virtualization

Enhancements
- Feg and Orc8r are build previously by other jobs. Those images are uploaded to magma core jfrog docker registry. Instead of building them again in this task, we could point the vm docker to the JFROG repo and grab the already built image from there.

## Test Plan

For this test we triggered the job to run Federated Integ test on PRs. This is triggered here https://github.com/magma/magma/pull/12633/files#diff-6a95f663e4dfd08ee0988d079f574d7c7782a84b3b3ba2459a8b4891e48ca944R15-R24 . But once this work, we can remove those lines and just execute this on merge. This code is just here so we are able to execute the test and troubleshoot it

image

## Additional Information

### DIFF
--- a/.github/workflows/federated-integ-test.yml
+++ b/.github/workflows/federated-integ-test.yml
@@ -12,22 +12,67 @@ on:  # yamllint disable-line rule:truthy
       - 'v1.*'
     types:
       - completed
-  push:
-    branches:
-      - master
-      - 'v1.*'
-  pull_request:
-    branches:
-      - master
-      - 'v1.*'
-    types: [opened, reopened, synchronize]
 
 jobs:
+  # Build images on ubuntu which is faster than MacOs.
+  docker-build-orc8r:
+    runs-on: ubuntu-latest
+    env:
+      MAGMA_ROOT: "${{ github.workspace }}"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build Orc8r docker images
+        run: |
+          cd orc8r/cloud/docker
+          ./build.py --deployment all
+          docker images
+      - name: Export docker images to deploy them
+        run: |
+          mkdir images
+          cd images
+          docker images
+          docker save orc8r_nginx:latest | gzip > fed_orc8r_nginx.tar.gz
+          docker save orc8r_controller:latest  | gzip > fed_orc8r_controller.tar.gz
+          docker save orc8r_fluentd:latest  | gzip > fed_orc8r_fluentd.tar.gz
+          docker save orc8r_test:latest  | gzip > fed_orc8r_test.tar.gz
+      - uses: actions/upload-artifact@v2
+        with:
+          name: docker-build-orc8r-images
+          path: images
+
+  docker-build-feg:
+    runs-on: ubuntu-latest
+    env:
+      MAGMA_ROOT: "${{ github.workspace }}"
+    steps:
+      - uses: actions/checkout@v2
+      - name: pre requisites to build feg
+        run: |
+          cd ${{ env.MAGMA_ROOT }} && mkdir -p .cache/test_certs/ && mkdir -p .cache/feg/
+          cd ${{ env.MAGMA_ROOT }}/.cache/feg/ && touch snowflake
+      - name: Build FEG docker images
+        run: |
+          cd feg/gateway/docker
+          docker-compose build --force-rm --parallel
+          docker images
+      - name: Export docker images to deploy them
+        run: |
+          mkdir images
+          cd images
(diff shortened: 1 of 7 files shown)


uri200:
macos does not support docker. It can not be installed properly on a runner

maxhbr:
maybe use a minikube or similar as docker host. But it is not a no-brainer. I will think about that

.github/workflows/federated-integ-test.yml
```diff
@@ -0,0 +1,115 @@
...
+
+      - name: build_all
+        run: |
+          cd lte/gateway/python/integ_tests/federated_tests
+          export MAGMA_DEV_CPUS=3
+          export MAGMA_DEV_MEMORY_MB=9216
+          fab build_all:orc8r_on_vagrant=True
+
+
+      #- name: Run the integ test
```
maxhbr: this is currently a Draft?

uri200: We are testing some stuff on CI.

Let me add Work in progress tag

Neudrino: Hi @uri200,
if this is a draft, you could consider to convert it to a draft PR, to get it of peoples review list. :wink: 
grafik

.github/workflows/federated-integ-test.yml
```diff
@@ -0,0 +1,201 @@
...
+#      - completed
+#  push:
+#    branches:
+#      - master
+#      - 'v1.*'
+#  pull_request:
+#    branches:
+#      - master
+#      - 'v1.*'
+#    types: [opened, reopened, synchronize]
```
uri200: this is commented out to avoid this to be run
This will be fixed  by 
https://github.com/magma/magma/pull/13012

emakeev (approved):
LGTM

koolzz (approved):

.github/workflows/federated-integ-test.yml
```diff
@@ -0,0 +1,201 @@
...
+    branches:
+      - master
+      - 'v1.*'
+    types:
+      - completed
+  push:
+    branches:
+      - master
+      - 'v1.*'
+  pull_request:
```
crasu: I guess this has to go before we merge this

.github/workflows/federated-integ-test.yml
```diff
@@ -0,0 +1,201 @@
...
+          cd lte/gateway
+          export MAGMA_DEV_CPUS=3
+          export MAGMA_DEV_MEMORY_MB=9216
+          fab federated_integ_test:build_all=False,orc8r_on_vagrant=True
+      - name: Get test results
+        if: always()
+        run: |
+          cd lte/gateway
+          fab get_test_summaries:dst_path="test-results"
+          ls -R
```
crasu, line 149: Do we need the ls here?

crasu (approved):

ardzoht (approved):

.github/workflows/federated-integ-test.yml
uri200: yep, this si gone now

.github/workflows/federated-integ-test.yml
uri200: I am not totally sure. Per my understanding that was being used by `magma-ci`. So I guess that next step is to integrate it with magma-ci and use those summaries

crasu: ok fine

crasu:
@uri200 I guess we can merge this now

uri200:
> @uri200 I guess we can merge this now

Hey @crasu  for some reason it does not let me merge. Not sure if there is a missing approval ?

uri200:
@Neudrino  can you approve this PR to see if then I can merge it?

tmdzk (approved):

PR #12638, opened 2022-05-05
chore(ci): merge bazel base and devcontainer - first pr

nstng:
## Summary

This change makes the devcontainer image be build on top of the bazel base image.
* until now a lot of changes needed to be done in both images
* bazel base provides the main functionality for enabling builds with Bazel
* the devcontainer also provides QoL features for developers in vscode and also a couple of dependencies for make builds

This change is split up in 4 PRs - this is the first one
1. largest change - commits:
 a. move the bazel base image from experimental to .devcontainer - the bazel base image is not experimental anymore
 b. merge the two images and build the devcontainer image on top of bazel base. basically remove duplicated apt dependencies and change building from scratch to magma core repository dependencies when possible.
 c. build the bazel base and devcontainer image in one CI job (sequentially)
 d. do not set the latest tag for newly created bazel base and devcontainer images - set latestv2 instead. **This will cause current PRs to still use the latest images! I.e., current PRs will not fail because of not compatible images! Also jobs on master will still use latest and thus they will not fail because the new images are not created immediately.**
2. make CI workflows use latestv2 images - this will cause existing PR workflows to still use the latest tag (i.e., they will not fail) and new and rebased PRs to use the latestv2 images. This PR should be created shortly after the first PR when the image build job run successfully.
3. The image create job will use the latest tag again. This PR will be created after a couple of days when existing PRs are sufficiently created from (or rebased on) master.
4. CI workflows use the latest tag again - this can be done shortly after the third PR when the build job was run successfully.

## Test Plan

* Please note: `Build Docker image for Bazel Base and DevContainer` is expected to fail on this PR as the bazel base image is here not pushed to the magma github repo - this should succeed on master - see next bullet point for test on fork.
* run build job on fork https://github.com/nstng/magma/actions/runs/2275637718 and see that packages are created with correct tag(s) https://github.com/nstng?tab=packages&repo_name=magma (in `.devcontainer/Dockerfile` the `FROM` property was pointing temporally to fork repo)
* start vscode using the newly created devontainer - `ghcr.io/nstng/magma/devcontainer:latestv2`
* run relevant devcontainer and bazel base workflows on fork with new images (relevant for PR 2-3)
  * https://github.com/nstng/magma/actions/workflows/bazel.yml
  * https://github.com/nstng/magma/actions/workflows/gcc-problems.yml
  * https://github.com/nstng/magma/actions/workflows/bazel-cache-push.yml
  * https://github.com/nstng/magma/actions/workflows/agw-workflow.yml

## Additional Information


.devcontainer/Dockerfile
```diff
@@ -42,124 +45,60 @@ RUN echo "Install general purpose packages" && \
...
+        clangd-12 \
         g++-9 \
         gcc-9 \
-        # iputils-ping is a dependency of python services (e.g. magmad)
-        iputils-ping \
-        lcov \
         gdb \
-        git \
-        gnupg \
+###        gnupg \
```
Neudrino: There is a commented out package here. What is the intention?
Could we either install it or remove the line?

CODEOWNERS
```diff
@@ -69,7 +69,7 @@
 /.github/workflows/bazel.yml @magma/approvers-bazel
 /.github/workflows/docker-builder-bazel-base.yml @magma/approvers-bazel
 /bazel/ @magma/approvers-bazel
-/experimental/bazel-base/ @magma/approvers-bazel
+/.devcontainer/bazel-base/ @magma/approvers-bazel
```
Neudrino: Changes of the CODEOWNER file are pulling a very special set of reviewers (with limited availability) into the PR, which might not be necessary for the other functional changes. So in the interest of quick review and progress, I suggest to put this oneline change in a separate PR. That would also have the advantage for this specific review group to only see quick and easy the relevant code change.

.devcontainer/bazel-base/Dockerfile
```diff
@@ -0,0 +1,110 @@
+################################################################
```
Neudrino, line 1: This file does not show as moved, but as deleted and newly added. This makes it hard to make out the changed differences. Would it be possible to apply a move action in git (`git mv --help`) as to potentially retain good diff?

.devcontainer/bazel-base/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+# Builder Image (can also be used as developer's image)
+################################################################
+FROM ubuntu:focal as bazel_builder
+
+ARG PYTHON_VERSION=3.8
+
+ENV TZ=America/New_York \
```
Neudrino: Is there a reason to prefer this timezone in favour of any other one?
For a globally developed and used project, I think UTC should be used.

.devcontainer/bazel-base/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+    DEBIAN_FRONTEND=noninteractive
```
Neudrino: I found an [article which discourages](https://bobcares.com/blog/debian_frontendnoninteractive-docker/) the use of this setting.
The conclusion there was, that if at all, it should only be set as `ARG` or prefacing commands.

.devcontainer/bazel-base/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+
+#  MAGMA_ROOT is needed by python tests (e.g. freedomfi_one_tests in enodebd)
+ENV MAGMA_ROOT=/workspaces/magma
+
+RUN echo "Install general purpose packages" && \
+    apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get upgrade -y && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
```
Neudrino: However the usage of `DEBIAN_FRONTEND=noninteractive` might be in general, this would be a duplication to the above setting of `ENV` (or `ARG`). I.e. either set it above or here.
I would be in favour of using `apt-get -q` instead.

.devcontainer/bazel-base/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+        # dependency of pip systemd
+        libsystemd-dev \
+        lld \
+        # dependency of python services (e.g. magmad)
+        net-tools \
+        # dependency of python services (e.g. pipelined)
+        netbase \
+        python${PYTHON_VERSION} \
+        python-is-python3 \
+        # neded for post install script of libfolly-dev
```
Neudrino: Spelling `needed`.

.devcontainer/bazel-base/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+        systemd \
+        unzip \
+        # dependency of liagent
+        uuid-dev \
+        vim \
+        wget \
+        zip
+
+RUN echo "Install Folly dependencies" && \
+    apt-get install -y --no-install-recommends \
```
Neudrino, line 69: Inconsistent usage of `DEBIAN_FRONTEND=noninteractive`. Also further below.
As above I am in favour of dropping the debian frontend thing and keep it here as is.
Command line parameters for apt should be synced though.

.devcontainer/bazel-base/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+    libdouble-conversion-dev \
+    libboost-chrono-dev \
+    libiberty-dev && \
+    echo "Install libtins and connection tracker dependencies" && \
+    apt-get install -y --no-install-recommends \
+    libpcap-dev=1.9.1-3 \
+    libmnl-dev=1.0.4-2
+
+RUN apt-get -y install --no-install-recommends \
+    libtool=2.4.6-14
```
Neudrino: This is inconsistent in so far, that we use one `RUN` for multiple apt install commands above but not include this one.
Unless there is a specific reason, it should be included in the before `RUN`. An additional `echo` to describe it would be optional for me, but welcome.

Neudrino: I wonder why yamllint does not complain about to RUNs after each other? :man_shrugging:

.devcontainer/bazel-base/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+
+# setup magma artifactories and install magma dependencies
+RUN echo 'deb https://artifactory.magmacore.org/artifactory/debian-test focal-ci main' > /etc/apt/sources.list.d/magma.list && \
+    echo 'deb https://artifactory.magmacore.org/artifactory/debian-test/ focal-1.7.0 main' > /etc/apt/sources.list.d/magma.list && \
```
Neudrino: I would usually prefer the `add-apt-repository` command over a manual editing of those files manually. I would assume that the command has additional helpful assertions for that action.

.devcontainer/bazel-base/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+    wget -qO - https://artifactory.magmacore.org:443/artifactory/api/gpg/key/public -O /tmp/public && \
+    apt-key add /tmp/public && \
```
Neudrino: Avoiding a temporary file which clutters the image, it can be shortened to
```
wget -qO - https://artifactory.magmacore.org:443/artifactory/api/gpg/key/public | apt-key add -
```

.devcontainer/Dockerfile
```diff
@@ -42,124 +45,60 @@ RUN echo "Install general purpose packages" && \
     apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get upgrade -y && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y \
```
Neudrino: As below/above `DEBIAN_FRONTEND=noninteractive`...

.devcontainer/Dockerfile
```diff
@@ -42,124 +45,60 @@ RUN echo "Install general purpose packages" && \
...
     echo "Install Nettle requirements" && \
-    apt-get install -y libgoogle-glog-dev libconfig-dev libxml2-dev \
+    apt-get install -y libxml2-dev \
     libyaml-cpp-dev nlohmann-json3-dev && \
-    echo "Install ZeroMQ" && \
-    apt-get install -y libczmq-dev && \
-    echo "Install dependencies for Sentry Native" && \
-    apt-get install -y libcurl4-openssl-dev
+    echo "GRPC and it's dependencies" && \
+    apt-get install -y grpc-dev
```
Neudrino: - There is inconsistency regarding the apt-get parameters here. They should be aligned for the installations in this file.
- Above there is a one-package-per-line philosophy followed, which is much more readable. I think we should also follow this approach here.

CODEOWNERS
nstng: and also the `Codeowners Validator / sanity (pull_request)` does not seem to like adding approvers for files not on master :) will pull thiss into a follw-up

.devcontainer/bazel-base/Dockerfile
nstng: TZ is needed for tzdata. DEBIAN_FRONTEND=noninteractive disables the manual configuration for tzdata. Changed to approach in https://rtfm.co.ua/en/docker-configure-tzdata-and-timezone-during-build/ for doing the configuration.

.devcontainer/bazel-base/Dockerfile
nstng: see above

.devcontainer/bazel-base/Dockerfile
nstng: `git mv` was used and it shows up in my ide correctly - where does it now show as moved?

.devcontainer/Dockerfile
nstng: removed

CODEOWNERS
nstng: removed change from commit

Neudrino (approved):

nstng:
for reviewers - forced pushed: bazel base is again building folly from scratch and does not install via magma artifactory. Reason: the linking in bazel would change from static to dynamic linking and this corrupts the caches , i.e., builds that pull caches from github will fail.

.devcontainer/bazel-base/Dockerfile
```diff
@@ -0,0 +1,130 @@
...
+    make install && \
+    cd / && \
+    rm -rf fmt
+
+# Facebook Folly C++ lib
+# Note: "Because folly does not provide any ABI compatibility guarantees from
+#        commit to commit, we generally recommend building folly as a static library."
+# Here we checkout the hash for v2021.02.22.00 (arbitrary recent version)
+RUN git clone https://github.com/facebook/folly && cd folly && \
+    git checkout tags/v2021.02.15.00 && \
```
Neudrino: I would prefer to use the full capability of `git clone ${YOUR_REPO} --branch ${CHECKOUT_BRANCH}` in scripts to be more efficient.

.devcontainer/bazel-base/Dockerfile
```diff
@@ -0,0 +1,130 @@
...
+    mkdir _build && cd _build && \
+    cmake .. && \
+    make -j"$(nproc)" && \
+    make install && \
+    cd / && \
```
Neudrino, line 104: Comparing to line 97 (see comment above), where a relativ change of directory was made, I would also make a `cd ..` here. Or use absolute path in line 97 as well, e.g. `cd /folly`.

Neudrino (commented):
Was the folly code copied as is from somewhere?
On might want to verify that the changes of directory are actually correct.
And mybe also have one bash command per line as a style improvement.

nstng:
> Was the folly code copied as is from somewhere? On might want to verify that the changes of directory are actually correct. And mybe also have one bash command per line as a style improvement.

yes, this is the code from the current bazel base Dockerfile

LKreutzer (approved):
Reviewed in pairing

PR #12666, opened 2022-05-10
refactor(orc8r): Omit nil-checks for multierror.Append calls

wolfseb:
## Summary

`multierror.Append` already handles nil-checks, so checking before calling it is unnecessary, except when the error is formatted with e.g. `fmt.Errorf`. This PR follows this [discussion](https://github.com/magma/magma/pull/12627#discussion_r866240479).

## Test Plan

Run unit tests

## Additional Information

### DIFF
--- a/orc8r/cloud/go/syncstore/store_writer.go
+++ b/orc8r/cloud/go/syncstore/store_writer.go
@@ -196,9 +196,7 @@ func (l *syncStore) collectGarbageSQL(tracked []string) error {
 			return nil, err
 		}
 		_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
-		if err != nil {
-			errs = multierror.Append(errs, errors.Wrapf(err, "collect garbage for table %+v", tableName))
-		}
+		errs = multierror.Append(errs, errors.Wrapf(err, "collect garbage for table %+v", tableName))
 	}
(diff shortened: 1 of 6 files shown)


orc8r/cloud/go/syncstore/store_writer.go
```diff
@@ -196,9 +196,7 @@ func (l *syncStore) collectGarbageSQL(tracked []string) error {
 			return nil, err
 		}
 		_, err := sqorc.ExecInTx(l.db, nil, nil, txFn)
-		if err != nil {
-			errs = multierror.Append(errs, errors.Wrapf(err, "collect garbage for table %+v", tableName))
-		}
+		errs = multierror.Append(errs, errors.Wrapf(err, "collect garbage for table %+v", tableName))
```
alexzurbonsen, line 199: Removing this nil check might make things more complicated if we intend to get rid of errors.Wrapf later? We will have to reintroduce it then, I suppose.

uri200 (approved):

orc8r/cloud/go/syncstore/store_writer.go
wolfseb: Yes, you are right. Since it is merged now we'll have to change it in the PR that gets rid of Wrapf (#12682); I will arrange that.

PR #12668, opened 2022-05-10
refactor(nms): move fbc lib logging

Neudrino:
## Summary

- Logging is more or less a global cross cutting concern an thus deserves a top level folder.
- Depends on refactoring in https://github.com/magma/magma/pull/12656
  - The `require(...)` statements will be refactored in that PR.

## Test Plan

- `yarn run eslint . --quite`
- `docker compose build && docker compose up`


thmsschmitt (changes requested):
In retrospect moving this to `/shared` would make more sense. Sorry 😇

thmsschmitt (approved):

PR #12684, opened 2022-05-11
fix(nms): Sidebar version number overlaps with profile button

thmsschmitt:
## Summary
When the sidebar is expanded the version number partially overlaps with the profile button and blocks the lower part of the click area. 

version-cover

## Test Plan
Check that the bug described above is fixed.


andreilee (approved):
Please show screenshots for before/after next time.

PR #12685, opened 2022-05-11
ci(dp): DP Makefile and Actions Workflow  improvements

sklgromek:
## Summary
- Makefile targets cleanup was made
- Makefile target `integration_tests` was created for integration tests as a replacement for triggering them with `CI=true` env variable
- Adding ` _switch_k8s_context` target as protection, to running only on miniukube context

### DIFF
--- a/.github/workflows/dp-workflow.yml
+++ b/.github/workflows/dp-workflow.yml
@@ -382,6 +382,25 @@ jobs:
         run: |
           make ${{ matrix.test_type }}
 
+      - name: Collect Pods logs
+        if: always()
+        run: >-
+          mkdir /tmp/k8s-pods-logs;
+          for pod in $(kubectl get pods -o custom-columns=NAME:.metadata.name --no-headers);
+          do
+            if [[ "$pod" == "postgres-database-0" ]]; then
+              kubectl logs $pod --container postgres-test > /tmp/k8s-pods-logs/${pod}.log;
+            else
+              kubectl logs $pod > /tmp/k8s-pods-logs/${pod}.log;
+            fi;
+          done
+      - name: Collect Elasticsearch data
+        if: always()
+        run: >
+          kubectl exec
+          $(kubectl get pods -o custom-columns=NAME:.metadata.name | grep elasticsearch)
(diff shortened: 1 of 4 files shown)


dp/tools/deployment/vendor/elasticsearch.yml
```diff
@@ -24,6 +24,12 @@ spec:
           env:
             - name: discovery.type
               value: single-node
+          readinessProbe:
```
jkmar, line 27: Let's maybe investigate flake first before proposing this change, I'm not sure if it helps

jkmar (approved):

dp/tools/deployment/vendor/elasticsearch.yml
xbend: While it was not proven explicitly (we did not notice the integ tests failing) we did see an indication that this fix removes the flakiness. We saw that fluentd-forward did have problems connecting to ES, and entered a long backoff connection retry.
The test managed to succeed, but we did see backoff timeouts kicking in on fluentd-forward due to ES unavailability.

xbend (approved):

xbend:
@magma/approvers-ci 
This PR is now ready for review and merge. It should solve most if not all problems with flaky Domain Proxy integration test runs so the quicker we get it into master, the better.

.github/workflows/dp-workflow.yml
```diff
@@ -390,6 +409,8 @@ jobs:
           path: |
             ${{ env.TEST_DIR }}
             /tmp/${{ matrix.test_type }}.txt
+            /tmp/k8s-pods-logs
```
nstng, line 412: NIT: is this more robust if results files are collected in a dedicated /tmp/foo folder and this folder is uploaded? I.e., omit to list all files created in different places.

nstng (approved):
CI changes (dp-workflow.yml) lgtm - artifacts are uploaded successfully.

.github/workflows/dp-workflow.yml
sklgromek: Good point. I will improve it in another PR.

PR #12686, opened 2022-05-11
chore: Add size attribute to all bazel test targets

LKreutzer:

## Summary

- Add "size" attribute to every bazel test target.
- This sets the correct timeouts for all unit tests - see [bazel docs](https://bazel.build/reference/be/common-definitions#size).
- If a test took >~10s I used the "medium" label in order to avoid a one minute timeout on slow machines.

## Test Plan

- Check that all tests have the attribute:
```
bazel query 'attr(size, small , tests(//...))' > tests-with-size.txt && \
bazel query 'attr(size, medium , tests(//...))' >> tests-with-size.txt && \ 
bazel query 'attr(size, large , tests(//...))' >> tests-with-size.txt && \
bazel query 'kind(".*_test", attr("name",".*","//..."))' > anytest.txt && \ 
diff tests-with-size.txt anytest.txt
```
- Run all unit tests `bazel test //...`
- CI

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/test/gtpv2-c/BUILD.bazel
+++ b/lte/gateway/c/core/oai/test/gtpv2-c/BUILD.bazel
@@ -15,6 +15,7 @@ package(default_visibility = ["//visibility:private"])
 
 cc_test(
     name = "gtpv2_fteid_test",
+    size = "small",
     srcs = [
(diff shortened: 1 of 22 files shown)


nstng (approved):
lgtm, thank you for the detailed review queries

PR #12687, opened 2022-05-11
feat(cleanup):cleanup old registry

tmdzk:
Signed-off-by: backport-bot <tim@magmacore.org>

feat(cleanup):cleanup old registry

## Summary

When we cleanup the registry we focused on the agw folders, we totally forgot that some of the other components were also using it, I did some cleanup and tested everything locally, everything was working fine

## Test Plan

Most of my changes affect builds, if build all is successful then we can ship

## Additional Information

- [x] It was painful


ShubhamTatvamasi (approved):

.devcontainer/Dockerfile
```diff
@@ -344,7 +344,7 @@ RUN ${PYTHON_VENV_EXECUTABLE} -m pip install --quiet --upgrade --no-cache-dir "s
 ### install python3-aioeventlet and bcc-tools from the magma apt repo
 COPY orc8r/tools/ansible/roles/pkgrepo/files/jfrog.pub /tmp/
 RUN apt-key add /tmp/jfrog.pub && \
-    echo "deb https://facebookconnectivity.jfrog.io/artifactory/list/dev-focal/ focal main" > /etc/apt/sources.list.d/fbc.list && \
+    echo "deb https://artifactory.magmacore.org/artifactory/debian-test focal-ci main" > /etc/apt/sources.list.d/fbc.list && \
```
nstng: python3-aioeventlet seems to differ from the version in the fb artifactory - the patch in line 354+ fails. It looks like the patch is already applied in the magma artifactory version -> the patching step needs to be removed

nstng: NIT: the magmacore repo is still written into fbc.list - this applies for multiple changes in this PR

.devcontainer/Dockerfile
```diff
@@ -344,7 +344,7 @@ RUN ${PYTHON_VENV_EXECUTABLE} -m pip install --quiet --upgrade --no-cache-dir "s
 ### install python3-aioeventlet and bcc-tools from the magma apt repo
 COPY orc8r/tools/ansible/roles/pkgrepo/files/jfrog.pub /tmp/
 RUN apt-key add /tmp/jfrog.pub && \
```
nstng, line 217: Does it make sense to use here (and in other Dockerfiles where possible)
`RUN wget -qO - https://artifactory.magmacore.org:443/artifactory/api/gpg/key/public | apt-key add -`
so that the jobs in this PR are not dependent on the not yet updated `jfrog.pub`.

nstng (changes requested):
I still see the fb artifactory in use in
cwf/gateway/docker/go/Dockerfile
feg/gateway/docker/go/Dockerfile
are these relevant for this change?

lte/gateway/docker/services/c/Dockerfile
```diff
@@ -45,7 +45,7 @@ RUN apt-get update && \
 
 COPY orc8r/tools/ansible/roles/pkgrepo/files/jfrog.pub /tmp/jfrog.pub
 RUN apt-key add /tmp/jfrog.pub && \
-    apt-add-repository "deb ${EXTRA_REPO} ${OS_RELEASE} main"
```
ymasmoudi, line 48: Err:5 https://artifactory.magmacore.org/artifactory/debian-test focal-ci InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 9CECAD4616DD1B36
Reading package lists...
W: GPG error: https://artifactory.magmacore.org/artifactory/debian-test focal-ci InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 9CECAD4616DD1B36

ymasmoudi (changes requested):

lte/gateway/docker/services/c/Dockerfile
tmdzk: IT needs https://github.com/magma/magma/pull/12689/files to be merged

tmdzk:
> I still see the fb artifactory in use in cwf/gateway/docker/go/Dockerfile feg/gateway/docker/go/Dockerfile are these relevant for this change?

https://github.com/magma/magma/pull/12688 Did another PR cause I'm bumping the OS version

.devcontainer/Dockerfile
tmdzk: It makes sense, honestly I was just lazy and just copy pasted, it was painful to go through all the sources haha

nstng:
@tmdzk you need to rebase because of https://github.com/magma/magma/pull/12638 (but this actually should make the change smaller because here the magmacore repos are already used for the touched Dockerfiles).

PR #12689, opened 2022-05-11
fix(agw): Get rid of old registries

tmdzk:
Signed-off-by: tmdzk <me@tmdzk.com>

fix(agw): Get rid of old registries

## Summary

Get rid of the old registry in the local VM 

This is needed for PR #12688 and #12687 but it will break build when it lands that's why we need 12688 and 12687 they are in different PR because it's gonna be easier to rollback in case of further probllems

PR #12688, opened 2022-05-11
feat(cleanup):bump ubuntu from 18(bionic) to 20(focal)

tmdzk:
Signed-off-by: backport-bot <tim@magmacore.org>

feat(cleanup):bump ubuntu from 18(bionic) to 20(focal)

## Summary

In the build process we download specific packages from the registry, those dockerfile are using the old one. 
In order to use the new one that only support ubuntu 20 i'm bumping the images from bionic to focal 

(In order to migrate everything to the new registry)

## Test Plan

Let ci build everything and make sure it's green 

## Additional Information

- [x] No Additional Information

PR #12690, opened 2022-05-11
feat(ci): Pull from registry

tmdzk:
Pull from registry

## Summary

When doing integ test, pulling from registry

## Test Plan

Let CI go 

## Additional Information
DO NOT MERGE BEFORE ALL PR BELOW ARE MERGED
- [ ] #12687
- [ ] #12688
- [ ] #12689

### DIFF
--- a/circleci/fabfile.py
+++ b/circleci/fabfile.py
@@ -266,7 +266,7 @@ def _get_test_summaries_and_logs(test_result_code: int):
 def _run_remote_lte_integ_test(repo: str, magma_root: str):
     repo_name = _get_repo_name(repo)
     with cd(f'{repo_name}/{magma_root}/lte/gateway'):
-        test_result = run('fab integ_test', timeout=180 * 60, warn_only=True, build=False)
+        test_result = run('fab integ_test:build=False', timeout=180 * 60, warn_only=True)
 
         # Transfer test summaries into current directory
         run('fab get_test_summaries:dst_path="test-results"', warn_only=True)
--- a/circleci/fabfile.py
+++ b/circleci/fabfile.py
@@ -1,0 +1,429 @@
+"""
+Copyright 2020 The Magma Authors.
+
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


lte/gateway/fabfile.py
```diff
@@ -270,8 +271,8 @@ def integ_test(
         `magma_trfserver` vagrant box.
     """
 
-    destroy_vm = bool(strtobool(destroy_vm))
-    provision_vm = bool(strtobool(provision_vm))
+    destroy_vm = False
+    provision_vm = False
```
uri200, line 275: i think you forgot to set this as it was

circleci/fabfile.py
```diff
@@ -0,0 +1,429 @@
+"""
```
uri200, line 1: why is this file at circleci folder?

lte/gateway/fabfile.py
```diff
@@ -250,7 +250,8 @@ def federated_integ_test(
 
 def integ_test(
     gateway_host=None, test_host=None, trf_host=None,
-    destroy_vm='True', provision_vm='True',
+    destroy_vm='True', build=None, hash=None,
+    provision_vm='True',
```
uri200, line 254: can we just use `hash` or `github_hash`?

If hash is empty, then you build it from source. If we have a hash, then pull it from repo

```
def integ_test(
    gateway_host=None, test_host=None, trf_host=None,
    destroy_vm='True', github_hash='',
    provision_vm='True',
```

lte/gateway/fabfile.py
```diff
@@ -284,11 +285,16 @@ def integ_test(
     else:
         ansible_setup(gateway_host, "dev", "magma_dev.yml")
         gateway_ip = gateway_host.split('@')[1].split(':')[0]
+    print("bbuild is set to: %s" % build)
```
uri200, line 288: maybe move this inside the if and only print when we are building from a hash. Otherwise it is the regular flow of building from source

uri200: In my opinion, this would be more readable this way.

```
    if github_hash:
	print("Building from github hash:  %s" % github_hash)
	execute(_from_registry)
   else:
        execute(_dist_upgrade)
        execute(_build_magma)
        execute(_run_sudo_python_unit_tests)
        execute(_start_gateway)
```

uri200:
By the way, is there any plan to do the same of pulling from the registiry but for feg and orc8r?

uri200 (commented):
is it possible to include on this PR lte-integ test github action changes so it starts using this ?

PR #12696, opened 2022-05-11
fix(pipelined): use precise match filter for searching dhclient process.

pshelar:
On some hosts pgrep is returning self process ID, which can result
in exception when pipelineD tries to kill the process.
Following PR improves the pgrep match criteria to fix the issue.

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/uplink_bridge.py
+++ b/lte/gateway/python/magma/pipelined/app/uplink_bridge.py
@@ -476,7 +476,7 @@ class UplinkBridgeController(MagmaController):
 
         # Kill dhclient if running.
         pgrep_out = subprocess.Popen(
-            ["pgrep", "-f", "dhclient.*" + if_name],
+            ["pgrep", "-f", "^/sbin/dhclient.*" + if_name],
             stdout=subprocess.PIPE,
         )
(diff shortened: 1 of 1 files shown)


lte/gateway/python/magma/pipelined/app/uplink_bridge.py
```diff
@@ -476,7 +476,7 @@ def _kill_dhclient(self, if_name: str, af_inet: int):
 
         # Kill dhclient if running.
         pgrep_out = subprocess.Popen(
-            ["pgrep", "-f", "dhclient.*" + if_name],
+            ["pgrep", "-f", "^/sbin/dhclient.*" + if_name],
```
nstng, line 479: Are we sure that `/sbin/dhclient` is used for starting the processes? On my host:
```
sudo dhclient docker0
---
$ pgrep -fa dhclient.*docker0
1487058 sudo dhclient docker0
1487059 dhclient docker0
1487060 dhclient docker0
```
This is, `$ pgrep -f ^/sbin/dhclient.*docker0` is empty for me.

pshelar: On AGW dhclient is started by sys init which uses above signature.

nstng (approved):

PR #12697, opened 2022-05-12
refactor(nms): Remove usages of tabs on organisations and user entities

thmsschmitt:
## Summary

It used to be possible to configure accessible tabs for users and organisations. 
This no longer makes sense since all users have the same navigation.
All usages of the fields have been removed.

The fields will be removed from the entities later together with a migration. 

## Test Plan

- Make sure the navigation in existing organisations still works, also check for new users in existing organisations.
- Make sure the navigation work in new organisations.
- Make sure that rebuilding the docker container with an empty database still works.
- Make sure that organisations can be create via script.


nms/app/components/Main.js
```diff
@@ -79,20 +71,48 @@ function Main() {
...
   return (
     <AppContextProvider>
       <IndexWithoutNetwork />
     </AppContextProvider>
   );
 }
 
-export default () => (
-  <ApplicationMain>
+export default () => {
```
thmsschmitt, line 89: I must say having admin and settings under `/nms/{network_Id}` is rather annoying 😕

nms/fbc_js_core/ui/components/ProfileButton.js
```diff
@@ -126,7 +128,7 @@ const ProfileButton = (props: Props) => {
               onClick={() => {
                 GeneralLogger.info(Events.SETTINGS_CLICKED);
                 props.setMenuOpen(false);
-                navigate('settings');
+                navigate(networkId === null ? '/settings' : 'settings');
```
thmsschmitt: This is needed for the special case that a user lands and `/nms` without any networks and wants to go to the settings.

nms/fbc_js_core/sequelize_models/models/user.js
```diff
@@ -64,11 +63,13 @@ export default (
           return this.getDataValue('networkIDs') || [];
         },
       },
+      // Unused should be removed together with a migration as a breaking change
```
thmsschmitt: Actually it turns out that the migration is not a breaking change #12704

HannaFar (approved):

Issue #12004, opened 2022-03-08
AGW status is fluctuating under Dashboard (Connected/Disconnected) and Equipment (Good/Bad) in NMS UI


p00jaY:
**Your Environment**

     Version : v1.6.1
     Affected Component : NMS
     Affected Subcomponent : N/A
     Deployment Environment : Vagrant [(AGW)]

**Describe the Issue**
Added Gateway in NMS. The status of added gateway is not accurate(fluctuating) sometime it shows Good/Bad in NMS. In Dashboard, same behavior is observed. Gateway's connected status under Events is not accurate (fluctuating) sometime it shows Connected/Disconnected.  If any one explored or having knowledge on NMS code please let me know the exact behavior or exact fix for the same if it seems as a bug. Screenshots are attached below for same:

**To Reproduce**

    Go to NMS UI
    Click on Add Gateway button and add details such as Hardware ID and Challenge Key from AGW Host.
    After Adding Gateway, you can check the gateway status as Good/Bad.
    Then go to Dashboard and monitor the gateway connected status under Events as Connected/Disconnected.

**Expected behavior**

It has to show Good as a gateway status.
![Gateway_health_is_bad](https://user-images.githubusercontent.com/101162794/157183290-060e4583-0eef-4ebb-9fa2-bfb50a0216a9.PNG)
![Gateway_health_is_good](https://user-images.githubusercontent.com/101162794/157183299-5302d5aa-c0dd-43ce-a93f-2ad01b8d2fd4.PNG)
![Gateway_status_as_disconnected](https://user-images.githubusercontent.com/101162794/157183306-10ec3eca-3d0a-4531-9324-2f81fe1ad56c.PNG)
![Gatway_status_as_connected_under_Dashboard](https://user-images.githubusercontent.com/101162794/157183312-d86561fb-4c73-4a3e-8673-fd40b083a185.PNG)

Bharat0908:
@andreilee, requesting you to please suggest your inputs here.

p00jaY:
@andreilee @HannaFar  , is there any update on this issue??

p00jaY:
Hi @andreilee,

Found this similar issue, we request you to please provide your inputs here. 
https://github.com/magma/magma/issues/7645

Best Regards,
Pooja Yadav

p00jaY:
Hi Team,

We noticed the below observation in which we have seen the bad states.
[...]

PR #12699, opened 2022-05-12
fix(orc8r): Make check-in status check in the backend

sebathomas:
## Summary

At the moment, gateway health in NMS is checked by comparing the last check-in time from the backend to the current time in the browser. This check is unstable and results in an unreliable health status in many cases. See #12004 for details.

With this change, the check-in time check is instead performed in the backend, taking the configured check-in interval into account. The result will be used in NMS in #12735.

## Test Plan

Added a unit test and tried it starting NMS locally and connecting/disconnecting an AGW.

Example result for get-gateway call in local test:
```
{
  "gw1": {
    "cellular": {...},
    "checked_in_recently": false,
    "connected_enodeb_serials": [],
    "description": "Test Gateway",
    "device": {...},
    "id": "gw1",
    "magmad": {
      "autoupgrade_enabled": true,
      "autoupgrade_poll_interval": 60,
      "checkin_interval": 60,
      "checkin_timeout": 30,
      "dynamic_services": []
    },
    "name": "TestGateway",
    "status": {
      "checkin_time": 1652362850759,
      "hardware_id": "...",
      "machine_info": {...},
        "network_info": {...}
      },
      "meta": {...},
      "platform_info": {...},
      "system_status": {...}
    },
    "tier": "default"
  }
}

```

## Additional Information

I initially wanted to put the new field inside `status` but the corresponding state microservice doesn't have access to the check-in interval. The calculation needs both the `status` and `magmad` information so I put `checked_in_recently` top-level in the response.


orc8r/cloud/go/services/orchestrator/obsidian/models/conversion.go
```diff
@@ -626,6 +627,21 @@ func (m *GatewayVpnConfigs) FromBackendModels(ctx context.Context, networkID str
...
+	}
+	checkinTime := time.UnixMilli(int64(gatewayStatus.CheckinTime))
+
+	var checkinInterval time.Duration
+	if magmadConfig != nil {
+		checkinInterval = time.Duration(int64(magmadConfig.CheckinInterval)) * time.Second
+	} else {
+		checkinInterval = time.Minute
+	}
+	return time.Now().Before(checkinTime.Add(5 * checkinInterval))
```
sebathomas: The grace period of 5 intervals is somewhat arbitrary. It is based on existing behavior (grace period of 5 minutes for a default interval of 1 minute).

orc8r/cloud/go/services/orchestrator/obsidian/models/conversion.go
```diff
@@ -626,6 +627,21 @@ func (m *GatewayVpnConfigs) FromBackendModels(ctx context.Context, networkID str
...
+	if gatewayStatus == nil {
+		return false
+	}
+	checkinTime := time.UnixMilli(int64(gatewayStatus.CheckinTime))
+
+	var checkinInterval time.Duration
+	if magmadConfig != nil {
+		checkinInterval = time.Duration(int64(magmadConfig.CheckinInterval)) * time.Second
+	} else {
+		checkinInterval = time.Minute
```
sebathomas: Using the default interval if no config can be found.

thmsschmitt (approved):
NMS changes look good to me

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
```diff
@@ -1411,6 +1411,7 @@ func TestListAndGetGateways(t *testing.T) {
 	ctx := test_utils.GetContextWithCertificate(t, "hw1")
 	test_utils.ReportGatewayStatus(t, ctx, models.NewDefaultGatewayStatus("hw1"))
 
+	checkedInRecentlyFalse := lteModels.GatewayCheckedInRecently(false)
```
uri200: I think you are using the same `checkedInRecentlyFalse`  on multiple object.  This can create issues if you modify the content of if .  Also, since this is a test, it is always better to make them independent. Code redundancy in tests is not a problem

better if you use it like this: 
```
CheckedInRecently: lteModels.GatewayCheckedInRecently(false)
```
Besides, it looks better with the rest ;)

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
```diff
@@ -4425,3 +4443,28 @@ func newMutableGateway(id string) *lteModels.MutableLteGateway {
...
+			CheckinInterval:         15,
+			CheckinTimeout:          10,
+			AutoupgradePollInterval: 300,
+			AutoupgradeEnabled:      swag.Bool(true),
+		},
+		Cellular:               newDefaultGatewayConfig(),
+		ConnectedEnodebSerials: []string{},
+		Tier:                   "t0",
+		ApnResources:           lteModels.ApnResources{},
+		CheckedInRecently:      &checkedInRecently,
```
uri200: why not just 

```
CheckedInRecently: &lteModels.GatewayCheckedInRecently(false)
```

Since this is a test it is always good to see the object at one sight.

lte/cloud/go/services/lte/obsidian/models/conversion.go
```diff
@@ -200,6 +200,9 @@ func (m *LteGateway) FromBackendModels(
 	}
 	sort.Strings(m.ConnectedEnodebSerials)
 
+	checkedInRecently := GatewayCheckedInRecently(orc8rModels.LastGatewayCheckInWasRecent(m.Status, m.Magmad))
+	m.CheckedInRecently = &checkedInRecently
```
uri200, line 203: same here , you can save one variable 

```
m.CheckedInRecently = &GatewayCheckedInRecently(orc8rModels.LastGatewayCheckInWasRecent(m.Status, m.Magmad))

orc8r/cloud/go/services/orchestrator/obsidian/models/conversion.go
uri200: Can you create a constant, like `DEFAULTCHECKININTERVAL` and set it on top of this file?

orc8r/cloud/go/services/orchestrator/obsidian/models/conversion.go
uri200: Same here, lets create a constant and set it up on top. 

Add a comment on each constant to have some kind of idea what they mean

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
sebathomas: The problem is that `CheckedInRecently` is a pointer and I don't think Go allows you to create a variable and return the address in a one-liner. But you're right that using the same object for all these instances is problematic. I'll refactor it.

lte/cloud/go/services/lte/obsidian/models/conversion.go
sebathomas: Same problem here, I get a compile error if I don't create a variable before setting the pointer. I don't think Go has a more convenient way to do that.

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
sebathomas: Extracted a function to get the address in one line.

lte/cloud/go/services/lte/obsidian/handlers/handlers_test.go
sebathomas: Changed.

orc8r/cloud/go/services/orchestrator/obsidian/models/conversion.go
sebathomas: Done.

orc8r/cloud/go/services/orchestrator/obsidian/models/conversion.go
sebathomas: Done.

Neudrino (commented):
Could you also execute `nms/$ yarn test`, please?
I believe adding some default `checked_in_resently: false,` might break some test relying on this status to be `true`.

sebathomas:
> Could you also execute `nms/$ yarn test`, please? I believe adding some default `checked_in_resently: false,` might break some test relying on this status to be `true`.

These tests should be run by the CI:
https://github.com/magma/magma/runs/6418876080?check_suite_focus=true

Neudrino (approved):

alexzurbonsen (approved):

maxhbr (approved):
LGTM

PR #12700, opened 2022-05-12
ci(dp): Add Path filter for DP integration_tests

sklgromek:
- Add Path filter for integration tests job to avoid triggering them with no significant changes in the repo
- Fix reporting from integration tests

### DIFF
--- a/.github/workflows/dp-workflow.yml
+++ b/.github/workflows/dp-workflow.yml
@@ -1,4 +1,5 @@
-name: Domain proxy workflow
+---
+name: dp-workflow
 
 on:  # yamllint disable-line rule:truthy
   push:
@@ -11,6 +12,10 @@ on:  # yamllint disable-line rule:truthy
       - 'v1.*'
     types: [opened, reopened, synchronize]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
 jobs:
   path_filter:
     runs-on: ubuntu-latest
@@ -20,7 +25,7 @@ jobs:
       rc: ${{ steps.filter.outputs.rc }}
       db: ${{ steps.filter.outputs.db }}
       helm: ${{ steps.filter.outputs.helm }}
-      integration_tests_no_orc8r: ${{ steps.filter.outputs.integration_tests_no_orc8r }}
+      integration_tests: ${{ steps.filter.outputs.integration_tests }}
     steps:
       - uses: actions/checkout@v2
         if: github.event_name == 'push'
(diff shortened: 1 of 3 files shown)


.github/workflows/unit-test-workflow.yml
```diff
@@ -8,6 +8,7 @@ on:  # yamllint disable-line rule:truthy
       - agw-workflow
       - golang-build-test
       - nms-workflow
+      - Domain proxy workflow
```
nstng: The other workflows are linked via file name (without extension). Does this work also with the name of the workflow? Even if, it is more consistent if the file name is used and imho more robust (renaming the workflow name does not cause failures here).

.github/workflows/dp-workflow.yml
```diff
@@ -385,56 +407,33 @@ jobs:
...
-          path: test-results
-
-      - name: Publish integration test results
-        uses: EnricoMi/publish-unit-test-result-action@v1
-        with:
-          commit: ${{ github.event.workflow_run.head_sha }}
-          files: test-results/**/*.xml
-          comment_title: "Domain Proxy integration tests results"
-
+          path: ${{ env.TEST_DIR }}
```
nstng, line 430: Thanks for cleaning this up :)

.github/workflows/dp-workflow.yml
```diff
@@ -55,12 +59,28 @@ jobs:
...
               - 'dp/**'
+              - 'nms/**'
+              - 'src/**'
+              - 'scripts/**'
+              - 'protos/**'
             db:
               - '.github/workflows/dp-workflow.yml'
               - 'dp/cloud/python/magma/db_service/**'
-
+      - name: Save should_not_skip output
```
nstng, line 73: Why is this named "Save should_not_skip"? Why is the information from the uploaded file needed?

.github/workflows/unit-test-workflow.yml
sklgromek: I think it will work only with name of workflow, and other workflows have exactly that names, However, we can rename our workflow to look as others

nstng: See e.g., `golang-build-test.yml` has the name `Golang Build & Test`.

.github/workflows/dp-workflow.yml
sklgromek: I am not sure of that, other workflows which are using "Unit test reporting" workflow have the same step. I just copied it. It's required to artifact pr.zip exists in Workflow, because on that condition other steps are run from it.

.github/workflows/unit-test-workflow.yml
sklgromek: I am not sure of that but I think it will not work for that one. All others have a short name, and I saw that reports are being displayed in the comments section of the PR's, for that one I was not able to find any.  However, I changed the name of our workflow to unify.

nstng: Thank you for clarifying. The naming still confuses me - but I think this is easier to change in another PR if this is consistent for now.

.github/workflows/dp-workflow.yml
nstng: Thank you for clarifying. The naming still confuses me - but I think this is easier to change in another PR if this is consistent for now.

jkmar (approved):

nstng (approved):
CI changes LGTM - checked that expected artifacts are uploaded.

Out of scope here, but out of curiosity: why does dp-workflow use strategy matrix in all jobs, imho there is only one job where actually more then one case is considered.

sklgromek:
> CI changes LGTM - checked that expected artifacts are uploaded.
> 
> Out of scope here, but out of curiosity: why does dp-workflow use strategy matrix in all jobs, imho there is only one job where actually more then one case is considered.

Thanks @nstng, you are right, matrix-strategy is not necessary for most of them. I will cleanup it in another PR

PR #12703, opened 2022-05-12
feat(dp): Support single step registration without CPI in AMC

jkmar:
## Summary

Support single step registration without CPI in AMC


dp/cloud/go/active_mode_controller/internal/message_generator/sas/registration.go
```diff
@@ -9,16 +13,57 @@ func NewRegistrationRequestGenerator() *registrationRequestGenerator {
 }
 
 func (*registrationRequestGenerator) GenerateRequests(cbsd *active_mode.Cbsd) []*Request {
-	req := &registrationRequest{
-		UserId:           cbsd.GetUserId(),
-		FccId:            cbsd.GetFccId(),
-		CbsdSerialNumber: cbsd.GetSerialNumber(),
-	}
+	req := buildRegistrationRequest(cbsd)
```
WojSad, line 16: i think we shouldn't be generating a registration request for a radio unless it's either marked as "multi-step" or it's single-step && type A && has installation params filled out && is indoors (for now). These fields, in a single-step should be updated by enodebd first before we send a request.

dp/cloud/python/magma/radio_controller/services/active_mode_controller/service.py
```diff
@@ -152,11 +148,43 @@ def _list_cbsds(session: Session) -> State:
...
+            and_(
+                DBCbsd.single_step_enabled == True,
+                not_null(multi_step + single_step),
+            ),
+        ),
+    )
+
+
+def not_null(fields: List[Any]):
+    return and_(*[field != None for field in fields])  # noqa: E711
```
WojSad, line 187: some fields that are required for type B radios are optional for type A

dp/cloud/python/magma/radio_controller/services/active_mode_controller/service.py
```diff
@@ -152,11 +148,43 @@ def _list_cbsds(session: Session) -> State:
...
+                DBCbsd.should_deregister == True,
+                DBCbsd.is_deleted == True,
+            ),
+            and_(
+                DBCbsd.single_step_enabled == False,
+                not_null(multi_step),
+            ),
+            and_(
+                DBCbsd.single_step_enabled == True,
+                not_null(multi_step + single_step),
```
WojSad, line 180: I think we should add DBCbsd.cbsd_category == "a"  && DBCbsd.indoor_deployment == True to this condition for now.

dp/cloud/python/magma/radio_controller/services/active_mode_controller/service.py
```diff
@@ -152,11 +148,43 @@ def _list_cbsds(session: Session) -> State:
...
         populate_existing()
     )
 
 
+def _build_filter():
+    multi_step = [
+        DBCbsd.fcc_id, DBCbsd.user_id, DBCbsd.number_of_ports,
+        DBCbsd.min_power, DBCbsd.max_power, DBCbsd.antenna_gain,
+    ]
+    single_step = [
```
WojSad, line 161: maybe split it between type A and B single step, since some fields required for B are optional for A. See https://docs.google.com/document/d/1NMX3ZIzV-7nagjrmL4D03mCIE98ENlumYyj8bcW9oOM/edit#heading=h.sdkpl59t443e for reference

dp/cloud/python/magma/radio_controller/services/active_mode_controller/service.py
jkmar: as discussed, let's leave it like this for now

dp/cloud/python/magma/radio_controller/services/active_mode_controller/service.py
jkmar: done

dp/cloud/python/magma/radio_controller/services/active_mode_controller/service.py
jkmar: as discussed, let's leave it like this for now

WojSad (approved):

PR #12709, opened 2022-05-13
chore(sessiond): GRPC printing

vikramgreddy:
Signed-off-by: Vikram G <vikram.gurrappagaru@radisys.com>

## Summary
PrintGrpcMessage to DirectorydClient.cpp on request

## Test Plan
make integ_test
make test session_manager

### DIFF
--- a/lte/gateway/c/session_manager/DirectorydClient.cpp
+++ b/lte/gateway/c/session_manager/DirectorydClient.cpp
@@ -12,6 +12,7 @@
  */
 
 #include "lte/gateway/c/session_manager/DirectorydClient.hpp"
+#include "lte/gateway/c/session_manager/GrpcMagmaUtils.hpp"
 
 #include <grpcpp/channel.h>
 #include <orc8r/protos/common.pb.h>
@@ -44,6 +45,7 @@ void AsyncDirectorydClient::update_directoryd_record(
     std::function<void(Status status, Void)> callback) {
   auto local_response =
       new AsyncLocalResponse<Void>(std::move(callback), RESPONSE_TIMEOUT);
+  PrintGrpcMessage(static_cast<const google::protobuf::Message&>(request));
   local_response->set_response_reader(stub_->AsyncUpdateRecord(
       local_response->get_context(), request, &queue_));
 }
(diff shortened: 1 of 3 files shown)


uri200 (commented):
Thanks @vikramgreddy. 

This will print the `request`. We will also need to print the response. 
Unfortunately the way CPP manages grpc message does not allow us to use the same strategy here, so using your IDE, please search all the instances where those functions are called and print the response. For example here 
http://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/SessionProxyResponderHandler.cpp#L78-L78

uri200:
Also you can try to see the logs running [ S1AP test](https://docs.magmacore.org/docs/next/lte/s1ap_tests#s1ap-integration-tests) and enable grpc printing on sessiond.yml. To do that you could go your vagrant vm and change  `print_grpc_payload: True` at `/etc/magma/sessiond.yml`. Then from your host  (not vagrant) you can run 
```
cd magma/lte/gateway
 fab integ_test:provision_vm=False,rerun_fails=0,no_build=True,skip_unit_tests=True
```
To run the test.

From your Vagrant VM  you will be able to see sessiond logs using
```
sudo journalctl -fu magma@sessiond
```

In case you need it, you can add some extra prints to the function and rebuild sessiond to see if they print the message. 

Note that using s1ap tester will be very useful on the future to test out stuff that can't be tested with unit test

lte/gateway/c/session_manager/DirectorydClient.cpp
```diff
@@ -53,6 +55,7 @@ void AsyncDirectorydClient::delete_directoryd_record(
     std::function<void(Status status, Void)> callback) {
   auto local_response =
       new AsyncLocalResponse<Void>(std::move(callback), RESPONSE_TIMEOUT);
+  PrintGrpcMessage(static_cast<const google::protobuf::Message&>(request));
   local_response->set_response_reader(stub_->AsyncDeleteRecord(
       local_response->get_context(), request, &queue_));
 }
```
uri200, line 61: is there a reason why `get_all_directoryd_records` doesnt have a print?
Matybe too overkill to print all records? Or is there any other reason

If so maybe it would be good to print just that a message has been received.

vikramgreddy: get_all_directoryd_records is only called in RestartHandler::populate_sessions_to_terminate_with_retries. Inside this function we have print. Do you want to add in get_all_directoryd_records as well .?

uri200 (approved):

PR #12713, opened 2022-05-13
feat(ci): add eslint to reviewdog

Neudrino:
## Summary

`eslint` is used in the JavaScript universe and specifically for the NMS frontend in magma.
This integrates the linter with reviewdog for a visual feedback in Github.

The change to the yarn package install is necessary, because otherwise the npm install step fails in CI. `yarn.lock` automatically updated by running `yarn`.

## Test Plan

- CI
  - no changes: https://github.com/magma/magma/actions/runs/2319132027
  - good changes: https://github.com/magma/magma/actions/runs/2319593415
  - bad changes: https://github.com/magma/magma/actions/runs/2319630502

### DIFF
--- a/.github/workflows/reviewdog-workflow.yml
+++ b/.github/workflows/reviewdog-workflow.yml
@@ -21,6 +21,7 @@ jobs:
     outputs:
       changed_cpp: ${{ steps.changes.outputs.cpp }}
       changed_go: ${{ steps.changes.outputs.go }}
+      changed_javascript: ${{ steps.changes.outputs.javascript }}
       changed_python: ${{ steps.changes.outputs.python }}
       changed_terraform: ${{ steps.changes.outputs.terraform }}
     steps:
@@ -35,6 +36,8 @@ jobs:
               - ["lte/gateway/c/**", "orc8r/gateway/c/**"]
             go:
               - ["src/go/**"]
+            javascript:
+              - ["nms/**", "**/*.js"]
             python:
               - ["lte/gateway/python/**", "orc8r/gateway/python/**"]
             terraform:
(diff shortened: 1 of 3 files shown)


nms/package.json
```diff
@@ -117,7 +117,7 @@
     "prettier": "^2.0.5",
     "puppeteer": "5.5.0",
     "react-hot-loader": "^4.8.0",
-    "react-test-renderer": "^17.0.2",
+    "react-test-renderer": "^16.9.0",
```
thmsschmitt, line 120: Makes sense, we should use a test renderer that fits to our react version.

.github/workflows/reviewdog-workflow.yml
```diff
@@ -113,6 +116,26 @@ jobs:
           # Ignore DL3005-"Do not use apt-get upgrade or dist-upgrade"
           hadolint_ignore: DL3005
 
+  eslint:
```
thmsschmitt, line 119: Does that mean we are now running `eslint` twice once as part of the nms workflow and ones in the reviewdog workflow?

Neudrino: That is correct. However, I do not see an issue with that, as it will be run in parallel and thus not prolong the CI execution.
I would see this as first step in part of a larger consolidation of the CI, in this part of linters and formatters.
Anyhow, for now, it brings the advantage that people opening a PR see in the PR conversation, what issues the Linter shows, instead of only having those hidden in the run logs of the action.

jheidbrink (approved):

thmsschmitt (approved):

PR #12718, opened 2022-05-13
refactor(nms): Reset and restore mocks automatically after every test

thmsschmitt:
## Summary

Resets and restores jest mocks after every test per configuration option.
This makes sure that no test silently depends on mocking or calling of mocks in another test.

## Test Plan

```
yarn test
...
Test Suites: 52 passed, 52 total
Tests:       315 passed, 315 total
Snapshots:   0 total
Time:        18.444 s
Ran all test suites in 2 projects.
✨  Done in 20.20s.
```

### DIFF
--- a/nms/jest.config.js
+++ b/nms/jest.config.js
@@ -32,6 +32,8 @@ module.exports = {
       transform: {
         '^.+\\.js$': 'babel-jest',
       },
+      resetMocks: true,
+      restoreMocks: true,
     },
     {
       name: 'app',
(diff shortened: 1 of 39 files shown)


sebathomas (approved):
Github really doesn't make it too easy to see the changes here, but looking good to me.

nms/jest.config.js
```diff
@@ -32,6 +32,8 @@ module.exports = {
       transform: {
         '^.+\\.js$': 'babel-jest',
       },
+      resetMocks: true,
+      restoreMocks: true,
```
Neudrino, line 36: Imho `resetMocks: true,` is redundant as it is included in `restoreMocks: true,` if I read [the documentation](https://jestjs.io/docs/mock-function-api#mockfnmockreset) correctly.

thmsschmitt: I also understood the documentations this way, but unfortunately you really have to do both.

Neudrino (approved):

Neudrino (approved):

PR #12724, opened 2022-05-16
refactor(orc8r): replace deprecated errors module

wolfseb:
Replace functions from the github.com/pgk/errors package with
`fmt.Errorf`. Created in pairing with Moritz Huebner.

Signed-off-by: Sebastian Wolf <sebastian.wolf@tngtech.com>

## Summary

This PR implements the proposed solutions of ticket #12632 for the orc8r module.

A few instances of broken rollback error handling next to the changes
in this PR are also fixed.

## Test Plan

Perform Unit tests

## Additional Information

Worked in pairing with @MoritzThomasHuebner 

- [ ] This change is backwards-breaking

PR #12725, opened 2022-05-16
refactor(cwg): replace deprecated errors module

### LINKED ISSUE #12632: Check whether `github.com/pkg/errors` can be replaced with standard library functions
The module [`github.com/pkg/errors`](https://github.com/pkg/errors) is no longer maintained, see for example https://github.com/pkg/errors/issues/245.

It seems like it can be removed and replaced by standard lib functionality (mainly `errors.Errorf` --> `fmt.Errorf`), which includes all of the functionality, except for the stack traces, that are included in the error type from `pkg/errors`. See for example this comment https://github.com/microsoft/go-winio/pull/227#issuecomment-984390345.

We should probably check whether removal is sensible and if yes replace it.

## Progress:

this needs to be fixed in
- [x] ./orc8r : https://github.com/magma/magma/pull/12724
- [x] ./feg : https://github.com/magma/magma/pull/12726
- [x] ./cwf : https://github.com/magma/magma/pull/12725
- [x] ./src : https://github.com/magma/magma/pull/12729
- [x] ./dp : https://github.com/magma/magma/pull/12727
- [ ] ...

MoritzThomasHuebner:
Fixes #12632 for description

## Summary

Remove dependency on "github.com/pkg/errors"

- Replace use of `errors.Wrap[f]` with `fmt.Errorf`
- Replace use of pkg `errors.New` with standard lib `errors.New`
- Update go.mod files
- Replace use of `errors.New(fmt.Sprintf(...))` with `fmt.Errorf`
- The error handling in `addToStore` in `ue_store_utils.go` and in `getUE` in `uesim.go` was incorrect.
  - In these database transactions `commitErr` and `rollbackErr` were not logged.
  - We restructured the switch/case logic to if/else
  - In the else bracket  (`err != nil`), we now use a `multierror` to propagate both `err` and `rollbackErr`. The logic here is that the rollback should only be triggered if an error happened before, so both errors should be logged.
## Test Plan

Run unit tests

## Additional Information

Worked in pairing with @wolfseb 

### DIFF
--- a/cwf/cloud/go/services/cwf/obsidian/models/validate.go
+++ b/cwf/cloud/go/services/cwf/obsidian/models/validate.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
-	"github.com/pkg/errors"
 )
 
 func (m *CwfNetwork) ValidateModel(context.Context) error {
@@ -45,7 +44,7 @@ func (m *GatewayCwfConfigs) ValidateModel(context.Context) error {
 	for _, peer := range m.AllowedGrePeers {
 		for _, key := range set[peer.IP] {
 			if swag.Uint32Value(peer.Key) == key {
-				return errors.New(fmt.Sprintf("Found duplicate peer %s with key %d", peer.IP, key))
+				return fmt.Errorf("Found duplicate peer %s with key %d", peer.IP, key)
 			}
 		}
 		set[peer.IP] = append(set[peer.IP], swag.Uint32Value(peer.Key))
--- a/cwf/gateway/services/uesim/servicers/ue_store_utils.go
+++ b/cwf/gateway/services/uesim/servicers/ue_store_utils.go
@@ -14,18 +14,18 @@
 package servicers
 
 import (
+	"fmt"
+
 	cwfprotos "magma/cwf/cloud/go/protos"
 	"magma/orc8r/cloud/go/blobstore"
 	"magma/orc8r/lib/go/protos"
-
-	"github.com/pkg/errors"
 )
 
 func addUeToStore(srvstore blobstore.StoreFactory, ue *cwfprotos.UEConfig) {
 	blob, err := ueToBlob(ue)
 	store, err := srvstore.StartTransaction(nil)
 	if err != nil {
-		err = errors.Wrap(err, "Error while starting transaction")
+		err = fmt.Errorf("Error while starting transaction: %w", err)
 		err = ConvertStorageErrorToGrpcStatus(err)
 		return
 	}
@@ -33,12 +33,12 @@ func addUeToStore(srvstore blobstore.StoreFactory, ue *cwfprotos.UEConfig) {
 		switch err {
 		case nil:
 			if commitErr := store.Commit(); commitErr != nil {
-				err = errors.Wrap(err, "Error while committing transaction")
+				err = fmt.Errorf("Error while committing transaction: %w", err)
 				err = ConvertStorageErrorToGrpcStatus(err)
 			}
 		default:
 			if rollbackErr := store.Rollback(); rollbackErr != nil {
-				err = errors.Wrap(err, "Error while rolling back transaction")
+				err = fmt.Errorf("Error while rolling back transaction: %w", err)
 				err = ConvertStorageErrorToGrpcStatus(err)
 			}
 		}
--- a/cwf/gateway/services/uesim/servicers/ue_store_utils.go
+++ b/cwf/gateway/services/uesim/servicers/ue_store_utils.go
@@ -19,6 +19,8 @@ import (
 	cwfprotos "magma/cwf/cloud/go/protos"
 	"magma/orc8r/cloud/go/blobstore"
 	"magma/orc8r/lib/go/protos"
+
+	"github.com/hashicorp/go-multierror"
 )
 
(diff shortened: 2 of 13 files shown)


cwf/cloud/go/services/cwf/obsidian/models/validate.go
```diff
@@ -45,7 +44,7 @@ func (m *GatewayCwfConfigs) ValidateModel(context.Context) error {
 	for _, peer := range m.AllowedGrePeers {
 		for _, key := range set[peer.IP] {
 			if swag.Uint32Value(peer.Key) == key {
-				return errors.New(fmt.Sprintf("Found duplicate peer %s with key %d", peer.IP, key))
+				return fmt.Errorf("Found duplicate peer %s with key %d", peer.IP, key)
```
alexzurbonsen, line 47: Noticed that the PR description says this would be substituted with errors.New from std lib. I think this way might be nicer (you don't need the Sprintf). But may be adapt the PR summary?

cwf/gateway/services/uesim/servicers/ue_store_utils.go
```diff
@@ -14,32 +14,33 @@
...
+				err = fmt.Errorf("Error while committing transaction: %w", commitErr)
 				err = ConvertStorageErrorToGrpcStatus(err)
 			}
-		default:
+		} else {
 			if rollbackErr := store.Rollback(); rollbackErr != nil {
-				err = errors.Wrap(err, "Error while rolling back transaction")
-				err = ConvertStorageErrorToGrpcStatus(err)
+				errs := multierror.Append(err, fmt.Errorf("Error while rolling back transaction: %w", rollbackErr))
+				err = ConvertStorageErrorToGrpcStatus(errs.ErrorOrNil())
```
alexzurbonsen, line 43: Is there a reason to use multierror in one case but not in the other? And for using multierror at all?

And why do we start passing commitErr and rollbackErr? 

There is a similar case in cwf/gateway/services/uesim/servicers/uesim.go.

cwf/cloud/go/services/cwf/obsidian/models/validate.go
wolfseb: I agree. Also, I stumbled across a few cases like this in #12724, where there was a `Sprintf` inside an `errors.New`. When we just change to `errors.New` from the std lib, the linter will complain and suggest using `fmt.Errorf` instead, as it is done here.
But yes, we should add this case to the summary.

cwf/gateway/services/uesim/servicers/ue_store_utils.go
wolfseb: If `err` is nil, but the commit fails, there is only one error to be passed on. If `err` is not nil, then it tries to roll back, but if that also fails, we have two errors to pass: `err` and `rollbackErr`, that's why we use a multierror here.

We need to pass `commitErr` , because `err` is nil by default here due to the `if` clause, otherwise we would find an error, but the error we pass is nil, so in the end the error gets ignored.
In the other case we use the multierror to pass on both, `err` and `rollbackErr`.

cwf/cloud/go/services/cwf/obsidian/models/validate.go
MoritzThomasHuebner: I updated the summary.

cwf/gateway/services/uesim/servicers/ue_store_utils.go
MoritzThomasHuebner: I agree with @wolfseb

alexzurbonsen: Ah, so this goes beyond replacing pkg/errors and fixes broken error handling. Probably should mention that in the summary as well.

MoritzThomasHuebner: I added this to the summary.

cwf/gateway/services/uesim/servicers/uesim.go
```diff
@@ -397,9 +398,9 @@ func executeCommand(command string, argList []string) (*IperfResponse, error) {
 	rawOutput, err := cmd.Output()
 	output, _ := (&IperfResponse{}).FromBytes(rawOutput)
 	if err != nil {
-		newError := errors.Wrap(err, fmt.Sprintf(
+		newError := fmt.Errorf(fmt.Sprintf(
 			"error while executing \"%s %s\"\n output:\n%v",
-			command, strings.Join(argList, " "), string(rawOutput)))
+			command, strings.Join(argList, " "), string(rawOutput))+": %w", err)
```
Neudrino, line 403: From the rest of the code I would assume that binary operators (`+`)  should be surrounded by white-spaces.
Was the go formatter run?

voisey: The Go formatter seems to be happy with the current formatting. In particular, if I add spaces and then run `cd magma/cwf/gateway && go fmt ./...`, the spaces are removed. Note that this is also covered in the CI: https://github.com/magma/magma/pull/12749.

Neudrino (approved):

alexzurbonsen (approved):

alexzurbonsen:
@uri200 Want to bring this PR to your attention. Would be great to get a review. Thanks!

uri200 (approved):

PR #12735, opened 2022-05-16
feat(nms): use proper check in status from orcestrator for lteGateways

Neudrino:
:warning: Commit wise review recommended. ~~First Commit is part of #12699 and will vanish after its merge and a rebase.~~

## Summary

Within https://github.com/magma/magma/pull/12699 a orc8r calculated status was implemented.
This PR provides the implementation to actually display this status and not calculate it (wrongly) in the front-end.

As the health is dependent on the configured check-in interval, one cannot give an exact time period in the tool-tip, but has to stay more vague in the tool-tip description.

## Test Plan

- `yarn flow`
- `yarn run eslint . --quiet --fix`
- `yarn test --maxWorkers=2`

## Additional Information

- [ ] Requires https://github.com/magma/magma/pull/12699.

### DIFF
--- a/nms/app/views/equipment/GatewayDetailStatus.js
+++ b/nms/app/views/equipment/GatewayDetailStatus.js
@@ -89,7 +89,7 @@ export default function GatewayDetailStatus({refresh}: {refresh: boolean}) {
         status: gwInfo.checked_in_recently,
         tooltip: gwInfo.checked_in_recently
           ? 'Gateway checked in recently'
-          : "Gateway hasn't checked in within last 5 minutes",
+          : 'Gateway has not checked in recently',
       },
       {
         category: 'Last Check in',
--- a/nms/app/views/equipment/GatewayDetailStatus.js
+++ b/nms/app/views/equipment/GatewayDetailStatus.js
@@ -13,6 +13,7 @@
  * @flow strict-local
  * @format
  */
+
 import type {DataRows} from '../../components/DataGrid';
 
 import DataGrid from '../../components/DataGrid';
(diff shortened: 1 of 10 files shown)


nms/app/views/network/FEGServicingAccessGatewayTable.js
```diff
@@ -79,9 +78,8 @@ async function getServicedAccessGatewaysInfo(
         gatewayId: servicedAccessGatewayId,
         gatewayName:
           servicedAccessGateways[servicedAccessGatewayId]?.name || '',
-        gatewayHealth: isGatewayHealthy(
-          servicedAccessGateways[servicedAccessGatewayId] || {},
-        )
+        gatewayHealth: (servicedAccessGateways[servicedAccessGatewayId] || {})
```
thmsschmitt: ``` servicedAccessGateways[servicedAccessGatewayId]?.checked_in_recently ```

nms/app/views/equipment/GatewayDetailStatus.js
```diff
@@ -79,17 +80,17 @@ export default function GatewayDetailStatus({refresh}: {refresh: boolean}) {
     !!gwInfo.magmad.dynamic_services &&
     gwInfo.magmad.dynamic_services.includes(DynamicServices.MONITORD);
 
-  const isHealthy = isGatewayHealthy(gwInfo);
+  const isHealthy = gwInfo.checked_in_recently;
```
sebathomas: Could be inlined (or used for all fields).

sebathomas (changes requested):
It looks to me like this also adapts the logic for FEGs while the backend logic was only changed for AGWs. In that case, we either we need to keep the old logic for FEGs or implement the backend changes for them as well.

sebathomas (approved):
As discussed in person, I was confused by the file names and the logic is only used for LTE gateways, same as in the backend.

thmsschmitt (approved):

sebathomas:
I did an E2E test with check-in interval 30 and confirmed that the status switch works correctly in NMS. :+1:

sebathomas:
This PR can finally be rebased and opened properly. :)

Neudrino:
Rebased after merge of https://github.com/magma/magma/pull/12699.

thmsschmitt (approved):

PR #12743, opened 2022-05-17
chore(ci): Modify the bazel workflow to use remote caching

### LINKED ISSUE #12458: Investigate options for Bazel remote caches
This ticket tracks efforts in the investigation of remote caches for Bazel.

### Motivation

We are currently [using](https://github.com/magma/magma/blob/0de62a739c867ac35979452e50ad8acf637b17bd/.github/workflows/bazel.yml#L76) Github Actions cache for Bazel builds in the CI.

These come with some challenges:
* The total cache size is limited to 10 GiB. As long as we are not fully hermetic, we should use different cache keys for different build environments (ie VM / devcontainer). That might make us exceed the limit which could make Github evict caches that are still useful as described [here](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy).
* The Github Actions runners are limited in disk space. In order to prevent them from running full, we [remove some unsed SDKs](https://github.com/magma/magma/blob/0de62a739c867ac35979452e50ad8acf637b17bd/.github/workflows/bazel.yml#L67) and clear the cache [once it exceeds 6.5 GiB](https://github.com/magma/magma/blob/0de62a739c867ac35979452e50ad8acf637b17bd/.github/workflows/bazel.yml#L97)

By using remote caches, we are not limited to 10 GiB. Also, downloading cache results is much more fine-grained which can reduce the amount of data downloaded and hopefully abolish the need to maintain code that clears the cache in our runners.

### LINKED ISSUE #12626: Bazel remote caching setup/disaster recovery documentation is drafted
## Bazel remote caching setup und disaster recovery plan 

:warning: This is a draft documentation and should be moved into the official magma bazel docs when done.
See also #12458

### Disaster recovery in the case of poisoned caches
 - **Option 1**: Delete the entire cache (requires magma AWS access)
   - Step 1: Update the service to have desired task count zero. Wait for running tasks to stop.
   - Step 2: Manually empty the s3 bucket prefixed with "bazel-remote-cache-".
   - Step 3: Update the service to have desired task count one.
 - **Option 2**: Delete and redeploy the entire remote caching setup  (requires magma AWS and ci-infra repo access)
   -  Tear down the bazel-remote infrastructure by running `terraform destroy -target=aws_s3_bucket.S3CacheBucket` (the `force_destroy = true` options needs to have been enabled in the terraform code). Then redeploy with `terraform init` and `terraform apply`.
- **Option 3**: Invalidate the cache keys (requires CI codeowner approval)
  - With bazel-remote this can be done by changing the `--remote_cache` URL e.g. from `https://user:pw@url:9090/current-cache` to `https://user:pw@url:9090/new-cache`.
  - This needs to be changed in all affected workflows and might require rebasing.

### Disaster recovery in the case of broken remote cache service
 - **Option 1**: Delete and redeploy the entire remote caching setup  (requires magma AWS and ci-infra repo access)
   -  Tear down the bazel-remote infrastructure by running `terraform destroy -target=aws_s3_bucket.S3CacheBucket` (the `force_destroy = true` options needs to have been enabled in the terraform code). Then redeploy with `terraform init` and `terraform apply`.
 - **Option 2**: Re-implement the GH caches (requires CI codeowner approval).

LKreutzer:

## Summary

- The bazel workflow is modified to use the [remote cache](https://github.com/magma/magma/issues/12458) instead of the GH caches.
  - GH caching steps are removed.
  - If the secret is present read + write access is used, else read-only access is used.
  - The files `bazel/bazelrcs/remote_caching_rw.bazelrc` and `bazel/bazelrcs/remote_caching_ro.bazelrc` are templates which are filled in the workflow and then imported in the `.bazelrc`.
- A `workflow_dispatch` trigger is added to the workflow (for ease of testing).

## Test Plan

- CI
- See [run on this PR with read-only access](https://github.com/magma/magma/runs/6466203486?check_suite_focus=true).
- See [run on fork with secret present](https://github.com/LKreutzer/magma/actions/runs/2333026996).

One can check that the import of the templated `bazelrc` file works by finding a line like 
```
INFO: Found applicable config definition build:remote_caching_rw in file /workspaces/magma/remote-cache.bazelrc: --remote_cache=***bazel-remote-cache.magmacore-ci.org:9090/bazel-base-image
```
 at the beginning of the bazel output.

If there were remote cache hits one will find a statement similar to:
```
INFO: Elapsed time: 407.457s, Critical Path: 41.39s
INFO: 7188 processes: 5309 remote cache hit, 1614 internal, 2 local, 263 processwrapper-sandbox.
INFO: Build completed successfully, 7188 total actions
```
at the end of the bazel output, which lists `remote cache hit`s. 

## Additional Information

- See [Wiki page on remote cache disaster recovery](https://github.com/magma/magma/wiki/Bazel-remote-caching-disaster-recovery-plan) in case of poisoned caches or other issues with the remote cache.

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -1,6 +1,7 @@
 ---
 name: "Bazel Build & Test"
 on:  # yamllint disable-line rule:truthy
+  workflow_dispatch:
   pull_request:
     types:
       - opened
@@ -16,10 +17,8 @@ on:  # yamllint disable-line rule:truthy
     - cron: '0 0,6,12,18 * * *'
 env:
   BAZEL_BASE_IMAGE: "ghcr.io/magma/magma/bazel-base:latestv2"
-  BAZEL_CACHE: bazel-cache
-  BAZEL_CACHE_REPO: bazel-cache-repo
-  CACHE_SUB_KEY_BUILD_ALL: build-all
-  CACHE_SUB_KEY_TEST_ALL: test-all
+  CACHE_KEY: bazel-base-image
+
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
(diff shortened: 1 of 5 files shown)


.github/workflows/bazel.yml
```diff
@@ -73,48 +72,6 @@ jobs:
...
         uses: actions/checkout@v2
-      - name: Bazel Cache
-        uses: actions/cache@v2
-        with:
-          path: ${{ github.workspace }}/.${{ env.BAZEL_CACHE }}
-          key: ${{ runner.os }}-${{ env.BAZEL_CACHE }}-${{ env.CACHE_SUB_KEY_BUILD_ALL }}-${{ github.sha }}
-          restore-keys: |
-            ${{ runner.os }}-${{ env.BAZEL_CACHE }}-${{ env.CACHE_SUB_KEY_BUILD_ALL }}-
-
-      - name: Bazel Cache Repo
```
jheidbrink, line 84: Didn't we want to keep the repo cache? Anyway, I'm fine with removing it if that's not too bad for performance.

bazel/bazelrcs/remote_caching_ro.bazelrc
```diff
@@ -0,0 +1,2 @@
+build --config=remote_caching_ro
```
jheidbrink, line 3: Even though we already explain this in `.bazelrc`, I think it wouldn't hurt to also explain here that this is supposed to be templated and written to `remote-cache.bazelrc` where it is imported by `.bazelrc`

.github/workflows/bazel.yml
LKreutzer: At least with the bazel version that we are using (4.2.2) it does not seem possible to combine the repository cache with a remote cache. See open feature request https://github.com/bazelbuild/bazel/issues/6359. We also tested the `--experimental_remote_downloader` option, which did not work either in our setup.

jheidbrink (approved):

.github/workflows/bazel.yml
```diff
@@ -131,6 +88,17 @@ jobs:
           options: -v ${{ github.workspace }}:/workspaces/magma/
           run: |
             cd /workspaces/magma
+            if [ -n "${{ secrets.BAZEL_REMOTE_PASSWORD }}" ]
```
nstng: this setup is needed here in two places and we plan to use it in other workflows as well - can this be refactored in a script file with parameters pw and cache key?

bazel/bazelrcs/remote_caching_ro.bazelrc
LKreutzer: Comments added

.github/workflows/bazel.yml
jheidbrink: We cannot use the remote cache as a repository cache, but can we use `--repository_cache` flag to have an on-disk repository cache as before?

LKreutzer: We tested it and in bazel 4.2.2 it seems that the local `repository_cache` is ignored if a remote cache is present, same as for the disk cache.

.github/workflows/bazel.yml
LKreutzer: Added script and refactoring.

See CI runs with the changes:
- [With secret on fork](https://github.com/LKreutzer/magma/actions/runs/2337887288)
- [Without secret on PR](https://github.com/magma/magma/actions/runs/2337876477)

nstng (approved):
LGTM

PR #12744, opened 2022-05-17
refactor(orc8r): Remove unary package from orc8r lib

voisey:

## Summary

This removes a unused function, `CloudClientInterceptor`, from the unary package in the orc8r lib. This leaves the package with just one function, which is only called once in the orc8r. I have therefore moved this function to the file where it's called, along with the constants it uses.

## Test Plan

- [x] CI


orc8r/lib/go/registry/interceptors.go
```diff
@@ -45,5 +52,24 @@ func TimeoutInterceptor(ctx context.Context, method string, req, resp interface{
 func CloudClientTimeoutInterceptor(ctx context.Context, method string, req, resp interface{},
 	cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 
-	return TimeoutInterceptor(unary.OutgoingCloudClientCtx(ctx), method, req, resp, cc, invoker, opts...)
+	return TimeoutInterceptor(OutgoingCloudClientCtx(ctx), method, req, resp, cc, invoker, opts...)
+}
+
+// OutgoingCloudClientCtx amends outgoing cloud client context with magic client CSN metadata
+// if the original context doesn't already have client certificate serial number metadata
+func OutgoingCloudClientCtx(ctx context.Context) context.Context {
```
sebathomas:
```suggestion
func outgoingCloudClientCtx(ctx context.Context) context.Context {
```
I think this could be private now.

voisey: Good point, thanks! See https://github.com/magma/magma/pull/12744/commits/060e5b323367812baee3f07f701b825d4ed36274

sebathomas (approved):

maxhbr (approved):
Looks like a good refactoring that aligns with the initial idea of ab6dfc859556cb21700de0f280ac72f9e7f1180e, which created the client_identity.go

PR #12749, opened 2022-05-17
feat(ci): Add cwag formatting to ci and apply formatting changes

voisey:

## Summary

Running the precommit check `make -C ${MAGMA_ROOT}/cwf/gateway precommit` on master on the cwag_test VM leads to some formatting changes. This PR adds these changes, as well as adding a check to the CI to make sure that such formatting problems aren't introduced in the future.

To test the modification to the CI, I created a PR in which only the CI is modified: https://github.com/magma/magma/pull/12753. As expected, cwag-workflow breaks on this PR due to there being unstaged files: https://github.com/magma/magma/runs/6472265618?check_suite_focus=true, while the corresponding check is green on this PR, due to the files being added.

## Test Plan

- [x] cwag_test:  `make -C ${MAGMA_ROOT}/cwf/gateway test`
- [x] Host: `cd $MAGMA_ROOT/cwf/gateway && fab integ_test`
- [x] CI

### DIFF
--- a/.github/workflows/cwag-workflow.yml
+++ b/.github/workflows/cwag-workflow.yml
@@ -58,6 +58,10 @@ jobs:
             make -C ${MAGMA_ROOT}/cwf/gateway precommit
             cd ${MAGMA_ROOT}/cwf/gateway
             make -C ${MAGMA_ROOT}/cwf/gateway/integ_tests precommit
+      - name: Check precommit has not generated formatting changes
+        run: |
+            git status
+            git diff-index --quiet HEAD
       - name: Extract commit title
         if: failure() && github.event_name == 'push'
         id: commit
--- a/.github/workflows/cwag-workflow.yml
+++ b/.github/workflows/cwag-workflow.yml
@@ -54,14 +54,13 @@ jobs:
           timeout_minutes: 10
       - name: Run precommit
         run: |
-            cd ${MAGMA_ROOT}/cwf/gateway
             make -C ${MAGMA_ROOT}/cwf/gateway precommit
-            cd ${MAGMA_ROOT}/cwf/gateway
             make -C ${MAGMA_ROOT}/cwf/gateway/integ_tests precommit
       - name: Check precommit has not generated formatting changes
         run: |
             echo "Checking for changes caused by 'make -C \${MAGMA_ROOT}/cwf/gateway precommit' and 'make -C \${MAGMA_ROOT}/cwf/gateway/integ_tests precommit'."\
                  "Run these commands locally to see the respective changes."
+            cd ${MAGMA_ROOT}
             git status
             git diff-index --quiet HEAD
       - name: Extract commit title
--- a/.github/workflows/cwag-workflow.yml
+++ b/.github/workflows/cwag-workflow.yml
@@ -60,6 +60,8 @@ jobs:
             make -C ${MAGMA_ROOT}/cwf/gateway/integ_tests precommit
       - name: Check precommit has not generated formatting changes
         run: |
+            echo "Checking for changes caused by 'make -C \${MAGMA_ROOT}/cwf/gateway precommit' and 'make -C \${MAGMA_ROOT}/cwf/gateway/integ_tests precommit'."\
+                 "Run these commands locally to see the respective changes."
             git status
             git diff-index --quiet HEAD
       - name: Extract commit title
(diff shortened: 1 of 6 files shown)


sebathomas:
This could probably be part of the CI insync check?

voisey:
> This could probably be part of the CI insync check?

According to the [wiki](https://github.com/magma/magma/wiki/Contributing-Code#continuous-integration-ci--continuous-deployment-cd), the insync check is to "Ensure generated files are committed", so I think it's a bit of a different case. It is actually part of the [cwag-workflow](https://github.com/magma/magma/blob/master/.github/workflows/cwag-workflow.yml#L58), but I think it needs to be modified so that it checks the diff and fails if it's not empty.

sebathomas:
> > This could probably be part of the CI insync check?
> 
> According to the [wiki](https://github.com/magma/magma/wiki/Contributing-Code#continuous-integration-ci--continuous-deployment-cd), the insync check is to "Ensure generated files are committed", so I think it's a bit of a different case. It is actually part of the [cwag-workflow](https://github.com/magma/magma/blob/master/.github/workflows/cwag-workflow.yml#L58), but I think it needs to be modified so that it checks the diff and fails if it's not empty.

Ah, I thought it also handled the formatter check for Go. But I believe this CI check exists somewhere at least for the orc8r Go files, maybe not as part of the insync check though...

voisey:
Hi @sebathomas, I have included the following commit https://github.com/magma/magma/pull/12749/commits/70bcc0487ff479821b68212bce85543f00fd73de, so that the CI now checks whether there are any formatting changes to be made

sebathomas (approved):
Awesome, thank you! :+1:

themarwhal (approved):

uri200 (approved):

.github/workflows/cwag-workflow.yml
```diff
@@ -58,6 +58,10 @@ jobs:
             make -C ${MAGMA_ROOT}/cwf/gateway precommit
             cd ${MAGMA_ROOT}/cwf/gateway
             make -C ${MAGMA_ROOT}/cwf/gateway/integ_tests precommit
+      - name: Check precommit has not generated formatting changes
+        run: |
+            git status
+            git diff-index --quiet HEAD
```
nstng, line 65: Is the precommit check something that can be run locally without much effort?
* if yes, does it make sense to print the command in the failure case? this is just nit, because it can be seen in the previous step
* if no, does it make sense to show the actual diff?

sebathomas: I think that's the purpose of the status command. It will show a list of changed files.

nstng: But it is not visible, what needs to be changed. I.e., I would need to run locally
```
cd ${MAGMA_ROOT}/cwf/gateway
make -C ${MAGMA_ROOT}/cwf/gateway precommit
cd ${MAGMA_ROOT}/cwf/gateway
make -C ${MAGMA_ROOT}/cwf/gateway/integ_tests precommit
```
In order to see the changes and commit them. In github CI I see that this take ~2-3 minutes, i.e., I think it is reasonable to do this manually. But the PR owner would have to fetch these commands from the .yml file -> does it make sense to print these commands to the output in case of changes?

sebathomas: The list of files is visible in the PR in case of an error (see https://github.com/magma/magma/runs/6472265618?check_suite_focus=true from the test plan).

But it might be nice to print the precommit commands that fix the issue, I agree.

nstng: Discussed offline, @voisey it would be nice to have an additional echo for the check, something like:
echo "Checking for changes caused by make -C ${MAGMA_ROOT}/cwf/gateway precommit. Run this command locally to see respective changes."

voisey: Nice idea, thanks! Are you happy with this commit? https://github.com/magma/magma/pull/12753/commits/5b0556e74aea9a30d4d027fa7b222f9639f3a9be. You can see how it looks in the CI here: https://github.com/magma/magma/runs/6487122263?check_suite_focus=true. If so, I can cherry pick it into this PR.

sebathomas:
I think you need to rebase, there's a conflict now.

.github/workflows/cwag-workflow.yml
nstng: /me is happy :) thank you!

voisey:
@nstng @sebathomas I've also added this https://github.com/magma/magma/pull/12749/commits/09dc7dca9cca3fc27ffd7a81ee09721b4ba2033b, which removes a couple of unnecessary directory changes - they aren't needed because `make` is called with the `-C` flag.

nstng (approved):
lgtm :)

PR #12751, opened 2022-05-17
chore(orc8r): Add sql locker to sqorc lib

jkmar:
## Summary

Not all sql dialects support locking feature (SELECT FOR UPDATE),
so in order to use locking for different dialects,
a locker interface was introduced.

### DIFF
--- a/orc8r/cloud/go/sqorc/locker.go
+++ b/orc8r/cloud/go/sqorc/locker.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2022 The Magma Authors.
+
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sqorc
+
+import (
+	"os"
+	"strings"
+)
+
+// GetSqlLocker returns a lock statement generator for the configured SQL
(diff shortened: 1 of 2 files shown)


jkmar:
@magma/approvers-orc8r 
The PR is ready for review (coverage flake is unrelated to these changes), please review

orc8r/cloud/go/sqorc/locker.go
```diff
@@ -0,0 +1,35 @@
+package sqorc
```
maxhbr, line 14: missing license header

orc8r/cloud/go/sqorc/locker_test.go
```diff
@@ -0,0 +1,66 @@
+package sqorc_test
```
maxhbr, line 14: missing license header

maxhbr (commented):
This code looks unused. Can you either document, how it is used or add usages? It could also be helpful to add more  documentation

orc8r/cloud/go/sqorc/locker.go
jkmar: added

orc8r/cloud/go/sqorc/locker_test.go
jkmar: added

jkmar:
> This code looks unused. Can you either document, how it is used or add usages? It could also be helpful to add more documentation

@maxhbr
We need this change in domain proxy service, where we need to use `SELECT FOR UPDATE`,
but the point is that this construction is not supported in SQLite which is used for unit tests,
so this is why we need to have different behavior depending on SQL dialect,
and the sqorc lib is the place for defining such things.
We also wanted to keep PRs small, so that is why it contains only the library code,
it is unused now (although functions are exported), but it will be used in the upcoming domain proxy PRs

PR #12757, opened 2022-05-17
feat(nms): generate swagger api for typescript

### LINKED ISSUE #12756: Fix swagger.yml issue and generate typescript API
Generate Swagger API for TypeScript in parallel to the one for Flow

Neudrino:
## Summary
Generate the Swagger API for TypeScript in parallel to the one for Flow.
This is a preparatory step for migrating the NMS to TypeScript.

The tool the we are using for the flow API generation is discontinued and generates a broken TypeScript API.
Therefore we had to use a different tool for TypeScript. 

We found an issues in our current swagger spec that we had to work around. There are three endpoints that return a differently shaped response based on a query parameter. This has been realised by putting the URL a second time into the swagger spec with a hard coded query parameter, which is not allowed. We should probably fix that by splitting the endpoints at some point.

Another problem that we ran into is more related to the generation tool. It will create types with reserved keywords. Since this only effected a very sparingly used subtype we renamed the type to unblock us.  

The TypeScript API will be added in a separate PR https://github.com/magma/magma/pull/12761.

Fixes #12756

Done in pairing with @thmsschmitt .


nms/scripts/generateAPIFromSwaggerForTypeScript.sh
```diff
@@ -0,0 +1,63 @@
...
+ * This file is generated by "nms/scripts/generateAPIFromSwaggerForTypeScript.sh".
+ */
+'
+
+# shellcheck disable=SC2044
+for file in $(find "${OUTPUT}" -type f); do
+  TEMPORARY_FILE=$(mktemp)
+  (echo "${HEADER}"; cat "${file}") > "${TEMPORARY_FILE}" && mv "${TEMPORARY_FILE}" "${file}"
+  sed -i -e 's/?view=full(/_view_full(/g' "${file}"
+  sed -i -e 's/?verbose=false(/_verbose_false(/g' "${file}"
```
sebathomas: I assume this is the workaround for the hard-coded query parameters? Maybe add a TODO to remove them.

sebathomas (approved):
Looking good.

thmsschmitt (approved):

maxhbr:
> Another problem that we ran into is more related to the generation tool. It will create types with reserved keywords. Since this only effected a very sparingly used subtype we renamed the type to unblock us.

Changing the name of the type is not breaking API, right?

thmsschmitt:
> Changing the name of the type is not breaking API, right?

Correct, that is only an internal thing.

nms/scripts/generateAPIFromSwaggerForTypeScript.sh
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @generated
+ * This file is generated by "nms/scripts/generateAPIFromSwaggerForTypeScript.sh".
+ */
+'
+
+# shellcheck disable=SC2044
+for file in $(find "${OUTPUT}" -type f); do
+  TEMPORARY_FILE=$(mktemp)
+  (echo "${HEADER}"; cat "${file}") > "${TEMPORARY_FILE}" && mv "${TEMPORARY_FILE}" "${file}"
```
maxhbr: some suggestions that might result in cleaner code:

Maybe define a function like:
```
addHeader() {
    local file="$1"
    EMPORARY_FILE=$(mktemp)
    cat <<EOF > "${TEMPORARY_FILE}"
  /**
 * Copyright 2022 The Magma Authors.
 *
 * This source code is licensed under the BSD-style license found in the
 * LICENSE file in the root directory of this source tree.
 *
 * Unless required by applicable law or agreed to in writing, software
 * distributed under the License is distributed on an "AS IS" BASIS,
 * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 * See the License for the specific language governing permissions and
 * limitations under the License.
 *
 * @generated
 * This file is generated by "nms/scripts/generateAPIFromSwaggerForTypeScript.sh".
 */
 $(cat "${file})")
 EOF
     mv "${TEMPORARY_FILE}" "${file}"
}
```
and write:
```suggestion
  addHeader "${file}"
```

#### if `moreutils` would be installed, it can be even cleaner:
And there is a cli command called [`sponge`](https://linux.die.net/man/1/sponge) that (if present on the system) can get rid of the temporary file and you could just write
```suggestion
  (echo "${HEADER}"; cat "${file}") | sponge "${file}"
```

nms/scripts/generateAPIFromSwaggerForTypeScript.sh
```diff
@@ -0,0 +1,64 @@
...
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+set -e
```
maxhbr: better:
```suggestion
set -euo pipefail
```

nms/scripts/generateAPIFromSwaggerForTypeScript.sh
```diff
@@ -0,0 +1,64 @@
...
+  echo "$USAGE"
+  exit 1
+fi
+
+INPUT=${1}
+OUTPUT=${2}
+echo "Input file: $INPUT";
+echo "Output directory: $OUTPUT";
+
+yarn --silent openapi -i "${INPUT}" -o "${OUTPUT}"
```
maxhbr: maybe some helpful logging:
```suggestion
(set -x; 
 yarn --silent openapi -i "${INPUT}" -o "${OUTPUT}")
```

nms/scripts/generateAPIFromSwaggerForTypeScript.sh
```diff
@@ -0,0 +1,64 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @generated
+ * This file is generated by "nms/scripts/generateAPIFromSwaggerForTypeScript.sh".
+ */
+'
+
+# shellcheck disable=SC2044
+for file in $(find "${OUTPUT}" -type f); do
```
maxhbr: If someone accidentally uses an `$OUTPUT` folder that contains more files, this might break a lot. Please add a more restrictive way of finding the generated files.

maxhbr (commented):
a lot of nitpicking on Bash, but please replace the `find` in line 58 with something more specific.

orc8r/cloud/Makefile
```diff
@@ -202,7 +204,9 @@ nms_prereqs_osx:
 	npm --version || brew install npm
 	yarn --version || npm install --global yarn
 	yarn add swagger-js-codegen
+	yarn add openapi-typescript-codegen
 
 nms_gen:
 	# generate Swagger API bindings for NMS
 	$(MAGMA_ROOT)/nms/scripts/generateAPIFromSwagger.sh $(SWAGGER_V1_YML) $(SWAGGER_NMS_OUT)
+	$(MAGMA_ROOT)/nms/scripts/generateAPIFromSwaggerForTypeScript.sh $(SWAGGER_V1_YML) $(SWAGGER_NMS_OUT_TS)
```
maxhbr: I think it would be more idiomatic to call a `gen` target in the `nms` makefile. But since it was done that way also previously, it might be fine.

nms/scripts/generateAPIFromSwaggerForTypeScript.sh
thmsschmitt: We have to set `--client axios` because we also used this API from node in a few places.

nms/scripts/generateAPIFromSwaggerForTypeScript.sh
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @generated
+ * This file is generated by "nms/scripts/generateAPIFromSwaggerForTypeScript.sh".
+ */
+$(cat "${file}")
+EOF
+  mv "${TEMPORARY_FILE}" "${file}"
+}
+
+for file in "${OUTPUT}"/*.ts "${OUTPUT}"/**/*.ts; do
```
maxhbr, line 86: Are we sure all systems are already at bash 4? On mac there should still be no support for `**` in bash

thmsschmitt: The script is executed in a docker container, through `orc8r/cloud/docker/build.py --generate` although you could in theory execute it manually.
I'm on a Mac and the script doesn't work, but only because BSD `sed -i` is incompatible with GNU `sed -i`. Which we should be able to remove after https://github.com/magma/magma/issues/12768

maxhbr (approved):
LGTM

nms/scripts/generateAPIFromSwaggerForTypeScript.sh
```diff
@@ -0,0 +1,91 @@
...
+    i) INPUT="${OPTARG}" ;;
+    o) OUTPUT="${OPTARG}" ;;
+    h | *) usage ;;
+  esac
+done
+
+[[ -z "${INPUT}" ]] || [[ -z "${OUTPUT}" ]] && usage
+
+if [ "$FORCE" = false ] && [ -d "${OUTPUT}"  ] ; then
+    read -p "Are you sure you want to overwrite all files in ${OUTPUT} (y/N)? " REPLY
```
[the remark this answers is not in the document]
thmsschmitt: Should be fixed

maxhbr (approved):

Neudrino:
A follow up is at https://github.com/magma/magma/pull/12761.

PR #12761, opened 2022-05-17
refactor(nms): Add generated TypeScript API

### LINKED ISSUE #12747: Generate TypeScript API bindings
To be able to migrate to TypeScript API bindings need to be generated.
[swagger-js-codegen](https://www.npmjs.com/package/swagger-js-codegen) it is 3 years old and no longer maintained. It supports TypeScript but the generated API is a lot more verbose and every return type is `Promise<request.Response>`. Either we have to fix this or we need to use a different tool.

### Options

* [openapi-typescript-codegen](https://github.com/ferdikoomen/openapi-typescript-codegen) (seems to work well at first glance)
* [openapi-generator](https://openapi-generator.tech/) (The most standard option, but that needs a JVM)
See also https://tngtech.slack.com/archives/CLD2CLJ85/p1635340292001200

We have paths with query parameters in the swagger.yml. https://github.com/magma/magma/blob/131c294199c43b52f9437f956fe221642fffb465/orc8r/cloud/go/obsidian/swagger/v1/swagger.yml#L4751
This is not allowed https://swagger.io/docs/specification/2-0/paths-and-operations/#:~:text=Query%20String%20in%20Paths And at least openapi-typescript-codegen stumbles over this.

The package type needs to be renamed because it is a reserved keyword
https://github.com/magma/magma/blob/131c294199c43b52f9437f956fe221642fffb465/orc8r/cloud/go/obsidian/swagger/v1/swagger.yml#L10410

The goal of the story is to fix the swagger issues and to generate TypeScript API bindings next to the existing Flow bindings. The new bindings will unused for now.

### Subtasks

- [x] #12756
- [x] #12748

### LINKED ISSUE #12748: CI checks that all required generated files are checked in properly.

thmsschmitt:
# Summary

Switches to `openapi-generator` for TypeScript API generation. `openapi-typescript-codegen` that was added in #12757 skips the parameters of the path name when constructing method names. As a result paths link `/foo/{something}` and `/foo` get the same method name only differentiated by a counter.  
 
Adds generate TypeScript API. 

Fixes #12748


.gitignore
```diff
@@ -122,4 +123,4 @@ compile_commands.json
 
 # Domain Proxy
 dp/cloud/helm/dp/charts/domain-proxy/certificates/**/*
-orc8r/cloud/helm/orc8r/charts/secrets
```
thmsschmitt, line 125: Not my change

ShubhamTatvamasi (approved):
LGTM

nms/generated-ts/.openapi-generator/VERSION
```diff
@@ -0,0 +1 @@
+6.0.0
```
Neudrino, line 1: Should we add an EOL at EOF here?

maxhbr: looks like a generated file -> Ignore

maxhbr (approved):
orc8r part looks good. 

Just wondering if generated subfolders need to be listed somewhere, to be excluded from linting and other checks? It should probably also be added to https://github.com/magma/magma/blob/72e819c5861bec738f72acc824529939cdfd8212/CODEOWNERS#L76-L83

thmsschmitt:
> Just wondering if generated subfolders need to be listed somewhere, to be excluded from linting and other checks? It should probably also be added to ...

Linting is disabled for the generated files with an ignore comment `/* eslint-disable */`. I wasn't aware that there is a code ownership exclude list, I'll add it there.

Neudrino (approved):

Neudrino (approved):

maxhbr (approved):
also changes in `CODEOWNERS` look good

PR #12764, opened 2022-05-17
chore: stale default assignees for new features

lucasgonze:
Signed-off-by: Lucas Gonze <lucas@gonze.com>

## Summary

Clear default assignees for new feature requests

## Additional Information

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -3,7 +3,7 @@ name: "\U0001F680 Feedback form"
 about: Describe a problem, suggest a new idea or functionality for the project
 title: ''
 labels: 'type: feature'
-assignees: 
+assignees: magma/approvers-tsc
 
 ---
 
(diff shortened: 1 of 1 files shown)


lucasgonze:
I asked Aswin if he wanted to continue to triage feature requests. He said:

> This should be done by the TSC or by teams building on Magma who want to pick up the feature requests.

ekowtaylor:
Hi @lucasgonze , I think this change makes sense but unfortunately, I'm not a maintainer. I agree with @aswin-alexander that this should be handled by the TSC. I'll change the reviewers and point to them instead.

maxhbr (approved):
Looks good to me, and unless there is another proposal in https://magmacore.slack.com/archives/C01Q1T14YJ2/p1653586273189919, I think that this can be merged.

Neudrino (changes requested):
Assignees cannot be groups but only individual persons. So unfortunately this assignment will not work as proposed. You can add up to 10 individuals to an issue.

Further reading:
- https://docs.github.com/en/issues/tracking-your-work-with-issues/assigning-issues-and-pull-requests-to-other-github-users
- https://docs.github.com/en/enterprise-cloud@latest/rest/issues/assignees

The API call for available assignees does not give any groups. Please also double check in the web-UI. Groups are not select-able there.

maxhbr:
7. a github actions job could add assignees to newly created issues with the corresponding labels

lucasgonze:
> a github actions job could add assignees to newly created issues with the corresponding labels

That could work. I'm good with it, but the question of who would have ownership is out of scope for me. I suspect this project needs discussion at the TSC level. 

Let's consider closing it unmerged for the moment.

PR #12766, opened 2022-05-18
chore: Track all non bazelified AGW Python files

vktng:

## Summary

The non-bazelified Python files are tracked with GH issues. Files that do not need to be bazelified were moved in to the deny list. Deleted pkt_tester tester files were removed from the list.

## Test Plan

CI for removed deny list entries.
Other changes should not impact the build.

## Additional Information


nstng (approved):
lgtm - in pairing

PR #12767, opened 2022-05-18
fix(dp): Use retry for log fetching in tests

sklgromek:

ElasticSearch integration tests should use retry for log fetching

### DIFF
--- a/dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
+++ b/dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
@@ -52,10 +52,11 @@ class DomainProxyOrc8rTestCase(DomainProxyIntegrationTestCase):
         cbsd_id = self.given_cbsd_provisioned(builder)
 
         with self.while_cbsd_is_active():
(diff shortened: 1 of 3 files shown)


dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
```diff
@@ -52,13 +52,15 @@ def test_cbsd_sas_flow(self):
         cbsd_id = self.given_cbsd_provisioned(builder)
 
         with self.while_cbsd_is_active():
-            when_elastic_indexes_data()
+            @retry(stop_max_attempt_number=30, wait_fixed=1000)
+            def when_elastic_indexes_data():
```
jkmar: It would be nice if we move those retries to then_logs_are instead of creating a function here, but for that I think we need to ensure that elastic has proper timestamps, so maybe we need to do that task first

ShubhamTatvamasi (approved):
LGTM from Helm side.

xbend (approved):
LGTM

jkmar (approved):

PR #12787, opened 2022-05-19
refactor(orc8r): split function SetIdentityFromContext in middleware

wolfseb:
## Summary

Refactor parts of the function `SetIdentityFromContext` in orc8r middleware to separate functions to increase readability.
A Test for the function has also been created. The test cases are oriented by

- the possible cases described in the comment at the start of the function
- additional possible cases given in the code

The tests were first benchmarked on the original logic to ensure the refactoring did not change any potential outcome.

original_logic

refactored_logic

## Test Plan

Run unit tests and newly created test to ensure the same outcomes as before the refactoring.

## Additional Information

Created in pairing with @MoritzThomasHuebner 

### DIFF
--- a/orc8r/cloud/go/service/middleware/unary/identity_decorator.go
+++ b/orc8r/cloud/go/service/middleware/unary/identity_decorator.go
@@ -85,7 +85,6 @@ func init() {
 // SetIdentityFromContext will bypass the Identity checks for local callers
 // (other services on the cloud) and allowlisted RPCs (methods in
 // identityDecoratorBypassList)
-// TODO(hcgatewood): refactor this to clean up the convoluted logic
 func SetIdentityFromContext(ctx context.Context, _ interface{}, info *grpc.UnaryServerInfo) (newCtx context.Context, newReq interface{}, resp interface{}, err error) {
 	// There are 5 possible outcomes:
 	// 1. !ok -> type assertion: mdIncomingKey{} is present, but it's not of MD type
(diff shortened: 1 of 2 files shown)


orc8r/cloud/go/service/middleware/unary/identity_decorator.go
```diff
@@ -112,70 +112,97 @@ func SetIdentityFromContext(ctx context.Context, _ interface{}, info *grpc.Unary
 		return newCtx, newReq, resp, err
 	}
 
+	// check for allowlisted calls only once, since info does not change
+	// Bypasses for allowlisted methods are given further down, when the
+	// call would be rejected otherwise
+	allow := isAllowlistedCall(info)
```
emakeev: Alowlisted calls are supposed to be infrequent (bootstrapping happens once in 3-4 days, etc.), that's why original code first tries to handle most frequent Gateway client calls (client cert present & valid).

Here, you do a map lookup for something, you are not likely to use. Unneeded map lookup is not a huge deal, but: 1) it happens on every gRPC call to every orc8r service & 2) why do it if you have original code which works & avoids it?

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
```diff
@@ -112,70 +112,97 @@ func SetIdentityFromContext(ctx context.Context, _ interface{}, info *grpc.Unary
...
-			}
 		}
-	} else {
-		// No CSNs found, check there is also no Certificate Common Names
-		if _, ok = ctxMetadata[CLIENT_CERT_CN_KEY]; ok {
-			// CN header is present while SN header is missing - possible
-			// security hack, either both or neither of the headers should be
-			// set
-			glog.Infof("CCN is present without SCN in metadata: %+v", ctxMetadata)
+	} else if snok {
```
emakeev: I'd say that the original logic is more efficient (& subjectively easier to read): 
IF A and B 
else A... // checking A second time
vs 
IF A
  IF B
  else ...

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
```diff
@@ -112,70 +112,97 @@ func SetIdentityFromContext(ctx context.Context, _ interface{}, info *grpc.Unary
...
+		// Bypass method (Bootstrapper & Co.), shortcut...
+		return ok
 	}
+	return false
+}
 
-	// Only allow local clients if there is no previous errors (snok == false)
+// findIdentityFromCertificateSerialNumber will try to find the identity associated
+// with the one caller's Client Certificate Serial Number found in CTX
+func findIdentityFromCertificateSerialNumber(snlist []string, ctx context.Context, ctxMetadata metadata.MD) (newCtx context.Context, err error) {
```
emakeev: maybe makes sense to just pass a single string snlist[0] if the function is only using index 0. It'd also remove from the caller a burden of enforcing explicit & undocumented condition of passing of the lis list of len >= 1.

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
```diff
@@ -85,7 +85,6 @@ func init() {
 // SetIdentityFromContext will bypass the Identity checks for local callers
 // (other services on the cloud) and allowlisted RPCs (methods in
 // identityDecoratorBypassList)
-// TODO(hcgatewood): refactor this to clean up the convoluted logic
```
emakeev, line 88: Please see the review comment.

emakeev (dismissed):
The function could probably benefit from some cleanup, but I'd caution from taking the comment about convoluted logic as a proven fact & the reason to change the logic.

The original reason for the logic is an optimization for most frequent cases.
We assumed that the most frequent calls would come from Gateways (CSN is present and valid), 
second most frequent case - local clients (service/obsidian to service), this may not the case anymore since when orc8r was moved to kubernetes, interservice calls were decorated with special ORC8R_CLIENT_CERT_VALUE, and the least frequent case is the allowlisted APIs (bootstrapper & register).

Optimizing for the second most frequent case may be less relevant now & it also needs additional CTX/Metadata verifications to prevent erroneous client or hacking, but I'd suggest to keep in mind the most common case (a single CSN present & valid).

Maybe you can  use your isLocalClientCall, findIdentityFromCertificateSerialNumber & isAllowlistedCall functions to improve code readability but keep the original logic as?

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
wolfseb: Good point, thanks!

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
wolfseb: Ok, the code has been changed to check for allow listed calls only when necessary again. It should now be optimized again for the most common cases.

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
wolfseb: For the most common case it should now be the same efficiency, and for the case that there are multiple CSNs for the same key, it is just an if check of a bool more. I'd argue that having all distinct cases on the same level might be more readable, since there are quite a few cases (might be subjective)

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
```diff
@@ -102,82 +101,113 @@ func SetIdentityFromContext(ctx context.Context, _ interface{}, info *grpc.Unary
...
 		//   return newCtx, newReq, resp, ensureLocalPeer(ctx)
-		// but, it may present a se3curity risk
-		err = status.Error(codes.Unauthenticated, ERROR_MSG_NO_METADATA)
-		return newCtx, newReq, resp, err
+		// but, it may present a security risk
+		return newCtx, newReq, resp, noMetadata()
 	}
 
-	// First, try to find the caller's identity
 	snlist, snok := ctxMetadata[CLIENT_CERT_SN_KEY]
```
maxhbr, line 112: Might be even easier to read, if you would extract [line 112 till 123](https://github.com/magma/magma/pull/12787/files#diff-ad4e5f2d268e6211a55db9855205cdfeec4eb63ad3c5abf7899a74d7302c22b7R112-R123) into a function (e.g. `findIdentity`):
```suggestion
	newCtx, err = findIdentity(ctx, ctxMetadata)
	if err == nil {
		return newCtx, newReq, resp, nil
	}
```
with
```go
func findIdentity(...) ... {
	snlist, snok := ctxMetadata[CLIENT_CERT_SN_KEY]
	if snok {
		return findIdentityFromCSNlist(snlist, ctx, ctxMetadata)
	} else if _, ok = ctxMetadata[CLIENT_CERT_CN_KEY]; ok {
		// CN header is present while SN header is missing - possible
		// security hack, either both or neither of the headers should be
		// set
		return nil, unexpectedCCN(ctxMetadata)
	}
}
```

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
```diff
@@ -102,82 +101,113 @@ func SetIdentityFromContext(ctx context.Context, _ interface{}, info *grpc.Unary
...
-		if _, ok := identityDecoratorBypassList[info.FullMethod]; ok {
-			// Bypass method (Bootstrapper & Co.), shortcut...
-			return newCtx, newReq, resp, nil
-		}
+	if isAllowlistedCall(info) {
+		return newCtx, newReq, resp, nil
 	}
 
-	// Only allow local clients if there is no previous errors (snok == false)
 	if err == nil {
```
maxhbr: is the `err==nil` case reachable? in https://github.com/magma/magma/blob/830bf9beee8fc4b59351ad942d6ac4701bbaa1b2/orc8r/cloud/go/service/middleware/unary/identity_decorator.go#L113-L123 it looks like that both branches end on a non-nil `err`:

- if `err==nil` in the first branch, it is returned early
- `unexpectedCCN` never returns `nil`

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
```diff
@@ -102,82 +101,113 @@ func SetIdentityFromContext(ctx context.Context, _ interface{}, info *grpc.Unary
...
+}
+
+func findIdentityFromUniqueCSN(sn string, ctx context.Context, ctxMetadata metadata.MD) (newCtx context.Context, err error) {
+	// Check if SN is the reserved value used for all inter-orc8r calls
+	if sn == registry.ORC8R_CLIENT_CERT_VALUE {
+		return newCtx, nil
+	}
+	var gwIdentity *protos.Identity
+	var certExpTime int64
+	gwIdentity, certExpTime, err = findGatewayIdentity(ctx, sn, ctxMetadata)
```
maxhbr, line 149: isn't that equivalent to:
```suggestion
	gwIdentity, certExpTime, err := findGatewayIdentity(ctx, sn, ctxMetadata)
```

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
```diff
@@ -102,82 +101,113 @@ func SetIdentityFromContext(ctx context.Context, _ interface{}, info *grpc.Unary
...
+		newCtx, err := findIdentityFromUniqueCSN(snlist[0], ctx, ctxMetadata)
+		return newCtx, err
+	} else {
+		// there is a certificate serial number (CSN) list in CTX
+		// there can be only one CSN, error out if not
+		return nil, multipleCSN(ctxMetadata)
+	}
+}
+
+func findIdentityFromUniqueCSN(sn string, ctx context.Context, ctxMetadata metadata.MD) (newCtx context.Context, err error) {
```
maxhbr: why is `err` passed to that function, if it is not used?

maxhbr (commented):
some comments

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
wolfseb: Yes, when `ok==false` in the `else if` bracket, which is the case when both CSN and CCN are missing. It is a bit cryptic to see that, I tried refactoring a bit more and reversing the last cases (i.e. asking for `!ok` instead of `ok`) to make it more clear.

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
wolfseb: The problem is then to handle the return in line 114, since `findIdentity` can also return `err=nil` if both `if` brackets are skipped (i.e. when both CSN and CCN are missing). I changed the code so that each `if` bracket returns no matter what, to make it a bit more clear what is happening.

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
wolfseb: See comment below.

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
wolfseb: It isn't passed to it, it is defined within it. In go you can define variables for the return values in the function declaration like this (the second bracket are the return values, the first bracket are the variables passed to the function), so you can just write e.g. `err = ...` instead of `err := ...` within the function. 
This is also the reason why the variables `gwIdentity` and `certExpTime` are defined the way they are, because `err` is already defined. The main reason to do it like this is probably to have named return values, especially for `newCtx`, because statements like `return nil, nil` are a bit cryptic (or we'd have to define `newCtx` explicitly).
Unfortunately, we have to give names to either all return values or none.

maxhbr: Ah sorry, my misstake :). Failed to read the function definition correctly

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
maxhbr: ah, `else if` and not `else` :)

orc8r/cloud/go/service/middleware/unary/identity_decorator.go
maxhbr: Yes, missed the `else if` here again

maxhbr (approved):
LGTM

voisey:
Hi @emakeev, could you please take another look at this if you get the chance to see whether the updates have addressed your comments? Thanks!

wolfseb:
Hi @emakeev, could you please take another look at this if you get the chance to see whether the updates have addressed your comments? Thanks!

PR #12789, opened 2022-05-19
chore(dp): Add timestamps to DP logs

sklgromek:

Domain Proxy logs timestamp in ES should equal to logs generated date


ShubhamTatvamasi (approved):

jkmar (commented):
I'm afraid this is not enough (or is it?) and we have to change also the filtering

sklgromek:
> I'm afraid this is not enough (or is it?) and we have to change also the filtering

In fluentd ES output configuration, we set `time_key` as event_timestamp, in this case, records in ES will have @timestamp  value set based on event_timestamp

dp/cloud/python/magma/fluentd_client/dp_logs.py
```diff
@@ -55,9 +56,14 @@ def make_dp_log(
     return func(*args)
 
 
+def now() -> str:
+    return datetime.now(timezone.utc).isoformat()
+
+
 def _make_log_from_db_request(request: 'DBRequest') -> DPLog:  # noqa: F821
```
jkmar, line 63: It would be nice to have some UTs for this

jkmar (approved):

dp/cloud/python/magma/radio_controller/tests/unit/test_dp_logs.py
```diff
@@ -26,8 +27,10 @@
 SOME_SERIAL_NUMBER = 'some_serial_number'
 SOME_NETWORK_ID = 'some_network_id'
 CBSD_REGISTER = 'CBSDRegister'
+SOME_DATE = '2022-05-20T00:00:00+00:00'
 
 
+@freeze_time('2022-05-20')
```
jheidbrink: Should this be a reference to `SOME_DATE`? Also, given that this date is today, this might accidentally work only today, maybe choose a date in the past instead.

jheidbrink (approved):
I haven't thoroughly looked at the go changes but I assume you have tested them. From a CI perspective, the change to the path filter looks good to me :+1:

dp/cloud/python/magma/configuration_controller/tests/unit/test_dp_logs.py
```diff
@@ -114,3 +119,6 @@ def test_dp_log_created_from_db_response(self, with_cbsd, serial_num, fcc_id, ne
     def test_make_dp_log_returns_type_error_for_unknown_message_type(self):
         with self.assertRaises(TypeError):
             make_dp_log(IncorrectDPLog())
+
+    def test_datetime_now(self):
+        self.assertEqual(SOME_DATE, now())
```
jheidbrink, line 124: Isn't this only testing that the `@freeze_time` decorator works? I think that is already ensured because otherwise the `self.assertEqual(expected_log, actual_log)` lines should fail, right?

PR #12791, opened 2022-05-19
refactor(pipelined): Simplify treatment of default parameters

voisey:

## Summary

This tidies up a couple of cases in which default parameters are set in function bodies rather than being set when defining the function arguments. This also cleans up a couple of unnecessary comments I came across while making the changes.

## Test Plan

- [x] magma-dev: `make test_python_service UT_PATH=python/magma/pipelined/tests/`


pshelar (approved):
LGTM

PR #12803, opened 2022-05-20
feat(dp): Add enodebd update checker

jkmar:
## Summary

It checks if cbsd should be updated with parameters from enodebd

### DIFF
--- a/dp/cloud/go/services/dp/storage/diff.go
+++ b/dp/cloud/go/services/dp/storage/diff.go
@@ -1,0 +1,54 @@
+package storage
+
+import (
+	"math"
+)
+
+const (
+	radiusM            = 6371e3
+	distanceThresholdM = 10
+)
+
+func ShouldENodeBDUpdate(prev *DBCbsd, next *DBCbsd) bool {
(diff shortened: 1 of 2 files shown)


dp/cloud/go/services/dp/storage/diff.go
```diff
@@ -0,0 +1,54 @@
...
+	distanceThresholdM = 10
+)
+
+func ShouldENodeBDUpdate(prev *DBCbsd, next *DBCbsd) bool {
+	return canUpdate(prev, next) &&
+		(paramsChanges(next, prev) || coordinatesChanged(next, prev))
+}
+
+func canUpdate(prev *DBCbsd, next *DBCbsd) bool {
+	return !prev.CpiDigitalSignature.Valid && next.CbsdCategory.String == "A"
```
WojSad: && indoor_deployment == true. Also "a" should be lower case

jkmar: As discussed removed category completely

WojSad (approved):

PR #12817, opened 2022-05-23
refactor(nms): Start TypeScript migration

thmsschmitt:
## Summary

Initial setup for the NMS TypeScript migration. Makes it possible to use Flow and TypeScript in parallel.
Plus migration of a few files to TypeScript to test if the setup works.

**To be on the safe side we should not merge this before the 1.8 release branch has been created.**

# Test Plan

```
yarn test
yarn eslint .
yarn flow
yarn tsc
```


.github/workflows/nms-workflow.yml
```diff
@@ -85,6 +85,44 @@ jobs:
...
+        if: failure() && github.event_name == 'push'
+        uses: rtCamp/action-slack-notify@v2.2.0
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_CI }}
+          SLACK_TITLE: "NMS typescript failed"
+          SLACK_MESSAGE: "${{ steps.commit.outputs.title}}"
+          SLACK_USERNAME: "NMS workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+          SLACK_FOOTER: ' '
```
Neudrino, line 125: Do we really need this? I am in favour of dropping it.

nms/api/MagmaAPI.ts
```diff
@@ -0,0 +1,85 @@
+/**
+ * Copyright 2020 The Magma Authors.
```
Neudrino: Should be 2022, right?

nms/app/components/context/AppContext.d.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright 2020 The Magma Authors.
```
Neudrino: 2022?

nms/app/theme/default.d.ts
```diff
@@ -0,0 +1,97 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @format
```
Neudrino: Some of the files have the `@format`, some do not.
What is it good for? Can we have this consistent in all or non of the files?

Neudrino: Some of the files have the `@format`, some do not.
What is it good for? Can we have this consistent in all or non of the files?

nms/shared/types/index.d.ts
```diff
@@ -0,0 +1,22 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @flow
+ * @format
```
Neudrino: Should this not be removed for typescript files? Especially the `@flow` part?

Neudrino (commented):
It might make sense to split the infrastructure changes out into a second PR, which is independent and more easily merge-able.
One might also consider to split the actual typescript migration of the components to an extra PR, in contrast to the JavaScript environment setup?

We might consider using the [the tsc with ReviewDog](https://github.com/marketplace/actions/run-tsc-with-reviewdog) for a better Github integration.

Summary and Test plan should be added to the description. :wink:

nms/app/theme/default.d.ts
thmsschmitt: Copy paste error :( This is from prettier and has no effect since we don't use the require-pragma option https://prettier.io/docs/en/options.html#require-pragma

I'm trying to remove those together with the `@flow` during the migration.

.github/workflows/nms-workflow.yml
thmsschmitt: I don't need it, but if we drop it should be dropped from for everything in the NMS workflow

thmsschmitt:
> It might make sense to split the infrastructure changes out into a second PR

To be able to run `tsc` we need a need TypeScript config, so if would split it would have to wait for the initial setup.

Neudrino:
Closed in favour of the big **NMS migration PR**:
- https://github.com/magma/magma/pull/13098

PR #12821, opened 2022-05-23
refactor(nms): Move /data, /alerts and /views directories

HannaFar:

## Summary

Refactor nms top level directories 

## Additional Information


nms/app/alerts/cwfAlerts.js
```diff
@@ -14,7 +14,7 @@
  * @format
  */
 
-import type {prom_alert_config} from '../generated/MagmaAPIBindings';
+import type {prom_alert_config} from '../../generated/MagmaAPIBindings';
```
thmsschmitt: `sync.js` and everything else in `/alerts` is only imported from `alerts/routes.js`. Therefore everything should be moved to `server/alerts`.

thmsschmitt: The general plan is to have everything frontend related in `/app`, everything backend related in `/server`, everything that is shared between the two in `/shared` and then may a few special top level things like `/api` and `/generated`.

thmsschmitt (approved):
Looks good to me, but please move `/app/alerts` to `/server/alerts`.

Issue #12605, opened 2022-04-29
Consolidate Go installation across Magma infrastructure


Neudrino:
During the Go upgrade https://github.com/magma/magma/pull/12151, it was noticeable, that the Go installation method and installation source is different parts across Magma. This story aims to consolidate the Go installation across the Magma infrastructure

### User story

As a devops engineer I want to have only one source and method for Go installation in order to reduce the maintenance effort and have a single source of truth.

### Acceptance Criteria

- Go binaries are only fetched from one source.
  - This source should be the Magma Artifactory https://artifactory.magmacore.org .
- Only one version of Go should be used.
  - After merge of #12151 this should be Go version 1.18.
- The same installation method should be employed for comparable infrastructure.
  - I.e. all installations inside a docker image should be the same.
  - It should be taken care, that unnecessary files are cleaned up after installation.
    E.g. for docker installations, a starting point could be
    ```bash
    ARG GOLANG_VERSION="1.18"
    RUN GO_TARBALL="go${GOLANG_VERSION}.linux-amd64.tar.gz" \
     && wget --no-verbose https://artifactory.magmacore.org/artifactory/generic/${GO_TARBALL} \
     && tar --strip-components=2 -C /usr/local/bin -xzf ${GO_TARBALL} go/bin/go go/bin/gofmt \
     && rm ${GO_TARBALL}
    ```

PR #12824, opened 2022-05-23
chore: docker images use same go installation

ajahl:

## Summary
As introduced in #12605 , only one Go version should be used in the Dockerfiles. Now, all Dockerfiles use the Go binaries provided by the Magma Artifactory https://artifactory.magmacore.org/artifactory.

## Test Plan

* Build and start the DevContainer locally
* Pass  CI

## Additional Information

### DIFF
--- a/cwf/k8s/cwf_operator/docker/Dockerfile
+++ b/cwf/k8s/cwf_operator/docker/Dockerfile
@@ -20,9 +20,12 @@ RUN apt-get update && apt-get install -y bzr curl daemontools gcc
 
 # Install Golang 1.18
 WORKDIR /usr/local
-RUN curl https://artifactory.magmacore.org/artifactory/generic/go1.18.linux-amd64.tar.gz --remote-name --location && \
-    tar -xzf go1.18.linux-amd64.tar.gz && \
-    cp -r go/bin/* /usr/local/bin/
+ARG GOLANG_VERSION="1.18"
+RUN GO_TARBALL="go${GOLANG_VERSION}.linux-amd64.tar.gz" \
+ && curl https://artifactory.magmacore.org/artifactory/generic/${GO_TARBALL} --remote-name --location \
+ && tar -xzf ${GO_TARBALL} \
+ && ln -s /usr/local/go/bin/go /usr/local/bin/go \
+ && rm ${GO_TARBALL}
 
 ENV GOBIN /var/opt/magma/bin
 ENV PATH ${PATH}:${GOBIN}
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -77,9 +77,11 @@ RUN echo "Install general purpose packages" && \
 
 # Install golang 1.18
 WORKDIR /usr/local
-RUN curl https://artifactory.magmacore.org/artifactory/generic/go1.18.linux-amd64.tar.gz --remote-name --location && \
-    tar -xzf go1.18.linux-amd64.tar.gz && \
-    rm go1.18.linux-amd64.tar.gz
+ARG GOLANG_VERSION="1.18"
+RUN GO_TARBALL="go${GOLANG_VERSION}.linux-amd64.tar.gz" \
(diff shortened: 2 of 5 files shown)


ShubhamTatvamasi (approved):
LGTM

cwf/k8s/cwf_operator/docker/Dockerfile
```diff
@@ -20,9 +20,12 @@ RUN apt-get update && apt-get install -y bzr curl daemontools gcc
...
 # Install Golang 1.18
 WORKDIR /usr/local
-RUN curl https://artifactory.magmacore.org/artifactory/generic/go1.18.linux-amd64.tar.gz --remote-name --location && \
-    tar -xzf go1.18.linux-amd64.tar.gz && \
-    cp -r go/bin/* /usr/local/bin/
+ARG GOLANG_VERSION="1.18"
+RUN GO_TARBALL="go${GOLANG_VERSION}.linux-amd64.tar.gz" \
+ && curl https://artifactory.magmacore.org/artifactory/generic/${GO_TARBALL} --remote-name --location \
+ && tar -xzf ${GO_TARBALL} \
+ && cp -r go/bin/* /usr/local/bin/ \
```
Neudrino: If there is more than the go binary in the archive, we will have unwanted files in the image this way. If we would only extract the needed files, this could be avoided.
https://github.com/magma/magma/blob/a68bf5d633b8167cc5b96196e2dd0cdbfa2841c6/orc8r/cloud/deploy/orc8r_deployer/docker/Dockerfile#L48-L53

cwf/k8s/cwf_operator/docker/Dockerfile
```diff
@@ -20,9 +20,12 @@ RUN apt-get update && apt-get install -y bzr curl daemontools gcc
 
 # Install Golang 1.18
 WORKDIR /usr/local
-RUN curl https://artifactory.magmacore.org/artifactory/generic/go1.18.linux-amd64.tar.gz --remote-name --location && \
-    tar -xzf go1.18.linux-amd64.tar.gz && \
-    cp -r go/bin/* /usr/local/bin/
+ARG GOLANG_VERSION="1.18"
+RUN GO_TARBALL="go${GOLANG_VERSION}.linux-amd64.tar.gz" \
+ && curl https://artifactory.magmacore.org/artifactory/generic/${GO_TARBALL} --remote-name --location \
```
Neudrino, line 25: _optional_: I personally prefer `wget`, as it is a standard tool in Ubuntu, while `curl` is not and needs to be installed explicitly. However, as long as we are consistent in this file it is probably fine.

cwf/k8s/cwf_operator/docker/Dockerfile
Neudrino: This actually also holds for all other occurrences.
However, we could also avoid the separate copy-step here, by making use of all tar functionality.

cwf/k8s/cwf_operator/docker/Dockerfile
ajahl: I have tried the `wget` command before but it is also not part of the used base image.

cwf/k8s/cwf_operator/docker/Dockerfile
ajahl: To build the Dockerfile, Go has to be in both folders, `/usr/local/bin/` and `/usr/local/go/bin/`.
While the binaries are needed in `/usr/local/bin/` to run `go`, `go install ...` requires the content of the Tar file in `/usr/local/go/bin/`, including the source code and additional libs.

Neudrino: While I do not know, how big the package is, it is than usually a good idea to only sym-link the required binary in the `/usr/local/bin` directory than, imho.

ajahl: I replaced the copy instruction with a sym-link.

themarwhal (approved):
as long as cwf integ / feg integ pass lgtm!

uri200 (approved):

PR #12826, opened 2022-05-23
refactor(nms): Move WebClient.js to /app directory

HannaFar:

## Summary

Move WebClient.js to /app directory

## Additional Information


thmsschmitt:
We are not going to do this because the file will anyhow be removed during the TypeScript migration.

PR #12833, opened 2022-05-24
chore(pipelined): Fix selection of mypy errors

voisey:
## Summary

Running `mypy --ignore-missing-imports magma/lte/gateway/python/magma/pipelined` leads to 154 errors on master, which is reduced to 105 with this PR. This PR addresses many of the simpler fixes, with further fixes to come in a subsequent PR.

## Test Plan

- [x] magma-dev: `make test_python_service UT_PATH=python/magma/pipelined/tests/`

### DIFF
--- a/lte/gateway/python/magma/pipelined/qos/common.py
+++ b/lte/gateway/python/magma/pipelined/qos/common.py
@@ -15,7 +15,7 @@ import logging
 import threading
 import traceback
 from enum import Enum
-from typing import Dict, List  # noqa
+from typing import Dict, List, Tuple  # noqa
 
 from lte.protos.policydb_pb2 import FlowMatch
 from magma.common.redis.client import get_default_client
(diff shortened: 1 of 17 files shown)


lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -155,7 +155,7 @@ def remove_rule(self, rule_num: int) -> None:
     def find_rule(self, rule_num: int):
         return self.rules.get(rule_num)
 
-    def get_all_rules(self) -> List:
+    def get_all_rules(self) -> Dict:
```
sebathomas: Do you know what kind of dict it is (e.g. `Dict[str, int]`)? In general it's good to be more specific with container types.

sebathomas (approved):
Thank you for cleaning that up! :+1:

lte/gateway/python/magma/pipelined/qos/common.py
voisey: Good point. It seems to me it's `Dict[int, List[Tuple[FlowMatch.Direction, str]]]`. Shall I add this or do you think it's too long?

sebathomas: I think it's still good to have it even if it's not a very intuitive type... :)

voisey: Done!

pshelar (approved):

PR #12834, opened 2022-05-24
chore(feg): drop fbgo support

maxhbr:
## Summary

This drops an unused branch from a script that executes go.

This also fixes a path variable and adds a `.gitignore` file to exclude the `./radius/` build output folder.


feg/radius/src/run.sh
```diff
@@ -13,13 +13,6 @@
 ################################################################################
 
 OUTPUT_DIR=bin
```
maxhbr: Just tested, the output us placed under `radius` and not `bin`:
```suggestion
OUTPUT_DIR=radius
```

feg/radius/src/run.sh
```diff
@@ -12,14 +12,7 @@
...
 
-OUTPUT_DIR=bin
-WHICH_FBGO=$(command -v fbgo)
-WHICH_GO=$(command -v go)
-if [ -z "${WHICH_FBGO}" ]; then
-    GO=${WHICH_GO}
-else
-    GO=${WHICH_FBGO}
-fi
+OUTPUT_DIR=radius
```
Neudrino: - Are you sure this does not collide with the `radius` executable?
https://github.com/magma/magma/pull/12834/files#diff-6042bb9d924b2bd61a52556741827a08460a7f32c4ab4a435b2fac8ae4c16799R47
- This variable is only used once afaiks. So it might not be necessary at all?
https://github.com/magma/magma/pull/12834/files#diff-6042bb9d924b2bd61a52556741827a08460a7f32c4ab4a435b2fac8ae4c16799R42

maxhbr: The links do not work for me.

> * Are you sure this does not collide with the `radius` executable?

this just touches the cleanup function here. If I do `./run.sh build` I get a file `./radius` and I expect the clean command to get rid of that. There is no folder `bin` created. So probably the name `_DIR` should be fixed.

> * This variable is only used once afaiks. So it might not be necessary at all?

probably, but wanted to keep changes minimal, but inlining would also fix the above issue

crasu:
Looks fine

Neudrino (approved):

feg/radius/src/run.sh
Neudrino: > The links do not work for me.

They are probably pointing to a diff with a specific commit. If you force pushed, that commit might be gone, such that the diff is invalid. :man_shrugging:

crasu (approved):

crasu (approved):

crasu (approved):

PR #12836, opened 2022-05-24
fix(ci): Fix dockerfile build agw

tmdzk:


ymasmoudi (approved):
grpc-dev and  prometheus-cpp-dev moved to debian-test
LGTM

tmdzk:
UBN this is unbreaking agw container build (one remaining problem with oai)

PR #12837, opened 2022-05-25
fix(deps): Removing redundant deb packages on aarch64

m-govind:
fix(deps): Removing redundant deb packages


tmdzk (approved):

ymasmoudi (commented):
@m-govind you could use a compact one liner here instead.  [[ ]] && apt-get install -y ...

ymasmoudi (approved):

PR #12848, opened 2022-05-26
feat(feg): Added health metric for s8_proxy servicer

### LINKED ISSUE #12555: [FEG] S8 proxy add Health service
Some FEG services like [s6a proxy]( http://github.com/magma/magma/blob/master/feg/gateway/services/s6a_proxy/metrics/metrics.go#L35-L35), [session_proxy](http://github.com/magma/magma/blob/master/feg/gateway/services/session_proxy/metrics/metrics.go#L1-L1) have health/metrics service. S8_proxy is missing that

To add this service to s8_proxy you can follow the other `metrics.go` files you may find on our codebase (please check the link examples on this issue). You will  need to create metrics for all the request (Create Session, Create Bearer, Delete Bearer, etc)

MoritzThomasHuebner:
Fixes #12555 

## Summary

- Added `metrics.go` file with Session/Bearer Create/Delete Request/Fails counters
- Added health tracker and `GetHealthStatus` method to `s8_proxy.go`
- Added increments for the health metric counters in `s8_proxy.go`
- Added the S8 Proxy to the metricsd collection in `magmad.yml`
 
## Test Plan

- Run unit tests
- Run some manual tests to see if the counter actually work
- - Connected a local Orc8r with a local FEG
- - Run `docker-compose exec s8_proxy /var/opt/magma/bin/s8_cli cs -server -test` to create a faulty session creation request
- - Check in Prometheus via localhost:9090 if the metrics increment.

After running `docker-compose exec s8_proxy /var/opt/magma/bin/s8_cli cs -server -test` we see that the metrics have incremented three times as expected.

Screenshot from 2022-06-08 11-33-37
Screenshot from 2022-06-08 11-33-53

## Additional Information


feg/gateway/services/s8_proxy/metrics/metrics.go
```diff
@@ -0,0 +1,169 @@
+/*
+Copyright 2020 The Magma Authors.
```
sebathomas:
```suggestion
Copyright 2022 The Magma Authors.
```

feg/gateway/services/s8_proxy/servicers/s8_proxy.go
```diff
@@ -220,3 +241,46 @@ func validateDeleteSessionRequest(dsr *protos.DeleteSessionRequestPgw) error {
 	}
 	return nil
 }
+
+// GetHealthStatus retrieves a health status object which contains the current
+// health of the service
+func (s *S8Proxy) GetHealthStatus(ctx context.Context, req *orc8r_protos.Void) (*protos.HealthStatus, error) {
```
sebathomas, line 255: I'm a bit confused about how the health metrics are supposed to work in general (also the existing metrics for s6a). We have a public method in the servicer but it's not part of the GRPC interface and there are no callers?

feg/gateway/services/s8_proxy/servicers/s8_proxy.go
```diff
@@ -64,30 +66,35 @@ func NewS8ProxyWithEcho(config *S8ProxyConfig) (*S8Proxy, error) {
...
+		config:        config,
+		gtpClient:     cli,
+		healthTracker: metrics.NewS8HealthTracker(),
 	}
 	addS8GtpHandlers(s8p)
 	return s8p, nil
 }
 
 func (s *S8Proxy) CreateSession(ctx context.Context, req *protos.CreateSessionRequestPgw) (*protos.CreateSessionResponsePgw, error) {
+	metrics.SessionCreateRequests.Inc()
```
sebathomas, line 78: If there are existing unit tests for the methods that now produce metrics, you could try to add assertions that the correct counters are incremented.

feg/gateway/services/s8_proxy/metrics/metrics.go
MoritzThomasHuebner: Fixed

feg/gateway/services/s8_proxy/servicers/s8_proxy.go
MoritzThomasHuebner: I checked the other proxies and there are no such tests, as far as I can tell. In order to access the metric count we need to introduce the helper function below. This does indeed provide the correct counts. 

```
func getMetricValue(counter prometheus.Counter) float64 {
	// Helper function to access the prometheus counter value
    channel := make(chan prometheus.Metric, 1)
    counter.Collect(channel)
    metric := dto.Metric{}
    _ = (<-channel).Write(&metric)
    return *metric.Counter.Value
}
```

feg/gateway/services/s8_proxy/servicers/s8_proxy.go
MoritzThomasHuebner: I am also unsure how this exactly happens, however, it appears that the proxies (other than the one I wrote myself...) implement the `ServiceHealthServer` interface in `feg/cloud/go/protos/service_health.pb.go`, which is used in some other parts.

I assume there is some magic happening that makes it all work in the end.

feg/gateway/services/s8_proxy/metrics/metrics.go
```diff
@@ -0,0 +1,195 @@
...
+		Help: "Total number of create bearer requests that failed.",
+	})
+	BearerDeleteRequests = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "bearer_delete_requests_total",
+		Help: "Total number of delete bearer requests.",
+	})
+	BearerDeleteFails = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "bearer_delete_fails",
+		Help: "Total number of delete bearer requests that failed.",
+	})
```
uri200, line 66: I don't remember how this is going to be displayed but note that all those metrics correspond to s8, so maybe it would be good to use a prefix to differentiate them from session_proxy metric 

```s8_bearer_create_fails```

File: feg/gateway/services/s8_proxy/servicers/s8_proxy.go (hunk 2 above)
Comment (uri200) [reply feg/gateway/services/s8_proxy/servicers/s8_proxy.go:255]: disable/enable functions were created for HA. When health was bad. orc8r will send a disable message to the servicer, and enable the other HA. That only applies to Diameter services because they have some state. For GTP services all the state is at MME, so disable, enable does not matter.

The 3 functions here define the health interface. That is needed so the service can be added as a `RegisterServiceHealthServer`

File: feg/gateway/services/s8_proxy/metrics/metrics.go (hunk 4 above)
Comment (MoritzThomasHuebner) [reply feg/gateway/services/s8_proxy/metrics/metrics.go:66]: Thanks, we just managed to get the metrics displayed in a local instance of prometheus. I think you are correct that the names are currently not differentiated enough. There are metrics from other services such as "create_session_lat" or "session_timeouts_total", which may be confusing. I added the prefix as you suggested.

File: feg/gateway/configs/magmad.yml [hunk 5]
```diff
@@ -66,7 +66,6 @@ mconfig_modules:
   - feg.protos.mconfig.mconfigs_pb2
 
 metricsd:
-  log_level: INFO
```
sebathomas, line 69: Why remove the log level here?

uri200: I remember that a  couple of years ago we removed log_level from the config files so it could be controlled by orc8r.  If you put it here, that value will overwrite whatever is coming from orc8r. 
I know this worked in AGW but not 100% sure if this is the same case for feg. 
In any case, if you want to remove it you can leave a commentary like here http://github.com/magma/magma/blob/master/lte/gateway/configs/magmad.yml#L15-L15

uri200 (approved):

feg/gateway/configs/magmad.yml
crasu: The log level inside the service config is not picked up only the one at line 13 in this file.

PR #12859, opened 2022-05-27
fix(pipelined): Remove problematic type annotation

sebathomas:
## Summary

The annotation breaks sudo tests that don't run in the CI even though it's actually correct. Must be some issue with the generated protobuf types.

## Test Plan

```
sudo /home/vagrant/build/python/bin/nosetests -s magma/pipelined/tests/test_service_manager.py
```

## Additional Information


nstng (approved):
tested locally, LGTM

PR #12860, opened 2022-05-27
chore(orc8r): Adapt bootstrap test script to run on Linux

sebathomas:
## Summary

Some of the orc8r [interface testing tools](https://docs.magmacore.org/docs/orc8r/dev_security#debug-tools) were not compatible with Linux commands before.

## Test Plan

I tested that it runs on Linux now.

```
cd $MAGMA_ROOT/orc8r/tools/scripts/
bash bootstrap.bash $HWID
```

## Additional Information


maxhbr (approved):
LGTM

PR #12873, opened 2022-05-30
docs: typo in Security Debugging docs

alexzurbonsen:
## Summary

Replace `.sh` with `.bash` to make command work properly.

### DIFF
--- a/docs/docusaurus/versioned_docs/version-1.5.X/orc8r/dev_security.md
+++ b/docs/docusaurus/versioned_docs/version-1.5.X/orc8r/dev_security.md
@@ -242,7 +242,7 @@ openssl verify \
     gateway.crt  # copied from target gateway
 
 # Consolidate Magma and its imported protos, copying to /tmp/magma_protos/
-${MAGMA_ROOT}/orc8r/tools/scripts/consolidate_protos.sh
+${MAGMA_ROOT}/orc8r/tools/scripts/consolidate_protos.bash
 
 # Generate compiled proto definitions
 cd /tmp/magma_protos/ && \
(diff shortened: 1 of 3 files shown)


Neudrino (changes requested):
`.bash` is a rather uncommon file extension. Usually those files should also be `.sh` files with an `#!/usr/bin/env bash` (or similar) comment in the first line.
So I think, we should rather correct the file extensions to comply with usual standards in contrast  to modifying the documentation to bake in the odds.

alexzurbonsen:
> `.bash` is a rather uncommon file extension. Usually those files should also be `.sh` files with an `#!/usr/bin/env bash` (or similar) comment in the first line. So I think, we should rather correct the file extensions to comply with usual standards in contrast to modifying the documentation to bake in the odds.

Sounds reasonable.

Neudrino (approved):

voisey:
@alexzurbonsen Should this be updated in previous versions of the docs too?

voisey (changes requested):
It seems the file is .bash for magma versions >=1.5 (see https://github.com/magma/magma/commit/0c03dfeb007ac7e71f9cfaae230702be7e7c0a9c), so if you could please fix all versioned docs for >=1.5, that would be great!

alexzurbonsen:
> It seems the file is .bash for magma versions >=1.5 (see [0c03dfe](https://github.com/magma/magma/commit/0c03dfeb007ac7e71f9cfaae230702be7e7c0a9c)), so if you could please fix all versioned docs for >=1.5, that would be great!

Done :)

voisey (approved):

PR #12876, opened 2022-05-30
chore(dp): Make target with tests development POD

sklgromek:

Add `make` target which will leave `test-runner` POD,
for local tests running/debugging.

## Additional Information

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -51,6 +51,7 @@
 !orc8r/tools/ansible/roles/fluent_bit/files
 !orc8r/tools/docker/
 
+!dp/cloud/docker/python/test_runner/entrypoint.sh
 !dp/cloud/go/
 !dp/cloud/python/
 !dp/protos/
--- a/dp/tools/deployment/tests/test_runner_deployment_dev.yml
+++ b/dp/tools/deployment/tests/test_runner_deployment_dev.yml
@@ -2,9 +2,9 @@
 apiVersion: v1
 kind: Pod
 metadata:
-  name: test-runner-orc8r
+  name: test-runner-dev
   labels:
-    name: test-runner-orc8r
+    name: test-runner
     type: integration-tests
 spec:
   restartPolicy: Never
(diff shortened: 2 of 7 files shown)


dp/tools/deployment/tests/test_runner_deployment_orc8r.yml
```diff
@@ -16,7 +16,7 @@ spec:
     - image: test_runner_image
       name: test-runner-orc8
       imagePullPolicy: IfNotPresent
-      command: ["python"]
+      command: ["/entrypoint.sh"]
       args: ["-m", "pytest", "-vvv", "-s", "-m", "orc8r", "--junit-xml=/backend/test_runner/test-results/test_report.xml", "tests"]
```
jkmar, line 27: should args stay the same?

dp/tools/deployment/tests/test_runner_deployment_dev.yml
```diff
@@ -0,0 +1,38 @@
...
+  name: test-runner-dev
+  labels:
+    name: test-runner
+    type: integration-tests
+spec:
+  restartPolicy: Never
+  initContainers:
+    - name: wait-for-elasticsearch
+      image: curlimages/curl
+      command: ['sh', '-c', "until curl -sf http://elasticsearch:9200/_cluster/health; do echo \"waiting for elasticsearch...\"; sleep 0.2; done"]
```
Neudrino: Could we multi-line this command for better readability, please.

.dockerignore
```diff
@@ -51,6 +51,7 @@
 !orc8r/tools/ansible/roles/fluent_bit/files
 !orc8r/tools/docker/
 
+!dp/cloud/docker/python/test_runner/entrypoint.sh
```
Neudrino, line 54: Why would we want this, if we are adding this file to the docker image later on. Is that not contradicting?

sklgromek: All files are excluded in `.dockerignore` file by asterisk in the first line of this file, so I  have to explicitly exclude this location from it. If I need to be able to add this file to the image when using `$MAGMA_ROOT` as context for build.

dp/tools/deployment/tests/test_runner_deployment_dev.yml
sklgromek: Done

.dockerignore
Neudrino: I overlooked the `!` in the beginning of the line. Sorry! :mag:

Neudrino (approved):

jkmar (approved):

dp/tools/deployment/tests/test_runner_deployment_orc8r.yml
sklgromek: Yes, all of them are passed to python by `entrypoint.sh` when the `DEV_TESTS` env variable is not true

PR #12877, opened 2022-05-30
chore: bazel base can be build on arm and tests can be executed successfully

nstng:
## Summary

This change prepares the bazel base Docker container for CI runs on arm infrastructure.
* Bazelisk is downloaded for arm architecture if Docker flag `--build-arg DEB_PORT=arm64` is set (default is `amd64`)
* An obsolete arm liblfds patch is removed - this patch actually made the bazel build fail on arm. The patch does not seem to be applied anywhere else.

Note that this prepares to run the `bazel.yml` workflow on CI, but does not yet implement this! For this we need:
* github runners on arm machines
* a job that builds and persists the bazel base image for arm
* a job that executes the `bazel.yml` workflow on arm (based on the image above)

**Problems**:
* Bazel needs >8gb of memory and almost 28gb of space (including os, etc). The t4g.medium only has 4gb, the t4g.large 8gb which is still not enough. A swap file setup helps here - see test plan.

## Test Plan
This was tested manually:
* spin up t4g.large with 28gb of space and image `ami-00a0488e9d5582804` on `us-west-1b` (see [here](https://github.com/ShubhamTatvamasi/terraform-aws-spot-ec2-arm/blob/a6ee5e473a2541d6f16ad91e60e1b572cdd05689/local.tf#L8))
* ssh connect
* set up swap
```
sudo fallocate -l 4G /swapfile
sudo chmod 600 /swapfile
sudo mkswap /swapfile
sudo swapon /swapfile
persist for reboots:
  sudo vim /etc/fstab
  /swapfile swap swap defaults 0 0
verify:
  sudo swapon --show
```
* `git clone https://github.com/magma/magma.git`
* set up remote
```
cd magma
git remote add nils https://github.com/nstng/magma.git
git fetch nils
git switch -c bazel_base_arm nils/bazel_base_arm
```
* setup docker
```
sudo apt update
sudo apt install docker.io
sudo usermod -a -G docker $USER
reboot instance
```
* build bazel base image
```
cd ~/magma
docker build -f .devcontainer/bazel-base/Dockerfile --build-arg DEB_PORT=arm64 -t bb:latest .
```
* run bazel base image
```
export MAGMA_ROOT=~/magma
docker run -v ${MAGMA_ROOT}:/workspaces/magma/ -v ${MAGMA_ROOT}/lte/gateway/configs:/etc/magma/ -it bb:latest /bin/bash
```
* run `bazel build ...` and `bazel test ...`

green

## Additional Information


LKreutzer:
FYI it might be nice to use a [matrix job](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) for running the arm and amd `bazel.yml` workflows (once we are ready).

Neudrino (approved):
Thank you very much for the detailed and elaborate testing and documentation as test plan! :100:

LKreutzer (approved):
Tested locally

PR #12879, opened 2022-05-31
docs: updated url for go installation

ajahl:

## Summary

As introduced in #12605 , only the Go version from https://artifactory.magmacore.org/artifactory should be used.  The PR updates the installation guide.

## Additional Information

### DIFF
--- a/docs/readmes/basics/prerequisites.md
+++ b/docs/readmes/basics/prerequisites.md
@@ -66,7 +66,7 @@ Development can occur from multiple OS's, where **macOS** and **Ubuntu** are **e
    1. Download the tar file.
 
       ```bash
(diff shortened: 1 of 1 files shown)


Neudrino (commented):
Was the go update actually merged before the 1.7 release?
If so, the versioned documentation for v1.7 should also be updated.

Neudrino (approved):
If not, the change looks good to me.

ajahl:
> Was the go update actually merged before the 1.7 release? If so, the versioned documentation for v1.7 should also be updated.

The Go 1.18 upgrade is not part of 1.7.

voisey (approved):

PR #12881, opened 2022-05-31
chore(dp): Show timestamps in integrations_tests

sklgromek:

## Summary
Add timestamps to test_runner output and logs

## Test Plan
-  Creating this PR
-  Check if in `dp-workflow` summary logs saved as artifacts have timestamps

### DIFF
--- a/.github/workflows/dp-workflow.yml
+++ b/.github/workflows/dp-workflow.yml
@@ -412,9 +412,9 @@ jobs:
           for pod in $(kubectl get pods -o custom-columns=NAME:.metadata.name --no-headers);
           do
             if [[ "$pod" == "postgres-database-0" ]]; then
-              kubectl logs $pod --container postgres-test > ${TEST_DIR}/k8s-pods-logs/${pod}.log;
+              kubectl logs --timestamps=true $pod --container postgres-test > ${TEST_DIR}/k8s-pods-logs/${pod}.log;
             else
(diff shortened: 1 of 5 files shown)


dp/cloud/python/magma/test_runner/pytest.ini
```diff
@@ -15,3 +15,5 @@ markers =
     local: integration tests without orc8r
 log_cli = True
 log_level = DEBUG
+log_format = %(asctime)s %(levelname)s %(message)s
```
jkmar: do we need both this and k8s timestamps?

sklgromek: From k8s should be enough

Neudrino (approved):
Seems reasonable.

Was it tested?
Give a test plan in the PR description and execute it :wink:, please.

dp/cloud/python/magma/test_runner/pytest.ini
sklgromek: I reverted this change

WojSad (approved):

dp/cloud/python/magma/test_runner/pytest.ini
jkmar: hmm, I'm actually wondering it would better if we had only that change instead of kubeclt one, since it would only in one place (so it would work for both orc8r tests and without orc8r test) and be consistent with the other components which use their own timestamps instead of relying on k8s

jkmar (approved):

PR #12875, opened 2022-05-30
chore(pipelined): Fix selection of mypy errors

mpfirrmann:
## Summary

This is a follow-up to #12833, focusing on `encoding.py` and further type annotation errors. 
After the PR, the number of `mypy --ignore-missing-imports magma/lte/gateway/python/magma/pipelined` errors is decreased by 22.

## Test Plan
on magma-dev: 

- `bazel test lte/gateway/python/magma/pipelined/...`
- `./bazel/scripts/run_sudo_tests.sh lte/gateway/python/magma/pipelined/`

## Additional Information

- [ ] This change is backwards-breaking

PR #12884, opened 2022-05-31
chore(pipelined): More mypy fixes

voisey:
## Summary

Follow up to https://github.com/magma/magma/pull/12833 and https://github.com/magma/magma/pull/12875.

## Test Plan

- [x] magma-dev: `bazel test lte/gateway/python/magma/pipelined/...`
- [x] magma-dev: `./bazel/scripts/run_sudo_tests.sh lte/gateway/python/magma/pipelined/`

### DIFF
--- a/lte/gateway/python/magma/pipelined/openflow/flows.py
+++ b/lte/gateway/python/magma/pipelined/openflow/flows.py
@@ -654,8 +654,8 @@ def _check_resubmit_action(actions, parser):
 
 
 def send_stats_request(
-    datapath, tbl_num, cookie: hex = 0,
-    cookie_mask: hex = 0, retries: int = 3,
+    datapath, tbl_num, cookie: int = 0,
+    cookie_mask: int = 0, retries: int = 3,
 ):
     """
     Send a stats request msg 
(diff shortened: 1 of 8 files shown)


lte/gateway/python/magma/pipelined/openflow/flows.py
```diff
@@ -654,17 +654,17 @@ def _check_resubmit_action(actions, parser):
...
     """
     Send a stats request msg 
     Args:
         datapath (ryu.controller.controller.Datapath):
             Datapath to query from
         table (int): Table number to query for
-        cookie (hex): cookie value for the request
-        cookie_mask (hex): cookie mask for the request
+        cookie (int): a hex cookie value for the request
+        cookie_mask (int): a hex cookie mask for the request
```
sebathomas, line 667: What does "hex" mean in this context?

voisey: Good question. I meant things like [this](https://github.com/magma/magma/blob/2f85a25c11ad8ce9c03a901bc1c400f2bdc37d23/lte/gateway/python/magma/pipelined/openflow/flows.py#L28) which is input to the function I've modified [here](https://github.com/magma/magma/blob/3c21b360b1e5fc3969074f5c64ffd810b27b0aae/lte/gateway/python/magma/pipelined/tests/test_enforcement_stats.py#L827), for example.

Note that Python seems to consider it to be an integer as I have declared here:
```
>>> OVS_COOKIE_MATCH_ALL = 0xffffffff
>>> type(OVS_COOKIE_MATCH_ALL)
<class 'int'>
```
though the built-in hex function actually returns strings:
```
>>> int(str(OVS_COOKIE_MATCH_ALL), 0)
4294967295
>>> hex(4294967295)
'0xffffffff'
```

sebathomas: All right, so it is just an int for Python. I guess they write it in hexadecimal because the variable is used as a bitmask.

sebathomas (approved):

vktng (approved):
LGTM.

PR #12885, opened 2022-05-31
chore(pipelined): selection of mypy errors are fixed

mpfirrmann:
## Summary

This is a follow-up to several mypy clean-ups, i.e. #12833, #12875, focusing on invalid type comments and incompatible return value types. After the PR, the number of mypy --ignore-missing-imports magma/lte/gateway/python/magma/pipelined errors is decreased by 12.

## Test Plan
on magma-dev: 

- `bazel test lte/gateway/python/magma/pipelined/...`
- `./bazel/scripts/run_sudo_tests.sh lte/gateway/python/magma/pipelined/`

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/gtp_stats_collector.py
+++ b/lte/gateway/python/magma/pipelined/gtp_stats_collector.py
@@ -106,7 +106,7 @@ def _get_ovsdb_dump_params(params: OVSDBDumpCommandParams) -> List[str]:
 
 
 def _parse_ovsdb_dump_output(
-    stdout: str, stderr: str,
+    stdout: bytes, stderr: str,
     _,
 ) -> OVSDBCommandResult:
     """
--- a/lte/gateway/python/magma/pipelined/qos/common.py
+++ b/lte/gateway/python/magma/pipelined/qos/common.py
@@ -15,7 +15,7 @@ import logging
 import threading
 import traceback
 from enum import Enum
-from typing import Any, Dict, List, Set, Tuple
+from typing import Any, Dict, List, Optional, Set, Tuple
 
 from lte.protos.policydb_pb2 import FlowMatch
 from magma.common.redis.client import get_default_client
@@ -107,13 +107,11 @@ class SubscriberState(object):
     def remove_session(self, ip_addr: str) -> None:
         del self.sessions[ip_addr]
 
-    def find_session_with_rule(self, rule_num: int) -> SubscriberSession:
-        session_with_rule = None
+    def find_session_with_rule(self, rule_num: int) -> Optional[SubscriberSession]:
         for session in self.sessions.values():
             if rule_num in session.rules:
-                session_with_rule = session
-                break
-        return session_with_rule
+                return session
+        return None
 
     def _update_rules_map(
         self, ip_addr: str, rule_num: int, d: FlowMatch.Direction,
@@ -148,9 +146,9 @@ class SubscriberState(object):
                 )
                 if get_key_json(k) in self._redis_store:
                     del self._redis_store[get_key_json(k)]
+            session_with_rule.rules.remove(rule_num)
 
         del self.rules[rule_num]
-        session_with_rule.rules.remove(rule_num)
 
     def find_rule(self, rule_num: int):
         return self.rules.get(rule_num)
--- a/lte/gateway/python/magma/pipelined/rpc_servicer.py
+++ b/lte/gateway/python/magma/pipelined/rpc_servicer.py
@@ -15,7 +15,7 @@ import logging
 import os
 import queue
 from concurrent.futures import Future
-from typing import List, OrderedDict, Tuple
+from typing import List, OrderedDict
 
 import grpc
 from lte.protos import pipelined_pb2_grpc
@@ -1023,7 +1023,7 @@ class PipelinedRpcServicer(pipelined_pb2_grpc.PipelinedServicer):
 
     def _ng_qer_update(
         self, request: SessionSet, pdr_entry: PDRRuleEntry,
-    ) -> Tuple[List[RuleModResult], List[RuleModResult]]:
+    ) -> List[RuleModResult]:
         enforcement_res = []
         failed_policy_rules_results: List = []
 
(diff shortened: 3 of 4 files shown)


lte/gateway/python/magma/pipelined/rpc_servicer.py
```diff
@@ -156,7 +156,7 @@ def _get_setup_result(self, controller, context):
             context.set_details('SetupDefaultController processing timed out')
             return SetupFlowsResult(result=SetupFlowsResult.FAILURE)
 
-    def _setup_default_controller(self, controller, fut: 'Future(SetupFlowsResult)'):
+    def _setup_default_controller(self, controller, fut: 'Future[SetupFlowsResult]'):
```
sebathomas: The single quotes are not necessary (here and at other places in this file).

lte/gateway/python/magma/pipelined/gtp_stats_collector.py
```diff
@@ -144,7 +144,7 @@ def match_gtp_lines(lines):
         return create_error_result(stderr)
 
     try:
-        stdout_lines = stdout.decode('ascii').split('\n')
```
sebathomas: Are you sure this can be removed? The method signature claims stdout is a string but did you check that the caller really passes a string?

sebathomas (commented):
Thanks for the cleanup. Just two minor points.

lte/gateway/python/magma/pipelined/rpc_servicer.py
mpfirrmann: Removing the single quotes fails the unit tests and a change to `Awaitable` does not work with `mypy`, idk why tbh.

lte/gateway/python/magma/pipelined/gtp_stats_collector.py
mpfirrmann: Apparently this does not get tested in the unit / sudo tests. How should we proceed? Change `str` to `bytes` instead or keep my first version?

lte/gateway/python/magma/pipelined/rpc_servicer.py
```diff
@@ -961,7 +961,7 @@ def ng_update_session_flows(
                 if ret:
                     # Install the Rules
                     failed_policy_rule_results =\
-                        self._ng_qer_update(request, pdr_entries)
+        self._ng_qer_update(request, pdr_entries)
```
voisey: This looks a bit strange and it seems it's breaking the Python formatting check. Was it intentional?

mpfirrmann: No, this was not intentional. Thank you :+1:

lte/gateway/python/magma/pipelined/gtp_stats_collector.py
sebathomas: Change `str` to `bytes` (probably for both inputs).

Assuming that this code worked before (even if it is untested), the change would break it.

mpfirrmann: The `stderr` is parsed into the `OVSDBCommandResult` NamedTuple, typed there as an `Optional[str]`.

lte/gateway/python/magma/pipelined/rpc_servicer.py
sebathomas: I think you need to do `from __future__ import annotations`, then you can remove the single quotes and use `Future` as type.

lte/gateway/python/magma/pipelined/gtp_stats_collector.py
sebathomas: You're right, that's really strange. So only `stdout` seems to be bytes, `stderr` seems to be a string...

sebathomas: Or maybe it's actually putting a bytes object into `OVSDBCommandResult` and no one noticed. Hard to say without running the code.

lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -108,7 +108,7 @@ def remove_session(self, ip_addr: str) -> None:
         del self.sessions[ip_addr]
 
     def find_session_with_rule(self, rule_num: int) -> SubscriberSession:
-        session_with_rule = None
+        session_with_rule = SubscriberSession()
```
sebathomas: I think it would be better to change the return value to `Optional[SubscriberSession]` here. Returning a new session seems strange.

lte/gateway/python/magma/pipelined/qos/common.py
```diff
@@ -55,7 +55,7 @@ def list():
 
 
 class SubscriberSession(object):
-    def __init__(self, ip_addr: str):
+    def __init__(self, ip_addr: str = ""):
```
sebathomas: Then you might not need this default argument.

lte/gateway/python/magma/pipelined/qos/common.py
sebathomas: I think the caller cannot handle the case where `None` is returned at the moment. You may have to move the last line into the `if` block there.

sebathomas (approved):

vktng (approved):
LGTM.

lte/gateway/python/magma/pipelined/rpc_servicer.py
```diff
@@ -10,12 +10,16 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+# pylint: disable=unsubscriptable-object
```
sebathomas, line 13: Context: This is a bug and will not be needed for future versions of pylint.
https://github.com/PyCQA/pylint/issues/6523

voisey: Could you please create an issue about this so it gets removed when pylint is updated?

sebathomas: The fix in the new version of pylint is just to disable the check in all files that do `from __future__ important annotations`. So the line will have no effect and we won't need it any more but I'm not sure it's really worth tracking that.

PR #11668, opened 2022-02-17
chore(agw): Update metrics naming on exporting 

ardzoht:
## Summary

- Currently, metrics collector on magmad is using a deprecated proto file (`metricsd.proto`) to set the name for each Metric family (collection of metrics on a service)
- This change updates the filling of the metric name to use the name from the prometheus metric name rather from the protobuf description name. 
- Updates s1ap tests to remove `metricsd` proto enum values

## Test Plan

- `make integ_test`
- running local environment and making sure metrics are still collected and exported successfully to the cloud:
```
Feb 08 22:45:33 magma-dev-focal magmad[51255]: DEBUG:root:Collected 9 from state...
Feb 08 22:45:33 magma-dev-focal magmad[51255]: DEBUG:root:Collected 18 from pipelined...
Feb 08 22:45:33 magma-dev-focal magmad[51255]: DEBUG:root:Metrics upload success for service pipelined (chunk 0)
Feb 08 22:45:33 magma-dev-focal magmad[51255]: DEBUG:root:Metrics upload success for service magmad (chunk 0)
Feb 08 22:45:33 magma-dev-focal magmad[51255]: DEBUG:root:Metrics upload success for service sessiond (chunk 0)
Feb 08 22:45:33 magma-dev-focal magmad[51255]: DEBUG:root:Metrics upload success for service state (chunk 0)
Feb 08 22:45:33 magma-dev-focal magmad[51255]: DEBUG:root:Metrics upload success for service enodebd (chunk 0)
```

- `service303_cli.py metrics pipelined` before:
```
name: "50"
type: GAUGE
metric {
  label {
    name: "10"
    value: "3.8.5"
  }
  label {
    name: "11"
    value: "CPython"
  }
  label {
    name: "9"
    value: "3"
  }
  label {
    name: "8"
    value: "8"
  }
  label {
    name: "7"
    value: "5"
  }
  gauge {
    value: 1.0
  }
  timestamp_ms: 1644366222080
}

name: "309"
type: COUNTER

name: "550"
type: COUNTER
metric {
  counter {
    value: 2.0
  }
  timestamp_ms: 1644366222080
}

name: "350"
type: COUNTER

name: "351"
type: COUNTER
metric {
  counter {
    value: 0.0
  }
  timestamp_ms: 1644366222080
}

name: "352"
type: COUNTER

name: "353"
type: COUNTER
metric {
  counter {
    value: 0.0
  }
  timestamp_ms: 1644366222080
}
```
[...]

PR #12906, opened 2022-06-02
chore(orc8r): Cleanup metric naming to not use deprecated protobuf file

### LINKED ISSUE #12660: Deprecate proto enum parsing on cloud metricsd service
### Description
- Metrics on magma have been using a method of parsing names and values of the Prometheus metrics which is hard to maintain. The names need to be updated constantly on a protobuf file and adding or deleting metrics become hard and some of these values are already obsolete.
- With the change on https://github.com/magma/magma/pull/11668, we have removed the parsing of these names using the proto file on the AGW side.
- To fully deprecate the proto parsing process, we need to remove enum parsing on the `metricsd` service on Orchestrator. 

**Magma metrics starting guide**
https://docs.magmacore.org/docs/resources/ref_magma_metrics#life-of-a-magma-metric

**Code pointers**
- https://sourcegraph.com/github.com/magma/magma/-/blob/orc8r/protos/metricsd.proto
- https://sourcegraph.com/github.com/magma/magma/-/tree/orc8r/cloud/go/services/metricsd

**Test your changes**
- Option 1: After removing proto parsing, you can deploy an instance of Magma Orchestrator locally and the AGW VM, register and connect it to the Orchestrator, and perform a query for any metric (e.g. `magmad_ping_rtt_ms`) using PromQL https://prometheus.io/docs/prometheus/latest/querying/basics/ on the Swagger API `Metrics` endpoint.
- Option 2: You can deploy the full Magma stack (AGW + Orc8r + NMS) following the steps on https://docs.magmacore.org/docs/basics/quick_start_guide#using-the-nms-ui and look for any dashboard on the Metrics tab. It should display all metrics successfully.

vroon2703:
## Summary

Replaced the function calls to GetDecodedName with GetName.  
The GetDecodedName function made calls to a GetEnumNameIfPossible function that tried to parse the metricsd.proto file to extract the enum name (metricName) that corresponds to the enum value. 

But with the changes on the AGW side -  (refer https://github.com/magma/magma/pull/11668), the metric name itself is parsed instead of the enum value. This renders the GetEnumNameIfPossible function obsolete. 

## Test Plan

Updated the unittests to check that the metricName is indeed the name that is passed. 

Screen Shot 2022-06-02 at 2 59 14 PM

All unittests for metricsd/servicers pass for both protected and southbound after the change.


orc8r/cloud/go/services/metricsd/servicers/southbound/servicer_southbound_test.go
```diff
@@ -169,7 +170,7 @@ func TestCollect(t *testing.T) {
 	assert.Equal(t, 1, len(e.queue))
 	assert.Equal(t, strconv.FormatFloat(float, 'f', -1, 64), e.queue[0].Value())
 	// check that label protos are converted
-	assert.True(t, tests.HasLabelName(e.queue[0].Labels(), protos.GetEnumNameIfPossible(key, protos.MetricLabelName_name)))
+	// assert.True(t, tests.HasLabelName(e.queue[0].Labels(), protos.GetEnumNameIfPossible(key, protos.MetricLabelName_name)))
```
ssanadhya: Remove commented code and the corresponding commented text

orc8r/lib/go/protos/metricsd_helper.go
```diff
@@ -17,43 +17,18 @@ limitations under the License.
...
-	if err == nil {
-		nameStr = proto.EnumName(protoEnumMapping, int32(enumVal))
-	} else {
-		nameStr = toConvert
-	}
-	return nameStr
-}
 
-// GetDecodedLabel tries to convert the metric label name/value enums to their
-// enum names for display.
```
ssanadhya, line 41: Please update the description with what this function is doing now

uri200 (approved):
LGTM!

ardzoht (approved):
lgtm 👍

PR #12909, opened 2022-06-03
fix(ci): the latest package version coincides with the version apt would choose

### LINKED ISSUE #12894: Analyse and Fix build-all/agw-build/Build AGW
see #12893

It looks like there are discrepancies in the python3-* packages between the magma and fb repo. The current error can be passed (locally) by using
- using python3-astroid_2.4.2_all.deb from the fb artifactory
- using python3-pylint_2.6.0_all.deb from the fb artifactory
- using python3-ryu from the ubuntu repos (i.e., remove it from the magma repo)

agw-build was successful in https://github.com/magma/magma/runs/6707621279?check_suite_focus=true
but the post sentry_release fails now. See attached PR.

nstng:

## Summary

In pydep at one point for a package `python3-<NAME>` it is checked, what is the latest version that could be installed via apt. The latest version is calculated manually via "<"-comparison. This fails e.g., for `python3-webcolors` when it is available in versions `1.5-2.1` and `1.11.1`. The algorithm returns `1.5-2.1` while apt would install `1.11.1` as the latest version.

Here: do not calculate the latest version manually, but use the existing apt.package.candidate.version mechanics (candidate ^= the candidate of a package that would be installed by apt).

## Test Plan

CI - build-all/agw-build/Build AGW
But to be honest, this is hard to reproduce because this relies on certain repository states.

## Additional Information

### DIFF
--- a/lte/gateway/release/pydep
+++ b/lte/gateway/release/pydep
@@ -394,18 +394,14 @@ def get_latest_apt_pkg_version(pkgname: str, py2: bool = False) -> Optional[str]
     """
     cache = apt.Cache()
     sys_pkgname = gen_sys_package_name(pkgname, py2)
+
     try:
-        avail_versions = cache[sys_pkgname].versions
+        package = cache[sys_pkgname]
     except KeyError:
         # package isn't available
         return None
 
-    latest = None
-    for v in avail_versions:
-        if not latest or v.version > latest:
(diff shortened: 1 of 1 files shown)


nstng:
@tmdzk FYI one issue with the build_all workflow.

PR #12917, opened 2022-06-07
fix(agw): Remove unused patch reference on liblfds packet generation

gjalves:
Signed-off-by: Gustavo Junior Alves <gustavo@radtonics.com>

## Summary

Remove unused patch reference on liblfds packet generation.

## Test Plan

Without this patch, trying to build liblfds with liblfds_build.sh script fails to build.


nstng:
Hi, this might me my fault, sorry. See https://github.com/magma/magma/pull/12877 - I just did not see that the patch was applied here. Patch was removed, because it broke arm builds.

nstng (approved):

gjalves:
I´m facing some permission trouble setting the label as ready2merge. Someone can do the favor? Thanks.

PR #12926, opened 2022-06-07
fix(ci): tests are re-run for more stable coverage generation

nstng:

## Summary

TBD

## Additional Information


nstng:
Experiments here did not succeed. Analysis shows that this is because of a Bazel 4 bug where c/c++ codecoverage does not work reliable with remote caches (see https://github.com/bazelbuild/bazel/issues 13193 and 12013).
This will be solved with #12929.

Issue #12818, opened 2022-05-23
sctpd crashes due to "too many open files"


jrobsing:
This is like bug #8898 which was closed as "will not fix".
**Problem summary:** In a lab regression test of a gNB where the gNB is rebooted 1000 times and UEs connect and pass traffic, I observed that some random time into the test the sctpd process would crash because of "too many open files". This caused a service outage from several minutes to over 20 minutes in duration. This problem was observed multiple times on multiple test platforms.
**Debug:** I did an lsof every minute during the test and plotted the number of file handles sctpd was using and found that it increased during the test and only decreased when the application crashed and was restarted by systemd. Also, I did strace on the process and found that it never called close() or shutdown().
**Fix:** I implemented a fix in sctp_connection::Listen() that calls shutdown() and close() on client_sd when the status == DISCONNECT and this appears to have fixed the problem. 1000 cycles ran without a failure. However, I will continue testing.

jrobsing:
I have run 2000 test cycles and there were no failures.

edaspb:
would you like to make pull request?

wssrobs:
I've never done a pull request before. Instead of learning how to do it, I'll just include the 2-line patch here. The file is sctp_connection.cpp at line 155 in function SctpConnection::Listen(). I marked the two new lines with "+":
```
      } else {
        int client_sd = events[i].data.fd;

        auto status = HandleClientSock(client_sd);

        if ((status == SctpStatus::DISCONNECT) ||
            (status == SctpStatus::NEW_ASSOC_NOTIF_FAILED)) {
          if (epoll_ctl(epoll_fd, EPOLL_CTL_DEL, client_sd, nullptr) < 0) {
            MLOG_perror("epoll_ctl");
            std::terminate();
          }
+          shutdown(client_sd, SHUT_RDWR);
+          close(client_sd);
        }
```

MoritzThomasHuebner:
I ran the `test_multi_enb_partial_reset.py` with and without the proposed changes several thousand times on the `magma_test` VM. I was able to reproduce the issue and the proposed changes do indeed fix the issue. I have created a PR #12931.

sebathomas:
Should be fixed by https://github.com/magma/magma/pull/12931.

Thank you @wssrobs for proposing the solution!

PR #12931, opened 2022-06-08
fix(agw): Properly closes client_sd in SCTP Listener

MoritzThomasHuebner:
## Summary

Implements the solution in https://github.com/magma/magma/issues/12818

## Test Plan

- Run unit tests
- Reproduce errors on a local `magma_test` vagrant with the current state of the master branch. This worked, but I did not take a screenshot. The error occurs about once every hour after which the sctpd connection drops and restarts. The restart takes about 15 minutes.
- Run this branch manually on a local `magma_test` vagrant to see if the changes fix the described problem. Grepping for "Too many" to find the error message now yields now result.
image

## Additional Information


sebathomas (approved):

ssanadhya (approved):
Lgtm

PR #12935, opened 2022-06-08
chore: Upgrade to echo v4

jkmar:
## Summary
Upgrade echo library from version 3.3.10 to 4.7.2.
This change makes it possible to use some of the additional v4 only libraries,
for example library for collecting http prometheus metrics.

## Test Plan
- [x] Deployment on AWS
- [x] CI


WojSad (approved):

maxhbr (approved):
LGTM

uri200 (approved):

PR #12937, opened 2022-06-08
chore(feg): The s8_proxy expects ips for the local_address and pgw_ad…

crasu:
The s8_proxy expects ips for the local_address and pgw_address parameters. If hostnames are provided the s8_proxy crashes on startup.

This pr replaces the previous hostnames with ips.


sebathomas (approved):

uri200 (approved):

uri200:
this could also be fixed using `net.LookupIP(ipOrHost)` in here http://github.com/magma/magma/blob/master/feg/gateway/gtp/gtp_client.go#L68-L68

The challenge is that `LookupIP` returns an array of IPs so we would need a function to know which IP to use

PR #12940, opened 2022-06-09
chore(deps): upgrade fluentd in orc8r to 1.14.6

wolfseb:
## Summary

Upgrade fluentd and its dependencies in the orc8r to the latest version
(fluentd 1.14.6) as discussed here #10464.

## Test Plan

Spin up a minikube deployment and see if nothing breaks.
The fluentd setup on orc8r consists of two containers:
(actually, [here in the docs](https://docs.magmacore.org/docs/next/orc8r/dev_security) it says that fluentd is outside the orc8r, but [here it says the opposite](https://docs.magmacore.org/docs/next/lte/eventd). Is one of those outdated?)

1. a fluentd-daemon which forwards all logs in the orc8r to elasticsearch (index pattern: logstash)
2. a fluentd-forward which forwards all logs from gateways to elasticsearch (index pattern: fluentd)

I only tested a setup with just the orchestrator in the minikube on a linux machine. If the forwarding of logs from e.g. the AGW should also be tested please let me know (fluentd-forward sends one log to elasticsearch when starting, so it is clear that it can connect to elasticsearch and send its logs there at least)

To spin up the minikube with a local registry (localhost:5000) with fluentd and connect it to elasticsearch and kibana I took the following steps (I made a separate PR to add this process to the documentation here: #12942):

In addition to the steps in the documentation [to deploy the orc8r on minikube](https://docs.magmacore.org/docs/next/orc8r/dev_minikube), we need to make the following additions to the guide (I only did this on a linux machine, there might be differences on mac):

**Build and publish fluentd images**

The values.yaml file in `orc8r/cloud/helm/orc8r/charts/logging` needs to be changed to create the fluentd images.
For `fluentd_daemon` and `fluentd_forward`, change 

- `create: true`
- `repository: localhost:5000/fluentd_daemon` (and `.../fluentd_forward`, respectively)
- `tag: latest` (or choose the tag you want, it doesn't really matter)
- `elastic_host: "host.minikube.internal"`
The address `host.minikube.internal` points to the host machine from inside the minikube, so that fluentd can forward its logs to elasticsearch, which will be listening to port 9200 on the host machine.

It should look something like shown in the picture (and equivalent for fluentd_forward):
fluentd-daemon_setup_in_values-yaml

To build the new fluentd images, change directory to
`cd $MAGMA_ROOT/orc8r/cloud/docker`
and edit `build.py`: uncomment the line `'docker-compose.logging.yml'` as shown in the picture below
build_fluentd_images
Then run `./build.py --all` to create the images.

Publish the images with the command
```
cd $MAGMA_ROOT/orc8r/cloud/docker
for image in fluentd fluentd-forward fluentd-daemon ; do ${MAGMA_ROOT}/orc8r/tools/docker/publish.sh -r localhost:5000 -i ${image} -v YOUR_TAG -u '' -p '' ; done
```
where you replace `YOUR_TAG` with the tag you chose earlier in `values.yaml`.

**Generate secrets for fluentd**
 generate and apply the secrets for fluentd. The commands to generate these can be found in `orc8r/cloud/deploy/scripts/create_application_certs.sh`, which are the following lines:
```
openssl genrsa -out fluentd.key 2048
openssl req -new -key fluentd.key -out fluentd.csr -subj "/C=US/CN=fluentd.$domain"
openssl x509 -req -in fluentd.csr -CA certifier.pem -CAkey certifier.key -CAcreateserial -out fluentd.pem -days 3650 -sha256
```

**Apply fluentd secrets**
To apply the secrets, add the following two lines to the command in the documentation:
```
--set-file 'secret.certs.files.fluentd\.pem'=${CERTS_DIR}/fluentd.pem \
--set-file 'secret.certs.files.fluentd\.key'=${CERTS_DIR}/fluentd.key \
```

**Start the minikube**
Once you start the minikube and apply your helm charts as written int he documentation, there should be a daemonset `orc8r-fluentd-daemon` and a pod `orc8r-fluentd-daemon` in the namespace kube-system, and a pod `orc8r-fluentd-forward` in the namespace orc8r.

**Start elasticsearch and kibana**
To start the elasticsearch and kibana containers, simply run the commands
```
cd $MAGMA_ROOT/orc8r/cloud/docker
./run.py
```
You should now be able to access kibana from your browser under `localhost:5601` and find the index patterns as described above.

## Additional Information


orc8r/cloud/docker/fluentd_daemon/Gemfile
```diff
@@ -9,9 +9,9 @@ gem "fluent-plugin-prometheus", "~> 1.5.0"
 gem 'fluent-plugin-json-in-json-2', ">= 1.0.2"
 gem "fluent-plugin-record-modifier", "~> 2.0.0"
 gem "fluent-plugin-rewrite-tag-filter", "~> 2.2.0"
-gem "elasticsearch", "~> 7.1"
-gem "fluent-plugin-elasticsearch", "~> 3.5.5"
+gem "elasticsearch", "~> 7.13.0"
+gem "fluent-plugin-elasticsearch", "~> 5.2.1"
```
alexzurbonsen, line 13: This is probably just updating to the versions that are installed in the Dockerfile already?

alexzurbonsen, line 13: This probably just takes the versions that are installed in the Dockerfile already?

orc8r/cloud/docker/fluentd/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fluent/fluentd:v1.7-1
+FROM fluent/fluentd:v1.14.6-debian-1.0
```
alexzurbonsen, line 1: Debian is necessary here due to the `fluent-plugin-systemd` dependency?

orc8r/cloud/docker/fluentd_daemon/Gemfile
wolfseb: Yes. For some reason, the versions for fluentd_daemon were older than for fluentd_forward, not sure if that was intended or not, but I set everything to the same version now. But the Dockerfiles had to be changed at some parts as well. The main point is that the older versions of elasticsearch and the fluent-plugin do not support the newest version of fluentd.

orc8r/cloud/docker/fluentd/Dockerfile
wolfseb: Frankly speaking, it is the only one that worked for me. E.g. it returned an error stating it could not find 1.14-1. (the fluentd_daemon Dockerfile specified Debian before the update as well, the others didn't)

alexzurbonsen (approved):

orc8r/cloud/docker/fluentd/Dockerfile
Neudrino: Weird. The [v1.14-1 image is available at docker-hub](https://hub.docker.com/layers/fluentd/library/fluentd/v1.14-1/images/sha256-9624d5acdaf03449120293a1f5332154dc9c2796fddf4b1f998aeee5aaca7027?context=explore) and I have no issue pulling it.
grafik

Neudrino (approved):
We might want to specify exactly one version in the lock-file instead of ranges in future?

ShubhamTatvamasi (approved):

orc8r/cloud/docker/fluentd/Dockerfile
wolfseb: I tried again, and the problem only occurs when building the fluentd-daemon image. If I change the version to 1.14-1 in `.../fluentd_daemon/Dockerfile`, the build process cancels with this error:
Screenshot from 2022-07-08 09-41-45
If I use the version 1.14.6-debian-1.0 it works fine. The other images are not affected by this, i.e. using the other version in the respective Dockerfiles still works. Maybe this is the reason the debian package was used for the fluentd-daemon image before the update as well?
I changed it for the others because I think it might be better/more consistent to use the same package everywhere.

Neudrino: Thanks for the additional comment. So maybe I misunderstood you previous comment?

__To clarify:__
There is no problem with image availability for `fluentd:v1.14-1`.
The issue is quite clearly stated in your console output, that the `apt-get` command is not available in the `fluentd:v1.14-1` image. But this is expected, as the `fluentd:v1.14-1` image is an [Alpine Linux Image](https://github.com/fluent/fluentd-docker-image/blob/master/v1.14/alpine/Dockerfile#L4), which does not provide `apt-get`.

I.e. the custom installation steps are not designed to be run on Alpine. If this is desired, one would need to port that to Alpine, which would be a separate effort.

__FYI:__
To check for an image OS you could also download it and login to access the OS version.
```bash
HOST $ docker pull fluentd:v1.14-1 # download image
HOST $ docker run --rm -i -t fluentd:v1.14-1 sh # run image and login (be aware that alpine has no bash)
ALPINE $ cat /etc/os-release 
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.13.10
PRETTY_NAME="Alpine Linux v3.13"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://bugs.alpinelinux.org/"
```

wolfseb: Ok, thanks a lot for the clarification and the information!

PR #12942, opened 2022-06-09
docs: fluentd for orc8r minikube deployment documentation

wolfseb:
## Summary

Documentation for using fluentd, elasticsearch, and kibana for the Orchestrator minikube deployment.

### DIFF
--- a/docs/readmes/orc8r/dev_build.md
+++ b/docs/readmes/orc8r/dev_build.md
@@ -68,7 +68,16 @@ Build and publish Orchestrator images
 ```bash
 cd ${MAGMA_ROOT}/orc8r/cloud/docker
 ./build.py --all
-for image in controller nginx ; do ${PUBLISH} -r ${REGISTRY} -i ${image} -v ${MAGMA_TAG} ; done
+for image in controller nginx ; do ${PUBLISH} -r ${REGISTRY} -i ${image} -v ${MAGMA_TAG} -u '' -p '' ; done
+```
+
+Optionally, to build fluentd images for the Orchestrator, uncomment the line `docker-compose.logging.yaml` in `build.py`.
+Build and publish fluentd images
+
+```bash
+cd ${MAGMA_ROOT}/orc8r/cloud/docker
+./build.py --all
+for image in fluentd-forward fluentd-daemon ; do ${PUBLISH} -r ${REGISTRY} -i ${image} -v ${MAGMA_TAG} -u '' -p '' ; done
 ```
 
 Build and publish NMS images
--- a/docs/readmes/orc8r/dev_minikube.md
+++ b/docs/readmes/orc8r/dev_minikube.md
@@ -105,6 +105,28 @@ helm template orc8r charts/secrets \
   kubectl apply -f -
 ```
 
+### Optional: fluentd secrets
+
+To use fluentd on the minikube deployment, create additional fluentd secrets
+
+```bash
+cd ${CERTS_DIR}
+openssl genrsa -out fluentd.key 2048
+openssl req -new -key fluentd.key -out fluentd.csr -subj "/C=US/CN=fluentd.$domain"
+openssl x509 -req -in fluentd.csr -CA certifier.pem -CAkey certifier.key -CAcreateserial -out fluentd.pem -days 3650 -sha256
+```
+
+Apply the secrets
+
+```bash
+cd ${MAGMA_ROOT}/orc8r/cloud/helm/orc8r
+helm template orc8r charts/secrets \
+  --namespace orc8r \
+  --set-file 'secret.certs.files.fluentd\.pem'=${CERTS_DIR}/fluentd.pem \
+  --set-file 'secret.certs.files.fluentd\.key'=${CERTS_DIR}/fluentd.key |
+  kubectl apply -f -
+```
+
 ### Create values file
 
 A minimal values file is at `${MAGMA_ROOT}/orc8r/cloud/helm/orc8r/examples/minikube.values.yaml`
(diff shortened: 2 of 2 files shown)


docs/readmes/orc8r/dev_build.md
```diff
@@ -64,11 +64,18 @@ git checkout -b v1.6 origin/v1.6  # or desired Git target, e.g. master
 ```
 
 Build and publish Orchestrator images
+(Optionally, build and publish fluentd images for the Orchestrator by uncommenting the line `docker-compose.logging.yaml` in `build.py`.)
 
 ```bash
 cd ${MAGMA_ROOT}/orc8r/cloud/docker
 ./build.py --all
-for image in controller nginx ; do ${PUBLISH} -r ${REGISTRY} -i ${image} -v ${MAGMA_TAG} ; done
+for image in controller nginx ; do ${PUBLISH} -r ${REGISTRY} -i ${image} -v ${MAGMA_TAG} -u '' -p '' ; done
```
Neudrino, line 71: What does the additional `-u '' -p ''` do?

docs/readmes/orc8r/dev_build.md
```diff
@@ -64,11 +64,18 @@ git checkout -b v1.6 origin/v1.6  # or desired Git target, e.g. master
 ```
 
 Build and publish Orchestrator images
+(Optionally, build and publish fluentd images for the Orchestrator by uncommenting the line `docker-compose.logging.yaml` in `build.py`.)
```
Neudrino: This seems to be duplicate to line 75 or a comment which refines line 75. Thus it should probably go to line 76, to have all comments on this additional step in one place. It is somewhat confusing otherwise, if we talk in two places about optional steps, when they are essentially one.

docs/readmes/orc8r/dev_minikube.md
```diff
@@ -105,6 +116,13 @@ helm template orc8r charts/secrets \
   kubectl apply -f -
 ```
 
+Include the following two lines in the `helm template` command to apply the fluentd secrets:
+
+```bash
+--set-file 'secret.certs.files.fluentd\.pem'=${CERTS_DIR}/fluentd.pem \
+--set-file 'secret.certs.files.fluentd\.key'=${CERTS_DIR}/fluentd.key \
+```
+
```
Neudrino: If they need to be always added, they should be appended in line 115.
Otherwise the condition under which they should be added must be clearly stated.

docs/readmes/orc8r/dev_minikube.md
```diff
@@ -120,6 +138,26 @@ This section describes how to install based on local charts. However, you can al
...
+    tag: IMAGE_TAG
+    pullPolicy: IfNotPresent
+
+  env:
+    elastic_host: "host.minikube.internal"
+    elastic_port: "9200"
+    elastic_scheme: "http"
+```
+
+and equivalent edits for `fluentd_forward`.
```
Neudrino: ~~It is clearer to _not_ split the files to be edited, i.e. naming one before and one after the edit block. Either name both before (preferred) or both after.~~

Neudrino: This can be confusing. Rephrase to be more clear or five a full example with both blocks.

docs/readmes/orc8r/dev_minikube.md
```diff
@@ -120,6 +138,26 @@ This section describes how to install based on local charts. However, you can al
...
+Again, replace `IMAGE_REGISTRY_URL` with your registry and `IMAGE_TAG` with your tag.
```
Neudrino: `Again` is superfluous.

docs/readmes/orc8r/dev_minikube.md
```diff
@@ -156,6 +194,27 @@ orc8r-user-grafana-6498bb6959-rchx5              1/1     Running   0          2m
 postgresql-0                                     1/1     Running   4          6d23h
 ```
 
+If using fluentd, there should be a pod `orc8r-fluentd-forward` running as well. Additionally,
+
+```bash
+kubectl --namespace kube-system get pods
+```
+
+should show a pod `orc8r-fluentd-daemon` running.
```
Neudrino: Do not split the list of running pods before and after the command. Better give the command to run first and than list all pods you expect.

docs/readmes/orc8r/dev_minikube.md
```diff
@@ -74,6 +74,17 @@ export CERTS_DIR=${MAGMA_ROOT}/.cache/test_certs
 cd ${MAGMA_ROOT}/orc8r/cloud/docker && ./build.py && ./run.py && sleep 30 && docker-compose down && ls -l ${CERTS_DIR} && cd -
 ```
 
+To use fluentd on the minikube deployment, create the fluentd certs with:
```
voisey:
```suggestion
To use fluentd on the minikube deployment, create the fluentd certificates with:
```

docs/readmes/orc8r/dev_build.md
wolfseb: It sets the user and password. In my experience it is necessary, otherwise the command just gives this output but doesn't actually do anything else:
image

docs/readmes/orc8r/dev_minikube.md
wolfseb: Thanks!

docs/readmes/orc8r/dev_minikube.md
wolfseb: Ok, I added the output with the additional pods created in one block

docs/readmes/orc8r/dev_minikube.md
wolfseb: Thanks!

docs/readmes/orc8r/dev_minikube.md
wolfseb: Yeah, you're right. I added the block for fluent_forward.

docs/readmes/orc8r/dev_minikube.md
wolfseb: I moved everything about the fluentd secrets into a separate block, and added a standalone code block instead of referencing where to add lines.

docs/readmes/orc8r/dev_build.md
wolfseb: The reason for splitting it like that was so that `./build.py --all` only needs to be run once, because it takes quite long to complete. But I can see how this might be confusing to read, so I just moved everything about fluentd into one block below.

docs/readmes/orc8r/dev_minikube.md
```diff
@@ -105,6 +105,31 @@ helm template orc8r charts/secrets \
...
 
+### Optional: fluentd secrets
+
+To use fluentd on the minikube deployment, create additional fluentd secrets
+
+```bash
+cd ${CERTS_DIR}
+openssl genrsa -out dp_fluentd.key 2048
+openssl req -new -key dp_fluentd.key -out dp_fluentd.csr -subj "/C=US/CN=dp-fluentd.$domain"
+openssl x509 -req -in dp_fluentd.csr -CA certifier.pem -CAkey certifier.key -CAcreateserial -out dp_fluentd.pem -days 3650 -sha256
```
Neudrino: Created but not applied in lines 125ff. Why?

wolfseb: Sorry, my bad. These are indeed not necessary, I removed them from the documentation.

Neudrino (approved):

voisey (approved):

PR #12948, opened 2022-06-09
chore(ci): Bazel workflows send slack message

vktng:

## Summary

Because more code gets bazelified more regression of bazel build files can happen. In order to react proactively, failures of bazel workflows are sent to #bazel-ci. This way the bazel team can react fast and provide support for bazel related workflows on master.

## Test Plan
Additional commit with intentional bazel failure was created. Messages were sent to #bazel-ci.

## Additional Information

The message is null if the workflow was manually triggered (workflow dispatch event), because the event does not have the property `commits`.

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -213,6 +213,17 @@ jobs:
         run: |
           echo "Available storage:"
           df -h
+      - name: Notify failure to slack
+        if: failure() && github.event_name == 'push'
+        uses: rtCamp/action-slack-notify@v2.2.0
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_BAZEL_CI }}
+          SLACK_TITLE: "C/C++ unit tests with Bazel"
+          SLACK_USERNAME: "agw-workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+          SLACK_FOOTER: ' '
+          MSG_MINIMAL: actions url,commit
 
   li_agent_test:
     needs: path_filter
@@ -369,6 +380,7 @@ jobs:
             cp $C_BUILD/coverage.info $MAGMA_ROOT
       - name: Run coverage with Bazel (COMMON / SESSIOND / SCTPD / LIAGENT / CONNECTIOND)
         if: always()
+        id: bazel-codecoverage
         uses: addnab/docker-run-action@v3
         with:
           image: ${{ env.BAZEL_BASE_IMAGE }}
@@ -420,6 +432,17 @@ jobs:
         run: |
           echo "Available storage:"
           df -h
+      - name: Notify Bazel failure to slack
+        if: steps.bazel-codecoverage.outcome=='failure' && github.event_name == 'push'
+        uses: rtCamp/action-slack-notify@v2.2.0
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_BAZEL_CI }}
+          SLACK_TITLE: "C / C++ code coverage with Bazel"
+          SLACK_USERNAME: "agw-workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+          SLACK_FOOTER: ' '
+          MSG_MINIMAL: actions url,commit
 
   lint-clang-format:
     needs: path_filter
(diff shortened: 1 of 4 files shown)


Neudrino (commented):
- There is an inconsistency regarding the positioning of the `uses` statement.
I would prefer to have it fully the same for all additions, e.g. always have the `uses` before the `env` block.

- While I believe and trust that the implementation is always the same, it seams as if this is not consistently working for all notifications. Some do display a title and icon some do not.
  Do you  have any idea why that is the case?
  
grafik

.github/workflows/bazel-cache-push.yml
```diff
@@ -82,6 +82,16 @@ jobs:
 
           aws s3 cp ${{ env.BAZEL_CACHE_MAGMA_VM_TAR }} ${{ env.S3_BUCKET_PATH }}/${{ env.BAZEL_CACHE_MAGMA_VM_TAR}}
           aws s3 cp ${{ env.BAZEL_CACHE_REPO_MAGMA_VM_TAR }} ${{ env.S3_BUCKET_PATH }}/${{ env.BAZEL_CACHE_REPO_MAGMA_VM_TAR}}
+      - name: Notify failure to slack
+        if: failure()
+        uses: rtCamp/action-slack-notify@v2.2.0
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_BAZEL_CI }}
+          SLACK_TITLE: "VM Caches"
+          SLACK_USERNAME: "Push Bazel Cache To S3"
```
Neudrino, line 91: There seems to be information duplication in those two entries. I would not see a reason, why the title and the username should be changed to something like a title.

Neudrino (commented):
To me it seems as if `Ref` and `Event` do not provide sensible information in the slack message. We might consider removing them by:
```yaml
MSG_MINIMAL: actions,commit
```

vktng:
> * There is an inconsistency regarding the positioning of the `uses` statement.
>       I would prefer to have it fully the same for all additions, e.g. always have the `uses` before the `env` block.
> 
>     * While I believe and trust that the implementation is always the same, it seams as if this is not consistently working for all notifications. Some do display a title and icon some do not.
>       Do you  have any idea why that is the case?
> 
> 
> grafik

The title and icon is displayed once for every workflow run. If more than one job fails during one run then only the first will have them.

.github/workflows/bazel-cache-push.yml
vktng: The username should be the same for all entries in the same workflow so messages will be grouped by Slack. (The name of the workflow makes sense.) By the title the relevant jobs or steps are separated. In this case the job does not have a proper name, but "VM caches" describes it sufficiently.

vktng: > To me it seems as if `Ref` and `Event` do not provide sensible information in the slack message. We might consider removing them by:
> 
> ```yaml
> MSG_MINIMAL: actions,commit
> ```

Done.

.github/workflows/agw-workflow.yml
```diff
@@ -213,6 +213,17 @@ jobs:
...
+        if: failure() && github.event_name == 'push'
+        uses: rtCamp/action-slack-notify@v2.2.0
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_BAZEL_CI }}
+          SLACK_TITLE: "C/C++ unit tests with Bazel"
+          SLACK_USERNAME: "agw-workflow"
+          SLACK_ICON_EMOJI: ":boom:"
+          SLACK_COLOR: "#FF0000"
+          SLACK_FOOTER: ' '
+          MSG_MINIMAL: actions url,commit
```
Neudrino, line 226: Should be comma separated, I guess.

vktng: `actions url` is one item. I tried to do it the way you commented, but ` url` is needed too. You can check the difference in #bazel-ci.

Neudrino: Thanks for the clarification. I have misread the documentation and was confused by a line break there.

Neudrino (approved):

.github/workflows/agw-workflow.yml
```diff
@@ -213,6 +213,17 @@ jobs:
         run: |
           echo "Available storage:"
           df -h
+      - name: Notify failure to slack
```
nstng, line 216: Should a notification also be added to `Run coverage with Bazel (COMMON / SESSIOND / SCTPD / LIAGENT / CONNECTIOND)`?

vktng: We can add the functionality to that job too, but it already has a "Notify failure to slack" job (that is restricted to failure in step `c-cpp-codecov-upload`). What would be the best?:
1) Just name it something different (e.g.: "Notify failure to slack (#bazel-ci)"
2) Rename the other job too

nstng: discussed offline

nstng (approved):
lgtm

PR #12950, opened 2022-06-09
chore(pipelined): Make mypy green

sebathomas:
## Summary

Fix various type issues for pipelined.
- Correct a few wrong argument or variable types
- Add type stubs for mixins
- Ignore a false positive
- Raise ValueError instead of NPE
- Extract method to handle all types of error in tc_ops_pyroute2
- Add null checks in test code

## Test Plan

```
cd $MAGMA_ROOT/lte/gateway
mypy --ignore-missing-imports python/magma/pipelined/
```
```
Success: no issues found in 131 source files
```

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/pipelined/app/BUILD.bazel
+++ b/lte/gateway/python/magma/pipelined/app/BUILD.bazel
@@ -204,6 +204,7 @@ py_library(
 py_library(
     name = "restart_mixin",
     srcs = ["restart_mixin.py"],
+    deps = [":startup_flows"],
 )
 
 py_library(
--- a/lte/gateway/python/magma/pipelined/app/restart_mixin.py
+++ b/lte/gateway/python/magma/pipelined/app/restart_mixin.py
@@ -10,13 +10,20 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+from __future__ import annotations
+
 from abc import ABCMeta, abstractmethod
-from typing import Dict, List
+from asyncio import Future
+from logging import Logger
+from typing import Callable, Dict, List, Optional
 
 from lte.protos.pipelined_pb2 import SetupFlowsResult
 from magma.pipelined.app.base import ControllerNotReadyException
+from magma.pipelined.app.startup_flows import StartupFlows
 from magma.pipelined.openflow import flows
+from magma.pipelined.openflow.messages import MessageHub
 from magma.pipelined.policy_converters import ovs_flow_match_to_magma_match
+from ryu.controller.controller import Datapath
 from ryu.ofproto.ofproto_v1_4_parser import OFPFlowStats
 
 DefaultMsgsMap = Dict[int, List[OFPFlowStats]]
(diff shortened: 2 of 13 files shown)


lte/gateway/python/magma/pipelined/encoding.py
```diff
@@ -137,7 +137,7 @@ def get_hash(s, hash_function) -> bytes:
 
 def encode_str(s: str, encoding_type) -> str:
     if encoding_type == PipelineD.HEConfig.BASE64:
-        s = codecs.encode(codecs.decode(s, 'hex'), 'base64').decode()
+        s = codecs.encode(codecs.decode(s, 'hex'), 'base64').decode()  # type: ignore
```
sebathomas, line 140: This seems to be an issue in mypy, see: https://github.com/python/typeshed/issues/300

lte/gateway/python/magma/pipelined/app/BUILD.bazel
```diff
@@ -204,6 +204,9 @@ py_library(
 py_library(
     name = "restart_mixin",
     srcs = ["restart_mixin.py"],
+    deps = [
```
nstng: NIT for one liner lists we usually do not indent, i.e., `deps = [":startup_flows"],`

nstng: Curious, why was this not needed before? I see that you added the import to `restart_mixin`, but the `_startup_flow_controller` was there before, right?

LKreutzer: The import `from magma.pipelined.app.startup_flows import StartupFlows` was newly added

nstng: > I see that you added the import to `restart_mixin`

My question was, why the import was not needed before.

LKreutzer: To me it looks like both lines are new?

LKreutzer: Current master does this: Screenshot from 2022-06-10 13-05-01

nstng: > but the `_startup_flow_controller` was there before

yes, but the object that contains the `StartupFlows` was existing before

sebathomas: Yes but it wasn't typed here before so the Python code didn't know what it was in the abstract superclass.

nstng (approved):
lgtm (feel free to ignore the nit if you do not plan to do other changes). sudo tests locally green.

lte/gateway/python/magma/pipelined/ebpf/ebpf_manager.py
```diff
@@ -11,6 +11,8 @@
 limitations under the License.
 """
 
+from __future__ import annotations
```
voisey, line 14: What effect does this have here?

sebathomas: In this case, it allows you to use `Array[ctypes.c_ubyte]` as a type. In general, it makes type hints behave in this file like they will in Python 3.10:
https://peps.python.org/pep-0563/

voisey: Makes sense, thanks!

lte/gateway/python/magma/pipelined/app/restart_mixin.py
```diff
@@ -28,6 +35,17 @@ class RestartMixin(metaclass=ABCMeta):
...
+    logger: Logger
+    tbl_num: int
+    cleanup_state: Callable
+    delete_all_flows: Callable
+    finish_init: Callable
+    _msg_hub: MessageHub
+    _datapath: Datapath
+    _startup_flow_controller: Optional[StartupFlows]
+    _startup_flows_fut: Future[StartupFlows]
+    _clean_restart: Callable
```
voisey: Is `_clean_restart` definitely callable? It seems to me that it usually gets set to the `clean_restart` part of the config, which is a boolean.

sebathomas: Good catch, that's an error.

(And it's bad that mypy didn't notice...)

voisey (approved):

PR #12956, opened 2022-06-10
chore(mobilityd): Make mypy green

mpfirrmann:
## Summary

Fix mypy issues for mobilityd.

- Argument has incompatible type
- Incompatible types in assigment

## Test Plan
tested with mypy 0.960

```
cd $MAGMA_ROOT/lte/gateway
mypy --ignore-missing-imports python/magma/mobilityd/
```
```
Success: no issues found in 31 source files
```
Unit tests: 
```
vagrant@magma-dev-focal:~/magma$ bazel test lte/gateway/python/magma/...
```
```
Executed 60 out of 60 tests: 60 tests pass.
```

## Additional Information

### DIFF
--- a/lte/gateway/python/magma/mobilityd/dhcp_client.py
+++ b/lte/gateway/python/magma/mobilityd/dhcp_client.py
@@ -16,7 +16,7 @@ import datetime
 import logging
 import threading
 import time
-from ipaddress import IPv4Network, ip_address
+from ipaddress import IPv4Network
 from threading import Condition
 from typing import MutableMapping, Optional
 
@@ -278,7 +278,7 @@ class DHCPClient:
 
                 dhcp_router_opt = self._get_option(packet, "router")
                 if dhcp_router_opt is not None:
-                    router_ip_addr = ip_address(dhcp_router_opt)
+                    router_ip_addr = dhcp_router_opt
                 else:
                     # use DHCP as upstream router in case of missing Open 3.
                     router_ip_addr = dhcp_server_ip
--- a/lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
+++ b/lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
@@ -132,7 +132,7 @@ class IPAllocatorDHCP(IPAllocator):
         if not dhcp_desc or not dhcp_allocated_ip(dhcp_desc):
             dhcp_desc = self._alloc_ip_address_from_dhcp(mac, vlan_id)
 
-        if dhcp_desc and dhcp_allocated_ip(dhcp_desc):
+        if dhcp_desc and dhcp_desc.subnet and dhcp_allocated_ip(dhcp_desc):
             ip_block = ip_network(dhcp_desc.subnet)
             ip_desc = IPDesc(
                 ip=ip_address(dhcp_desc.ip),
@@ -146,9 +146,7 @@ class IPAllocatorDHCP(IPAllocator):
 
             return ip_desc
         else:
-            msg = "No available IP addresses From DHCP for SID: {} MAC {}".format(
-                sid, mac,
-            )
+            msg = f"No available IP addresses From DHCP for SID: {sid} MAC {mac}"
             raise NoAvailableIPError(msg)
 
     def release_ip(self, ip_desc: IPDesc):
--- a/lte/gateway/python/magma/mobilityd/subscriberdb_client.py
+++ b/lte/gateway/python/magma/mobilityd/subscriberdb_client.py
@@ -28,7 +28,7 @@ class NetworkInfo:
     ):
         gw_ip_parsed = None
         try:
-            gw_ip_parsed = ipaddress.ip_address(gw_ip)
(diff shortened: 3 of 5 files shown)


lte/gateway/python/magma/mobilityd/dhcp_client.py
```diff
@@ -278,10 +278,10 @@ def _process_dhcp_pkt(self, packet, state: DHCPState):
 
                 dhcp_router_opt = self._get_option(packet, "router")
                 if dhcp_router_opt is not None:
-                    router_ip_addr = ip_address(dhcp_router_opt)
+                    router_ip_addr = str(ip_address(dhcp_router_opt))
```
sebathomas: This makes little sense. Why convert from (presumably) str to ip back to str?

lte/gateway/python/magma/mobilityd/dhcp_client.py
```diff
@@ -278,10 +278,10 @@ def _process_dhcp_pkt(self, packet, state: DHCPState):
...
                 else:
                     # use DHCP as upstream router in case of missing Open 3.
-                    router_ip_addr = dhcp_server_ip
+                    router_ip_addr = str(dhcp_server_ip)
```
sebathomas: What is dhcp_server_ip?

lte/gateway/python/magma/mobilityd/ip_allocator_dhcp.py
```diff
@@ -133,7 +133,7 @@ def alloc_ip_address(self, sid: str, vlan_id: int) -> IPDesc:
             dhcp_desc = self._alloc_ip_address_from_dhcp(mac, vlan_id)
 
         if dhcp_desc and dhcp_allocated_ip(dhcp_desc):
```
sebathomas:
```suggestion
        if dhcp_desc and dhcp_desc.subnet and dhcp_allocated_ip(dhcp_desc):
```
Mypy is complaining that the input may be `None`. Converting to a string doesn't actually help in that case, it will also fail for the string "None".

lte/gateway/python/magma/mobilityd/subscriberdb_client.py
```diff
@@ -28,7 +28,7 @@ def __init__(
     ):
         gw_ip_parsed = None
         try:
-            gw_ip_parsed = ipaddress.ip_address(gw_ip)
+            gw_ip_parsed = ipaddress.ip_address(str(gw_ip))
```
sebathomas: Same here, but we actually catch the ValueError that happens if the IP is `None`. I think you can just add a `type: ignore`.

lte/gateway/python/magma/mobilityd/uplink_gw.py
```diff
@@ -132,7 +132,7 @@ def update_ip(
             vlan_id: vlan of the GW, None in case of no vlan used.
```
sebathomas, line 132: Same for the changes in this file. We catch the error so we can add a `type: ignore`.

sebathomas (changes requested):

lte/gateway/python/magma/mobilityd/dhcp_client.py
mpfirrmann: Both `dhcp_router_opt` and `dhcp_server_ip` come from scapy's DHCP.options and are indeed `str` (https://github.com/secdev/scapy/blob/b3726ca2bf2b0e7d4999c6165807a7a314bf962f/scapy/layers/dhcp.py#L207).
As the `router_ip_addr` is `str` in `self.dhcp_gw_info.update_ip` and `DHCPDescriptor` any convertion is not needed.

lte/gateway/python/magma/mobilityd/dhcp_client.py
mpfirrmann: See above.

lte/gateway/python/magma/mobilityd/dhcp_client.py
```diff
@@ -132,11 +132,7 @@ def send_dhcp_packet(
             pkt_xid = self._msg_xid
             ciaddr = dhcp_desc.ip
         else:
-            LOG.warning(
-                "Unknown egress request mac %s state %s",
-                str(mac),
-                state,
-            )
+            LOG.warning(f"Unknown egress request mac {str(mac)} state {state}")
```
sebathomas: For logging we actually enforce the other syntax because it is lazy (see the failing linter test).

lte/gateway/python/magma/mobilityd/uplink_gw.py
```diff
@@ -177,17 +177,16 @@ def update_mac(
             mac: mac address of the GW.
         """
         try:
-            ip_addr = ipaddress.ip_address(ip)
+            ip_addr = ipaddress.ip_address(str(ip))
```
sebathomas: This should also be an `ignore` instead I think.

lte/gateway/python/magma/mobilityd/dhcp_client.py
sebathomas: Thanks for looking into that! :+1:

sebathomas (approved):

PR #12963, opened 2022-06-10
chore(ci): Run the LTE integ tests on bazel-built services

LKreutzer:

## Summary

- Create LTE integ test workflow that runs the integ tests on the bazel-built services.
  - Create service files for all services that start the bazel-built executables.
    - This is done by copying the standard service files and changing the ExecStart command.
    - The services that use the template service file get new service files derived from the template. 
  - Service files are copied at runtime into the `/etc/systemd/system/` folder.

## Test Plan

Manually run the workflow.
- [#1 Run on fork, with log level "warning".](https://github.com/LKreutzer/magma/actions/runs/2475783198)
- [#2 Run on fork, with log level "warning".](https://github.com/LKreutzer/magma/actions/runs/2475079087)
- [#3 Run on fork, with log level "warning".](https://github.com/LKreutzer/magma/actions/runs/2488809869)
- [#4 Run on fork, with log level "warning".](https://github.com/LKreutzer/magma/actions/runs/2488821759)
- ( [Run on fork, with log level "fatal".](https://github.com/LKreutzer/magma/actions/runs/2475079087))

## Additional Information

- Note: The log level of bazel needs to be limited in order to not block the SSH connection of fabric. Here this is done with `--ui_event_filters=warning` (see https://github.com/bazelbuild/bazel/issues/4867).
  - Higher log levels lead to much longer run times of the workflow.
  - A better solution to this issue will be investigated. This will also benefit the other parts of the workflow and possibly the standard LTE workflow. E.g.:
    - Do not use fabric
    - Suppress all bazel logging and upload `/tmp/bazel/command.log` afterwards
    - Reconfigure fabric to allow more traffic through SSH if possible (?)
    - ...
- Note: We do not use a GH matrix workflow here in order to keep the complexity low, to not disrupt the result reporting from the standard LTE integ test workflow and because we expect further modifications of this workflow that may be incompatible with the standard workflow.
- Note: The `/etc/gai.conf` file is temporarily - for the bazel build - modified in order to prefer IPv4 connections which speeds up the bazel build.

### DIFF
--- a/.github/workflows/lte-integ-test-bazel.yml
+++ b/.github/workflows/lte-integ-test-bazel.yml
@@ -1,0 +1,85 @@
+---
+name: LTE integ test bazel
+
+on:  # yamllint disable-line rule:truthy
+  workflow_dispatch:
+  workflow_run:
+    workflows:
+      - build-all
+    branches:
+      - master
+      - 'v1.*'
+    types:
+      - completed
+
+jobs:
+  lte-integ-test-bazel:
+    if: github.repository_owner == 'magma' || github.event_name == 'workflow_dispatch'
+    runs-on: macos-10.15
+    env:
+      SHA: ${{ github.event.workflow_run.head_commit.id || github.sha }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ env.SHA }}
+      - name: setup pyenv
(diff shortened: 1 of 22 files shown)


LKreutzer:
Replaced by improved version https://github.com/magma/magma/pull/13279

Issue #12907, opened 2022-06-02
Security: Subscriber Keys Stored in Memory on Host #84 


lucasgonze:
# Type of issue 

Data Exposure

# Step-by-step instructions to reproduce the issue

1. Gain local root access to the Linux host
2. Access subscriber private keys leading to complete compromise of the subscribers in the database. 

# Any special configuration required to reproduce the issue

Local root access to the Linux host

# Impact of the issue, including how an attacker might exploit the issue

Private keys of subscribers, if disclosed, can be used to clone USIMs and/or decrypt user traffic.

# Component

LTE

Location: The subscriberdb element running on the EPC host.

# Severity [*](https://github.com/magma/security/blob/main/Scoring%20rubric%20for%20Magma%20weaknesses.xlsx)

Risk: high
- Impact: high
- Exploitability: medium

# Description

In its current form, subscriberdb running on the Magma Linux host is updated from the cloud every minute with the actual subscriber private keys, which are subsequently stored in an in-memory SQLite database. An attacker that can gain local root access to the Linux host can access these keys leading to complete compromise of the subscribers in the database. 

# Recommendation 

The future migration path plans to host the HSS only in the cloud, with authentication vectors (AVs) generated and forwarded to the local MME in the Magma gateway device. Considering the importance of subscriber private keys in the LTE system, these keys should never be exposed in plaintext on the gateway. Only derived keys should ever be forwarded to the gateway.

# Source

This is also tracked as https://github.com/magma/security/issues/84, which is identical.

This issue was identified by NCC Group in their December 2017 review of security vulnerabilities. Because it is roughly four years old, it may no longer exist, and it is unlikely to be urgent.

<img width="897" alt="image" src="https://user-images.githubusercontent.com/111581/171748358-d64a81c4-e61c-4bfb-a259-48ea98f45beb.png">

lucasgonze:
Close when https://github.com/magma/magma/issues/13249 is complete and merged. 

But first discuss this with the TSC: "Default is to always stream to AGW, that is to keep the existing setup. We can revisit that once Varun has made all the changes. If the TSC agrees, making the default switch is very straightforward."

lucasgonze:
> "Default is to always stream to AGW, that is to keep the existing setup. We can revisit that once Varun has made all the changes. If the TSC agrees, making the default switch is very straightforward."

Maxim comment in Slack:

[...]

PR #12964, opened 2022-06-13
chore(mme): static init is refactored

nstng:
## Summary

This is a preparation for building mme oai with statically linked libraries. This is needed for building a hermetically-ish mme artifact with bazel.

**Problem (in a build with statically linked libraries)**
* In a make build `protobuf` and `glog` are linked dynamically.
* In `S6AClient` are static variables that are initialized by using `protobuf` functionality (json to message mapping). This does not work with statically linked libraries. Furthermore, in the error handling `glog` is used which then also fails with dynamic linking.

**Solution here**
* There are two variables that are initialized with `protobuf` calls.
* The variables are refactored to do the `protobuf` calls after the initialization phase when the result is actually needed (but the results are cached as before).

Best practice suggestions to solve this more elegant are appreciated.

Note: disregard the branch name, as in a first approach one of the variables was removed as dead code.

## Test Plan

CI

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/lib/s6a_proxy/S6aClient.cpp
+++ b/lte/gateway/c/core/oai/lib/s6a_proxy/S6aClient.cpp
@@ -42,14 +42,9 @@ using grpc::Status;
 namespace magma {
 using namespace feg;
 
-static bool read_hss_relay_enabled(void);
-
-static bool read_mme_cloud_subscriberdb_enabled(void);
-
-static const bool hss_relay_enabled = read_hss_relay_enabled();
-
-static const bool cloud_subscriberdb_enabled =
-    read_mme_cloud_subscriberdb_enabled();
+// initialize with -1: not yet calculated/cached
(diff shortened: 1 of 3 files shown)


ssanadhya:
@nstng , the dead code identified in this PR will be active soon, since @vroon2703 is working on https://github.com/magma/magma/issues/12907 . Could you please revise this PR to only include the protobuf initialization fix?

ssanadhya (changes requested):
@nstng , the dead code identified in this PR will be active soon, since @vroon2703 is working on https://github.com/magma/magma/issues/12907 . Could you please revise this PR to only include the protobuf initialization fix?

nstng:
> @nstng , the dead code identified in this PR will be active soon, since @vroon2703 is working on #12907 . Could you please revise this PR to only include the protobuf initialization fix?

Sure, changes applied.

ssanadhya (approved):
Lgtm

Issue #11190, opened 2022-01-18
[agw] Replace hashtable OAI lib to a standard library through map.h interface


ardzoht:
### Description

- OAI makes use of its own internal library for hashtable structures. This is a wide-used library being used throughout all MME tasks mostly for the maintenance of both task’s and UE state. There are multiple types available to use, but mostly both uint64 => uint64 and uint64 => object PTR thread-safe mappings are the most used in the codebase.
- This library has shown evidence of being prone to instability across compiler versions, due to its complex memory management with value and key pointer objects: e.g. https://github.com/magma/magma/pull/11865

### Goal
- We want to iteratively migrate OAI’s hashtable and their usage to protobuf's generated object CPP hashmap structs, a starting point to refer can be: `/lte/gateway/c/core/oai/tasks/mme_app/mme_app_ip_imsi.cpp` or `lte/gateway/c/core/oai/include/map.h`

1. Replace hashtable lib usages in S1AP, NGAP task procedures to `magma::map_s`
2.  Replace hashtable lib usages in SPGW task procedures to `magma::map_s`
3. Replace hashtable lib usages in MME APP task procedures to `magma::map_s`
4. Replace hashtable lib usages in SGW_S8 task procedures to `magma::map_s`

ssanadhya:
@rsarwad , is there any hash table managed by the SGW_S8 task? If yes, please add that to the list of goals for this issue.

rsarwad:
Yes @ssanadhya, SGW_S8 task also includes hash tables. The estimation includes sgw_s8 task

shahrahman-fb:
Closing this issue as part of a repository cleanup effort. This issue has had no activity for over 2 years.

If this issue is still relevant, please reopen it or file a new issue with updated information. We are actively maintaining the project again and want to focus on current, actionable issues.

See: #15833 (Phase 1 cleanup), #15828 (tracking issue)

PR #12965, opened 2022-06-13
fix(agw): Migrated ue_id_coll hashlist to protobuf map

rsarwad:
## Summary
Migrated ue_id_coll hashlist to protobuf map. The PR partially addresses the issue, https://github.com/magma/magma/issues/11190

All tasks are initialized from oai_mme.c file in main function. As we port the individual tasks to cpp, initialize functions are called from main function. And declaration of initialization function of all tasks shall we moved to include/mme_init.hpp file
## Test Plan


lte/gateway/c/core/oai/include/mme_init.hpp
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright 2022 The Magma Authors.
```
ssanadhya, line 2: What's the purpose of creating this new file? 

Further, this is initializing the S1AP task, not the MME task. So the name is misleading.

rsarwad: 1. All tasks are initialized from oai_mme.c file in main function. 
2. As we port the individual tasks to cpp, initialize functions are called from main function. And all these initialize shall be put in this file.
3. Since we started with s1ap task, right now only s1ap task's functionalities are available

lte/gateway/c/core/oai/tasks/s1ap/s1ap_state_converter.cpp
```diff
@@ -145,21 +146,17 @@ void S1apStateConverter::proto_to_enb(const oai::EnbDescription& proto,
...
 
   // load ues
-  hashtable_rc_t ht_rc;
-  auto ht_name = bfromcstr("s1ap_ue_coll");
-
-  hashtable_uint64_ts_init(&enb->ue_id_coll, mme_config.max_ues, nullptr,
-                           ht_name);
-  bdestroy(ht_name);
+  proto_map_rc_t rc = {PROTO_MAP_OK};
+  enb->ue_id_coll.map = new google::protobuf::Map<uint32_t, uint64_t>();
```
ssanadhya, line 150: The proto map is using a uint64 key. Can we use the same width here?

lte/gateway/c/core/oai/tasks/s1ap/s1ap_state_converter.cpp
```diff
@@ -122,7 +122,8 @@ void S1apStateConverter::enb_to_proto(enb_description_t* enb,
   proto->set_ran_cp_ipaddr_sz(enb->ran_cp_ipaddr_sz);
 
   // store ue_ids
-  hashtable_uint64_ts_to_proto(&enb->ue_id_coll, proto->mutable_ue_ids());
+  proto_map_uint32_uint64_to_proto(enb->ue_id_coll, proto->mutable_ue_ids());
```
ssanadhya: Can we just pass the enb->ue_id_coll reference to initialize the proto field? Eventually, this conversion should not be needed as the enb->ue_id_coll should be directly serialized to S1ap state.

ssanadhya (changes requested):

lte/gateway/c/core/oai/include/mme_init.hpp
ssanadhya: Can you add this to the PR description and also the parent issue to guide the migration strategy ? Also, is the goal to eventually get rid of this file?

rsarwad: Updated the comment. Planning to retain this file as declaration of initialization function of all tasks shall be part of this file.

lte/gateway/c/core/oai/tasks/s1ap/s1ap_state_converter.cpp
rsarwad: We discussed that, since mme_ue_s1ap_if type uint32 and shall change the datatype in proto file.

lte/gateway/c/core/oai/tasks/grpc_service/S1apServiceImpl.cpp
```diff
@@ -14,14 +14,14 @@
  * For more information about the OpenAirInterface (OAI) Software Alliance:
  *      contact@openairinterface.org
  */
-#include "lte/gateway/c/core/oai/tasks/grpc_service/S1apServiceImpl.hpp"
 
 #include <string>
+#include "lte/gateway/c/core/oai/tasks/grpc_service/S1apServiceImpl.hpp"
```
ajahl, line 18: Why did you change the order? The include of the own interface should be on the first position.

nstng (approved):
bazel changes lgtm (only BUILD.bazel change)

lte/protos/oai/s1ap_state.proto
```diff
@@ -34,7 +34,7 @@ message EnbDescription {
   uint32 instreams = 8;        // sctp_stream_id_t
   uint32 outstreams = 9;      // sctp_stream_id_t
 
-  map<uint64, uint64> ue_ids = 10; // mme_ue_s1ap_id -> comp_s1ap_id
```
ssanadhya, line 37: You cannot modify the proto field since it will not be backwards compatible. We need to mark this as deprecated, and add a new field 
``` 
map<uint32, uint64> ue_id_map = 14;
```

ssanadhya (changes requested):
This version is still using `proto_map_uint32_uint64_to_proto` to copy the in-memory object to the proto generated object. Do we still need it after changing `s1ap_state.proto`?

lte/protos/oai/s1ap_state.proto
rsarwad: Sure.
Is this the way we mark unused field as deprecated, https://github.com/magma/magma/blob/master/lte/protos/oai/s1ap_state.proto#L8?

rsarwad:
> This version is still using `proto_map_uint32_uint64_to_proto` to copy the in-memory object to the proto generated object. Do we still need it after changing `s1ap_state.proto`?
You are right. We will not be using proto_map_uint32_uint64_to_proto to copy the in-memory object to the proto generated object.  My mistake, I had changes in another branch but forgot take those changes here.
Thanks for letting me know.

lte/gateway/c/core/oai/include/s1ap_types.hpp
```diff
@@ -164,8 +165,9 @@ typedef struct enb_description_s {
   /** UE list for this eNB **/
   /*@{*/
   uint32_t nb_ue_associated;  ///< Number of NAS associated UE on this eNB
-  hash_table_uint64_ts_t ue_id_coll;  ///< Contains comp_s1ap_id assoc to
-                                      ///< enodeb, key is mme_ue_s1ap_id;
+  magma::proto_map_uint32_uint64_t
+      ue_id_coll;  ///< Contains comp_s1ap_id assoc to
```
ssanadhya: The comment here can be simplified to

```
/// key: mme_ue_s1ap_id, value: comp_s1ap_id
```

lte/gateway/c/core/oai/include/mme_init.hpp
ssanadhya: But the init function for each task should be part of that task directory, instead of being in a common header. It will be good to retain this under s1ap_mme.hpp.

lte/gateway/c/core/oai/tasks/s1ap/s1ap_mme_handlers.cpp
```diff
@@ -560,22 +558,22 @@ status_code_e s1ap_mme_handle_s1_setup_request(s1ap_state_t* state,
...
+        enb_association->nb_ue_associated, enb_association->ue_id_coll.size());
     rc = s1ap_mme_generate_s1_setup_failure(
         assoc_id, S1ap_Cause_PR_transport,
         S1ap_CauseTransport_transport_resource_unavailable,
         S1ap_TimeToWait_v20s);
     increment_counter("s1_setup", 1, 2, "result", "failure", "cause",
                       "invalid_state");
     /* UE state at s1ap task is created on reception of initial ue message
-     * Hash list, ue_id_coll is updated after mme_app_task assigns and provides
+     * map, ue_id_coll is updated after mme_app_task assigns and provides
```
ssanadhya:
```suggestion
     * The map ue_id_coll is updated after mme_app_task assigns and provides
```

lte/gateway/c/core/oai/tasks/s1ap/s1ap_mme_handlers.cpp
```diff
@@ -3704,12 +3699,12 @@ status_code_e s1ap_handle_sctp_disconnection(s1ap_state_t* state,
 
   if (reset) {
     /* UE state at s1ap task is created on reception of initial ue message
-     * Hash list, ue_id_coll is updated after mme_app_task assigns and provides
+     * map, ue_id_coll is updated after mme_app_task assigns and provides
```
ssanadhya:
```suggestion
     * The map ue_id_coll is updated after mme_app_task assigns and provides
```

rsarwad:
> proto_map_uint32_uint64_to_proto



> This version is still using `proto_map_uint32_uint64_to_proto` to copy the in-memory object to the proto generated object. Do we still need it after changing `s1ap_state.proto`?

Incorporated review comment.

ssanadhya (commented):
Minor comments, will continue reviewing

lte/gateway/c/core/oai/tasks/s1ap/s1ap_mme_handlers.cpp
```diff
@@ -3704,12 +3699,12 @@ status_code_e s1ap_handle_sctp_disconnection(s1ap_state_t* state,
 
   if (reset) {
     /* UE state at s1ap task is created on reception of initial ue message
-     * Hash list, ue_id_coll is updated after mme_app_task assigns and provides
+     * The map, ue_id_coll is updated after mme_app_task assigns and provides
```
ssanadhya:
```suggestion
     * The map ue_id_coll is updated after mme_app_task assigns and provides
```

ssanadhya (approved):
Lgtm, with minor nit

PR #12971, opened 2022-06-13
fix(ci): Fix bazel remote cache config for bazel version 5

LKreutzer:

## Summary

- In bazel 5 the disk and repository cache configs (in `config:docker`) clash with the remote cache config and the PUT requests to the remote cache are not processed properly.
  - The disk and repository cache configs are disabled for the bazel base container.

## Test Plan

CI

## Additional Information

### DIFF
--- a/bazel/bazelrcs/docker.bazelrc
+++ b/bazel/bazelrcs/docker.bazelrc
@@ -1,1 +1,1 @@
-common --config=docker
+build --config=docker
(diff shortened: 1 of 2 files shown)


bazel/bazelrcs/docker.bazelrc
```diff
@@ -1 +1 @@
-common --config=docker
```
LKreutzer, line 1: This change is required because the `bazel query` in the test job otherwise uses an empty config which leads to an error.

nstng (approved):
lgtm - in pairing

PR #12972, opened 2022-06-13
chore(agw): Update Python dependencies [DO NOT MERGE]

sebathomas:
⚠️ At the moment, the build-all job is broken on master and cannot provide all necessary Python dependencies. This should be fixed before this is merged, and then it should be tested with the upgraded dependencies.

## Summary

Upgrade as many Python dependencies as possible now that we no longer have to support Python 3.5 for parts of the code base.

## Test Plan

I ran the integration tests on my fork. The failing test in the result is currently also red on master and will be fixed by #12979.

https://github.com/sebathomas/magma/actions/runs/2490580856

## Additional Information


lucasgonze:
This is a superset of https://github.com/magma/magma/pull/13222, so I'll close that.

sebathomas:
Closed in favor of #13482.

PR #12974, opened 2022-06-13
feat(ci): Add mypy check of AGW

voisey:
## Summary

This is for testing purposes.

### DIFF
--- a/.github/workflows/python-workflow.yml
+++ b/.github/workflows/python-workflow.yml
@@ -1,5 +1,5 @@
 ---
-name: Python Format Check
+name: Python Format & Typing Check
 on:  # yamllint disable-line rule:truthy
   push:
     branches:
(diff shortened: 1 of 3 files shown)


.github/workflows/python-workflow.yml
```diff
@@ -94,3 +98,26 @@ jobs:
...
+        run: |
+          pip3 install --upgrade pip
+          pip3 install mypy==0.960
+      - name: Run mypy in AGW
+        run: |
+          for file in ${{ needs.pre_job.outputs.files_changed_agw }};
+          do
+            set -e
+            echo "Running mypy on $file...";
+            mypy --ignore-missing-imports $MAGMA_ROOT/$file;
```
alexzurbonsen, line 122: We might consider to use `--install-types --non-interactive` as well, since for some package type annotations exist but have to be installed?

PR #12989, opened 2022-06-14
fix(agw): Deleted residual OVS flow rules after continuous random attach

VinashakAnkitAman:
## Title
fix(agw): Deleted residual OVS flow rules after continuous random attach

## Summary
The continuous random attach test case were leaving residual flow rules after execution, because of subsequent test cases with flow rules verification were failing. This PR removes residual flow rules after test case execution.

## Test plan
Verified with sanity


ardzoht (approved):

pruthvihebbani (approved):
lgtm

PR #12990, opened 2022-06-14
chore: pin github action

lucasgonze:
Fixes https://github.com/magma/security/issues/97

Version numbers were converted to hashes using https://github.com/mheap/pin-github-action/. We'll need to periodically re-run this in the future, and might want a monthly action.

I had to modify `/magma/.github/workflows/config/yamllint_config.yml` to accept the YAML output formatting from that script, as follows:

```
15,20c15,16
<   braces: 
<     max-spaces-inside: 1
<     min-spaces-inside: 1
<   brackets:
<     min-spaces-inside: 0
<     max-spaces-inside: 1
---
>   braces: enable
>   brackets: enable
25d20
<     min-spaces-from-content: 1
31,32c26
<   empty-lines: 
<     max: 3
---
>   empty-lines: enable
43c37
<   trailing-spaces: disable
---
>   trailing-spaces: enable
```

## Additional Information

### DIFF
--- a/.github/workflows/agw-docker-load-test.yaml
+++ b/.github/workflows/agw-docker-load-test.yaml
@@ -1,7 +1,7 @@
 ---
 name: AGW docker load test
 
-on:  # yamllint disable-line rule:truthy
+on:
   workflow_run:
     workflows:
       - build-all
@@ -25,14 +25,14 @@ jobs:
       WORK_DIR: "${{ github.workspace }}/experimental/cloudstrapper/playbooks"
       AGW_DOCKER_AMI: "ami-0150e153a94c122b5"
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@7884fcad6b5d53d10323aee724dc68d8b9096a2e # pin@v2
       - name: Run apt
         run: sudo apt-get update && sudo apt -y upgrade
       - name: setup pyenv
-        uses: "gabrielfalcao/pyenv-action@v8"
+        uses: "gabrielfalcao/pyenv-action@5327db2939908b2ef8f62d284403d678c4b611d0" # pin@v8
         with:
           default: 3.7.0
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@7f80679172b057fc5e90d70d197929d454754a5a # pin@v2
         with:
           python-version: '3.7'
       - name: Install Dependencies
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -1,7 +1,7 @@
 ---
 name: agw-workflow
 
-on:  # yamllint disable-line rule:truthy
+on:
   push:
     branches:
       - master
@@ -10,7 +10,7 @@ on:  # yamllint disable-line rule:truthy
     branches:
       - master
       - 'v1.*'
-    types: [opened, reopened, synchronize]
+    types: [ opened, reopened, synchronize ]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
@@ -28,9 +28,9 @@ jobs:
       should_not_skip: ${{ steps.changes.outputs.filesChanged }}
     steps:
       # Need to get git on push event
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@7884fcad6b5d53d10323aee724dc68d8b9096a2e # pin@v2
         if: github.event_name == 'push'
-      - uses: dorny/paths-filter@v2
+      - uses: dorny/paths-filter@b2feaf19c27470162a626bd6fa8438ae5b263721 # pin@v2
         id: changes
         with:
           filters: |
@@ -46,7 +46,7 @@ jobs:
         run: |
           mkdir -p ./pr
           echo -n ${{ steps.changes.outputs.filesChanged == 'false' }} > ./pr/skipped
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # pin@v3
         if: always()
         with:
           name: pr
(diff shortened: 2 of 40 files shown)


lucasgonze:
Marked ready for review despite two failures, because I think it's unlikely those failures were caused by my changes. 

Are these items longstanding problems?
```
agw-workflow / mme test job (pull_request) Failing after 38m 
agw-workflow / C / C++ code coverage (pull_request) 
```

LKreutzer:
> Marked ready for review despite two failures, because I think it's unlikely those failures were caused by my changes.
> 
> Are these items longstanding problems?
> 
> ```
> agw-workflow / mme test job (pull_request) Failing after 38m 
> agw-workflow / C / C++ code coverage (pull_request) 
> ```

Yes, these are currently known issues, restarting the workflow can fix it.

.github/workflows/agw-workflow.yml
```diff
@@ -156,33 +156,36 @@ jobs:
...
             echo "Pulled the devontainer image!"
       - name: Create a directory to store XML test results
-        uses: addnab/docker-run-action@v3
+        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 # pin@v3
         with:
           shell: bash
           image: ${{ env.BAZEL_BASE_IMAGE }}
           # TODO: Remove work-around mount of Github workspace to /magma (https://github.com/addnab/docker-run-action/issues/11)
-          options: -v ${{ github.workspace }}:/workspaces/magma/ -v ${{ github.workspace }}/lte/gateway/configs:/etc/magma
+          options: -v ${{ github.workspace }}:/workspaces/magma/ -v ${{ github.workspace
```
LKreutzer: This change looks somewhat strange - was this an auto-formatter? It looks like line 200 might even be longer and was not changed?

.github/workflows/agw-workflow.yml
```diff
@@ -156,33 +156,36 @@ jobs:
...
+            }}/lte/gateway/configs:/etc/magma
           run: |
             cd /workspaces/magma
             mkdir -p c-cpp-test-results/
             echo "created c-cpp-test-results/ directory"
-      - name: Run `bazel test //orc8r/gateway/c/...:* //lte/gateway/c/...:* --config=asan`
-        uses: addnab/docker-run-action@v3
+      - name: Run `bazel test //orc8r/gateway/c/...:* //lte/gateway/c/...:*
+          --config=asan`
```
LKreutzer: This line break might be unnecessary - as above. Can we configure this?

.github/workflows/agw-workflow.yml
```diff
@@ -156,33 +156,36 @@ jobs:
...
+        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 # pin@v3
         with:
           shell: bash
           image: ${{ env.BAZEL_BASE_IMAGE }}
           # TODO: Remove work-around mount of Github workspace to /magma (https://github.com/addnab/docker-run-action/issues/11)
-          options: -v ${{ github.workspace }}:/workspaces/magma/ -v ${{ github.workspace }}/lte/gateway/configs:/etc/magma
+          options: -v ${{ github.workspace }}:/workspaces/magma/ -v ${{ github.workspace
```
LKreutzer: As above.

.github/workflows/amis-workflow.yml
```diff
@@ -1,8 +1,8 @@
 ---
 name: amis-workflow
 
-on:  # yamllint disable-line rule:truthy
-  workflow_dispatch:
+on: # yamllint disable-line rule:truthy
+  workflow_dispatch: null
```
LKreutzer, line 5: Why is this needed?

LKreutzer (approved):
The version pinning looks good!

Regarding the auto-formatting of the `name`, `args` and `options` fields: I find that in many cases the formatted version is less readable and it might be better to allow for a greater line length limit.

.github/workflows/agw-workflow.yml
lucasgonze: Right, this is the auto-formatter. 

I have changed the configuration to break lines at 200 characters.

.github/workflows/agw-workflow.yml
lucasgonze: Agreed. In the latest version lines are auto-wrapped at 200 characters.

.github/workflows/amis-workflow.yml
lucasgonze: That is a default of the auto-formatter. See nullStr at https://eemeli.org/yaml/#tostring-options . I have changed the options to suppress that "null".

maxhbr:
This PR should also add documentation (or documentation should be added to the wiki) (maybe copy from security ticket). And it should give the tools to keep the pins up to date, or at least should explain how to do that.

.github/workflows/cwf-integ-test.yml
```diff
@@ -2,8 +2,9 @@
 
 name: CWF integ test
 
-on:  # yamllint disable-line rule:truthy
-  workflow_dispatch:
+on:
+  # yamllint disable-line rule:truthy
```
sebathomas: Did you move these yamllint comments accidentally (looks like it could be an autoformat issue)? It seems the lines now trigger the linter.

lucasgonze: I replaced the case-by-case `truthy` configurations with one at the global level. We were either overriding it or ignoring the warning pretty much everywhere.

sebathomas: I agree the check should be disabled globally, but it seems with the changes to these comments there are a lot of lint errors in this PR?

.github/workflows/agw-docker-load-test.yaml
```diff
@@ -1,7 +1,8 @@
 ---
 name: AGW docker load test
 
-on:  # yamllint disable-line rule:truthy
+on:
+  # yamllint disable-line rule:truthy
```
Neudrino: Those changes should be reverted everywhere. It needs to be in the same line.

Neudrino (changes requested):

.github/workflows/agw-docker-load-test.yaml
lucasgonze: The disable-line statement is vestigial since adding `truthy: disable` to config/yamllint_config.yml. Whether it's on the same line has no impact. In the latest commit I have deleted all of the extra "truthy" lines.

.github/workflows/cwf-integ-test.yml
lucasgonze: Could you help me find the lint warnings, @sebathomas ? I checked before submitting and found just the following, which I believe were present already:

```
$ yamllint *.{yml,yaml} -c config/yamllint_config.yml 
codeql-analysis.yml
  12:1      warning  missing document start "---"  (document-start)
  65:8      warning  missing starting space in comment  (comments)

gcc-problems.yml
  10:1      warning  missing document start "---"  (document-start)

golang-build-test.yml
  158:9     warning  comment not indented like content  (comments-indentation)
```

lucasgonze: I'm seeing a reduction in linter errors. In the current master, `yamllint *.{yml,yaml} -c config/yamllint_config.yml ` produces 31 lines, including the "missing document start" above, and errors as well as warnings. In the new version I'm seeing 10 lines, all warnings, nothing new.

.github/workflows/agw-docker-load-test.yaml
lucasgonze: This is moot because none of these lines has an impact since adding that same setting in the global configuration file. To make that more clear I have eliminated all the line-by-line "yamllint disable-line rule:truthy" statements.

lucasgonze:
> This PR should also add documentation (or documentation should be added to the wiki) (maybe copy from security ticket). And it should give the tools to keep the pins up to date, or at least should explain how to do that.

I have created documentation here:
https://github.com/magma/magma/wiki/Pinning-Actions-By-Hash

maxhbr:
> I have created documentation here:
https://github.com/magma/magma/wiki/Pinning-Actions-By-Hash

there is nothing under that URL

maxhbr:
The following steps currently appear with multiple versions, can you maybe switch them to the later one:
      2 actions/github-script
      2 actions/upload-artifact
      2 addnab/docker-run-action

lucasgonze:
> steps currently appear with multiple versions

I have consolidated the versions so only one is used. In the cases of breaking changes, I have verified that the later version should work.

lucasgonze:
> there is nothing under that URL

https://github.com/magma/magma/wiki/Secure-Use-of-Github-Actions

.github/workflows/cwf-integ-test.yml
sebathomas: So the reviewdog comments in the Changed Files are based on an old version of the PR and don't get removed automatically? That's what's confusing me.

.github/workflows/cwf-integ-test.yml
```diff
@@ -2,8 +2,8 @@
 
 name: CWF integ test
 
-on:  # yamllint disable-line rule:truthy
-  workflow_dispatch:
+on:
+  workflow_dispatch: null
```
Neudrino, line 6: One of the previous comments (https://github.com/magma/magma/pull/12990#discussion_r899375308) sounds as if there should not be a `null` added here?

.github/workflows/config/yamllint_config.yml
```diff
@@ -34,6 +39,5 @@ rules:
   new-lines: enable
   octal-values: disable
   quoted-strings: disable
-  trailing-spaces: enable
-  truthy:
-    level: warning
+  trailing-spaces: disable
+  truthy: disable
```
Neudrino, line 43: _For the record:_ This is the change, which should remove the "truthy warning". (See also https://github.com/magma/magma/pull/12990#discussion_r902820945)

.github/workflows/config/yamllint_config.yml
```diff
@@ -34,6 +39,5 @@ rules:
   new-lines: enable
   octal-values: disable
   quoted-strings: disable
-  trailing-spaces: enable
-  truthy:
-    level: warning
+  trailing-spaces: disable
```
Neudrino: Could you comment, why this change was necessary?
I would see it rather desired, to have no trailing spaces in the yaml files.

Neudrino (approved):

Neudrino:
> https://github.com/magma/magma/wiki/Secure-Use-of-Github-Actions

Thanks for creating such a detailed documentation! :100:

Neudrino:
_For reference:_ The process of conversion is tool based. Further automation will require https://github.com/mheap/pin-github-action/issues/83.

.github/workflows/cwf-integ-test.yml
Neudrino: [The workflow is triggered `on: pull_request_target`](https://github.com/magma/magma/blob/master/.github/workflows/reviewdog-workflow.yml#L4). That means, that the actual file which is executed on this PR itself, is the code from current `master`, where the respective change is not yet active.

To remove the warnings here, one would need a two step process.
1. Changing the "yaml lint config".
2. Changing everything else.

lucasgonze: I have created a single-purpose PR to change the yamllint config:

https://github.com/magma/magma/pull/13053

.github/workflows/config/yamllint_config.yml
lucasgonze: The yaml generator sometimes leaves trailing spaces. 

I have re-enabled the trailing spaces rule and edited all the generated code to prevent the warning. 

I have also cleaned up yaml formatting causing three other warnings, so we can have 100% clean yamllint output unless a new problem is introduced, which will make the output more useful:

1. comment not indented like content
2. missing starting space in comment
3. missing document start "---"

Neudrino:
> [...] I also like that Dependabot would go to the issue @maxhbr raised early on about fragility caused by missing upgrades.
> 
> I have created a ticket so these don't get lost.

:arrow_right: https://github.com/magma/security/issues/105

.github/workflows/cwf-integ-test.yml
Neudrino: > I have created a single-purpose PR to change the yamllint config:
> - #13053

#13053 was closed. This is ok, as the linter warnings are only in this PR, and we decided to ignore them here, as the configuration will be adopted.

Neudrino: I should have made clearer that mentioning the 2 step process was a "theoretical workaround", which is not necessary to apply here now, imho.
My appologies, @lucasgonze !

lucasgonze:
> I am happy with that now (although I am also still afraid of the fragility and the raise of complexity).

If you have specifics on fragility and complexity, I will gladly work on them. Outside of this PR I will be making contributions to the pin-github-action tool to improve developer experience - I made my first PR there yesterday.

> But also: can you add to the wiki documentation how and when to update the pinned jobs? E.g. when not to update and what to check before updating.

I have updated https://github.com/magma/magma/wiki/Secure-Use-of-Github-Actions with this information.

lucasgonze:
I have submitted a PR to pin-github-action that should make maintaining pinned actions almost as convenient and therefore robust as go.sum or packages.lock. 

To update all actions by a highly trusted publisher:
`pin-github-action --only actions/*`

To update (and pin) to a tagged version:
`pin-github-action --only foo/bar@v1.2.3`

To update a single package to the latest hash for the version in the yaml:
`pin-github-action --only foo/bar`

Neudrino:
Why is there a commit by @jheidbrink on this PR now?
I think this should be removed, and only your changes, @lucasgonze , should be in this PR.

lucasgonze:
> commit by jheidbrink

Thank you for catching this. I have removed the ready2merge label for the moment.

tmdzk (approved):
Thanks for your service. this looks like a painful PR

Neudrino (approved):

PR #13002, opened 2022-06-15
fix(nms): applying date filter on event component in dashboard screen

talesmota:
Signed-off-by: Tales Mota <tales@radtonics.com>

## Summary
Applying the Date Filter on Event Component on Dashboard Screen.

screenshot

Files related:
- EventsTable.js

## Test Plan

Manually checking the date on Http Request.


nms/app/views/events/EventsTable.js
```diff
@@ -261,6 +261,8 @@ export default function EventsTable(props: EventTableProps) {
   }, [startDate, endDate]);
 
   useEffect(() => {
+    setStartDate(props.inStartDate);
```
thmsschmitt: Apart from the formating issues, there are several problems here:

- `props.inStartDate` and  `props.inEndDate` can be undefined. Setting the start and end date to undefined breaks the `EventsTable` this happening if you open the gateway details and crashes the page.
- `setStartDate` and `setEndDate` are missing in the dependencies of useEffect. See the linter warning below. But to be able to add them we need to wrap the callback for `useRefreshingDateRange` into a `useCallback`.
- if start and end date are changed `useRefreshingDateRange` will already refresh the table to there is no need to call `onQueryChange` again.

thmsschmitt (commented):
Unfortunately there are a few issues here

nms/app/views/events/EventsTable.js
talesmota: Hi @thmsschmitt, very kind you feedback. I'll check all these points here and I'll update the PR.

nms/app/views/events/EventsTable.js
```diff
@@ -262,11 +262,13 @@ export default function EventsTable(props: EventTableProps) {
 
   useEffect(() => {
     if (props.inStartDate && props.inEndDate) {
+      props.inStartDate !== undefined && setStartDate(props.inStartDate);
+      props.inEndDate !== undefined && setEndDate(props.inEndDate);
       if (tableRef.current) {
         tableRef.current.onQueryChange();
```
talesmota: this behaviour at line 268 had already been present before my commit so I didn't touch although @thmsschmitt  said to remove.

nms/app/views/events/EventsTable.js
```diff
@@ -262,11 +262,13 @@ export default function EventsTable(props: EventTableProps) {
...
       }
     }
-  }, [props.inStartDate, props.inEndDate]);
```
thmsschmitt, line 269: As I mentioned above to be able to to use `setStartDate` and `setEndDate` as dependencies we need to make sure that they are stable. And to so we need to stabilise the callback for `useRefreshingDateRange` i.e. wrap it into useCallback.

```
const {startDate, endDate, setStartDate, setEndDate} = useRefreshingDateRange(
   isAutoRefreshing,
   30000,
   useCallback(() => {
     tableRef.current?.onQueryChange();
   }, []),
);
```

nms/app/views/events/EventsTable.js
```diff
@@ -262,11 +262,13 @@ export default function EventsTable(props: EventTableProps) {
 
   useEffect(() => {
     if (props.inStartDate && props.inEndDate) {
```
thmsschmitt: Because of the this if statement the checks below don't really do anything.

nms/app/views/events/EventsTable.js
thmsschmitt: I think the motivation for `tableRef.current.onQueryChange` here was to refetch if props.inStartDate or props.inEndDate change. Which as you noticed doesn't work because the query uses the state and not the props. We don't have to `onQueryChange` explicitly because  `setStartDate` and `setEndDate` do also call it.

I would suggest something like:

```
useEffect(() => {
  if (props.inStartDate) {
    setStartDate(props.inStartDate);
  }
}, [props.inStartDate, setStartDate]);

useEffect(() => {
  if (props.inEndDate) {
    setEndDate(props.inEndDate);
  }
}, [props.inEndDate, setEndDate]);
```

nms/app/views/events/EventsTable.js
talesmota: I made the suggested adjustments...

thmsschmitt (approved):
This looks great now, thanks for your effort!

PR #13003, opened 2022-06-16
chore(ci): cache Vagrant boxes with Github Actions Cache

jheidbrink:
## Summary

The Github Actions workflows see a lot of HTTP 429 when downloading Vagrant boxes from the Vagrant Cloud.

As a quick fix, simply cache the `~/.vagrant.d/boxes` directory using the `actions/cache` action.

## Test Plan

Spin up the `magma_dev`, `magma_test` and `magma_trfserver` VMs and verify the cache is used: https://github.com/jheidbrink/magma/actions/runs/2507266285

## Additional Information

All our boxes (lte, cwf) currently use around 8 GiB of disk space. That is close to the cache size limit of 10GiB. Given that we don't use the Github Actions caches elsewhere and don't regularly update the boxes, this should be ok for now. Once the boxes receive more regular upgrades, we need to find a more elaborate solution.

### DIFF
--- a/.github/workflows/bazel-cache-push.yml
+++ b/.github/workflows/bazel-cache-push.yml
@@ -30,6 +30,11 @@ jobs:
       - run: echo "::set-output name=date::$(date +'%m-%d-%Y--%H-%M-%S')"
         id: date
       - uses: actions/checkout@v2
+      - name: Cache Vagrant Boxes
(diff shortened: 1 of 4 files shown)


nstng (approved):
lgtm - discussed offline

PR #13005, opened 2022-06-16
fix(ci): Fix Bazel failure Slack message in agw-workflow

vktng:
## Summary

The failure reporting step was not triggered on failure. We have to still include failure() to override the default status check of success() that is automatically applied to if conditions that don't contain a status check function.

## Test Plan

CI.
Also tested on fork: https://github.com/vktng/magma/runs/6917739513?check_suite_focus=true

## Additional Information


nstng (approved):
lgtm, nice to figure out that success() is added implicitly :+1:

PR #13006, opened 2022-06-16
chore: Link glog library dynamically in Bazel builds

vktng:

## Summary

The glog library is not working correctly when it is statically linked. (Sometimes log files or links to them are not created.)
By linking it dynamically we avoid this problem.

## Test Plan

CI.

## Additional Information

### DIFF
--- a/bazel/cpp_repositories.bzl
+++ b/bazel/cpp_repositories.bzl
@@ -24,13 +24,6 @@ def cpp_repositories():
     )
 
     http_archive(
-        name = "com_github_google_glog",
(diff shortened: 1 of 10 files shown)


nstng (approved):
lgtm :+1:
We should track that at one point we look into refactoring mme to allow for a statically linked glog. But for now this is imho the better solution - mainly because the make artifacts links glog also dynamically.

jheidbrink (approved):

themarwhal (approved):

PR #13009, opened 2022-06-17
refactor(agw): Replaced folly with nlohmann in OperationalStatesHandler

MoritzThomasHuebner:
Signed-off-by: Moritz Huebner <moritz.huebner@tngtech.com>

## Summary

We want to move to replace `folly` with `nlohmann` within magma. `folly` is a large C++ library that is only used for json conversion and IP parsing. `nlohmann` is a much smaller dependency that only handles jsons. Replacing `folly` thus means that we can get rid of a large dependency. This PR is a small step towards it.

- Replaced with nlohmann package in `OperationalStatesHandler.cpp` and `OperationalStatesHandler.hpp`.
- Updated Bazel file

## Test Plan

- Run unit tests

## Additional Information

### DIFF
--- a/lte/gateway/c/session_manager/BUILD.bazel
+++ b/lte/gateway/c/session_manager/BUILD.bazel
@@ -92,6 +92,7 @@ cc_library(
         "@cpp_redis",
         "@system_libraries//:libglog",
         "@yaml-cpp//:yaml-cpp",
+        "@github_nlohmann_json//:json",
     ],
 )
 
@@ -119,7 +120,7 @@ cc_library(
         "//lte/protos:pipelined_cpp_grpc",
         "//lte/protos:session_manager_cpp_grpc",
         "//orc8r/gateway/c/common/logging",
-        "@system_libraries//:folly",
+        "@github_nlohmann_json//:json",
         "@system_libraries//:libglog",
     ],
 )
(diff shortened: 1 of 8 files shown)


MoritzThomasHuebner:
I trimmed this down to only replace folly from the OperationalStatesHandler since there are some issues with the StoredState.

themarwhal:
Do we have any unit tests that check the operational states handler logic in SessionD? If not, we might benefit from adding a small one.

LKreutzer (commented):
Do we have an issue where we track the progress of the folly to nlohmann migration?

MoritzThomasHuebner:
> Do we have any unit tests that check the operational states handler logic in SessionD? If not, we might benefit from adding a small one.

There is this test, or do you have something else in mind @themarwhal ?

https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/test/test_operational_states_handler.cpp

themarwhal:
> > Do we have any unit tests that check the operational states handler logic in SessionD? If not, we might benefit from adding a small one.
> 
> 
> 
> There is this test, or do you have something else in mind @themarwhal ?
> 
> 
> 
> https://github.com/magma/magma/blob/master/lte/gateway/c/session_manager/test/test_operational_states_handler.cpp

Yeah this looks great!

themarwhal (approved):
lgtm! thanks for the cleanup

PR #13012, opened 2022-06-17
Ci add federated integ test

crasu:
Test PR for planning around with the feg CI setup.

This PR is a clone of #12633. 

The feg integ tests now work on this branch. However #12633 still needs some polishing before it can  be merged into master.

### DIFF
--- a/.github/workflows/federated-integ-test.yml
+++ b/.github/workflows/federated-integ-test.yml
@@ -23,11 +23,65 @@ on:  # yamllint disable-line rule:truthy
     types: [opened, reopened, synchronize]
 
 jobs:
+  # Build images on ubuntu which is faster than macos
+  docker-build-orc8r:
+    runs-on: ubuntu-latest
+    env:
+      MAGMA_ROOT: "${{ github.workspace }}"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build Orc8r docker images
+        run: |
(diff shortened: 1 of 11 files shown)


uri200 (commented):
lets separate those changes into groups.

I woukd advise to put hss changes on a different PR and work from there

uri200 (commented):
btw good catch with the hss issue!

lte/gateway/fabfile.py
```diff
@@ -219,23 +220,86 @@ def s1ap_setup_cloud():
...
+
         local("fab configure_orc8r")
-        local("fab test_connectivity")
+
+        #FEG may need to be restarted to reload some configs
+        print("Wait for configuration to be available")
+        sleep(60)
+        print("Restart hss")
+        local('fab restart_feg_hss')
+        sleep(20)
```
uri200: can you please check if the test works without those lines (296 to 300) ?

Your patch may have fixed the need for that hack

.github/workflows/federated-integ-test.yml
```diff
@@ -0,0 +1,220 @@
+---
+
+name: Federated integ test
+
+on:  # yamllint disable-line rule:truthy
```
Neudrino, line 5: With https://github.com/magma/magma/pull/12990 merged, we can actually remove the lint ignore.

PR #13013, opened 2022-06-17
feat(dp): domain proxy services metrics implementation

WojSad:

feat(dp): domain proxy services metrics implementation

## Summary

Added prometheus metrics support to Domain Proxy
Created a metricsd client on the Domain Proxy's side which can be used to collect metrics produced by DP's services.
The metrics are pushed from Domain Proxy to Metricsd to a new gRPC method `PushRaw`. From there on out the metrics are processed the same way as for all other services

## Test Plan

Tests included

## Additional Information


dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
```diff
@@ -432,6 +434,30 @@ def given_single_step_cbsd_provisioned(self, builder: CbsdAPIDataBuilder) -> int
...
+            'dp_cc_pending_requests_fetching_seconds_sum',
+            'dp_cc_request_processing_seconds_bucket',
+            'dp_cc_request_processing_seconds_count',
+            'dp_cc_request_processing_seconds_sum',
+            'dp_cc_response_processing_seconds_bucket',
+            'dp_cc_response_processing_seconds_count',
+            'dp_cc_response_processing_seconds_sum',
+        ]
+        for m in metrics:
+            resp = requests.get(prom + '/api/v1/query', params={'query': m})
```
jkmar: Is it possible to check all metrics with just one http request?

dp/cloud/python/magma/configuration_controller/response_processor/response_db_processor.py
```diff
@@ -45,6 +46,7 @@ def __init__(
         self.request_states_map = {}
         self.fluentd_client = fluentd_client
 
+    @SAS_RESPONSE_PROCESSING_TIME.time()
```
jkmar, line 49: Not sure about this, I though we were interested in time between sending request to SAS and receiving response, but here it is just internal response processing time.

dp/cloud/python/magma/metricsd_client/client.py
```diff
@@ -0,0 +1,54 @@
...
+from orc8r.protos.metricsd_pb2 import RawMetricsContainer
+from orc8r.protos.metricsd_pb2_grpc import CloudMetricsControllerStub
+from prometheus_client import CollectorRegistry
+
+logging.basicConfig(
+    level=logging.DEBUG,
+    datefmt='%Y-%m-%d %H:%M:%S',
+    format='%(asctime)s %(levelname)-8s %(message)s',
+)
+logger = logging.getLogger("configuration_controller.run")
```
jkmar: why is this called configuration controller?

dp/cloud/python/magma/radio_controller/metrics.py
```diff
@@ -0,0 +1,27 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+from prometheus_client import Histogram
+
+# Metrics for current configuration controller status
+GRPC_REQUEST_PROCESSING_TIME = Histogram(
```
jkmar, line 17: Maybe add a TODO to check if there is similar library that orc8r uses for grpc metrics

dp/cloud/python/magma/radio_controller/run.py
```diff
@@ -48,6 +52,19 @@ def run():
     """
     logger.info("Starting grpc server")
     config = get_config()
+    scheduler = BackgroundScheduler()
```
jkmar, line 55: Configuration controller uses the same scheduler can we extract the common part?

jkmar (approved):

dp/cloud/python/magma/configuration_controller/response_processor/response_db_processor.py
WojSad: true, i just put it there as just another metric. to get more granularity in case if it affects the overall request processing time

dp/cloud/python/magma/metricsd_client/client.py
WojSad: the power of Copy-Paste method :)

dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
WojSad: maybe it's just me but i couldnt find a suitable endpoint in the documentation

WojSad:
@magma/approvers-orc8r please review this PR's orc8r part. It affects metricsd by adding gRPC method to it.

sebathomas:
Just for my understanding: Why do we need a new interface in metricsd instead of using the existing one?

Did you run an E2E test (locally or with a deployed orc8r) that the metrics are transmitted correctly?

WojSad:
@sebathomas There's several reason for the new grpc endpoint:
1. I needed to push metrics from Domain Proxy's internal services `Configuration Controller` and `Radio Controller`, which are not orc8r services, but I wanted them to be processed by metricsd the same way as all other orc8r services. I needed something exactly like the `metricsd`'s `ConsumeCloudMetrics` method, but instead of polling, I wanted to be able to push my metrics. So I extracted some code from `ConsumeCloudMetrics` into  `processMetricFamily` and used it in my new `PushRaw` endpoint. I couldn't simply use the existing polling mechanism, because the mentioned services are not in orc8r's service registry and thus had to be pushed to metricsd.
2. The existing `Push` endpoint does not process metrics the same way `ConsumeCloudMetrics` does. As far as I can tell is only used by `nms` and it requires a composite struct called `PushedmetricContainer`. Once this struct arrives at metricsd, it's then stripped into raw Prometheus metrics and context by the `pushedMetricsToMetricsAndContext` method like so: ```for _, e := range metricsExporters {
		metricsToSubmit := pushedMetricsToMetricsAndContext(in)
		err := e.Submit(metricsToSubmit)```
I didn't want to wrap my raw metrics into a struct just so that they can later be unwrapped back into raw metrics.
3. Even if I did, `PushedmetricContainer` doesn't have fields such as `hostName` and `service` that are needed to display the metrics the same way as other orc8r services and so I needed a different type of struct to be pushed.

sebathomas (approved):
Orc8r changes look good to me, thanks for the detailed response! :+1:

PR #13019, opened 2022-06-18
fix(dp): Fix typo in domain-proxy Makefile

### LINKED ISSUE #12855: [Proposal] Make `Domain proxy integration tests with orc8r` required check
## Brief Context & scope

We would like to have the `Domain proxy integration tests with orc8r` Check from `dp-workflow` required in CI. 
Changes in orc8r possibly can have an impact on the Domain Proxy component, we would like to avoid merging changes which break compatibility between Orchestrator and Domain Proxy.
The Check was improved and recently is robust and we didn't see flakiness in this job

Link to job definition: https://github.com/magma/magma/blob/7fe7fde1607dd015c4210b2d70de0f8efacd78d8/.github/workflows/dp-workflow.yml#L370

## Goals
Prevent any PRs that break this test from being merged into master.

## Proposal
- Make an announcement in #dev that this test will now be mandatory.
- Wait 3 days for any objections
- Turn on the required check!
- Update documentation on GitHub wiki - move check from "non-blocking" to "blocking" table

sklgromek:

## Summary
Fix typo iuntroduced by https://github.com/magma/magma/pull/12999

## Additional Information

### DIFF
--- a/.github/workflows/dp-workflow.yml
+++ b/.github/workflows/dp-workflow.yml
@@ -6,7 +6,7 @@ on:  # yamllint disable-line rule:truthy
     branches:
       - master
       - 'v1.*'
-  pull_request_target:
+  pull_request:
     branches:
       - master
       - 'v1.*'
--- a/dp/Makefile
+++ b/dp/Makefile
@@ -20,7 +20,7 @@ else
 endif
 
 ifdef DP_METRICS
-APPEND_METRICS_PROFILE = ,metrics
+APPEND_METRICS_PROFILE=,metrics
 endif
 
 .PHONY: local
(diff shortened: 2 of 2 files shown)


xbend (approved):

xbend:
@magma/approvers-ci 
Please review and approve the changes in this PR. It fixes master CI runs, and since the dp workflow will soon be made a required CI job, it's pretty urgent we get this into master.

.github/workflows/dp-workflow.yml
```diff
@@ -6,7 +6,7 @@ on:  # yamllint disable-line rule:truthy
     branches:
       - master
       - 'v1.*'
-  pull_request_target:
+  pull_request:
```
maxhbr, line 9: this reverts the change suggested in https://github.com/magma/magma/issues/12855#issuecomment-1149883837. Can you elaborate why this was done? I think @Neudrino had some reason for his suggestion

maxhbr (commented):
The changes look good, just the `pull_request_target` -> `pull_request` change is not explained

.github/workflows/dp-workflow.yml
xbend: The suggestion was to use the `pull_request_target` trigger action to make secrets available. We took the advice without actually verifying the validity of the suggestion.
`pull_request_target` is documented as follows:
```
This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.
```
We are building code from the pull request, as that is the whole idea of running integration tests.

Secondly, the only secret we use is the `GITHUB_TOKEN` which is an exception to availability of secrets in PRs from forks:
```
With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.
```

Now, since we have introduced the `pull_request_target`, the CI was run from the base of the PRs, rather from the head of the PR branch, and that hid the mistake in the `Makefile`.

maxhbr (approved):
LGTM

PR #13023, opened 2022-06-20
chore(agw): Determine Ubuntu AMI ID dynamically

jheidbrink:
## Summary

Currently one has to manually change the AMI ID when deploying to a different AWS region. This change uses the `amazon.aws.ec2_ami_info` Ansible module to dynamically determine the Ubuntu AMI ID during Ansible invocation, making it easier to deploy to a different region.

## Test Plan

Followed [experimental/cloudstrapper/README.md](https://github.com/magma/magma/blob/c82dcc560582bc293f4cb08d25b017ff906d8ac3/experimental/cloudstrapper/README.md) and bits of [lte/gateway/docker/readme.md
](https://github.com/magma/magma/blob/c82dcc560582bc293f4cb08d25b017ff906d8ac3/lte/gateway/docker/readme.md) and verified I could set up an AGW instance on AWS

## Additional Information

### DIFF
--- a/.github/workflows/amis-workflow.yml
+++ b/.github/workflows/amis-workflow.yml
@@ -83,7 +83,6 @@ jobs:
             sed -i -e '/^stackDevOpsCloudstrapper: /s/:.*$/: publish-amis-to-marketplace-stack-devopscloustrapper/' ${{ env.VARS_DIR }}/defaults.yaml
             sed -i -e '/^stackCloudstrapper: /s/:.*$/: publish-amis-to-marketplace-stack-cloustrapper/' ${{ env.VARS_DIR }}/defaults.yaml
             sed -i -e '/^devOpsAmi: /s/:.*$/: cloudstrapper-'"$VERSION"'/' ${{ env.VARS_DIR }}/defaults.yaml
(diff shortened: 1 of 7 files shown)


experimental/cloudstrapper/playbooks/roles/cloudstrapper-infra/tasks/main.yaml
```diff
@@ -7,7 +7,7 @@
   ec2_ami_info:
     filters:
       "tag:Name": "{{ devOpsAmi }}"
-  register: valAmi
```
Neudrino: This might be undesired, as in line 14 `valAmi` is still used.

Neudrino (changes requested):
Looks good overall.
One inline comment to be addressed.

_General question:_ 
I understand some Ansible mechanism is used to determine the image AMI.
Why do we not use any Amazon provided mechanism instead?
Would the [Amazon given dynamic endpoints for AMI resolution](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/finding-an-ami.html#finding-an-ami-parameter-store) also work?

jheidbrink:
> I understand some Ansible mechanism is used to determine the image AMI.
> Why do we not use any Amazon provided mechanism instead?

I think the `ec2_ami_info` Ansible module will simply call an AWS API endpoint and make the result available to Ansible.

Neudrino (approved):

PR #13027, opened 2022-06-20
fix(nms): Fix settings link on host portal

thmsschmitt:
## Summary
The link to the account settings is broken on host portal.

## Test Plan
Check that account settings can be opened form:
- the host portal
- the nms portal with a network
- the nms portal **without** a network
Screenshot 2022-06-20 at 12 57 31


Neudrino (approved):
Checked deployment from branch. Can successfully load the page in question.

PR #13028, opened 2022-06-20
fix(ci): bazel coverage failures are mitigated by several flags (see #13026)

### LINKED ISSUE #13026: Bazel coverage in CI is resolved
Since we switched to remote caches in Bazel git hub workflows in the workflow `agw-workflow` in the job `c-cpp-codecov` the step `Run coverage with Bazel (COMMON / SESSIOND / SCTPD / LIAGENT / CONNECTIOND)` has become very flaky with the message:
```
cp: cannot stat 'bazel-out/_coverage/_coverage_report.dat': No such file or directory
```
This file contains (should) the merged coverage of all tests in scope.

Assumption:
* bazel has currently problems with coverage and remote caches - this was to be solved for bazel 5
* it seems our combination of two tests marked as flaky (re-run up to four times if fails) and exclusive (run alone) are causing this behavior - this is determined from observing multiple executions

Approach:
* inspired by https://github.com/bazelbuild/bazel/issues 4685 local experiments are successful using the flags
  * `--[no]experimental_split_coverage_postprocessing` default: "false"
      If true, then Bazel will run coverage postprocessing for test in a new spawn. 
  * `--[no]experimental_fetch_all_coverage_outputs` default: "false"
      If true, then Bazel fetches the entire coverage data directory for each test during a coverage run. 
  * `--remote_download_outputs=<all, minimal or toplevel>` default: "all"
      If set to 'minimal' doesn't download any remote build outputs to the local machine, except the ones required by local actions. If set to 'toplevel' behaves like'minimal' except that it also downloads outputs of top level targets to the local machine. Both options can significantly reduce build times if network bandwidth is a bottleneck.
    * Note: if this is not set to `all` then the build fails with an error - assumption: the other flags are changing this value without documentation
  * see https://docs.bazel.build/versions/main/command-line-reference.html for reference

nstng:

## Summary

see #13026

## Test Plan

* setup remote caching with rw locally
* execute `bazel coverage lte/gateway/c/session_manager/test/...`
  * see that `bazel-out/_coverage/_coverage_report.dat` is (ok) or is not (failure) generated
  * do `bazel clean` between runs
  * you may need multiple runs to see that coverage is not generated
* execute `bazel coverage --experimental_split_coverage_postprocessing --experimental_fetch_all_coverage_outputs --remote_download_outputs=all lte/gateway/c/session_manager/test/...`
  * see that coverage is always generated

## Additional Information

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -390,10 +390,11 @@ jobs:
             cd $MAGMA_ROOT
             bazel/scripts/remote_cache_bazelrc_setup.sh "${{ env.CACHE_KEY }}" "${{ secrets.BAZEL_REMOTE_PASSWORD }}"
             # Collecting coverage with Bazel can be slow. We can follow this thread to see if this can be improved: https://github.com/bazelbuild/bazel/issues/8178
-            # Omit OAI coverage until it is tested. We need to determine what the behavior is for doing both CMake and Bazel based coverage at the same time
-            # TODO: GH11936
+            # Coverage in bazel is flaky for remote caches - the flags below are helping. See GH13026 for details.
+            # TODO: GH11936 Omit OAI coverage until it is tested. We need to determine what the behavior is for doing both CMake and Bazel based coverage at the same time
             bazel coverage \
               --profile=Bazel_test_coverage_profile \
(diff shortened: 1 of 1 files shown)


jheidbrink (approved):

LKreutzer (approved):
Tested locally (but difficult to know for sure - without the change I had 1 failure in 6 runs)

PR #13048, opened 2022-06-21
feat(agw): authentication done via eps_authentication instead of subscriberdb

### LINKED ISSUE #12907: Security: Subscriber Keys Stored in Memory on Host #84

vroon2703:
Signed-off-by: Varun Madathil <varun27@fb.com>

## Summary

Added logic to call eps_authentication service in the case of cloud_subscriberdb_enabled is set  true in the configurations. 

## Test Plan
Attach successful before and after the cloud_subscriberdb_enabled flag is set to true. 

Screen Shot 2022-06-24 at 12 04 06 PM

Screen Shot 2022-06-24 at 12 04 20 PM

## Additional Information

### DIFF
--- a/lte/gateway/c/core/oai/lib/s6a_proxy/S6aClient.cpp
+++ b/lte/gateway/c/core/oai/lib/s6a_proxy/S6aClient.cpp
@@ -138,7 +138,7 @@ S6aClient::S6aClient(bool enable_s6a_proxy_channel) {
   } else if (get_cloud_subscriberdb_enabled()) {
     auto channel = ServiceRegistrySingleton::Instance()->GetGrpcChannel(
         "eps_authentication", ServiceRegistrySingleton::CLOUD);
-    // Create stub for eps_authentication gRPC service
+    // Create S6aProxy stub for eps_authentication gRPC service
     stub_ = S6aProxy::NewStub(channel);
   } else {
     auto channel = ServiceRegistrySingleton::Instance()->GetGrpcChannel(
--- a/lte/gateway/c/core/oai/lib/s6a_proxy/S6aClient.cpp
+++ b/lte/gateway/c/core/oai/lib/s6a_proxy/S6aClient.cpp
@@ -130,20 +130,17 @@ S6aClient::S6aClient(bool enable_s6a_proxy_channel) {
   // then create a channel towards the FeG.
   // Otherwise, create a channel towards either local or cloud-based
   // subscriberdb.
-  std::cout << "[INFO] cloud authentication: " << mme_cloud_subscriberdb_enabled << std::endl;
   if ((get_s6a_relay_enabled() == true) && (enable_s6a_proxy_channel)) {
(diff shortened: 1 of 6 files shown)


lte/gateway/c/core/oai/lib/s6a_proxy/S6aClient.cpp
```diff
@@ -135,6 +135,11 @@ S6aClient::S6aClient(bool enable_s6a_proxy_channel) {
         "s6a_proxy", ServiceRegistrySingleton::CLOUD);
     // Create stub for S6aProxy gRPC service
     stub_ = S6aProxy::NewStub(channel);
+  } else if (get_cloud_subscriberdb_enabled()) {
+    auto channel = ServiceRegistrySingleton::Instance()->GetGrpcChannel(
+        "eps_authentication", ServiceRegistrySingleton::CLOUD);
+    // Create stub for eps_authentication gRPC service
```
ssanadhya: Please add a comment that EPS Authentication service uses the S6aProxy stub

ssanadhya (approved):
Lgtm, with minor comment

emakeev (approved):
LGTM, probably needs a rebase now.

ssanadhya:
@arunuke @ShubhamTatvamasi , could you please review this PR? Note that this is part of the solution for https://github.com/magma/magma/issues/12907

ShubhamTatvamasi (approved):
LGTM

PR #13050, opened 2022-06-22
chore(feg): Pr/fix hss parameter check for feg integ test

crasu:
## Summary

This pr was split from #13012. This fixes the test hss parameter validation. The is a precondition to make the feg integration run on github actions.

### DIFF
--- a/feg/gateway/services/testcore/hss/servicers/config_test.go
+++ b/feg/gateway/services/testcore/hss/servicers/config_test.go
@@ -1,0 +1,47 @@
+package servicers_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"magma/feg/cloud/go/protos/mconfig"
+	"magma/feg/gateway/services/testcore/hss/servicers"
+	gwMconfig "magma/gateway/mconfig"
+)
+
+func TestGetHssProxyConfig(t *testing.T) {
+	conf := generateHssMconfig(t, test_config)
+	assert.Equal(t, "magma-fedgw.magma.com", conf.Server.DestHost)
+}
+
+func generateHssMconfig(t *testing.T, configString string) *mconfig.HSSConfig {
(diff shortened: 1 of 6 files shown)


feg/gateway/services/testcore/hss/servicers/config_test.go
```diff
@@ -0,0 +1,66 @@
+package servicers_test
```
maxhbr, line 1: new files need license headers

uri200:
Can you describe how those changes fixed the issue we were seeing during initialization of federated integ test please?

uri200 (approved):

crasu:
@uri200 
- In the feg integration  test the hss tries to get it config from the orc8r which fails (probably because the orc8r is not yet started)
- After that it is trying to load the config from various files which are empty
- Finally it does start the service with an empty local and remote address which results in the hss service not binding any ports
- The actual fix is in this [commit](https://github.com/magma/magma/pull/13050/commits/d2d98c27d7743d6bda40063b93bef846bac66a6f)
- The hss container/service simply fails when not providing a valid Server.Address/Server.LocalAddress
- After that the container is restarted by docker

OpenQuestions:
- I am unsure if the port should really be optional for these parameters.

uri200:
I think behaviour is now fine: If the config is not available service will keep restarting. Test has some sleeps so if after those sleeps config is not working the service failing will cause the whole test to fail.

Since this is a mock service it is fine to leave some degree of flexibility on how we define the ports.

sebathomas:
The DCO check seems to insist on the umlaut in your name. :shrug:

I think the failing tests are probably just flaky.

crasu:
@sebathomas 
The feg build test is really failing - I am currently looking into it.

sebathomas:
> @sebathomas The feg build test is really failing - I am currently looking into it.

Ah, I missed that one. I think the cloud and agw tests are probably flaky.

PR #13058, opened 2022-06-23
chore: Split mme binary by features

vktng:
## Summary

# Problem

- mme builds for features agw_of (EMBEDDED_SGW=1, S6A_OVER_GRPC=1) and mme_oai (EMBEDDED_SGW=0, DS6A_OVER_GRPC=0) are actually build with differing sets of source files.
- This is set via conditionally adding sub-directories and single files in CMakeLists.txt files.
- There are files that will not even compile if added. Example: pgw_config.c is not added in a make build if EMBEDDED_SGW=0. And an included #if (!EMBEDDED_SGW) is damaged.
- In general it might be a good idea to build artifacts only with needed files.

# Approach Here

- Only add source files needed for feature.
- Some transitive dependencies were removed, because this file is already too big and this will reduce cyclical dependencies in MME.

## Test Plan

`bazel build lte/gateway/c/core:agw_of`
`bazel build lte/gateway/c/core:mme_oai`

If you want to get rid of the unnecessary bloated logging, then you can add this flag:  `--output_filter=DONT_MATCH_ANYTHING`

## Additional Information


LKreutzer (changes requested):
Currently the build is failing because the target
```
cc_test(
    name = "lib_mme_app_map_test",
    size = "small",
    srcs = [
        "test_mme_app_map.cpp",
    ],
    deps = [
        "//lte/gateway/c/core",
        "@com_google_googletest//:gtest_main",
    ],
)
```
references `//lte/gateway/c/core` with expands to `//lte/gateway/c/core:core`, which you removed. Can we specify the target here? I guess `"//lte/gateway/c/core:lib_agw_of",` would be correct?

LKreutzer (changes requested):
For the required Semantic PR / semantic-pr check you can simply rename the PR to `chore: Split mme binary by features`. This is necessary if you have several commits, as the PR title does not keep the prefix.

LKreutzer (commented):
What I don't like about this approach is that the sources, headers, deps etc. are not being sorted by the buildifier, I guess there is no way of avoiding this without duplicating almost everything?

lte/gateway/c/core/BUILD.bazel
```diff
@@ -26,1117 +26,1168 @@ cc_asn1_library(
...
+        "-DS6A_OVER_GRPC=0",
     ],
+    deps = MME_DEPS,
+)
+
+cc_binary(
+    name = "agw_of",
+    srcs = ["oai/oai_mme/oai_mme.c"],
+    copts = ["-DEMBEDDED_SGW=1"],
     linkstatic = False,
```
LKreutzer, line 1170: Regarding both MME binaries: Can MME be linked statically now that we wrapped glog?

vktng:
> Currently the build is failing because the target
> 
> ```
> cc_test(
>     name = "lib_mme_app_map_test",
>     size = "small",
>     srcs = [
>         "test_mme_app_map.cpp",
>     ],
>     deps = [
>         "//lte/gateway/c/core",
>         "@com_google_googletest//:gtest_main",
>     ],
> )
> ```
> 
> references `//lte/gateway/c/core` with expands to `//lte/gateway/c/core:core`, which you removed. Can we specify the target here? I guess `"//lte/gateway/c/core:lib_agw_of",` would be correct?

Fixed.

lte/gateway/c/core/BUILD.bazel
vktng: Yes, this will be a separate PR.

vktng:
> For the required Semantic PR / semantic-pr check you can simply rename the PR to `chore: Split mme binary by features`. This is necessary if you have several commits, as the PR title does not keep the prefix.

Done.

vktng:
> What I don't like about this approach is that the sources, headers, deps etc. are not being sorted by the buildifier, I guess there is no way of avoiding this without duplicating almost everything?

It would be nice to be able to use the buildifier on them, but this is the best solution we have currently.

LKreutzer (approved):
Tested locally in pairing, including `--config=asan` and `lsan`.

Neudrino:
There was a large number of approver groups added, which seemed uncorrelated to this change. I removed them, as it seemed that their review request was unintentional.
Please re-request review from those groups if it was indeed desired.

PR #13070, opened 2022-06-23
fix(lte): Use last resort default subscriber profile in EPS UpdateLoation

emakeev:
…

## Summary

Use last resort, system wide default subscriber profile in EPS UpdateLoation similar to the logic of local subscriberdb

## Test Plan

unit tests

## Additional Information

### DIFF
--- a/lte/cloud/go/services/eps_authentication/servicers/ul.go
+++ b/lte/cloud/go/services/eps_authentication/servicers/ul.go
@@ -30,12 +30,20 @@ import (
 )
 
 const (
-	serviceSelection                  = "oai.ipv4"
+	serviceSelection                  = "magma.ipv4"
 	qosProfileClassID                 = 9
 	qosProfilePriorityLevel           = 15
 	qosProfilePreemptionCapability    = true
 	qosProfilePreemptionVulnerability = false
+
+	defaultMaxUlBitRate uint64 = 2000000000
+	defaultMaxDlBitRate uint64 = 4000000000
 )
+
+var defaultSubscriberProfile = models.NetworkEpcConfigsSubProfilesAnon{
+	MaxDlBitRate: defaultMaxUlBitRate,
+	MaxUlBitRate: defaultMaxDlBitRate,
+}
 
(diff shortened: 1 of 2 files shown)


lte/cloud/go/services/eps_authentication/servicers/ul.go
```diff
@@ -35,8 +35,16 @@ const (
 	qosProfilePriorityLevel           = 15
 	qosProfilePreemptionCapability    = true
 	qosProfilePreemptionVulnerability = false
+
+	defaultMaxUlBitRate uint64 = 2000000000
```
ssanadhya, line 39: This is 2 Gbps uplink and 4Gbps downlink per UE. Can we use the same defaults as [S1AP integ tests](https://github.com/magma/magma/blob/master/lte/gateway/python/integ_tests/s1aptests/s1ap_wrapper.py#L360-L361), i.e. DL: 200 Mbps and UL: 100 Mbps

Also note the default APN name is "magma.ipv4", instead of "oai.ipv4"

ssanadhya (changes requested):

lte/cloud/go/services/eps_authentication/servicers/ul.go
emakeev: I just took the default values from subscriberdb default configs (https://github.com/magma/magma/blob/master/lte/gateway/configs/subscriberdb.yml#L41-L43), to have parity with subscriberdb. Are we planning to switch the integration tests to EPS Auth path?  

will update oai.ipv4

ssanadhya: Ah, I see! Let's keep it at 2 Gbps and 4 Gbps then.

The goal is to switch integration tests to EPS Auth path, by modifying EPS Auth service to run inside AGW dev VM for integration testing. Let's discuss more offline.

ssanadhya (approved):
Lgtm

uri200 (approved):

PR #13071, opened 2022-06-24
chore: Consolidate the bazel build and test options

LKreutzer:

## Summary

- See detailed explanation of the motivation in the issue https://github.com/magma/magma/issues/13073
- The `bazel test` options are consolidated with the `bazel build` options in order to address the same cache entries.
  - This is only required for the `test` options that are marked as `affects_outputs` in [the bazel command line docs](https://bazel.build/reference/command-line-reference#test-options)
- The `--config=mme_unit_test` option is only used in CI if this config should be used during a build step. The default is that `MME_UNIT_TEST` is only set for `bazel test`.
- The `--flaky_test_attempts=5` flag is removed as it is not needed.

## Test Plan
- Check that the same artifacts are generated by `build` and `test`, if `--config=mme_unit_test` is set.
- Check that the commands that inherit from build are compatible with the consolidated options. See:
Screenshot from 2022-06-24 09-37-23

- CI

## Additional Information

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -12,9 +12,9 @@ build --cxxopt=-std=c++14
 build --experimental_strict_action_env
 build --compilation_mode=dbg
 
-# Default test configuration
+# DEFAULT TEST CONFIGURATION
+# Please read the GH issue #13073 before adding "test" options.
 test --test_output=errors
-test --flaky_test_attempts=5
 
 # Use dynamically linked folly library
 build --define=folly_so=1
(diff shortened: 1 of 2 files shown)


vktng (approved):
LGTM.

The query `bazel cquery 'filter("(^[^@])", filter("^(?!.*external)", deps(//lte/gateway/c/core:oai_mme)))' --define species=excelsior --noimplicit_deps` shows the same build configuration ID after `bazel build --config=mme_unit_test ...` as after `bazel test ...`.

PR #13076, opened 2022-06-24
chore(dp): modify get cbsd state

WojSad:
chore(dp): modify get cbsd state

## Summary

Add carrier_agggregation_enabled to GetCbsdState result + return a list of channels instead of 1 channel.

## Test Plan

Tests amended


dp/protos/enodebd_dp.proto
```diff
@@ -25,8 +25,9 @@ message CBSDRequest {
 }
 
 message CBSDStateResult {
-  LteChannel channel = 1;
+  repeated LteChannel channels = 1;
   bool radio_enabled = 2;
```
jkmar, line 29: Do we still need this?

jkmar (approved):

PR #13080, opened 2022-06-24
chore(ci): Add check that all Go versions are consistent

sebathomas:
## Summary

Adds a workflow step to grep yml files and Dockerfiles for Go
versions that are installed, and to make sure they are all set
to the expected version

## Test Plan

- [x] Run the script locally
- [x] Run the new step in the CI
  - Failed run: https://github.com/magma/magma/runs/7043019811?check_suite_focus=true
  - Successful run: https://github.com/magma/magma/runs/7235650661?check_suite_focus=true

## Additional Information

### DIFF
--- a/.github/workflows/golang-build-test.yml
+++ b/.github/workflows/golang-build-test.yml
@@ -19,6 +19,15 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  check_go_version:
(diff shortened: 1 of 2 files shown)


sebathomas:
> Could we assert on the changes done in #13107 too, please?

Added a check for that too.

Neudrino (approved):

Issue #13036, opened 2022-06-20
[subscriberdb] Subscribers not showing up after IMSI deconcealing fix (12567)


vroon2703:
### Your Environment

- **Version:** master
- **Affected Component:** Access Gateway 
- **Affected Subcomponent:** SubscriberDB
- **Deployment Environment:** Vagrant (AGW)

### Describe the Issue

Possibly related commit: https://github.com/magma/magma/pull/12567

After adding a new subscriber via swagger and running subscriber_cli.py list on the gateway VM returns no subscribers. The debug logs indicate the following error: 

ERROR:root:Fetch suci profiles error! [StatusCode.UNKNOWN] handler panic: runtime error: invalid memory address or nil pointer dereference; 

**To Reproduce**
1. Add a subscriber using Swagger (/lte/{network_id}/subscribers) 
2. On the AGW VM, run `subscriber_cli.py list ` 
3. Observe that the added subscriber does not show up. 
4. Run `sudo journalctl -u  magma@subscriberdb` to see the logs

**Expected behavior**
IMSI of the added subscriber shows up. 

**Screenshots**
GET shows that IMSI has been added. 
<img width="1234" alt="Screen Shot 2022-06-20 at 3 52 36 PM" src="https://user-images.githubusercontent.com/105877840/174686424-efb7031b-8b42-46da-9a31-c7c8bbea7352.png">

`subscriber_cli.py list `  does not print any IMSIs
<img width="475" alt="Screen Shot 2022-06-20 at 3 53 18 PM" src="https://user-images.githubusercontent.com/105877840/174686466-fd63ffab-fa69-4c0c-a284-a971d79f3008.png">

Error message shows up:  fetch suci profiles error!

<img width="1637" alt="Screen Shot 2022-06-20 at 3 57 17 PM 1" src="https://user-images.githubusercontent.com/105877840/174686781-4a9770af-ef47-43b3-bc69-e8fcd019dfa1.png">

Akshayp77:
On latest master code base, we have added a new subscriber(IMSI001010000000004) through swagger API and it is getting reflected only after 5 minutes in the subscriberdb. Please refer attached log file and snapshots.
[sys-subdb.log](https://github.com/magma/magma/files/8965824/sys-subdb.log)

![Screenshot (27)](https://user-images.githubusercontent.com/87304387/175263415-84089b08-244c-4fd2-9294-723926d9ada0.png)
[...]

PR #13081, opened 2022-06-24
feat(subscriberdb): Handle NW NGC config nil check correctly

moinuddin1980:
Signed-off-by: Moinuddin Khan <moinuddin.khan@wavelabs.ai>

feat(subscriberdb): Handle NW NGC config nil check correctly

## Summary
Done the following as part of the PR:
- Dereference the NetworkNgcConfigs only if it is not nil and populate the SUCI profiles relevant entries.

## Test Plan
Tested in local VM using docker based Orc8r and AGW. 

## Additional Information
#13036

### DIFF
--- a/lte/cloud/go/services/subscriberdb/load.go
+++ b/lte/cloud/go/services/subscriberdb/load.go
@@ -237,16 +237,19 @@ func LoadSuciProtos(ctx context.Context, networkID string) ([]*lte_protos.SuciPr
 		return nil, fmt.Errorf("network loading failed: %w", err)
 	}
 
+	suciProtos := []*lte_protos.SuciProfile{}
+
 	ngcModel := &lte_models.NetworkNgcConfigs{}
-	ngcConfig := ngcModel.GetFromNetwork(network).(*lte_models.NetworkNgcConfigs)
+	ngcConfig := ngcModel.GetFromNetwork(network)
 	if ngcConfig == nil {
-		return nil, fmt.Errorf("ngcConfig is nil: %w", err)
+		return suciProtos, fmt.Errorf("ngcConfig is nil: %w", err)
 	}
 
-	suciProfiles := ngcConfig.SuciProfiles
-	suciProtos := []*lte_protos.SuciProfile{}
-	for _, suciProfile := range suciProfiles {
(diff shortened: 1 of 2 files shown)


panyogesh:
[Test-Logs-13081.txt](https://github.com/magma/magma/files/8980079/Test-Logs-13081.txt)

lte/cloud/go/services/subscriberdb/load.go
```diff
@@ -237,16 +237,19 @@ func LoadSuciProtos(ctx context.Context, networkID string) ([]*lte_protos.SuciPr
...
 	}
 
+	suciProtos := []*lte_protos.SuciProfile{}
+
 	ngcModel := &lte_models.NetworkNgcConfigs{}
-	ngcConfig := ngcModel.GetFromNetwork(network).(*lte_models.NetworkNgcConfigs)
+	ngcConfig := ngcModel.GetFromNetwork(network)
 	if ngcConfig == nil {
-		return nil, fmt.Errorf("ngcConfig is nil: %w", err)
+		return suciProtos, fmt.Errorf("ngcConfig is nil: %w", err)
```
sebathomas, line 245: I don't think returning an empty `suciProtos` here will change anything. The GRPC call will still return an error (see [`subscriberdb_servicer`](https://github.com/magma/magma/blob/f9812c669427199c0cdfa107122b31ca3ef57c7b/lte/cloud/go/services/subscriberdb/servicers/southbound/subscriberdb_servicer.go#L222)).

sebathomas (approved):
Looking good, I think this will fix the issue, I just wonder if we can make it a bit more stable.

In the Python client on the gateway, we currently fetch the suci profiles first and skip all other sync calls if the suci call fails. See [client.py](https://github.com/magma/magma/blob/f9812c669427199c0cdfa107122b31ca3ef57c7b/lte/gateway/python/magma/subscriberdb/client.py#L113). Would it be better to do the suci call last? At the moment, the response doesn't seem to be used in the client.

sebathomas (approved):

lte/cloud/go/services/subscriberdb/load.go
moinuddin1980: Modified client.py to perform suci call at the last.

sebathomas (approved):

panyogesh (approved):
LGTM

PR #13061, opened 2022-06-23
[WIP] Try to downgrade ubuntu to fix CWAG failures

crasu:
Signed-off-by: Nils Semmelrock <nils.semmelrock@tngtech.com>

## Additional Information

- [ ] This change is backwards-breaking

PR #13084, opened 2022-06-26
fix(ci): update quotas and relax asserts in cwf integ tests in order to make them less flaky

crasu:
### ***Summary***
While #13061 simply tries to downgrade to the previous ubuntu version, this pr tries to address the causes of the cwf flakiness.

The general cause of the cwf integ tests seems to be timing. Ubuntu 20.04 seems to handle traffic generation a bit quicker(?) so the following test have become unstable:

**OmniRulesTest**: Fails because the session is terminated before GetPolicyUsage is called. Lowering the generated traffic seems to fix the issue. With Ubuntu 18.04 it worked because the session was terminated half a second slower. Why the termination happens at all is still unclear to me. The quota for the session shoud be sufficient.
**TestGyAbortSessionRequest**: Fails because the total allowed traffic in one request is limited to 5MB. So the session is already terminated before the ASR gets through. Why it is not requesting the last 3MB in another request is unclear to me.
**TestGxReAuthQuotaRefill**: Failed because the test now generates about 100kb more traffic. Increasing the buffer size fixes the issue.
**TestGxUsageReportEnforcement**: Fails because pcrf expectation is slightly to low, increased the acceptable delta in order to make the test more stable
**TestGyCreditExhaustionRestrict**: Adjust policy usage limits. Increased lower bound, because there seems to be delayed traffic from before the rules were installed. Increased upper bound to make sure there is enough volume left for delayed traffic.

Currently: **TestGxMidSessionRuleRemovalWithCCA_U** is still unstable I am looking into it but its hard to reproduce the failure locally.

Further more I checked the docker key into the repo cause the download was failing sometimes and fixed the ipv6 networking for the traffic server.

### *Testing*
I ran the entire testsuite about ~10 times locally and had no failed run. On the CI I had one failed run.


.github/workflows/cwf-integ-test.yml
```diff
@@ -12,6 +12,11 @@ on:  # yamllint disable-line rule:truthy
       - 'v1.*'
     types:
       - completed
+  pull_request:
+    branches:
+      - master
+      - 'v1.*'
+    types: [opened, reopened, synchronize]
```
uri200: I understand this is just temporal, right?

crasu: yes

nstng: Does this still needs to be removed?

.github/workflows/cwf-integ-test.yml
```diff
@@ -101,6 +106,7 @@ jobs:
           sudo mkdir -p /etc/vbox/
           sudo touch /etc/vbox/networks.conf
           sudo sh -c "echo '* 192.168.0.0/16' > /etc/vbox/networks.conf"
+          sudo sh -c "echo '* 3001::/64' >> /etc/vbox/networks.conf"
```
nstng, line 104: Why is this needed for cwf VMs? I was under the impression that this is an lte VM necessity. Ok, found it in the logs, the trfserver is using it - sry, missed that in the ipv6 review that introduced this change :/

cwf/gateway/integ_tests/gy_enforcement_test.go
```diff
@@ -589,7 +589,7 @@ func TestGyAbortSessionRequest(t *testing.T) {
 
 	err = setNewOCSConfig(
 		&fegprotos.OCSConfig{
-			MaxUsageOctets: &fegprotos.Octets{TotalOctets: ReAuthMaxUsageBytes},
+			MaxUsageOctets: &fegprotos.Octets{TotalOctets: 8 * MegaBytes}, //we generate more then 5Mbyte traffic, if this is set below 7MB this session will terminate before the ASR goes through
```
nstng, line 592: NIT: ReAuthMaxUsageBytes is now only used in gy_reauth_test.go. Could be moved there.

orc8r/tools/ansible/roles/docker/tasks/install_debian.yml
```diff
@@ -29,9 +29,14 @@
...
+    src: docker.key
+    dest: /tmp/docker.key
+
 - name: Add Docker’s official GPG key
   apt_key:
-    url: https://download.docker.com/linux/ubuntu/gpg
+    file: /tmp/docker.key
     state: present
 
 - name: Verify that you now have the key with the fingerprint
```
nstng: NIT: do we still need to verify vs a fingerprint if the key is checked in?

nstng (commented):
Cool findings! I would be interested in your debugging workflow (offline) :) I can not really reproduce this locally, because for me local runs were still green (ubuntu 20.04) - but I am fine with CI runs to verify this.
Feel free to ignore the NITs.

.github/workflows/cwf-integ-test.yml
crasu: Yes - I would like to remove this at the very last minute. Because without the cwf integ tests will not run on this branch.

orc8r/tools/ansible/roles/docker/tasks/install_debian.yml
crasu: done

nstng (approved):
ci changes lgtm (% remove of ci trigger before merge)

voisey:
> Cool findings! I would be interested in your debugging workflow (offline) :) I can not really reproduce this locally, because for me local runs were still green (ubuntu 20.04) - but I am fine with CI runs to verify this. Feel free to ignore the NITs.

@nstng For what it's worth, I was unable to get two tests -- TestGyAbortSessionRequest and TestOmnipresentRules -- to pass at all when I ran them locally on master. I've just run them each 20 times on this branch and they passed every time.

uri200 (approved):

PR #13109, opened 2022-06-28
chore(agw): Replace deprecated include with include_task

jheidbrink:
## Summary

Replace `include` with `include_tasks` in order to fix an Ansible deprecation warning

## Test Plan

I provisioned an Ubuntu EC2 instance configured to host the containerized AGW using the existing playbooks after making this change.

## Additional Information


mpfirrmann (approved):
LGTM. Thanks for cleaning :broom:

PR #13125, opened 2022-06-29
chore(feg): fix slack username for feg

crasu:
## Summary

Very small fix for the feg username in slack messages

## Test Plan

Did no test because the trigger only runs on master


maxhbr (approved):
LGTM

PR #13142, opened 2022-06-30
chore: fix misspelled error codes in protobuf

wolfseb:
Signed-off-by: Sebastian Wolf <sebastian.wolf@tngtech.com>

## Summary

Fix the incorrect spelling for the error codes `COMMAND_UNSUPPORTED`, `INVALID_HDR_BITS`, and `UNKNOWN_SERVING_NODE` in s6a_proxy and diam_errors. All usages of the error codes were fixed as well, which seem to only be for `COMMAND_UNSUPPORTED`. The corresponding *.pb.go were generated by running `./build.py --generate.`

## Test Plan

Run unit tests

## Additional Information


rsarwad (commented):
Changes look good. But posted one query

feg/protos/s6a_proxy.proto
```diff
@@ -52,14 +52,14 @@ enum ErrorCode {
...
-    COMMAND_UNSUPORTED       = 3001;
+    COMMAND_UNSUPPORTED      = 3001;
     UNABLE_TO_DELIVER        = 3002;
     REALM_NOT_SERVED         = 3003;
     TOO_BUSY                 = 3004;
     LOOP_DETECTED            = 3005;
     REDIRECT_INDICATION      = 3006;
     APPLICATION_UNSUPPORTED  = 3007;
-    INVALIDH_DR_BITS         = 3008;
+    INVALID_HDR_BITS         = 3008;
```
rsarwad, line 62: @ssanadhya, can we modify the proto field as it is not backward compatible. But error codes have fixed values.

sebathomas: I think it's no problem for binary serialization (see for example https://stackoverflow.com/a/50863982). I am not 100% sure that's what we use in this interface though.

rsarwad: As per my understanding field and values of proto needs to be same across various release versions.

ssanadhya: @rsarwad , this should be okay, since it is not changing an actual field in the proto message, but the name of an Enum value. As @sebathomas highlighted, the value 3008 will be sent on the wire with type as int before and after this change.

Further, since these are standard Diameter error codes, we would want to map 3008 to "Invalid header bits"

rsarwad: Thanks @ssanadhya  for clarification

sebathomas (approved):

voisey (approved):

voisey:
Hi @GANESH-WAVELABS, just in case you missed it, could you possibly take a look at this if you get the chance? Thanks!

crasu (approved):

ajahl (approved):
LGTM

PR #13144, opened 2022-06-30
chore(ci): Reuse the bazel build artifacts for the bazel tests

### LINKED ISSUE #13131: Comparison of bazel build times and remote cache usage in GH workflows
# Comparison of bazel build times and remote cache usage in GH workflows

## Motivation:

The goals of this analysis are:
- Analyze differences in build times depending on the structure of the GH workflow
- Impact of bazel remote caching options, e.g. `--remote_download_minimal`
- How to maximize reuse of build outputs and minimize the remote cache download volume and traffic

## Setup:

There are four different scenarios in this analysis:
|  Scenario  | Mounting  | Structure  | Flags  | GH actions URL |
|---|---|---|---|---|
|  1 |  None  | Seperate steps  |  `--remote_download_minimal` |  [URL](https://github.com/LKreutzer/magma/actions/runs/2582758255) |
|  2 |  None  | All in one step  |  `--remote_download_minimal`  |   [URL](https://github.com/LKreutzer/magma/actions/runs/2583543597) |
|  3 | Mount bazel output base  | Seperate steps  | `--remote_download_minimal`  |   [URL](https://github.com/LKreutzer/magma/actions/runs/2582704698) |
|  4 | Mount bazel output base |  Seperate steps  | None  |  [URL](https://github.com/LKreutzer/magma/actions/runs/2582719484)  |

- The "Mounting" column refers to a mounting of the bazel output base directory, which currently is at `/tmp/bazel`. This mounting - in principle - allows the reuse of output files for the following steps in github workflows.
- The "Structure" column refers to whether bazel commands are executed in separate steps in the workflow or within the same step.
- The "Flags" column refers to the bazel remote caching options that are set, in this case `--remote_download_minimal` .

### Results for the `bazel test //...` command after the `bazel build //...`

| Scenario  | Run time  | Remote cache hits  | Internal  |  Local  | Sandboxed  |
|---|---|---|---|---|---|
| 1  |  392.501s | 5475  |  1366 | 2  | 262  |
| 2  | 279.262s  |  229  | 0  | 2   |  267 |
|  3 | 425.362s  | 4538  | 0  | 2  | 267  |
| 4  |  366.891s | 1  | 0  | 2  | 267  |

When comparing scenarios 3 and 1 we find that mounting in combination with the `--remote_download_minimal` (or `--remote_download_toplevel`) flag does not work well, as most things will need to be loaded again. We do however see a slight improvement compared to not mounting at all.

Looking at scenario 4 we see that mounting without the remote cache flags works well and essentially all build outputs are reused in the following steps. The improvement in the build time is however small.

:question: Why are there remote cache hits in scenario 2? All the build outputs should be reused, as in scenario 4.

#### Results for the initial `bazel build //...`
If we compare how much is downloaded in the initial `bazel build //...` we find that scenario 4 is considerably slower in the build, which might indicate a download bottle-neck (?).
 
| Scenario  | Run time  | Remote cache hits  | Internal  |  Local  | Sandboxed  |
|---|---|---|---|---|---|
| 1  | 222.673s  | 5469  |  1456 | 0  |  0 |
| 2  | 430.521s   | 5469  |  1456 | 0  |  0 |
|  4 | 669.292s  | 5411  | 1798  | 0 | 58  |

:question: The build time in scenario 2 is varying between 270s - 450s, why? The build should be the same as in sc. 1.
Overall runtime for the sc. 2 workflow is ~15 min.

LKreutzer:

## Summary

- To maximize the reuse of build outputs in the bazel workflow the `bazel build ...`, `bazel test ...`, starlark format check and python module import checks are all done in the same step. See https://github.com/magma/magma/issues/13131 for an analysis of the expected impact and comparison to other approaches.
- The relevant commands in this workflow step are chained by `&&` to give the correct exit value for the action.
  - Run with test failure (on purpose): [action run URL](https://github.com/LKreutzer/magma/runs/7128581017?check_suite_focus=true)
  - If a command fails the logging persists and ends with the exit of the failing command.
- One of the drawbacks of this approach is that the logging becomes somewhat difficult to read.
  - To mitigate this some `printf` statements are added to separate the command outputs.
  - Because much of the bazel logging is to `stderr` the `printf` statements are redirected to `stderr` to keep them in the right position.
  - The newline character `\n` does not work in this situation and `\r` has to be used.
  - Note: There seems to be a bug in the GH actions which prints most logging three times while it is running and when the workflow finishes it only appears once. This behaviour does not originate from the changes made here and I have observe this behaviour also on the current master runs.

## Test Plan

- CI

## Additional Information

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -48,14 +48,14 @@ jobs:
               - '**/*.bazel'
               - '**/*.bzl'
 
-  bazel_build:
+  bazel_build_and_test:
     needs: path_filter
     # Only run workflow if this is a scheduled run on master branch,
     # or a pull_request that skip-duplicate-action wants to run again.
     if: |
       (github.event_name == `schedule` && github.ref == 'refs/heads/master')
         || ${{ needs.path_filter.outputs.files_changed == 'true' }}
-    name: Bazel Build Job `bazel build //...`
+    name: Bazel Build & Test Job
     runs-on: ubuntu-latest
     steps:
       - name: Maximize build space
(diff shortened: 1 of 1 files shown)


nstng (approved):
LGTM - reviewed and discussed offline in pairing

PR #13041, opened 2022-06-21
fix(orc8r): add reindexing of gateway subscriber state to subscriberdb indexer

alexzurbonsen:
## Summary

This PR adds capabilities to subscriberdb indexer to write state of type `gateway_subscriber_state` into the `gateway_subscriber_states` table created in #12949. The logic implemented there splits up the state into its entries and renews the entire state with every indexing run (old entries are deleted and new entries are written into `gateway_subscriber_states` table).

The second commit adds annotations such that the indexer is triggered for this type as well in docker and k8s deployments. 

This is the follow up PR to #12949 and implements step 3 of the plan outlined in [#8621 - comment](https://github.com/magma/magma/issues/8621#issuecomment-1144762649).

## Test Plan

- [x] Unit tests of subscriberdb: indexer and subscriber storage
- [x] manual test of orc8r deployed locally with docker compose

Manual testing is done with grpcurl and the following commands.

1) Send gateway state with three IMSIs.

```
grpcurl -vv \
    -insecure \
    -key /tmp/magma_protos/gateway.key \
    -cert /tmp/magma_protos/gateway.crt \
    -authority state-controller.magma.test \
    -protoset /tmp/magma_protos/out.protoset \
    -d '{"states": [{"type": "gateway_subscriber_state",
[...]

PR #12949, opened 2022-06-09
fix(orc8r): Add new table for reindexed subscriber state

sebathomas:
## Summary

Adds a new table to orc8r's subscriberdb that will contain subscriber state data from gateways after reindexing. The table has a combined primary key:

* network ID, gateway ID, imsi

The interface of the table is designed to allow updating all data for a certain gateway in order to remove stale subscriber entries. This implements step 2 of the plan outlined in https://github.com/magma/magma/issues/8621#issuecomment-1144762649. Follow-up PRs will add a reindexer that calls this interface every time a state update is received from a gateway.

## Test Plan

* Added unit tests for CRUD logic.
* Manual test of the new storage logic using PostgreSQL by calling the methods directly in subscriberdb.
gateway_subscriber_states_dbeaver

## Additional Information

- [ ] This change is backwards-breaking

Issue #8621, opened 2021-08-12
Sanitize stale session state in Orc8r


themarwhal:
Original discussion: https://github.com/magma/magma/discussions/7681

## Problem
As part of AGW state sent to Orc8r periodically, SessionD sends a snapshot of all existing sessions keyed by its IMSI.
As SessionD can only keep track of active sessions in the system, there is no way for SessionD to remove subscriber state in the Orc8r once a subscriber leaves the AGW. As Orc8r does not clean up stale state and NMS also does not check for stale state before displaying it, we end up with a growing number of stale subscribers that show active sessions.

## Possible solutions
1. SessionD sends a per-AGW state as a single object. This way, every state reported has the most updated snapshot of all subscribers. We will need to do indexing on the Orc8r side to enable per-subscriber queries from this state.
2. Orc8r periodically goes through all reported states to remove stale entries.

One caveat here is that with approach 1, we will still have stale state if an AGW goes offline. However, this may be OK if we also display the last checkin time to let the user know the state may be stale. 

## Actions
1. @themarwhal to add an additional SessionD state that encapsulates all subscribers in one object.
2. @hcgatewood to add any Orc8r modifications needed to support the same subscriber state API endpoint.
3. Once we have migrated to the new state, @themarwhal can remove per-subscriber state entries from SessionD's side.

ekowtaylor:
Phase out implementation and land design in this sprint. @themarwhal  to help break this down for tracking in this sprint

bhuvaneshne:
Hello @themarwhal , @hcgatewood ,
I was searching for existing bugs to report this exact issue when I found this thread.
Wanted to check how this is coming along.
Regards,
Bhuvanesh

sebathomas:
Concrete plan for solution 1:
1. Adapt sessiond's `OperationalStatesHandler` to produce a new kind of state with type `gateway_subscriber_state`. This will get the hardware ID as a key and then contain a map of all IMSIs to the corresponding subscriber state. This will ensure that there will also be a state update if the list of subscribers is empty. For the moment, sessiond can send both types of state until the changes on the orc8r side are finished.
2. Add a new table `gateway_subscriber_states` to subscriberdb that will contain the reindexed state. The table will have the following structure:

| Column  | Type |
| ------------- | ------------- |
| network_id  | text  |
| gateway_id  | text (hardware UUID)  |
| imsi | text |
| last_updated_at | int8 (unix timestamp) |
[...]

PR #13151, opened 2022-06-30
fix(orc8r): Read from subscriber storage by IMSI

alexzurbonsen:
⚠️ Merge after #13212

## Summary

In this PR we enable NMS to read `subscriber_state` from the new `gateway_subscriber_states` table (#12949). It follows up on #13041 and implements step 4 from [#8621 - comment](https://github.com/magma/magma/issues/8621#issuecomment-1144762649).

We adapt the obsidian handlers of subscriberdb to read all states of type `lte.SubscriberStateType = 'subscriber_state'`
from the `gateway_subscriber_states` table instead of the `states` table. To that end we add a new method to the SubscriberStorage interface, which enables to read state by network ID and imsi (`GetSubscribersForIMSIs`). Accordingly we also add an index for network ID and imsi to the `gateway_subscriber_states` table.

The handlers that are getting `subscriber_state`, are now hard coded to always get it the new interface method by calling the `getSubscribersForIMSIs` function.

Tests are adapted accordingly.

## Test Plan

- [x] Check in postgres database that index was created.
- [x] run handlers_test.go
- [x] Test manually in NMS whether all relevant routes are loading state correctly
         - /magma/v1/lte/:network_id/subscribers (called in subscribers tab)
         - /magma/v1/lte/:network_id/subscribers/:subscriber_id (view of single subscriber)
         - /magma/v1/lte/:network_id/subscriber_state (called in subscribers tab)
         - /magma/v1/lte/:network_id/subscriber_state/:subscriber_id (called if autorefresh in single subscriber view is activated)
- [x] e2e test with simple integration test

## Additional Information

- Done in pairing with @sebathomas.

### DIFF
--- a/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
+++ b/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
@@ -34,6 +34,7 @@ import (
 	policydbmodels "magma/lte/cloud/go/services/policydb/obsidian/models"
 	"magma/lte/cloud/go/services/subscriberdb"
 	subscribermodels "magma/lte/cloud/go/services/subscriberdb/obsidian/models"
+	subscriberstorage "magma/lte/cloud/go/services/subscriberdb/storage"
 	"magma/orc8r/cloud/go/orc8r"
 	"magma/orc8r/cloud/go/serde"
 	"magma/orc8r/cloud/go/services/configurator"
@@ -65,16 +66,16 @@ const (
 	ParamVerbose   = "verbose"
 )
 
-func GetHandlers() []obsidian.Handler {
+func GetHandlers(subscriberStorage subscriberstorage.SubscriberStorage) []obsidian.Handler {
 	ret := []obsidian.Handler{
-		{Path: ListSubscribersPath, Methods: obsidian.GET, HandlerFunc: listSubscribersHandler},
+		{Path: ListSubscribersPath, Methods: obsidian.GET, HandlerFunc: getListSubscribersHandler(subscriberStorage)},
 		{Path: ListSubscribersPath, Methods: obsidian.POST, HandlerFunc: createSubscribersHandler},
-		{Path: ManageSubscriberPath, Methods: obsidian.GET, HandlerFunc: getSubscriberHandler},
+		{Path: ManageSubscriberPath, Methods: obsidian.GET, HandlerFunc: getSubscriberHandler(subscriberStorage)},
 		{Path: ManageSubscriberPath, Methods: obsidian.PUT, HandlerFunc: updateSubscriberHandler},
 		{Path: ManageSubscriberPath, Methods: obsidian.DELETE, HandlerFunc: deleteSubscriberHandler},
 
-		{Path: ListSubscriberStatePath, Methods: obsidian.GET, HandlerFunc: listSubscriberStateHandler},
-		{Path: ManageSubscriberStatePath, Methods: obsidian.GET, HandlerFunc: getSubscriberStateHandler},
+		{Path: ListSubscriberStatePath, Methods: obsidian.GET, HandlerFunc: getListSubscriberStateHandler(subscriberStorage)},
+		{Path: ManageSubscriberStatePath, Methods: obsidian.GET, HandlerFunc: getSubscriberStateHandler(subscriberStorage)},
 
 		{Path: ActivateSubscriberPath, Methods: obsidian.POST, HandlerFunc: makeSubscriberStateHandler(subscribermodels.LteSubscriptionStateACTIVE)},
 		{Path: DeactivateSubscriberPath, Methods: obsidian.POST, HandlerFunc: makeSubscriberStateHandler(subscribermodels.LteSubscriptionStateINACTIVE)},
@@ -111,7 +112,6 @@ var subscriberStateTypesKeyedByIMSI = []string{
 	lte.MMEStateType,
 	lte.S1APStateType,
 	lte.SPGWStateType,
-	lte.SubscriberStateType,
 	orc8r.DirectoryRecordType,
 }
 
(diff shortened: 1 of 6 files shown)


lte/cloud/go/services/subscriberdb/storage/subscriber_state_test.go
```diff
@@ -101,4 +109,90 @@ func TestSubscriberStorage(t *testing.T) {
...
+		foundStates, err = s.GetSubscribersForIMSIs(nid1, selectedIMSIs)
+		assert.NoError(t, err)
+		assert.True(t, cmp.Equal(foundStates, subscriberStatesGw2.Subscribers))
+		active_duration_sec_found := foundStates["IMSI001010000000123"]["magma.ipv4"].([]interface{})[0].(map[string]interface{})["active_duration_sec"]
+		assert.Equal(t, active_duration_sec_found, float64(1))
+	})
+}
+
+func (ss *subscriberStorage) setTestGatewaySubscriberStateWithIncreasingTimestamp(networkID string, gatewayID string, subscriberStates *GatewaySubscriberState) error {
+	timeSec := clock.Now()
```
alexzurbonsen: Currently this method is only used to set the states in the final case of the test, the duplicate IMSIs. It uses system time (as `SetAllSubscribersForGateway`) and adds one second for every entry.
I.e. the first entry of `double_IMSI` is registered at system time (via `SetAllSubscribersForGateway`), the duplicate is registered at system time + 2 Seconds (via this method, `setTestGatewaySubscriberStateWithIncreasingTimestamp`).

More work, but may be less confusing could be to fix the time and set all states via this method? What do you think?

lte/cloud/go/services/subscriberdb/storage/subscriber_state.go
```diff
@@ -182,6 +195,57 @@ func (ss *subscriberStorage) DeleteSubscribersForGateway(networkID string, gatew
...
+			err = rows.Scan(&imsi, &rawState, &lastUpdated)
+			if err != nil {
+				return nil, fmt.Errorf("GetSubscribersForIMSIs, SQL row scan error: %w", err)
+			}
+			reportedState := make(state.ArbitraryJSON)
+			err = json.Unmarshal(rawState, &reportedState)
+			if err != nil {
+				return nil, fmt.Errorf("GetSubscribersForIMSIs, error unmarshaling state: %w", err)
+			}
+			if lastUpdated > lastUpdates[imsi] {
```
alexzurbonsen: What do we do in the = case? The compared times are given in seconds. I am not sure whether it is realistic to have an IMSI registering with two gateways within one second. But maybe that is an edge case we should take care of?

sebathomas: In the current implementation, we will randomly pick the state that is read first. I think that is fine. The conflict will only be temporary and should be solved when we get an update from one of the AGWs.

lte/cloud/go/services/subscriberdb/storage/subscriber_state_test.go
sebathomas: Yes, I think it would be clearer to just give them explicit times and not rely on the system time + x seconds, and to create both states by the same method.

lte/cloud/go/services/subscriberdb/storage/subscriber_state_test.go
```diff
@@ -101,4 +107,90 @@ func TestSubscriberStorage(t *testing.T) {
...
+		}
+		foundStates, err = s.GetSubscribersForIMSIs(nid1, selectedIMSIs)
+		assert.NoError(t, err)
+		assert.True(t, cmp.Equal(foundStates, expectedStates))
+
+		// get latest entry for IMSI appearing in two gateways
+		// the order of the imsis in imsisGw2 is important. The target
+		// doubled imsi "IMSI001010000000123", has to be the second entry,
+		// since its two entries are distinguished by the value of
+		// "active_duration_sec", which wouldn't differ otherwise
```
sebathomas: Maybe you can do something else to ensure they are different (e.g. modify the states after creation), and then assert on the entire state instead of only comparing `active_duration_sec`.

alexzurbonsen: Sounds reasonable.

lte/cloud/go/services/subscriberdb/storage/subscriber_state_test.go
```diff
@@ -101,4 +107,86 @@ func TestSubscriberStorage(t *testing.T) {
...
+		assert.NoError(t, err)
+		foundStates, err = s.GetSubscribersForIMSIs(nid1, selectedIMSIs)
+		assert.NoError(t, err)
+		assert.True(t, cmp.Equal(foundStates, subscriberStatesGw2.Subscribers))
+	})
+}
+
+func (ss *subscriberStorage) setTestGatewaySubscriberStateWithIncreasingTimestamp(t *testing.T, networkID string, gatewayID string, subscriberStates *GatewaySubscriberState, frozenTime int64) error {
+	clock.SetAndFreezeClock(t, time.Unix(frozenTime, 0))
+	defer clock.UnfreezeClock(t)
```
sebathomas: Can't you just put the ints into the database without setting or reading the clock?

alexzurbonsen: Oh, right. Scusi. Didn't think of that.

lte/cloud/go/services/subscriberdb/storage/subscriber_state_test.go
```diff
@@ -101,4 +105,82 @@ func TestSubscriberStorage(t *testing.T) {
 		err = s.DeleteSubscribersForGateway(nid1, gwid2)
 		assert.NoError(t, err)
 	})
+
+	t.Run("Test getting subscriber states by imsi and getting all subscribers", func(t *testing.T) {
+
```
sebathomas: Nitpick: unneeded empty line

lte/cloud/go/services/subscriberdb/storage/subscriber_state_test.go
```diff
@@ -14,10 +14,14 @@
...
 
 import (
+	"database/sql"
+	"fmt"
 	"testing"
 
+	"github.com/Masterminds/squirrel"
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
+	"golang.org/x/exp/maps"
```
sebathomas: I guess this leads to all the go.mod changes? :(

lte/cloud/go/services/subscriberdb/storage/subscriber_state_test.go
```diff
@@ -101,4 +105,82 @@ func TestSubscriberStorage(t *testing.T) {
...
+		frozenTime = 200000
+		err = s.(*subscriberStorage).setTestGatewaySubscriberStateWithIncreasingTimestamp(t, nid1, gwid2, &subscriberStatesGw2, frozenTime)
+		assert.NoError(t, err)
+		foundStates, err = s.GetSubscribersForIMSIs(nid1, selectedIMSIs)
+		assert.NoError(t, err)
+		assert.True(t, cmp.Equal(foundStates, subscriberStatesGw2.Subscribers))
+	})
+}
+
+func (ss *subscriberStorage) setTestGatewaySubscriberStateWithIncreasingTimestamp(t *testing.T, networkID string, gatewayID string, subscriberStates *GatewaySubscriberState, frozenTime int64) error {
```
sebathomas: I think the name of this function is based on an old version, don't see the increasing timestamp any more.

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
```diff
@@ -2316,3 +2358,14 @@ func newPolicy(id string) *policydbModels.PolicyRule {
 	}
 	return policy
 }
+
+// newTestSubscriberStorage sets up the gateway_subscriber_states table, without
+// spinning up the entire subscriberdb Test Service, which could be done
+// alternatively.
```
sebathomas:
```suggestion
// spinning up the entire subscriberdb Test Service.
```

lte/cloud/go/services/subscriberdb/storage/subscriber_state.go
```diff
@@ -182,6 +195,58 @@ func (ss *subscriberStorage) DeleteSubscribersForGateway(networkID string, gatew
...
+				lastUpdates[imsi] = lastUpdated
+				mappings[imsi] = reportedState
+			}
+		}
+		err = rows.Err()
+		if err != nil {
+			return nil, fmt.Errorf("GetSubscribersForIMSIs, SQL rows error: %w", err)
+		}
```
sebathomas: Idea: Maybe this could be extracted to a new method `convertRowsToMap` (or similar).

lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers.go
```diff
@@ -901,6 +918,20 @@ func mergeStates(s1 state_types.StatesByID, s2 state_types.StatesByID) state_typ
 	return s1
 }
 
+func getSubscribersForIMSIs(networkID string, imsis []string, subscriberStorage subscriberstorage.SubscriberStorage) (state_types.StatesByID, error) {
```
sebathomas: Maybe a better name:
```suggestion
func getSubscriberStatesForIMSIs(networkID string, imsis []string, subscriberStorage subscriberstorage.SubscriberStorage) (state_types.StatesByID, error) {
```

sebathomas (commented):
Looking good, just some minor points.

lte/cloud/go/services/subscriberdb/storage/subscriber_state_test.go
alexzurbonsen: Yes... it is just used to merge two maps. Could also write our own method for that?

sebathomas (approved):

PR #13170, opened 2022-07-01
fix(dp): add filtering logs by networkId

WojSad:
fix(dp): add filtering logs by networkId

## Summary

enabled filtering dp logs by network id

## Test Plan
Tests amended


dp/cloud/go/services/dp/obsidian/log/handlers.go
```diff
@@ -160,6 +161,9 @@ func (r *ListLogsRequest) toElasticSearchQuery() *elastic.BoolQuery {
 	if responseCode != nil {
 		matchQueries = append(matchQueries, elastic.NewMatchQuery("response_code", responseCode))
 	}
+	if networkId != "" {
```
jkmar: network Id should always be in the filter (we shouldn't have this check)

jkmar (approved):

Issue #13086, opened 2022-06-27
Memory leak found for network initiated detach procedure


rsarwad:
### Your Environment

- **Version: Master code base
- **Affected Component: Access Gateway
- **Affected Subcomponent: MME service 
- **Deployment Environment: Local Vagrant setup
### Describe the Issue
Memory leak is observed while executing s1ap sanity test suite. From trace it appears that the leak is observed for network initiated detach procedure. Below is the back trace 

Jun 15 12:03:23 magma-dev-focal mme[1182337]: =================================================================
Jun 15 12:03:23 magma-dev-focal mme[1182337]: ==1182337==ERROR: LeakSanitizer: detected memory leaks
Jun 15 12:03:23 magma-dev-focal mme[1182337]: Direct leak of 120 byte(s) in 1 object(s) allocated from:
Jun 15 12:03:23 magma-dev-focal mme[1182337]:     #0 0x7f5f1e5c7dc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
Jun 15 12:03:23 magma-dev-focal mme[1182337]:     #1 0x55e96891e3f4 in mme_app_handle_nw_init_ded_bearer_actv_req /home/vagrant/magma/lte/gateway/c/core/oai/tasks/mme_app/mme_app_bearer.c:3207
Jun 15 12:03:23 magma-dev-focal mme[1182337]:     #2 0x55e9688dbeef in handle_message /home/vagrant/magma/lte/gateway/c/core/oai/tasks/mme_app/mme_app_main.c:372
Jun 15 12:03:23 magma-dev-focal mme[1182337]:     #3 0x7f5f1d7ca5c6 in zloop_start (/lib/x86_64-linux-gnu/libczmq.so.4+0x295c6)
Jun 15 12:03:23 magma-dev-focal mme[1182337]: Direct leak of 48 byte(s) in 1 object(s) allocated from:
Jun 15 12:03:23 magma-dev-focal mme[1182337]:     #0 0x7f5f1e5c7dc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
Jun 15 12:03:23 magma-dev-focal mme[1182337]:     #1 0x7f5f1d7c5c0f in zlist_new (/lib/x86_64-linux-gnu/libczmq.so.4+0x24c0f)
[...]

PR #13173, opened 2022-07-04
fix(agw): Fixed the issue, 13086

rsarwad:
## Summary
Provides the fixes for the issue, https://github.com/magma/magma/issues/13086
Fixed the memory leak found for below scenario, recreated using test case, test_dedicated_bearer_activation_idle_mode_paging_timer_expiry
1. Dedicated bearer establishment procedure is initiated while UE is in idle mode. So first MME does Paging, bring the UE to Connected state and then send Activate Dedicated Bearer Request. 
2. Test case, test_dedicated_bearer_activation_idle_mode_paging_timer_expiry simulates paging timer expiry. So UE doesn't respond to Paging. Then UE sends Detach Request. As part of Detach MME needs to clear all contexts. But pending dedicated bearer was not freed. Fixed the leak

## Test Plan


pruthvihebbani (approved):
lgtm

voisey:
@rsarwad Just in case you aren't aware, you can rebase to get the new required job (nms-typescript) to run

PR #13196, opened 2022-07-05
fix(ci): full path required for misspell excludes

Neudrino:
It seems, as if excludes for misspell need fully qualified path names.
The documentation does not specify this explicitly, but [the examples also seem to indicate it](https://github.com/reviewdog/action-misspell#exclude).

Log from job

### DIFF
--- a/.github/workflows/reviewdog-workflow.yml
+++ b/.github/workflows/reviewdog-workflow.yml
@@ -172,7 +172,7 @@ jobs:
           reporter: github-pr-review
           locale: "US"
           exclude: |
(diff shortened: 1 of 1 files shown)


voisey (approved):

nstng (approved):
lgtm, tested via find
```
nils@nilstng:~/magma/nms$ find . -not -path yarn.lock -name "*.lock"
./yarn.lock
nils@nilstng:~/magma/nms$ find . -not -path ./yarn.lock -name "*.lock"
# no result
```

PR #13197, opened 2022-07-05
feat(ci): create cleanup GH workflow

### LINKED ISSUE #13199: Cleanup Github actions tab
### Describe the Issue

- **Current status:**
The [Github Actions tab](https://github.com/magma/magma/actions) is untidy and cluttered by a lot of workflows, which are outdated, abandoned, only used on forks, only used on branches, only used short term for testing, always broken, etc.
That makes it difficult to find the correct job, if searching for a specific workflow or action.
- **Actions from forks:**
Anyone can implement a new action on his fork. Given that this individual also creates a PR against the Magma repository, those newly defined workflows will also show up in the Github Actions overview. The current development policy does not allow to restrict this behaviour. Thus one will have to live with new workflows poping up once in a while and becoming outdated at some point.
- **Actions from branches:**
Same as for forks, but with the addition, that branches on Magma were once in the past the way to develop but are not any more. Still there exist some and contributers do in principle have the ability to create new branches.

### Describe the Target State

The [Github Actions tab](https://github.com/magma/magma/actions) should be tidy. This can be rephrased as
- Unused workflows should be removed.
  - Workflows are seen as unused after **90 days of inactivity**.
  - Inactivity means that a specific workflow had no execution in a time interval.
- A cleanup process should be implemented. The cleanup process should be automatic.

### Additional information

- The workflows themselves cannot be removed. They could only be disabled.
- Any workflow run can be removed.
- A workflow with no workflow runs will be removed from the display in Github Actions automatically. (Or they will not be displayed in first place, if they did never run.)

Neudrino:
... to cleanup the mess in the Github actions tab.

Using [Delete Workflow Runs](https://github.com/marketplace/actions/delete-workflow-runs) Github Action.

`keep_minimum_runs` needs to be set to `0` to also clean up those workflow runs which were generated by some manual tests from forks or outdated workflows from old releases.

Fixes https://github.com/magma/magma/issues/13199.

### DIFF
--- a/.github/workflows/cleanup-old-workflows.yml
+++ b/.github/workflows/cleanup-old-workflows.yml
@@ -1,0 +1,26 @@
+# Copyright 2022 The Magma Authors.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "Cleanup Github Workflows"
+
+on:
+  schedule:
+    - cron: '42 * * * *'
+
+jobs:
+  cleanup-workflows:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Mattraks/delete-workflow-runs@c1a2f0de3fa4134a2357de83c92acff6e298bb8a # pin@v2.0.3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          retain_days: 90
+          keep_minimum_runs: 0


.github/workflows/ci-cleanup-old-workflow-runs.yml
```diff
@@ -0,0 +1,16 @@
+---
+name: "CI Cleanup workflow runs"
+
+on:
```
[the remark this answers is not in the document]
Neudrino: :arrow_right: https://github.com/magma/magma/pull/12990/files#r903505808

.github/workflows/ci-cleanup-old-workflow-runs.yml
```diff
@@ -0,0 +1,16 @@
+---
```
maxhbr: missing license header

.github/workflows/ci-cleanup-old-workflow-runs.yml
```diff
@@ -0,0 +1,16 @@
...
+    - cron: '42 6 * * 0'
+
+jobs:
+  ci-workflow-cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dmvict/clean-workflow-runs@9b0fba98c68369add0884d6af67e6866578cbf98 # pin@v1.0.0
+        with:
+          token: ${{ secrets.PRIVATE_GITHUB_TOKEN }}
+          save_period: 30
```
voisey: I like the idea of the PR but was just wondering whether `save_period: 30` might be overly restrictive. Perhaps there are some situations when one would like to look back at runs from over a month ago? I'm not super against 30, but I would feel more comfortable with, say, 60 or 90.

Neudrino: Do you have any concrete example, where a longer history was actually necessary?
I would rather go for the 30 days now and only adopt this to a longer period, on the occasion of a specific need rather than on a vague feeling.

voisey: My feeling was that it was useful to look back at what the state of the CI was before it broke recently, and in particular it took over 30 days to fix.

.github/workflows/ci-cleanup-old-workflow-runs.yml
Neudrino: The headers are missing in all workflow files. To track the effort of fixing this everywhere, I created https://github.com/magma/magma/issues/13280 to do it in a separate PR in bulk.

.github/workflows/ci-cleanup-old-workflow-runs.yml
```diff
@@ -0,0 +1,27 @@
...
+  schedule:
+    - cron: '42 * * * *'
+
+jobs:
+  cleanup-workflows:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Mattraks/delete-workflow-runs@c1a2f0de3fa4134a2357de83c92acff6e298bb8a # pin@v2.0.3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
voisey: According to the [documentation](https://github.com/Mattraks/delete-workflow-runs#2-repository), `repository` is also a required input. Or is there a reason it isn't required here?

Neudrino: Yes, is required, however [it has a default](https://github.com/Mattraks/delete-workflow-runs#default--githubrepository-), which is fine for us.

voisey: Oops, I somehow overlooked this.

voisey (approved):

.github/workflows/cleanup-old-workflows.yml
```diff
@@ -0,0 +1,26 @@
...
+    - cron: '42 * * * *'
+
+jobs:
+  cleanup-workflows:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Mattraks/delete-workflow-runs@c1a2f0de3fa4134a2357de83c92acff6e298bb8a # pin@v2.0.3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          retain_days: 90
```
jheidbrink, line 25: While I like 90 days better than 30 days, I would still prefer to keep workflow runs for a longer time, like a year. I think we should also raise that question in Slack as it's strictly speaking a breaking change, and we shouldn't surprise people with this. I assume not everybody with a stake in this is looking at this review.

.github/workflows/cleanup-old-workflows.yml
```diff
@@ -0,0 +1,26 @@
...
+
+on:
+  schedule:
+    - cron: '42 * * * *'
+
+jobs:
+  cleanup-workflows:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Mattraks/delete-workflow-runs@c1a2f0de3fa4134a2357de83c92acff6e298bb8a # pin@v2.0.3
```
jheidbrink, line 22: How does this behave with rate-limiting by the Github API? Does it slow down when encountering HTTP 429s? Might this break other workflows that call the Github API in case different runners share a request rate limit?

.github/workflows/cleanup-old-workflows.yml
Neudrino: A public post in Slack was made: https://magmacore.slack.com/archives/C018J8UMGMR/p1661418380688709.

.github/workflows/cleanup-old-workflows.yml
Neudrino: From the [respective part of the documentation](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-github-actions) I read, that there is a limit of 1000 requests per hour for usage of the `GITHUB_TOKEN`.
While there is a way to [inspect the rate limit for access with a _personal_ access token](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#checking-your-rate-limit-status), I think this inspection method does not apply for the `GITHUB_TOKEN` as it is not a personal access token.

That being said towards the general topic of rate limits, I did test this workflow, and had no issue with rate limits whatsoever. There was no impact on other workflows observed during (or after) the test.
It should be mentioned that the execution policy was adopted to minimize the impact on rate limits. I.e. the execution was set to hourly, such that every hour only a view requests are made, and they will not matter with the hourly reset of the quota.

.github/workflows/cleanup-old-workflows.yml
LKreutzer: I also think that it would be nice to be able to analyse how build times etc. changed within the last year - at least for the main workflows, incl. `bazel.yml`. For workflows that are not actively used all runs can be removed of course.

Neudrino:
Closing as per argument in https://github.com/magma/magma/issues/13199#issuecomment-1243723508.

Issue #13102, opened 2022-06-28
Bazel commands only work in the magma directory


vktng:
Since #12987 was merged, the bazel commands will only work in the magma folder, because the cache setup was reworked to use a relative path: `bazel/bazelrcs/cache.bazelrc`

LKreutzer:
This also causes some make targets to fail (such as `build_c`) because they invoke bazel form `lte/gateway`.
```
vagrant@magma-dev-focal:~/magma/lte/gateway$ make build_c
BAZEL_FLAGS 
MAIN_FEATURE agw_of
OAI_FLAGS -DS6A_OVER_GRPC=True -DEMBEDDED_SGW=True
bazel build  //lte/gateway/c/session_manager:sessiond //lte/gateway/c/sctpd/src:sctpd //lte/gateway/c/connection_tracker/src:connectiond //lte/gateway/c/li_agent/src:liagentd
[INFO 09:42:48.881 src/main/cpp/option_processor.cc:407] Looking for the following rc files: /etc/bazel.bazelrc,/home/vagrant/magma/.bazelrc,/home/vagrant/.bazelrc
[INFO 09:42:48.881 src/main/cpp/rc_file.cc:56] Parsing the RcFile /home/vagrant/magma/.bazelrc
[INFO 09:42:48.881 src/main/cpp/rc_file.cc:56] Parsing the RcFile bazel/bazelrcs/cache.bazelrc
[FATAL 09:42:48.881 src/main/cpp/blaze.cc:1295] Unexpected error reading .blazerc file 'bazel/bazelrcs/cache.bazelrc'
make: *** [Makefile:154: build_c] Error 37
```

PR #13200, opened 2022-07-05
chore: remove experimental_strict_action_env option from bazelrc

LKreutzer:

## Summary

- The bazel option `--experimental_strict_action_env` is removed from the `.bazelrc`
 - This option was introduced in https://github.com/magma/magma/pull/10642 to wrap some bazel commands in Make targets
   - The Make targets do not seem to be used in CI but may be convenient for local development

## Test Plan

- CI
- Run the Make targets `build_c`, `test_c` and `build_oai_clang`.
  - The Make calls need to happen in `magma/lte/gateway`, however there is currently a bug (see issue https://github.com/magma/magma/issues/13102) which requires a slight modification of the `.bazelrc` for testing.
  - The line `import bazel/bazelrcs/cache.bazelrc` needs to be replaced by `import /home/vagrant/magma/bazel/bazelrcs/cache.bazelrc` for testing inside the magma-vm.
  - Builds are successful and no change in behaviour is observed compared to the run with the flag.

## Additional Information
- The initial run of the bazel workflow is very slow on this PR because the removal of the flag prevents most cache hits. ~~Once the cache is filled on master the build time should be unaffected.~~ 
   - I filled the cache from the fork and now the workflow has a normal runtime. 
- The option has been renamed to `incompatible_strict_action_env` (see https://github.com/bazelbuild/bazel/issues/6648)
- The option has the following description in the current [docs](https://bazel.build/reference/command-line-reference):
Screenshot from 2022-07-05 16-35-06
  - Since we have a relatively reproducible build environment in CI it should not affect the cache hits once they are generated without this flag.
  - :warning: It might be good to monitor the remote cache hits in the weeks following a potential merge of this PR.
- We specify `build:asan --action_env=ASAN_OPTIONS=detect_leaks=1:color=always` in the current `.bazelrc`

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -9,7 +9,6 @@ build:production --config=lsan --strip=never --copt=-O3
 
 # C/C++ CONFIGS
 build --cxxopt=-std=c++14
-build --experimental_strict_action_env
 build --compilation_mode=dbg
 
 # DEFAULT TEST CONFIGURATION


nstng (approved):
LGTM - discussed offline

PR #12938, opened 2022-06-08
chore(agw): python syntax is adapted

mpfirrmann:
## Summary
Some changes in the Python syntax after upgrading to 3.8 on `magma_test`. 
This PR is also related to #12435 and #12673.

## Test Plan

S1AP integ tests were executed together with the VM upgrade.

## Additional Information

- [ ] This change is backwards-breaking

PR #13202, opened 2022-07-05
Revert "chore(agw): python syntax is adapted"

mpfirrmann:
Reverts magma/magma#12938

This code is used by the `magma_trfserver` as well, which runs with python 3.5 on debian stretch. Python >3.6 is needed for this code to work.

### DIFF
--- a/lte/gateway/python/integ_tests/common/service303_utils.py
+++ b/lte/gateway/python/integ_tests/common/service303_utils.py
@@ -21,13 +21,13 @@ from orc8r.protos.common_pb2 import Void
 from orc8r.protos.service303_pb2 import ServiceInfo
 from orc8r.protos.service303_pb2_grpc import Service303Stub
 
-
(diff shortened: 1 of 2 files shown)


ssanadhya (approved):
Lgtm, @mpfirrmann , can we plan to upgrade the python version on `magma_trfserver` VM too?

VinashakAnkitAman:
> Lgtm, @mpfirrmann , can we plan to upgrade the python version on `magma_trfserver` VM too?

Hi @ssanadhya I think it would be better to migrate TRF Server VM to Ubuntu 20 if possible to keep the software versions in sync with Magma-Dev and Magma-Test VMs because the other 2 VMs are already in Ubuntu 20.04 version..

mpfirrmann:
Agree with @VinashakAnkitAman. Upgrading all VMs to Ubuntu focal would probably also be easier and more beneficial.

PR #13210, opened 2022-07-06
chore(dp): Use magma artifactory in DP helm chart

sklgromek:

## Summary
Currently, the Domain-proxy helm chart doesn't point to any external registry, so it will be not possible to install it, if images are not available locally. Rather than the Magma artifactory should be used here

## Test plan
- [x] Take care to put images in Magma artifactory
- [x] Try to install the domain-proxy helm chart


jkmar (approved):

xbend (approved):
LGTM

PR #13223, opened 2022-07-06
chore: bump eventlet

lucasgonze:
Fixes a long-standing work item, https://github.com/magma/magma/security/dependabot/2

This upgrade fell in the cracks about a year ago.

Signed-off-by: Lucas Gonze <lucas@gonze.com>

## Summary

 Bump eventlet from 0.30.2 to 0.31.0 in /lte/gateway/python

## Test Plan

CI tests

## Additional Information

### DIFF
--- a/lte/gateway/python/setup.py
+++ b/lte/gateway/python/setup.py
@@ -119,7 +119,7 @@ setup(
         'idna==2.8',
         'python-dateutil==2.8.1',
         'six>=1.12.0',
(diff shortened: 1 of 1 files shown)


lucasgonze:
There is not yet a version of eventlet that both works and has the security patch we need. We are waiting on a tagged release for this PR: https://github.com/benoitc/gunicorn/pull/2581

PR #13231, opened 2022-07-07
refactor(nms): remove refreshingContext for EnodebContext

### LINKED ISSUE #13218: Replace useRefreshingContext with EnodebContext.refetch
* Define `refetch` on `EnodebContext `
* Implement  `refetch` in `EnodebContext `. It should fetch all enodebs or a single enodeb if an ID is provided and update the state.
* Replace
```
useRefreshingContext({
    context: EnodebContext,
   ...
```
with
```   
  useInterval(() => ctx.refetch(...), refresh ? REFRESH_INTERVAL : null);
``` 
and use the `EnodebContext ` instead of the state from `useRefreshingContext`. Also replace the existing usage of `useInterval( ...  FetchEnodebs`.

Neudrino:
## Summary

- Fixes #13218 

## Test Plan

- CI
- Manual testing, that the requests to refetch are executed in **three** different panels in the UI

### DIFF
--- a/nms/app/components/lte/LteContext.tsx
+++ b/nms/app/components/lte/LteContext.tsx
@@ -44,7 +44,6 @@ import type {
   Tier,
 } from '../../../generated';
 import type {EnodebInfo} from './EnodebUtils';
-import type {EnodebState} from '../context/EnodebContext';
 import type {gatewayPoolsStateType} from '../context/GatewayPoolsContext';
 
 import * as cbsdState from '../../state/lte/CbsdState';
@@ -56,6 +55,7 @@ import {
   SubscriberId,
 } from '../../../shared/types/network';
 import {
+  FetchEnodebs,
   FetchGateways,
   InitEnodeState,
   InitGatewayPoolState,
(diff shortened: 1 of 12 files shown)


nms/app/views/equipment/EquipmentEnodeb.tsx
```diff
@@ -241,25 +241,15 @@ type EnodebRowType = {
...
   const [refresh, setRefresh] = useState(true);
   const [currRow, setCurrRow] = useState<EnodebRowType>({} as EnodebRowType);
-  const params = useParams();
-  const networkId: string = nullthrows(params.networkId);
-  const REFRESH_INTERVAL = 30000;
-  useInterval(
-    () => {
-      void FetchEnodebs({networkId, enqueueSnackbar}).then(enodebs => {
-        if (enodebs) {
-          void ctx.setState('', undefined, {enbInfo: enodebs});
```
thmsschmitt, line 254: We should also be able to remove this bit
```
  if (newState) {
    setEnbInfo(newState.enbInfo);
    return;
  }
```
from `SetEnodebState`

nms/app/components/lte/LteContext.tsx
```diff
@@ -326,6 +327,19 @@ export function EnodebContextProvider(props: Props) {
...
           });
         },
+        refetch: id => {
+          void FetchEnodebs({
+            id: id,
+            networkId,
+            enqueueSnackbar,
+          }).then(enodebs => {
+            if (enodebs) {
+              setEnbInfo(enodebState =>
```
thmsschmitt, line 348: Nit: a bit clearer would maybe be 
```
setEnbInfo(currentEnodebs => id ? {...currentEnodebs, ...enodebs} : enodebs)
```

thmsschmitt (changes requested):
Looks good, I left some review comments.
And some `jest.spyOn(hooks, 'useRefreshingContext')` need to be replaced see https://github.com/magma/magma/runs/7235675529?check_suite_focus=true

thmsschmitt (approved):

PR #13232, opened 2022-07-07
docs(orc8r): Fixes juju action name in how-to guide

saltiyazan:
## Summary

fixes action name in the how-to guide for deploying with juju


Neudrino (approved):
Watch out for the DCO check, please.

PR #13233, opened 2022-07-07
refactor(nms): Replace useRefreshingContext with SubscriberContext.refetch

### LINKED ISSUE #13216: Replace useRefreshingContext with SubscriberContext.refetchSessionState
* Define `refetchSessionState ` on `SubscriberContextType`
* Implement  `refetchSessionState` in ` LteContext`. It should fetch the session state and update the state.
* Replace
```
useRefreshingContext({
    context: SubscriberContext,
   ...
```
with
```   
  useInterval(() => ctx.refetchSessionState(), refresh ? REFRESH_INTERVAL : null);
``` 
and use the  `SubscriberContext` instead of the state from `useRefreshingContext`.

HannaFar:
## Summary

- Fixes https://github.com/magma/magma/issues/13216

## Additional Information


nms/app/components/lte/LteContext.tsx
```diff
@@ -428,8 +428,25 @@ export function SubscriberContextProvider(props: Props) {
             key,
             value,
             newState,
-            newSessionState,
           }),
+        refetchSessionState: (id?: SubscriberId) => {
```
thmsschmitt, line 463: A bit shorter would be:
```
refetchSessionState: (id?: SubscriberId) => {
     void FetchSubscriberState({networkId, id}).then(subscriberState => {
        if(subscriberState) {   
            setSessionState(sessionState => id ? {
              ...sessionState,
              [id]: subscriberState,
            } : subscriberState);
        }
     });
}
```

We could also cleanup `FetchSubscriberState` a bit so that it returns `{ [id]: subscriberState }` then it would have a consistent return type. Although be aware that we would still need to do   
`setSessionState(currentState => id ? {...currentState, ...subscriberState} : subscriberState)` Because we never want to merge the full new state with current state otherwise it would not be possible to delete entries.

nms/app/views/equipment/GatewayDetailSubscribers.tsx
```diff
@@ -53,7 +44,7 @@ export default function GatewayDetailSubscribers(props: GatewayDetailType) {
     (serialNum: string) => {
       //TODO[TS-migration] Something is seriously wrong here?
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-member-access
-      const subscriberInfo = (subscriberState as any).state?.[
```
thmsschmitt, line 56: There is no `state` on the `sessionState`. Shouldn't this be  `subscriberCtx.state?.[serialNum]` ?
Which also means we can get rid of the cast and the TODO.

thmsschmitt: The reason for the refetch here might be to update`subscriberCtx.gwSubscriberMap`.

nms/app/components/lte/LteContext.tsx
```diff
@@ -428,8 +428,25 @@ export function SubscriberContextProvider(props: Props) {
...
+                setSessionState(sessionState => ({
+                  ...sessionState,
+                  [id]: subscriberState,
+                }));
+              }
+            });
+          } else {
+            void FetchSubscriberState({networkId}).then(subscriberState => {
+              if (subscriberState) {
+                setSessionState(subscriberState);
```
thmsschmitt: I find it very confusing that we use two different terms for this.

nms/app/components/lte/LteContext.tsx
```diff
@@ -453,8 +453,21 @@ export function SubscriberContextProvider(props: Props) {
...
-            newSessionState,
           }),
+        refetchSessionState: (id?: SubscriberId) => {
+          void FetchSubscriberState({networkId, id}).then(sessionState => {
+            if (sessionState) {
+              setSessionState(currentSessionState =>
+                id
+                  ? {
+                      ...currentSessionState,
+                      sessionState,
```
thmsschmitt: `sessionState` is of type `Record<string, SubscriberState>` so this needs to be `...sessionState`.
I'm a bit puzzled why TypeScript does not complain here.

 We should also type the state i.e. 
 `const [sessionState, setSessionState] = useState<Record<string, SubscriberState>>({});`
 it still would not complain but at least we are giving it a chance.

nms/app/state/lte/SubscriberState.ts
```diff
@@ -102,12 +102,16 @@ export async function FetchSubscriberState(props: FetchProps) {
   const {networkId, enqueueSnackbar, id} = props;
   if (id !== null && id !== undefined) {
     try {
-      return (
-        await MagmaAPI.subscribers.lteNetworkIdSubscriberStateSubscriberIdGet({
-          networkId,
-          subscriberId: id,
-        })
-      ).data;
+      return {
```
thmsschmitt: We should no longer need the cast in line 200 now.

nms/app/components/lte/LteContext.tsx
thmsschmitt: While we are at it we should also type `const [subscriberMap, setSubscriberMap] = useState({});`
useState will infer the type from the initial state if nothing is given, so it will end up with `{}` here.

nms/app/state/lte/SubscriberState.ts
thmsschmitt: To make the code more explicit we could also annotate the return type of `FetchSubscriberState`

nms/app/components/lte/LteContext.tsx
```diff
@@ -453,8 +453,21 @@ export function SubscriberContextProvider(props: Props) {
             key,
             value,
             newState,
-            newSessionState,
           }),
+        refetchSessionState: (id?: SubscriberId) => {
+          void FetchSubscriberState({networkId, id}).then(sessionState => {
```
thmsschmitt: The variable `sessionState` here shadows the state defined above

nms/app/state/lte/SubscriberState.ts
```diff
@@ -102,12 +102,16 @@ export async function FetchSubscriberState(props: FetchProps) {
...
+        [id]: (
+          await MagmaAPI.subscribers.lteNetworkIdSubscriberStateSubscriberIdGet(
```
thmsschmitt: If the response it empty this will result in a state like:
``` { IMSI1234 : {} }```
but when fetching all subscriber states this imsi will not be included.
This would only make a difference if we iterate over the keys though.

thmsschmitt (approved):
The bug when refetching a single session state should be fixed, other then that this looks good to me.

PR #13240, opened 2022-07-08
chore(dp): Improve Domain-proxy GH actions workflow

sklgromek:

## Summary
As it's impossible to mark one item from GH matrix-strategy as a required job, we get rid of it and replace it with GH composite action, as was suggested in https://github.com/magma/magma/issues/12855#issuecomment-1177547133

Also, path_filters dor that jobs was adjusted for DP integration_tests

## Test Plan
- [ ] Check GH actions reports

## Additional Information

### DIFF
--- a/.github/workflows/composite/dp-integ-tests/action.yml
+++ b/.github/workflows/composite/dp-integ-tests/action.yml
@@ -1,0 +1,72 @@
+---
+name: dp-integ-tests
+description: Run DP integration tests
+
+inputs:
+  make_tests_target:
+    required: true
+    description: Make target for running relevant tests
+  test_dir:
+    required: true
+    description: Directory to store tests results
+    default: /tmp/integration-tests-results
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@7884fcad6b5d53d10323aee724dc68d8b9096a2e # pin@v2
+
+    - name: Set env
+      shell: bash
+      run: |
+        echo "MINIKUBE_DP_MAX_MEMORY=$(grep MemTotal /proc/meminfo | awk '{printf "%dm",$2/1024 - 1}')" >> $GITHUB_ENV
+
+    - name: Install Minikube
+      uses: manusa/actions-setup-minikube@cdef63c020a1c7b3d4787f30b3787ca1095ed9a7 # pin@v2.4.1
+      with:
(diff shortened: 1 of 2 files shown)


.github/workflows/composite/dp-integ-tests/action.yml
```diff
@@ -0,0 +1,72 @@
...
+  make_tests_target:
+    required: true
+    description: Make target for running relevant tests
+  test_dir:
+    required: true
+    description: Directory to store tests results
+    default: /tmp/integration-tests-results
+  github_token:
+    required: true
+    description: Github token
```
Neudrino: The Github token does not need to be handed in. It can be accessed via `${{ github.token }}`.

.github/workflows/composite/dp-integ-tests/action.yml
```diff
@@ -0,0 +1,72 @@
...
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@7884fcad6b5d53d10323aee724dc68d8b9096a2e # pin@v2
+    - name: Set env
+      shell: bash
+      run: |
+        echo "MINIKUBE_DP_MAX_MEMORY=$(grep MemTotal /proc/meminfo | awk '{printf "%dm",$2/1024 - 1}')" >> $GITHUB_ENV
```
Neudrino, line 22: Watch out for consistent spacing between steps. The other ones are separated by a "free line".

Neudrino (commented):
Good in general. Two minor comments inline.

I did not look at the pipeline in detail, as it is probably only copied over from before and was already reviewed and working.

.github/workflows/composite/dp-integ-tests/action.yml
sklgromek: I aligned it

sklgromek:
> I did not look at the pipeline in detail, as it is probably only copied over from before and was already reviewed and working.
Yes, it's mostly copied with small adjustments needed by composite actions

.github/workflows/composite/dp-integ-tests/action.yml
sklgromek: Thanks. I didn't know that

Neudrino (approved):

xbend (approved):

PR #13261, opened 2022-07-11
fix(nms):  Fix update Audience in Notifications on Alert Rule Edit Screen

talesmota:
## Summary

When the Audience changes, all routes for the specific previous user are deleted.
These fix deletes only the route for the ruleName for the previous user.

- There are others routes for my user like Picture bellow.
Screenshot from 2022-07-08 13-54-19

- When I change the Audience, all routes for my user are deleted, or changed.
Screenshot from 2022-07-08 13-55-12

I made a fix to exclude or change only the receiver from ruleName that action was made.

Files changed:
-  nms/app/views/alarms/components/hooks.ts
- nms/app/views/alarms/components/\_\_tests\_\_/FilterRouteByRuleName-test.ts

## Test Plan

Testing on NMS changing the receivers and with test file.
- nms/app/views/alarms/components/\_\_tests\_\_/FilterRouteByRuleName-test.ts


nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
```diff
@@ -0,0 +1,227 @@
+/**
+ * Copyright 2020 The Magma Authors.
```
Neudrino: 2022?

nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
```diff
@@ -0,0 +1,227 @@
...
+    const ruleName = 'High Disk Usage Alert';
+    const initialReceiver = 'User1';
+    const error = () => {
+      filterRouteByRuleName(_response, initialReceiver, ruleName);
+    };
+    expect(error).toThrow();
+  });
+});
+
+describe('Testint filterUpdatedFilterRoutes', () => {
```
Neudrino: Typo: `Testing`?

nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
```diff
@@ -0,0 +1,227 @@
...
+      {
+        receiver: 'User1',
+        match: {alertname: 'Bootstrap Exception Alert'},
+      },
+      {
+        receiver: 'User2',
+        match: {alertname: 'Gateway Checkin Failure'},
+      },
+    ],
+  };
```
Neudrino: There seems to be a lot of code duplication here. Should we maybe abstract this?

nms/app/views/alarms/components/hooks.ts
```diff
@@ -77,6 +77,57 @@ export function useLoadRules<TRuleUnion>({
...
+    return response.routes || [];
+  }
+  if (!initialReceiver || typeof initialReceiver === 'undefined') {
+    return response.routes || [];
+  }
+  const {routes} = response;
+  const filteredRoutes = (routes || []).filter((route: AlertRoutingTree) => {
+    const filterToDelete =
+      route.receiver === initialReceiver && route.match?.alertname === ruleName;
+    return route.receiver !== initialReceiver || !filterToDelete;
```
Neudrino: `Not (A and B)` is equivalent to `Not A or not B` thus that I believe the logic here could be simplified.

Neudrino (commented):
Some inconsistency in whitespaces (double spaces) as well as capitalising.

nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
talesmota: I moved this const outside of both describers.

nms/app/views/alarms/components/hooks.ts
talesmota: Thanks... I applied this modification!

Neudrino: The condition on the `route.receiver` is still duplicated.

Neudrino:
> Some inconsistency in whitespaces (double spaces) as well as capitalising.

Those inconsistencies still exist and should be straightened.

nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
```diff
@@ -0,0 +1,209 @@
...
+
+describe('Testing filterRouteByRuleName', () => {
+  test('Remove only  High Disk Usage Alert from User1', () => {
+    const ruleName = 'High Disk Usage Alert';
+    const initialReceiver = 'User1';
+    const routes = filterRouteByRuleName(response, initialReceiver, ruleName);
+    expect(routes.length).toBe(3);
+  });
+
+  test('Remove only  Gateway Checkin Failure from User2', () => {
```
Neudrino: E.g. 1:  (more below)
```js
  test('Remove only Gateway Checkin Failure from User2', () => {
```

nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
```diff
@@ -0,0 +1,209 @@
...
+    const ruleName = 'Gateway Checkin Failure';
+    const initialReceiver = 'User2';
+    const routes = filterRouteByRuleName(response, initialReceiver, ruleName);
+    expect(routes.length).toBe(3);
+  });
+
+  test('return all routes if initialReceiver is empty', () => {
```
Neudrino: E.g. 2: Capitalization is different from above. It should be consistent. More below.

nms/app/views/alarms/components/hooks.ts
```diff
@@ -101,8 +101,8 @@ export function filterRouteByRuleName(
   const {routes} = response;
   const filteredRoutes = (routes || []).filter((route: AlertRoutingTree) => {
     const filterToDelete =
-      route.receiver === initialReceiver && route.match?.alertname === ruleName;
-    return route.receiver !== initialReceiver || !filterToDelete;
+      route.receiver !== initialReceiver || route.match?.alertname !== ruleName;
+    return route.receiver !== initialReceiver || filterToDelete;
```
thmsschmitt: This can be simplified to 
```return route.receiver !== initialReceiver || route.match?.alertname !== ruleName;```

File: nms/app/views/alarms/components/hooks.ts [hunk 8]
```diff
@@ -77,6 +77,57 @@ export function useLoadRules<TRuleUnion>({
...
+ * @param {*} initialReceiver string
+ * @param {*} ruleName string
+ * @returns
+ */
+export function filterRouteByRuleName(
+  response: AlertRoutingTree | null | undefined,
+  initialReceiver: string | null | undefined,
+  ruleName: string | null | undefined,
+) {
+  if (!response || typeof response === 'undefined') {
```
thmsschmitt: undefined is already falsy so the `typeof response === 'undefined'` has no effect. Same for all the checks below.

nms/app/views/alarms/components/hooks.ts
```diff
@@ -77,6 +77,57 @@ export function useLoadRules<TRuleUnion>({
...
 }
 
+/**
+ * Filter a route from routes for a especific ruleName and receiver
+ * @param {*} response from apiUtil.useAlarmsApi
+ * @param {*} initialReceiver string
+ * @param {*} ruleName string
+ * @returns
+ */
+export function filterRouteByRuleName(
```
thmsschmitt, line 87: All those `null/undefined` sanity checks really clutter the code. This function could become a two liner if we get rid of them. TypeScript can help us here. `ruleName` for instance is always a string but might be empty. And the call site already checks that `initialReceiver` is not null. If we set up the  `initialReceiver` state so that it can't be undefined and don't pass in response but the rules or an empty array all sanity checks can be removed.

nms/app/views/alarms/components/hooks.ts
```diff
@@ -77,6 +77,57 @@ export function useLoadRules<TRuleUnion>({
...
+  if (!initialReceiver || typeof initialReceiver === 'undefined') {
+    return response.routes || [];
+  }
+  const {routes} = response;
+  const filteredRoutes = (routes || []).filter((route: AlertRoutingTree) => {
+    const filterToDelete =
+      route.receiver !== initialReceiver || route.match?.alertname !== ruleName;
+    return route.receiver !== initialReceiver || filterToDelete;
+  });
+  return filteredRoutes || [];
```
thmsschmitt: no need for the `|| []` we can directly return the result of filter

nms/app/views/alarms/components/hooks.ts
```diff
@@ -77,6 +77,57 @@ export function useLoadRules<TRuleUnion>({
...
+  receiver: string | null | undefined,
+) {
+  if (!initialReceiver || !receiver || !ruleName) return response?.routes;
+  if (!response) throw new Error("Response can't be null.");
+  return (response?.routes || []).map((route: AlertRoutingTree) => {
+    if (
+      initialReceiver === route.receiver &&
+      route.match?.alertname === ruleName
+    ) {
+      if (!!receiver) {
```
thmsschmitt: `receiver` was already checked in line 116

Neudrino:
It would be very nice if you could **rebase** onto the current `master` branch instead of merging that branch into your feature branch. There is some documentation regarding the usual workflow in the [Contributer Guide](https://github.com/magma/magma/wiki/Contributing-Code#developing-workflow). :slightly_smiling_face:

nms/app/views/alarms/components/hooks.ts
```diff
@@ -77,6 +77,57 @@ export function useLoadRules<TRuleUnion>({
...
+  const {routes} = response;
+  const filteredRoutes = (routes || []).filter((route: AlertRoutingTree) => {
+    const filterToDelete =
+      route.receiver !== initialReceiver || route.match?.alertname !== ruleName;
+    return route.receiver !== initialReceiver || filterToDelete;
+  });
+  return filteredRoutes || [];
+}
+
+export function filterUpdatedFilterRoutes(
```
thmsschmitt, line 104: How about `updateReceiver`.

nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
```diff
@@ -0,0 +1,209 @@
...
+    expect(error).toThrow();
+  });
+});
+
+describe('Testing filterUpdatedFilterRoutes', () => {
+  test('update only  High Disk Usage Alert from User1 to User2', () => {
+    const ruleName = 'High Disk Usage Alert';
+    const initialReceiver = 'User1';
+    const receiver = 'User2';
+    const initialState = response.routes.filter(
```
thmsschmitt: There is more logic in the test than in the function that is tested. And we only really test the first here anyhow.
Much simpler would be something like:
```
    const routes = filterUpdatedFilterRoutes(
      response,
      initialReceiver,
      ruleName,
      receiver,
    );

    expect(routes![0].receiver).toBe(receiver);
    expect(routes![1].receiver).toBe(initialReceiver);
```

nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
```diff
@@ -0,0 +1,209 @@
...
+      expect(routes).not.toBe(undefined);
+    }
+  });
+
+  test('return exact routes if initial receiver undefined', () => {
+    const ruleName = 'High Disk Usage Alert';
+    const initialReceiver = undefined;
+    const receiver = 'User2';
+
+    const routes:
```
thmsschmitt: No need to specify the return type. TypesScript already knows this.

nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
```diff
@@ -0,0 +1,209 @@
...
+      | Array<AlertRoutingTree>
+      | undefined = filterUpdatedFilterRoutes(response, '', ruleName, receiver);
+    if (routes3) {
+      expect(response.routes).toBe(routes3);
+    }
+  });
+  test('return exact routes if receiver undefined', () => {
+    const ruleName = 'High Disk Usage Alert';
+    const initialReceiver = 'User1';
+    const receiverUndefined = undefined;
```
thmsschmitt: We don't need to redefine undefined or null

nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
```diff
@@ -0,0 +1,209 @@
...
+    const receiverNull = null;
+    const receiverEmpty = '';
+
+    const routesUndefined:
```
thmsschmitt: Also here no need to have spell out the return types

nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
```diff
@@ -0,0 +1,209 @@
...
+      match: {alertname: 'Bootstrap Exception Alert'},
+    },
+    {
+      receiver: 'User2',
+      match: {alertname: 'Gateway Checkin Failure'},
+    },
+  ],
+};
+
+describe('Testing filterRouteByRuleName', () => {
```
thmsschmitt, line 37: I general this are way too many tests compared to how simple the logic is. Each of the functions really only has only relevant condition. I would limit the test to a few realistic scenarios

thmsschmitt (changes requested):

nms/app/views/alarms/components/__tests__/FilterRouteByRuleName-test.ts
```diff
@@ -0,0 +1,209 @@
...
+  test('Remove only  High Disk Usage Alert from User1', () => {
```
thmsschmitt: There are double whitespaces here, also below

nms/app/views/alarms/components/hooks.ts
talesmota: Hi Thanks for feedback @thmsschmitt . Could you help on this? My _Eslint_ is complaining about the funciton _filterUpdatedFilterRoutes_ because on ' //update existing route' blocks the variables **initialReceiver** and **receiver** can be null or undefined. So because of that I made all these verification. But analyzing what you said, is it enough if i just use the **initialReceiver!** and **receiver!** to prevent null and undefined while calls the _filterUpdatedFilterRoutes_?


```
// update existing route
      const _updatedRoutes = filterUpdatedFilterRoutes(
        response,
        initialReceiver!,
        ruleName,
        receiver!,
      );
      updatedRoutes = {
        ...updatedRoutes,
        routes: _updatedRoutes,
      };
```

talesmota (commented):
I've made changes changes all suggestions. I hope everything goes well.

nms/app/views/alarms/components/hooks.ts
thmsschmitt: No not really, there are three branches:

-  receiver === null && initialReceiver !== null
-  receiver !== null &&  initialReceiver === null 
-  receiver === null && initialReceiver === null || receiver !== null && initialReceiver !== null

`filterUpdatedFilterRoutes` is called in the last case so it needs to handle the case of both being null.
`filterRouteByRuleName` is called in the first case so it can assume that initialReceiver is not null.

`undefined` should be taken out of the picture by setting up the states like so:

```
const [initialReceiver, setInitialReceiver] = React.useState<string | null>(null);
const [receiver, setReceiver] = React.useState<string | null>(null); 
```

nms/app/views/alarms/components/hooks.ts
```diff
@@ -134,9 +178,7 @@ export function useAlertRuleReceiver({
       // remove the route
       updatedRoutes = {
         ...updatedRoutes,
-        routes: (response?.routes || []).filter(
-          route => route.receiver !== initialReceiver,
-        ),
+        routes: filterRouteByRuleName(response, initialReceiver, ruleName),
```
thmsschmitt, line 181: Sorry, I looked at this a bit more closely. The hook and the dialog itself should probably be rewritten at some point. All of this is over complicating things. 

A given rule can only have one subscriber, we don't need the initial subscriber to update the routes. So for now we can simplify this to
```ts
    if (
      (receiver == null || receiver.trim() === '') &&
      initialReceiver != null
    ) {
      // remove the route
      updatedRoutes = {
        ...updatedRoutes,
        routes: (updatedRoutes.routes || []).filter(
          route => route.match?.alertname !== ruleName,
        ),
      };
    } else if (receiver != null && initialReceiver == null) {
      // creating a new route
      const newRoute: AlertRoutingTree = {
        receiver: receiver,
        match: { alertname: ruleName },
      };
      updatedRoutes = {
        ...updatedRoutes,
        routes: [...(updatedRoutes.routes || []), newRoute],
      };
    } else if (initialReceiver !== null && receiver !== null) {
      updatedRoutes = {
        ...updatedRoutes,
        routes: (updatedRoutes.routes || []).map(route => {
          if (route.match?.alertname === ruleName) {
            route.receiver = receiver;
          }
          return route;
        }),
      };
    }
```    

The tests should test against the UI then we don't have to touch them if we refactor the logic like this.

thmsschmitt (changes requested):

thmsschmitt:
@talesmota do you mind if we take this over?

PR #13268, opened 2022-07-12
fix(agw): Adapt nonsanity tests for Python 3.8

sebathomas:
## Summary

`time.clock()` is no longer supported.

## Test Plan

```
make nonsanity
```

## Additional Information

### DIFF
--- a/lte/gateway/python/integ_tests/s1aptests/test_attach_and_mme_restart_loop_detach_and_mme_restart_loop_multi_ue.py
+++ b/lte/gateway/python/integ_tests/s1aptests/test_attach_and_mme_restart_loop_detach_and_mme_restart_loop_multi_ue.py
@@ -101,7 +101,7 @@ class TestAttachAndMmeRestartLoopDetachAndMmeRestartLoopMultiUe(
 
         for ue in ue_ids:
             # Now detach the UE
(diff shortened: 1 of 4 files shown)


lte/gateway/python/integ_tests/s1aptests/test_attach_and_mme_restart_loop_detach_and_mme_restart_loop_multi_ue.py
```diff
@@ -14,6 +14,7 @@
 import random
 import time
 import unittest
+from datetime import datetime
```
VinashakAnkitAman: Since we are still importing `time` library, instead of importing one more library `datetime`, can we use something from `time` library like `time.time()` or `time.perf_counter()` or `time.process_time()`

sebathomas: Good point, I now use `time.time()` instead.

VinashakAnkitAman:
Hi @sebathomas.. This PR needs rebase on latest master before merge.

VinashakAnkitAman (approved):
LGTM

PR #13269, opened 2022-07-12
fix(agw): Migrate hash list of mmeid2associd to protobuf_map

rsarwad:
Modified code to replace hashlist with protobuf map for mmeid2associd hash table in s1ap module
The PR partially addresses the issue, https://github.com/magma/magma/issues/11190

Test Plan
Executed s1ap sanity test suite and unit test case.
Verified manually with test_attact_detach_with_mme_restart.py test case, to check whether contents of map, mmeid2associd are written correctly into DB.


ssanadhya (approved):
Lgtm

PR #13273, opened 2022-07-12
fix(nms): Disable changing the ID of a gateway pool

### LINKED ISSUE #13183: Changing the ID of a gateway pool creates a new one
1. Create a gateway pool through the NMS
2. Edit the gateway pool and change the ID
3. BUG - Saving the gateway pool with the changed ID creates a new gateway pool in addition to the old one.

Screenshot 2022-07-04 at 16 53 34

### Environment

- **Version:** master
- **Affected Component:**  NMS
- **Deployment Environment:**  Vagrant (AGW)

thmsschmitt:
## Summary

Disable The ID field when editing an existing gateway pool.
- Fixes: #13183

## Test Plan

* Make sure the ID of an existing  gateway pool can't be changed.  
* Makes sure that new gateways pools can be created with an ID.

### DIFF
--- a/nms/app/views/equipment/__tests__/EquipmentGatewayPoolTest.tsx
+++ b/nms/app/views/equipment/__tests__/EquipmentGatewayPoolTest.tsx
@@ -28,7 +28,7 @@ import {
   SetGatewayPoolsState,
   UpdateGatewayPoolRecords,
 } from '../../../state/lte/EquipmentState';
-import {fireEvent, render, wait} from '@testing-library/react';
+import {fireEvent, render, wait, waitFor} from '@testing-library/react';
 import {mockAPI} from '../../../util/TestUtils';
 import {useEnqueueSnackbar} from '../../../hooks/useSnackbar';
 import {useState} from 'react';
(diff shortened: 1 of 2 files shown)


nms/app/views/equipment/__tests__/EquipmentGatewayPoolTest.tsx
```diff
@@ -237,6 +237,60 @@ describe('<GatewayPools />', () => {
...
+    expect(poolId).toBeDisabled();
+
+    if (
+      name instanceof HTMLInputElement &&
+      mmeGroupId instanceof HTMLInputElement
+    ) {
+      fireEvent.change(name, {target: {value: 'foo'}});
+      fireEvent.change(mmeGroupId, {target: {value: '4'}});
+    } else {
+      throw 'invalid type';
```
Neudrino: Does that really only mean that the type is incorrect? Or does it moreover mean that components are missing on the page?
Maybe we could have a more speaking and precise error here as to make it more useful when occuring.

Neudrino: Second thought. As we are in a test here, I thought we fully control the environment, such that errors would only occur where we expect them?

thmsschmitt: You are right, this is only here to make the typing happy. I was lazy and copied another test. I'll change this to type casts.

Neudrino (approved):

PR #13293, opened 2022-07-14
fix(dp): add max_ibw_mhz validation

hejjestemjarek:
Signed-off-by: Jarosław Jaszczuk <jaroslaw.jaszczuk@freedomfi.com>

## Summary
Added validation for `max_ibw_mhz` field, that ensures the value is:
- `>= 5`
- `<= 150`
- multiple of 5

## Test Plan
Added test cases.

## Additional Information

### DIFF
--- a/dp/cloud/go/services/dp/obsidian/models/swagger.v1.yml
+++ b/dp/cloud/go/services/dp/obsidian/models/swagger.v1.yml
@@ -22,7 +22,6 @@ definitions:
         type: integer
         x-nullable: false
         maximum: 150
-        minimum: 5
         multipleOf: 5
       min_power:
         description: min tx power available on cbsd
--- a/dp/cloud/go/services/dp/obsidian/models/swagger.v1.yml
+++ b/dp/cloud/go/services/dp/obsidian/models/swagger.v1.yml
@@ -21,6 +21,9 @@ definitions:
         description: this is the maximum allowed difference in MHz between leftmost end of leftmost channel and rightmost end of rightmost channel used by a Base Station (eNB)
         type: integer
         x-nullable: false
+        maximum: 150
+        minimum: 5
+        multipleOf: 5
       min_power:
         description: min tx power available on cbsd
         type: number
--- a/dp/cloud/go/services/dp/obsidian/models/validation_test.go
+++ b/dp/cloud/go/services/dp/obsidian/models/validation_test.go
@@ -4,13 +4,12 @@ import (
 	"context"
 	"testing"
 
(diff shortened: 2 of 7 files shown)


dp/cloud/go/services/dp/obsidian/models/swagger.v1.yml
```diff
@@ -21,6 +21,9 @@ definitions:
         description: this is the maximum allowed difference in MHz between leftmost end of leftmost channel and rightmost end of rightmost channel used by a Base Station (eNB)
         type: integer
         x-nullable: false
+        maximum: 150
+        minimum: 5
```
jkmar: Can we change it to 20, so that is it always greater than bandwidth of a single channel?

hejjestemjarek: Removed `minimum: 5` from swagger validation, but modified `MutableCbsd`'s `ValidateModel` func to ensure that `max_ibw_mhz` is not less than `bandwidth_mhz`.

dp/cloud/go/services/dp/obsidian/models/validation_test.go
```diff
@@ -83,6 +83,14 @@ func TestMutableCbsd_Validate(t *testing.T) {
...
 		data:          b.NewMutableCbsdModelPayloadBuilder().WithMaxIbwMhz(0).Payload,
 		expectedError: "max_ibw_mhz in body is required",
+	}, {
+		name:          "Should validate max ibw mhz too high",
+		data:          b.NewMutableCbsdModelPayloadBuilder().WithMaxIbwMhz(151).Payload,
+		expectedError: "max_ibw_mhz in body should be less than or equal to 150",
+	}, {
+		name:          "Should validate max ibw mhz not multiple of 5",
+		data:          b.NewMutableCbsdModelPayloadBuilder().WithMaxIbwMhz(7).Payload,
+		expectedError: "max_ibw_mhz in body should be a multiple of 5",
```
jkmar, line 95: I would add a test for `if m.Capabilities.MaxIbwMhz < m.FrequencyPreferences.BandwidthMhz` and also for this `!*m.GrantRedundancy && *m.CarrierAggregationEnabled`, since I've noticed that it is missing for some reason

hejjestemjarek: Added them in a separate test, since original `TestMutableCbsd_Validate` covers only swagger generated `Validate` func and would not cover `ValidateModel` func.

jkmar (approved):

PR #13295, opened 2022-07-14
fix(agw): Added testcase with support to large number of UEs

VinashakAnkitAman:
## Title
fix(agw): Added testcase with support to large number of UEs

## Summary
By default a testcase is capable of handling only 243 UEs at this point of time because of the hardcoded configurations. This PR adds a reference testcase to handle IP pool configuration and UE inactivity timer configuration to support large number of UEs.

## Test plan
Verified with sanity


lte/gateway/python/integ_tests/s1aptests/test_scalability_attach_detach_multi_ue.py
```diff
@@ -0,0 +1,95 @@
...
+import s1ap_wrapper
+
+
+class TestScalabilityAttachDetachMultiUe(unittest.TestCase):
+    """Integration Test: TestScalabilityAttachDetachMultiUe"""
+
+    def setUp(self):
+        """Initialize before test case execution"""
+        self.default_ip_block = s1ap_wrapper.TestWrapper.TEST_IP_BLOCK
+        s1ap_wrapper.TestWrapper.TEST_IP_BLOCK = "192.168.128.0/17"
```
rsarwad, line 32: In case if someone wants to execute this test case for different number of UEs, can we add comment to inform that mask value needs to be changed with logic indicated in s1ap_wrapper.py file or same comment can be added here as well

lte/gateway/python/integ_tests/s1aptests/test_scalability_attach_detach_multi_ue.py
```diff
@@ -0,0 +1,95 @@
...
+
+        for ue in ue_ids:
+            print("************************* Calling detach for UE id:", ue)
+            self._s1ap_wrapper.s1_util.detach(
+                ue,
+                s1ap_types.ueDetachType_t.UE_NORMAL_DETACH.value,
+            )
+
+        # Reset the inactivity timer value to default 2 mins (120000 ms)
+        print("Resetting the inactivity timer value to dafault value (2 mins)")
```
pruthvihebbani: typo default

lte/gateway/python/integ_tests/s1aptests/test_scalability_attach_detach_multi_ue.py
```diff
@@ -0,0 +1,95 @@
...
+            self._s1ap_wrapper.s1_util.attach(
+                req.ue_id,
+                s1ap_types.tfwCmd.UE_END_TO_END_ATTACH_REQUEST,
+                s1ap_types.tfwCmd.UE_ATTACH_ACCEPT_IND,
+                s1ap_types.ueAttachAccept_t,
+            )
+            # Wait for EMM Information from MME
+            self._s1ap_wrapper._s1_util.receive_emm_info()
+            ue_ids.append(req.ue_id)
+
```
pruthvihebbani, line 79: Can we add flow verification before detaching

pruthvihebbani (changes requested):

lte/gateway/python/integ_tests/s1aptests/test_scalability_attach_detach_multi_ue.py
VinashakAnkitAman: Added comment with reference to s1ap_wrapper.py as well as S1APTester documentation to get information from whichever is updated latest.

lte/gateway/python/integ_tests/s1aptests/test_scalability_attach_detach_multi_ue.py
VinashakAnkitAman: The flow rules creation verification and deletion verification needs a minimum of 8 seconds (3+5) for every UE which is equal to around 2.5 hours additional wait time for this test case with 1024 UEs. Similarly adding only flow rules deletion verification before detach with1024 UEs expects around 1.5 hours additional wait time. The wait duration will increase as the number of UEs increase (which will be part of the later plan).

This is why I have intentionally not added the code to verify flow rules verification for keeping this test case simple to basic attach-detach only to verify till how many UEs magma can support with its increasing stability. In future we need to add more functionality oriented test cases for load testing including inactivity, service and data. There we will be adding this flow rules verification as well.

rsarwad (approved):
LGTM

pruthvihebbani (approved):

pruthvihebbani:
Not able to merge as mme test is failing. Please rebase

PR #13301, opened 2022-07-14
fix(agw): both IPv4 and IPv6 downlink route returns are evaluated

mpfirrmann:
Signed-off-by: Marco Pfirrmann <marco.pfirrmann@tngtech.com>

## Summary

At the moment, the return values of the downlink routes for IPv4 and IPv6 are stored in the same variable and the function return value only evaluates the IPv6 return. This PR changes this behavior. 

## Additional Information

### DIFF
--- a/lte/gateway/python/integ_tests/s1aptests/util/traffic_util.py
+++ b/lte/gateway/python/integ_tests/s1aptests/util/traffic_util.py
@@ -135,15 +135,15 @@ class TrafficUtil(object):
 
     def update_dl_route(self, ue_ip_block):
         """Update downlink route in TRF server"""
-        ret_code = self.exec_command(
+        ret_code_ipv4 = self.exec_command(
             "sudo ip route flush via 192.168.129.1 && sudo ip route "
             "replace " + ue_ip_block + " via 192.168.129.1 dev eth2",
         )
-        ret_code = self.exec_command(
+        ret_code_ipv6 = self.exec_command(
             "sudo ip -6 route flush via " + self.agw_ipv6 + " && sudo ip -6 route "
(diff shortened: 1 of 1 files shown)


voisey (approved):

lte/gateway/python/integ_tests/s1aptests/util/traffic_util.py
```diff
@@ -135,15 +135,15 @@ def exec_command(self, command):
...
         """Update downlink route in TRF server"""
-        ret_code = self.exec_command(
+        ret_code_ipv4 = self.exec_command(
             "sudo ip route flush via 192.168.129.1 && sudo ip route "
             "replace " + ue_ip_block + " via 192.168.129.1 dev eth2",
         )
-        ret_code = self.exec_command(
+        ret_code_ipv6 = self.exec_command(
             "sudo ip -6 route flush via " + self.agw_ipv6 + " && sudo ip -6 route "
             "replace " + self.ue_ipv6_block + " via " + self.agw_ipv6 + " dev eth3",
```
wolfseb, line 144: Maybe out of scope, but where the `ue_ip_block` and the `ue_ipv6_block` are coming from is inconsistent, same with the hard-coded ipv4 and the `agw_ipv6`.

mpfirrmann: That is correct. Both, `self.agw_ipv6` and `self.ue_ipv6_block`, are only used here. @pruthvihebbani?

pruthvihebbani: Ipv6 addresses are hardcoded for now. I will create a new PR to handle this similar to ipv4.

pruthvihebbani (approved):

PR #13311, opened 2022-07-15
chore(dp): Get rid of NMS UI from integration tests targets

sklgromek:

## Summary
Remove NMS from integration tests make targets, as is not using there

### DIFF
--- a/dp/cloud/docker/python/test_runner/entrypoint.sh
+++ b/dp/cloud/docker/python/test_runner/entrypoint.sh
@@ -2,6 +2,7 @@
 set -e
 
 if [ "$TESTS_DEV" = true ] ; then
+    echo "This container will sleep infinitely when TESTS_DEV env is set to true, it's for tests development"
     sleep infinity
 else
     python "$@"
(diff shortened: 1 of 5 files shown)


dp/cloud/docker/python/test_runner/entrypoint.sh
```diff
@@ -2,6 +2,7 @@
 set -e
 
 if [ "$TESTS_DEV" = true ] ; then
+    echo "This container will sleep infinitely when TESTS_DEV env is set to true, it's for tests development"
```
jkmar, line 5: Is it still deployed then?

sklgromek: Yes, it's deployed with `dev_integration_tests` make target, it gives oportunity to enter the container, and run tests manually. I added this comment cause the container with just sleep can confuse a user

jkmar: Hmm, I thought we are going to remove even building NMS when running integration tests, since that can be flaky, because it downloads a lot of external dependencies, but maybe it is not possible

sklgromek: Oh, Ok I think I misunderstood. I removed NMS in this PR. This change which you commented out was made by the way, and is not related with NMS disabling

jkmar (approved):

PR #13313, opened 2022-07-15
feat(nms): Add carrier aggregation fields to CBSD

ivan-freedomfi:
## Summary
- Add 3 carrier aggregation inputs with default values to the CBSD add/edit form:
  - `carrier_aggregation_enabled` (default `false`)
  - `grant_redundancy` (default `true`)
  - `capabilities.max_ibw_mhz` (default `150`)
- Send them in the POST and PUT requests.
- Add carrier aggregation and grant redundancy interdependence logic:
  - If `carrier_aggregation_enabled` checkbox is set to `true`, `grant_redundancy` checkbox is set to `true` automatically.
  - If `grant_redundancy` checkbox  is set to `false`, `carrier_aggregation_enabled` checkbox is set set to `false` automatically.

Signed-off-by: Ivan Sergiienko <ivan@freedomfi.com>

### Fields in the form

new-fields

### Checkbox interdependence

checkboxes

### POST request (create new CBSD)

image

### PUT request (edit CBSD)

image

### DIFF
--- a/nms/app/views/domain-proxy/CbsdEdit.tsx
+++ b/nms/app/views/domain-proxy/CbsdEdit.tsx
@@ -117,6 +117,9 @@ type CbsdFormData = {
   frequenciesMhz: Array<number | string>;
   cbsdCategory: CbsdCategoryEnum;
   singleStepEnabled: boolean;
+  grantRedundancyEnabled: boolean;
+  carrierAggreGationEnabled: boolean;
+  maxIbw: number | string;
 };
 
 const convertToCbsdFormData = (cbsd?: Cbsd): CbsdFormData => {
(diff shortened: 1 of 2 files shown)


thmsschmitt (approved):
Looks good to me

Neudrino (commented):
What is the style guideline here? Would it not be better to have the check boxes on the left side of the respective text? I think this is what most applications do nowadays.
If the checkbox and the text are not in the same line, it can be quite confusing, which check box belongs to which line of text.

thmsschmitt:
> What is the style guideline here? Would it not be better to have the check boxes on the left side of the respective text? I think this is what most applications do nowadays. If the checkbox and the text are not in the same line, it can be quite confusing, which check box belongs to which line of text.

You are right I somehow assumed that this is our usual style, but it isn't. We are typically using toggle switches instead of checkboxes.
Screenshot 2022-07-19 at 12 39 57
We should also do this for CBSDs.

ivan-freedomfi:
> > What is the style guideline here? Would it not be better to have the check boxes on the left side of the respective text? I think this is what most applications do nowadays. If the checkbox and the text are not in the same line, it can be quite confusing, which check box belongs to which line of text.
> 
> You are right I somehow assumed that this is our usual style, but it isn't. We are typically using toggle switches instead of checkboxes. Screenshot 2022-07-19 at 12 39 57 We should also do this for CBSDs.

I'm a bit busy this week, let's merge it as is to have it in `v1.8.0`, and I can change the design later.
Ok? :)

PR #13288, opened 2022-07-13
feat(pipelined): Support for Non-NAT mode in 5G-SA

mehul-jindal:
Signed-off-by: mehul-jindal-wavelabs <mehul.jindal@wavelabs.ai>

## Description
- If in `pipelined.yml` the following flag is set as false: `enable_nat = false`,
- then, fetch the `ovs_uplink_port_name` value from pipelined config and set the corresponding port number to `uplink_port`
### Logs
[table0.log](https://github.com/magma/magma/files/9104231/table0.log)

image

PR #13093, opened 2022-06-27
chore(agw): several warnings in the python unittests are …

mpfirrmann:
…fixed

Signed-off-by: Marco Pfirrmann <marco.pfirrmann@tngtech.com>

## Summary

After migrating the python unittests from `nose` to `pytest` (#13052), several deprecation warning are displayed. This PR fixes the first warnings, focusing on

- deprecated `assertEquals`
- deprecated `asyncio.Task`
- classes named `*Test` do not contain tests 

## Test Plan

```vagrant@magma-dev-focal:~/magma/lte/gateway$ make test_python```

before: 
```
======================== 434 passed, 4 skipped, 118 warnings in 99.23s (0:01:39) ========================
======================== 154 passed, 3 skipped, 14 warnings in 527.74s (0:08:47) ========================
======================== 142 passed, 123 warnings in 54.91s ======================================== 
```
after:
```
======================= 434 passed, 4 skipped, 115 warnings in 111.66s (0:01:51) =======================
======================== 154 passed, 3 skipped, 5 warnings in 536.91s (0:08:56) =========================
======================== 142 passed, 122 warnings in 52.82s =========================================
```

## Additional Information

- [ ] This change is backwards-breaking

PR #13320, opened 2022-07-18
fix(pipelined): missing sudo test is bazelified

nstng:

This PR fixes two things:
* in #13288 a new pipelined sudo test was added that was not bazelified
  * this broke a CI check that tests if all python files are bazelified - it is under investigation why the slack notification was not triggered
* in #13093 `lte/gateway/python/magma/pipelined/tests/pipelined_test_util.py` was refactored and the changes were not reflected in Bazel
  * this broke the bazel sudo tests in general - they are not yet executed in CI

## Additional Information


LKreutzer (approved):
Lgtm, sudo test was succesful locally

PR #13321, opened 2022-07-18
refactor(orc8r): Replace moriyoshi routewrapper

MoritzThomasHuebner:
## Summary

Replaces "github.com/moriyoshi/routewrapper" package with "github.com/vishvananda/netlink".

I also did refactored some of the existing logic into separate functions. Unfortunately, the individual commits may be somewhat messy.

## Test Plan

`routewrapper` uses the `ip route` command to retrieve routing table information, so we can manually compare the new routing table with the changes to the one we get with `ip route` and also the one we used to get with `netlink`. 

I spun up some docker images and the vagrant machines to add some more routing table entries.

In the image below, I used the IntelliJ debugger to look in the routing table before (left) and after (right) I made the changes. The order has changed somewhat because of how the information is retrieved. However, the order in the routing table does not matter. I highlighted in red the routing table entries that are identical.

Manual comparison

## Additional Information

### DIFF
--- a/orc8r/gateway/go/status/gateway_status.go
+++ b/orc8r/gateway/go/status/gateway_status.go
@@ -156,49 +156,30 @@ func GetNetworkInfo() *NetworkInfo {
 	}
 }
 
-func getRoutes(hRoutes []netlink.Route, recursiveResolving bool) []*Route {
+func getRoutes(hostRoutes []netlink.Route, recursiveResolving bool) []*Route {
 	interfaceToIP := make(map[string]string)
 	var unlinkedHostRoutes []netlink.Route
 	var routes []*Route
 
-	for _, r := range hRoutes {
-		src := getSourceIP(r)
+	for _, hostRoute := range hostRoutes {
+		src := getSourceIP(hostRoute)
 
 		if src == "" {
-			unlinkedHostRoutes = append(unlinkedHostRoutes, r)
+			unlinkedHostRoutes = append(unlinkedHostRoutes, hostRoute)
 			continue
 		}
 
-		index := r.LinkIndex
-
-		dest := getDestinationIP(r).IP.To4()
+		dest := getDestinationIP(hostRoute).IP.To4()
 		if dest == nil {
 			continue
 		}
-		gw := r.Gw.To4()
-		if gw == nil {
-			gw = r.Gw
-			if len(gw) == 0 {
-				if len(dest) == net.IPv4len {
-					gw = []byte{0, 0, 0, 0}
-				} else {
-					gw = net.IP([]byte{0, 0, 0, 0}).To16()
-				}
-			}
-		}
+		gw := getGatewayIP(hostRoute, dest)
+		maskStr := getMaskStr(hostRoute, dest)
 
-		maskStr := getDestinationIP(r).Mask.String()
-		if len(dest) == net.IPv4len {
-			maskV4 := net.IP(getDestinationIP(r).Mask).To4()
-			if maskV4 != nil {
-				maskStr = maskV4.String()
-			}
-		}
-
-		convertedIface := getNetInterface(index)
+		convertedIface := getNetInterface(hostRoute.LinkIndex)
 		route := &Route{
 			DestinationIp:      dest.String(),
-			GatewayIp:          gw.String(),
+			GatewayIp:          gw,
 			Genmask:            maskStr,
 			NetworkInterfaceId: convertedIface.Name,
 		}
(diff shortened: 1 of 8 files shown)


orc8r/gateway/go/status/host_info.go
```diff
@@ -78,10 +78,7 @@ func init() {
 	}
 	disksInfo, _ = disk.Partitions(true)
 
-	w, err := routewrapper.NewRouteWrapper()
-	if err == nil {
-		hostRoutes, _ = w.Routes()
-	}
+	hostRoutes, _ = netlink.RouteList(nil, 0)
```
sebathomas: Maybe you can use `netlink.RouteList(nil, netlink.FAMILY_V4)` here to avoid ipv6 addresses instead of filtering later.

MoritzThomasHuebner: Turns out this is the right idea. I made the change.

orc8r/gateway/go/status/gateway_status.go
```diff
@@ -147,43 +148,68 @@ func GetNetworkInfo() *NetworkInfo {
...
+	var unlinkedHostRoutes []netlink.Route
+	var routes []*Route
+
+	for _, hostRoute := range hostRoutes {
+		src := getSourceIP(hostRoute)
+
+		if src == "" {
+			// No source IP is stored in hostRoute, we try to resolve this later
+			unlinkedHostRoutes = append(unlinkedHostRoutes, hostRoute)
+			continue
```
sebathomas, line 187: Not sure if I'm missing something but if we build the map `linkIndexToIP` at the beginning of the function, couldn't we resolve the source IP here and then avoid the recursion?

```suggestion
			src = linkIndexToIP[hostRoute.LinkIndex]
```

orc8r/gateway/go/status/gateway_status.go
```diff
@@ -249,3 +275,66 @@ func GetGatewayStatus() *GatewayStatus {
...
+		Index:        iface.Index,
+		MTU:          iface.MTU,
+		Name:         iface.Name,
+		HardwareAddr: parsedAddr,
+	}
+	return convertedIface
+}
+
+// getSourceIP resolves the source IP as a string given a host route.
+// Returns an empty string if no source IP is given.
```
sebathomas: This function is small enough that the comment is not needed. It's as easy to see what it does from reading the code as from reading the comment (and that's a good thing :)).

Same for the functions below.

sebathomas (commented):
Tested that it works on my machine. :+1:

Just added some ideas for refactoring.

orc8r/gateway/go/status/gateway_status.go
MoritzThomasHuebner: This isn't quite as easy, we need to fully populate the `interfaceToIP` map first to build `linkIndexToIP`. 

Since we do not know in which order we find the link indices, we always will need to iterate twice over the host routes. The first time we iterate to find all where the source IP is given, and the second time to try to find matching link indices for the host routes that have no stated source IP.

I chose recursion since it seemed like the simplest way to avoid redundant logic.

MoritzThomasHuebner: I removed the recursion. It now looks also a bit different, so maybe have another look.

orc8r/gateway/go/status/gateway_status.go
```diff
@@ -147,43 +148,73 @@ func GetNetworkInfo() *NetworkInfo {
...
+
+// getRoutingTable resolves the routing table based on the hostRoutes. In the first
+// call to getRoutes, we obtain routes that can have an associated source IPs.
+// We use the remaining outputs to infer the source IPs of the other hostRoutes
+// and extract the desired format with another call to getRoutes.
+func getRoutingTable(hostRoutes []netlink.Route) []*Route {
+	routes, unlinkedHostRoutes, interfaceToIP := getRoutes(hostRoutes)
+	unlinkedHostRoutes = linkRoutes(unlinkedHostRoutes, interfaceToIP)
+	// Get the routing table entries for the newly linked host routes.
+	// recursiveResolving is false because we can no longer resolve any new
```
sebathomas: This comment needs to be updated too, still has a reference to recursive resolving.

sebathomas (approved):
Looking good, there's just a left-over comment from the old version that needs to be adapted or removed.

orc8r/gateway/go/status/gateway_status.go
MoritzThomasHuebner: Done

PR #13331, opened 2022-07-18
ci: Remove bazel-cache-push workflow

jheidbrink:
## Summary

Remove the "Push Bazel Cache To S3" workflow. In #13330 we stop using the zipped disk caches and use remote caches instead.

## Test Plan

not tested

## Additional Information

When this is merged, the secrets `AWS_ACCESS_KEY_AMI_BAZEL_CACHE_S3` and `AWS_SECRET_ACCESS_KEY_AMI_BAZEL_CACHE_S3` can be removed.

- [ ] This change is backwards-breaking (when pull/13330 isn't merged yet, this can slightly increase bazel build times, but not break anything)


LKreutzer (approved):
Lgtm, obviously should only be merged after https://github.com/magma/magma/pull/13330. 

Are the VM caches used anywhere at the moment?

We should keep the configs used here: `production`, `limited_mem`, `asan` for VM/devcontainer in mind as they should also be added to the bazel workflow (this is already planned).

jheidbrink:
> obviously should only be merged after https://github.com/magma/magma/pull/13330.

It's not a strict requirement though, the codespaces would just get an older cache until #13330 is merged.

> Are the VM caches used anywhere at the moment?

No, the caches are built and uploaded, but seem unused.

> We should keep the configs used here: production, limited_mem, asan for VM/devcontainer in mind as they should also be added to the bazel workflow (this is already planned).

Yes, good point!

maxhbr (approved):
LGTM

maxhbr:
I removed the secrets

PR #13341, opened 2022-07-19
ci: Create composite action for maximizing the build space

LKreutzer:

## Summary

- A composite GH action is created to maximize the build space in the GH runners.

## Test Plan

- Compare workflow logging on PR to logging on master.
  - Build space left at the end of the workflows is more than before the action cleaned up. 

## Additional Information

### DIFF
--- a/.github/workflows/agw-workflow.yml
+++ b/.github/workflows/agw-workflow.yml
@@ -146,20 +146,11 @@ jobs:
     name: C/C++ unit tests with Bazel
     runs-on: ubuntu-latest
     steps:
-      - name: Maximize build space
-        shell: bash
-        run: |
-          echo "Available storage before:"
-          df -h
-          echo "Removing /usr/share/dotnet, /usr/local/lib/android, /opt/ghc"
-          sudo rm -rf /usr/share/dotnet
-          sudo rm -rf /usr/local/lib/android
-          sudo rm -rf /opt/ghc
-          echo "Available storage after:"
-          df -h
       - name: Check Out Repo
         # This is necessary for overlays into the Docker container below.
         uses: actions/checkout@7884fcad6b5d53d10323aee724dc68d8b9096a2e # pin@v2
+      - name: Maximize build space
+        uses: ./.github/workflows/composite/maximize-build-space
       - name: Setup Bazel Base Image
         uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 # pin@v3
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/composite/maximize-build-space/action.yml
```diff
@@ -0,0 +1,18 @@
+---
+name: maximize-build-space
+description: Maximize build space on github runner
```
Neudrino: Should the name `Github` be capitalized?

.github/workflows/composite/maximize-build-space/action.yml
```diff
@@ -0,0 +1,18 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: Maximize build space
+      shell: bash
+      run: |
+        echo "Available storage before:"
+        df -h
+        echo "Removing /usr/share/dotnet, /usr/local/lib/android, /opt/ghc"
```
Neudrino, line 15: Is that not redundant documentation?

.github/workflows/composite/maximize-build-space/action.yml
```diff
@@ -0,0 +1,18 @@
...
+        sudo rm -rf /usr/share/dotnet
+        sudo rm -rf /usr/local/lib/android
+        sudo rm -rf /opt/ghc
```
Neudrino: - Do we have no password sudo permissions on the runner?
- Should we delete all folders in one call using line breaks?

jheidbrink (approved):

.github/workflows/agw-workflow.yml
```diff
@@ -146,20 +146,10 @@ jobs:
...
-          echo "Removing /usr/share/dotnet, /usr/local/lib/android, /opt/ghc"
-          sudo rm -rf /usr/share/dotnet
-          sudo rm -rf /usr/local/lib/android
-          sudo rm -rf /opt/ghc
-          echo "Available storage after:"
-          df -h
       - name: Check Out Repo
         # This is necessary for overlays into the Docker container below.
         uses: actions/checkout@7884fcad6b5d53d10323aee724dc68d8b9096a2e # pin@v2
+      - uses: ./.github/workflows/composite/maximize-build-space
```
jheidbrink: Nitpick: All the other steps here have a `name: ` entry with a descriptive name, we could adhere to that pattern.

.github/workflows/composite/maximize-build-space/action.yml
LKreutzer: The `rm` calls take ~1min which does not produce any logging, so I think it is justified to print one line to give feedback.

.github/workflows/composite/maximize-build-space/action.yml
```diff
@@ -0,0 +1,18 @@
+---
+name: maximize-build-space
```
jheidbrink, line 2: Nitpick: We could include a short rationale why we don't use an existing action for this (as far as I'm concerned the reason is that this one does less and is simpler)

.github/workflows/composite/maximize-build-space/action.yml
LKreutzer: Done

.github/workflows/composite/maximize-build-space/action.yml
LKreutzer: 1. `sudo` is needed
2. Done

.github/workflows/agw-workflow.yml
LKreutzer: Done. (Currently almost all local action calls do not have a name field.)

.github/workflows/composite/maximize-build-space/action.yml
LKreutzer: Done

.github/workflows/composite/maximize-build-space/action.yml
Neudrino: Sorry I missed a dash in the question. It should have been:
1. Do we have no-password sudo permissions on the runner?
I.e. if we can use `sudo` for the active user without password confirmation.

LKreutzer: Yes, it works as it is.

Neudrino (approved):

PR #13344, opened 2022-07-19
chore(lte): Bazelify the S1AP extended tests

### LINKED ISSUE #13417: Bazelify the S1AP extended tests

vktng:
## Summary

This PR bazelifies the extended S1AP integration tests and introduces a wrapper script for executing them.
The script has multiple ways to be executed, these are detailed in the help message:
`$MAGMA_ROOT/bazel/scripts/run_integ_tests.sh --help`
```
Executes all integration tests.
Usage:
   run_integ_tests.sh
      Execute all integration tests in the magma repository.
   run_integ_tests.sh path_to_tests_directory:bazel_test_target_name
      Execute the specified test.
   run_integ_tests.sh --list
      List all integration tests.
   run_integ_tests.sh --list-extended
      List all tests that use the traffic server.
   run_integ_tests.sh --list-traffic-server
      List the extended integration tests.
   run_integ_tests.sh --setup-extended
      Execute the setup test for the extended tests.
   run_integ_tests.sh --teardown-extended
      Execute the teardown test for the extended tests.
   run_integ_tests.sh --skip-setup-teardown-extended
      Execute all tests in the magma repository,
      except the setup and teardown for extended tests.
   run_integ_tests.sh --skip-setup-teardown-extended path_to_tests_directory:bazel_test_target_name
      Execute the specified test, without executing
      the setup and teardown for extended tests.
   run_integ_tests.sh --help
      Display this help message.
```

## Test Plan

Use the script as mentioned above.
The script runs on the magma_test VM and it needs the same preparation as if the integration tests were executed with make. More info about the setup here: https://docs.magmacore.org/docs/lte/s1ap_tests#s1ap-integration-tests

## Additional Information

### DIFF
--- a/bazel/python_test.bzl
+++ b/bazel/python_test.bzl
@@ -34,7 +34,7 @@ from bazel.python_utils.coverage_decorator import coverage_decorator
 
 if __name__ == "__main__":
     with coverage_decorator():
-        args =  ["-ra", "-vv"]  + %s + sys.argv[1:] + %s
+        args =  ["-s", "-vv"]  + %s + sys.argv[1:] + %s
         sys.exit(pytest.main(args))"""
 
 PYTEST_DEPS = [
--- a/bazel/scripts/run_integ_tests.sh
+++ b/bazel/scripts/run_integ_tests.sh
@@ -1,0 +1,261 @@
+#!/usr/bin/env bash
+
+################################################################################
+# Copyright 2022 The Magma Authors.
+
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+set -euo pipefail
+
+###############################################################################
+# FUNCTION DECLARATIONS
+###############################################################################
+
+help() {
+    echo "Executes all integration tests."
+    echo "Usage:"
+    echo "   $(basename "$0")" 
+    echo "      Execute all integration tests in the magma repository." 
+    echo "   $(basename "$0") path_to_tests_directory:bazel_test_target_name"
+    echo "      Execute the specified test."
+    echo "   $(basename "$0") --list"
+    echo "      List all integration tests."
+    echo "   $(basename "$0") --list-extended"
+    echo "      List the extended integration tests."
+    echo "   $(basename "$0") --list-traffic-server"
+    echo "      List all integration tests that use the traffic server."
+    echo "   $(basename "$0") --setup-extended"
+    echo "      Execute the setup test for the extended tests."
+    echo "   $(basename "$0") --teardown-extended"
+    echo "      Execute the teardown test for the extended tests."
+    echo "   $(basename "$0") --skip-setup-teardown-extended"
+    echo "      Execute all integration tests in the magma repository," 
+    echo "      except the setup and teardown for extended tests." 
+    echo "   $(basename "$0") --skip-setup-teardown-extended path_to_tests_directory:bazel_test_target_name"
+    echo "      Execute the specified test, without executing"
+    echo "      the setup and teardown for extended tests." 
+    echo "   $(basename "$0") --help"
+    echo "      Display this help message."
+}
+
+categorize_test() {
+    local TARGET=$1
+    if bazel query "attr(tags, extended_test, kind(py_test, ${TARGET}))";
+    then
+        EXTENDED_TEST_TARGETS=( "${TARGET}" )
+    else
+        echo "ERROR: Could not categorize the provided test."
+        exit 1
+    fi
+    # TODO: other tests
+}
+
+create_test_targets() {
+    if [[ "${TARGET_PATH}" == *":"* ]];
+    then
+        echo "Single target specified - running test:"
+        categorize_test "${TARGET_PATH}"
+    else
+        echo "Multiple targets specified - running tests:"
+        create_extended_test_targets
+        # TODO: other tests
+    fi
+    if [[ "${#EXTENDED_TEST_TARGETS[@]}" -eq 0 ]]; # TODO other tests
+    then
+        echo "ERROR: No test found."
+        help
+        exit 1
+    fi
+    for TARGET in "${EXTENDED_TEST_TARGETS[@]}"
+    do
+        echo "${TARGET}"
+    done
+}
+
+create_extended_test_targets() {
+    mapfile -t EXTENDED_TEST_TARGETS < <(bazel query "attr(tags, extended_test, kind(py_test, //lte/gateway/python/integ_tests/s1aptests/...))")
+}
+
+list_all_tests() {
+    echo "All integration tests:"
+    list_extended_tests
+    # TODO: other tests
+    exit 0
+}
+
+list_extended_tests() {
+    echo "Extended tests:"
+    create_extended_test_targets
+    for TARGET in "${EXTENDED_TEST_TARGETS[@]}"
+    do
+        echo "${TARGET}"
+    done
+}
+
+list_traffic_server_tests() {
+    echo "Tests that require the traffic server:"
+    bazel query "attr(tags, traffic_server_test, kind(py_test, //lte/gateway/python/integ_tests/s1aptests/...))"
+    exit 0
+}
+
+setup_extended_tests() {
+    echo "Setting up the environment for the extended tests."
+    echo "Building..."
+    bazel build "//lte/gateway/python/integ_tests/s1aptests:test_modify_mme_config_for_sanity" --define=on_magma_test=1
+    echo "Executing..."
+    sudo "${MAGMA_ROOT}/bazel-bin/lte/gateway/python/integ_tests/s1aptests/test_modify_mme_config_for_sanity"
+    echo "Setup finished successfully."
+}
+
+teardown_extended_tests() {
+    if [[ -f "${EXTENDED_TEST_CLEANUP_FILE_NAME}" ]];
+    then
+        echo "Cleaning up the environment after the extended tests."
+        echo "Building..."
+        bazel build "//lte/gateway/python/integ_tests/s1aptests:test_restore_mme_config_after_sanity" --define=on_magma_test=1
+        echo "Executing..."
+        sudo "${MAGMA_ROOT}/bazel-bin/lte/gateway/python/integ_tests/s1aptests/test_restore_mme_config_after_sanity"
+        echo "Cleanup finished successfully."
+    else
+        echo "No backup file found, skipping cleanup."
+    fi
+}
+
+run_test() {
+    local TARGET=$1
+    local TARGET_PATH=${TARGET%:*}
+    local SHORT_TARGET=${TARGET#*:}
+    (
+        echo "BUILDING TEST: ${TARGET}"
+        set -x
+        bazel build "${TARGET}" --define=on_magma_test=1
+        set +x
+        echo "RUNNING TEST: ${TARGET}"
+        set -x
+        sudo "${MAGMA_ROOT}/bazel-bin/${TARGET_PATH}/${SHORT_TARGET}"
+    )
+}
+
+print_summary() {
+    local NUM_SUCCESS=$1
+    local TOTAL_TESTS=$2
+    echo "SUMMARY: ${NUM_SUCCESS}/${TOTAL_TESTS} tests were successful."
+    for TARGET in "${!TEST_RESULTS[@]}"
+    do
+        echo "  ${TARGET}: ${TEST_RESULTS[${TARGET}]}"
+    done
+}
+
+###############################################################################
+# SCRIPT SECTION
+###############################################################################
+
+declare -a EXTENDED_TEST_TARGETS
+declare -A TEST_RESULTS
+NUM_SUCCESS=0
+NUM_RUN=1
+
+cd "${MAGMA_ROOT}"
+
+EXTENDED_TEST_SETUP="lte/gateway/python/integ_tests/s1aptests:test_modify_mme_config_for_sanity"
+EXTENDED_TEST_TEARDOWN="lte/gateway/python/integ_tests/s1aptests:test_restore_mme_config_after_sanity"
+SKIP_EXTENDED_SETUP_AND_TEARDOWN="false"
+EXTENDED_TEST_CLEANUP_FILE_NAME="${MAGMA_ROOT}/lte/gateway/configs/templates/mme.conf.template.bak"
+
+declare -a POSITIONAL_ARGS
+
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    --list)
+      list_all_tests
+      ;;
+    --list-extended)
+      list_extended_tests
+      exit 0
+      ;;
+    --list-traffic-server)
+      list_traffic_server_tests
+      ;;
+    --setup-extended)
+      setup_extended_tests
+      exit 0
+      ;;
+    --teardown-extended)
+      teardown_extended_tests
+      exit 0
+      ;;
+    --skip-setup-teardown-extended)
+      SKIP_EXTENDED_SETUP_AND_TEARDOWN="true"
+      shift
+      ;;
+    --help)
+      help
+      exit 0
+      ;;
+    --*|-*)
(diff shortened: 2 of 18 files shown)


bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,175 @@
...
+
+###############################################################################
+# FUNCTION DECLARATIONS
+###############################################################################
+
+help() {
+    echo "Executes all bazelified integration tests."
+    echo "Usage:"
+    echo "   $(basename "$0")" 
+    echo "      Execute all tests in the magma repository."
```
LKreutzer: Do you really mean "all tests in the magma repository" including all unit/sudo tests?

vktng: I specified which tests in line 23 and in line 26 I'm referring to all those tests. I see now that this might be confusing.

bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,239 @@
...
+        exit 1
+    fi
+    # TODO: other tests
+}
+
+create_test_targets() {
+    if [[ "${TARGET_PATH}" == *":"* ]];
+    then
+        echo "Single target specified - running test:"
+        categorise_test "${TARGET_PATH}"
```
LKreutzer: We should fix the spelling here too, otherwise the function name is wrong: should be `categorize_test "${TARGET_PATH}"`

LKreutzer (changes requested):

bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,239 @@
...
+    echo "   $(basename "$0")" 
+    echo "      Execute all integration tests in the magma repository." 
+    echo "   $(basename "$0") path_to_tests_directory:bazel_test_target_name"
+    echo "      Execute the specified test."
+    echo "   $(basename "$0") --list"
+    echo "      List all integration tests."
+    echo "   $(basename "$0") --list-extended"
+    echo "      List all tests that use the traffic server."
+    echo "   $(basename "$0") --list-traffic-server"
+    echo "      List the extended integration tests."
```
LKreutzer, line 32: I think line 32 and 34 got mixed up somehow?

LKreutzer (changes requested):

bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,239 @@
...
+    echo "   $(basename "$0") --setup-extended"
+    echo "      Execute the setup test for the extended tests."
+    echo "   $(basename "$0") --teardown-extended"
+    echo "      Execute the teardown test for the extended tests."
+    echo "   $(basename "$0") --skip-setup-teardown-extended"
+    echo "      Execute all tests in the magma repository,"
```
LKreutzer: Perhaps we can insert another "integration" in this line?

LKreutzer (changes requested):

bazel/scripts/run_integ_tests.sh
vktng: Yes, I did not notice.

bazel/scripts/run_integ_tests.sh
vktng: Agreed, I forgot to change it.

vktng: Fixed.

bazel/scripts/run_integ_tests.sh
vktng: Fixed.

bazel/scripts/run_integ_tests.sh
vktng: Fixed.

nstng:
After some experiments I agree that the tests are really flaky :) But I got them green without any make usage. I have the feeling that there is some statefullness, i.e., especially the traffic server tests need some other test to be executed first. But I expect this to be independent of Bazel.

One last finding: executing all tests seems to execute the restore test twice - the second time fails (maybe the setup test is also executed twice - without failure).

bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,239 @@
...
+    echo "Cleanup finished successfully."
+}
+
+run_test() {
+    local TARGET=$1
+    local TARGET_PATH=${TARGET%:*}
+    local SHORT_TARGET=${TARGET#*:}
+    (
+        set -x
+        echo "BUILDING TEST: ${TARGET}"
```
LKreutzer, line 137: Not super important: 
The commands are also printed and because of this the echo is printed first as a command and then again in the execution. Is there something that can be done to avoid these doubled outputs?

Screenshot from 2022-07-22 14-48-58

bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,239 @@
...
+        exit 1
+    fi
+    for TARGET in "${EXTENDED_TEST_TARGETS[@]}"
+    do
+        echo "${TARGET}"
+    done
+}
+
+create_extended_test_targets() {
+    mapfile -t EXTENDED_TEST_TARGETS < <(bazel query "attr(tags, extended_test, kind(py_test, //lte/gateway/python/integ_tests/s1aptests/...))")
```
LKreutzer, line 84: :heavy_check_mark: Checked that all extended tests are covered.

bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,239 @@
...
+
+TOTAL_TESTS=${#EXTENDED_TEST_TARGETS[@]}
+# TODO TOTAL_TESTS=$((TOTAL_TESTS + PRECOMMIT))
+
+if [[ "${SKIP_EXTENDED_SETUP_AND_TEARDOWN}" == "false" ]];
+then
+    setup_extended_tests
+fi
+
+for TARGET in "${EXTENDED_TEST_TARGETS[@]}"
```
LKreutzer, line 240: I think it would be nice if there was an option to rerun failing tests for a given number of times. 

In the integ test workflow we should have at least a 3x retry on failure. If we want to simply execute the script in the workflow then it would be nice if there was at least the option to do: `$MAGMA_ROOT/bazel/scripts/run_integ_tests.sh --retry_on_failure` which could run all tests and retry 3x on failure. That should not add much complexity to the script - mainly calculating the success quota becomes somewhat more difficult. What do you think?

bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,239 @@
...
+    done
+}
+
+###############################################################################
+# SCRIPT SECTION
+###############################################################################
+
+TARGET_PATH="${1:-}"
+
+if [[ "${TARGET_PATH}" == "--help" ]];
```
LKreutzer: Have we thought about using a separate tool or other mechanism for parsing the command line arguments? Right now the order is also relevant, e.g. when adding the `--skip-setup-teardown-extended` option (line 202ff) it does not work if this is the second argument and a test is passed on before. 

But it might add too much complexity. 

Some ideas can be found here:
- https://github.com/matejak/argbash
- https://stackoverflow.com/questions/192249/how-do-i-parse-command-line-arguments-in-bash

bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,239 @@
...
+    exit 0
+fi
+
+if [[ "${TARGET_PATH}" == "--list-traffic-server" ]];
+then
+    list_traffic_server_tests
+    exit 0
+fi
+
+if [[ "${TARGET_PATH}" == "--setup-extended" || "${TARGET_PATH}" == "${EXTENDED_TEST_SETUP}" ]];
```
LKreutzer: Just an observation: I think this relies on the fact that users always enter the bazel target with a `//` at the beginning. Same for line 196.

In fact running `$MAGMA_ROOT/bazel/scripts/run_integ_tests.sh lte/gateway/python/integ_tests/s1aptests:test_restore_mme_config_after_sanity` first does the setup job and then fails.

bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,239 @@
...
+        TEST_RESULTS["${TARGET}"]="PASSED"
+    else
+        TEST_RESULTS["${TARGET}"]="FAILED"
+    fi
+    NUM_RUN=$((NUM_RUN + 1))
+done
+
+if [[ "${SKIP_EXTENDED_SETUP_AND_TEARDOWN}" == "false" ]];
+then
+    teardown_extended_tests
```
LKreutzer, line 255: When running all tests the teardown seems to fail and I don't get to see the test summary.
Screenshot from 2022-07-22 15-51-15

- Maybe it would be better to first print the summary and then attempt the teardown?
- Do we know why the teardown fails in this case?

bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,239 @@
...
+
+if [[ "${TARGET_PATH}" == "--setup-extended" || "${TARGET_PATH}" == "${EXTENDED_TEST_SETUP}" ]];
+then
+    setup_extended_tests
+    exit 0
+fi
+
+if [[ "${TARGET_PATH}" == "--teardown-extended" || "${TARGET_PATH}" == "${EXTENDED_TEST_TEARDOWN}" ]];
+then
+    teardown_extended_tests
```
LKreutzer, line 226: It would be great if running `$MAGMA_ROOT/bazel/scripts/run_integ_tests.sh //lte/gateway/python/integ_tests/s1aptests:test_restore_mme_config_after_sanity` would not fail if there is nothing to restore. Perhaps we can check for the existence of the config files? If there are no config files we simply skip?

LKreutzer (changes requested):

bazel/python_test.bzl
```diff
@@ -34,7 +34,7 @@ from bazel.python_utils.coverage_decorator import coverage_decorator
 
 if __name__ == "__main__":
     with coverage_decorator():
-        args =  ["-ra", "-vv"]  + %s + sys.argv[1:] + %s
```
LKreutzer, line 37: Why is this needed?

lte/gateway/python/integ_tests/s1aptests/util/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+load("@python_deps//:requirements.bzl", "requirement")
+load("@rules_python//python:defs.bzl", "py_library")
+
+package(default_visibility = ["//visibility:public"])
+
+MAGMA_ROOT = "../../../../../"
+
+ORC8R_ROOT = "{}orc8r/gateway/python".format(MAGMA_ROOT)
+
+LTE_ROOT = "{}lte/gateway/python".format(MAGMA_ROOT)
```
LKreutzer: The roots are not used in this file and can be removed.

lte/gateway/python/integ_tests/s1aptests/BUILD.bazel
```diff
@@ -0,0 +1,489 @@
...
+pytest_test(
+    name = "test_implicit_detach_timer_with_mme_restart",
+    size = "large",
+    srcs = ["test_implicit_detach_timer_with_mme_restart.py"],
+    imports = [LTE_ROOT],
+    tags = TAG_EXTENDED_TEST,
+    deps = [":s1ap_wrapper"],
+)
+
+pytest_test(
```
LKreutzer, line 479: Do we really want the setup and tear-down to be modeled as a `pytest_test` target? If we execute all tests with `bazel ...` it will not execute due to the manual tag, but any query for e.g. `bazel query "kind(pytest_.*, //orc8r/... union //lte/...)"` does return these as well.

LKreutzer (changes requested):

LKreutzer:
In general I think it is working well, it is much more comfortable to run individual tests like this! 

In particular when running some specific tests:
- Setup: `$MAGMA_ROOT/bazel/scripts/run_integ_tests.sh --setup-extended`
- Run some tests: `$MAGMA_ROOT/bazel/scripts/run_integ_tests.sh --skip-setup-teardown-extended //lte/gateway/python/integ_tests/s1aptests:test_guti_attach_with_zero_mtmsi`
- ...
- Tear down: `$MAGMA_ROOT/bazel/scripts/run_integ_tests.sh --teardown-extended`

Actually I do not see any flakiness when running the extended tests, I could not get a single test to fail (I had a fresh VM and the tcp-checksumming was disabled everywhere).

bazel/python_test.bzl
vktng: This makes pytest show the output in real time instead of only after failures.

bazel/scripts/run_integ_tests.sh
vktng: In my experience with these tests just running them again does not really help, usually the problem is resolved when an other test is run. (Running test_attach_detach.py usually helps with failing tests.)

bazel/scripts/run_integ_tests.sh
vktng: Fixed.

bazel/scripts/run_integ_tests.sh
vktng: Good idea. Fixed.

bazel/scripts/run_integ_tests.sh
vktng: Fixed.

bazel/scripts/run_integ_tests.sh
vktng: I would guess that you deleted the backup file or stashed it in git during the long execution. The teardown will only run if there was a setup before, except when you execute the teardown direcly, but then you do not have a summary to worry about. I added an extra check, to only clean up if there is a backup file. I do not think putting it after the summary is worth it, this way you do not have to scroll for the results.

bazel/scripts/run_integ_tests.sh
vktng: Fixed.

lte/gateway/python/integ_tests/s1aptests/util/BUILD.bazel
vktng: Fixed.

lte/gateway/python/integ_tests/s1aptests/BUILD.bazel
vktng: What would you suggest as an alternative?
In the query you can exclude manual tests.
The setup and teardown are still tests if you look at the code. IMO if you want to change it to something else you should change the code too to be a normal script.

vktng:
> After some experiments I agree that the tests are really flaky :) But I got them green without any make usage. I have the feeling that there is some statefullness, i.e., especially the traffic server tests need some other test to be executed first. But I expect this to be independent of Bazel.
> 
> One last finding: executing all tests seems to execute the restore test twice - the second time fails (maybe the setup test is also executed twice - without failure).

I could not repoduce your finding (executing the teardown twice). I have also made a few changes since then, now even if you execute it twice it will not fail.

bazel/scripts/run_integ_tests.sh
```diff
@@ -0,0 +1,261 @@
...
+      ;;
+    --skip-setup-teardown-extended)
+      SKIP_EXTENDED_SETUP_AND_TEARDOWN="true"
+      shift
+      ;;
+    --help)
+      help
+      exit 0
+      ;;
+    -*|--*)
```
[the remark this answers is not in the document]
LKreutzer: Perhaps we could use `--*|-*)` here to avoid the warning? Otherwise this will keep showing up with every modification of this file.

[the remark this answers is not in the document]
LKreutzer: Same as above.

bazel/scripts/run_integ_tests.sh
LKreutzer: I still think it would be a nice-to-have option, but we can always add it later on when we need it.

lte/gateway/python/integ_tests/s1aptests/BUILD.bazel
LKreutzer: Since we have the specific tags 
```
    "TAG_EXTENDED_TEST_SETUP",
    "TAG_EXTENDED_TEST_TEARDOWN",
```
I guess we are distinguishing them already. Agree, perhaps we should create a task to refactor these?

LKreutzer (changes requested):
See comment on the shellcheck findings.

LKreutzer (approved):
Tested locally

nstng (approved):
lgtm!

ajahl (approved):

PR #13347, opened 2022-07-20
fix(ci): AWS AMI Build fixed

ShubhamTatvamasi:
## Additional Information


experimental/cloudstrapper/playbooks/roles/magma-agw/tasks/magma-setup.yaml
```diff
@@ -14,11 +14,16 @@
 
 - name: Add magma user to wheel group
   user:
-    name=magma
-    groups=magma
-    append=yes
-    state=present
-    createhome=yes
```
ShubhamTatvamasi, line 21: Default values for **state** is `present` and for **createhome** is `yes`, so removing these two.

PR #13359, opened 2022-07-20
feat(dp): Add CRL validation for SAS SSL certificate

hejjestemjarek:
Signed-off-by: Jarosław Jaszczuk <jaroslaw.jaszczuk@freedomfi.com>

## Summary
- Added new `CRLValidator` module responsible for validating SSL certificates against their CRLs 
- Certificates are cached for set amount of time and updated in the background
- When making requests to SAS `RequestRouter` can use `CRLValidator` to validate SAS SSL certificate

## Test Plan
- Added unit tests

## Additional Information


dp/cloud/python/magma/configuration_controller/crl_validator/certificate.py
```diff
@@ -0,0 +1,103 @@
...
+        hostname: host from which certificate will be retrieved
+        port: SSL port of the host
+
+    Returns:
+        x509.Certificate
+    """
+    context = ssl.SSLContext(protocol=ssl.PROTOCOL_SSLv23)
+    context.minimum_version = ssl.TLSVersion.TLSv1_2
+
+    sock = socket.create_connection((hostname, port))
```
jkmar: What if there is an exception somewhere in the middle of this function?
The socket won't be closed, so I think we use `with`

dp/cloud/python/magma/configuration_controller/tests/unit/test_crl_validator.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+@requests_mock.Mocker()
+@freeze_time(CERTIFICATES_DATE)
+class CRLValidatorTestCase(TestCase):
+    sas_url = 'https://fake.sas.url/'
+
+    @parameterized.expand([
+        (VALID_CRL_CERT, VALID_CERT_CRLS_DATA),
+        (NO_CRL_CERT, {}),
```
jkmar, line 45: No CRL is valid?

dp/cloud/python/magma/configuration_controller/crl_validator/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""
```
jkmar, line 1: [dp/cloud/python/magma/__init__.py](https://github.com/magma/magma/pull/13359/files#diff-b6ec112d156f2e00319f5f0cdb1c6aacc529d290b25d023e9b6d16d64504d4d6) is missing header

dp/cloud/python/magma/configuration_controller/tests/unit/test_crl_validator.py
```diff
@@ -0,0 +1,149 @@
...
+
+        update_rate = 1  # seconds
+        validator = self._create_validator(urls=[self.sas_url], update_rate=update_rate)
+        valid = validator.is_valid(url=self.sas_url)
+        self.assertIs(valid, True)
+
+        # When
+        self._set_mocket_cert(certificate=REVOKED_CRL_CERT)
+        self._set_mocker_CRL(mocker=mocker, crls=REVOKED_CERT_CRLS_DATA)
+        time.sleep(update_rate + 1)  # Wait for certs update.
```
jkmar: can we mock time, since sleep is flake and can possible finish faster that expected when some signal is received

dp/cloud/python/magma/configuration_controller/tests/unit/test_crl_validator.py
```diff
@@ -0,0 +1,149 @@
...
+        self._set_mocker_CRL(mocker=mocker, crls=VALID_CERT_CRLS_DATA)
+
+        update_rate = 1  # seconds
+        validator = self._create_validator(urls=[self.sas_url], update_rate=update_rate)
+        valid = validator.is_valid(url=self.sas_url)
+        self.assertIs(valid, True)
+
+        # When
+        self._set_mocket_cert(certificate=REVOKED_CRL_CERT)
+        self._set_mocker_CRL(mocker=mocker, crls=REVOKED_CERT_CRLS_DATA)
```
jkmar, line 96: can we also check that after setting those certs nothing happened yet (cert is still valid)?

dp/cloud/python/magma/configuration_controller/crl_validator/crl_validator.py
```diff
@@ -0,0 +1,162 @@
...
+    ) -> List[x509.CertificateRevocationList]:
+        """ Get cached CertificateRevocationLists for given certificate, fetch new if they does not exist.
+
+        Args:
+            certificate: certificate that CRLs were attached to
+
+        Returns:
+            list of Certificate Revocation Lists
+        """
+        serial_number = certificate.serial_number
```
jkmar: possible race condition between updating certs and reading them

dp/cloud/python/magma/configuration_controller/crl_validator/crl_validator.py
```diff
@@ -0,0 +1,162 @@
...
+        self._update_rate = update_rate
+
+    def run(self) -> None:
+        """ Start thread.
+
+        Returns: None
+        """
+        while True:
+            self._update_certificates()
+            time.sleep(self._update_rate)
```
jkmar: suggest using threading.Event or threading.Timer

dp/cloud/python/magma/configuration_controller/crl_validator/certificate.py
hejjestemjarek: Socket connection is now used as context manager.

dp/cloud/python/magma/configuration_controller/tests/unit/test_crl_validator.py
hejjestemjarek: There are sites that have certificates which do not provide CRLs but are perfectly valid, `stackoverflow.com` or `letsencrypt.org` could be an examples. They may rely on OCSP or some other solution which might be a better idea, but certification process requires us to check CRLs specifically. Although CRLs are not mandatory in any way, so we cannot assume that certificate is not valid because it does not supply CRLs.

dp/cloud/python/magma/configuration_controller/tests/unit/test_crl_validator.py
hejjestemjarek: Added validity check after setting new certificate, but before it's updated in `CRLValidator`.

dp/cloud/python/magma/configuration_controller/crl_validator/crl_validator.py
hejjestemjarek: Switched to using `threading.Timer`.

dp/cloud/python/magma/configuration_controller/crl_validator/crl_validator.py
hejjestemjarek: We're caching certificates and their CRLs for minutes anyway, so I don't consider this a problem.   
We might get outdated result on this specific check while certs are being updated, but we were getting outdated result for the last 5 minutes because of the cache.

dp/cloud/python/magma/configuration_controller/tests/unit/test_crl_validator.py
hejjestemjarek: We could mock the timer in the updater thread, but then certs would get instant updated and we couldn't test if it's happening in the background and in the set intervals of time. We can't really mock sleep in the test, since we have to actually wait for the updater thread to update the certs.
 
We could change the test to rule out flakiness:
 ```python
        sas_host = get_host(url=self.sas_url)
        old_certificate = validator._certificates[sas_host]
 
        # When / Then
        self._set_mocket_cert(certificate=REVOKED_CRL_CERT)
        self._set_mocker_CRL(mocker=mocker, crls=REVOKED_CERT_CRLS_DATA)

        # Ensure certs are not instant updated
        valid = validator.is_valid(url=self.sas_url)
        self.assertIs(valid, True)

        # Set timeout in case certs won't get updated at all
        def _timeout(*_) -> None:
            raise TimeoutError('Timed out waiting for certs update')
        signal.signal(signal.SIGALRM, _timeout)
        signal.alarm(10)

        # Wait for updater to refresh certs
        while validator._certificates[sas_host] == old_certificate:
            time.sleep(update_rate)
        signal.alarm(0)    # Disable alarm after successful update

        # Ensure certs are updated
        with self.assertRaises(SSLError):
            validator.is_valid(url=self.sas_url)
 ```
but we would still have to wait and additionally check for timeout to not wait for infinity if certs would not get updated at all for whatever reason.

dp/cloud/python/magma/configuration_controller/crl_validator/__init__.py
hejjestemjarek: Removed the file, as it's unnecessary and was pushed by mistake.

dp/cloud/python/magma/configuration_controller/crl_validator/crl_validator.py
```diff
@@ -0,0 +1,171 @@
...
+        self._timer: threading.Timer
+
+    def run(self) -> None:
+        """ Update certificates and set timer for the next update.
+        The timer calls this exact method, so it automatically sets new timer after each update.
+
+        Returns: None
+        """
+        self._update_certificates()
+        self._set_update_timer()
```
jkmar: I think we should run this just once - the timer would then wake up the tread once in every `self._update_rate` seconds

dp/cloud/python/magma/configuration_controller/crl_validator/crl_validator.py
jkmar: I'm not sure it's that simple, because of two things:
1. Update is not doing atomic operations and also it is not guaranteed that even dict operations are thread safe (they are for example in cpython, but it is not guaranteed in any way), so for example you may be left in some inconsistent state (for example dict insert may require changing its internal structure and if you read something in the other thread it may not even the value you are looking for), which cal lead to weird undefined behavior. Also bug related to concurrency tend to be very difficult to debug, since they are not deterministic and also you have to consider multiple different scenarios of threads overlapping with each other
2. This function is doing not only reads, but also writes, so in fact if there is no cert for given serial number, then each call to this function would result in an additional query, which is what we were trying to avoid by introducing a background job

So I suggest a few things here:

1. Use mutex for `_crls` access/modification (or even better RWmutex, unfortunately not in python standard library), there are only up to 7 threads using it, so it should drop the performance that much, but I think it is very important to have thread safe code
2. Do not do any writes in this function, only reads. If we don't have cert for some serial number then, we don't get it, similarly (this is even the current behavior) if we have outdated cert we return it. This is the point of having background cert manager - not to query for new crls by ourselves.
4. This is more of a question - do you know how often CRLs needs to be updated? I've seen some guidelines saying that doing it once every 24h should be enough, so we don't have to worry about certs being outdated for 5 min and maybe we can even make refresh interval larger (for example 1h or more)

dp/cloud/python/magma/configuration_controller/tests/unit/test_crl_validator.py
jkmar: Hmm, maybe we can just test update_certificates method synchronously and run cert updater as job, like other entities in configuration controller, for example here:
``` 
    scheduler.add_job(
            process_requests,
            args=[consumer, processor, router, session_manager, fluentd_client],
            trigger=IntervalTrigger(
                seconds=config.REQUEST_PROCESSING_INTERVAL_SEC,
            ),
            max_instances=1,
            name=f"{req_type}_job",
        )
```
this would be consistent with current configuration controller design, and also would separate background job responsibility from updater.
So in the end updater would have just to methods:
- update certs
- get cert

And it would be run from run.py

dp/cloud/python/magma/configuration_controller/crl_validator/crl_validator.py
hejjestemjarek: `CertificatesUpdater` removed, configuration controller's `BackgroundScheduler` is now responsible for refershing certs.

dp/cloud/python/magma/configuration_controller/crl_validator/crl_validator.py
hejjestemjarek: 1. Added `rwmutex.RWLock` for read and write to `_certificates` and `_crls`.
2. Removed the writes from `is_valid` method, if we don't have specific cert then `KeyError` is raised, informing that this specific host's certs are not available since this host was not given on validator initialization and it's certs were not prefetched. 
3. I've also seen the 24h mark. I think 1h cache on our end is perfectly reasonable, changed cache time in config to 1h.

dp/cloud/python/magma/configuration_controller/tests/unit/test_crl_validator.py
hejjestemjarek: Removed `CertificatesUpdater` and delegated updating the certs to configuration controller's `BackgroundScheduler` as you suggested . It pretty much solved all the testing problems and simplified validator's code a lot.

jkmar (approved):

PR #13379, opened 2022-07-21
feat(dp): Add carrier aggregation logic to AMC

jkmar:
## Summary

Add carrier aggregation logic to AMC


xbend (approved):
LGTM

Issue #13245, opened 2022-07-08
magma_trfserver not starting up correctly on Mac OS


alexzurbonsen:
### Your Environment

- **Version:** master
- **Affected Component:** magma_trfserver
- **Deployment Environment:** Vagrant 2.2.19, Virtualbox 6.1.34 r150636, MacOS 11.6.5

### Describe the Issue

When running `vagrant up magma_trfserver` after destroying it (`vagrant destroy magma_trfserver`) on MacOS (but not on Linux), the process fails with the following error

```
==> magma_trfserver: Configuring and enabling network interfaces...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

/sbin/ifup 'eth1'
/sbin/ifup 'eth2'
/sbin/ifup 'eth3'

Stdout from the command:



Stderr from the command:

ifup: interface eth1 already configured
ifup: interface eth2 already configured
RTNETLINK answers: File exists
ifup: failed to bring up eth3
```

Repeating the `vagrant up magma_trfserver` finishes the provisioning but cannot set the symlink to traffic_server.py because the file does not exist.

```
TASK [trfserver : Create symlink for trfserver scripts] ************************
Friday 08 July 2022  11:40:00 +0200 (0:00:00.019)       0:00:04.219 *********** 
changed: [magma_trfserver] => (item=traffic_server.py) => {"ansible_loop_var": "item", "changed": true, "dest": "/usr/local/bin/traffic_server.py", "item": "traffic_server.py", "src": "/home/vagrant/magma/lte/gateway/deploy/roles/trfserver/files/traffic_server.py", "warnings": ["Cannot set fs attributes on a non-existent symlink target. follow should be set to False to avoid this."]}
[WARNING]: Cannot set fs attributes on a non-existent symlink target. follow
should be set to False to avoid this.
``` 

**Expected behavior**
should run without error, as it does on Linux. (Or when reverting commit #13010 locally)

alexzurbonsen:
Some observations:

the magma vm (ubuntu focal) has the same network interfaces
```
magma.vm.network "private_network", ip: "192.168.60.142", nic_type: "82540EM"
# iperf3 trfserver routable IP.
magma.vm.network "private_network", ip: "192.168.129.1", nic_type: "82540EM"
magma.vm.network "private_network", ip: "3001::10", nic_type: "82540EM"
```
as the trfserver (debian stretch)
```
[...]

PR #13385, opened 2022-07-21
chore(ci): github runner for build_all is bumped from macos10.15 to 12

### LINKED ISSUE #13337: Upgrade Github runners to macOS 12 due to deprecation of macOS 10.15
### Subtasks

- [x] Change runners on to the new macos version
- [x] Fix pyenv command and make sure traffic server runs on ubunu-20.04
- [x] Run action on private fork to make sure actions still work with macOS 12

### Describe the Issue

The Github runners with macOS 10.15 are being deprecated soon enough. This will lead to lower availability and in the consequence to longer waiting times for job execution.

#### Tests
LTE integ test run: https://github.com/crasu/magma/runs/7555541971?check_suite_focus=true
CWF integ test run: https://github.com/crasu/magma/runs/7559855082?check_suite_focus=true
FEG integ test run: https://github.com/crasu/magma/actions/runs/2752365494

#### Suggested Resolution

Upgrade any macOS runners in CI to version **12**.

#### Target date

The deprecation date is give as **2022-08-30**!

#### Additional context

- Find additional information and details about the deprecation at
https://github.com/actions/virtual-environments/issues/5583
- Find an example usage of macOS 10.15 runners at https://github.com/magma/magma/blob/master/.github/workflows/lte-integ-test.yml#L19.
- There will be several "brown-out" periods. You will see the following errors during that times with all runs failing on those runners.
grafik

nstng:

## Summary

**Context**
* github is deprecating macos-10.15 runners. Alternatives are macos-11 and 12 runners.
* This goes along with "brownouts" - one day per week macos-10.15 runners are not usable - see https://github.com/actions/virtual-environments/issues/5583
* Slack discussion: https://magmacore.slack.com/archives/C01MDUQDFC3/p1658411534798269
* We have 4(5) jobs using macos-10.15 runners
  * CWF
  * FED
  * LTE integ tests (make and bazel)
  * build_all
* CWF, FED, LTE are using the traffic server which has a known issue on macos >=11 - see https://github.com/magma/magma/issues/13245
  * here we need the base image bumped to ubuntu 20.04 - this is ongoing (issue link TBD)

**Here**
* the build_all job is not using the trf server and can be bumped
* we decided to use macos 12 because
  * its more recent
  * it already has vagrant and virtualbox pre-installed - the 11 image does not
* I also took the liberty to add a **workflow_dispatch** trigger for easy testing
* setup pyenv (gabrielfalcao/pyenv-action, with py 3.8.5) does not work with macos 11/12 - but it seems in a test run that it is not needed ... looks like tech debt -> removed

## Test Plan

* real change will only run on master first
* test run on branch (with tweaks to be runable): https://github.com/nstng/magma/actions/runs/2713097965

## Additional Information

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -2,6 +2,7 @@
 name: build-all
 
 on:
+  workflow_dispatch: null
   push:
     branches:
       - master
@@ -101,7 +102,7 @@ jobs:
           python ci-scripts/helm_repo_rotation.py
   agw-build:
     if: github.event_name == 'push' && github.repository_owner == 'magma'
-    runs-on: macos-10.15
+    runs-on: macos-12
     outputs:
       artifacts: ${{ steps.publish_packages.outputs.artifacts }}
     steps:
@@ -113,10 +114,6 @@ jobs:
         with:
           path: ~/.vagrant.d/boxes/magmacore-VAGRANTSLASH-magma_dev
           key: vagrant-box-magma-dev
-      - name: setup pyenv
-        uses: "gabrielfalcao/pyenv-action@5327db2939908b2ef8f62d284403d678c4b611d0" # pin@v8
-        with:
-          default: 3.8.5
       - uses: actions/setup-python@7f80679172b057fc5e90d70d197929d454754a5a # pin@v2
         with:
           python-version: '3.8.10'


Neudrino (approved):
> setup pyenv (gabrielfalcao/pyenv-action, with py 3.8.5) does not work with macos 11/12 - but it seems in a test run that it is not needed ... looks like tech debt -> removed

_Optional idea:_ If this is tech-depth, should we maybe create a ticket, to remove it from other setups as well?

nstng:
> _Optional idea:_ If this is tech-depth, should we maybe create a ticket, to remove it from other setups as well?
imho an extra issue is not needed as this needs to be done when the other jobs are bumped to macos-12 (and also it will be checked there if the pyenv is needed in those jobs)

LKreutzer (approved):

PR #13389, opened 2022-07-22
chore(dp): Remove old integration-tests job from CI

sklgromek:

## Summary
Removing old integration tests run without  orc8r as they are not necessary anymore

## Test Plan
- [ ] Check CI dp-workflow results

## Additional Information


jkmar (approved):

Neudrino (changes requested):
:heavy_check_mark: CI Workflow change looks good in principle.

However, the job is still required. It needs to be removed from the list of required jobs before merging this PR. Otherwise CI will be broken. Thus I will mark it red.
grafik

@maxhbr 
The check to be removed is probably named something out of:
- integration_tests_no_orc8r
- Domain proxy integration tests without orc8r

sklgromek:
> heavy_check_mark CI Workflow change looks good in principle.
> 
> However, the job is still required. It needs to be removed from the list of required jobs before merging this PR. Otherwise CI will be broken. Thus I will mark it red. grafik
> 
> @maxhbr The check to be removed is probably named something out of:
> 
> * integration_tests_no_orc8r
> * Domain proxy integration tests without orc8r

Indeed, I noticed that this check is required on this PR. It shouldn't be. We requested to mark "Domain proxy integration tests with orc8r" as the required check. That one probably was marked as required by mistake.
I just reopened the issue in which we requested to mark tests as required
https://github.com/magma/magma/issues/12855

Neudrino (approved):
As per https://github.com/magma/magma/issues/12855#issuecomment-1192468039 @maxhbr has removed the required check.

Thus approving from CI side.

PR #13392, opened 2022-07-22
fix(ci): path filter is always true is prevented

nstng:

## Summary

It seems a job guard of the example form `condition1 && ${{ condition2 }}` is problematic - and resulted in the case here to return always true. Assumption is that the syntax must not be mixed.
This is, use either `condition1 && condition2` or `${{ condition1 && condition2 }}`.

## Test Plan

CI

## Additional Information


LKreutzer (approved):
In pairing

PR #13394, opened 2022-07-22
feat(dp): Remove grant attempts

jkmar:
## Summary

Grant attemts have been obsolete since frequency preferences was introduced.


xbend (approved):
LGTM but please rebase.

Issue #11291, opened 2022-01-27
[NMS] magmalte_magmalte_1 container unhealthy and PostgreSQL database missing the "tabs" column from "Users" table


rogerpueyo:
Dear developers,

I am experiencing some trouble while deploying the NMS with Magma versions v1.5.3 and v1.6.1. In particular:
- the NMS's "magmalte_magmalte_1" container is in an unhealthy status
- the `dev_setup.sh` script fails to run

In the following I'm reporting the details for v1.6.1 only, but they are identical for v1.5.3.

## Environment
My environment is in a VM with 4 cores and 8 GB RAM, running Ubuntu 20.04 LTS with up-to-date packages and the required software (Docker, Docker Compose, etc.) from the prerequisites of the quick start guide.

## Procedure
1. Magma source code is at `~/magma`, on branch v1.6 (1 commit ahead of tag `v1.6.1`:
```
cd ~magma
git log
commit 80a9762a064b13a6f32758b174a4bb3242735b64 (HEAD -> v1.6, origin/v1.6)
Author: Pravin Shelar <pbshelar@fb.com>
Date:   Wed Nov 24 17:14:31 2021 -0800

    feat: draft: Add support for L3 transport for PGW traffic (#9750)
    
    * feat(agw): introduce AGW L3 transport (#8726)
    
    Following patch introduces SGi tunnel transport based on wireguard.
    Design doc: #8723
    
    This initial PR is targeted for NATed AGW.
    
    Signed-off-by: Pravin B Shelar <pbshelar@fb.com>
    
    * feat(agw): create wireguard tunnel for PGW. (#8944)
    
    Add config option for SGW to enable PGW packet routing
    over WG tunnel.
    
    Signed-off-by: Pravin B Shelar <pbshelar@fb.com>

commit 346e86d7588dd4f88cbc2bea337f1c4747ecfe00 (tag: v1.6.1)
Author: quentinDERORY <15911421+quentinDERORY@users.noreply.github.com>
Date:   Wed Nov 17 10:51:23 2021 -0800

    fix(ci): Backport fix deployment of agw (#10393)
    
    Will fix broken agw ami workflow
    
    Signed-off-by: quentinDERORY <15911421+quentinDERORY@users.noreply.github.com>
```
2. Build and run Orc8r:
```
cd ~/magma/orc8r/cloud/docker
./build.py
 [...]
./run.py
```
3. Orc8r is up and running, the containers are running and I can access the Swagger web GUI using the browser.
[...]

PR #13402, opened 2022-07-25
chore(nms): Upgrade sequelize and umzug

### LINKED ISSUE #13123: Upgrade sequelize
See https://sequelize.org/docs/v6/other-topics/upgrade/
There is also an alpha version https://github.com/sequelize/sequelize/milestone/20 but upgrading to this might be too early.

Make sure that database version we use watch the requirements https://sequelize.org/releases/

This was tried in the past already
https://github.com/magma/fbc-js-core/pull/153
Possible issue with this
https://github.com/magma/fbc-js-core/pull/159

thmsschmitt:
## Summary

Upgrades sequelize and umzug to the latest version.

- fixes #13123 

## Test Plan
* Create and edit organisations 
    * make sure you can navigate to the newly created organisations
* Destroy the NMS docker volume and set it up from scratch make sure the migrations are executed
    * Run `dev_setup.sh` 
    * Make sure `magma-test` exists and can be accessed  
    * Restart the NMS docker container with destroying the volume, make sure the migration are not executed


nms/shared/sequelize_models/index.ts
```diff
@@ -31,15 +31,12 @@ import sequelizeConfig from './sequelizeConfig';
 const env = process.env.NODE_ENV || 'development';
 const config = sequelizeConfig[env];
 
-export const sequelize: Sequelize =
-  env === 'test'
```
thmsschmitt, line 35: I realised that this is already happing through the `config`

nms/shared/sequelize_models/migrations/20190509000000-make-organization-unique.js
```diff
@@ -27,9 +27,10 @@ module.exports = {
    * @param {QueryInterface} queryInterface
    */
   up: queryInterface => {
-    return queryInterface.addConstraint('Organizations', ['name'], {
+    return queryInterface.addConstraint('Organizations', {
```
thmsschmitt, line 29: This was a breaking changed introduced to sequelize 6 and probably the root cause of this issue https://github.com/magma/magma/issues/11291

nms/scripts/createOrganization.ts
```diff
@@ -58,12 +58,10 @@ async function createOrUpdateOrganization(
   organizationObject: OrganizationObject,
 ) {
   const organization = await Organization.findOne({
-    where: {
```
thmsschmitt, line 61: This worked and still does although it is semantically somewhat incorrect. The doc are unfortunately really vage here. But with the upgrade the types definitions got better and this is now forbidden.

Neudrino (approved):

PR #13403, opened 2022-07-25
feat(nms): Add new style to the Login page

HannaFar:

## Summary

Add new style to the Login page

Organization portal sso enabled: 

Capture d’écran 2022-07-27 à 11 09 26

Organization portal: 
Capture d’écran 2022-07-27 à 11 08 56

Host portal: 
Capture d’écran 2022-07-27 à 11 09 05

## Additional Information

### DIFF
--- a/nms/app/components/FormField.tsx
+++ b/nms/app/components/FormField.tsx
@@ -55,6 +55,13 @@ const useStyles = makeStyles<Theme>(theme => ({
     fontWeight: 400,
     marginLeft: '8px',
   },
+  label: {
+    fontSize: '14px',
+    fontFamily: 'Inter',
+    fontStyle: 'normal',
+    fontWeight: 500,
+    lineHeight: '20px',
+  },
   secondaryHeading: {
     flexBasis: '66.66%',
   },
@@ -87,6 +94,7 @@ type Props = {
   isOptional?: boolean;
   // If true, the left and right padding is removed.
   disableGutters?: boolean;
+  className?: string;
 };
 
 export default function FormField(props: Props) {
@@ -115,9 +123,12 @@ export default function FormField(props: Props) {
 export function AltFormField(props: Props) {
   const classes = useStyles();
   return (
-    <ListItem dense={props.dense} disableGutters={props.disableGutters}>
+    <ListItem
+      dense={props.dense}
+      disableGutters={props.disableGutters}
+      className={props.className}>
       <Grid container>
-        <Grid item xs={12}>
+        <Grid item xs={12} className={classes.label}>
           {props.label}
           {props.isOptional && (
             <Typography
--- a/nms/app/views/login/LoginForm.tsx
+++ b/nms/app/views/login/LoginForm.tsx
@@ -19,27 +19,61 @@ import FormLabel from '@material-ui/core/FormLabel';
 import React from 'react';
 import Text from '../../theme/design-system/Text';
 import TextField from '@material-ui/core/TextField';
+
+import {AltFormField} from '../../components/FormField';
+import {colors} from '../../theme/default';
 import {withStyles} from '@material-ui/core/styles';
 
 const ENTER_KEY = 13;
 const styles = {
+  capitalize: {
+    textTransform: 'capitalize',
+  },
   card: {
     maxWidth: '400px',
-    margin: '10% auto 0',
+    margin: '24px auto 0',
+    padding: '20px 0',
+  },
+  cardContent: {
+    padding: '0 24px',
   },
   input: {
     display: 'inline-flex',
     width: '100%',
-    margin: '5px 0',
   },
   footer: {
-    padding: '18px 16px 16px',
-    float: 'right',
+    padding: '0 24px',
+  },
+  login: {
+    marginTop: '10%',
   },
   title: {
-    marginBottom: '16px',
     textAlign: 'center',
+    display: 'flex',
+    justifyContent: 'center',
+    margin: '12px auto 0',
+    flexDirection: 'column',
+    maxWidth: '400px',
+    alignItems: 'start',
+  },
+  formTitle: {
+    marginBottom: '8px',
+    textAlign: 'left',
     display: 'block',
+    fontSize: '20px',
+  },
+  FormField: {
+    paddingBottom: '0',
+  },
+  submitButton: {
+    width: '100%',
+    marginTop: '16px',
+  },
+  regular: {
+    color: colors.primary.mirage,
+  },
+  gray: {
+    color: colors.primary.gullGray,
   },
 } as const;
 
@@ -53,6 +87,9 @@ type Props = {
   ssoAction?: string;
 };
 
+const HOST_PORTAL_TITLE = 'Magma Host Portal';
+const ORGANIZATION_PORTAL_TITLE = 'Magma Organization Portal';
+
 class LoginForm extends React.Component<Props> {
   form: HTMLFormElement | null = null;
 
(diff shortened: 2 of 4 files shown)


nms/app/views/login/LoginForm.tsx
```diff
@@ -16,30 +16,62 @@ import Card from '@material-ui/core/Card';
...
+    marginBottom: '16px',
+    textAlign: 'left',
     display: 'block',
+    fontSize: '20px',
+  },
+  submitButtom: {
+    width: '100%',
+  },
+  regular: {
+    color: colors.common.black,
```
thmsschmitt: Pure black should never be used https://uxplanet.org/basicdesign-never-use-pure-black-in-typography-36138a3327a6
MaterialUI uses `rgba(0, 0, 0, 0.87);` if you don't define anything

nms/app/views/login/LoginForm.tsx
```diff
@@ -16,30 +16,62 @@ import Card from '@material-ui/core/Card';
...
+  },
+  gray: {
+    color: colors.grey[500],
```
thmsschmitt: We should use our colors i.e. `import {colors} from '../../theme/default';`

nms/app/views/login/LoginForm.tsx
```diff
@@ -65,67 +100,119 @@ class LoginForm extends React.Component<Props> {
     const params = new URLSearchParams(window.location.search);
     const to = params.get('to');
 
+    const hostname = window.location.hostname;
+    const organization: string = hostname.slice(0, hostname.indexOf('.'));
```
thmsschmitt: Nit `const [organization] = window.location.hostname.split('.', 1)`

nms/app/components/FormField.tsx
```diff
@@ -117,7 +124,7 @@ export function AltFormField(props: Props) {
   return (
```
thmsschmitt, line 125: I was wondering why everything looks a bit different/off compared to the rest of page.
Turns out we don't render a `CssBaseline` in login.tsx

Screenshot 2022-07-26 at 09 28 53

nms/app/views/login/LoginForm.tsx
```diff
@@ -16,30 +16,62 @@ import Card from '@material-ui/core/Card';
...
   },
   input: {
     display: 'inline-flex',
     width: '100%',
     margin: '5px 0',
   },
   footer: {
-    padding: '18px 16px 16px',
-    float: 'right',
+    padding: '0 24px',
```
thmsschmitt, line 45: I know this is pixel pushing but the distance between the input fields is much larger than the distance to the button.
May be can even this out a little

current
Screenshot 2022-07-26 at 09 31 01
new
Screenshot 2022-07-26 at 09 32 58

nms/app/views/login/LoginForm.tsx
```diff
@@ -65,67 +100,119 @@ class LoginForm extends React.Component<Props> {
     const params = new URLSearchParams(window.location.search);
     const to = params.get('to');
 
+    const hostname = window.location.hostname;
+    const organization: string = hostname.slice(0, hostname.indexOf('.'));
+    const hostPortal: boolean = organization === 'host';
+
     if (ssoEnabled) {
```
thmsschmitt: if SSO would be enabled it looks something like this 
Screenshot 2022-07-26 at 09 43 54

Neudrino (commented):
The NMS e2e test job is failing: https://github.com/magma/magma/runs/7498662277?check_suite_focus=true.

nms/app/views/login/LoginForm.tsx
```diff
@@ -65,67 +100,119 @@ class LoginForm extends React.Component<Props> {
...
-              color="primary">
-              Login
-            </Button>
-          </CardActions>
-        </form>
-      </Card>
+          </Grid>
+          <Grid item xs={12}>
+            <Card raised={true} className={classes.card}>
+              <form
```
thmsschmitt, line 148: Instead of duplicating the header for SSO we could do something like
```
<Card raised={true} className={classes.card}>
   {ssoEnabled ? ( ... ) : ( ... )}
```

Neudrino:
There are several warnings [in the yarn test job](https://github.com/magma/magma/runs/7498663012?check_suite_focus=true#step:6:367) about certain elements not used in the correct places. E.g. `alignItems`, `xs`. A good search term for the logs is `must be used on`.

nms/app/views/login/LoginForm.tsx
```diff
@@ -65,67 +100,119 @@ class LoginForm extends React.Component<Props> {
...
-        <form
-          ref={ref => (this.form = ref)}
-          method="post"
-          action={this.props.action}>
-          <input type="hidden" name="_csrf" value={csrfToken} />
-          <input type="hidden" name="to" value={to!} />
-          <CardContent>
-            <Text className={classes.title} variant="h5">
-              {this.props.title}
+      <Grid className={classes.login}>
```
thmsschmitt: I think this can be simplified to
```
 <div className={classes.login}>
        <div className={classes.title}>
         ...
        </div>
        <Card raised={true} className={classes.card}>
        ...
```

nms/app/components/FormField.tsx
```diff
@@ -55,6 +55,13 @@ const useStyles = makeStyles<Theme>(theme => ({
     fontWeight: 400,
     marginLeft: '8px',
   },
+  label: {
+    fontSize: '14px',
+    fontFamily: 'Inter',
```
thmsschmitt, line 60: What is that fontFamily about?

nms/app/views/login/LoginForm.tsx
```diff
@@ -16,30 +16,62 @@ import Card from '@material-ui/core/Card';
...
   },
   footer: {
-    padding: '18px 16px 16px',
-    float: 'right',
+    padding: '0 24px',
+  },
+  login: {
+    marginTop: '10%',
   },
   title: {
```
thmsschmitt, line 50: This should be sufficient
Screenshot 2022-07-26 at 10 13 38

thmsschmitt (changes requested):

nms/app/views/login/LoginForm.tsx
HannaFar: it was when I used the Grid component but now I need to keep most of it

nms/app/views/login/LoginForm.tsx
```diff
@@ -65,67 +102,117 @@ class LoginForm extends React.Component<Props> {
...
-              color="primary">
-              Login
-            </Button>
-          </CardActions>
-        </form>
-      </Card>
+      <div className={classes.login}>
+        {ssoEnabled ? (
+          <>
+            <div className={classes.title}>
```
thmsschmitt: Ins't the title/sub-title exactly the same as without SSO and can therefore be moved out of ternary?

Neudrino:
> There are several warnings [in the yarn test job](https://github.com/magma/magma/runs/7498663012?check_suite_focus=true#step:6:367) about certain elements not used in the correct places. E.g. `alignItems`, `xs`. A good search term for the logs is `must be used on`.

They still appear in the last job. :detective:

thmsschmitt:
> > There are several warnings [in the yarn test job](https://github.com/magma/magma/runs/7498663012?check_suite_focus=true#step:6:367) about certain elements not used in the correct places. E.g. `alignItems`, `xs`. A good search term for the logs is `must be used on`.
> 
> They still appear in the last job. 🕵️

I fixed some errors that are unrelated to login here https://github.com/magma/magma/pull/13430

nms/app/views/login/LoginForm.tsx
thmsschmitt: I see, this is because the title and subtitle are spans :|

nms/app/components/FormField.tsx
thmsschmitt: I looked a bit around and this is indeed a standard font that we use.
Although we typically use `fontFamily: '"Inter", sans-serif',`

This changes all labels across the app, but it does look better and it consistent with outer places so I'm fine with this change.

thmsschmitt (approved):

PR #13404, opened 2022-07-25
chore(dp): Add harness certification service

sklgromek:

## Summary
Deploying harness certification service instead of fake_sas when `DP_CERTIFICATION` env variable is set

## Test Plan
- [x] Deploy magma locally by invoking `DP_CERTIFICATION=true make orc8r`
- [x] Check if fake_sas contains the harness app, and is it possible to run it

## Additional Information


xbend (approved):
LGTM

PR #13410, opened 2022-07-25
refactor(nms): move FEG Gateway Add Button to equipment page

### LINKED ISSUE #13371: Migrate adding and editing FEGs gateways to new style
The design of the dialog to add and edit FEG gateways is outdated and does not match with the design that is used for LTE gateways.

FEG
Image

LTE
Image

There is an extra gateway page for FEG that also uses an outdated design and that has the same functionality as the equipment page.  This page does not exist for LTE networks. The only extra functionality that the FEG gateway page provides is the "Add Gateway" button.

Image

Requirements:
* The "Add Gateway" should be moved to the equipment page
* The equipment should become the second page in the sidebar (like it is for LTE)
* The gateway page should be removed
* The FEGGatewayDialog should be styled to look like the corresponding dialog for LTE

Neudrino:
## Summary

- Contributes to #13371 

## Test Plan

grafik


nms/app/components/feg/FEGGatewayDialog.tsx
```diff
@@ -179,6 +179,27 @@ function virtualApnRulesToObject(
     });
 }
 
+export function FEGAddEditGatewayButton({title}: {title: string}) {
```
thmsschmitt: I don't think there is any editing here.

nms/app/views/equipment/FEGEquipmentDashboard.tsx
```diff
@@ -52,6 +52,7 @@ function EquipmentDashboardInternal() {
             label: 'Federation Gateways',
             to: 'gateway',
             icon: CellWifiIcon,
+            filters: <FEGAddEditGatewayButton title="Add New" />,
```
thmsschmitt: If the button is only for adding gateways we could also hard code the title.

thmsschmitt (approved):
Looks good, I left some minor comments.

PR #13423, opened 2022-07-26
fix(agw): Enhanced debugging in CI for agw

VinashakAnkitAman:
## Title
fix(agw): Enhanced debugging in CI for agw

## Summary
During the CI run, many test cases are there which restart MME causing multiple MME log files generation in the Dev VM. However, after execution of lte integ test, we are fetching only the last MME log. This PR adds code to fetch all MME logs from Dev VM. Moreover the integ test execution is not updating the status properly after execution because of which artifact logs are not published when the CI runs are failing. This PR fixes the integ test status issue. The PR changes does the following:
- Fetches all the MME logs from the Dev VM after integ test execution
- Fetches the final integ test status and update the command execution output with failure for failing executions
- Shifts the ignoring of failing test case from overall integ test case execution to the command execution individually
- Removes the unnecessary sleep after test case execution

## Test plan
Verified with sanity and in CI
Successful CI execution log:
https://github.com/VinashakAnkitAman/magma/actions/runs/2740579763
https://github.com/VinashakAnkitAman/magma/actions/runs/2803524233


lte/gateway/python/integ_tests/Makefile
```diff
@@ -22,17 +22,18 @@ $(PYTHON_BUILD)/setupinteg_env:
...
 else
 	FLAKY_CMD_ARGS :=
 endif
+# Don't remove the ending comment from function execute_test. Somehow without
+# any ending statement the failing testcases are missing status update
 define execute_test
 	echo "Running test: $(1)"
 	timeout --foreground -k 930s 900s sudo -E PATH=$(PATH) PYTHONPATH=$(PYTHONPATH):$(S1AP_TESTER_PYTHON_PATH) $(BIN)/pytest $(FLAKY_CMD_ARGS) --capture=tee-sys --junit-xml=$(RESULTS_DIR)/$(basename $(notdir $(1))).xml -x $(1) || (echo "fail" > $(MAGMA_ROOT)/test_status.txt && exit 1)
-	sleep 1
+	# Testcase $(1) execution completed
```
pruthvihebbani, line 36: Is this comment needed here?

pruthvihebbani (changes requested):

lte/gateway/python/integ_tests/Makefile
VinashakAnkitAman: This comment is not mandatory, but when exiting for the failing the test cases in the previous statement, if there is no following statement after that, somehow the failure status is not updated to the terminal. This is the reason why there was a sleep added there earlier. I have replaced the sleep with the comment to reduce the execution time. Moreover I have mentioned this point in line 31 above for clarity when working with this file.

lte/gateway/fabfile.py
```diff
@@ -875,7 +876,7 @@ def _run_integ_tests(gateway_ip='192.168.60.142', tests=None, federated_mode=Fal
         ' sudo ethtool --offload eth1 rx off tx off; sudo ethtool --offload eth2 rx off tx off;'
         ' source ~/build/python/bin/activate;'
         ' export GATEWAY_IP=%s;'
-        ' make -i %s enable-flaky-retry=true %s\''
+        ' make %s enable-flaky-retry=true %s\''
```
ardzoht, line 879: Will not this be removing the continuous tests execution?

VinashakAnkitAman: Hi @ardzoht There were 2 ways to ignore failures and go for continuous execution. One was by passing this `-i` flag here in the make command itself and the 2nd one was by adding `-` before the command inside a makefile. This `-i` flag is working fine but creates problem in getting final failure status from the integ test run and for that reason artifact logs were also not published properly from VM. So, instead of this `-i`, I have added `-` inside the makefile for the commands when flaky retry is enabled. Please have a look on line numbers 60/77 in the Makefile changes in this PR where I have added `-` when we want to run integ test or fed integ test with flaky retry enabled. I have removed this `-i` flag from ci run but the `-i` flag is still valid after this change and if we want to run this make command manually with `-i`, that will work fine.

ardzoht (approved):
lgtm

pruthvihebbani (commented):
LGTM

pruthvihebbani (approved):

PR #13428, opened 2022-07-26
fix(nms): ignore errors from DP when it's disabled

ivan-freedomfi:
## Summary

Temporary hotfix for https://github.com/magma/domain-proxy/issues/469.

Ignore the error from Domain Proxy `CBSDContext` when Domain Proxy is disabled. Log error to console instead of showing a notification to the user.

Signed-off-by: Ivan Sergiienko <ivan@freedomfi.com>

### Docker compose
- Use [Quick Start Guide](https://magma.github.io/magma/docs/next/basics/introduction.html) to start `orc8r` and `nms` using `docker-compose`.
- Open NMS for some network.
- Now the error is logged to console, no notification is shown to the user:

dp-502

### Production deployment with helm charts
@sklgromek tested kubernetes.
Kubernetes deployment results into `404` response from Domain Proxy, rather than `502` that is seen with `docker-compose`.
We can emulate this behaviour with docker compose deployment by running the following commands after startup:
```
docker exec -it orc8r_controller_1 rm -rf /etc/magma/configs/dp
docker exec -it orc8r_controller_1 supervisorctl restart all
```

The console message for this case is a bit different:

nms-new-error

### CBSDs load successfully when DP is enabled
If I run Magma with Domain Proxy enabled, it works as expected:

image

## Additional Information

- This is just a quick fix to get v1.8 out.

### DIFF
--- a/nms/app/context/CbsdContext.tsx
+++ b/nms/app/context/CbsdContext.tsx
@@ -11,6 +11,7 @@
  * limitations under the License.
  */
 import * as React from 'react';
+import {AxiosError} from 'axios';
 import {useCallback, useEffect, useMemo, useState} from 'react';
 
 import MagmaAPI from '../api/MagmaAPI';
@@ -58,6 +59,33 @@ async function fetch(params: {
   } = params;
   if (networkId == null) return;
 
+  /**
+   * Temporary hotfix for https://github.com/magma/domain-proxy/issues/469.
+   * Ignore error when domain proxy module is disabled,
+   * so that an error is not shown on the main page.
+   */
+  const handleError = (error: unknown) => {
+    const status = (error as AxiosError)?.response?.status;
+
+    if (status === 404) {
+      console.error(
+        'CBSD endpoint not found. Is Domain Proxy module enabled in orc8r deployment?',
+      );
+      return;
+    }
+
+    if (status === 502) {
+      console.error(
+        'CBSD endpoint returns 502. Is Domain Proxy module up and running?',
+      );
+      return;
+    }
+
+    enqueueSnackbar?.('failed fetching CBSDs information', {
+      variant: 'error',
+    });
+  };
+
   try {
     setIsLoading(true);
 
@@ -69,10 +97,8 @@ async function fetch(params: {
       })
     ).data;
     setFetchResponse(response);
-  } catch {
-    enqueueSnackbar?.('failed fetching CBSDs information', {
-      variant: 'error',
-    });
+  } catch (error) {
+    handleError(error);
   } finally {
     setIsLoading(false);
   }


nms/app/context/CbsdContext.tsx
```diff
@@ -69,10 +62,13 @@ async function fetch(params: {
       })
     ).data;
     setFetchResponse(response);
-  } catch {
-    enqueueSnackbar?.('failed fetching CBSDs information', {
-      variant: 'error',
-    });
+  } catch (error) {
+    console.error(error);
+    // #469 Ignore error when domain proxy module is disabled,
```
thmsschmitt: Maybe link to https://github.com/magma/domain-proxy/issues/469 instead because it is not clear that this isn't an issue from the main repo.

thmsschmitt (dismissed):
This is very much a quick fix, but it does the job of unblocking the release. At some point I would love to see this resolved properly though.

nms/app/context/CbsdContext.tsx
ivan-freedomfi: Makes sense, corrected :)

thmsschmitt:
@ivan-freedomfi we should not close https://github.com/magma/domain-proxy/issues/469 with this PR. There needs to a proper fix at some point.

nms/app/context/CbsdContext.tsx
```diff
@@ -69,10 +70,21 @@ async function fetch(params: {
...
+    // https://github.com/magma/domain-proxy/issues/469.
+    // Ignore error when domain proxy module is disabled,
+    // so that an error is not shown on the main page.
+    const is404 = (error as AxiosError)?.response?.status === 404;
+    if (is404) {
+      console.error(
+        'CBSD endpoint not found. Is domain proxy module disabled in orc8r?',
+      );
+    } else {
+      enqueueSnackbar?.('failed fetching CBSDs information', {
```
thmsschmitt: The Error is back because I get a 502. I’m concerned that now everybody has to explicitly disable DP.  
Screenshot 2022-07-28 at 17 59 05

sklgromek: @thmsschmitt  DP in orc8r deployment by docker-compose is partial for now, and it's not deploying, all components which are required by DP.
It contains only DP module which is in orc8r, for that reason you are getting 502 error.
For building DP development environment we are using minikube and helm charts. Described here https://github.com/openEPC/magma/blob/master/dp/README.md
And in the helm charts, DP was disabled by default by this PR  https://github.com/magma/magma/pull/13445

thmsschmitt: `docker-compose` is the standard setup for developers, also for new people. And we should not give people the impression that something is broken if they follow the normal setup. Can we please go back to suppressing all errors.

sklgromek: Ok, it's not a big deal for now if also `502 ` for DP will be hidden. We will consider to prepare docker-compose also for the DP component. Sorry for the confusion

thmsschmitt (changes requested):
We should go back to initial version because it also covers the `docker-compose` setup.

thmsschmitt (approved):

nms/app/context/CbsdContext.tsx
ivan-freedomfi: I've updated the code to suppress both `404` and `502`.
So it works fine with `docker-compose` now.

PR #13431, opened 2022-07-27
feat(ci): Added initial build and upload workflow

### LINKED ISSUE #13449: Create CI to upload ONE deb artifact to JFrog Artifactory
## Acceptance criteria

The artifact type should be `deb` and hence should the repository type be and **debian package repository**.

The CI job should
- have a manual trigger
- should not have an automatic trigger
- allow to configure the target repository
  - the default should be set to a testing, i.e. `*-test` repository

## To be clarified

- building the artifact:
  - we might want to have this separately, as building different artifacts is very different, but upload is always the same
- add build-number to version, so that every upload is unique
- test the uploaded artifact before uploading

MoritzThomasHuebner:
Relating to #13450

Created in pairing with @alexzurbonsen 

## Summary
- Workflow builds nettle into a .deb file using fpm
- Workflow installs nettle with the .deb to test if the .deb works
- Extra check to see if the libnettle.so file exists
- Upload artifact to github
- Upload to artifactory
- Access is currently restricted to a certain set of users.

The actor can configure:
- The repository (default debian-test)
- The distribution (default focal-ci)
- Whether to do a try run, i.e. not actually upload the .deb (default true)

The magma repository probably still needs an access token for the artifactory. Currently, authentication works via username/password secrets. 

## Test Plan
Manually tested using the dry-run option. Also tested with a temporary folder on the artifactory (janheidbrink) with a temporary token:
https://artifactory.magmacore.org/ui/repos/tree/General/janheidbrink%2Fpool%2Ffocal-ci%2Foai-nettle_2.5-1ubuntu20.04_amd64.deb

## Additional Information

### DIFF
--- a/.github/workflows/build_magma_dep.yml
+++ b/.github/workflows/build_magma_dep.yml
@@ -1,0 +1,78 @@
+---
+name: "Build Nettle and Upload to Artifactory"
+on:
+  workflow_dispatch:
+    inputs:
+      repo:
+        description: 'Artifactory Repository'
+        required: false
+        default: 'debian-test'
+      distribution:
+        description: 'Distribution'
+        required: false
+        default: 'focal-ci'
+      dry_run:
+        description: 'Dry Run (only preview all files that would be uploaded)'
+        required: true
+        default: true
+
+env:
+  MAGMA_PACKAGE_DIR: /home/runner/magma-packages
+  LIBNETTLE_PATH: /usr/lib/libnettle.so
+  ARTIFACTORY_URL: https://artifactory.magmacore.org:443/artifactory/
+
+jobs:
+  build_deps:
+    runs-on: ubuntu-20.04
+    if: contains('["maxhbr", "nstng", "Neudrino", "tmdzk", "alexzurbonsen", "MoritzThomasHuebner"]', github.actor)
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # pin@v3
+      - name: Install fpm
+        shell: bash
+        run: |
+          sudo apt update
+          sudo apt install -y ruby ruby-dev rubygems build-essential
+          sudo gem install --no-document fpm
+      - name: Build nettle
+        shell: bash
+        run: |
+          mkdir -p ${MAGMA_PACKAGE_DIR}
+          cd third_party/build
+          ./build.py nettle --no-install
+          cp *.deb $MAGMA_PACKAGE_DIR
+      - name: Testing if .deb file can be installed
+        shell: bash
+        run: |
+          sudo apt install -y ${MAGMA_PACKAGE_DIR}/*.deb
+      - name: Testing is libnettle.so exists.
+        shell: bash
+        run: |
+          if test -f "$LIBNETTLE_PATH"; then
+              echo "$LIBNETTLE_PATH exists."
+          else
+              echo "$LIBNETTLE_PATH does not exist. Build failed. Exiting..."
+              exit 2
+          fi
+      - name: Upload zipped nettle.deb to Github
+        uses: actions/upload-artifact@v3
+        with:
+          name: nettle
+          path: ${{ env.MAGMA_PACKAGE_DIR }}
+      - name: Setup JFROG CLI
+        uses: jfrog/setup-jfrog-cli@v2
+        env:
+          JF_ENV_1: ${{ secrets.JFROG_TOKEN }}
+      - name: Set dry run environment variable
+        id: is_dry
+        if: ${{ github.event.inputs.dry_run == 'true' }}
+        run: |
+          echo "IS_DRY=--dry-run" >> $GITHUB_ENV
+      - name: Publish debian packages
+        run: |
+          jf rt u \
+            --recursive=false \
+            --url=${ARTIFACTORY_URL} \
+            --detailed-summary \
+            ${{ env.IS_DRY }} \
+            --target-props="deb.distribution=${{ github.event.inputs.distribution }};deb.component=main;deb.architecture=amd64" \
+            "~/magma-packages/(*).deb" ${{ github.event.inputs.repo }}/pool/${{ github.event.inputs.distribution }}/{1}.deb


.github/workflows/build_magma_dep.yml
```diff
@@ -0,0 +1,73 @@
...
+        default: 'focal-ci'
+      dry_run:
+        description: 'Dry Run (only preview all files that would be uploaded)'
+        required: true
+        default: true
+
+jobs:
+  build_deps:
+    runs-on: ubuntu-20.04
+    if: contains('["maxhbr","nstng", "Neudrino", "tmdzk", "alexzurbonsen", "MoritzThomasHuebner"]', github.actor)
```
Neudrino: _nitpick:_ inconsistent spacing here.

.github/workflows/build_magma_dep.yml
```diff
@@ -0,0 +1,73 @@
...
+    steps:
+      - uses: actions/checkout@7884fcad6b5d53d10323aee724dc68d8b9096a2e # pin@v2
```
Neudrino: As this is a new workflow, can we us the newest v3 immediately, please.

.github/workflows/build_magma_dep.yml
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          fetch-depth: 0
```
Neudrino: Why would that be necessary?
We should not be doing anything on foreign commits in this workflow.

.github/workflows/build_magma_dep.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Install fpm
+        shell: bash
+        run: |
+          sudo apt update
+          sudo apt install -y ruby ruby-dev rubygems build-essential
+          sudo gem install --no-document fpm
+      - name: Build nettle
+        shell: bash
+        run: |
+          mkdir ~/magma-packages
```
Neudrino: - In scripts always use mkdir with the `-p` option
- Is it safe to use the home dir on a Github runner?

.github/workflows/build_magma_dep.yml
```diff
@@ -0,0 +1,73 @@
...
+          cd third_party/build && ./build.py nettle --no-install && cp *.deb ~/magma-packages
+      - name: Testing if .deb file can be installed
+        shell: bash
+        run: |
+          sudo apt install -y ~/magma-packages/*.deb
```
Neudrino: This is the third time `~/magma-packages` is used. We should really use an `env` for it than.

.github/workflows/build_magma_dep.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Testing is libnettle.so exists.
+        shell: bash
+        run: |
+          FILE=/usr/lib/libnettle.so
```
Neudrino: Consider using a [Github env](https://docs.github.com/en/actions/learn-github-actions/environment-variables) instead.

.github/workflows/build_magma_dep.yml
```diff
@@ -0,0 +1,73 @@
...
+          name: nettle
+          path: ~/magma-packages/
+      - name: Setup JFROG CLI
+        uses: jfrog/setup-jfrog-cli@v2
+        env:
+          JF_ENV_1: ${{ secrets.JFROG_TOKEN }}
+      - name: Dry run - publish debian packages
+        if: ${{ github.event.inputs.dry_run == 'true'}}
+        run: |
+          jf rt u --recursive=false --url=https://artifactory.magmacore.org:443/artifactory/ --detailed-summary \
```
Neudrino: - Every command option should have its own line.
- We might be able to get rid of the URL by [including it into the token](https://github.com/marketplace/actions/setup-jfrog-cli#creating-the-configuration-on-your-local-machine)?

.github/workflows/build_magma_dep.yml
```diff
@@ -0,0 +1,73 @@
...
+            --dry-run \
+            --target-props="deb.distribution=${{ github.event.inputs.distribution }};deb.component=main;deb.architecture=amd64" \
+            "~/magma-packages/(*).deb" ${{ github.event.inputs.repo }}/pool/${{ github.event.inputs.distribution }}/{1}.deb
+      - name: Publish debian packages
+        if: ${{ github.event.inputs.dry_run == 'false' }}
+        run: |
+          jf rt u --recursive=false --url=https://artifactory.magmacore.org:443/artifactory/ --detailed-summary \
+            --target-props="deb.distribution=${{ github.event.inputs.distribution }};deb.component=main;deb.architecture=amd64" \
+            "~/magma-packages/(*).deb" ${{ github.event.inputs.repo }}/pool/${{ github.event.inputs.distribution }}/{1}.deb
```
Neudrino: One should probably avoid this superfluous duplication if the only difference is the `--dry-run` option.

A solution could involve [environment variables](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable).
Something similar to the following:
```yaml
steps:
  - name: Set the value
    id: is_dry
    if: ${{ github.event.inputs.dry_run == 'true' }}
    run: |
      echo "IS_DRY='--dry-run'" >> $GITHUB_ENV
  - name: Use the value
    id: publish
    run: |
      jf rt u --all_other_options_here \ 
        "${{ env.IS_DRY }}"
```

.github/workflows/build_magma_dep.yml
MoritzThomasHuebner: Fixed

.github/workflows/build_magma_dep.yml
MoritzThomasHuebner: :+1:

.github/workflows/build_magma_dep.yml
MoritzThomasHuebner: I removed this

.github/workflows/build_magma_dep.yml
MoritzThomasHuebner: I created the env

.github/workflows/build_magma_dep.yml
MoritzThomasHuebner: I created an env

.github/workflows/build_magma_dep.yml
MoritzThomasHuebner: - I added the `-p` on the `mkdir` command.
- By using the home directory do you mean the use of `~` or just being in `/home/runner/` in general?

.github/workflows/build_magma_dep.yml
MoritzThomasHuebner: - I reformatted so that every command has its own line
- It seems certainly possible, however this would be hard for me to test since I don't have access to the magma github secrets. Other workflows also state the artifactory URL explicitly.

.github/workflows/build_magma_dep.yml
MoritzThomasHuebner: I made the change as you suggested but it took me while to figure out what was going wrong. I had to remove some of the quotation marks to make it work.

.github/workflows/build_magma_dep.yml
Neudrino: Sorry, should have made it clear though, that we should still pin the version to an exact hash. :innocent:

.github/workflows/build_magma_dep.yml
Neudrino: `home dir`: The latter. Using `~` should be fine. However, I was not sure, if and how home directory would be usable on those runners. But if it works, it is probably fine.

.github/workflows/build_magma_dep.yml
MoritzThomasHuebner: All good, pinning versions is not consistent across all workflows, so I was unsure.

Neudrino: Since https://github.com/magma/magma/pull/12990 was merged, all Github actions should be pinned to hashes.

Neudrino: But I am confused now, the comment says it is version 2 again.
We should have it pinned to the hash of the latest version 3 though.

MoritzThomasHuebner: Sorry, that was my bad. The comment should now say it's version 3 and the git commit hash should be 2541b1294d2704b0964813337f33b291d3f8596b

Neudrino (approved):

PR #13446, opened 2022-07-28
chore: Remove Go imports for initflag

LKreutzer:

## Summary

- The Go imports of the initflag package are removed
- The package is still used by openwrt
  - This usage should be cleaned up at a later point
- The initflag Bazel libraries are removed 
 
## Test Plan

- CI
- Run `cd ${MAGMA_ROOT}/orc8r/gateway/go &&  go test ./...`

## Additional Information


orc8r/lib/go/initflag/syslog.go
```diff
@@ -1,161 +0,0 @@
...
-	"flag"
-	"fmt"
-	"log"
-	"log/syslog"
-	"os"
-	"path/filepath"
-	"strings"
-)
-
-const syslogFlag = "stderr2syslog"
```
crasu: This seems to be used in the magma openwrt deployment here:
./openwrt/gateway/configs/etc/init.d/magma.template

crasu (changes requested):

orc8r/lib/go/initflag/syslog.go
crasu: I do not know if we can remove the openwrt stuff as well.

nstng:
converted to draft - initflag still needed in other places. needs to be adapted

LKreutzer:
Changed the commit to only remove the side-effect imports in the Go code but the package itself since it is still used in the openwrt code.

LKreutzer:
Added commit to also remove the imports in the bazel code.

LKreutzer:
As discussed with @crasu @sebathomas @jheidbrink @alexzurbonsen:

- Add initflag functionality to bazel in https://github.com/magma/magma/pull/13532
- Closing this PR and move refactoring of initflag functionality to future task.

PR #13447, opened 2022-07-28
fix(agw): Ported c files to cpp for sgw_s8 module

rsarwad:
## Summary
Moved the c files to cpp from sgw_s8 module
## Test Plan


lte/gateway/c/core/oai/tasks/sgw_s8/sgw_s8_handlers.cpp
```diff
@@ -1476,7 +1486,7 @@ void sgw_s8_handle_s11_create_bearer_response(
     itti_s11_nw_init_actv_bearer_rsp_t* s11_actv_bearer_rsp, imsi64_t imsi64) {
   OAILOG_FUNC_IN(LOG_SGW_S8);
   uint32_t msg_bearer_index = 0;
-  sgw_eps_bearer_ctxt_t dedicated_bearer_ctxt = {0};
```
LKreutzer, line 1479: Why are these changes necessary? It seems that it would be safer to keep the initialization?

lte/gateway/c/core/oai/tasks/sgw_s8/sgw_s8_handlers.cpp
```diff
@@ -1655,12 +1665,12 @@ int sgw_s8_handle_delete_bearer_request(
 }
 
 // Handle NW-initiated dedicated bearer dectivation response from MME
-status_code_e sgw_s8_handle_s11_delete_bearer_response(
+int sgw_s8_handle_s11_delete_bearer_response(
     sgw_state_t* sgw_state,
     const itti_s11_nw_init_deactv_bearer_rsp_t* const
         s11_delete_bearer_response_p,
     imsi64_t imsi64) {
-  uint32_t rc = RETURNok;
```
LKreutzer, line 1663: If these type changes are not necessary we should prefer to specify the size explicitly, as `int` may vary depending on the compiler and architecture. Are these changes needed?

lte/gateway/c/core/oai/tasks/sgw_s8/sgw_s8_handlers.cpp
```diff
@@ -1565,7 +1575,7 @@ void sgw_s8_send_failed_create_bearer_response(
     sgw_state_t* sgw_state, uint32_t sequence_number, char* pgw_cp_address,
     gtpv2c_cause_value_t cause_value, Imsi_t imsi, teid_t pgw_s8_teid) {
   OAILOG_FUNC_IN(LOG_SGW_S8);
-  itti_s11_nw_init_actv_bearer_rsp_t s11_actv_bearer_rsp = {0};
```
LKreutzer, line 1568: See above.

LKreutzer (changes requested):

LKreutzer (commented):
Running all MME unit tests with `asan` and `lsan` for 10 runs with

- `bazel test lte/gateway/c/core/... --config=lsan --runs_per_test=10`
- `bazel test lte/gateway/c/core/... --config=asan --runs_per_test=10`

is passing (on the codespaces runner in the devcontainer).

lte/gateway/c/core/oai/tasks/sgw_s8/sgw_s8_handlers.cpp
rsarwad: > Why are these changes necessary? It seems that it would be safer to keep the initialization?

Hi @LKreutzer, Thanks for reviewing the PR
The structure, sgw_eps_bearer_ctxt_t contains nested structure, if these structures contains enumeration structures, the compiler reports an error saying that "invalid conversion from ‘int’ to enum type"
62 typedef struct sgw_eps_bearer_ctxt_s {
 63   paa_t paa;
 64   ebi_t eps_bearer_id;  ///< An EPS bearer identity uniquely identifies an EPS
 65                         ///< bearer for one UE accessing via E-UTRAN
 66   traffic_flow_template_t tft;  ///< Traffic Flow Template
 67
 68   ip_address_t
 69       p_gw_address_in_use_up;
 <contains few more fields>
 }

To fix the compiler, we initialized with empty braces and verified with unit test cases that it gets initialized with zero values
image

lte/gateway/c/core/oai/tasks/sgw_s8/sgw_s8_handlers.cpp
rsarwad: Thanks for bringing into notice.
Actually I made mistake to return an int. It should be status_code_e because in unit test case we validate the procedure based on return value.
https://github.com/magma/magma/blob/master/lte/gateway/c/core/oai/test/sgw_s8_task/sgw_dedicated_bearer.cpp#L199

LKreutzer (approved):
Lgtm, thanks for the cleanup!

lte/gateway/c/core/oai/test/sgw_s8_task/sgw_s8_utility.hpp
```diff
@@ -13,24 +13,26 @@
...
-#include "lte/gateway/c/core/oai/include/sgw_context_manager.h"
-#include "lte/gateway/c/core/oai/include/spgw_types.h"
 #include "lte/gateway/c/core/oai/tasks/gtpv1-u/gtpv1u.h"
-#include "lte/gateway/c/core/oai/tasks/sgw/sgw_handlers.hpp"
-#include "lte/gateway/c/core/oai/tasks/sgw_s8/sgw_s8_defs.h"
-#include "lte/gateway/c/core/oai/tasks/sgw_s8/sgw_s8_s11_handlers.h"
 }
 
+#include "lte/gateway/c/core/oai/include/sgw_context_manager.hpp"
+#include "lte/gateway/c/core/oai/include/spgw_types.hpp"
```
ajahl: Is there a reason why you put these includes separately?

rsarwad: Thank you! corrected the order of header in other files as well

ajahl (approved):

ssanadhya (approved):
Lgtm

Issue #13418, opened 2022-07-26
[Regression]Service Request is getting rejected as AMF is storing 'ffffffff' as TMSI in case of Registration-accept retry cases.


NeelimaKalpana:
### Your Environment

Version :- 1.8.0-1658821072-44f94ca1
Affected Component :- AGW
Affected Subcomponent :- AGW
Deployment Environment :- Baremetal - VM based

#### To Reproduce

1. Trigger the registration from UE and de-reg by storing the TMSI value.
2. Try to go for GUTI registration and wait for reg-accept message and wait till maximum retries are happening and see implicit reg-complete is happening.
3. Now try to establish PDU-Session.
4. Once PDU is up try to make UE to idle .
5. Once UE is idle , trigger service request procedure from UE.
6. See that service accept is getting triggered. But AMF is sending service reject as IMSI stored is not proper.

#### Expected behavior
AMF should send service accept.

#### Screenshots
![image.png](https://images.zenhubusercontent.com/6139f1931bcb9da2af1bf651/e348af1d-8c71-4319-84e4-81287bd60080)

#### Additional context
Please find the attached logs for more info.
[https://app.zenhub.com/files/170803235/82e368c0-95e5-4022-bafa-cba4a2a356d2/download](https://app.zenhub.com/files/170803235/82e368c0-95e5-4022-bafa-cba4a2a356d2/download)

NeelimaKalpana:
Closing this bug as issue is not observed in latest master.

Validated Build: 1.8.0-1659442984-a5290e7d

Logs attached for the same.
[https://app.zenhub.com/files/170803235/a4eb699f-2eac-44b5-b4d5-330a454c07fb/download](https://app.zenhub.com/files/170803235/a4eb699f-2eac-44b5-b4d5-330a454c07fb/download)


[https://app.zenhub.com/files/170803235/15880e53-d50f-4af7-b590-2c854e508ae9/download](https://app.zenhub.com/files/170803235/15880e53-d50f-4af7-b590-2c854e508ae9/download)

PR #13454, opened 2022-07-28
fix(amf): Reverting Guti creation during reg-accept

Sathyaj27:
Signed-off-by: Sathyaj27 <sathya.jayadev@wavelabs.ai>

fixing #13418 

## Summary

Reverting the Guti creation during registration accept due to CI failure.

## Test Plan
- Tested on abot
- 
cicd

## Additional Information

- [ ] This change is not backwards-breaking


panyogesh (approved):
LGTM

PR #13456, opened 2022-07-28
feat(nms): Integrate FEG gateway upgrade UI into equipment page

### LINKED ISSUE #13372: Integrate FEG gateway upgrade UI into equipment page
For FEG networks there is a configure page for gateway upgrades.

Image

For LTE networks this page does not exist and the corresponding UI lives on the equipment page.

Image

Image

Image

Requirements
* FEG should use the same UI as LTE
* If sensible existing components should be reused
* The FEG configure page should be removed

HannaFar:

## Summary
Add gateway upgrade button and gateways upgrade view to feg gateway
fix minor bug in gateway equipment upgrade

 
Capture d’écran 2022-07-28 à 17 47 50

Closes #13372

## Additional Information

### DIFF
--- a/nms/app/context/FEGGatewayContext.tsx
+++ b/nms/app/context/FEGGatewayContext.tsx
@@ -25,11 +25,8 @@ import {useEffect, useState} from 'react';
 
 export type FEGGatewayContextType = {
   state: Record<string, FederationGateway>;
-  setState: (
-    key: GatewayId,
-    val?: MutableFederationGateway,
-    newState?: Record<string, FederationGateway>,
-  ) => Promise<void>;
+  setState: (key: GatewayId, val?: MutableFederationGateway) => Promise<void>;
+  updateGateway: (props: UpdateFegGatewayParams) => Promise<void>;
   refetch: (id?: GatewayId) => void;
   health: Record<GatewayId, FederationGatewayHealthStatus>;
   activeFegGatewayId: GatewayId;
@@ -84,7 +81,6 @@ async function initGatewayState(params: {
  * @property {(gatewayId:gateway_id) => void} setActiveFegGatewayId Sets the active gateway id.
  * @property {gateway_id} key Id of the gateway to be added, deleted or edited.
  * @property {mutable_federation_gateway} value New Value for the gateway with the id: key.
- * @property {[gateway_id]: federation_gateway} newState New State of the Federation Gateway.
  * @property {(msg, cfg,) => ?(string | number),} enqueueSnackbar Snackbar to display error
  */
 type GatewayStateParams = {
@@ -157,6 +153,22 @@ async function setGatewayState(params: GatewayStateParams) {
   setActiveFegGatewayId(
     await getActiveFegGatewayId(networkId, fegGateways, enqueueSnackbar),
   );
+}
+
+export type UpdateFegGatewayParams = {
+  gatewayId: GatewayId;
+  tierId: string;
+};
+
+async function updateGateway(
+  params: {networkId: NetworkId} & UpdateFegGatewayParams,
+) {
+  const {networkId, gatewayId, tierId} = params;
+  await MagmaAPI.lteGateways.lteNetworkIdGatewaysGatewayIdTierPut({
+    networkId,
+    gatewayId: gatewayId,
+    tierId: JSON.stringify(`"${tierId}"`),
+  });
 }
 
 /**
@@ -338,6 +350,11 @@ export function FEGGatewayContextProvider(props: {
             enqueueSnackbar,
           });
         },
+        updateGateway: props =>
+          updateGateway({
+            networkId,
+            ...props,
+          }),
         refetch: (id?: GatewayId) => {
           if (id) {
             void fetchFegGateway({id, networkId, enqueueSnackbar}).then(
--- a/nms/app/views/equipment/FEGGatewayTable.tsx
+++ b/nms/app/views/equipment/FEGGatewayTable.tsx
@@ -21,13 +21,21 @@ import CellWifiIcon from '@material-ui/icons/CellWifi';
 import CheckIcon from '@material-ui/icons/Check';
 import DeviceStatusCircle from '../../theme/design-system/DeviceStatusCircle';
 import FEGGatewayContext from '../../context/FEGGatewayContext';
+import GatewayTierContext from '../../context/GatewayTierContext';
 import Grid from '@material-ui/core/Grid';
 import Link from '@material-ui/core/Link';
+import MenuItem from '@material-ui/core/MenuItem';
+import OutlinedInput from '@material-ui/core/OutlinedInput';
 import React, {useContext, useState} from 'react';
+import Select from '@material-ui/core/Select';
+import Text from '../../theme/design-system/Text';
+
 import withAlert from '../../components/Alert/withAlert';
 import {GatewayId} from '../../../shared/types/network';
 import {GatewayTypeEnum, HEALTHY_STATUS} from '../../components/GatewayUtils';
 import {REFRESH_INTERVAL} from '../../context/AppContext';
+import {SelectEditComponent} from '../../components/ActionTable';
+import {getErrorMessage} from '../../util/ErrorUtils';
 import {useEnqueueSnackbar} from '../../hooks/useSnackbar';
 import {useInterval} from '../../hooks';
 import {useNavigate} from 'react-router-dom';
@@ -45,6 +53,10 @@ type EquipmentFegGatewayRowType = {
   csfb: string;
 };
 
+const ViewTypes = {
+  STATUS: 'Status',
+  UPGRADE: 'Upgrade',
+};
 /**
  * Displays the number of federation gateways alonside with a table showing
  * each federation gateway.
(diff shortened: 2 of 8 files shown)


nms/app/context/FEGGatewayContext.tsx
```diff
@@ -143,16 +145,24 @@ async function setGatewayState(params: GatewayStateParams) {
...
-      networkId: networkId,
-      gatewayId: key,
-    });
-    const newFegGateways = {...fegGateways};
-    const newFegGatewaysHealthStatus = {...fegGatewaysHealthStatus};
-    delete newFegGateways[key];
-    delete newFegGatewaysHealthStatus[key];
-    setFegGateways(newFegGateways);
-    setFegGatewaysHealthStatus(newFegGatewaysHealthStatus);
+    if (tierId !== undefined && tierId !== '') {
```
Neudrino: Would it be possible to use JS truthy here? Would that be shorter?
```js
if (!tierId) {
```

nms/app/views/equipment/FEGGatewayTable.tsx
```diff
@@ -45,13 +50,23 @@ type EquipmentFegGatewayRowType = {
   csfb: string;
 };
 
+const UPGRADE_VIEW = 'UPGRADE';
+
+const ViewTypes = {
+  STATUS: 'Status',
+  UPGRADE: 'Upgrade',
+};
```
Neudrino: Do these two need to stay in sync (manually)?
It seems as if the information of line 53 is already included in line 57?

nms/app/views/equipment/FEGGatewayTable.tsx
```diff
@@ -65,16 +80,36 @@ export default function GatewayTable() {
...
-              <AutorefreshCheckbox
-                autorefreshEnabled={refresh}
-                onToggle={() => setRefresh(current => !current)}
+              <TypedSelect
+                input={<OutlinedInput />}
+                value={currentView}
+                items={{
+                  STATUS: 'Status',
+                  UPGRADE: 'Upgrade',
+                }}
```
Neudrino: This seems to be a duplication of `viewTypes`?

nms/app/views/equipment/FEGGatewayTable.tsx
```diff
@@ -65,16 +80,36 @@ export default function GatewayTable() {
...
+                onChange={setCurrentView}
               />
             </Grid>
           </Grid>
         )}
       />
-      <StatusTable refresh={refresh} />
+      {currentView === 'UPGRADE' ? (
```
Neudrino: Compare to [line 83](https://github.com/magma/magma/pull/13456/files#diff-d8679aafc0dd22fd8dffd050f2e322b40dc01b2349acc81453b93b606de243eaR83). This seems to be inconsistent.

nms/app/views/equipment/FEGGatewayTable.tsx
```diff
@@ -209,3 +244,103 @@ function GatewayStatusTable(props: WithAlert & {refresh: boolean}) {
...
+type EquipmentGatewayUpgradeType = {
+  name: string;
+  id: GatewayId;
+  hardwareId: string;
+  tier: string;
+  currentVersion: string;
+};
+function UpgradeTable() {
+  const ctx = useContext(GatewayTierContext);
+  const gwCtx = useContext(FEGGatewayContext);
```
Neudrino: _optional:_ I would prefer a clearer naming of the elements, such that the distinction is clear on every usage.

nms/app/context/FEGGatewayContext.tsx
```diff
@@ -28,7 +28,7 @@ export type FEGGatewayContextType = {
   setState: (
     key: GatewayId,
     val?: MutableFederationGateway,
-    newState?: Record<string, FederationGateway>,
```
thmsschmitt, line 31: I would much rather have a separate `updateTier` method. Or maybe even and `updateGateway` method to keep it consistent with LTE.

nms/app/context/FEGGatewayContext.tsx
```diff
@@ -143,16 +145,24 @@ async function setGatewayState(params: GatewayStateParams) {
...
-    setFegGateways(newFegGateways);
-    setFegGatewaysHealthStatus(newFegGatewaysHealthStatus);
+    if (tierId !== undefined && tierId !== '') {
+      await MagmaAPI.lteGateways.lteNetworkIdGatewaysGatewayIdTierPut({
+        networkId,
+        gatewayId: key,
+        tierId: JSON.stringify(`"${tierId}"`),
+      });
+    } else {
+      await MagmaAPI.federationGateways.fegNetworkIdGatewaysGatewayIdDelete({
```
thmsschmitt: So if I pass in an empty string as `tierId` the gateway gets deleted 🧐
These multipurpose methods as just asking for things to go wrong. See comment above.

nms/app/views/equipment/FEGGatewayTable.tsx
```diff
@@ -209,3 +244,103 @@ function GatewayStatusTable(props: WithAlert & {refresh: boolean}) {
...
 const StatusTable = withAlert(GatewayStatusTable);
+
+type EquipmentGatewayUpgradeType = {
+  name: string;
+  id: GatewayId;
+  hardwareId: string;
+  tier: string;
+  currentVersion: string;
+};
+function UpgradeTable() {
```
thmsschmitt, line 257: This is in largely a copy of the `EquimentGateway.UpgradeTable`. That is a bit sad, but I understand that it might be    somewhat awkward to abstract this. Have you tried?

thmsschmitt (changes requested):
We should at least split the `setState` and clean up the `ViewTypes`

nms/app/views/equipment/FEGGatewayTable.tsx
HannaFar: I agree

nms/app/context/FEGGatewayContext.tsx
```diff
@@ -338,6 +349,11 @@ export function FEGGatewayContextProvider(props: {
             enqueueSnackbar,
           });
         },
+        updateGateway: props =>
+          updateGateway({
+            networkId,
+            ...props,
+          } as UpdateFegGatewayParams),
```
thmsschmitt: The cast here hides potential problems. Like a missing gatewayId.
We should type the method on the context correctly. See comment above.

nms/app/context/FEGGatewayContext.tsx
```diff
@@ -25,11 +25,8 @@ import {useEffect, useState} from 'react';
 
 export type FEGGatewayContextType = {
   state: Record<string, FederationGateway>;
-  setState: (
-    key: GatewayId,
-    val?: MutableFederationGateway,
-    newState?: Record<string, FederationGateway>,
-  ) => Promise<void>;
+  setState: (key: GatewayId, val?: MutableFederationGateway) => Promise<void>;
+  updateGateway: (props: Partial<UpdateFegGatewayParams>) => Promise<void>;
```
thmsschmitt: We should type the props correct to protect us against misuse.

Maybe something like 
```
type UpdateFegGatewayParams = {
   gatewayId: GatewayId;
   tierId?: string;
}
```

nms/app/context/FEGGatewayContext.tsx
```diff
@@ -159,6 +155,21 @@ async function setGatewayState(params: GatewayStateParams) {
   );
 }
 
+export type UpdateFegGatewayParams = {
+  gatewayId: GatewayId;
+  networkId: NetworkId;
+  tierId: string;
+};
+
+async function updateGateway(params: UpdateFegGatewayParams) {
```
thmsschmitt: Given the comment above this could be `params: { networkId: NetworkId } & UpdateFegGatewayParams`

nms/app/context/FEGGatewayContext.tsx
```diff
@@ -159,6 +155,21 @@ async function setGatewayState(params: GatewayStateParams) {
...
+  const {networkId, gatewayId, tierId} = params;
+  await MagmaAPI.lteGateways.lteNetworkIdGatewaysGatewayIdTierPut({
```
thmsschmitt, line 167: We should also handle the case that the `tierId` is undefined

nms/app/views/equipment/FEGGatewayTable.tsx
thmsschmitt: The crux here is that `TypedSelect` is working on the keys. Which make the `ViewTypes` useless because we can't use `ViewTypes.STATUS` and `ViewTypes.UPGRADE` as you would for an enum. Instead we have to compare against `'UPGRADE'` and `'STATUS'` string literals and duplicate everything.

One could construct something like
```
const UPGRADE_VIEW = 'UPGRADE';
const STATUS_VIEW = 'STATUS';

const ViewTypes = {
  [STATUS_VIEW]: 'Status',
  [UPGRADE_VIEW]: 'Upgrade',
};
```
but that also feels overly complicated

nms/app/views/equipment/FEGGatewayTable.tsx
thmsschmitt: If you agree why did you resolve this without changing the names 🤔

If you want to keep short names `tierCtx` and `gwCtx` would already be clearer.

thmsschmitt (changes requested):
I left some comments that should be addressed

thmsschmitt (approved):

PR #13465, opened 2022-07-29
chore(dp): do not create idle grants

hejjestemjarek:

## Summary
There is no such state for grant as "idle". If grant is idle in DP it means it does not exist in the SAS, so it should also not exist in DP.
* If grant is about to be created, but it would be in the idle state - do not create such grant
* If grant is about to be updated into the idle state - delete such grant
* Remove all idle grant state references from code
* Remove all idle grants from DB upon migration
* Remove idle grant state from DB upon migration

## Test Plan
* Modified Configuration Controller test cases: when they would check for an idle grant - they now check if grant does not exist

## Additional Information


dp/cloud/python/magma/radio_controller/services/active_mode_controller/service.py
```diff
@@ -157,21 +150,11 @@ def _list_cbsds(session: Session) -> State:
     # It might be possible to use join instead of nested queries
     # however it requires some serious investigation on how to use it
     # with eager_contains and filter (aliases)
```
jkmar: this comment should also be removed

dp/cloud/python/magma/db_service/migrations/versions/98f7ccfbd2f8_.py
```diff
@@ -0,0 +1,31 @@
+"""empty message
```
jkmar: It would be nice to change this filename (not content, only filename) to match the convention started by the previous migration (016_...)

dp/cloud/python/magma/db_service/migrations/versions/98f7ccfbd2f8_.py
```diff
@@ -0,0 +1,31 @@
...
+    Run upgrade
+    """
+    op.execute("DELETE FROM grants WHERE state_id = (SELECT id FROM grant_states WHERE name = 'idle');")
+    op.execute("DELETE FROM grant_states WHERE name = 'idle';")
+
+
+def downgrade():
+    """
+    Run downgrade
+    """
```
jkmar: We should add the idle grant in downgrade

dp/cloud/python/magma/db_service/migrations/versions/98f7ccfbd2f8_.py
```diff
@@ -0,0 +1,31 @@
...
+
+Revision ID: 98f7ccfbd2f8
+Revises: 58a1b16ef73c
+Create Date: 2022-07-26 14:46:13.599502
+
+"""
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '98f7ccfbd2f8'
```
jkmar: We should add migration unit test for this migration

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
```diff
@@ -202,30 +209,33 @@ def process_heartbeat_response(obj: ResponseDBProcessor, response: DBResponse, s
...
         None
     """
+    grant = _get_grant_from_response(response=response, session=session)
 
-    grant = _get_or_create_grant_from_response(obj, response, session)
-    if not grant:
+    if response.response_code == ResponseCodes.TERMINATED_GRANT.value:
+        if grant:
+            unset_frequency(grant=grant)
+            session.delete(instance=grant)
```
jkmar: Why sometimes there is `session.delete(instance=grant)` and the other times it is `session.delete(grant)` (I prefer the second one)

dp/cloud/python/magma/db_service/migrations/versions/98f7ccfbd2f8_.py
hejjestemjarek: Added.

dp/cloud/python/magma/db_service/migrations/versions/98f7ccfbd2f8_.py
hejjestemjarek: Added unit tests.

dp/cloud/python/magma/db_service/migrations/versions/98f7ccfbd2f8_.py
hejjestemjarek: Filename changed.

dp/cloud/python/magma/radio_controller/services/active_mode_controller/service.py
hejjestemjarek: Removed.

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
hejjestemjarek: I'm fan of Python's keyword args due to readability, but you're right I should've matched preexisting code convention for the given repo.   
Switched to positional args.

jkmar (approved):

PR #13388, opened 2022-07-22
chore(agw): Changes for new magma-dev base image

sebathomas:
## Summary

Creating a new base image of `magma-dev` on Ubuntu focal.

- Make many steps preburn again that changed in the last year
- openvswitch version `*.2.15.4-9-magma` is installed
- Several Ansible hints are addressed.
- Packing of VM is improved in the same way as in https://github.com/magma/magma/pull/13427

Closes #13179.

## Test Plan

- [x] [AGW build](https://github.com/crasu/magma/runs/7628256379?check_suite_focus=true)
- [x] Unit tests
- [x] Sudo tests
- [x] [LTE integration tests](https://github.com/sebathomas/magma/runs/7762509725) 
- [x] [Federated integration tests](https://github.com/sebathomas/magma/actions/runs/2780364051)

## Additional Information

- [ ] This change is backwards-breaking

PR #13471, opened 2022-08-02
chore(agw): Changes for new magma-dev base image

sebathomas:
## Summary

Includes all changes from #13388 but doesn't update the image tag yet.

## Test Plan

See #13388.

## Additional Information


sebathomas:
Doesn't work because the preburn changes will break the old magma-dev VM.

PR #13472, opened 2022-08-02
fix(amf): Adding Reject if Static ip configured is in subset of ip_pool

Sathyaj27:
Signed-off-by: Sathyaj27 <sathya.jayadev@wavelabs.ai>

## Summary
When static ip is configured in subset of ip_pool reject should be handled instead of no response.

## Test Plan
-Tested on UEransim
Rajeshsnew

4g Cases working fine. Ran s1ap integration test.
4G

## Additional Information

- [ ] This change is not backwards-breaking

### DIFF
--- a/lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
+++ b/lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
@@ -928,6 +928,16 @@ status_code_e amf_smf_handle_ip_address_response(
                  reinterpret_cast<char*>(response_p->pdu_session_id));
     return rc;
   }
+  if (response_p->result != SGI_STATUS_OK) {
+    rc = amf_pdu_session_establishment_reject(
+        ue_context->amf_ue_ngap_id, response_p->pdu_session_id, response_p->pti,
+        AMF_CAUSE_PROTOCOL_ERROR);
+    ue_context->amf_context.smf_ctxt_map.erase(response_p->pdu_session_id);
+    OAILOG_ERROR(LOG_AMF_APP,
+                 "Ip address allocation failed. Rejecting with cause %d",
+                 AMF_CAUSE_PROTOCOL_ERROR);
+    OAILOG_FUNC_RETURN(LOG_AMF_APP, rc);
+  }
 
   rc = amf_update_smf_context_pdu_ip(smf_ctx, &(response_p->paa));
 


lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
```diff
@@ -991,6 +991,11 @@ status_code_e amf_smf_handle_ip_address_response(
         OAILOG_ERROR(LOG_AMF_APP, "Create IPV4V6 Session \n");
       }
     }
+  } else {
+    amf_ue_ngap_id_t ue_id = ue_context->amf_ue_ngap_id;
+    int amf_cause = AMF_CAUSE_PROTOCOL_ERROR;
+    rc = amf_pdu_session_establishment_reject(ue_id, response_p->pdu_session_id,
```
rsarwad: This can coded as amf_pdu_session_establishment_reject(e_context->amf_ue_ngap_id, response_p->pdu_session_id, response_p->pti, AMF_CAUSE_PROTOCOL_ERROR);
So that we don't need 2 local variables

lte/gateway/python/magma/mobilityd/ip_allocator_static.py
```diff
@@ -138,6 +138,7 @@ def _allocate_static_ip(self, sid: str) -> Optional[IPDesc]:
                     ip_addr_info.ip,
                     ip_pool,
                 )
+                return IPDesc()
```
sebathomas: The function above will raise an exception so I don't think the new `return` is ever reached.

Sathyaj27: Addressed this.

lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
Sathyaj27: Addressed.

lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
```diff
@@ -991,6 +991,10 @@ status_code_e amf_smf_handle_ip_address_response(
         OAILOG_ERROR(LOG_AMF_APP, "Create IPV4V6 Session \n");
       }
     }
+  } else {
```
rsarwad: On failing to allocate IP or on reception of gRPC status as "not ok"
1. Then amf/smf should not update the context with received IP address as it is done in line, https://github.com/magma/magma/blob/69d17a7a6640fa1161dc7cc3c38e47d0d975f44f/lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp#L932
2. This condition shall be present at start of function.
3. At UE, on reception of pdu_session_establishment_reject for initial request type, then UE shall not create pdu session. So I think amf/smf should clear the session contexts. Correct me if I am wrong

Sathyaj27: Addressed the comments.

lte/gateway/c/core/oai/tasks/amf/amf_smf_send.cpp
```diff
@@ -928,6 +928,13 @@ status_code_e amf_smf_handle_ip_address_response(
                  reinterpret_cast<char*>(response_p->pdu_session_id));
     return rc;
   }
+  if (response_p->result != SGI_STATUS_OK) {
+    rc = amf_pdu_session_establishment_reject(
```
rsarwad, line 932: Add an error log to understand from where exactly Pdu Session Reject is sent

rsarwad (approved):
LGTM

panyogesh (approved):
LGTM

PR #13481, opened 2022-08-03
chore(dp): add 401 GRANT_CONFLICT response handling

hejjestemjarek:

## Summary
Add handling of 401 GRANT CONFLICT response to Configuration Controller, so upon receiving such response from SAS grant is created in UNSYNC state, which should get it relinquished and resolve the conflict.

## Test Plan
Added unit tests cases.

## Additional Information

### DIFF
--- a/dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
+++ b/dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
@@ -176,7 +176,7 @@ def process_grant_response(obj: ResponseDBProcessor, response: DBResponse, sessi
     if response.response_code == ResponseCodes.GRANT_CONFLICT.value:
         _unsync_conflict_from_response(obj, response, session)
         return
-    elif response.response_code != ResponseCodes.SUCCESS.value:
+    if response.response_code != ResponseCodes.SUCCESS.value:
         _remove_grant_from_response(response, session, unset_freq=True)
         return
 
@@ -384,16 +384,16 @@ def _terminate_all_grants_from_response(response: DBResponse, session: Session) 
 def _unsync_conflict_from_response(obj: ResponseDBProcessor, response: DBResponse, session: Session) -> None:
     state = obj.grant_states_map[GrantStates.UNSYNC.value]
 
-    conflicts_ids = set(response.payload.get("response", {}).get("responseData", []))
-    existing_grants = session.query(DBGrant).filter(DBGrant.grant_id.in_(conflicts_ids)).all()
-    existing_grants_ids = {g.id for g in existing_grants}
+    conflicts_ids = response.payload.get("response", {}).get("responseData", [])
+    existing_grants = session.query(DBGrant.grant_id).filter(DBGrant.grant_id.in_(conflicts_ids)).all()
+    existing_grants_ids = {g.grant_id for g in existing_grants}
 
-    unsync_grants = list(conflicts_ids - existing_grants_ids)
-    if not unsync_grants:
+    for grant_id in conflicts_ids:
+        if grant_id in existing_grants_ids:
+            continue
+
+        _create_grant_from_response(response, state, session, grant_id=grant_id)
         return
-
-    grant_id = unsync_grants[0]
-    _create_grant_from_response(response, state, session, grant_id=grant_id)
 
 
 def _unregister_cbsd(response: DBResponse, session: Session) -> None:
--- a/dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
+++ b/dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
@@ -334,10 +334,11 @@ def _remove_grant_from_response(
     if not grant:
         return
 
+    logger.info(f'Terminating grant {grant.grant_id}')
+
     if unset_freq:
         unset_frequency(grant)
     session.delete(grant)
-    logger.info('Terminated grant')
 
 
 def _update_grant_from_request(response: DBResponse, grant: DBGrant) -> None:
--- a/dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
+++ b/dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
@@ -171,21 +171,17 @@ def process_grant_response(obj: ResponseDBProcessor, response: DBResponse, sessi
     Returns:
         None
     """
-
-    grant = _get_grant_from_response(response, session)
-
     # Grant response codes worth considering here also are:
     # 400 - INTERFERENCE
-    # 401 - GRANT_CONFLICT
-    # Might need better processing, for now we set the state to IDLE in all cases other than SUCCESS
-    if response.response_code != ResponseCodes.SUCCESS.value:
-        if grant:
-            unset_frequency(grant)
-            session.delete(grant)
-            logger.info(f'process_grant_responses: Terminated grant')
+    if response.response_code == ResponseCodes.GRANT_CONFLICT.value:
+        _unsync_conflict_from_response(obj, response, session)
+        return
+    elif response.response_code != ResponseCodes.SUCCESS.value:
+        _remove_grant_from_response(response, session, unset_freq=True)
         return
 
     new_state = obj.grant_states_map[GrantStates.GRANTED.value]
+    grant = _get_grant_from_response(response, session)
     if not grant:
         grant = _create_grant_from_response(response, new_state, session)
     else:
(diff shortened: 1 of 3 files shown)


dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
```diff
@@ -337,6 +327,19 @@ def _create_grant_from_response(
...
+        response: DBResponse, session: Session, unset_freq: bool = False,
+) -> None:
+    grant = _get_grant_from_response(response, session)
+    if not grant:
+        return
+
+    if unset_freq:
+        unset_frequency(grant)
+    session.delete(grant)
+    logger.info('Terminated grant')
```
xbend: Please add at least grantId to the logger for tracking purposes and log debugging.

hejjestemjarek: Added.

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
```diff
@@ -322,13 +312,13 @@ def _create_grant_from_response(
         response: DBResponse,
         state: DBGrantState,
         session: Session,
+        grant_id: str = None,
 ) -> Optional[DBGrant]:
-    cbsd_id = response.payload.get(CBSD_ID) or response.request.payload.get(CBSD_ID)
-    grant_id = response.payload.get(GRANT_ID) or response.request.payload.get(GRANT_ID)
+    grant_id = grant_id or response.grant_id
```
jkmar, line 317: it looks like duplication of `response.grant_id`

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
```diff
@@ -171,21 +171,17 @@ def process_grant_response(obj: ResponseDBProcessor, response: DBResponse, sessi
...
-    # Might need better processing, for now we set the state to IDLE in all cases other than SUCCESS
-    if response.response_code != ResponseCodes.SUCCESS.value:
-        if grant:
-            unset_frequency(grant)
-            session.delete(grant)
-            logger.info(f'process_grant_responses: Terminated grant')
+    if response.response_code == ResponseCodes.GRANT_CONFLICT.value:
+        _unsync_conflict_from_response(obj, response, session)
+        return
+    elif response.response_code != ResponseCodes.SUCCESS.value:
```
jkmar: there was return in previous `if`, so `el` is redundant

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
```diff
@@ -377,6 +381,21 @@ def _terminate_all_grants_from_response(response: DBResponse, session: Session)
     session.query(DBChannel).filter(DBChannel.cbsd == cbsd).delete()
 
 
+def _unsync_conflict_from_response(obj: ResponseDBProcessor, response: DBResponse, session: Session) -> None:
+    state = obj.grant_states_map[GrantStates.UNSYNC.value]
+
+    conflicts_ids = set(response.payload.get("response", {}).get("responseData", []))
+    existing_grants = session.query(DBGrant).filter(DBGrant.grant_id.in_(conflicts_ids)).all()
```
jkmar: this fetches all fields, while you need only ids

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
```diff
@@ -377,6 +381,21 @@ def _terminate_all_grants_from_response(response: DBResponse, session: Session)
     session.query(DBChannel).filter(DBChannel.cbsd == cbsd).delete()
 
 
+def _unsync_conflict_from_response(obj: ResponseDBProcessor, response: DBResponse, session: Session) -> None:
+    state = obj.grant_states_map[GrantStates.UNSYNC.value]
+
+    conflicts_ids = set(response.payload.get("response", {}).get("responseData", []))
```
jkmar: I don't think you need a set for ids in response data, I'd rather just iterate this list and for of the check if grant with given id exists (you could still use set for existing grant ids)

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
```diff
@@ -188,6 +179,72 @@ def test_grant_state_after_response(
...
+        (
+            GRANT_REQ, ResponseCodes.GRANT_CONFLICT.value,
+            ['test_grant_id_for_1', 'test_grant_id_for_2'],
+            [GrantStates.GRANTED.value, GrantStates.UNSYNC.value],
+        ),
+        (
+            GRANT_REQ, ResponseCodes.TERMINATED_GRANT.value, None, [],
+        ),
+        (
+            HEARTBEAT_REQ, ResponseCodes.SUCCESS.value, None, [GrantStates.AUTHORIZED.value],
```
jkmar, line 198: Why there are heartbeat in this test? Wasn't this 401 handling suppose to be only for grant requests?

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
```diff
@@ -188,6 +179,72 @@ def test_grant_state_after_response(
...
+
+        # When
+        self._process_response(
+            request_type_name=request_type_name, response=response, db_requests=db_requests,
+        )
+
+        # Then
+        self.assertListEqual(
+            expected_grants_states,
+            [g.state.name for g in self.session.query(DBGrant).all()],
```
jkmar: this can be flaky, since query is not necessarily ordered

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
```diff
@@ -188,6 +179,72 @@ def test_grant_state_after_response(
...
+        ),
+        (
+            HEARTBEAT_REQ, ResponseCodes.UNSYNC_OP_PARAM.value, None, [GrantStates.UNSYNC.value],
+        ),
+        (
+            RELINQUISHMENT_REQ, ResponseCodes.SUCCESS.value, None, [],
+        ),
+    ])
+    @responses.activate
+    def test_preexisting_grant_state_after_response(
```
jkmar, line 214: What is the difference between this test and the previous one?

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
hejjestemjarek: True, removed `el`.

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
hejjestemjarek: It's not.   

When receiving `401 GRANT CONFLICT` conflicting grants are listed in `response.payload.response.responseData`. We have to iterate over them and create the first one that does not exist as unsync grant.   
To my understanding it might be the same grant as in `response.grant_id`, but it might not. Passing `grant_id` as optional arg is suboptimal way of handling this situation, but the only one I thought about that wouldn't require major refactor of `response_processing.py`. 

If we're sure that grant from `response.payload.response.responseData` that needs to be created as unsync will be the same grant that the one in`response.grant_id` then it would simplify the code a lot.

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
hejjestemjarek: Changed.

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
hejjestemjarek: True.

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
hejjestemjarek: 401 handling is only for grant requests, but this test covers more cases. Having preexisting grant in the scenario allows to test cases where grant already exists and has to be removed or it's state must be changed. It increased code coverage in `response_processing.py` from 88% to 95%.    
Code for these cases was present before and was working, it just wasn't covered in tests.

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
hejjestemjarek: This one covers scenario when one of conflicting grants already exists. It's allows to ensure that preexisting grant's state is not changed while handling 401, only new one is created as unsync. This test also covers few additional cases as to my previous comment.

dp/cloud/python/magma/configuration_controller/tests/unit/test_response_processor.py
hejjestemjarek: Didn't consider this, and since we can have multiple grants in different states it might become the real issue. Added `order_by`.

jkmar (approved):

PR #13470, opened 2022-08-01
feat(agw): Write per subscriber APN specific data into subscriberDB store when MME receives Update Location Answer

vroon2703:
## Summary

This PR has the following new functions: 

1. SQLite functions that allow adding subscribers 
2. A function to convert ula response to subscriber data which is then written into the sqlite database

## Test Plan
Unit tests for the function that converts ula to sub_data
Test that an attach/detach is still successful 

## Additional Information

- [ ] This change is backwards-breaking

PR #13495, opened 2022-08-03
chore(orc8r): adding s1ap testing for cloud-based authentication

vroon2703:
## Summary

Adding test case for cloud-based authentication for the changes made in https://github.com/magma/magma/pull/13470

changes made: 

1. Added a register subscriber function to dev_tools. This function adds one subscriber with IMSI IMSI001010000000001. 
2. Updated the ConfigUE function to now call the register subscriber function in the case that a cloud flag is set to true. More specificallly, if the flag is true, then the subscriber is added using another function called add_sub_cloud_and_update_nw_config which registers a client as well updates the network configuration. 
3. Once ConfigUE is called, then the test restarts all the services using magmaUtil's restart_all_services

## Test Plan

Will check that the test case passes.


Neudrino (changes requested):
This should have been opened as "Draft pull request" to avoid so many reviewer groups to get notified, when it is still far from finished.

- A lot of checks fail.
- This needs a clean rebase on master.
- There seem to be lot of uncorrelated changes in this PR.

PR #13498, opened 2022-08-04
fix(nms): Repair hot module replacement

### LINKED ISSUE #13496: Repair Hot Module Replacement
Hot Module Replacement is not fully setup and we use the error message overlay.
It should either be fixed or removed.

thmsschmitt:
## Summary

Renable hot module replacement using [react fast refresh](https://javascript.plainenglish.io/react-fast-refresh-the-new-react-hot-reloader-652c6645548c).
Also make sure that this work when upgrading to webpack 5. 

- fixes #13496

## Test Plan

- rebuild the docker container
- open the app in the browser
- change some code
   - make sure the code changes are reflected in the browser without reloading
- break the code
  - make that an error page is shown
- repair the code
  - make sure that error page disappears


nms/babel.config.js
```diff
@@ -11,33 +11,46 @@
...
-  ignore: ['node_modules'],
-  presets: [
-    [
-      '@babel/preset-env',
-      {
-        targets: {
-          node: 'current',
-          chrome: '58',
+function isWebpack(caller) {
+  return !!(caller && caller.name === 'web');
```
Neudrino, line 15: Why `!!`? Would `( a && a.name === 'web')` not already give a truthy back?

Neudrino (approved):

nms/babel.config.js
thmsschmitt: I just followed there example https://babeljs.io/docs/en/config-files#apicallercb :) 
It is not clear to me how exactly the return value is used internally to configure the cache, it might make a difference if there are three or two possible return values.

PR #13503, opened 2022-08-04
chore(ci): reboot after dist upgrade in fabfile

maxhbr:
## Summary

fab does do a dist upgrade, but that only has full effect if there is a
reboot afterwards (e.g. to load new kernel modules).

## Test Plan

tbd.


sebathomas:
Python formatting is wrong:
https://github.com/magma/magma/runs/7672630528?check_suite_focus=true

sebathomas:
I think here it would be good to do an integration test run on your fork before the merge.

maxhbr:
> I think here it would be good to do an integration test run on your fork before the merge.


> I think here it would be good to do an integration test run on your fork before the merge.

I have done that before creation of the PR

sebathomas (approved):

PR #13509, opened 2022-08-04
chore(dp): make CRLValidator always enabled for SAS POSTs, add error handling

hejjestemjarek:

## Summary
`CRLValidator` is used only to validate specific cert against its CRLs, because `requests` package does its own certificates validation upon sending requests. Additionally CRLs are not required in certs, so we cannot revoke a cert because for example we had a problem while trying to download CRLs.   
With this `CRLValidator` required additional changes to be production ready:
* make SAS CRL validation always enabled for Configuration Controller
* mark certificate as valid  if unable to get cert or CRLs
* add error handling for getting certs and CRLs 

## Test Plan
* Added new test cases covering the changes made

## Additional Information

### DIFF
--- a/dp/cloud/python/magma/configuration_controller/crl_validator/certificate.py
+++ b/dp/cloud/python/magma/configuration_controller/crl_validator/certificate.py
@@ -10,27 +10,28 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+import logging
 import socket
 import ssl
-from typing import List
+from typing import List, Optional
 
 import requests
 from cryptography import x509
-from cryptography.hazmat.backends import default_backend
-from requests.exceptions import SSLError
+from requests.exceptions import RequestException, SSLError
 
-_x509_backend = default_backend()
+logger = logging.getLogger(__name__)
 
 
-def get_certificate(
-        hostname: str,
-        port: int = 443,
-) -> x509.Certificate:
-    """ Retrieve certificate of given host.
+def get_certificate(hostname: str, port: int = 443) -> Optional[x509.Certificate]:
+    """ Retrieve SSL certificate of given host.
 
     Args:
         hostname: host from which certificate will be retrieved
         port: SSL port of the host
+
+    Raises:
+        ConnectionError: if there was an error while trying to get certificate from host
+        ValueError: if data received from host is not valid DER certificate
 
     Returns:
         x509.Certificate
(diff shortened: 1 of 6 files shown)


dp/cloud/python/magma/configuration_controller/crl_validator/certificate.py
```diff
@@ -10,46 +10,56 @@
...
-        binary_certificate = sock.getpeercert(binary_form=True)
-        sock.shutdown(socket.SHUT_RDWR)
+    try:
+        with socket.create_connection((hostname, port)) as sock:
+            sock = context.wrap_socket(sock, server_hostname=hostname)
+            binary_certificate = sock.getpeercert(binary_form=True)
+            sock.shutdown(socket.SHUT_RDWR)
+    except socket.error as e:
+        msg = f'Unable to get SSL certificate for {hostname}: {e}'
+        logger.warning(msg)
```
jkmar: we are logging this message multiple times (here and in `get_certificate_crls`, can we log only there (or maybe in top level not to have logs cluttered)

dp/cloud/python/magma/configuration_controller/crl_validator/crl_validator.py
```diff
@@ -72,11 +74,19 @@ def is_valid(self, url: str) -> bool:
         try:
             with self._lock.read:
                 certificate = self._certificates[host]
-                crls = self._crls[certificate.serial_number]
+                if certificate is CERTIFICATE_UNAVAILABLE:
```
jkmar: shouldn't we use `==` instead of `is`?

dp/cloud/python/magma/configuration_controller/run.py
```diff
@@ -83,7 +83,7 @@ def run():
         max_overflow=config.SQLALCHEMY_ENGINE_MAX_OVERFLOW,
     )
     session_manager = SessionManager(db_engine=db_engine)
-    ssl_validator = CRLValidator(urls=[config.SAS_URL]) if config.VERIFY_SAS_CRL else None
+    ssl_validator = CRLValidator(urls=[config.SAS_URL])
```
jkmar, line 86: Can crl verification still be disabled?

jkmar (commented):
This pr besides adding error handling also changes some logic regarding crl verifier enablement, can pr title be updated to reflect that?

dp/cloud/python/magma/configuration_controller/crl_validator/certificate.py
hejjestemjarek: Moved this logging from `get_certificate` to `update_certificates`.   
Error in `get_certificate_crls` is about being unable to get CRL, not the cert itself, and is logged and suppressed after that.

dp/cloud/python/magma/configuration_controller/crl_validator/crl_validator.py
hejjestemjarek: Changed to `==`.

dp/cloud/python/magma/configuration_controller/run.py
hejjestemjarek: It's required by certification process to perform CRL validation when making requests to SAS.   
It's always enabled and cannot be disabled, but it fetches only SAS cert and CRLs and is used only in `RequestRouter.post_to_sas` method.

jkmar (approved):

xbend (approved):

PR #13516, opened 2022-08-04
fix(agw): adding sqlite3 dependencies

vroon2703:
Signed-off-by: vroon2703 <varun27@fb.com>

## Summary
When cloud subscriber authentication is done, the MME writes to the subscriberdb SQLite database. This requires cpp sqlite functions. This PR adds the SQLite dependencies for those functions. 

## Test Plan
Before this change:
ls /usr/include/*sql* will return nothing
After this change:
ls /usr/include/*sql* will return 
/usr/include/sqlite3.h  /usr/include/sqlite3ext.h


lte/gateway/deploy/roles/magma/tasks/main.yml
```diff
@@ -517,3 +517,12 @@
...
     remote_src: yes
   when: full_provision
+
+# TODO: preburn this and add it under MME(OAI) dependencies
+- name: Install SQLite3 C++ library
+  retries: 5
+  apt:
+    state: latest
+    pkg:
+      - libsqlite3-dev
```
nstng, line 527: Would this be better suited under "Install common Magma dependencies"?

lte/gateway/c/core/BUILD.bazel
```diff
@@ -82,6 +82,7 @@ MME_SRCS = [
     "oai/lib/sms_orc8r_client/SMSOrc8rClient.cpp",
     "oai/lib/sms_orc8r_client/itti_msg_to_proto_msg.cpp",
     "oai/lib/sms_orc8r_client/sms_orc8r_client_api.cpp",
+    "oai/lib/store/sqlite.cpp",
```
nstng: Changes in this file must be done in #13470 - all files must be present during build time.

bazel/external/system_libraries.BUILD
```diff
@@ -115,3 +115,8 @@ native_binary(
     src = "usr/local/bin/asn1c",
     out = "asn1c",
 )
+
+cc_library(
```
nstng: This change should be done in #13470.

lte/gateway/docker/services/c/Dockerfile
```diff
@@ -89,7 +89,8 @@ RUN apt-get update && apt-get install -y \
   libtspi-dev \
   libconfig-dev \
   libgmp3-dev \
-  libczmq-dev
+  libczmq-dev \
```
nstng, line 92: For Bazel CI the dependency needs to be added to `.devcontainer/bazel-base/Dockerfile`. From a fast glance, I think mme is actually not build with Bazel in this Dockerfile - but imo it does not hurt to already add it (at some point mme will be build here with Bazel too).

nstng (changes requested):
Hi, see comments for needed changes. You can test these changes in context with #13470 by cherry-picking this change on your local #13470 branch, rebuilding the VM or the bazel base container and running `bazel build ...` in magma root (be aware, the build will take some time because you do not have caches on your fresh setup).

FYI: the bazel CI runs on the bazel-base Docker container. This container is rebuild only after changes are merged back to master, i.e., CI on this PR will run on container without the sqlite library.

lte/gateway/c/core/BUILD.bazel
```diff
@@ -1070,6 +1070,7 @@ MME_DEPS = [
     "@system_libraries//:libfd",
     "@system_libraries//:libgnutls",
     "@system_libraries//:libnettle",
+    "@system_libraries//:libsqlite3-dev",
```
nstng: The test is failing because this library does not exists. This change also must be done in #13470.

nstng (changes requested):

lte/gateway/deploy/roles/magma/tasks/main.yml
ssanadhya: @nstng , @vroon2703  note that "Install common Magma dependencies" is a preburn action, so moving this library to that action will only be effective if we update the `magma-dev` box base image. It will not install the library if we just destroy and re-build the VM locally or in CI.

ssanadhya: Since updating the magma-dev box image is out of scope for @vroon2703 's project plan, my suggestion would be to keep the install step for sqlite3 package as a `full_provision` action for this PR and someone from DevOps team can take over the box update task separately.

nstng: @ssanadhya nice catch, I overlooked the "when" - sorry @vroon2703 you need your change from before with the comment that this dependency should be added to "Install common Magma dependencies" when the next preburn imaged is created.

ssanadhya (approved):
Lgtm

lte/gateway/deploy/roles/magma/tasks/main.yml
vroon2703: Done!

ssanadhya:
@nstng , I am force merging this to unblock PR #13470 . I checked that all the requested changes have been made and all required CI checks are passing. If further fixes are needed, we can create a new PR.

PR #13517, opened 2022-08-05
feat(dp): Merge get state and enodebd update

WojSad:
feat(dp): Merge get state and enodebd update

## Summary

Merge get state and enodebd update
Remove old Radio Controller API.
There are no longer separate gRPC calls for getting cbsd state and updating CBSD by enodebd
enodebd does both in one step - updates a cbsd and in return receives its state.
Number of sql queries has been decreased too.
The whole logic has been moved from Python to Go.
Changed Enodebd's state machine behaviour accordingly.
Removed obsolete files related to old Python API.

## Test Plan
Unit / Integration tests for DP included
Unit tests for Enodebd's state machines modified


dp/cloud/go/services/dp/obsidian/log/handlers.go
```diff
@@ -111,14 +111,14 @@ func getListLogsHandler(client *elastic.Client) func(c echo.Context) error {
 }
 
 type LogsFilter struct {
-	LogFrom             string
-	LogTo               string
-	Name                string
-	SerialNumber        string
-	FccId               string
-	ResponseCode        *int64
-	BeginTimestampMilli *int64
```
jkmar, line 120: Why we are dropping support for milliseconds? (we also need to change swagger and api)

dp/cloud/go/services/dp/storage/cbsd_manager.go
```diff
@@ -99,13 +100,35 @@ func (c *cbsdManager) UpdateCbsd(networkId string, id int64, data *MutableCbsd)
...
+		return runner.fetchDetailedCbsd(sq.Eq{"cbsd_serial_number": data.CbsdSerialNumber}, grantJoinClause)
 	})
-	return cbsd.(*DBCbsd), makeError(err, c.errorChecker)
+	if err != nil {
+		return nil, makeError(err, c.errorChecker)
+	}
+	return result.(*DetailedCbsd), nil
+}
+
+func getGrantJoinClauseForEnodebdUpdate() sq.Sqlizer {
```
jkmar: can we move it under public methods?

dp/cloud/go/services/dp/storage/cbsd_manager.go
```diff
@@ -99,13 +100,35 @@ func (c *cbsdManager) UpdateCbsd(networkId string, id int64, data *MutableCbsd)
...
+	grantJoinClause := getGrantJoinClauseForEnodebdUpdate()
+	result, err := sqorc.ExecInTx(c.db, nil, nil, func(tx *sql.Tx) (interface{}, error) {
 		runner := c.getInTransactionManager(tx)
-		cbsd, err := runner.enodebdUpdateCbsd(data)
-		return cbsd, err
+		err := runner.enodebdUpdateCbsd(data)
+		if err != nil {
+			return nil, err
+		}
+		return runner.fetchDetailedCbsd(sq.Eq{"cbsd_serial_number": data.CbsdSerialNumber}, grantJoinClause)
```
jkmar, line 111: Hmm, we are fetching cbsd twice (once in enodebdUpdateCbsd and the other time here), I suggest adding a todo or creating a ticket to optimize it

dp/cloud/go/services/dp/storage/cbsd_manager.go
```diff
@@ -245,28 +269,35 @@ func (c *cbsdManagerInTransaction) updateCbsd(networkId string, id int64, data *
 	return err
 }
 
-func (c *cbsdManagerInTransaction) enodebdUpdateCbsd(data *DBCbsd) (*DBCbsd, error) {
+func (c *cbsdManagerInTransaction) enodebdUpdateCbsd(data *DBCbsd) error {
 	identifiers := []string{"cbsd_serial_number", "network_id"}
 	maskFields := append(identifiers, getEnodebdWritableFields()...)
+	maskFields = append(maskFields, "single_step_enabled")
```
jkmar: We can do this in one step:
```
maskFields = append(identifiers, "single_step_enabled", getEnodebdWritableFields()...)
```
And also a question why we are changing "single_step_enabled" (shouldn't this be only set by users)?

dp/cloud/go/services/dp/servicers/cbsd_manager.go
```diff
@@ -267,6 +297,9 @@ func grantsFromDatabase(grants []*storage.DetailedGrant) []*protos.GrantDetails
 }
 
 func makeErr(err error, wrap string) error {
+	if err == nil {
```
jkmar: not sure about this, especially since it's used only in one place, I think it's better to check error before calling this function, since usually when you have a function with multiple return arguments and error is returned all of the other parameters should rather be nil/zero values

dp/cloud/go/services/dp/servicers/cbsd_manager_test.go
```diff
@@ -571,13 +699,27 @@ func (s *stubCbsdManager) DeregisterCbsd(networkId string, id int64) error {
...
+	expectedLog := p.getExpectedLog(log)
+	if expectedLog == nil {
+		p.t.Fail()
+	}
+	assert.Equal(p.t, expectedLog, log)
+	return nil
+}
+
+func (p *LogPusher) getExpectedLog(log *logs_pusher.DPLog) *logs_pusher.DPLog {
+	if log.LogName == "EnodebdUpdateCbsd" {
```
jkmar: `switch`

dp/cloud/go/services/dp/storage/cbsd_manager.go
```diff
@@ -245,28 +269,35 @@ func (c *cbsdManagerInTransaction) updateCbsd(networkId string, id int64, data *
...
+	if ShouldEnodebdUpdateInstallationParams(cbsd, data) {
+		cols := append(getEnodebdWritableFields(), "should_deregister")
+		columns = append(columns, cols...)
+		data.ShouldDeregister = db.MakeBool(true)
 	}
-	data.ShouldDeregister = db.MakeBool(true)
-	columns := append(getEnodebdWritableFields(), "should_deregister")
-	models, err := db.NewQuery().
+
+	data.LastSeen = db.MakeTime(clock.Now().UTC())
```
jkmar: Can we pass this in data (when calling this function) and not set/change it here?

dp/cloud/go/services/dp/storage/cbsd_manager.go
```diff
@@ -245,28 +269,35 @@ func (c *cbsdManagerInTransaction) updateCbsd(networkId string, id int64, data *
...
+	_, err = db.NewQuery().
 		WithBuilder(c.builder).
 		From(data).
 		Select(db.NewIncludeMask(identifiers...)).
```
jkmar: I think we no longer want that and we should select nothing (or maybe just an id)

dp/cloud/go/services/dp/builders/builders.go
```diff
@@ -211,22 +216,42 @@ type DBGrantBuilder struct {
...
+	}
+	return b
+}
+
 func (b *DBGrantBuilder) WithId(id int64) *DBGrantBuilder {
 	b.Grant.Id = db.MakeInt(id)
 	return b
 }
 
+func (b *DBGrantBuilder) WithLowFrequency(f int64) *DBGrantBuilder {
```
jkmar: Is there a case where we only want to set one of the grant params? I suggest to combine them

dp/cloud/go/services/dp/storage/diff.go
```diff
@@ -9,13 +9,13 @@ const (
...
 
-func ShouldENodeBDUpdate(prev *DBCbsd, next *DBCbsd) bool {
+func ShouldEnodebdUpdateInstallationParams(prev *DBCbsd, next *DBCbsd) bool {
 	return canUpdate(prev) &&
 		(paramsChanges(next, prev) || coordinatesChanged(next, prev))
 }
 
 func canUpdate(prev *DBCbsd) bool {
-	return !prev.CpiDigitalSignature.Valid
+	return prev.SingleStepEnabled.Bool == true && !prev.CpiDigitalSignature.Valid
```
jkmar: I think we should also update mutlistep cbsd (for example with antenna gain) we did that previously

dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
```diff
@@ -630,32 +614,26 @@ def _check_for_cbsd(self, serial_number: str, should_exist: bool = True) -> Opti
         if should_exist:
             return cbsds[0]
 
+    def _get_current_sas_filters(self, serial_number: str) -> Dict[str, Any]:
```
jkmar: why? it has nothing to do with class

dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
```diff
@@ -50,20 +54,22 @@
...
 
 @pytest.mark.orc8r
-class DomainProxyOrc8rTestCase(DomainProxyIntegrationTestCase, Orc8rIntegrationTestCase):
+class DomainProxyOrc8rTestCase(Orc8rIntegrationTestCase):
     def setUp(self) -> None:
+        self.now = now()
         self.serial_number = self._testMethodName + '_' + str(uuid4())
         # TODO why do we do that? Setup was supposed to be done by deployment...
         self.prometheus_url = f'http://{config.PROMETHEUS_SERVICE_HOST}:{config.PROMETHEUS_SERVICE_PORT}'
+        self.update_request = EnodebdUpdateCbsdRequest(serial_number=self.serial_number)
```
jkmar: why instance variable?

dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
```diff
@@ -75,138 +81,103 @@ def test_cbsd_sas_single_step_flow(self):
...
-            .with_serial_number(self.serial_number) \
-            .with_cbsd_category("b")
-        self.when_cbsd_is_created(builder.payload)
-        cbsd = self.when_cbsd_is_fetched(self.serial_number)
-        self.then_cbsd_is(
-            cbsd, builder.with_indoor_deployment(False).with_is_active(False).with_state(UNREGISTERED).payload,
-        )
-
-        req = EnodebdUpdateCbsdRequest(
+        update_request_1 = update_request_2 = EnodebdUpdateCbsdRequest(
```
jkmar: I'm not sure about this, since after this operation both `update_request_1` and `update_request2` point to the same object and later you change `update_request_2`.
Here is an example:
```
class A:
  def __init__(self):
    self.v = 0
x = y = A()
y.v = 1
print(x.v) # this prints 1 instead of 0
```

WojSad: it actually doesn't matter that they point to the same request, but I'll remove update_request_2 since in fact it's not needed

dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
WojSad: so that I don't need to declare it in every single test case and so that I have access to self.serial_number

dp/cloud/go/services/dp/storage/cbsd_manager.go
WojSad: We need network_id and serial_number for Logs

dp/cloud/go/services/dp/storage/cbsd_manager.go
WojSad: ok

dp/cloud/go/services/dp/servicers/cbsd_manager_test.go
WojSad: ok

dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
WojSad: moved out of class

dp/cloud/go/services/dp/servicers/cbsd_manager.go
WojSad: changed

dp/cloud/go/services/dp/builders/builders.go
WojSad: done

dp/cloud/go/services/dp/storage/cbsd_manager.go
```diff
@@ -245,28 +254,33 @@ func (c *cbsdManagerInTransaction) updateCbsd(networkId string, id int64, data *
...
-		return nil, err
+		return err
 	}
-	if !ShouldENodeBDUpdate(cbsd, data) {
-		return cbsd, nil
+
+	columns := []string{"last_seen"}
+
+	if ShouldEnodebdUpdateInstallationParams(cbsd, data) {
+		cols := append(getEnodebdWritableFields(), "should_deregister")
```
jkmar, line 271: I'd rather do this `columns = append(getEnodebdWritableFields(), "should_deregister", "last_seen")` to avoid unnecessary copying (3 times as much)

jkmar (approved):

WojSad:
@magma/approvers-bazel @magma/approvers-agw please have a look at the bazel and agw parts of the PR

vktng (approved):
Bazel changes LGTM.

sebathomas (approved):
Enodebd changes looking good.

PR #13523, opened 2022-08-05
chore: Allow Bazel commands to run in repository subdirectories

### LINKED ISSUE #13102: Bazel commands only work in the magma directory

vktng:

## Summary

Since https://github.com/magma/magma/pull/12987 was merged, the bazel commands will only work in the magma folder, because the cache setup was reworked to use a relative path: bazel/bazelrcs/cache.bazelrc
This PR changes the path to use the workspace root.

## Test Plan

Try using Bazel in subdirectories.
For example: `cd $MAGMA_ROOT/lte/gateway && bazel build //:buildifier`

## Additional Information

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -48,7 +48,7 @@ run:remote_caching_ro --remote_download_toplevel
 
 # Import the bazel caching config, with default disk cache,
 # which in CI gets overwritten by the remote caching config. 
-import bazel/bazelrcs/cache.bazelrc
+import %workspace%/bazel/bazelrcs/cache.bazelrc
 
 # TEST CONFIGS
 # Bazel test runtime default: PATH=/bin:/usr/bin:/usr/local/bin


.bazelrc
```diff
@@ -48,7 +48,7 @@ run:remote_caching_ro --remote_download_toplevel
 
 # Import the bazel caching config, with default disk cache,
 # which in CI gets overwritten by the remote caching config. 
-import bazel/bazelrcs/cache.bazelrc
+import /var/cache/cache.bazelrc
```
jheidbrink: We could also use `import %workspace%/bazel/bazelrcs/cache.bazelrc`, see https://docs.bazel.build/versions/main/guide.html#imports

vktng: This is a really clean solution, the other PR setting up symbolic links is not even needed.

LKreutzer: Is it possible to replace all links by `%workspace%/.bazel-cache/` etc?

LKreutzer (approved):
Tested locally

.bazelrc
vktng: I tried `%workspace%/.bazel-cache/` and it works fine.

LKreutzer: Thanks, good to know. Then we might refactor the mechanics at some point to get rid of all the linking. But for now let us just keep it as it is as this is a larger refactoring.

vktng: I stand corrected:
workspace_not_working_bazel_cache
But it might work with a little more thought put into it.

jheidbrink (approved):

PR #13524, opened 2022-08-05
chore: Set symbolic links in the VMs and the devcontainer for the baz…

vktng:
…elrc

## Summary

Since https://github.com/magma/magma/pull/12987 was merged, the bazel commands only work in the magma folder, because the cache setup was reworked to use a relative path: bazel/bazelrcs/cache.bazelrc
This PR uses a symbolic link to provide an absolute path to this file on all environments. This is a preparation for #13523.

## Test Plan

CI.

## Additional Information

### DIFF
--- a/.devcontainer/bazel-base/Dockerfile
+++ b/.devcontainer/bazel-base/Dockerfile
@@ -17,69 +17,69 @@ RUN echo "Install general purpose packages" && \
     apt-get update && \
     apt-get upgrade -y && \
     apt-get install -y --no-install-recommends \
-        apt-transport-https \
-        apt-utils \
-        # dependencies of FreeDiameter
-        bison \
-        build-essential \
-        ca-certificates \
-        cmake \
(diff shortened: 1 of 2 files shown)


.devcontainer/bazel-base/Dockerfile
```diff
@@ -17,69 +17,69 @@ RUN echo "Install general purpose packages" && \
     apt-get update && \
     apt-get upgrade -y && \
     apt-get install -y --no-install-recommends \
-        apt-transport-https \
```
LKreutzer, line 20: Can we keep the formatting as it is in this file? The extra spaces are here because these are packages for the `apt-get install` and it is more readable this way.

LKreutzer (changes requested):

vktng:
This PR is no longer needed, because there is a better solution and #13523 does not need symbolic links anymore.

PR #13531, opened 2022-08-08
feat(ci): Add federated integ test for HSS

orion-belt:
Signed-off-by: kharade <rohan.kharade@openairinterface.org>

## Summary

This PR adds Federated Integ Test  to validate HSS s6a interface. Here we are reusing existing  [lte-integ-tests](https://github.com/magma/magma/tree/master/lte/gateway/python/integ_tests/s1aptests) in fedrated mode.
#### Following tests are added in this PR which leverages HSS s6a interface over federated gateway -
```bash
test_attach_auth_failure.py
test_no_auth_response.py
test_nas_non_delivery_for_auth.py
test_sctp_abort_after_auth_req.py
test_sctp_shutdown_after_auth_req.py
test_no_auth_resp_with_mme_restart_reattach.py
test_send_error_ind_for_dl_nas_with_auth_req.py
test_attach_auth_mac_failure.py
test_no_auth_response_with_mme_restart.py
test_no_security_mode_complete.py
test_no_attach_complete.py
test_attach_detach_security_algo_eea0_eia0.py
test_attach_detach_security_algo_eea1_eia1.py
test_attach_detach_security_algo_eea2_eia2.py
test_attach_security_mode_reject.py
test_attach_missing_imsi.py
test_duplicate_attach.py
test_attach_emergency.py
test_attach_detach_after_ue_context_release.py
test_attach_esm_information_wrong_apn.py
test_attach_detach_secondary_pdn_invalid_apn.py
test_standalone_pdn_conn_req_with_apn_correction.py
```
## Test Plan

We are reusing some of existing and relevant tests from  lte-integ-tests. The [sheet attached](https://docs.google.com/spreadsheets/d/1oI1FvX1AFaYQDuH85oMJsCu7X20tox248mXsuCFes0c/edit#gid=0) describes the priority and relevancy of tests. Please feel free to update sheet accordingly.

## Additional Information


lte/gateway/python/integ_tests/federated_tests/s1aptests/test_attach_detach_after_ue_context_release.py
```diff
@@ -0,0 +1,84 @@
+"""
+Copyright 2020 The Magma Authors.
```
VinashakAnkitAman: 2020 => 2022 (Same for other test cases)

lte/gateway/python/integ_tests/defs.mk
```diff
@@ -278,7 +278,30 @@ s1aptests/test_restore_config_after_non_sanity.py
 
 # Add the s1aptester integration tests with federation gateway
 FEDERATED_TESTS = federated_tests/s1aptests/test_attach_detach.py \
-federated_tests/s1aptests/test_attach_detach_multi_ue.py
+federated_tests/s1aptests/test_attach_detach_multi_ue.py \
```
VinashakAnkitAman: We are adding the same testcases in 2 folders `magma/lte/gateway/python/integ_tests` and `magma/lte/gateway/python/integ_tests/federated_tests/s1aptests/`. Its almost impossible to always maintain same changes in both the places. I see the only difference between the test cases at the 2 places is:
```
<         self._s1ap_wrapper = s1ap_wrapper.TestWrapper(federated_mode=True)
---
>         self._s1ap_wrapper = s1ap_wrapper.TestWrapper()
```

Is it not possible to use the same test cases for both the scenario? We may get this variable `federated_mode` by either passing one argument to pytest command or setting one ENVIRONMENT variable and getting it inside the test case to understand whether `federated_mode` is `true` or `false`. This will reduce the effort to maintain the new changes in both the places.

VinashakAnkitAman (changes requested):
Left some comments

lte/gateway/python/integ_tests/defs.mk
rohanrkharade: We are re-using test here for s6a over FEG but there will be new tests (or modification of some of these tests) for validation Gx, Gy, SGs, and 5G interfaces like nPCF etc. I also noted that current LTE integ tests does not cover test related to Gx, Gy, SGs etc.

Also when we run test from integ_tests, then we need to make sure that, not all test are relevant/or run against Fed setup
e.g. `make fed_integ_test TESTS=s1aptests/test_x2_handover.py`  may not be relevant/supported against fed gateway tests.

What is your view @VinashakAnkitAman @uri200 @voisey ?

lte/gateway/python/integ_tests/defs.mk
```diff
@@ -278,7 +278,30 @@ s1aptests/test_restore_config_after_non_sanity.py
 
 # Add the s1aptester integration tests with federation gateway
 FEDERATED_TESTS = federated_tests/s1aptests/test_attach_detach.py \
```
crasu: Have you tried to run these tests on github actions? How long did it take? Did they run?

rohanrkharade: Tests run is here https://github.com/orion-belt/magma/runs/7745520159?check_suite_focus=true and it takes ~11 min to run these tests.

lte/gateway/python/integ_tests/defs.mk
crasu: I guess if new tests are to be added/changed in the future they could still be copied once significant changes are made.

You could use pytest addoption to pass a federated flag to the tests:
https://docs.pytest.org/en/7.1.x/example/simple.html

To make the test execution for "non feg enabled integ test" impossible, we could fail during test setup for those tests.

crasu (changes requested):

orion-belt:
Hi @crasu 
Thanks for the clarification. I tried to mimic the way the first test which was added by @uri200 and @voisey.

I agree its better to maintain all the S1AP related test which don't have any significant changes in `integ_tests/s1aptests` test folder and remove the folder `federated_tests/s1aptests/` folder at the moment. But later in around a month I will add Gx, Gy, SGs and other 5G related tests so those should be hosted in a dedicated folder like `federated_tests/s1aptests/` as those test are only relevant for fed-integ testing.

orion-belt:
I have addressed fed test run with environment variable now.
https://github.com/orion-belt/magma/runs/7859498266?check_suite_focus=true

rdefosse (approved):

orion-belt:
hi @VinashakAnkitAman @crasu 
Will you please review changes now ?
thanks

crasu:
> hi @VinashakAnkitAman @crasu Will you please review changes now ? thanks

I will do it today

lte/gateway/python/integ_tests/s1aptests/s1ap_wrapper.py
```diff
@@ -81,6 +81,12 @@ def __init__(
         self._s1_util = S1ApUtil()
         self._enBConfig(ip_version)
 
+        if os.environ.get("FED_MODE") == "True":
```
crasu: nit: Please call the env variable FEDERATED_MODE so it matches with the python var. If you want to use one.

docs/readmes/feg/s1ap_federated_tests.md
```diff
@@ -195,14 +192,16 @@ vagrant up magma_trfserver
 Once you have built all of the VMs, you can try to run the tests from the
 `magma_test` VM.
 
+##### Note: Only [these LTE integ](https://github.com/magma/magma/blob/master/lte/gateway/python/integ_tests/defs.mk#L288) tests can be run in federated mode
```
crasu: I guess it would be nicer to use something along the lines of skipif (see https://docs.pytest.org/en/7.1.x/how-to/skipping.html). To just not execute the tests. 

Or even better a custom decorator like in this example: https://stackoverflow.com/questions/67200501/pytest-run-only-tests-with-specific-marker-attribute

lte/gateway/python/integ_tests/Makefile
```diff
@@ -63,6 +63,7 @@ endif
 endif
 
 .PHONY: fed_integ_test
+export FED_MODE = True
```
crasu: nit: I would have liked the pytest option more as it is more explicit.

crasu (changes requested):

lte/gateway/python/integ_tests/Makefile
VinashakAnkitAman: I agree that we should use pytest option like `pytest addoption` instead of using environment variables here. The only issue which I had seen earlier was that we are using `python unittest` along with `pytest` and for accepting these addoption options, we needed to add some hacks in unittest main.

From my end, environment variable is also ok for now. But if you have bandwidth, its always good to check if you can easily use pytest options and implement it.

VinashakAnkitAman (approved):
LGTM. Please have a look on the review comments from @crasu

lte/gateway/python/integ_tests/s1aptests/s1ap_wrapper.py
```diff
@@ -81,6 +81,12 @@ def __init__(
...
+            federated_mode = True
+            print("\n************************* Running the test in Federated Mode")
+        else:
+            print("\n************************* Running the test in Non-Federated Mode")
+
```
Neudrino: Is that not a bit overly bloated? I think something along the following lines would be more concise.
```python
federated_mode = (os.environ.get("FED_MODE") == "True")
print(f"\n *** Running the test in {'Non-' if not federated_mode else ''}Federated Mode")
```

docs/readmes/feg/s1ap_federated_tests.md
Neudrino: - A header is mis-used for an emphasis here. Rather use a bold font.
- Full stop missing.
- "These test" links to one line, but is for me still kind of vague. Could we name those tests explicitly. E.g. "only the attach-detach-tests..."

rohanrkharade: HI @Neudrino 
Sorry for the typos. I have addressed your comments in latest amend.

orion-belt:
hi @crasu 
I agree with your recommendation to use pytest adoption instead of env var.
I will create issue for tracking and address this in separate PR, is that okay ?

crasu (approved):

sebathomas (approved):

Neudrino (approved):

PR #13532, opened 2022-08-08
chore(orc8r): Add sources to initflag library in bazel

LKreutzer:

## Summary

- Sources are added to the `initflag` bazel target
  - The sources are needed for the `init()` function to run during side-effect imports. 

## Test Plan

- `bazel build //orc8r/lib/go/initflag`
- In the `orc8r/lib/go/initflag/initflag.go` file add a line `	panic("This is the init function!")` to the `init` function and run `bazel run //feg/gateway/services/envoy_controller`, see that the `init` function is used. If the sources are not added then the `panic` does not have an effect.

## Additional Information

### DIFF
--- a/orc8r/lib/go/initflag/BUILD.bazel
+++ b/orc8r/lib/go/initflag/BUILD.bazel
@@ -11,10 +11,16 @@
 
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
 
-# This empty library is needed to satisfy the side-effect imports of
+# This library is needed to satisfy the side-effect imports of
 # this package. E.g. in orc8r/lib/go/service/config/service_config.go
 go_library(
     name = "initflag",
+    srcs = [
+        "initflag.go",
+        "parse_flag_debug.go",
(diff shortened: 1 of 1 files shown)


jheidbrink (approved):
I verified that the test plan produces the expected results.

orc8r/lib/go/initflag/BUILD.bazel
```diff
@@ -11,10 +11,16 @@
...
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
 
-# This empty library is needed to satisfy the side-effect imports of
+# This library is needed to satisfy the side-effect imports of
 # this package. E.g. in orc8r/lib/go/service/config/service_config.go
 go_library(
     name = "initflag",
+    srcs = [
+        "initflag.go",
+        "parse_flag_debug.go",
```
jheidbrink, line 20: Do we need this one?

jheidbrink: Never mind

PR #13554, opened 2022-08-09
fix(ci): re-enable backport

tmdzk:
Signed-off-by: tmdzk <timodzik@gmail.com>

## Summary
Re-enabling backport. changing var name cause github doesn't allow you to prepend with GITHUB

## Additional Information

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -19,7 +19,7 @@ jobs:
           python-version: '3.8.10'
       - run: |
           mkdir ~/.magma
-          echo ${{ secrets.GITHUB_TOKEN }} > ~/.magma/github_access_token
+          echo ${{ secrets.GIT_TOKEN }} > ~/.magma/github_access_token
       - name: Install python prerequisites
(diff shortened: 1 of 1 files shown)


maxhbr (approved):
LGTM

.github/workflows/backport.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           python-version: '3.8.10'
       - run: |
           mkdir ~/.magma
-          echo ${{ secrets.GITHUB_TOKEN }} > ~/.magma/github_access_token
+          echo ${{ secrets.GIT_TOKEN }} > ~/.magma/github_access_token
```
Neudrino, line 22: @tmdzk @maxhbr 
This seems rather wrong to me. Does that really fix a thing?
https://docs.github.com/en/actions/security-guides/automatic-token-authentication

maxhbr: ~~I think~~ `GIT_TOKEN` was added as a special secret.

PR #13562, opened 2022-08-10
chore(ci): Run the bazel workflow as a matrix job with relevant configs

### LINKED ISSUE #13363: The Bazel build and test job is run as a matrix job with all relevant configs
The Bazel build and test job is run as a GitHub actions matrix job with all relevant configs.

Some of the bazel configs (use `--config=`) are:
- `native`
- `asan`
- `limit_mem`
- `production`
- `lsan` 
  - Do we need this if we have `production`?
- ... are there others?

Here: start with native and production and create follow-up issue for other values (to be changed after some observation period)

Test locally (on fork) with read-write caches before and look at increase in cache size.
- Might be useful to test with other remote cache, e.g. with buildbuddy to measure the size separately.

Hint: Look at this [example matrix workflow](https://github.com/LKreutzer/GH-actions-testing/blob/main/.github/workflows/matrix_workflow.yml)

LKreutzer:
## Summary

- The bazel workflow is run as a matrix workflow with the configs `native`, `asan` and `production`.
  -  :exclamation:  The setup is also explained in detail on this magma wiki page [:book: How-to-set-up-a-required-matrix-workflow-on-GitHub-actions](https://github.com/magma/magma/wiki/How-to-set-up-a-required-matrix-workflow-on-GitHub-actions). 
  - See detailed analysis on: https://github.com/magma/magma/issues/13363
- :warning: **Just before** this PR is **merged** the branch protection settings need to be changed to **require the following jobs**:
  - [x] **NOT** the overall workflow or `Bazel Build & Test Job` (this is blocking this PR and will lead to wrong requirements)
  - [x] **NOT** the `C/C++ unit tests with Bazel ` (this is blocking this PR and the job was deleted)
  - [x]  `c_cpp_file_check` / `Check if there are non-bazelified c or c++ files` (already true ?)
  - [x]  `python_file_check` / `Check if there are not bazelified python files` (already true ?)
  - [x] `report_result_bazel_build_and_test` / `Bazel build and test status` after n (n=2) days
- The bazel asan C/C++ tests are removed from the agw-workflow as they are now included in the bazel workflow. 
- Minor refactoring: the shell is set to bash and `set -euo pipefail` is used.
  - This makes it possible to use multiple lines for the execution steps. 

## Test Plan

- [x] If the build step is skipped it will still return green at the end: 
  - Test run  https://github.com/LKreutzer/magma/actions/runs/2826379794
  - Test PR on fork: https://github.com/LKreutzer/magma/pull/1
  - Screenshot from 2022-08-10 13-28-27
- [x] If one of the shards fails the job will fail (and the required status will be blocked):
  - Test PR on fork:  https://github.com/LKreutzer/magma/pull/2
  - Screenshot from 2022-08-10 13-28-44
- [x] If all shards are successful the status job will be green (and the required status will be green):
  - Screenshot from 2022-08-10 13-38-21
- For general results see again: **https://github.com/magma/magma/issues/13363** 

## Additional Information

- :exclamation: For the required status of the matrix job to work additional jobs are needed.
  - See https://brunoscheufler.com/blog/2022-04-09-the-required-github-status-check-that-wasnt for the general setup.
  - There is one job that **only runs if all shards of the matrix are successful** and one job that **always runs** and evaluates the results. The final job is the one that should be required.
  - The approach from the blog needs to be adapted due to the path filter step, which may be skipped itself or lead to a skipped bazel job. This can be handled by modifying the if conditional in the last job.  
- Filled cache for `asan` and `production` configs in [this run](https://github.com/LKreutzer/magma/actions/runs/2832131969).

- Related https://github.com/actions/runner/issues/952


nstng:
Idea for merge strategy: remove required settings for the bazel workflow completely, rerun workflows here, merge, re-enable after a couple of days (when rebases are expected)
@LKreutzer what do you think?

LKreutzer:
> Idea for merge strategy: remove required settings for the bazel workflow completely, rerun workflows here, merge, re-enable after a couple of days (when rebases are expected) @LKreutzer what do you think?

That would certainly work - there is a chance of regression during that time, but we also have the slack notifications on master.

.github/workflows/bazel.yml
```diff
@@ -131,13 +153,56 @@ jobs:
...
+
+          echo "The status of this job is determined by the statuses of the previous jobs in this workflow."
+          echo "For more details on this matrix workflow please look at the following wiki page or the PR #13562:"
+          echo "https://github.com/magma/magma/wiki/How-to-set-up-a-required-matrix-workflow-on-GitHub-actions"
+          echo "https://github.com/magma/magma/pull/13562"
+
+          echo "bazel_build_and_test_success: $bazel_build_and_test_success"
+          echo "files_changed: $files_changed"
+
+          if [[ $bazel_build_and_test_success == "true" || $files_changed != 'true' ]];
```
Neudrino: There is a inconsistency here for `'` and `"` usage.

Neudrino (approved):
The Github actions part looks good. Did not check the Bazel part though.

.github/workflows/bazel.yml
LKreutzer: Pushed

nstng (approved):
lgtm

Issue #13537, opened 2022-08-08
Fix flaky cloud tests


sebathomas:
* [x] Stabilize TestSingletonRun
  * Fixed in #13646
* [x] Stabilize TestGetSubscriberState
  * Made less flaky in #13565
  * Update: Looks like the error [can still occur](https://github.com/magma/magma/runs/7896850066), it's just much more rare now. We might think about an explicit sleep?
  * Update: An explicit sleep of 1 second [doesn't seem to help](https://github.com/magma/magma/pull/13701)...
* [x] Prepare internal review

Two cloud tests are frequently failing in the CI:

```
=== Failed
=== FAIL: services/state/indexer/reindex TestSingletonRun (4.08s)
I0704 04:22:09.190181    9485 service_config.go:161] Successfully loaded 'orc8r::accessd' service configs from '/etc/magma/configs/orc8r/accessd.yml'
I0704 04:22:09.834319    9485 service_config.go:161] Successfully loaded 'orc8r::certifier' service configs from '/etc/magma/configs/orc8r/certifier.yml'
I0704 04:22:09.834690    9485 service_config.go:161] Successfully loaded 'orc8r::configurator' service configs from '/etc/magma/configs/orc8r/configurator.yml'
I0704 04:22:09.835375    9485 service_config.go:161] Successfully loaded 'orc8r::device' service configs from '/etc/magma/configs/orc8r/device.yml'
I0704 04:22:09.841370    9485 service_config.go:161] Successfully loaded 'orc8r::state' service configs from '/etc/magma/configs/orc8r/state.yml'
    access.go:59: Test Gateway Certificate SN: 344F38CD5A71A1882E54B0EF229CD9F5
    access.go:59: Test Gateway Certificate SN: F6EA3AC40A3238D0D96284E61E9E3432
    access.go:59: Test Gateway Certificate SN: A2ED53B6720C280A294085442E804445
W0704 04:22:10.435740    9485 service_config.go:189] Cannot find '/etc/magma/configs/orc8r/some_indexerid_0.yml': stat /etc/magma/configs/orc8r/some_indexerid_0.yml: no such file or directory, or Legacy Service Registry Configuration: '/etc/magma/orc8r/some_indexerid_0.yml': stat /etc/magma/orc8r/some_indexerid_0.yml: no such file or directory
[...]

PR #13565, opened 2022-08-10
test(orc8r): Stabilize flaky test TestListSubscriberStates

sebathomas:
## Summary

Most likely, the test is flaky because there's a race condition between the reindexer and the assertion. With this change, the gateway subscriber state is updated before all other state updates, which gives the reindexer more time to run before the state is asserted.

This turns out to make the test reliable without adding any explicit waiting.

Part of #13537.

## Test Plan

I have not been able to get the test red on my machine even in over 100 runs.

- [x] Retrigger it in the CI and check if it's still flaky.
  - 6 successful CI runs on this PR, no failures of TestListSubscriberStates so far. I don't have statistics but I am confident we should have seen an error by now if the test was still as flaky as it is on master.
    - https://github.com/magma/magma/runs/7766828563
    - https://github.com/magma/magma/runs/7767411021
    - https://github.com/magma/magma/runs/7767936329
    - https://github.com/magma/magma/runs/7768551842
    - https://github.com/magma/magma/runs/7769078532
    - https://github.com/magma/magma/runs/7770489016
  - There was a single failed CI run due to another flaky test.
    - https://github.com/magma/magma/runs/7769685515

## Additional Information

### DIFF
--- a/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
+++ b/lte/cloud/go/services/subscriberdb/obsidian/handlers/handlers_test.go
@@ -768,30 +768,7 @@ func TestListSubscriberStates(t *testing.T) {
 	clock.SetAndFreezeClock(t, time.Unix(frozenClock, 0))
 	defer clock.UnfreezeClock(t)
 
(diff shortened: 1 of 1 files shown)


alexzurbonsen (approved):
LGMT

maxhbr (approved):
LGTM

PR #13568, opened 2022-08-10
fix(agw): Fixed file not found error in publishing integ test status

VinashakAnkitAman:
## Title
fix(agw): Fixed file not found error in publishing integ test status

## Summary
Whenever LTE integ test run fails in CI before triggering the actual integ test execution (during system bring-up or configuration phase), the status file goes missing and it leads to file not found exception and CI status is not updated for the integ test. This PR fixes the issue by getting the status of the integ test only if status file exists.

The error log:
```
  File "ci-scripts/firebase_publish_report.py", line 114, in <module>
    args.func(args)
  File "ci-scripts/firebase_publish_report.py", line 68, in lte_integ_test
    with open('test_status.txt', 'r') as file:
FileNotFoundError: [Errno 2] No such file or directory: 'test_status.txt'
Error: Process completed with exit code 1.
```

## Test plan
Verified with sanity and in CI
Successful CI execution log:
https://github.com/VinashakAnkitAman/magma/actions/runs/2832919308
Reference CI run: https://github.com/magma/magma/runs/7752890817?check_suite_focus=true


ci-scripts/firebase_publish_report.py
```diff
@@ -65,8 +65,12 @@ def lte_integ_test(args):
...
     # test as success. Here we read the CI status from file for more accurate
     # lte integ test execution status
-    with open('test_status.txt', 'r') as file:
-        verdict = file.read().rstrip()
+    if os.path.exists("test_status.txt"):
+        with open('test_status.txt', 'r') as file:
+            status_file_content = file.read().rstrip()
+            expected_verdict_list = ["pass", "fail"]
+            if status_file_content in expected_verdict_list:
+                verdict = status_file_content
```
crasu, line 73: If we cannot find the test status should we report a "fail" verdict?

VinashakAnkitAman: Hi @crasu I agree to this point to some extent that if test status is not available that means it is failing. But I believe when we are saying status of lte integ test, then it should be sanity test suite execution status and not the environment setup status. So if it fails during environment setup, this is CI setup issue (because sanity suite may still be passing, but could not execute because of ci environment issues) and the integ test status for that ci run is basically inconclusive. If it fails during the sanity test suite run, then it is actually the failure of lte integ test.

ardzoht (approved):

ci-scripts/firebase_publish_report.py
crasu: Both is fine with me

crasu (approved):

Neudrino (approved):

nstng:
already in 1.8 - setting respective label

Issue #13546, opened 2022-08-08
[Feg] Cleanup references to XWF


ssanadhya:
Once https://github.com/magma/magma/pull/13545 is merged, the XWF code will be removed.
- [x] remove scripts (PR: https://github.com/magma/magma/pull/13592)
- [x] remove frontend xwf implementation (PR: https://github.com/magma/magma/pull/13603)
- [x] remove option to configure xwf from swagger/backend (https://github.com/magma/magma/pull/13627)
- [x] remove the pass through controller in pipelined (PR: https://github.com/magma/magma/pull/13621)
- [x] remove xwf references from helm charts (PR: https://github.com/magma/magma/pull/13619)
- [x] after all PRs have been merged, check for remaining xwf occurrances again (orc8r/cloud/api/v1/ is not considered here, since it is not kept up to date. Regenerating leads to a massive set of changes, 1000+ files.)
   https://github.com/magma/magma/pull/13668
   https://github.com/magma/magma/pull/13667
   https://github.com/magma/magma/pull/13677
- [x] prepare sprint review
- [x] open follow up ticket for radius (https://github.com/magma/magma/issues/13671)
- [x] remove xwf approvers team from Github

Go radius is out of scope of this ticket.

This issue is to track clean-up for XWF references from other components of Magma core, e.g. Federation Gateway (feg)

crasu:
There are only a few references left
```
rg --files-with-matches xwf
CODEOWNERS
bazel/scripts/check_py_bazel.sh
nms/generated/api.ts
orc8r/tools/docker/install_gateway.sh
nms/server/network/routes.ts
nms/server/grafana/handlers.ts
nms/shared/types/network.ts
feg/radius/src/deploy.sh
lte/gateway/python/magma/pipelined/BUILD.bazel
lte/gateway/python/magma/pipelined/service_manager.py
feg/radius/src/server/sse_listener_test.go
feg/radius/src/loader/static_loader.go
feg/radius/src/monitoring/init.go
nms/app/components/layout/__tests__/useSections-test.tsx
lte/gateway/python/magma/pipelined/app/xwf_passthru.py
lte/gateway/python/magma/pipelined/app/BUILD.bazel
orc8r/cloud/api/v1/go/models/network_carrier_wifi_configs.go
feg/radius/lib/go/libgraphql/xwf_customer.go
feg/radius/src/monitoring/scuba/scuba_test.go
feg/radius/src/monitoring/scuba/scuba.go
[...]

PR #13545, opened 2022-08-08
chore(xwf): Remove the Express WiFi directory since code is not being maintained

ssanadhya:
Signed-off-by: Shruti Sanadhya <ssanadhya@fb.com>

## Summary

Support for Express WiFi with Magma is not being maintained any more, hence cleaning up this code.

Once this PR is merged, we will remove the [approvers-xwf](https://github.com/orgs/magma/teams/approvers-xwf) team

## Test Plan

N/A

PR #13570, opened 2022-08-10
fix(xwf): remove xwf reference from installation and deploy scripts

crasu:
## Summary

Remove xwf references in scripts

## Test Plan

Tested if installation still runs 

Relates to #13546


bazel/scripts/check_py_bazel.sh
```diff
@@ -45,7 +45,6 @@ DENY_LIST_NOT_RELEVANT=(
   "./protos"
   "./show-tech"
   "./third_party"
-  "./xwf/gateway/deploy"
```
crasu: can we remove this already before #13545 is merged @LKreutzer

LKreutzer: Removing the line while the folder still exists should lead to failing bazel workflows on PRs. So we should keep it and remove it in #13545.

crasu: ok moved it there

crasu:
@alexzurbonsen  Could you take over this pr?

sebathomas (approved):

alexzurbonsen:
Changes are now part of https://github.com/magma/magma/pull/13570.

PR #13571, opened 2022-08-10
ci(fix): Remove disksize plugin and disksize setting as the ubuntu ba…

crasu:
Remove disksize plugin and disksize setting as the ubuntu base box is already 130GB+

Signed-off-by: Christian Krämer <christian.kraemer@tngtech.com>

## Test Plan

Started the vm without error


Neudrino (approved):

crasu:
Closed in favour of #13587 which removes even more stuff

PR #13559, opened 2022-08-10
Revert "fix(nms): Repair hot module replacement (#13498)"

thmsschmitt:
## Summary

This reverts commit bf71a9175ccb8a103a349af7506d80238b18ef0d.
It somehow broke the e2e build even-though it looks identical to the regular DEV build.  

## Test Plan

e2e test are green again. https://github.com/magma/magma/runs/7760800528?check_suite_focus=true

PR #13572, opened 2022-08-10
fix(nms): Make sure hot module replacement works for the e2e test 

thmsschmitt:
## Summary

This re-enables the hot module replacement as it was initially configured here #13559
The e2e tests were broken because there was an unhandled exception that lead to an overlay that blocked the UI. This has been fixed.

## Test Plan

* e2e tests are green again https://github.com/magma/magma/runs/7770506908?check_suite_focus=true
* The hot module replacement works
   * rebuild the NMS docker container
   * make sure the UI updates when code is changed without reloading


nms/babel.config.js
```diff
@@ -12,7 +12,7 @@
  */
 
 function isWebpack(caller) {
-  return !!(caller && caller.name === 'web');
```
thmsschmitt: `caller.target` was `web` so the plugin was never enabled. Somehow it work without, but I kept it because the docs say you should use it.

nms/app/context/VersionContext.tsx
```diff
@@ -38,23 +38,21 @@ export function VersionContextProvider(props: Props) {
...
+        setOrc8rVersion(`v${version.container_image_version!}`);
+      } catch (e) {
+        enqueueSnackbar?.('failed fetching NMS version information', {
+          variant: 'error',
+        });
+      }
     };
 
     void fetchNmsVersion();
-  });
```
thmsschmitt, line 57: prevent unnecessary requests

Neudrino (approved):

PR #13574, opened 2022-08-10
[Do not merge] rules_docker POC

jheidbrink:
## Summary

Play with rules_docker


lte/gateway/c/li_agent/src/BUILD.bazel
```diff
@@ -25,6 +26,19 @@ cc_binary(
...
+    name = "liagentd_image",
+    srcs = ["main.cpp"],
+    deps = [
+        ":interface_monitor",
+        "//orc8r/gateway/c/common/config:mconfig_loader",
+        "//orc8r/gateway/c/common/logging",
+        "//orc8r/gateway/c/common/sentry:sentry_wrapper",
+        "//orc8r/gateway/c/common/service303",
+    ],
+)
```
jheidbrink, line 79: The arguments to `cc_image` are the same as for the above `cc_binary` target.

lte/gateway/python/magma/enodebd/BUILD.bazel
```diff
@@ -39,6 +44,34 @@ py_binary(
...
+    main = "main.py",
+    visibility = ["//visibility:public"],
+    deps = [
+        ":enodebd_lib",
+        "//lte/gateway/python/magma/enodebd/tr069:server",
+        "//lte/protos:mconfigs_python_proto",
+        "//orc8r/gateway/python/magma/common:sentry",
+        "//orc8r/gateway/python/magma/common:service",
+    ],
+)
```
jheidbrink, line 91: The arguments are the same as for the `py_binary` above, but without `python_version` because `py3_image` already sets that.

lte/gateway/c/li_agent/src/BUILD.bazel
```diff
@@ -25,6 +30,36 @@ cc_binary(
     ],
 )
 
+download_pkgs(
+    name = "liagent_requisites",
+    image_tar = "@ubuntu_focal//image",
+    packages = [
```
jheidbrink, line 39: The required dependencies can by found with `ldd` or one-by-one by simply executing the binary.

WORKSPACE.bazel
```diff
@@ -162,3 +162,52 @@ http_archive(
...
+)
+
+_cc_image_repos()
+
+load(
+    "@io_bazel_rules_docker//container:container.bzl",
+    "container_pull",
+)
+
+container_pull(
```
nstng, line 199: Are the two container pulls happen here for every build, e.g., executing a python unit test? Have you tried to add these two targets in a BUILD.bazel file?

bazel/BUILD.bazel
```diff
@@ -8,7 +8,3 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
nstng, line 11: why was it possible to remove all the python toolchain configurations?

lte/gateway/python/BUILD.bazel
```diff
@@ -8,3 +8,46 @@
...
+)
+
+download_pkgs(
+    name = "python_interpreters_packages",
+    packages = [
+        "python",  # Python 2 seems to be required during the build
+        "python3",
+        "python3-distutils",  # make strtobool from deprecated distutils work
+        ],
+    image_tar = select({
```
nstng, line 37: why is this also needed in install_pkgs?

WORKSPACE.bazel
nstng: discussed offline - needed in WORKSPACE

bazel/BUILD.bazel
jheidbrink: We don't seem to require the toolchain configurations. The `py_binary` targets will generated an entrypoint that finds a Python interpreter on the system it is executed on. The toolchain required during the build is probably similarly autodetected.

lte/gateway/python/BUILD.bazel
jheidbrink: Both `download_pkgs` and `install_pkgs` execute apt-get inside a docker container. The output of `download_pkgs` is only a `.tar` file containing the downloaded packages, which is passed to `install_pkgs`. But that `.tar` containing the packages file has no information about the docker image, so that needs to be passed to `install_pkgs` as well.

MoritzThomasHuebner:
Closing because this will obviously never be merged

PR #13580, opened 2022-08-11
fix(ci): allow backports to v1.8 branch

Neudrino:
## Summary

- branch `v1.8` needs to be made available for backports to work
- additionally added a manual trigger, such that one has not to wait for the next push on master.

## Test Plan

- Possible only after merge for items from [the respective list](https://github.com/magma/magma/issues?q=label%3Aapply-v1.8+is%3Aclosed).

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


maxhbr (approved):
LGTM

PR #13581, opened 2022-08-11
refactor(enodebd): package code for enodebd devices

hejjestemjarek:

Some of devices files in enodebd are pretty large, which makes them difficult to work with and understand what is going on. Wanted to propose an idea to make it a bit easier. Could you take a look and tell what do you think about it?

As an example I refactored 2 biggest files `baicells_qrtb.py` (1100+ lines) and `freedomfi_one.py` (1600+ lines) into separate packages with new files that group the code by its type/logic/purpose. That is:
* `data_model.py` with device's implementation of `DataModel` class
* `impl.py` with device's implementation of `Handler` and `ConfigurationInitializer`, as in `state_machines.enb_acs_impl.py`
* `params.py` with device's additional parameters
* `states.py` with device's custom states, as in `state_machines.enb_acs_states.py`

## Summary
* refactor `freedomfi_one.py` file into a new, separate package
* refactor `baicells_qrtb.py` file into a new, separate package

## Test Plan
* no changes to code's logic
* all exists test cases pass

## Additional Information


LKreutzer (approved):
Bazel changes lgtm, thanks for the refactoring!

ajahl (approved):

jkmar (approved):

PR #13600, opened 2022-08-12
fix(feg): fix snowflake issue in feg integ test

### LINKED ISSUE #13540: Fix snowflake issue in FeG integ test
## Tasks:
- [x] Find out how the snowflake folder in `magma/.cache/feg` is created and change it to a file
- [ ] Prepare internal review

When following the instructions here https://docs.magmacore.org/docs/next/feg/s1ap_federated_test#s1ap-federated-integration-test, problems are encountered to do with a snowflake file. In particular, the problem is encountered with the following step of the manual build: `fab --fabfile=dev_tools.py register_feg_gw`. The problem is due to `magma/.cache/feg/snowflake` being a directory rather than a file. It can be fixed locally by deleting the directory and replacing it with a file of the same name containing a v1 UUID key, which can be generated e.g. here [https://www.uuidgenerator.net/.](https://www.uuidgenerator.net/)

The task is to find a general fix to this problem by finding the step that generates the snowflake directory, and then replacing the directory generation with the desired file generation.

Hint:
- A possible culprit is a docker volume mount. Look for places where the path to the file is mounted.
- Also, why is it checked in on github? https://github.com/magma/magma/blob/master/lte/gateway/python/.cache/feg/snowflake

wolfseb:
## Summary

fixes #13540 

Merge after #13781 

This PR creates a file `run.py`, which is to be used instead of `docker-compose up -d` to start the FeG containers. It creates the snowflake file in `/magma/.cache/feg/` if it does not exist already (if there is a directory `snowflake` in its place instead, it removes it first). Further related changes in this PR:
- adapt the automatic and semi-automatic build processes to use `./run.py`
- change the documentation of the manual build process accordingly
- remove a superfluous creation of a snowflake file in the wrong directory in the ci workflow
- remove an unused, empty snowflake file in `lte/gateway/python/.cache/feg/`

## Test Plan

- Spin up the AGW, FeG, and orc8r and run the FeG integ tests and the connectivity test (automatic, semiautomatic, and manual should all work).
- Check that if there is a folder `magma/.cache/feg/snowflake/`, the script will now automatically replace it with a snowflake file, which gets filled with an UUID.

## Additional Information

### DIFF
--- a/lte/gateway/python/integ_tests/federated_tests/docker/run.py
+++ b/lte/gateway/python/integ_tests/federated_tests/docker/run.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3
+
+"""
+Copyright 2022 The Magma Authors.
+
+This source code is licensed under the BSD-style license found in the
+LICENSE file in the root directory of this source tree.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import argparse
+import os
+import subprocess
+import sys
+
+MAGMA_ROOT = os.environ["MAGMA_ROOT"]
+SNOWFLAKE_PATH = MAGMA_ROOT + '/.cache/feg/'
+SNOWFLAKE_FILE = MAGMA_ROOT + '/.cache/feg/snowflake'
+
+
+def main() -> None:
+    """ create a snowflake file if necessary, then start docker containers """
+    args = _parse_args()
+    if not os.path.isfile(SNOWFLAKE_FILE):
+        _create_snowflake_file()
+    _exec_docker_cmd(args)
+
+
+def _create_snowflake_file() -> None:
+    if os.path.isdir(SNOWFLAKE_FILE):
+        _exec_cmd(['rm', '-r', SNOWFLAKE_FILE])
+    print("Creating snowflake file")
+    _exec_cmd(['mkdir', '-p', SNOWFLAKE_PATH])
+    _exec_cmd(['touch', SNOWFLAKE_FILE])
+
+
+def _exec_docker_cmd(args) -> None:
+    cmd = ['docker-compose', 'up', '-d']
+    if args.down:
+        cmd = ['docker-compose', 'down']
+    print(f"Running {' '.join(cmd)}...")
+    _exec_cmd(cmd)
+
+
+def _exec_cmd(cmd) -> None:
+    try:
+        subprocess.run(cmd, check=True)
+    except subprocess.CalledProcessError as err:
+        sys.exit(err.returncode)
+
+
+def _parse_args() -> argparse.Namespace:
+    """ Parse the command line args """
+    parser = argparse.ArgumentParser(description='FeG run tool')
+
+    # Other actions
+    parser.add_argument(
+        '--down', '-d',
+        action='store_true',
+        help='Stop running containers',
+    )
+    args = parser.parse_args()
+    return args
+
+
+if __name__ == '__main__':
+    main()
--- a/.github/workflows/federated-integ-test.yml
+++ b/.github/workflows/federated-integ-test.yml
@@ -99,7 +99,7 @@ jobs:
       - name: Vagrant Host prerequisites for federated integ test
         run: |
           cd ${{ env.AGW_ROOT }} && fab open_orc8r_port_in_vagrant
-          cd ${{ env.MAGMA_ROOT }} && mkdir -p .cache/test_certs/ && mkdir -p .cache/feg/ && touch snowflake
+          cd ${{ env.MAGMA_ROOT }} && mkdir -p .cache/test_certs/ && mkdir -p .cache/feg/
           cd ${{ env.MAGMA_ROOT }}/.cache/feg/ && touch snowflake
       - name: Open up network interfaces for VM
         run: |
--- a/docs/readmes/feg/s1ap_federated_tests.md
+++ b/docs/readmes/feg/s1ap_federated_tests.md
@@ -118,9 +118,9 @@ You can then [run the tests manually](#run-tests-manually).
 
 If you want to build the environment manually, you can carry out the following steps.
 
-*Note that commands for the AGW have to be run inside the Vagrant VM. For this reason,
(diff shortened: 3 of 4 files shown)


lte/gateway/python/integ_tests/federated_tests/docker/run.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
```
maxhbr, line 1: not sure how this python script compares against the following lines of bash. 

```bash
#!/usr/bin/env bash

SNOWFLAKE_FILE='$MAGMA_ROOT/.cache/feg/snowflake'
SNOWFLAKE_DIR="$(dirname "$SNOWFLAKE_FILE")"
rm -rf "$SNOWFLAKE_DIR"
mkdir -p "$SNOWFLAKE_DIR"
touch "$SNOWFLAKE_FILE"

if [[ $# -eq 0 ]] ; then
  docker-compose up -d
else
  docker-compose "$@"
fi
```

wolfseb: The only logical difference would be that the bash script deletes the `.cache/feg` directory by default and then creates the path and snowflake file, whereas the python script only does anything regarding the snowflake file if it doesn't exist in the correct way yet. But since it is just a cache it probably doesn't matter in the end, unless there is something else in the `feg` folder there (which, to my knowledge, there is not so far).
I made a python script to keep it consistent with how the containers are started for the orc8r (it just seemed convenient to have the same commands to do the same things for different parts of the project).

wolfseb: How about a slightly adapted version:
```bash
#!/usr/bin/env bash

SNOWFLAKE_DIR="$MAGMA_ROOT/.cache/feg"
SNOWFLAKE_FILE="$SNOWFLAKE_DIR/snowflake"

if [[ ! -f "$SNOWFLAKE_FILE" ]] ; then
  if [[ -d "$SNOWFLAKE_FILE" ]] ; then
    rm -rf "$SNOWFLAKE_FILE"
  fi
  mkdir -p "$SNOWFLAKE_DIR"
  touch "$SNOWFLAKE_FILE"
fi

if [[ $# -eq 0 ]] ; then
  docker-compose up -d
else
  docker-compose "$@"
fi
```

lte/gateway/python/integ_tests/federated_tests/docker/run.py
```diff
@@ -0,0 +1,69 @@
...
+        _create_snowflake_file()
+    _exec_docker_cmd(args)
+
+
+def _create_snowflake_file() -> None:
+    if os.path.isdir(SNOWFLAKE_FILE):
+        _exec_cmd(['rm', '-r', SNOWFLAKE_FILE])
+    print("Creating snowflake file")
+    _exec_cmd(['mkdir', '-p', SNOWFLAKE_PATH])
+    _exec_cmd(['touch', SNOWFLAKE_FILE])
```
voisey, line 36: At what point does the file get filled with a UUID?

voisey, line 39: At what point does the file get filled with a UUID?

wolfseb: To my understanding that is done by the magmad services on the feg via the snowflake.py package, i.e. not anywhere in the touched code. 
You can check that by e.g. replacing the snowflake file with a directory and checking the logs of the magmad container on the feg. It will throw an error:
```
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.8/dist-packages/magma/magmad/main.py", line 316, in <module>
    main()
  File "/usr/local/lib/python3.8/dist-packages/magma/magmad/main.py", line 61, in main
    logging.info('Starting magmad for UUID: %s', snowflake.make_snowflake())
  File "/usr/local/lib/python3.8/dist-packages/snowflake.py", line 43, in make_snowflake
    _write_new_id(snowflake_file)
  File "/usr/local/lib/python3.8/dist-packages/snowflake.py", line 33, in _write_new_id
    f = open(snowflake_file, "w")
IsADirectoryError: [Errno 21] Is a directory: '/etc/snowflake'
```

sebathomas: Yes, it's filled by the Python library if there's no folder there.

lte/gateway/python/integ_tests/federated_tests/docker/run.py
sebathomas: Both Python or Bash would be fine with me. In this case, the Bash version seems a little easier. On the other hand, we have a `run.py` file for other components too.

lte/gateway/python/integ_tests/federated_tests/docker/run.py
voisey: Great, thanks!

voisey (approved):

lte/gateway/python/integ_tests/federated_tests/docker/run.py
wolfseb: After discussion we decided to stick with the python script, to stay consistent with different parts of the project.

lte/gateway/python/integ_tests/federated_tests/docker/run.py
```diff
@@ -0,0 +1,69 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import argparse
+import os
+import subprocess
+
+SNOWFLAKE_PATH = '../../../../../../.cache/feg/'
+SNOWFLAKE_FILE = '../../../../../../.cache/feg/snowflake'
```
ajahl: Can we replace  `../../../../../..` with MAGMA_ROOT?
Something like `os.environ["MAGMA_ROOT"]`

wolfseb: Good idea, I changed the code accordingly.

ajahl (approved):

jheidbrink (approved):

voisey:
Could someone from @magma/approvers-agw-integtests please take a look at this? It fixes the local running ot the FeG integ tests :)

ulaskozat (approved):
LGTM

PR #13602, opened 2022-08-12
fix(ci): token needs to be specified during clone

Neudrino:
... to allow for push access.

:warning: This would only be a workaround. I hope that Github will mask the token in the log output.

It would be much cleaner, to use a proper Github action, e.g. https://github.com/tibdex/backport.
I'd personally prefer the cleaner solution over the overly complicated current implementation.


Neudrino:
Closing in favour of https://github.com/magma/magma/pull/13607.

Issue #13579, opened 2022-08-11
Required Jobs for v1.* branches are outdated and not in sync with master


maxhbr:
There are currently the following jobs required:
![2022-08-11_10:50:38](https://user-images.githubusercontent.com/1187050/184098078-25ffa32c-a229-494e-b66b-7aff6d0d35aa.png)
This is out of sync and does not align with the current state of the project.

But changing this also potentially interferes with old PRs that do not yet have the necessary gh-actions files.

Neudrino:
Should `master` and `v1.X` actually be covered by the same checks or are there differences needed?
In case it is the same, why not just extend the branch protection rules for master to the respective release branches?

maxhbr:
I created a new rule that just applies to `v1.8` and overwrites `v1.*` for that. It is a plain copy of the current `master` rules
![2022-08-13_11:47:56](https://user-images.githubusercontent.com/1187050/184478573-1841badf-162f-4a7a-adc5-073518fc9e07.png)


This should mitigate the issue for now but is not a long term solution and for the next release we again have the same issue.

Issue #13280, opened 2022-07-13
Add license headers to all Github workflows


Neudrino:
### Describe the Issue

The Github workflows currently do not include any license header.

#### Expected behavior

All files including also Github workflows should include a license header.

#### Acceptance criteria

- Add a license header to all Github workflows.
- Set 2022 as year in the newly added license headers.

PR #13607, opened 2022-08-12
feat(ci): implement proper backporting workflow

Neudrino:
## Issues with the current process

The current PR backporting workflow is broken in several aspects. Especially, the current backporting workflow
- requires special credentials with push access to protected branches. This is an anti-pattern, as these credentials could be misused elsewhere.
- directly pushes on a branch, effectively circumventing any checks, which might be in place to protect that branch. I.e. a commit being good for branch `master` could completely break the branch it is backported to.
- breaches peer review for (to be stable) release branches, due to the direct push.
- is overly complex and unmaintainable.
- has security issues as it would need to expose credentials to work correctly (https://github.com/magma/magma/pull/13602)

## Proposed new process

To avoid the above described issues, the following workflow is implemented in this PR.
It uses **https://github.com/marketplace/actions/backport-action**.

Whenever a PR is merged or a already merged PR is labeled for backporting, a new PR is created based on the changes from the original PR but against the desired backporting target. This new PR behaves like any normal PR, such that all security measures are in place for it, such as CI checks and peer review. With this process the above mentioned issues are largely avoided and a higher quality of the releases can be achieved post release.

Most significantly the improvements are:
- no credentials with elevated access required
- CI checks against the changes to decrease probability of breaking backport targets

## Test Plan

- The used Github action was tested on https://github.com/Neudrino/magma/pull/10
  - :heavy_check_mark: For the case that the PR was not merged and should be skipped. https://github.com/Neudrino/magma/runs/7807431043?check_suite_focus=true
  - :heavy_check_mark: For the case that a non existing backport target is given. https://github.com/Neudrino/magma/runs/7807442976?check_suite_focus=true
  - :heavy_check_mark: For the case that a valid backport target is given. https://github.com/Neudrino/magma
/runs/7807499100?check_suite_focus=true
  - :heavy_check_mark: For the case that the backport label is added before the merge and than merged.
    - https://github.com/Neudrino/magma/pull/15
    - https://github.com/Neudrino/magma/actions/runs/2868130983
  - :heavy_check_mark: Label added to unmerged PR: https://github.com/Neudrino/magma/actions/runs/2868126604 :arrow_right: skipped.
  - :heavy_check_mark: For the case of no backport label. :arrow_right: Workflow skipped on merge.
    - https://github.com/Neudrino/magma/pull/16
    - https://github.com/Neudrino/magma/actions/runs/2868097753
  - :heavy_check_mark: Label added after PR was merged:
    - https://github.com/Neudrino/magma/pull/16
    - https://github.com/Neudrino/magma/actions/runs/2868143268
  - :heavy_check_mark: The opened PR has the correct targets and commits. https://github.com/Neudrino/magma/pull/11

## Additional Information

- That the branch protection rules for the respective branches need to be sensible, is out of scope of this PR. https://github.com/magma/magma/issues/13579
- A well defined CI for the backport target branches is out of scope of this PR.
- The `auto_backport_label_prefix` needed to be inlined, as the [`if` context has not access to the global `env` context](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability). https://github.com/actions/runner/issues/1189


.github/workflows/backport-pull-request.yml
```diff
@@ -0,0 +1,39 @@
...
+      && (
+        github.event.action == 'closed'
+        || (
+          github.event.action == 'labeled'
+          && contains(github.event.label.name, env.backport-label)
+        )
+      )
+    steps:
+      - name: Backport Action
+        uses: sqren/backport-github-actionf54e19901f2a57f8b82360f2490d47ee82ec82c6 # pin@v8.9.3
```
maxhbr:
```suggestion
        uses: sqren/backport-github-action@f54e19901f2a57f8b82360f2490d47ee82ec82c6 # pin@v8.9.3
```

.github/workflows/backport-pull-request.yml
```diff
@@ -0,0 +1,39 @@
...
+      - labeled
+
+jobs:
+  backport:
+    name: Backport
+    runs-on: ubuntu-latest
+    # Only react to merged PRs for security reasons.
+    # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target.
+    env:
+      backport-label: apply-
```
maxhbr: better name for that variable would be `auto_backport_label_prefix`

And the label that flags a PR as "should be backported" needs to be something different.

.github/workflows/backport-pull-request.yml
```diff
@@ -0,0 +1,39 @@
...
+    if: >
+      github.event.pull_request.merged
+      && (
+        github.event.action == 'closed'
```
maxhbr, line 30: Isn't that true for every PR that is merged and the event.action is closed? Does that backport every merged PR? I think you missed the "and has the label `auto-backport`".

.github/workflows/backport-pull-request.yml
```diff
@@ -0,0 +1,39 @@
...
+        || (
+          github.event.action == 'labeled'
+          && contains(github.event.label.name, env.backport-label)
```
maxhbr:
```suggestion
          && github.event.label.name == 'auto-backport'
```

.github/workflows/backport-pull-request.yml
maxhbr:
```suggestion
        github.event.action == 'closed' && contains(github.event.pull_request.labels.*.name, 'auto-backport')
```

maxhbr (changes requested):
I think the logic that decides what to backport is not right. And also the usage of the `backport-label` does not look right.

.github/workflows/backport-pull-request.yml
```diff
@@ -0,0 +1,39 @@
+---
```
maxhbr, line 1: missing license header. Since it contains copied code from documentation also its header might be necessary to be added

maxhbr:
And creating a PR would require codeowner reviews again?

.github/workflows/backport-pull-request.yml
Neudrino: As for the other CI PR, as this is consistently missing for all workflows, I would rather prefer to do it in one commit for all workflow files later on.
- See https://github.com/magma/magma/issues/13280.

.github/workflows/backport-pull-request.yml
Neudrino: I chose a name, which is in particular different from the key of line 31, in order to avoid any chances to get name conflicts on the one hand and to not raise the impression that this would be the only place of usage.

I will change it to `backport_label_prefix`, such that there is still no exact match but a even more specific name.

.github/workflows/backport-pull-request.yml
Neudrino: _For reference:_
Another backporting action (which does not work for me however) with the respective filter can be found at https://github.com/tibdex/backport/blob/main/.github/workflows/backport.yml.

.github/workflows/backport-pull-request.yml
Neudrino: Adding the additional condition
```yaml
      && contains(join(github.event.issue.labels.*.name, ', '), env.backport_label_prefix)`
```
seem good enough for now, safe enough to try for me.

Neudrino:
> And creating a PR would require codeowner reviews again?

_No, not necessarily._
It is true, that one (or multiple [or no]) group(s) of people can be freely configured to be peer review responsible for any merge happening to the release branch. However, we could also call them **"Releasing team"** or **"Backport approvers"**. This group of people can be completely independent of any other approvers group. Moreover, it is also freely configurable for every release branch, in case that there are different teams for every branch.

Of course this would mean that one needs a well defined team for such a task and the release process, or - to be more specific - the branching process, must be well defined and respecting those changes when branching off. However, defining such a team or a proper release process is out of scope of this PR for me. This PR merely enables to implement such a well defined process easily.

.github/workflows/backport-pull-request.yml
```diff
@@ -0,0 +1,40 @@
...
+  # Only react to merged PRs for security reasons.
+  # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target.
+  # The label prefix needs to be inlined here due to the unavailability of `env` within the `if` block.
+  # https://github.com/actions/runner/issues/1189
+  backport:
+    name: Backport
+    runs-on: ubuntu-latest
+    if: >
+      github.event.pull_request.merged
+      && contains(join(github.event.pull_request.labels.*.name, ', '), 'apply-')
```
Neudrino, line 28: - The `auto_backport_label_prefix` needed to be inlined, as the `if` context has not access to the global `env` context. https://github.com/actions/runner/issues/1189

.github/workflows/backport-pull-request.yml
maxhbr: there is no reason to not start here. Especially since it contains code inspired by example configuration

.github/workflows/backport-pull-request.yml
```diff
@@ -0,0 +1,40 @@
...
+      && (
+        github.event.action == 'closed'
+        || (
+          github.event.action == 'labeled'
+          && contains(github.event.label.name, 'apply-')
```
maxhbr: minor improvement:
```suggestion
          && startsWith(github.event.label.name, 'apply-')
```

.github/workflows/backport-pull-request.yml
maxhbr: you could do:
```suggestion
      && (startsWith(join(github.event.pull_request.labels.*.name, ', '), 'apply-') || contains(join(github.event.pull_request.labels.*.name, ', '), ', apply-'))
```

Neudrino: I think this additional complexity is not worth the additional benefit. Especially if we have a cleanup implemented (https://github.com/magma/magma/pull/13197).

maxhbr (approved):
LGTM

PR #13613, opened 2022-08-15
refactor(orc8r): Simplify protected server logic

### LINKED ISSUE #13484: Clean up Service303 protected port usage
Tasks:
- [x] Implement the suggestions see newServiceWithOptions in `orc8r/lib/go/service/service.go`
- [x] Start k8s setup and check logfiles
- [x] Make sure the feg integ tests run
- [x] Prepare sprint review

Follow up story to #12608, #12941, and #12564

service303 methods are defined here:
`lib/go/protos/service303.pb.go`

Protected Port:
Just a separate port most likely not piped through the https proxy.

Services are registered to the service registry; to connect to a service, `registry.GetConnection` in local_registry.go is called.

Previous PRs aimed aims to move the Service303 functionality behind a protected port.

Client is currently hardcoded to use a protected port in orc8r `orc8r/lib/go/service/client/client_api.go`.

Questions:
- If Service303 is behind a protected port, should the service not also be moved behind a protected port? probably not because monitoring of the service could be internal and the rest of the service could still be public?
- Can we disable the non protected port if there is a protected port defined for the entire service? Probably not see above.
- Can we only move service303 to the protected port if a protected port is defined for that service? why not?

Suggestion:
- Protected&public port defined -> run 303 on protected port
- Only public port defined -> run 303 on public port
- Only protected port defined -> run 303 on protected port
- Whether the service is protected should be part of the config map in `orc8r/lib/go/service/service.go`

MoritzThomasHuebner:
Fixes #13484

Signed-off-by: Moritz Huebner <moritz.huebner@tngtech.com>

## Summary

- Consolidated `NewGatewayServiceWithOptions` and `NewOrc8rServiceWithOptions` back into `NewServiceWithOptions`
- Eliminated `protected303Server` argument in `NewServiceWithOptions`
- Service type is now identified by the available ports, using `registry.GetServicePort`
  - `registry.GetServicePort` returns `0` if and only if not protected port exists
  
## Test 

- [x] Manually check that services get the correct port:
  - [x] cwf
  - [x] feg
  - [x] orc8r (using minikube)
- [x] Run federated integ tests.

## Additional Information

### DIFF
--- a/cwf/gateway/services/gateway_health/gateway_health/main.go
+++ b/cwf/gateway/services/gateway_health/gateway_health/main.go
@@ -45,7 +45,7 @@ const (
 
 func main() {
 	// Create the service
-	srv, err := service.NewGatewayServiceWithOptions(registry.ModuleName, registry.GatewayHealth)
+	srv, err := service.NewServiceWithOptions(registry.ModuleName, registry.GatewayHealth)
 	if err != nil {
 		glog.Fatalf("Error creating %s service: %s", registry.GatewayHealth, err)
 	}
--- a/cwf/gateway/services/uesim/uesim/main.go
+++ b/cwf/gateway/services/uesim/uesim/main.go
@@ -48,7 +48,7 @@ func main() {
 	flag.Parse()
 	glog.Info("Starting UESim service")
 
-	srv, err := service.NewGatewayServiceWithOptions(registry.ModuleName, registry.UeSim)
+	srv, err := service.NewServiceWithOptions(registry.ModuleName, registry.UeSim)
 	if err != nil {
 		glog.Fatalf("Error creating UeSim service: %s", err)
 	}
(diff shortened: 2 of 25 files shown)


orc8r/gateway/go/services/magmad/main.go
```diff
@@ -65,6 +65,8 @@ var (
 )
 
 func main() {
+	err := service.StartMagmadServer()
+	fmt.Print(err)
```
sebathomas: This seems like leftover debug code.

sebathomas (changes requested):
The cleanup looks good, just take a look at the main.go changes.

orc8r/gateway/go/services/magmad/main.go
MoritzThomasHuebner: Weird, I have no recollection of writing and committing this code. I removed it anyway.

sebathomas (approved):

crasu (approved):

PR #13621, opened 2022-08-16
refactor: Remove xwf from pipelined

alexzurbonsen:
## Summary

This PR relates to https://github.com/magma/magma/issues/13546. It removes xwf functionality from pipelined.

## Test Plan

- CI
- Integ Tests (https://github.com/alexzurbonsen/magma/actions/runs/2876430498)

## Additional Information


lte/gateway/python/magma/pipelined/app/egress.py
```diff
@@ -140,10 +140,6 @@ def _get_default_flow_msgs(self, datapath) -> DefaultMsgsMap:
     def initialize_on_connect(self, datapath):
         self._datapath = datapath
         self._setup_non_nat_monitoring()
-        # TODO possibly investigate stateless XWF(no sessiond)
-        if self.config.setup_type == 'XWF':
-            self.delete_all_flows(datapath)
-            self._install_default_flows(datapath)
```
nstng, line 146: is `_install_default_flows` now unused?

lte/gateway/python/magma/pipelined/app/ingress.py
```diff
@@ -276,10 +276,6 @@ def finish_init(self, _):
 
     def initialize_on_connect(self, datapath):
         self._datapath = datapath
-        # TODO possibly investigate stateless XWF(no sessiond)
-        if self.config.setup_type == 'XWF':
-            self.delete_all_flows(datapath)
-            self._install_default_flows(datapath)
```
nstng, line 282: is `_install_default_flows` now unused?

lte/gateway/python/magma/pipelined/app/middle.py
```diff
@@ -180,10 +180,6 @@ def finish_init(self, _):
 
     def initialize_on_connect(self, datapath):
         self._datapath = datapath
-        # TODO possibly investigate stateless XWF(no sessiond)
-        if self.config.setup_type == 'XWF':
-            self.delete_all_flows(datapath)
-            self._install_default_flows(datapath)
```
nstng, line 186: is `_install_default_flows` now unused?

nstng (changes requested):

nstng:
`lte/gateway/python/magma/pipelined/tests/snapshots/test_inout.InOutTestXWF.testFlowSnapshotMatch.snapshot` should be obsolete now.

lte/gateway/python/magma/pipelined/app/egress.py
alexzurbonsen: Yes, your are right. There are no callers within the class anymore.

lte/gateway/python/magma/pipelined/app/ingress.py
alexzurbonsen: same as above

lte/gateway/python/magma/pipelined/app/middle.py
alexzurbonsen: same as above

alexzurbonsen:
> `lte/gateway/python/magma/pipelined/tests/snapshots/test_inout.InOutTestXWF.testFlowSnapshotMatch.snapshot` should be obsolete now.

Deleted it.

alexzurbonsen:
A little explanation for egress.py, ingress.py and middle.py:

Since the `_install_default_flows` method is deleted in all three files and `_get_default_flow_msgs` and `_wait_for_responses` do not have any other callers (exception: in egress.py's `EgressController` the method `_wait_for_responses` is still used).

Thus, one might think they should be deleted. `_get_default_flow_msgs` implements an abstract method from `RestartMixin`. One thus might prefer an implementation with empty return value.
But both methods are actually used in `handle_restart` in `RestartMixin`, even though `_wait_for_responses` is not defined as an abstract method.
Since `handle_restart` may be called from some outside caller, it is probably not sensible to mess with it.
  
Accordingly, `_get_default_flow_msgs` and `_wait_for_responses` are not deleted here.

nstng (approved):
lgtm - run sudo tests locally (/)

PR #13623, opened 2022-08-16
chore: new integ tests are bazelified

### LINKED ISSUE #13588: Remove S1AP tests from python script list for bazelification check
Most of the S1AP tests were bazelified, but there are still some remaining (which are currently not executed anywhere). The deny list in `bazel/scripts/check_py_bazel.sh` should be updated accordingly.

nstng:
## Summary

Main target here was to do #13588. But it came up that one python file - `gpp_types.py` - was not yet bazelified (this is used in integ tests, but somehow bazel has access on magma_test) and that there are three new integ tests that need to be bazelified.

~~Note that two of the integ tests seem to be setup and tear down "tests" for not yet activated ipv6 tests. But imho it is a good idea to bazelify them for now so that this logic is tested as well.~~ The basilification of those tests was removed after a review discussion. This will be re-added when respective tests for these setup and teardown "tests" are added. Or they are included in the respective tests directly.

### new tests
* `./bazel/scripts/run_integ_tests.sh lte/gateway/python/integ_tests/s1aptests:test_attach_detach_flaky_retry_success`

### `gpp_types`
On a vm `!= magma_test` check if `bazel run` produces ModuleNotFound exceptions (the tests will fail! but the ModuleNotFound exception should come before the test failure). E.g., use
```
#!/usr/bin/env bash

LOGGING_FILE="/tmp/test_import_logging.log"

A=$(bazel query "kind(py_test, //lte/gateway/python/integ_tests/s1aptests/...)")
for test in $A
do
  echo "CHECKING ${test}"
  bazel run "${test}" --define=on_magma_test=1 2>&1  | tee "${LOGGING_FILE}"
  RESULT=$(grep -m 1 "ModuleNotFoundError" "${LOGGING_FILE}" || [[ $? == 1 ]])
  if [[ "${RESULT}" != "" ]];
  then
    exit 1
  fi
done
```

### updated py files check
In `bazel/scripts/check_py_bazel.sh` remove line 56-81and check that the output corresponds to the removed lines. Note that some files are covered by excluding complete directories. 

## Additional Information

### DIFF
--- a/bazel/scripts/check_py_bazel.sh
+++ b/bazel/scripts/check_py_bazel.sh
@@ -33,6 +33,7 @@ DENY_LIST_NOT_RELEVANT=(
   "./lte/gateway/deploy"
   # manually executed script
   "./lte/gateway/c/core/oai/tasks/s1ap/messages/asn1/asn1tostruct.py"
+  "./lte/gateway/python/build"
   # used for manual testing
   "./lte/gateway/python/magma/pipelined/tests/envoy-tests/http-serve.py"
   "./lte/gateway/python/magma/tests/pylint_wrapper.py"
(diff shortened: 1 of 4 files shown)


bazel/scripts/check_py_bazel.sh
```diff
@@ -46,13 +46,39 @@ DENY_LIST_NOT_RELEVANT=(
   "./show-tech"
   "./third_party"
   "./hil_testing"
+  "./.git"
```
vktng: Why was this needed?

lte/gateway/python/integ_tests/s1aptests/BUILD.bazel
```diff
@@ -47,12 +48,16 @@ py_library(
     ],
 )
 
+py_library(
+    name = "gpp_types",
```
vktng, line 52: Checked that all tests that import `gpp_types` are included in this PR.

lte/gateway/python/integ_tests/s1aptests/BUILD.bazel
```diff
@@ -1119,6 +1181,24 @@ pytest_test(
     deps = [":s1ap_wrapper"],
 )
 
+pytest_test(
+    name = "test_enable_ipv6_iface",
+    size = "small",
+    srcs = ["test_enable_ipv6_iface.py"],
+    imports = [LTE_ROOT],
+    tags = TAG_PRECOMMIT_TEST,
```
vktng: We should add new tags to the file `bazel/test_constants.bzl`:
```
TAG_IPV6_TEST_ENABLE = ["ipv6_enable"] + TAG_MANUAL
TAG_IPV6_TEST_DISABLE = ["ipv6_disable"] + TAG_MANUAL
```
And tag the tests with these labels.
Even if it does not seem to break something, we should not run them out of order with the other precommit tests.

vktng (changes requested):

bazel/scripts/check_py_bazel.sh
nstng: sometimes there .py files in the local git cache - its should only be relevant for local execution. (just solves anyoing false positives)

lte/gateway/python/integ_tests/s1aptests/BUILD.bazel
nstng: discussed offline - remove from BUILD.bazel and add to check again.
* these "tests" are for setup and do not test something currently
* we will ask the owner if this can be included in specific setup/teardown steps in tests (the two steps seem to be sufficiently fast)

vktng (approved):
LGTM

LKreutzer (approved):

PR #13625, opened 2022-08-16
fix(agw): Remove unused python symlink

sebathomas:
## Summary

The python symlink had a typo. Since it didn't break anything, we can just remove it and rely on calling python3 explicitly.

## Test Plan

CI

## Additional Information

### DIFF
--- a/lte/gateway/docker/services/python/Dockerfile
+++ b/lte/gateway/docker/services/python/Dockerfile
@@ -57,8 +57,6 @@ RUN apt-get update && apt-get install -y \
   libsystemd-dev \
   libprotobuf-dev
 
-
-RUN cd /usr/local/bin && ln -s /usr/bin/python3 python
 RUN gem install fpm
 
 COPY . $MAGMA_ROOT/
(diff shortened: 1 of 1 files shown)


lte/gateway/docker/services/python/Dockerfile
```diff
@@ -43,6 +43,7 @@ RUN apt-get update && apt-get install -y \
   sudo \
   ruby-dev \
   docker.io \
+  python-is-python3 \
```
jheidbrink: do we actually need this in the containerized agw? The python invocations I'm aware of mention python3 explicitly.

jheidbrink (approved):

lte/gateway/docker/services/python/Dockerfile
sebathomas: We probably don't need it actually. From what I saw, Python 2 isn't installed and there is no `python` symlink without a typo at the moment.

I was trying to preserve the intended behavior but I could also remove it, what do you think?

sebathomas: I only looked at the "dev/production" image, not the "builder" image though.

jheidbrink: I don't have a strong opinion, but I'm slightly leaning towards not having a `python` binary and encouraging people to be explicit when invoking python. That is, as long as that's feasible. In this case I guess we don't need the symlink, given that the old symlink was broken anyway and the images still seemed to work.

sebathomas: Done.

maxhbr (commented):
LGTM

maxhbr (approved):

PR #13628, opened 2022-08-16
chore: Improve integ test wrapper script

### LINKED ISSUE #13525: Add improvements to the bazel/scripts/run_integ_tests.sh script
Two new features were proposed to make the use of the script more convenient:

1. Add option to rerun failed test a number of times (default 2?) - default: no rerun, else: provide number how often
2. optional: Make it possible to specify 2 or more targets from any category, also mixed (one from precommit one from extended etc.)

Hints:
* use pytest parameters for flaky rerun in python_test.bzl - need to be arguments so that it can be used for CI only
* `lte/gateway/python/integ_tests/Makefile:28`
* if not feasibly possible via bazel arguments, think about using an environment variable

vktng:
## Summary

The integration test wrapper script for Bazel was lacking some desired features, this PR aims to resolve 4 of them:

1. The option `--retry-on-failure` and `--retry-attempts N` was added. With these you can rerun failing tests N times and they will only be considered failed if all attempts fail.
2. Multiple test targets can be specified. You can provide your own list of targets that can contain tests from all categories and the order does not matter.
3. After every run there will be a Junit XML report generated.
4. With the `--rerun-previously-failed` option you can rerun all previously failed tests. 

## Test Plan

Use options as described in `bazel/scripts/run_integ_tests.sh  --help`.

Some examples:

- Quick, only precommit tests:
```bash
bazel/scripts/run_integ_tests.sh  //lte/gateway/python/integ_tests/s1aptests:test_attach_detach //lte/gateway/python/integ_tests/s1aptests:test_attach_detach_static_ip //lte/gateway/python/integ_tests/s1aptests:test_gateway_metrics_attach_detach //lte/gateway/python/integ_tests/s1aptests:test_attach_without_ips_available
```
- More flaky and longer nonsanity tests:
```bash
bazel/scripts/run_integ_tests.sh  //lte/gateway/python/integ_tests/s1aptests:test_outoforder_erab_setup_rsp_dedicated_bearer  //lte/gateway/python/integ_tests/s1aptests:test_attach_detach_multi_ue_looped //lte/gateway/python/integ_tests/s1aptests:test_attach_detach_nw_triggered_delete_secondary_pdn //lte/gateway/python/integ_tests/s1aptests:test_attach_detach_ipv6
```
- Retry:
```bash
bazel/scripts/run_integ_tests.sh --retry-on-failure //lte/gateway/python/integ_tests/s1aptests:test_outoforder_erab_setup_rsp_dedicated_bearer --retry-attempts 1 //lte/gateway/python/integ_tests/s1aptests:test_attach_detach_multi_ue_looped //lte/gateway/python/integ_tests/s1aptests:test_attach_detach_nw_triggered_delete_secondary_pdn //lte/gateway/python/integ_tests/s1aptests:test_attach_detach_ipv6
```
- Rerun:
```bash
bazel/scripts/run_integ_tests.sh  --rerun-previously-failed
```

## Additional Information

### DIFF
--- a/lte/gateway/python/integ_tests/s1aptests/BUILD.bazel
+++ b/lte/gateway/python/integ_tests/s1aptests/BUILD.bazel
@@ -9,7 +9,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("@python_deps//:requirements.bzl", "requirement")
 load("@rules_python//python:defs.bzl", "py_library")
 load("//bazel:python_test.bzl", "pytest_test")
 load(
@@ -1199,10 +1198,7 @@ pytest_test(
     srcs = ["test_attach_detach_flaky_retry_success.py"],
     imports = [LTE_ROOT],
     tags = TAG_EXTENDED_TEST,
-    deps = [
-        ":s1ap_wrapper",
-        requirement("flaky"),
-    ],
+    deps = [":s1ap_wrapper"],
 )
 
 pytest_test(
--- a/lte/gateway/python/scripts/runtime_report.py
+++ b/lte/gateway/python/scripts/runtime_report.py
@@ -46,12 +46,16 @@ def merge_all_report(working_dir, list_xml_report_paths, output_path):
         xml_file_path = working_dir + "/" + xml_file_path
         test_result = ET.parse(xml_file_path)
         test_suites_data = test_result.getroot()
-
+        # for the integration tests pytest generates the XML files with a slightly different structure
+        integration_tests = test_suites_data.attrib == {}
+        if integration_tests:
+            test_suites_data.attrib = test_suites_data[0].attrib
         num_all_failures += int(test_suites_data.attrib['failures'])
         num_all_tests += int(test_suites_data.attrib['tests'])
         total_time += float(test_suites_data.attrib['time'])
         num_all_errors += int(test_suites_data.attrib['errors'])
-        num_all_disabled += int(test_suites_data.attrib['disabled'])
+        if not integration_tests:
+            num_all_disabled += int(test_suites_data.attrib['disabled'])
         init_time = min(init_time, datetime.fromisoformat(test_suites_data.attrib['timestamp']))
 
         for single_test_suite_data in test_suites_data:
--- a/bazel/external/requirements.in
+++ b/bazel/external/requirements.in
@@ -65,8 +65,8 @@ fakeredis==1.7.1
 lupa==1.10
 parameterized==0.8.1
 freezegun
-pytest
-pytest-cov
+pytest==7.1.2
+pytest-cov==3.0.0
 coverage-lcov
 
 # s1ap test requirements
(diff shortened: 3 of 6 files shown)


nstng (changes requested):
As discussed offline:
* `$ ./bazel/scripts/run_integ_tests.sh --retry-on-failure` results in `./bazel/scripts/run_integ_tests.sh: line 111: POSITIONAL_ARGS: unbound variable`
* `report/merged_report/report_all_tests.xml` should not be localted under magma root, e.g., use something related to `RESULTS_DIR := /var/tmp/test_results` in make integ tests
-> in the repo is not good
* fix result xml - make it publishable

LKreutzer (approved):

vktng:
> As discussed offline:
> 
>     * `$ ./bazel/scripts/run_integ_tests.sh --retry-on-failure` results in `./bazel/scripts/run_integ_tests.sh: line 111: POSITIONAL_ARGS: unbound variable`
> 
>     * `report/merged_report/report_all_tests.xml` should not be localted under magma root, e.g., use something related to `RESULTS_DIR := /var/tmp/test_results` in make integ tests
>       -> in the repo is not good
> 
>     * fix result xml - make it publishable

Fixed.

lte/gateway/python/scripts/runtime_report.py
```diff
@@ -46,12 +46,15 @@ def merge_all_report(working_dir, list_xml_report_paths, output_path):
         xml_file_path = working_dir + "/" + xml_file_path
         test_result = ET.parse(xml_file_path)
         test_suites_data = test_result.getroot()
-
+        integration_tests = test_suites_data.attrib == {}
```
sebathomas, line 50: I'm a bit confused by the these changes. So if we're running integration tests, there's an empty dict here and the real data is in the first element? Maybe it would be good to add a comment.

vktng: When the xml is created with pytest then the structure is slightly different from what this script was intended to work with. If the attributes are missing, then we have to copy them from the child node.

sebathomas: I see. :+1:

By the way, there's an unused variable in this function called `c` that you could remove if you like.

sebathomas (approved):
AGW changes looking good.

lte/gateway/python/integ_tests/s1aptests/BUILD.bazel
```diff
@@ -1199,10 +1198,7 @@ pytest_test(
     srcs = ["test_attach_detach_flaky_retry_success.py"],
     imports = [LTE_ROOT],
     tags = TAG_EXTENDED_TEST,
-    deps = [
-        ":s1ap_wrapper",
-        requirement("flaky"),
```
vktng, line 1204: Since "flaky" is a dependency of pytest from now on you do not need to specify it. (It will lead to a duplication error.)
See `bazel/python_test.bzl`

sebathomas:
Needs to be rebased to get the fix for the DP CI issue from master.

nstng (approved):

PR #13636, opened 2022-08-17
chore(dp): optimizing queries

WojSad:
## Summary

Removed unnecessary selects, introduced the USING clause to deletes, subqueries, changed logic behind some of the updates. Removed logs that were making db calls.
Added indexes to a few columns on cbsds and channels tables
Cached frequently used entities in radio controller

## Test Plan

Migration tests included, existing unit and integration tests passing


dp/cloud/python/magma/configuration_controller/response_processor/response_db_processor.py
```diff
@@ -72,17 +73,14 @@ def process_response(self, db_requests: List[DBRequest], response: requests.Resp
             return
 
     def _populate_grant_states_map(self, session):
-        self.grant_states_map = {
-            GrantStates.GRANTED.value: session.query(DBGrantState).filter(
-                DBGrantState.name == GrantStates.GRANTED.value,
-            ).scalar(),
-            GrantStates.AUTHORIZED.value: session.query(DBGrantState).filter(
+        grant_states = session.query(DBGrantState).filter(
```
jkmar: why not just `session.query(DBGrantState).all()`?

dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
```diff
@@ -492,7 +492,7 @@ def then_state_is_eventually(self, expected, request):
     @retry(stop_max_attempt_number=30, wait_fixed=1000)
     def then_logs_are(self, filters: Dict[str, Any], expected: List[str], network: Optional[str] = None):
         actual = self._get_log_types(filters, network)
-        self.assertListEqual(expected, actual)
+        self.assertCountEqual(expected, actual)
```
jkmar, line 495: Maybe we should check if sets are equal if we don't care about the order, checking only count may be wrong for example in case there are no filters

dp/cloud/python/magma/db_service/models.py
```diff
@@ -200,12 +200,12 @@ class DBCbsd(Base):
     state_id = Column(
         Integer, ForeignKey(
             "cbsd_states.id", ondelete="CASCADE",
-        ), nullable=False,
+        ), nullable=False, index=True,
```
jkmar: Does this really help? This table has only 2 possible values

dp/cloud/python/magma/db_service/models.py
```diff
@@ -295,7 +295,7 @@ class DBChannel(Base):
     """
     __tablename__ = "channels"
     id = Column(Integer, primary_key=True, autoincrement=True)
-    cbsd_id = Column(Integer, ForeignKey("cbsds.id", ondelete="CASCADE"))
+    cbsd_id = Column(Integer, ForeignKey("cbsds.id", ondelete="CASCADE"), index=True)
```
jkmar, line 292: I think we should also do that for other tables with `ondelete="CASCADE"`, since we've already experienced problems with those (especially with requests)

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
```diff
@@ -374,11 +375,15 @@ def _terminate_all_grants_from_response(response: DBResponse, session: Session)
     ) or response.request.payload.get(CBSD_ID)
     if not cbsd_id:
         return
-    cbsd = session.query(DBCbsd).filter(DBCbsd.cbsd_id == cbsd_id).scalar()
     logger.info(f'Terminating all grants for {cbsd_id=}')
-    session.query(DBGrant).filter(DBGrant.cbsd == cbsd).delete()
-    logger.info(f"Deleting all channels for {cbsd_id=}")
-    session.query(DBChannel).filter(DBChannel.cbsd == cbsd).delete()
+    with session.no_autoflush:
```
jkmar, line 379: this looks great, can we maybe move to some top level function?

dp/cloud/python/magma/configuration_controller/response_processor/response_db_processor.py
```diff
@@ -72,17 +73,14 @@ def process_response(self, db_requests: List[DBRequest], response: requests.Resp
...
                 DBGrantState.name == GrantStates.AUTHORIZED.value,
-            ).scalar(),
-            GrantStates.UNSYNC.value: session.query(DBGrantState).filter(
+                DBGrantState.name == GrantStates.GRANTED.value,
                 DBGrantState.name == GrantStates.UNSYNC.value,
-            ).scalar(),
-        }
+            ),
+        ).all()
+        self.grant_states_map = {gs.name: gs for gs in grant_states}
```
jkmar, line 75: maybe we can also do that for other enum tables

dp/cloud/python/magma/test_runner/tests/test_dp_with_orc8r.py
WojSad: the name of the method is misleading, it actually compares the content of the containers regardless of the order

dp/cloud/python/magma/db_service/models.py
WojSad: you're right

dp/cloud/python/magma/configuration_controller/response_processor/response_db_processor.py
WojSad: true, didn't notice these were the only possible states

dp/cloud/python/magma/configuration_controller/response_processor/response_db_processor.py
WojSad: you mean grant_states, cbsd_states and request_types? they are only populated once during db initialization so I don't think it makes much difference. On the opposite, grant_states_map is populated with each requests batch coming in

dp/cloud/python/magma/configuration_controller/response_processor/strategies/response_processing.py
WojSad: leaving as is as discussed

dp/cloud/python/magma/configuration_controller/response_processor/response_db_processor.py
WojSad: did that for grant_states and request_types in radio controller

dp/cloud/python/magma/db_service/models.py
WojSad: deleted

dp/cloud/python/magma/db_service/models.py
WojSad: leaving as is as discussed

jkmar (approved):

xbend (approved):

PR #13637, opened 2022-08-17
fix(ci): refine backport title & allow dependent job exection

Neudrino:
## Summary

- The PR title for the new backporting workflow does not align with the semantic commit:
  - https://github.com/magma/magma/pull/13633
  - https://github.com/magma/magma/pull/13634
  - https://github.com/magma/magma/pull/13635

  As a new title format it is proposed to append the `[backport to v1.8]` instead of prefixing it.

## Test Plan

- The target branch is post-pended not prepended:
  - https://github.com/sqren/backport/blob/main/docs/config-file-options.md#prtitle
  - Postpending works: https://github.com/Neudrino/magma/pull/20

## Additional information

- This is a follow up to https://github.com/magma/magma/pull/13607.

### DIFF
--- a/.github/workflows/backport-pull-request.yml
+++ b/.github/workflows/backport-pull-request.yml
@@ -33,6 +33,10 @@ jobs:
         )
       )
     steps:
+      - run: |
+          echo '{
+            "prTitle": "{commitMessages} [backport to {targetBranch}]"
+          }' > .backportrc.json
       - name: Backport Action
         uses: sqren/backport-github-action@f54e19901f2a57f8b82360f2490d47ee82ec82c6 # pin@v8.9.3
         with:


.github/workflows/backport-pull-request.yml
```diff
@@ -33,6 +33,10 @@ jobs:
         )
       )
     steps:
+      - run: |
+          echo '{
+            "prTitle": "{commitMessages} [backport to {targetBranch}]"
```
jheidbrink, line 38: The linked example from the test plan has the suffix _backport backport-v1.8_ which doesn't match this pattern. Was that one created with an older version?

jheidbrink: Can we assume that `commitMessages` will always be just one message ("merge PR ...")?

.github/workflows/semantic-pr.yml
```diff
@@ -52,7 +52,7 @@ jobs:
     steps:
       - uses: amannn/action-semantic-pull-request@db6e259b93f286e3416eef27aaae88935d16cf2e # pin@v3.4.0
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jheidbrink: Why do we need a change in the semantic-PR workflow?

Neudrino: Point 2 of the description: Ability to run actions on opened PRs.

.github/workflows/backport-pull-request.yml
Neudrino: > The linked example from the test plan has the suffix _backport backport-v1.8_ which doesn't match this pattern. Was that one created with an older version?

Yes. Noticed the missing `to` after testing.

Neudrino: > Can we assume that `commitMessages` will always be just one message ("merge PR ...")?

- I cannot answer this, as I have no access to the respective repository settings.
- If the repository only allows squash merge commits, the answer is "yes".
- From my experience the repository setting is only allowing "squash merge".

.github/workflows/semantic-pr.yml
jheidbrink: From the linked docs I understand that the backport workflow needs a different token than `GITHUB_TOKEN` so that the push/PR events it creates trigger the test suite.

I don't understand why the semantic-PR workflow needs to create events that trigger workflow runs. Which events are those? And which workflows?

Neudrino: Oh sorry. You are correct, this was accidentally edited in the wrong file, due to too many long open parallel CI PRs. :innocent:

Neudrino: The required change was already made in https://github.com/magma/magma/pull/13640.

jheidbrink (approved):

PR #13638, opened 2022-08-17
fix(ci): the semantic-pr should also run on additional actions

Neudrino:
## Summary

- Semantic PR should also run on reopened PRs
- And while we are at it, running it at the transition from draft to real PR should also be included imho


maxhbr (approved):
LGTM

PR #13647, opened 2022-08-18
feat(ci): add non activated option to really use host docker socket

Neudrino:
## Summary

- The implementation of using host docker sock was not fully complete with the respect that an option was still missing.
- The option was added with a comment.
- The option was defaulted to `off` as it has security implications, but is _very_ helpful for development. Every developer should take an educated decision.

- The IntelliJ json linter complains about trailing commata. Thus I removed them.

## Test Plan

- Tested by @jheidbrink before.

## Additional Information

- Done in pairing with @jheidbrink.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -27,7 +27,7 @@
 		"stackbuild.bazel-stack-vscode-cc",
 		"augustocdias.tasks-shell-input",
 		"ryuta46.multi-command",
-		"coolchyni.beyond-debug",
+		"coolchyni.beyond-debug"
 	],
 	"initializeCommand": "docker pull ghcr.io/magma/magma/devcontainer:latest",
 	"image": "ghcr.io/magma/magma/devcontainer:latest",
@@ -44,7 +44,7 @@
 		},
 		"files.watcherExclude": {
 			"**/.bazel-cache/**": true,
-			"**/.bazel-cache-repo/**": true,
+			"**/.bazel-cache-repo/**": true
 		},
 		"bsv.bazel.buildFlags": [],
 		"bsv.bazel.testFlags": [],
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -61,14 +61,14 @@
...
-			"--background-index",
+			"--background-index"
 		],
 		"clangd.onConfigChanged": "restart",
 		"editor.formatOnSave": true,
 		// Update this field with any new Bazel targets that need compilation database generation
 		"bsv.cc.compdb.targets": [
 			"//orc8r/gateway/c/...",
-			"//lte/gateway/c/...",
+			"//lte/gateway/c/..."
```
jheidbrink, line 71: I would prefer that we are a bit more minimally-invasive and don't change formatting completely unrelated to our actual change. Especially since this style change is very subjective. Trailing commas produce nicer diffs. Since I think we don't have a style guide for this, we should simply respect the status quo and not go around changing things to our personal liking everywhere.

Neudrino: I disagree. Especially the change is not "subjective".
- The state of the file was very inconsistent before, i.e. there was no clear style at all in the file, and there would sometimes be trailing commas and sometimes not. And even if there is no conclusive definition on how it should be, the consistency of a file in itself should be of high value. A change to reach a consistent state is always an improvement and never "subjective" to me.
- While I am aware of the nicer diffs with trailing commas and am very much fan of it in most languages (e.g. Python), the [JSON specification does not allow for trailing commas](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas) (according to most references I found) such that I would see a reasonable argument for this change. And while this file might not need to follow the exact JSON specifications (comments seem to be allowed too), it is still a JSON and following the JSON specifications where possible is reason for this change.
- Other parts of the file where mis-formatted too. I.e. the indentation was wrong. To me this would be a change required by the "boy scouts mentality". If you find a mis-formatted file, you format it correctly and **consistently**.

jheidbrink (approved):

PR #13659, opened 2022-08-18
feat(agw): Run integ tests vs containerized AGW

### LINKED ISSUE #13475: Run integration test against containerized AGW
## Acceptance criteria
- [x] Deploy the containerized AGW following the [documentation](https://github.com/magma/magma/blob/master/lte/gateway/docker/README.md#running-the-containerized-agw-locally-on-the-magma-vm).
- [x] Run AGW integration tests against the containerized AGW. Get test_attach_detach to run.
- [x] (Optional) Run more tests. Ones involving the traffic server would be a bonus.
- [x] Document
  - the necessary steps to run the integration tests against containerized AGW
  - any obstacles while trying to run the tests against the containerized AGW
  - the test results
- [x] Prepare internal review meeting

## Implementation details
- Service restarts via systemd need to be adapted (?)
- Reference from the integ test run against the deb deployed vm: https://github.com/magma/magma/issues/13263 
- Start the containerized AGW on the current dev VM.

## Additional information
- https://github.com/magma/magma/blob/master/lte/gateway/docker/README.md
- https://github.com/magma/magma/pull/8771

voisey:
## Summary

Works towards getting all integration tests running versus the containerized AGW, starting with `test_attach_detach`. Modifies system restarts to account for Docker. Also modifies the readme to explain the present situation and how to run the test, as well as adding sctpd to the services that need to be stopped before starting the Docker containers.   

This aims to be a non-breaking PR which can be built upon in the future to get further tests running.

## Test Plan

At present, `test_attach_detach` runs successfully if `MAGMA_DEV_MODE` is set to 1 in the `.env` file.

Notes:
- In the future, we may want to use a Docker override file to set the required environment variable instead of modifying the `.env` file.
- Containers can be started following the instructions here: https://github.com/magma/magma/blob/master/lte/gateway/docker/README.md#running-the-containerized-agw-locally-on-the-magma-vm.

### DIFF
--- a/lte/gateway/docker/README.md
+++ b/lte/gateway/docker/README.md
@@ -40,7 +40,7 @@ containerized AGW by running the following steps inside the VM:
 ```
 cd $MAGMA_ROOT/lte/gateway && make run  # You can skip this if you have built the AGW with make before
 for component in redis nghttpx td-agent-bit; do cp "${MAGMA_ROOT}"/{orc8r,lte}/gateway/configs/templates/${component}.conf.template; done
-sudo systemctl stop 'magma@*'  # We don't want the systemd-based AGW to run when we start the containerized AGW
+sudo systemctl stop 'magma@*' 'sctpd' # We don't want the systemd-based AGW to run when we start the containerized AGW
 cd $MAGMA_ROOT/lte/gateway/docker
 docker-compose build
 docker-compose up
@@ -51,3 +51,18 @@ on a VM. However we are not there yet as the containerized AGW currently depends
 on a patched Open vSwitch installation on the host machine. The magma VM happens
 to have the right packages installed, and thus can currently be used as a quick
 and dirty way to run the containers locally.
+
+### Running the S1AP integration tests against the containerized AGW
+
+At present, of the S1AP integration tests, it has been verified that `test_attach_detach` can be run versus the containerized AGW.
+To run this test, development mode needs to be enabled when starting the containers, which can be done by using an override
+file `docker-compose.dev.yaml`:
+
+```
+cd $MAGMA_ROOT/lte/gateway/docker
+docker-compose down # If containers are already running
+docker-compose -f docker-compose.yaml -f docker-compose.dev.yaml up
+```
+
+The test VM can then be set up and the test executed by following
+[these instructions](https://docs.magmacore.org/docs/next/lte/s1ap_tests#test-vm-setup).
--- a/lte/gateway/docker/docker-compose.dev.yaml
+++ b/lte/gateway/docker/docker-compose.dev.yaml
@@ -1,0 +1,102 @@
+---
+# Copyright 2022 The Magma Authors.
+
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+version: "3.7"
+
+services:
+  magmad:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  redis:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  directoryd:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  subscriberdb:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  enodebd:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  state:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  policydb:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  health:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  monitord:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  redirectd:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  smsd:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  control_proxy:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  ctraced:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  sctpd:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  oai_mme:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  pipelined:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  sessiond:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  mobilityd:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  td-agent-bit:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  eventd:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  connectiond:
+    environment:
+      - MAGMA_DEV_MODE=1
+
+  liagentd:
+    environment:
+      - MAGMA_DEV_MODE=1
--- a/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
+++ b/lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
@@ -814,6 +814,11 @@ class SubscriberUtil(object):
         self._subscriber_client.wait_for_changes()
 
 
+class InitMode(Enum):
+    SYSTEMD = 1
+    DOCKER = 2
+
+
 class MagmadUtil(object):
     """Utility class to trigger system commands in Magma"""
 
@@ -823,6 +828,8 @@ class MagmadUtil(object):
     health_service_cmds = Enum("health_service_cmds", "DISABLE ENABLE")
     ha_service_cmds = Enum("ha_service_cmds", "DISABLE ENABLE")
     config_ipv6_iface_cmds = Enum("config_ipv6_iface_cmds", "DISABLE ENABLE")
+
+    _init_system = None
 
     def __init__(self, magmad_client):
         """
@@ -848,6 +855,9 @@ class MagmadUtil(object):
             "{user}@{host} {command}"
         )
 
+        if self._init_system is None:
+            self._init_system = self.detect_init_system()
+
     def exec_command(self, command):
         """Run a command remotely on magma_dev VM.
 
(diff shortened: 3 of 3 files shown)


lte/gateway/docker/README.md
```diff
@@ -51,3 +51,15 @@ on a VM. However we are not there yet as the containerized AGW currently depends
...
 to have the right packages installed, and thus can currently be used as a quick
 and dirty way to run the containers locally.
+
+### Running the S1AP integration tests against the containerized AGW
+
+At present, of the S1AP integration tests, it has been verified that `test_attach_detach` can be run versus the containerized AGW.
+Before running this test, one should execute the following command to set a required environment variable:
+
+```
+sed -i 's/MAGMA_DEV_MODE=0/MAGMA_DEV_MODE=1/g' $MAGMA_ROOT/lte/gateway/docker/.env
```
nstng: Is this possible after the images are created, i.e., during runtime?

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -868,6 +878,21 @@ def exec_command(self, command):
             stderr=subprocess.DEVNULL,
         )
 
+    def detect_init_system(self):
+        """Detect whether services are running with Docker or systemd."""
+
+        docker_running = "magmad" in self.exec_command_output("docker ps --filter 'name=magmad' --format '{{.Names}}'")
```
nstng: is this robust against environments where docker/systemd is not installed?

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -935,9 +960,13 @@ def corrupt_agw_state(self, key: str):
...
         """Restart all magma services on magma_dev VM"""
-        self.exec_command(
-            "sudo service magma@* stop ; sudo service magma@magmad start",
-        )
+        if self._init_system == InitMode.SYSTEMD:
+            self.exec_command(
+                "sudo service magma@* stop ; sudo service magma@magmad start",
+            )
+        elif self._init_system == InitMode.DOCKER:
+            self.exec_command("cd /home/vagrant/magma/lte/gateway/docker")
```
nstng: I am surprised that this works, I would expect this to translate to two ssh commands. How does the second command know about the `cd`?

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -1196,7 +1225,10 @@ def restart_sctpd(self):
         """
         Restart sctpd service explicitly because it is not managed by magmad
         """
-        self.exec_command("sudo service sctpd restart")
+        if self._init_system == InitMode.SYSTEMD:
+            self.exec_command("sudo service sctpd restart")
+        elif self._init_system == InitMode.DOCKER:
+            self.exec_command("sudo docker restart sctpd")
```
nstng: why is `sudo` needed here and not for the `docker-compose` restart?

lte/gateway/docker/.env
```diff
@@ -28,3 +28,6 @@ CONFIGS_DEFAULT_VOLUME=../configs
 CONFIGS_TEMPLATES_PATH=../configs/templates
 
 LOG_DRIVER=journald
+
+# Change to 1 to run integration tests
+MAGMA_DEV_MODE=0
```
nstng: iiuc this is now defined as an environment variable on the host and in this file - is there a better way to handle this?

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -868,6 +878,21 @@ def exec_command(self, command):
...
         )
 
+    def detect_init_system(self):
+        """Detect whether services are running with Docker or systemd."""
+
+        docker_running = "magmad" in self.exec_command_output("docker ps --filter 'name=magmad' --format '{{.Names}}'")
+        systemd_running = self.exec_command("systemctl is-active magma@magmad") == 0
+
+        if docker_running and systemd_running:
+            raise AssertionError("Magmad is running with both Docker and systemd.")
```
jheidbrink: nitpick: I think `AssertionError` is really just meant to be raised by `assert`. When manually raising an exception there is usually something more specific that fits better. That said, I'm not sure what would be a good one here ;)

lte/gateway/docker/README.md
voisey: We tried various things but nothing seemed to work. For example, you can parse an environment variable as part of a `docker compose` command, but not as part of the deprecated `docker-compose` command we're using. We thought we might try to migrate this to an override file, for example, but seeing as this was just an attempt to get the first test running, we thought this might be okay for the moment.

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -868,6 +878,21 @@ def exec_command(self, command):
             stderr=subprocess.DEVNULL,
         )
 
+    def detect_init_system(self):
```
sebathomas:
```suggestion
    def detect_init_system(self) -> InitMode:
```
I think it's always good to document the return and input types of new functions.

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
sebathomas: I think people often use `ValueError` for generic precondition errors like here even if it is meant for something more specific (strictly-speaking). Another option would be a generic `RuntimeError`.

jheidbrink: Not sure if this qualifies for `OSError`, otherwise a more general one would be `RuntimeError`

sebathomas: There's also an `EnvironmentError` that's an alias of `OSError` that might fit. On the other hand, `OSError` refers to I/O issues in its docstring and that doesn't really fit...

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
voisey: Good point, thanks! I've removed the `sudo` here.

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
voisey: Thanks for pointing this out and for the suggestions! To me `EnvironmentError` makes the most intuitive sense but it seems to have been deprecated, so I think I'll just go for `RuntimeError`.

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
voisey: Good point, ta! I've added that now.

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
voisey: You're right, thanks for spotting this! I've combined the two lines into one command.

lte/gateway/docker/.env
voisey: You're right. There may well be a better way. Do you have any thoughts on this @jheidbrink?

jheidbrink: https://docs.docker.com/compose/env-file/ reads "Values present in the environment at runtime always override those defined inside the .env file". That sounds like using the host variable should just work. Haven't tested it yet though.

jheidbrink: I tried it and it seems it doesn't work with `docker-compose`. The docs also mention `docker compose -e`, but again this does not work with `docker-compose`.

jheidbrink (approved):
I confirmed that I can run `test_attach_detach` on this branch and it succeeds.

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
voisey: Good point! This actually fails on `magma_deb` where Docker is not installed. How's something like this?
```
    def detect_init_system(self) -> InitMode:
        """Detect whether services are running with Docker or systemd."""

        docker_running = False
        systemd_running = False

        try:
            docker_running = "magmad" in self.exec_command_output("docker ps --filter 'name=magmad' --format '{{.Names}}'")
        except:
            logging.info("Docker is not installed")

        try:
            systemd_running = self.exec_command("systemctl is-active magma@magmad") == 0
        except:
            logging.info("systemd is not installed")
        ...
```
Btw, is it possible `systemctl` would not be installed but `systemd` would be?

jheidbrink: `systemctl` should be present when `systemd` is installed, so I think we can safely assume its presence. In the `except` we should be specific which exceptions we want to catch (probably `CalledProcessError`).

jheidbrink: Personally I would not initialize the `_running` variables, but add a `docker_running = False` in the except block, but that's a bit subjective, feel free to do it the way you prefer.

lte/gateway/docker/README.md
voisey: I realise you were asking a difference question. I've updates the docs with the sentence below stating that you need to restart the containers.

sebathomas (approved):

lte/gateway/docker/docker-compose.yaml
```diff
@@ -25,6 +25,7 @@ x-generic-service: &service
   logging: *logging_anchor
   restart: always
   network_mode: host
+  env_file: .env
```
nstng: might this line be redundant? is .env used by default?

nstng (approved):
lgtm (modulo minor comment)

lte/gateway/docker/docker-compose.yaml
voisey: The line was required to set the added environment variable in the containers, however the paths in `.env` were already being used in `docker-compose.yaml` as default. That said, this modification wasn't ideal so it has been replaced by the use of an override file, `docker-compose.dev.yaml`.

lte/gateway/docker/docker-compose.dev.yaml
```diff
@@ -0,0 +1,102 @@
+---
```
crasu, line 1: Instead of using this file you can add MAGMA_DEV_MODE (without any values) to the .env file. This will pass the value of the env variable from the host system to the container.

crasu: Ok this will not work as it does not propagate the variable into the container

sebathomas:
Could someone from @magma/approvers-agw-integtests have a look at this PR? It's the first step for integration tests of the dockerized AGW.

lte/gateway/python/integ_tests/s1aptests/s1ap_utils.py
```diff
@@ -935,9 +969,12 @@ def corrupt_agw_state(self, key: str):
 
     def restart_all_services(self):
         """Restart all magma services on magma_dev VM"""
-        self.exec_command(
-            "sudo service magma@* stop ; sudo service magma@magmad start",
-        )
+        if self._init_system == InitMode.SYSTEMD:
+            self.exec_command(
+                "sudo service magma@* stop ; sudo service magma@magmad start",
```
rsarwad, line 974: Hi @sebathomas here it looks good as all services have restarted. Do we need similar handling for test case where an individual service have started. Say for test case, test_sctp_shutdown_while_mme_is_stopped.py mme, mobilityd, pipelined services have stopped and started mme service.
Please find the test case here, https://github.com/magma/magma/blob/master/lte/gateway/python/integ_tests/s1aptests/test_sctp_shutdown_while_mme_is_stopped.py

sebathomas: Thanks for looking into this. Yes, I expect more changes will be necessary to run all tests, this will be handled in the follow-up ticket #13684.

rsarwad (commented):
S1ap tester changes looks good to me

sebathomas:
> S1ap tester changes looks good to me

Can you approve the PR then? :)

rsarwad (approved):
s1ap changes LGTM

sebathomas:
Thank you! I think the PR needs a rebase now because of changes to the semantic-pr CI, but apart from that it's ready to merge.

rsarwad:
> Thank you! I think the PR needs a rebase now because of changes to the semantic-pr CI, but apart from that it's ready to merge.
Yes

PR #13675, opened 2022-08-19
chore: bump postcss

lucasgonze:
## Summary

Fix Regular Expression Denial of Service in postcss by bumping all transitive paths to postcss.

There appear to be few interdependent facets of these upgrades, so little risk in upgrading to the latest versions.

## Test Plan

`cd nms; yarn test`

`./e2e_test_setup.sh`

## Additional Information


lucasgonze:
I could use feedback on how to test this. I think it's low risk, but it would be good to discuss webpack with a front-end dev.

lucasgonze:
Work log:

- testing with `cd /nms; yarn run test` shows no new test failures
- asked for advice in #nms-dev. No replies yet.

Neudrino:
@lucasgonze wrote in https://magmacore.slack.com/archives/C03C1C3KP44/p1661460386387119
> I’m unsure how to test this PR
> This affects:
> - css-loader, which is used in webpack.
> - autoprefixer,
>
> My best guess about a failure of this component is that a failure would cause CSS  in docusaurus to not load.
> Do our automated tests cover CSS, or do we do that kind of verification with functional - IE hands-on - testing?
>
> The only part of our code where this is used explicitly is nms//config/webpack.config.ts. Where does a failure in the webpack config file manifest?

Neudrino:
- DocuSaurus is not affected by this change, only NMS.
- Running `yarn test` was already a good start. :+1: 
- Additionally, also run the e2e tests for NMS. (https://docs.magmacore.org/docs/nms/dev_testing#test-nms)
- To check that the CSS is applied correctly, you can run an NMS instance and have a look, that it is looking normal.
  - https://docs.magmacore.org/docs/basics/prerequisites
  - https://docs.magmacore.org/docs/basics/quick_start_guide
  - You do not need to AGW parts here, running orc8r and NMS will be enough.

Btw. to me, the section `Test Plan` is dedicated to everything, which was undertaken or will be done to test and validate the changes made. So you could add your ideas and the additionally requested actions there. :slightly_smiling_face:

lucasgonze:
`./e2e_test_setup.sh` passes

sebathomas (approved):

PR #13688, opened 2022-08-22
feat(ci): deactivate useless linter warnings for yaml document start

Neudrino:
The document start is really a big pain and only bloats the files. This the yaml linter for it is deactivated with this commit.

This PR is required for a PR coming to add copyright headers (https://github.com/magma/magma/issues/13280). The PR will be opened if this one is merged, to avoid the heaps of annotations, which would be generated otherwise.

### DIFF
--- a/.github/workflows/config/yamllint_config.yml
+++ b/.github/workflows/config/yamllint_config.yml
@@ -12,7 +12,7 @@ ignore: |
   **/templates/**/*.yaml
 
 rules:
-  braces: 
+  braces:
     max-spaces-inside: 1
     min-spaces-inside: 1
   brackets:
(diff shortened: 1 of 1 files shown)


LKreutzer (approved):
Lgtm. 

- Do you plan to re-enable the check or is it permanently turned off?
  - If it is permanent then perhaps we should clean up the document start markers and make them uniform in all files as a follow up task.

Neudrino:
>       * If it is permanent then perhaps we should clean up the document start markers and make them uniform in all files as a follow up task.

Yes. This was already prepared with https://github.com/magma/magma/pull/13696, but could not be opened, until the first one was merged. Otherwise it would have been swamped with linter warnings. :wink:

PR #13701, opened 2022-08-23
test(orc8r): Wait in flaky subscriberdb test

sebathomas:
## Summary

Previously the test setup was modified to make it less flaky. Unfortunately, it is still sometimes red so it might be useful to add an explicit waiting step.

See #13565 for more context.

Still happens with a 1 second sleep statement so this is probably not a good solution.
https://github.com/magma/magma/runs/7998325221

## Test Plan

Run repeatedly in the CI.

## Additional Information


sebathomas:
Still happens with a 1 second sleep statement so this is probably not a good solution.
https://github.com/magma/magma/runs/7998325221

